PR #97, opened 2020-05-13
[Csv] add tests for csv dataset script

patrickvonplaten:
Adds dummy data tests for csv.


datasets/csv/csv.py
```diff
@@ -11,10 +10,11 @@
 @dataclass
 class CsvConfig(nlp.BuilderConfig):
     """BuilderConfig for CSV."""
+
     skip_rows: int = 0
     header_as_column_names: bool = True
     delimiter: str = ","
-    quote_char: str = "\""
+    quote_char: str = '"'
```
patrickvonplaten, line 17: @thomwolf - black did this automatically - is that ok?

datasets/csv/csv.py
```diff
@@ -2,7 +2,6 @@
 
 from dataclasses import dataclass
 
-import pyarrow as pa
```
patrickvonplaten, line 5: Deleted this import since it doesn't seem to be in use @thomwolf

tests/test_dataset_common.py
```diff
@@ -91,7 +92,16 @@ def check_load_dataset(self, dataset_name, configs, is_local=False):
                     is_local=is_local,
                 )
 
-                # build dataset from dummy data
+                if dataset_builder.__class__.__name__ == "Csv":
```
patrickvonplaten, line 95: @thomwolf - a bit hacky here, but I think it's ok for now

patrickvonplaten:
@thomwolf - can you check and merge if ok?

datasets/csv/csv.py
thomwolf: Yes!

datasets/csv/csv.py
thomwolf: Ok!

tests/test_dataset_common.py
thomwolf: Ok! I trust you with the tests :-)

PR #98, opened 2020-05-14
Webis tl-dr

jplu:
Add the Webid TL:DR dataset.


julien-c:
Should that rather be in an organization scope, @thomwolf @patrickvonplaten ?

patrickvonplaten:
> Should that rather be in an organization scope, @thomwolf @patrickvonplaten ?

I'm a bit indifferent - both would be fine for me!

patrickvonplaten:
@jplu - if creating the dummy_data is too tedious, I can do it as well :-)

jplu:
There is dummy_data here, no ?

julien-c:
Yeah I think naming it webis/tl_dr would be best @jplu if that works for you

jplu:
No problem at all!! On it^^

patrickvonplaten:
> There is dummy_data here, no ?

Some path were wrong - the structure is really confusing and the error message don't really help either - I have to think about how to make this easier to understand!

jplu:
> Some paths were wrong - the structure is really confusing and the error message don't really help either - I have to think about how to make this easier to understand!

Oh ok! I haven't noticed that sorry :(

> Hope it was ok that I fiddled with your PR !

Of course not :)

jplu:
@julien-c Looks like what you have in mind?

```python
import nlp
nlp.load_dataset("datasets/webis", "tl_dr")

#Output: Downloading and preparing dataset webis/tl_dr (download: Unknown size, generated: Unknown size, total: Unknown size) to /home/jplu/.cache/huggingface/datasets/webis/tl_dr/1.0.0...
```

patrickvonplaten:
Merging this for now. Maybe we can see whether to rename it in a different PR @julien-c ?

shahbazsyed:
Hi, 
Author here of the webis-tldr corpus. Any plans on integrating this dataset into the hub? I remember we could access it in the previous versions of the library. If there is a particular issue that I can help with, do let me know.

Thanks!

julien-c:
Hi @shahbazsyed, this dataset _is_ inside the hub but it's namespaced by the organization name `webis`.

You can load it following the steps described in https://huggingface.co/datasets/webis/tl_dr

Here's a Colab showcasing that it works: https://colab.research.google.com/drive/11IrzRVpnMLJZ8_UFFHLR8FhiajjAHRUU?usp=sharing

The reason the code is in S3 and not in this repo is that the dataset is namespaced under the `webis` organization. We don't have a lot of namespaced datasets yet but this should become the main way we add more datasets in the future.
Let us know if that's an issue for you. Thank you!

PR #102, opened 2020-05-14
Run save infos

lhoestq:
I replaced the old checksum file with the new `dataset_infos.json` by running the script on almost all the datasets we have. The only one that is still running on my side is the cornell dialog


patrickvonplaten:
Haha that cornell dialogue dataset - that ran for 3h on my computer as well. The `generate_examples` method in this script is one of the most inefficient code samples I've ever seen :D

lhoestq:
Indeed it's been 3 hours already
```73111 examples [3:07:48,  2.40 examples/s]```

PR #103, opened 2020-05-14
[Manual downloads] add logic proposal for manual downloads and add wikihow

patrickvonplaten:
Wikihow is an example that needs to manually download two files as stated in: https://github.com/mahnazkoupaee/WikiHow-Dataset. 

The user can then store these files under a hard-coded name: `wikihowAll.csv` and `wikihowSep.csv` in this case in a directory of his choice, e.g. `~/wikihow/manual_dir`.

The dataset can then be loaded via:

```python
import nlp
nlp.load_dataset("wikihow", data_dir="~/wikihow/manual_dir")
```

I added/changed so that there are explicit error messages when using manually downloaded files.


datasets/wikihow/wikihow.py
```diff
@@ -0,0 +1,191 @@
...
+        """Returns SplitGenerators."""
+        dl_path = dl_manager.download_and_extract(_URLS)
+        titles = {k: set() for k in dl_path}
+        for k, path in dl_path.items():
+            with open(path) as f:
+                for line in f:
+                    titles[k].add(line.strip())
+
+        path_to_manual_file = os.path.join(
+            os.path.abspath(os.path.expanduser(dl_manager.manual_dir)), self.config.filename
```
patrickvonplaten, line 132: The manual_dir can be accessed via `dl_manager.manual_dir`

datasets/wikihow/wikihow.py
```diff
@@ -0,0 +1,191 @@
...
+        # Version 1.1.0 remove empty document and summary strings.
+        # Version 1.2.0 add train validation test split, add cleaning & filtering.
+        super(WikihowConfig, self).__init__(version=nlp.Version("1.2.0"), **kwargs)
+        self.filename = filename
+
+
+class Wikihow(nlp.GeneratorBasedBuilder):
+    """WikiHow: A Large Scale Text Summarization Dataset."""
+
+    MANUAL_DOWNLOAD_INSTRUCTIONS = """\
```
patrickvonplaten, line 89: made the instruction a tad clearer

src/nlp/builder.py
```diff
@@ -344,7 +348,9 @@ def download_and_prepare(
             download_config.cache_dir = os.path.join(self._cache_dir_root, "downloads")
             download_config.force_download = download_mode == FORCE_REDOWNLOAD
 
-            dl_manager = DownloadManager(dataset_name=self.name, download_config=download_config,)
+            dl_manager = DownloadManager(
+                dataset_name=self.name, download_config=download_config, data_dir=self.config.data_dir
```
patrickvonplaten, line 352: @thomwolf - the DownoadManager has a nice property called `manual_dir` which refers to the `data_dir` the `data_dir` of the DownloadManager was previously never set to any value so I change that here and set it equal to `self.config.data_dir`.

mariamabarham:
> Wikihow is an example that needs to manually download two files as stated in: https://github.com/mahnazkoupaee/WikiHow-Dataset.
> 
> The user can then store these files under a hard-coded name: `wikihowAll.csv` and `wikihowSep.csv` in this case in a directory of his choice, e.g. `~/wikihow/manual_dir`.
> 
> The dataset can then be loaded via:
> 
> ```python
> import nlp
> nlp.load_dataset("wikihow", data_dir="~/wikihow/manual_dir")
> ```
> 
> I added/changed so that there are explicit error messages when using manually downloaded files.

wouldn't be nicer if we can have `manual_dir/wikihow`?

patrickvonplaten:
> > Wikihow is an example that needs to manually download two files as stated in: https://github.com/mahnazkoupaee/WikiHow-Dataset.
> > The user can then store these files under a hard-coded name: `wikihowAll.csv` and `wikihowSep.csv` in this case in a directory of his choice, e.g. `~/wikihow/manual_dir`.
> > The dataset can then be loaded via:
> > ```python
> > import nlp
> > nlp.load_dataset("wikihow", data_dir="~/wikihow/manual_dir")
> > ```
> > 
> > 
> > I added/changed so that there are explicit error messages when using manually downloaded files.
> 
> wouldn't be nicer if we can have `manual_dir/wikihow`?

Sure, I mean the user can decide whatever he likes best :-) The path one puts in `data_dir` will be used as the path to the manual dir. `nlp.load_dataset("wikihow", data_dir="~/manual_dir/wikihow")` would work as well as any other path ;-)

thomwolf:
Perfect! You can merge!

PR #106, opened 2020-05-14
Add data dir test command

lhoestq:


patrickvonplaten:
Nice - I think we can merge this. I will update the checksums for `wikihow` then as well

PR #107, opened 2020-05-14
add writer_batch_size to GeneratorBasedBuilder

lhoestq:
You can now specify `writer_batch_size` in the builder arguments or directly in `load_dataset`

### DIFF
--- a/src/nlp/builder.py
+++ b/src/nlp/builder.py
@@ -556,6 +556,10 @@ class GeneratorBasedBuilder(DatasetBuilder):
     (`_split_generators`). See the method docstrings for details.
     """
 
+    def __init__(self, *args, **kwargs):
+        super(GeneratorBasedBuilder, self).__init__(*args, **kwargs)
+        self._writer_batch_size = kwargs.get("writer_batch_size")
+
     @abc.abstractmethod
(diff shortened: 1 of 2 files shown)


patrickvonplaten:
Awesome that's great!

src/nlp/builder.py
```diff
@@ -556,6 +556,10 @@ class GeneratorBasedBuilder(DatasetBuilder):
     (`_split_generators`). See the method docstrings for details.
     """
 
+    def __init__(self, *args, **kwargs):
+        super(GeneratorBasedBuilder, self).__init__(*args, **kwargs)
+        self._writer_batch_size = kwargs.get("writer_batch_size")
```
patrickvonplaten, line 561: maybe it's nice to use `DEFAULT_MAX_BATCH_SIZE` here as a default for the `.get()` function,  but both ways are fine with me!

lhoestq: I think this parameter is more linked to the implementation of the `ArrowWriter`, that's why I think it's better if the default value is set in the writer itself

patrickvonplaten: Fair!

PR #112, opened 2020-05-14
Qa4mre - add dataset

patrickvonplaten:
Added dummy data test only for the first config. Will do the rest later.
I had to do add some minor hacks to an important function to make it work. 
There might be a cleaner way to handle it - can you take a look @thomwolf ?


src/nlp/features.py
```diff
@@ -400,10 +400,17 @@ def encode_nested_example(schema, obj):
     elif isinstance(schema, Sequence):
         # We allow to reverse list of dict => dict of list for compatiblity with tfds
         if isinstance(schema.feature, dict):
-            return dict(
-                (k, [encode_nested_example(sub_schema, o) for o in sub_obj])
-                for k, (sub_schema, sub_obj) in utils.zip_dict(schema.feature, obj)
-            )
+            if isinstance(obj, (list, tuple)):
```
patrickvonplaten, line 403: @thomwolf - would be nice if you can take a look here. 
qa4mre is a very special case and I had to make some changes here to make the dataset work. For `qa4mre`, `obj` is a list of `dicts` and so need to be treaded differently.

src/nlp/features.py
```diff
@@ -400,10 +400,17 @@ def encode_nested_example(schema, obj):
...
+                zipped_dict = utils.zip_dict(schema.feature, *obj)
+            else:
+                zipped_dict = utils.zip_dict(schema.feature, obj)
+            list_dict = {}
+            for k, dict_tuples in zipped_dict:
+                sub_schema = dict_tuples[0]
+                sub_objs = dict_tuples[1:] if len(dict_tuples) > 2 else dict_tuples[1]
```
patrickvonplaten, line 410: need this if statement here to be sure all previous datasets work exactly the same as before

thomwolf: Ok but won't this break if `obj` is a list (ie. the special case of qa4mre) but with only one element (i.e. probably `len(dict_tuples)`== 2)?
Maybe the test here should be again `if isinstance(obj, (list, tuple))`?

patrickvonplaten: True! There was currently no case of this, but could be - I think the logic as it is now, is better.

src/nlp/features.py
```diff
@@ -400,10 +400,19 @@ def encode_nested_example(schema, obj):
...
-            )
+            # dict of list to fill
+            list_dict = {}
+            if isinstance(obj, (list, tuple)):
+                # obj is a list of dict
+                for k, dict_tuples in utils.zip_dict(schema.feature, *obj):
+                    list_dict[k] = [encode_nested_example(dict_tuples[0], o) for o in dict_tuples[1:]]
+                return list_dict
+            else:
+                # obj is a single dict
```
patrickvonplaten, line 411: Previous case

src/nlp/features.py
```diff
@@ -400,10 +400,19 @@ def encode_nested_example(schema, obj):
...
-                (k, [encode_nested_example(sub_schema, o) for o in sub_obj])
-                for k, (sub_schema, sub_obj) in utils.zip_dict(schema.feature, obj)
-            )
+            # dict of list to fill
+            list_dict = {}
+            if isinstance(obj, (list, tuple)):
+                # obj is a list of dict
+                for k, dict_tuples in utils.zip_dict(schema.feature, *obj):
+                    list_dict[k] = [encode_nested_example(dict_tuples[0], o) for o in dict_tuples[1:]]
+                return list_dict
```
patrickvonplaten, line 409: The new special case @thomwolf

src/nlp/features.py
```diff
@@ -400,10 +400,19 @@ def encode_nested_example(schema, obj):
...
+            else:
+                # obj is a single dict
+                for k, (sub_schema, sub_objs) in utils.zip_dict(schema.feature, obj):
+                    list_dict[k] = [encode_nested_example(sub_schema, o) for o in sub_objs]
+                return list_dict
```
patrickvonplaten, line 414: previous case

PR #122, opened 2020-05-15
Final cleanup of readme and metrics

thomwolf:


metrics/gleu/gleu.py
```diff
@@ -78,8 +78,10 @@ def get_gleu_stats(scores) :
             'ci': ci}
 
 class Gleu(nlp.Metric):
+    def __init__(self, **kwargs):
+        raise NotImplementedError("Gleu is currently under construction.")
```
Ayushk4, line 82: Hi. Are there any plans to include GLEU into this library? Are there any other codebases that you may recommend?

PR #124, opened 2020-05-15
Xsum, require manual download of some files

mariamabarham:


tests/utils.py
```diff
@@ -98,7 +98,11 @@ def dummy_file(self):
 
     @property
     def manual_dir(self):
-        return self.dummy_file
+        # return full path if its a dir
```
patrickvonplaten, line 101: @mariamabarham - need to handle the special case when the downloaded path is not extracted to a dir, but just a single file

PR #135, opened 2020-05-15
Fix print statement in READ.md

codehunk628:
print statement was throwing generator object instead of printing names of available datasets/metrics


thomwolf:
Indeed, thanks!

PR #140, opened 2020-05-16
[Tests] run local tests as default

patrickvonplaten:
This PR also enables local tests by default

I think it's safer for now to enable both local and aws tests for every commit. The problem currently is that when we do a PR to add a dataset, the dataset is not yet on AWS on therefore not tested on the PR itself. Thus the PR will always be green even if the datasets are not correct. This PR aims at fixing this.

## Suggestion on how to commit to the repo from now on:
Now since the repo is "online", I think we should adopt a couple of best practices:
1) - No direct committing to the repo anymore. Every change should be opened in a PR and be well documented so that we can find it later
2) - Every PR has to be reviewed by at least x people (I guess @thomwolf you should decide here) because we now have to be much more careful when doing changes to the API for backward compatibility, etc...

### DIFF
--- a/datasets/kor_nli/kor_nli.py
+++ b/datasets/kor_nli/kor_nli.py
@@ -1,12 +1,12 @@
 """TODO(kor_nli): Add a description here."""
 
-from __future__ import absolute_import
-from __future__ import division
-from __future__ import print_function
+from __future__ import absolute_import, division, print_function
+
+import csv
+import os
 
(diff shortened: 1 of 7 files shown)


jplu:
You are right and I think those are usual best practice :) I'm 100% fine with this^^

patrickvonplaten:
Merging this for now to unblock other PRs.

PR #146, opened 2020-05-16
Add BERTScore to metrics

felixgwu:
This PR adds [BERTScore](https://arxiv.org/abs/1904.09675) to metrics.
Here is an example of how to use it.

```sh
import nlp
bertscore = nlp.load_metric('metrics/bertscore') # or simply nlp.load_metric('bertscore') after this is added to huggingface's s3 bucket
predictions = ['example', 'fruit']
references = [['this is an example.', 'this is one example.'], ['apple']]
results = bertscore.compute(predictions, references, lang='en')
print(results)
```

### DIFF
--- a/metrics/bertscore/bertscore.py
+++ b/metrics/bertscore/bertscore.py
@@ -15,6 +15,7 @@
 """ BERTScore metric. """
 
 import nlp
+import bert_score
 
 _CITATION = """\
 @inproceedings{bert-score,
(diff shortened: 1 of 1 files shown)


metrics/bertscore/bertscore.py
```diff
@@ -0,0 +1,131 @@
...
+        device=None,
+        batch_size=64,
+        nthreads=4,
+        all_layers=False,
+        rescale_with_baseline=False,
+    ):
+        try:
+            import bert_score
+        except ImportError:
+            raise ImportError("Please install bert_score using 'pip install bert-score'.")
```
thomwolf, line 99: You can have this import at the top of the script like a normal import.

When loading the script we are actually scanning the imports, testing them, and display an error message similar to the one you are throwing here if an import is not available (see [here](https://github.com/huggingface/nlp/blob/master/src/nlp/load.py#L284-L295))

felixgwu: Awesome! I didn't notice that the ImportError has been handled in such a nice way.
I've clean it up in the new commit. I can help with adding the doc regarding BERTScore once some examples are added in the future.

Please feel free to let me know if there is something that needs to be changed.

PR #167, opened 2020-05-19
[Tests] refactor tests

patrickvonplaten:
This PR separates AWS and Local tests to remove these ugly statements in the script:
```python
        if "/" not in dataset_name:
            logging.info("Skip {} because it is a canonical dataset")
            return
```

To run a `aws` test, one should now run the following command: 

```python 
pytest -s tests/test_dataset_common.py::AWSDatasetTest::test_builder_class_wmt14
```

The same `local` test, can be run with:
```python 
pytest -s tests/test_dataset_common.py::LocalDatasetTest::test_builder_class_wmt14
```

### DIFF
--- a/datasets/ubuntu_dialogs_corpus/ubuntu_dialogs_corpus.py
+++ b/datasets/ubuntu_dialogs_corpus/ubuntu_dialogs_corpus.py
@@ -1,12 +1,12 @@
 """TODO(ubuntu_dialogs_corpus): Add a description here."""
 
-from __future__ import absolute_import
-from __future__ import division
-from __future__ import print_function
+from __future__ import absolute_import, division, print_function
 
-import nlp
 import csv
 import os
+
+import nlp
+
 
 # TODO(ubuntu_dialogs_corpus): BibTeX citation
 _CITATION = """\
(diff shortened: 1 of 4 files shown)


lhoestq:
Nice !

datasets/ubuntu_dialogs_corpus/ubuntu_dialogs_corpus.py
```diff
@@ -1,13 +1,13 @@
 """TODO(ubuntu_dialogs_corpus): Add a description here."""
 
-from __future__ import absolute_import
-from __future__ import division
-from __future__ import print_function
+from __future__ import absolute_import, division, print_function
```
lhoestq, line 3: is there a problem with black smh ?

patrickvonplaten: I don't know whether that's `black` or `isort`, but I don't think it's a problem. 

It's just that we don't check quality on the `datasets` folder at the moment - see the .circle_ci folder, but I added `./datasets` to the `Make style` file. So I think it's just that some people upload datasets without formatting them and I correct it :D

patrickvonplaten: IMO, we should add `./datasets` to the circle ci quality check!

Issue #172, opened 2020-05-20
Clone not working on Windows environment


codehunk628:
Cloning in a windows environment is not working because of use of special character '?' in folder name ..
Please consider changing the folder name ....
Reference to folder -
nlp/datasets/cnn_dailymail/dummy/3.0.0/3.0.0/dummy_data-zip-extracted/dummy_data/uc?export=download&id=0BwmD_VLjROrfM1BxdkxVaTY2bWs/dailymail/stories/

error log:
fatal: cannot create directory at 'datasets/cnn_dailymail/dummy/3.0.0/3.0.0/dummy_data-zip-extracted/dummy_data/uc?export=download&id=0BwmD_VLjROrfM1BxdkxVaTY2bWs': Invalid argument

lhoestq:
Should be fixed on master now :)

codehunk628:
Thanks @lhoestq 👍  Now I can uninstall WSL and get back to work with windows.🙂

PR #173, opened 2020-05-20
Rm extracted test dirs

lhoestq:
All the dummy data used for tests were duplicated. For each dataset, we had one zip file but also its extracted directory. I removed all these directories

Furthermore instead of extracting next to the dummy_data.zip file, we extract in the temp `cached_dir` used for tests, so that all the extracted directories get removed after testing.

Finally there was a bug in the `mock_download_manager` that would let it create directories with invalid names, as in #172. I fixed that by encoding url arguments. I had to rename the dummy data for `scientific_papers` and `cnn_dailymail` (the aws tests don't pass for those 2 in this PR, but they will once aws will be synced, as the local ones do)

Let me know if it sounds good to you @patrickvonplaten . I'm still not entirely familiar with the mock downloader


src/nlp/utils/file_utils.py
```diff
@@ -216,7 +215,7 @@ def cached_path(url_or_filename, download_config=None, **download_kwargs,) -> Op
         # We avoid '.' in dir name and add "-extracted" at the end: "./model.zip" => "./model-zip-extracted/"
         output_dir, output_file = os.path.split(output_path)
         output_extract_dir_name = output_file.replace(".", "-") + "-extracted"
-        output_path_extracted = os.path.join(output_dir, output_extract_dir_name)
+        output_path_extracted = os.path.join(cache_dir, output_extract_dir_name)
```
patrickvonplaten, line 218: So `output_dir` is not needed anymore? Is that safe in a sense that `output_path_extracted` is unique for every dataset if we don't use the `output_dir`? Maybe safer to do `output_path_extracted = os.path.join(cache_dir, output_dir, output_extract_dir_name)`

And are we replacing the `output_dir` here only because of the tests? We could also just put `dummy_data-extracted` in `.gitignore`...

tests/utils.py
```diff
@@ -175,10 +176,11 @@ def create_dummy_data_dict(self, path_to_dummy_data, data_url):
         dummy_data_dict = {}
         for key, abs_path in data_url.items():
             # we force the name of each key to be the last file / folder name of the url path
+            # if the url has arguments, we need to encode them with urllib.parse.quote_plus
             if isinstance(abs_path, list):
-                value = [os.path.join(path_to_dummy_data, x.split("/")[-1]) for x in abs_path]
+                value = [os.path.join(path_to_dummy_data, urllib.parse.quote_plus(x.split("/")[-1])) for x in abs_path]
```
patrickvonplaten, line 181: I guess ok for me...Over the course of adding all these datasets we have been adding more and more fixes to the MockDlManager to make it work in all cases to a point where it is now quite hard for a user to correctly name the `dummy_data` folder structure. 

One idea would be to write a `create_dumm_data_folder` command that does that automatically for the user. Or maybe we have to rethink the architecture / logic here a bit...

patrickvonplaten: What do you think @lhoestq ?

patrickvonplaten:
Thanks for cleaning up the extracted dummy data folders! Instead of changing the file_utils we could also just put these folders under `.gitignore` (or maybe already done?).

src/nlp/utils/file_utils.py
lhoestq: I see. There might be collisions in names. However i like the idea of extracting in the cache directory and not right next to the zipped file (in general, not only for tests).

tests/utils.py
lhoestq: This is probably what we will end up doing indeed

src/nlp/utils/file_utils.py
patrickvonplaten: Agree! Do you think we could go with `output_path_extracted = os.path.join(cache_dir, output_dir, output_extract_dir_name)` ?

Maybe @thomwolf can chime in here as well

lhoestq: `output_dir` can be inside many subfolders. We could re-use the `url_to_filename` to hash it like a url ? Something like `os.path.join(cache_dir, url_to_filename(output_path))` ?

src/nlp/utils/file_utils.py
```diff
@@ -213,10 +212,9 @@ def cached_path(url_or_filename, download_config=None, **download_kwargs,) -> Op
             return output_path
 
         # Path where we extract compressed archives
-        # We avoid '.' in dir name and add "-extracted" at the end: "./model.zip" => "./model-zip-extracted/"
-        output_dir, output_file = os.path.split(output_path)
-        output_extract_dir_name = output_file.replace(".", "-") + "-extracted"
-        output_path_extracted = os.path.join(output_dir, output_extract_dir_name)
+        # We extract in the cache dir, and get the extracted path name by hashing the original path"
+        abs_output_path = os.path.abspath(output_path)
+        output_path_extracted = os.path.join(cache_dir, hash_url_to_filename(abs_output_path))
```
lhoestq, line 217: This should fix the collision problem when extracting in the cache dir @patrickvonplaten

patrickvonplaten: looks great!

patrickvonplaten:
Awesome! I guess you might have to add the changes for the MockDLManager now in a different file though because of my last PR - sorry!

PR #176, opened 2020-05-20
[Tests] Refactor MockDownloadManager

patrickvonplaten:
Clean mock download manager class. 
The print function was not of much help I think. 
We should think about adding a command that creates the dummy folder structure for the user.


datasets/wiki40b/wiki40b.py
```diff
@@ -20,9 +20,9 @@
 import logging
 import os
```
patrickvonplaten, line 22: Now this looks like we have different versions of isort :-/

lhoestq: I missed this bug so much

1
```diff
@@ -0,0 +1,157 @@
+import os
```
lhoestq, line 1: I'm not sure this file is that the right place nor have the right name ^^ Looks like it's called `1` and is placed at the root

patrickvonplaten: uuups :D thanks :-)

PR #169, opened 2020-05-19
Adding Qanta (Quizbowl) Dataset

EntilZha:
This PR adds the qanta question answering datasets from [Quizbowl: The Case for Incremental Question Answering](https://arxiv.org/abs/1904.04792) and [Trick Me If You Can: Human-in-the-loop Generation of Adversarial Question Answering Examples](https://www.aclweb.org/anthology/Q19-1029/) (adversarial fold)

This partially continues a discussion around fixing dummy data from https://github.com/huggingface/nlp/issues/161

I ran the following code to double check that it works and did some sanity checks on the output. The majority of the code itself is from our `allennlp` version of the dataset reader.

```python
import nlp
# Default is full question
data = nlp.load_dataset('./datasets/qanta') 
# Four configs
# Primarily useful for training
data = nlp.load_dataset('./datasets/qanta', 'mode=sentences,char_skip=25')  
# Primarily used in evaluation
data = nlp.load_dataset('./datasets/qanta', 'mode=first,char_skip=25')  
data = nlp.load_dataset('./datasets/qanta', 'mode=full,char_skip=25')  
# Primarily useful in evaluation and "live" play
data = nlp.load_dataset('./datasets/qanta', 'mode=runs,char_skip=25')  
```

PR #185, opened 2020-05-22
[Commands] In-detail instructions to create dummy data folder

patrickvonplaten:
### Dummy data command 

This PR adds a new command `python nlp-cli dummy_data <path_to_dataset_folder>` that gives in-detail instructions on how to add the dummy data files. 

It would be great if you can try it out by moving the current dummy_data folder of any dataset in `./datasets` with `mv datasets/<dataset_script>/dummy_data datasets/<dataset_name>/dummy_data_copy` and running the command `python nlp-cli dummy_data ./datasets/<dataset_name>` to see if you like the instructions. 

### CONTRIBUTING.md
Also the CONTRIBUTING.md is made cleaner including a new section on "How to add a dataset". 

### Current PRs 
It would be nice if we can try out if this command helps current PRs, *e.g.* #169  to add a dataset. I comment on those PRs.


src/nlp/commands/dummy_data.py
```diff
@@ -0,0 +1,114 @@
...
+            dummy_file_name = mock_dl_manager.dummy_file_name
+
+            for split in generator_splits:
+                logger.info(f"Collecting dummy data file paths to create for {split.name}")
+                split_names.append(split.name)
+                gen_kwargs = split.gen_kwargs
+                generator = dataset_builder._generate_examples(**gen_kwargs)
+
+                try:
+                    dummy_data_guidance_print = "\n" + 30 * "=" + "DUMMY DATA INSTRUCTIONS" + 30 * "=" + "\n"
```
patrickvonplaten, line 73: The idea is here to actually run the `_generate_examples` function and catch the `FileNotFoundError`. It will only catch the first `FileNotFoundError`, but most of the times there is only one `open(...)` statement in the function.

src/nlp/utils/mock_download_manager.py
```diff
@@ -0,0 +1,136 @@
+# coding=utf-8
```
patrickvonplaten, line 1: move the `MockDownloadManager` to the actualy `src/nlp` code base since it is used in one of the commands now - is that fine for you @lhoestq @thomwolf ?

src/nlp/commands/dummy_data.py
```diff
@@ -0,0 +1,123 @@
...
+                        "- In order to create the dummy data for "
+                        + config_string
+                        + f"{self._dataset_name}, please go into the folder '{dummy_data_folder}' with `cd {dummy_data_folder}` . \n\n"
+                    )
+
+                    # trigger generate function
+                    for key, record in generator:
+                        pass
+
+                    dummy_data_guidance_print += f"- It appears that the function `_generate_examples(...)` expects one or more files in the folder {dummy_file_name} using the function `glob.glob(...)`. In this case, please refer to the `_generate_examples(...)` method to see under which filename the dummy data files should be created. \n\n"
```
lhoestq, line 85: Does arriving at this point always mean that there is a `glob.glob` ?

lhoestq:
awesome !

src/nlp/commands/dummy_data.py
patrickvonplaten: yeah actually I just checked the dataset scripts and this is quite an exception. I think there are only 2,3 datasets where this is releventa. One is `nlp/datasets/aeslc/aeslc.py`. 
I guess there could also be another command than `glob.glob()` that lists the files, but I think we always use `glob/glob` or @mariamabarham ?

PR #194, opened 2020-05-26
Add Dataset: Qanta

patrickvonplaten:
Fixes dummy data for #169 @EntilZha


patrickvonplaten:
@lhoestq - the config name is rather special here: *E.g.* `mode=first,char_skip=25`. It includes `=` and `,` - will that be a problem for windows folders, you think? 

Apart from that good to merge for me.

lhoestq:
It's ok to have `=` and `,`.
Windows doesn't like things like `?`, `:`, `/` etc.

I'll add some lines to raise an error if the config name is invalid.

EntilZha:
Thanks for fixing things up! I'm curious to take a look at the zip files now to know the format for future reference.

PR #195, opened 2020-05-26
[Dummy data command] add new case to command

patrickvonplaten:
Qanta: #194 introduces a case that was not noticed before. This change in code helps community users to have an easier time creating the dummy data.

### DIFF
--- a/src/nlp/commands/dummy_data.py
+++ b/src/nlp/commands/dummy_data.py
@@ -59,7 +59,7 @@ class DummyDataCommand(BaseTransformersCLICommand):
                     f"Dataset {self._dataset_name} with config {config} seems to already open files in the method `_split_generators(...)`. You might consider to instead only open files in the method `_generate_examples(...)` instead. If this is not possible the dummy data has to be created with less guidance. Make sure you create the file {e.filename}."
                 )
 
-            files_to_create = []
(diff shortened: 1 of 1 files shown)


patrickvonplaten:
@lhoestq - tiny change in the dummy data command, should be good to merge.

src/nlp/commands/dummy_data.py
```diff
@@ -90,8 +90,8 @@ def run(self):
             split_names = ", ".join(split_names)
             if len(files_to_create) > 0:
                 # no glob.glob(...) in `_generate_examples(...)`
-                if len(files_to_create) == 1 and files_to_create[0] == dummy_file_name:
-                    dummy_data_guidance_print += f"- Please create a single dummy data file called '{files_to_create[0]}' from the folder '{dummy_data_folder}'. Make sure that the dummy data file provides at least one example for the split '{split_names}' \n\n"
+                if len(set(files_to_create)) == 1 and files_to_create[0] == dummy_file_name:
```
lhoestq, line 93: maybe we could use a set directly for `files_to_create` ?

patrickvonplaten: yeah good point - changed it to use a set

PR #199, opened 2020-05-26
Fix GermEval 2014 dataset infos

stefan-it:
Hi,

this PR just removes the `dataset_info.json` file and adds a newly generated `dataset_infos.json` file.


stefan-it:
Hopefully. this also fixes https://huggingface.co/nlp/viewer/ then :)

lhoestq:
Oh good catch ! This should fix it indeed

PR #200, opened 2020-05-26
[ArrowWriter] Set schema at first write example

lhoestq:
Right now if the schema was not specified when instantiating `ArrowWriter`, then it could be set with the first `write_table` for example (it calls `self._build_writer()` to do so).

I noticed that it was not done if the first example is added via `.write`, so I added it for coherence.


thomwolf:
Good point!

I guess we could add this to `write_batch` as well (before using `self._schema` in the first line of this method)?

Issue #187, opened 2020-05-23
[Question] How to load wikipedia ? Beam runner ?


richardyy1188:
When `nlp.load_dataset('wikipedia')`, I got
* `WARNING:nlp.builder:Trying to generate a dataset using Apache Beam, yet no Beam Runner or PipelineOptions() has been provided. Please pass a nlp.DownloadConfig(beam_runner=...) object to the builder.download_and_prepare(download_config=...) method. Default values will be used.`
* `AttributeError: 'NoneType' object has no attribute 'size'`

Could somebody tell me what should I do ? 

# Env
On Colab,
```
git clone https://github.com/huggingface/nlp
cd nlp
pip install -q .
```
```
%pip install -q apache_beam mwparserfromhell
-> ERROR: pydrive 1.3.1 has requirement oauth2client>=4.0.0, but you'll have oauth2client 3.0.0 which is incompatible.
ERROR: google-api-python-client 1.7.12 has requirement httplib2<1dev,>=0.17.0, but you'll have httplib2 0.12.0 which is incompatible.
ERROR: chainer 6.5.0 has requirement typing-extensions<=3.6.6, but you'll have typing-extensions 3.7.4.2 which is incompatible.
```
```
pip install -q apache-beam[interactive]
ERROR: google-colab 1.0.0 has requirement ipython~=5.5.0, but you'll have ipython 5.10.0 which is incompatible.
```

# The whole message
```
WARNING:nlp.builder:Trying to generate a dataset using Apache Beam, yet no Beam Runner or PipelineOptions() has been provided. Please pass a nlp.DownloadConfig(beam_runner=...) object to the builder.download_and_prepare(download_config=...) method. Default values will be used.

Downloading and preparing dataset wikipedia/20200501.aa (download: Unknown size, generated: Unknown size, total: Unknown size) to /root/.cache/huggingface/datasets/wikipedia/20200501.aa/1.0.0...

---------------------------------------------------------------------------

AttributeError                            Traceback (most recent call last)

/usr/local/lib/python3.6/dist-packages/apache_beam/runners/common.cpython-36m-x86_64-linux-gnu.so in apache_beam.runners.common.DoFnRunner.process()

44 frames

/usr/local/lib/python3.6/dist-packages/apache_beam/runners/common.cpython-36m-x86_64-linux-gnu.so in apache_beam.runners.common.PerWindowInvoker.invoke_process()

[...]

Issue #129, opened 2020-05-15
[Feature request] Add Google Natural Question dataset


elyase:
Would be great to have https://github.com/google-research-datasets/natural-questions as an alternative to SQuAD.

thomwolf:
Indeed, I think this one is almost ready cc @lhoestq

lhoestq:
I'm doing the latest adjustments to make the processing of the dataset run on Dataflow

ronakice:
Is there an update to this? It will be very beneficial for the QA community!

lhoestq:
Still work in progress :)
The idea is to have the dataset already processed somewhere so that the user only have to download the processed files. I'm also doing it for wikipedia.

richardyy1188:
Super appreciate your hard work !!
I'll cross my fingers and hope easily loadable wikipedia dataset will come soon.

lhoestq:
Quick update on NQ: due to some limitations I met using apache beam + parquet I was not able to use the dataset in a nested parquet structure in python to convert it to our Apache Arrow format yet.
However we had planned to change this conversion step anyways so we'll make just sure that it enables to process and convert the NQ dataset to arrow.

lhoestq:
NQ was added in #427 🎉

PR #204, opened 2020-05-27
Add Dataflow support + Wikipedia + Wiki40b

lhoestq:
# Add Dataflow support + Wikipedia + Wiki40b

## Support datasets processing with Apache Beam

Some datasets are too big to be processed on a single machine, for example: wikipedia, wiki40b, etc. Apache Beam allows to process datasets on many execution engines like Dataflow, Spark, Flink, etc.

To process such datasets with Beam, I added a command to run beam pipelines `nlp-cli run_beam path/to/dataset/script`. Then I used it to process the english + french wikipedia, and the english of wiki40b.
The processed arrow files are on GCS and are the result of a Dataflow job.

I added a markdown documentation file in `docs` that explains how to use it properly.

## Load already processed datasets

Now that we have those datasets already processed, I made it possible to load datasets that are already processed. You can do `load_dataset('wikipedia', '20200501.en')` and it will download the processed files from the Hugging Face GCS directly into the user's cache and be ready to use !

The Wikipedia dataset was already asked in #187 and this PR should soon allow to add Natural Questions as asked in #129 

## Other changes in the code

To make things work, I had to do a few adjustments:
- add a `ship_files_with_pipeline` method to the `DownloadManager`. This is because beam pipelines can be run in the cloud and therefore need to have access to your downloaded data. I used it in the wikipedia script:
    ```python
    if not pipeline.is_local():
            downloaded_files = dl_manager.ship_files_with_pipeline(downloaded_files, pipeline)
    ```
- add parquet to arrow conversion. This is because the output of beam pipelines are parquet files so we need to convert them to arrow and have the arrow files on GCS
- add a test script with a dummy beam dataset
- minor adjustments to allow read/write operations on remote files using `apache_beam.io.filesystems.FileSystems` if we want (it can be connected to gcp, s3, hdfs, etc...)


datasets/wikipedia/wikipedia.py
```diff
@@ -441,6 +441,8 @@ def _base_url(lang):
 
             # Use dictionary since testing mock always returns the same result.
         downloaded_files = dl_manager.download({"xml": xml_urls})
+        if not pipeline.is_local():
+            downloaded_files = dl_manager.ship_files_with_pipeline(downloaded_files, pipeline)
```
lhoestq, line 445: Beam pipelines can't download files (and this is so sad). So we have to upload the files somewhere the pipeline have access. Here it ships the files to the temp dir used by the pipeline

datasets/wikipedia/wikipedia.py
```diff
@@ -454,7 +456,7 @@ def _build_pcollection(self, pipeline, filepaths, language):
         def _extract_content(filepath):
             """Extracts article content from a single WikiMedia XML file."""
             logging.info("generating examples from = %s", filepath)
-            with open(filepath, "rb") as f:
+            with beam.io.filesystems.FileSystems.open(filepath) as f:
```
lhoestq, line 459: use the beam filesystem in beam dataset scripts to access to remote files

src/nlp/arrow_reader.py
```diff
@@ -28,10 +29,13 @@
...
 from .arrow_dataset import Dataset
 from .naming import filename_for_dataset_split
-from .utils import py_utils
+from .utils import cached_path, py_utils
 
 
+logger = logging.getLogger(__name__)
+
 _BUFFER_SIZE = 8 << 20  # 8 MiB per file.
+HF_GCP_BASE_URL = "https://storage.googleapis.com/huggingface-nlp/cache/datasets"
```
lhoestq, line 38: I can still change the url if you want

src/nlp/arrow_writer.py
```diff
@@ -209,49 +221,74 @@ def __init__(
...
+        It converts the resulting parquet files to arrow and it completes the info from the pipeline metrics.
+
+        Args:
+            metrics_query_result: `dict` obtained from pipeline_results.metrics().query(m_filter). Make sure
+                that the filter keeps only the metrics for the considered split, under the namespace `split_name`.
+        """
+        import apache_beam as beam
+        from .utils import beam_utils
+
+        # Convert to arrow
```
lhoestq, line 260: Here os the conversion from parquet to arrow after each pipeline run

src/nlp/builder.py
```diff
@@ -743,39 +775,22 @@ def _build_pcollection(pipeline, extracted_dir):
         """
         raise NotImplementedError()
 
-    def _as_dataset(self, split: Split = Split.TRAIN):
```
lhoestq, line 746: We don't use the parquet reader anymore as we can't do memory mapping on parquet

src/nlp/commands/test.py
```diff
@@ -12,7 +12,8 @@
 def test_command_factory(args):
     return TestCommand(
         args.dataset,
-        args.config,
+        args.name,
```
lhoestq, line 15: I just updated the naming here, nothing special

src/nlp/arrow_writer.py
lhoestq: By default the conversion is done by streaming data like this:
parquet file on gsc -> local machine -> arrow file one gcs

However if you have a bad connection (as I do), it might happen that the socket connection fails (if you have a stabke connection it should happen). In this case I do the basic download locally + convert + upload...

lhoestq: I couldn't do it directly inside the Beam pipeline because writing to files is done asynchronously and there is no way in python to know when the parquet file writing has finished (apparently the Java sdk does but not python...)

lhoestq: I am thinking of using something else that the local machine to do it. We could have a separate beam pipeline just for conversion, and this pipeline could simply be run after the previous one have finished. Or we could use any other tool...

PR #208, opened 2020-05-28
[Dummy data] insert config name instead of config 

patrickvonplaten:
Thanks @yjernite for letting me know. in the dummy data command the config name shuold be passed to the dataset builder and not the config itself. 

Also, @lhoestq fixed small import bug introduced by beam command I think.

### DIFF
--- a/src/nlp/commands/run_beam.py
+++ b/src/nlp/commands/run_beam.py
@@ -7,8 +7,8 @@ import apache_beam as beam
 
 from nlp.builder import FORCE_REDOWNLOAD, HF_DATASETS_CACHE, REUSE_CACHE_IF_EXISTS, DatasetBuilder, DownloadConfig
 from nlp.commands import BaseTransformersCLICommand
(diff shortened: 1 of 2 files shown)


src/nlp/commands/run_beam.py
```diff
@@ -7,8 +7,8 @@
 
 from nlp.builder import FORCE_REDOWNLOAD, HF_DATASETS_CACHE, REUSE_CACHE_IF_EXISTS, DatasetBuilder, DownloadConfig
 from nlp.commands import BaseTransformersCLICommand
+from nlp.info import DATASET_INFOS_DICT_FILE_NAME
 from nlp.load import import_main_class, prepare_module
-from nlp.utils.info_utils import DATASET_INFOS_DICT_FILE_NAME
```
patrickvonplaten, line 11: @lhoestq this threw an error for me when running a command

lhoestq: good catch !

PR #214, opened 2020-05-28
[arrow_dataset.py] add new filter function

patrickvonplaten:
The `.map()` function is super useful, but can IMO a bit tedious when filtering certain examples.
I think, filtering out examples is also a very common operation people would like to perform on datasets.

This PR is a proposal to add a `.filter()` function in the same spirit than the `.map()` function.

Here is a sample code you can play around with:

```python
ds = nlp.load_dataset("squad", split="validation[:10%]")


def remove_under_idx_5(example, idx):
    return idx < 5


def only_keep_examples_with_is_in_context(example):
    return "is" in example["context"]


result_keep_only_first_5 = ds.filter(remove_under_idx_5, with_indices=True, load_from_cache_file=False)
result_keep_examples_with_is_in_context = ds.filter(only_keep_examples_with_is_in_context, load_from_cache_file=False)

print("Original number of examples: {}".format(len(ds)))
print("First five examples number of examples: {}".format(len(result_keep_only_first_5)))
print("Is in context examples number of examples: {}".format(len(result_keep_examples_with_is_in_context)))
```


src/nlp/arrow_dataset.py
```diff
@@ -557,3 +558,43 @@ def apply_function_on_filtered_inputs(inputs, indices):
...
+
+    def filter(
+        self,
+        function,
+        with_indices=False,
+        **kwargs
+    ):
+        def map_function(batch, *args):
+            # flatten dict to list
+            flattened_batch = flatten_nested(batch)
```
lhoestq, line 570: I'm not a big fan of using `flatten_nested` here

src/nlp/arrow_dataset.py
```diff
@@ -557,3 +558,43 @@ def apply_function_on_filtered_inputs(inputs, indices):
...
+            num_items = len(batch.keys())
+            num_examples = len(flattened_batch) // num_items
+            result = defaultdict(list)
+
+            # create single examples
+            for example_pos in range(num_examples):
+                example = {}
+                for i, key in enumerate(batch.keys()):
+                    example[key] = flattened_batch[i * num_examples + example_pos]
```
lhoestq, line 579: It's probably not going to work for nested structures

lhoestq:
I agree that a `.filter` method would be VERY useful and appreciated. I'm not a big fan of using `flatten_nested` as it completely breaks down the structure of the example and it may create bugs. Right now I think it may not work for nested structures. Maybe there's a simpler way that we've not figured out yet.

lhoestq:
Instead of flattening everything and rebuilding the example, maybe we can try to access the examples like this:
```python
for i in range(num_examples):
    example = map_nested(lambda x: x[i], batch)
    # ... then test to keep it or not
```

src/nlp/arrow_dataset.py
patrickvonplaten: Works for `trivia_qa` as well :-) It essentially just merges the value lists of all values in a `dict`

patrickvonplaten:
> Instead of flattening everything and rebuilding the example, maybe we can try to access the examples like this:
> 
> ```python
> for i in range(num_examples):
>     example = map_nested(lambda x: x[i], batch)
>     # ... then test to keep it or not
> ```

Awesome I'll check it out :-)

src/nlp/arrow_dataset.py
lhoestq: Ok I see but if we have a batch of size two like this one:
```python
{
   "qa": {
        "question": ["foo", "bar],
        "answer": ["1", "2"]
    }
}
```
Then the reconstructed examples will be different as `flatten_nested` will return something like `["foo", "bar", "1", "2"]` and you only reconstruct the first depth of the structure (`len(batch.keys())`=1)

patrickvonplaten:
> Instead of flattening everything and rebuilding the example, maybe we can try to access the examples like this:
> 
> ```python
> for i in range(num_examples):
>     example = map_nested(lambda x: x[i], batch)
>     # ... then test to keep it or not
> ```

Awesome this function is definitely much nicer!

lhoestq:
Actually I just realized that `map_nested` might not work either as it applies the function at the very las list of the structure. However we can imagine that a single example has also a list in its structure:
```python
one_example = {
    "title": "blabla",
    "paragraphs": [
        "p1", "p2", ...
    ]
}
```

lhoestq:
We'll probably have to take into account the `dset._data.schema` to extract the examples from the batch.

patrickvonplaten:
> Actually I just realized that `map_nested` might not work either as it applies the function at the very last list of the structure. However we can imagine that a single example has also a list in its structure:
> 
> ```python
> one_example = {
>     "title": "blabla",
>     "paragraphs": [
>         "p1", "p2", ...
>     ]
> }
> ```

They both work. I'm using it on trivia_qa which is pretty nested. If you use the option `with_dict=True` I think it's fine.

src/nlp/arrow_dataset.py
```diff
@@ -557,3 +558,39 @@ def apply_function_on_filtered_inputs(inputs, indices):
...
             return self
+
+    def filter(self, function, with_indices=False, **kwargs):
+        def map_function(batch, *args):
+            result = defaultdict(list)
+            num_examples = len(batch[next(iter(batch.keys()))])
+
+            # create single examples
+            for i in range(num_examples):
+                example = map_nested(lambda x: x[i], batch, dict_only=True)
```
patrickvonplaten, line 569: So we assume that `batch` is a `dict{key_1: list, key_2: list, ...}`. With `dict_only=True` this means that the function has to be applied directly to every list in the `dict` and it should be fine @lhoestq

patrickvonplaten:
> We'll probably have to take into account the `dset._data.schema` to extract the examples from the batch.

Why?

src/nlp/arrow_dataset.py
lhoestq: Ok I see what you mean !

lhoestq:
Actually it's fine. I guess this is going to be yet another thing to be unit-tested just to make sure ^^

patrickvonplaten:
Yes, I will need to add tests and documentation! 
@thomwolf - would a function like this be ok? It abstracts `.map()` a bit which might be hard to understand.

lhoestq:
I tried on some datasets with nested structure and it works fine ! Great work :D

patrickvonplaten:
Awesome :-), I will add documentation and some simple unittests

src/nlp/arrow_dataset.py
```diff
@@ -557,3 +558,66 @@ def apply_function_on_filtered_inputs(inputs, indices):
                 return Dataset.from_buffer(buf_writer.getvalue())
         else:
             return self
+
+    def filter(self, function, with_indices=False, **kwargs):
+        """ Apply a filter function to all the elements in the table in batches
+            and update the table so that the dataset only includes examples according to the filter function.
+
+            Args:
```
patrickvonplaten, line 566: added docstring here

tests/test_arrow_dataset.py
```diff
@@ -0,0 +1,114 @@
...
+
+import numpy as np
+import pyarrow as pa
+
+from nlp.arrow_reader import BaseReader
+from nlp.info import DatasetInfo
+from nlp.splits import SplitDict, SplitInfo
+
+
+class ReaderTester(BaseReader):
```
patrickvonplaten, line 13: @lhoestq - it's slightly different from what you used so I copy pasted the code. I guess we could refactor this into `test_utils.py` at one point.

tests/test_arrow_dataset.py
```diff
@@ -0,0 +1,114 @@
...
+            filename_skip_take["skip"] if "skip" in filename_skip_take else None,
+            filename_skip_take["take"] if "take" in filename_skip_take else None,
+        )
+        pa_table = pa.Table.from_pydict({"filename": [filename + "_" + str(x) for x in np.arange(100).tolist()]})
+        if skip is not None and take is not None:
+            pa_table = pa_table.slice(skip, take)
+        return pa_table
+
+
+class BaseDatasetTest(TestCase):
```
patrickvonplaten, line 32: @thomwolf @lhoestq  - added 3 tests for `.map()` for now and 1 test for `.filter()`

patrickvonplaten:
Ok merging!

datasets/wikipedia/wikipedia.py
```diff
@@ -25,9 +25,9 @@
 import xml.etree.cElementTree as etree
 
 import apache_beam as beam
-import mwparserfromhell
 import six
 
+import mwparserfromhell
```
lhoestq, line 30: hmmmmmmmmmmmmmmm

PR #220, opened 2020-05-29
dataset_arcd

tayciryahmed:
Added Arabic Reading Comprehension Dataset (ARCD): https://arxiv.org/abs/1906.05394

PR #221, opened 2020-05-29
Fix tests/test_dataset_common.py

tayciryahmed:
When I run the command `RUN_SLOW=1 pytest tests/test_dataset_common.py::LocalDatasetTest::test_load_real_dataset_arcd` while working on #220. I get the error ` unexpected keyword argument "'download_and_prepare_kwargs'"` at the level of  `load_dataset`. Indeed, this [function](https://github.com/huggingface/nlp/blob/master/src/nlp/load.py#L441) no longer has the argument `download_and_prepare_kwargs` but rather `download_config`. So here I change the tests accordingly.

### DIFF
--- a/tests/test_dataset_common.py
+++ b/tests/test_dataset_common.py
@@ -170,9 +170,7 @@ class LocalDatasetTest(parameterized.TestCase):
             download_config.download_mode = GenerateMode.FORCE_REDOWNLOAD
 
             dataset = load_dataset(
-                "./datasets/" + dataset_name,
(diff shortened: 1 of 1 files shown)


lhoestq:
Thanks ! Good catch :)

To fix the CI you can do:
1 - rebase from master
2 - then run `make style` as specified in [CONTRIBUTING.md](https://github.com/huggingface/nlp/blob/master/CONTRIBUTING.md) ?

PR #226, opened 2020-06-01
add BlendedSkillTalk dataset

mariamabarham:
This PR add the BlendedSkillTalk dataset, which is used to fine tune the blenderbot.


lhoestq:
Awesome :D

PR #235, opened 2020-06-04
Add experimental datasets

yjernite:
## Adding an *experimental datasets* folder

After using the 🤗nlp library for some time, I find that while it makes it super easy to create new memory-mapped datasets with lots of cool utilities, a lot of what I want to do doesn't work well with the current `MockDownloader` based testing paradigm, making it hard to share my work with the community.

My suggestion would be to add a **datasets\_experimental** folder so we can start making these new datasets public without having to completely re-think testing for every single one. We would allow contributors to submit dataset PRs in this folder, but require an explanation for why the current testing suite doesn't work for them. We can then aggregate the feedback and periodically see what's missing from the current tests.

I have added a **datasets\_experimental** folder to the repository and S3 bucket with two initial datasets: ELI5 (explainlikeimfive) and a Wikipedia Snippets dataset to support indexing (wiki\_snippets)

#### Dataset description
This allows people to download the [ELI5: Long Form Question Answering](https://arxiv.org/abs/1907.09190) dataset, along with two variants based on the r/askscience and r/AskHistorians. Full Reddit dumps for each month are downloaded from [pushshift](https://files.pushshift.io/reddit/), filtered for submissions and comments from the desired subreddits, then deleted one at a time to save space. The resulting dataset is split into a training, validation, and test dataset for r/explainlikeimfive, r/askscience, and r/AskHistorians respectively, where each item is a question along with all of its high scoring answers.

#### Issues with the current testing
1. the list of files to be downloaded is not pre-defined, but rather determined by parsing an index web page at run time. This is necessary as the name and compression type of the dump files changes from month to month as the pushshift website is maintained.  Currently, the dummy folder requires the user to know which files will be downloaded.
2. to save time, the script works on the compressed files using the corresponding python packages rather than first running `download\_and\_extract` then filtering the extracted files.  

#### Dataset description
This script creates a *snippets* version of a source Wikipedia dataset: each article is split into passages of fixed length which can then be indexed using ElasticSearch or a dense indexer. The script currently handles all **wikipedia** and **wiki40b** source datasets, and allows the user to choose the passage length and how much overlap they want across passages. In addition to the passage text, each snippet also has the article title, list of titles of sections covered by the text, and information to map the passage back to the initial dataset at the paragraph and character level.

#### Issues with the current testing
1. The DatasetBuilder needs to call `nlp.load_dataset()`. Currently, testing is not recursive (the test doesn't know where to find the dummy data for the source dataset)


patrickvonplaten:
I think it would be nicer to not create a new folder `datasets_experimental` , but just put your datasets also into the folder `datasets` for the following reasons:

- From my point of view, the datasets are not very different from the other datasets (assuming that we soon have C4, and the beam datasets) so I don't see why we require a new dataset folder

- I'm not a big fan of adding a boolean flag to the `load_dataset()` function that basically switches between folder names on S3. The user has to know whether a dataset script is experimental or not. User installing nlp with pip won't see that there are folders called `datasets` and `datasets_experimental`

- If we do this just to bypass the test, I think a good solution could be: For all tests that are too complicated to be currently tested with the testing framework, we can add a class variable called `do_test = False` and skip all tests that have that variable in the dataset test framework similar to what is done to beam datasets: https://github.com/huggingface/nlp/blob/2e0a8639a79b1abc848cff5c669094d40bba0f63/tests/test_dataset_common.py#L79 

- This way the datascripts are backward compatible and can be used with earlier versions of `nlp` (not that this matters too much atm) 

What is your opinion on this @lhoestq ?

lhoestq:
Very cool to have add those datasets :)
I understand that making the dummy data for this case is not fun. I'm sure we'll be able to add them soon. For now it's still interesting to have them in the library, even if we can't test all the code with dummy data.

I like the idea of the `do_tests=False` class variable. 
However it would be cool to test at least that we can load the module and instantiate the builder (only ignore the dummy data test for now). In that case a better name could be `test_dummy_data=False` or something like that.

If we want to be picky we can also add a warning in `_download_and_prepare` to tell the user that datasets with `test_dummy_data=False` are still experimental.

yjernite:
Yeah I really like the idea of a partial test.

My main concern with the class variable is visibility, but having a warning would help with that. Maybe even get the user to agree > "are you sure you want to go ahead?"

patrickvonplaten:
> Very cool to have add those datasets :)
> I understand that making the dummy data for this case is not fun. I'm sure we'll be able to add them soon. For now it's still interesting to have them in the library, even if we can't test all the code with dummy data.
> 
> I like the idea of the `do_tests=False` class variable.
> However it would be cool to test at least that we can load the module and instantiate the builder (only ignore the dummy data test for now). In that case a better name could be `test_dummy_data=False` or something like that.
> 
> If we want to be picky we can also add a warning in `_download_and_prepare` to tell the user that datasets with `test_dummy_data=False` are still experimental.

`test_dummy_data=False` sounds good to me!

yjernite:
There we go: added a `test_dummy_data` class variable that is `False` by default for the `BeamBasedBuilder` and `True` for everyone else (except the new `explainlikeimfive` and `wiki_snippets`)

Note that `wiki_snippets` should become obsolete as soon as @lhoestq adds in the `IndexedDataset` class

patrickvonplaten:
Great! LGTM!

datasets/wiki_snippets/wiki_snippets.py
```diff
@@ -179,7 +181,15 @@ def _info(self):
         )
 
     def _split_generators(self, dl_manager):
+        # default configuration downloads Wiki40b
+        if not hasattr(self.config, "wikipedia_name"):
```
lhoestq, line 185: Not sure why we need this ? The config always has the wikipedia_name attribute no ?

datasets/explainlikeimfive/explainlikeimfive.py
```diff
@@ -0,0 +1,388 @@
...
+
+    def __init__(self, **kwargs):
+        """BuilderConfig for ExplainLikeImFive.
+    Args:
+      **kwargs: keyword arguments forwarded to super.
+    """
+        super(ExplainLikeImFiveConfig, self).__init__(**kwargs)
+
+
+class ExplainLikeImFive(nlp.GeneratorBasedBuilder):
```
lhoestq, line 258: In terms of naming I think we're using snake case/camel case correspondences between the class name and the script name. Therefore we should probably either rename the class name to `Explainlikeimfive` OR rename the script to `explain_like_im_fiveexplain_like_im_five.py`

lhoestq: Or we can name the dataset `eli5` to be able to do `load_dataset("eli5")`.
In that case the class name would be `Eli5` and script name `eli5/eli5.py`

datasets/wiki_snippets/wiki_snippets.py
yjernite: There seems to be cases where the Builder is launched with `name='default'` and uses a default `nlp.BuilderConfig`.

Did I miss something / is there a way to ensure that `default` actually uses a `WikiSnippetsConfig`?

datasets/explainlikeimfive/explainlikeimfive.py
yjernite: Yes let's go back to plain Eli5 :)

datasets/wiki_snippets/wiki_snippets.py
lhoestq: This should not happen. Do you know in which cases you end up with `name='default'` ?

yjernite: fixed the `BUILDER_CONFIG_CLASS`

datasets/eli5/eli5.py
```diff
@@ -0,0 +1,387 @@
...
+        super(Eli5Config, self).__init__(**kwargs)
+
+
+class Eli5(nlp.GeneratorBasedBuilder):
+    """ELI5: Explain Like I'm Five long form question answering dataset."""
+
+    BUILDER_CONFIG_CLASS = Eli5Config
+    _DATA_SPLIT_URL = "https://s3.amazonaws.com/datasets.huggingface.co/nlp/datasets/eli5/reddit_data_split.json"
+
+    name = "ELI5_LFQA"
```
lhoestq, line 264:
```suggestion
```

We should remove this line imo. The name will be replaced by "eli5" anyway when the builder is initialized

datasets/wiki_snippets/wiki_snippets.py
```diff
@@ -0,0 +1,203 @@
...
+        super(WikiSnippetsConfig, self).__init__(**kwargs)
+        self.wikipedia_name = wikipedia_name
+        self.wikipedia_version_name = wikipedia_version_name
+        self.snippets_length = snippets_length
+        self.snippets_overlap = snippets_overlap
+
+
+class WikiSnippets(nlp.GeneratorBasedBuilder):
+    BUILDER_CONFIG_CLASS = WikiSnippetsConfig
+    name = "wiki_snippets"
```
lhoestq, line 140:
```suggestion
```
same here

PR #236, opened 2020-06-04
CompGuessWhat?! dataset 

aleSuglia:
Hello,

Thanks for the amazing library that you put together. I'm Alessandro Suglia, the first author of CompGuessWhat?!, a recently released dataset for grounded language learning accepted to ACL 2020 ([https://compguesswhat.github.io](https://compguesswhat.github.io)).

This pull-request adds the CompGuessWhat?! splits that have been extracted from the original dataset. This is only part of our evaluation framework because there is also an additional split of the dataset that has a completely different set of games. I didn't integrate it yet because I didn't know what would be the best practice in this case. Let me clarify the scenario.

In our paper, we have a main dataset (let's call it `compguesswhat-gameplay`) and a zero-shot dataset (let's call it `compguesswhat-zs-gameplay`). In the current code of the pull-request, I have only integrated `compguesswhat-gameplay`. I was thinking that it would be nice to have the `compguesswhat-zs-gameplay` in the same dataset class by simply specifying some particular option to the `nlp.load_dataset()` factory. For instance:

```python

cgw = nlp.load_dataset("compguesswhat")
cgw_zs = nlp.load_dataset("compguesswhat", zero_shot=True)
```

The other option would be to have a separate dataset class. Any preferences?


mariamabarham:
Hi @aleSuglia, thanks for this great PR. Indeed you can have both datasets in one file. You need to add a config class which will allows you to specify the different subdataset names and then  you will be able to load them as follow.
nlp.load_dataset("compguesswhat", "compguesswhat-gameplay") 
nlp.load_dataset("compguesswhat", "compguesswhat-zs-gameplay").

Maybe you can refer to this file https://github.com/huggingface/nlp/blob/master/datasets/discofuse/discofuse.py

aleSuglia:
@mariamabarham Thanks for your suggestions. I've followed your advice and integrated the additional dataset using another `DatasetConfig` class. It looks like all tests passed. What do you think?

mariamabarham:
great @aleSuglia. I requested an additional review from @thomwolf @lhoestq and @patrickvonplaten @jplu . You can merge it after an approval from one of them

patrickvonplaten:
Looks great! Thanks for adding the dummy data :-)

aleSuglia:
Not sure whether it's the most appropriate place but I'll ask another design question. For Vision+Language dataset, is very common to have visual features associated with each example. At the moment, for instance, I'm only integrating the image identifier so that people can later on lookup the image features during training. Do you recommend this approach or do you think it should be done in a different way?

Thank you for your answer!

datasets/compguesswhat/compguesswhat.py
```diff
@@ -0,0 +1,224 @@
...
+
+import gzip
+import json
+import os
+
+import nlp
+
+
+class CompguesswhatConfig(nlp.BuilderConfig):
+    """ BuilderConfig for Discofuse"""
```
thomwolf, line 11: detail but can you change Discofuse in Compguesswhat?

datasets/compguesswhat/compguesswhat.py
```diff
@@ -0,0 +1,224 @@
...
+
+    def __init__(self, data_url, splits, gameplay_scenario, **kwargs):
+        """
+
+        Args:
+            balanced: to specify if we want to load the balanced file or the full file
```
thomwolf, line 17: I don't see `balanced` used but maybe you can explain `gameplay_scenario` in the docstring?

thomwolf:
Hi @aleSuglia your remark on the visual features is a good point.

We haven't started to dive deeply into how CV datasets are usually structured (cc @sgugger)

Do you have a pointer to how visual features are currently loaded and accessed by people using GuessCompWhat?

aleSuglia:
@thomwolf As far as I know, people using Language+Vision tasks they typically have their reference dataset (either in JSON or JSONL format) and for each example in it they have an identifier that specifies the reference image. Currently, images are represented by either pooling-based visual features (average pooling of ResNet or VGGNet features, see [DeVries et.al, 2017](https://arxiv.org/abs/1611.08481), [Shekhar et.al, 2019](https://www.aclweb.org/anthology/N19-1265.pdf)) where you have a single vector for every image. Another option is to use a set of feature maps for every image extracted from a specific layer of a CNN (see [Xu et.al, 2015](https://arxiv.org/abs/1502.03044)). A more common and recent option, especially with large-scale multi-modal transformers [Li et. al, 2019](https://arxiv.org/abs/1908.03557), is to use FastRCNN features. 

For all these types of features, people use either H5DF or NumPy compressed representations. In my personal projects, I've ditched altogether H5DF because it doesn't have out-of-the-box support for multi-processing (unless you have an ad-hoc installation of it). I've been successfully using a NumPy compressed file for each image so that I can store any sort of information in it (see [numpy.savez](https://numpy.org/doc/stable/reference/generated/numpy.savez.html)). However, I believe that Apache Arrow would be a really good fit for this type of features. 

Looking forward to hearing your thoughts about it!

datasets/compguesswhat/dataset_infos.json
```diff
@@ -0,0 +1,580 @@
...
+    "download_checksums": {
+      "https://www.dropbox.com/s/gawk420r77huc01/compguesswhat_games.zip?dl=1": {
+        "num_bytes": 106007842,
+        "checksum": "8ee244654ff6c9968d0bcccc903e5fd81b90ad7a6dae0f47f0f99861f84a088f"
+      }
+    },
+    "download_size": 106007842,
+    "dataset_size": 176646094,
+    "size_in_bytes": 282653936
+  },
```
lhoestq, line 185: If I'm not mistaken there are two configs 'compguesswhat-original' and 'compguesswhat-zeroshot', so this one should be removed:
```suggestion
```
It was probably something that was generated before you switched to having two configs with those names

aleSuglia: Good catch. Yes I just regenerated them. Thanks!

lhoestq:
Awesome work on this one thanks :)

aleSuglia:
@thomwolf I was thinking that I should create an issue regarding the visual features so that we can keep track of it for future work. I think it would be great to have it in NLP and I'll be happy to contribute. Let me know what you think :)

PR #244, opened 2020-06-05
Add Allociné Dataset

TheophileBlard:
This is a french binary sentiment classification dataset, which was used to train this model: https://huggingface.co/tblard/tf-allocine.

Basically, it's a french "IMDB" dataset, with more reviews.

More info on [this repo](https://github.com/TheophileBlard/french-sentiment-analysis-with-bert).


mariamabarham:
great work @TheophileBlard

patrickvonplaten:
LGTM, thanks a lot for adding dummy data tests :-) Was it difficult to create the correct dummy data folder?

TheophileBlard:
It was pretty easy actually. Documentation is on point !

datasets/allocine/allocine.py
```diff
@@ -0,0 +1,89 @@
...
+            ),
+            nlp.SplitGenerator(name=nlp.Split.TEST, gen_kwargs={"filepath": os.path.join(data_dir, self._TEST_FILE)}),
+        ]
+
+    def _generate_examples(self, filepath):
+        """Generate Allocine examples."""
+        with open(filepath) as f:
+            for row in f:
+                data = json.loads(row)
+                id_ = data["film-url"]
```
TheophileBlard, line 86: As there are multiple reviews of the same movie in each splits, the `id_` field won't be unique. Is this an issue ?

patrickvonplaten: good question! @mariamabarham @lhoestq - do you know more here? I remember that it was a problem it tfds, but maybe not anymore for us?

lhoestq: Right now it wouldn't change anything but I would still suggest to use unique ids. One day we will probably leverage them

thomwolf: Yep, we don't care about unique ids but tfds does so maybe good to have them unique so you can adapt your datasets to tfds if you want one day.

TheophileBlard: Thanks for the answers !  I updated the code so that each example as now a unique id (abf565d).

lhoestq: Thanks :)

PR #247, opened 2020-06-06
Make all dataset downloads deterministic by applying `sorted` to glob and os.listdir

patrickvonplaten:
This PR makes all datasets loading deterministic by applying `sorted()` to all `glob.glob` and `os.listdir` statements.

Are there other "non-deterministic" functions apart from `glob.glob()` and `os.listdir()` that you can think of @thomwolf @lhoestq @mariamabarham @jplu ?

**Important** 
It does break backward compatibility for these datasets because
1. When loading the complete dataset the order in which the examples are saved is different now
2. When loading only part of a split, the examples themselves might be different.

@patrickvonplaten - the nlp / longformer notebook has to be updated since the examples might now be different

### DIFF
--- a/datasets/aeslc/aeslc.py
+++ b/datasets/aeslc/aeslc.py
@@ -81,7 +81,7 @@ class Aeslc(nlp.GeneratorBasedBuilder):
 
     def _generate_examples(self, pattern=None):
         """Yields examples."""
-        for filename in glob.glob(pattern):
+        for filename in sorted(glob.glob(pattern)):
             email_body, subject_line = _parse_email_file(filename)
             key = os.path.basename(filename).rstrip(".subject")
             yield key, {_DOCUMENT: email_body, _SUMMARY: subject_line}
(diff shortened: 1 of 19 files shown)


thomwolf:
That's great!

I think it would be nice to test "deterministic-ness" if we can do it (should be left for future PR of course)
Here is a possibility (maybe there are other ways to do it): given that we may soon have efficient and large-scale hashing (cf our discussion on versioning/tracability), we could incorporate a hash of the final datasets to the `dataset.json` file and have a test on it as well as CI on a diversity of platform to test the hash (Win/Mac/Linux + various python/env). What do you think @lhoestq @patrickvonplaten?

patrickvonplaten:
> That's great!
> 
> I think it would be nice to test "deterministic-ness" of datasets in CI if we can do it (should be left for future PR of course)
> 
> Here is a possibility (maybe there are other ways to do it): given that we may soon have efficient and large-scale hashing (cf our discussion on versioning/tracability), we could incorporate a hash of the final Arrow Dataset to the `dataset.json` file and have a test on it as well as CI on a diversity of platform to test the hash (Win/Mac/Linux + various python/env).
> What do you think @lhoestq @patrickvonplaten?

I think that's a great idea! The test should be a `RUN_SLOW` test, since it takes a considerable amount of time to download the dataset and generate the examples, but I think we should add some kind of hash test for each dataset.

jplu:
Really nice!!

PR #251, opened 2020-06-08
Better access to all dataset information

thomwolf:
Moves all the dataset info down one level from `dataset.info.XXX` to `dataset.XXX`
This way it's easier to access `dataset.feature['label']` for instance

Also, add the original split instructions used to create the dataset in `dataset.split`
Ex:
```
from nlp import load_dataset
stsb = load_dataset('glue', name='stsb', split='train')
stsb.split
>>> NamedSplit('train')
```

### DIFF
--- a/src/nlp/arrow_dataset.py
+++ b/src/nlp/arrow_dataset.py
@@ -38,8 +38,9 @@ logger = logging.getLogger(__name__)
 
 class DatasetInfoMixin(object):
     """ This base class exposes some attributes of DatasetInfo
-        at the base level of the Dataset for easy access. 
+        at the base level of the Dataset for easy access.
     """
+
     def __init__(self, info, split):
         self._info = info
         self._split = split
--- a/src/nlp/arrow_dataset.py
+++ b/src/nlp/arrow_dataset.py
@@ -37,6 +37,9 @@ logger = logging.getLogger(__name__)
 
 
 class DatasetInfoMixin(object):
+    """ This base class exposes some attributes of DatasetInfo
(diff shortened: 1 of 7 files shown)


src/nlp/arrow_dataset.py
```diff
@@ -522,7 +588,7 @@ def apply_function_on_filtered_inputs(inputs, indices):
                 cache_file_name = self._get_cache_file_path(function, cache_kwargs)
             if os.path.exists(cache_file_name) and load_from_cache_file:
                 logger.info("Loading cached processed dataset at %s", cache_file_name)
-                return Dataset.from_file(cache_file_name)
+                return Dataset.from_file(cache_file_name, info=self.info, split=self.split)
```
lhoestq, line 591: Nice !

src/nlp/arrow_reader.py
```diff
@@ -138,21 +138,22 @@ def _get_dataset_from_filename(self, filename_skip_take):
...
+            original_instructions: store the original instructions used to build the dataset split in the dataset
         """
         pa_batches = []
         for f_dict in files:
             pa_table: pa.Table = self._get_dataset_from_filename(f_dict)
             pa_batches.extend(pa_table.to_batches())
         if pa_batches:
             pa_table = pa.Table.from_batches(pa_batches)
-        ds = Dataset(arrow_table=pa_table, data_files=files, info=info)
+        ds = Dataset(arrow_table=pa_table, data_files=files, info=info, split=original_instructions)
```
lhoestq, line 156: Even nicer !

PR #262, opened 2020-06-11
Add new dataset ANLI Round 1

easonnie:
Adding new dataset [ANLI](https://github.com/facebookresearch/anli/).

I'm not familiar with how to add new dataset. Let me know if there is any issue. I only include round 1 data here. There will be round 2, round 3 and more in the future with potentially different format. I think it will be better to separate them.

### DIFF
--- a/datasets/bookcorpus/bookcorpus.py
+++ b/datasets/bookcorpus/bookcorpus.py
@@ -24,6 +24,7 @@ import re
 
 import nlp
 
+
 _DESCRIPTION = """\
 Books are a rich source of both fine-grained information, how a character, \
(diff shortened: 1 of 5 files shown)


lhoestq:
Hello ! Thanks for adding this one :)

This looks great, you just have to do the last steps to make the CI pass.
I can see that two things are missing:
1. the dummy data that is used to test that the script is working as expected
2. the json file with all the infos about the dataset

You can see the steps to help you create the dummy data and generate the dataset_infos.json file right [here](https://github.com/huggingface/nlp/blob/master/CONTRIBUTING.md#how-to-add-a-dataset)

PR #265, opened 2020-06-11
Add pyarrow warning colab

lhoestq:
When a user installs `nlp` on google colab, then google colab doesn't update pyarrow, and the runtime needs to be restarted to use the updated version of pyarrow.

This is an issue because `nlp` requires the updated version to work correctly.

In this PR I added en error that is shown to the user in google colab if the user tries to `import nlp` without having restarted the runtime. The error tells the user to restart the runtime.

### DIFF
--- a/src/nlp/__init__.py
+++ b/src/nlp/__init__.py
@@ -20,6 +20,7 @@
 
 __version__ = "0.2.0"
 
+import pyarrow
 from pyarrow import total_allocated_bytes
 
 from . import datasets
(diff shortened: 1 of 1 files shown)


src/nlp/__init__.py
```diff
@@ -34,3 +35,10 @@
 from .splits import NamedSplit, Split, SplitBase, SplitDict, SplitGenerator, SplitInfo, SubSplitInfo, percent
 from .utils import *
 from .utils.tqdm_utils import disable_progress_bar
+
+
+if int(pyarrow.__version__.split(".")[1]) < 16 or int(pyarrow.__version__.split(".")[0]) > 0:
```
rmorain, line 40: This conditional is incorrect. Should be `if int(pyarrow.__version__.split(".")[0]) <= 0 and int(pyarrow.__version__.split(".")[1]) < 16:`

Issue #259, opened 2020-06-10
documentation missing how to split a dataset


fotisj:
I am trying to understand how to split a dataset ( as arrow_dataset). 
I know I can do something like this to access a split which is already in the original dataset : 

`ds_test = nlp.load_dataset('imdb, split='test') `

But how can I split ds_test into a test and a validation set (without reading the data into memory and keeping the arrow_dataset as container)?
I guess it has something to do with the module split :-) but there is no real documentation in the code but only a reference to a longer description: 

> See the  [guide on splits](https://github.com/huggingface/nlp/tree/master/docs/splits.md)  for more information.

But the guide seems to be missing.

To clarify: I know that this has been modelled after the dataset of tensorflow and that some of the documentation there can be used [like this one](https://www.tensorflow.org/datasets/splits). But to come back to the example above: I cannot simply split the testset doing this: 
`ds_test = nlp.load_dataset('imdb, split='test'[:5000]) `
`ds_val = nlp.load_dataset('imdb, split='test'[5000:])`

because the imdb test data is sorted by class (probably not a good idea anyway)

lhoestq:
Currently you can indeed split a dataset using `ds_test = nlp.load_dataset('imdb, split='test[:5000]')` (works also with percentages).

However right now we don't have a way to shuffle a dataset but we are thinking about it in the discussion in #166. Feel free to share your thoughts about it.

One trick that you can do until we have a better solution is to shuffle and split the indices of your dataset:
```python
import nlp
from sklearn.model_selection import train_test_split

imdb = nlp.load_dataset('imbd', split='test')
test_indices, val_indices = train_test_split(range(len(imbd)))
```

and then to iterate each split:
```python
for i in test_indices:
    example = imdb[i]
   ...
```

lhoestq:
I added a small guide [here](https://github.com/huggingface/nlp/tree/master/docs/splits.md) that explains how to split a dataset. It is very similar to the tensorflow datasets guide, as we kept the same logic.

fotisj:
Thanks a lot, the new explanation is very helpful!

About using train_test_split from sklearn: I stumbled across the [same error message as this user ](https://github.com/huggingface/nlp/issues/147 )and thought it can't be used at the moment in this context. Will check it out again.

One of the problems is how to shuffle very large datasets, which don't fit into the memory. Well, one strategy could be shuffling data in sections. But in a case where the data is sorted by the labels you have to swap larger sections first.

[...]

Issue #166, opened 2020-05-19
Add a method to shuffle a dataset


thomwolf:
Could maybe be a `dataset.shuffle(generator=None, seed=None)` signature method.

Also, we could maybe have a clear indication of which method modify in-place and which methods return/cache a modified dataset. I kinda like torch conversion of having an underscore suffix for all the methods which modify a dataset in-place. What do you think?

Colanim:
+1 for the naming convention

About the `shuffle` method, from my understanding it should be done in `Dataloader` (better separation between dataset processing - usage)

richardyy1188:
+1 for shuffle in `Dataloader`. 
Some `Dataloader` just store idxs of dataset and just shuffle those idxs, which might(?) be faster than do shuffle in dataset, especially when doing shuffle every epoch.

Also +1 for the naming convention.

vblagoje:
As you might already know the issue of dataset shuffling came up in the nlp code [walkthrough](https://youtu.be/G3pOvrKkFuk?t=3204) by Yannic Kilcher

lhoestq:
We added the `.shuffle` method :)

Closing this one.

PR #266, opened 2020-06-11
Add sort, shuffle, test_train_split and select methods

### LINKED ISSUE #147: Error with sklearn train_test_split
It would be nice if we could use sklearn `train_test_split` to quickly generate subsets from the dataset objects returned by `nlp.load_dataset`. At the moment the code:

```python
data = nlp.load_dataset('imdb', cache_dir=data_cache)
f_half, s_half = train_test_split(data['train'], test_size=0.5, random_state=seed)
```
throws:
```
ValueError: Can only get row(s) (int or slice) or columns (string).
```
It's not a big deal, since there are other ways to split the data, but it would be a cool thing to have.

thomwolf:
Add a bunch of methods to reorder/split/select rows in a dataset:
- `dataset.select(indices)`: Create a new dataset with rows selected following the list/array of indices (which can have a different size than the dataset and contain duplicated indices, the only constrain is that all the integers in the list must be smaller than the dataset size, otherwise we're indexing outside the dataset...)
- `dataset.sort(column_name)`: sort a dataset according to a column (has to be a column with a numpy compatible type)
- `dataset.shuffle(seed)`: shuffle a dataset rows
- `dataset.train_test_split(test_size, train_size)`: Return a dictionary with two random train and test subsets (`train` and `test` ``Dataset`` splits)

All these methods are **not** in-place which means they return new ``Dataset``.
This is the default behavior in the library.

Fix #147 #166 #259

### DIFF
--- a/src/nlp/arrow_dataset.py
+++ b/src/nlp/arrow_dataset.py
@@ -689,7 +689,8 @@ class Dataset(DatasetInfoMixin):
         # return map function
         return self.map(map_function, batched=True, with_indices=with_indices, arrow_schema=arrow_schema, **kwargs)
 
-    def sort(self,
+    def sort(
+        self,
         column: Optional[str] = None,
         indices: Optional[List[int]] = None,
         kind: str = None,
@@ -722,16 +723,18 @@ class Dataset(DatasetInfoMixin):
         if len(self) == 0:
             return self
 
-        assert (column is not None and isinstance(column, str)) or \
-            (indices is not None and isinstance(column, list)), "You must provide a column name or a list of indices"
-        assert column is None or indices is None, "You must provide either a column name or a list of indices but not both."
+        assert (column is not None and isinstance(column, str)) or (
+            indices is not None and isinstance(column, list)
+        ), "You must provide a column name or a list of indices"
+        assert (
+            column is None or indices is None
+        ), "You must provide either a column name or a list of indices but not both."
 
         # Check the column name
         if column is not None and column not in self._data.column_names:
             raise ValueError(
                 "Column name {} not in the dataset. Current columns in the dataset: {}".format(
-                    column,
-                    self._data.column_names,
+                    column, self._data.column_names,
                 )
             )
 
@@ -762,21 +765,12 @@ class Dataset(DatasetInfoMixin):
             writer = ArrowWriter(schema=self.schema, path=cache_file_name, writer_batch_size=writer_batch_size)
 
         if indices is None:
-            indices = self._getitem(
-                column,
-                format_type='numpy',
-                format_columns=None,
-                output_all_columns=False,
-            )
+            indices = self._getitem(column, format_type="numpy", format_columns=None, output_all_columns=False,)
             indices = np.argsort(indices, kind=kind)
 
         # Loop over single examples or batches and write to buffer/file if examples are to be updated
         for i in tqdm(indices):
-            example = self._getitem(
-                key=i,
-                format_type=None,
-                format_columns=None,
-            )
+            example = self._getitem(key=i.item(), format_type=None, format_columns=None,)
             writer.write(example)
 
         writer.finalize()  # close_stream=bool(buf_writer is None))  # We only close if we are writing in a file
--- a/src/nlp/arrow_dataset.py
+++ b/src/nlp/arrow_dataset.py
@@ -382,7 +382,9 @@ class Dataset(DatasetInfoMixin):
                     if format_type == "pandas":
                         outputs = self._data[key].to_pandas(zero_copy_only=False)
                     elif format_type == "numpy":
-                        outputs = np.concatenate([arr.to_numpy(zero_copy_only=False) for arr in self._data[key].chunks])
+                        outputs = np.concatenate(
+                            [arr.to_numpy(zero_copy_only=False) for arr in self._data[key].chunks]
+                        )
                     else:
                         outputs = self._convert_outputs(self._data[key].to_pylist(), format_type=format_type)
                 else:
@@ -1033,8 +1035,10 @@ class Dataset(DatasetInfoMixin):
                     test_cache_file_name = self._get_cache_file_path(self.train_test_split, test_kwargs)
             if os.path.exists(train_cache_file_name) and os.path.exists(test_cache_file_name) and load_from_cache_file:
                 logger.info("Loading cached split dataset at %s and %s", train_cache_file_name, test_cache_file_name)
-                return {'train': Dataset.from_file(train_cache_file_name, info=self.info, split=self.split),
-                        'test': Dataset.from_file(test_cache_file_name, info=self.info, split=self.split)}
+                return {
+                    "train": Dataset.from_file(train_cache_file_name, info=self.info, split=self.split),
+                    "test": Dataset.from_file(test_cache_file_name, info=self.info, split=self.split),
+                }
 
         if not shuffle:
             train_indices = np.arange(n_train)
--- a/src/nlp/arrow_dataset.py
+++ b/src/nlp/arrow_dataset.py
@@ -380,9 +380,9 @@ class Dataset(DatasetInfoMixin):
             if format_type is not None:
                 if format_columns is None or key in format_columns:
                     if format_type == "pandas":
-                        outputs = self._data[key].to_pandas()
+                        outputs = self._data[key].to_pandas(zero_copy_only=False)
                     elif format_type == "numpy":
-                        outputs = np.concatenate([arr.to_numpy() for arr in self._data[key].chunks])
+                        outputs = np.concatenate([arr.to_numpy(zero_copy_only=False) for arr in self._data[key].chunks])
                     else:
                         outputs = self._convert_outputs(self._data[key].to_pylist(), format_type=format_type)
                 else:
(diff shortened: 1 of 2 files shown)


lhoestq:
Nice !

Also it looks like we can have a train_test_split method for free:
```python
train_indices, test_indices = train_test_split(range(len(dataset)))
train = dataset.sort(indices=train_indices)
test = dataset.sort(indices=test_indices)
```

and a shuffling method for free:
```python
shuffled_indices = shuffle(range(len(dataset)))
shuffled_dataset = dataset.sort(indices=shuffled_indices)
```

Maybe we can have a specific API for train_test_split and shuffle. They are two features asked quite often (see #147, #166)

src/nlp/arrow_dataset.py
```diff
@@ -689,25 +689,87 @@ def map_function(batch, *args):
         # return map function
         return self.map(map_function, batched=True, with_indices=with_indices, arrow_schema=arrow_schema, **kwargs)
 
+
+    def index(
```
lhoestq, line 693: Maybe we can choose a name that is less ambiguous (cf 'indexing' with embeddings for example) ?
I can think of 'reorder' (or 'rearrange', or 'apply_permutation').

thomwolf: Two things here:
- I'm really happy you do this comment because that's exactly the main thing I was bothered about in this PR, this too generic name :) What an OSS connection!
- I like `reorder` let's go with this one!

thomwolf: I'm hesitating with `index_gather` like in PyTorch as well. In particular I would like to emphasize that the list of indices given doesn't have to be a permutation but can be shorter than the dataset size or contain duplicated indices.

thomwolf: Overall, I think the word `index` should stay FAISS/Elastic internals and not spread into the whole library. What do you think?

lhoestq: Yes `index_select` is fine too !
I agree with you to keep the word `index` for internals of the Faiss/ES similarity search feature.

thomwolf: Ok, let's go for `select` for the moment and see how we like it.

src/nlp/arrow_dataset.py
```diff
@@ -688,3 +689,365 @@ def map_function(batch, *args):
...
+        self,
+        test_size: Union[float, int, None] = None,
+        train_size: Union[float, int, None] = None,
+        shuffle: bool = True,
+        seed: Optional[int] = None,
+        generator: Optional[np.random.Generator] = None,
+        default_test_size: Optional[int] = 0.25,
+        keep_in_memory: bool = False,
+        load_from_cache_file: bool = True,
+        cache_file_name: Optional[str] = None,
```
lhoestq, line 899: I think we need two of them

src/nlp/arrow_dataset.py
```diff
@@ -688,3 +689,365 @@ def map_function(batch, *args):
...
+                    "seed": seed,
+                    "keep_in_memory": keep_in_memory,
+                    "load_from_cache_file": load_from_cache_file,
+                    "cache_file_name": cache_file_name,
+                    "writer_batch_size": writer_batch_size,
+                }
+                cache_file_name = self._get_cache_file_path(self.train_test_split, cache_kwargs)
+            if os.path.exists(cache_file_name) and load_from_cache_file:
+                logger.info("Loading cached sorted dataset at %s", cache_file_name)
+                return Dataset.from_file(cache_file_name, info=self.info, split=self.split)
```
lhoestq, line 1024: this one should be a dict

thomwolf:
Ok, I think this one is ready to merge.

@patrickvonplaten @jplu @mariamabarham you may want to give it a look, it adds a bunch of methods to reorder/split/select rows in a dataset:
- `dataset.select(indices)`: Create a new dataset with rows selected following the list/array of indices (which can have a different size than the dataset and contain duplicated indices, the only constrain is that all the integers in the list must be smaller than the dataset size, otherwise we're indexing outside the dataset...)
- `dataset.sort(column_name)`: sort a dataset according to a column (has to be a column with a numpy compatible type)
- `dataset.shuffle(seed)`: shuffle a dataset rows
- `dataset.train_test_split(test_size, train_size)`: Return a dictionary with two random train and test subsets (`train` and `test` ``Dataset`` splits)

All these methods are **not** in-place which means they return new ``Dataset``, which is the default behavior in the library.

src/nlp/arrow_dataset.py
```diff
@@ -379,9 +380,11 @@ def _getitem(
             if format_type is not None:
                 if format_columns is None or key in format_columns:
                     if format_type == "pandas":
-                        outputs = self._data[key].to_pandas()
+                        outputs = self._data[key].to_pandas(zero_copy_only=False)
```
lhoestq, line 383: Is there an issue with pandas too ?

thomwolf: Did that by precaution but maybe not indeed

src/nlp/arrow_dataset.py
```diff
@@ -688,3 +691,380 @@ def map_function(batch, *args):
...
+
+        writer.finalize()  # close_stream=bool(buf_writer is None))  # We only close if we are writing in a file
+
+        # Create new Dataset from buffer or file
+        if buf_writer is None:
+            return Dataset.from_file(cache_file_name, info=self.info, split=self.split)
+        else:
+            return Dataset.from_buffer(buf_writer.getvalue(), info=self.info, split=self.split)
+
+    def sort(
```
lhoestq, line 757: Could we have `reverse: bool = False` as in `sorted()` ? If it's true, you just have to reverse the indices `indices[::-1]`

src/nlp/arrow_dataset.py
```diff
@@ -688,3 +691,380 @@ def map_function(batch, *args):
...
+        )
+
+    def train_test_split(
+        self,
+        test_size: Union[float, int, None] = None,
+        train_size: Union[float, int, None] = None,
+        shuffle: bool = True,
+        seed: Optional[int] = None,
+        generator: Optional[np.random.Generator] = None,
+        default_test_size: Optional[int] = 0.25,
```
lhoestq, line 898: this one is not used anywhere ?

src/nlp/arrow_dataset.py
```diff
@@ -688,3 +691,380 @@ def map_function(batch, *args):
...
+                test_kwargs = cache_kwargs.deepcopy()
+                test_kwargs["split"] = "test"
+
+                if train_cache_file_name is None:
+                    train_cache_file_name = self._get_cache_file_path(self.train_test_split, train_kwargs)
+                if test_cache_file_name is None:
+                    test_cache_file_name = self._get_cache_file_path(self.train_test_split, test_kwargs)
+            if os.path.exists(train_cache_file_name) and os.path.exists(test_cache_file_name) and load_from_cache_file:
+                logger.info("Loading cached split dataset at %s and %s", train_cache_file_name, test_cache_file_name)
+                return {
```
lhoestq, line 1038: awesome !

src/nlp/arrow_dataset.py
```diff
@@ -688,3 +691,380 @@ def map_function(batch, *args):
...
+        )
+        test_split = self.select(
+            indices=test_indices,
+            keep_in_memory=keep_in_memory,
+            load_from_cache_file=load_from_cache_file,
+            cache_file_name=test_cache_file_name,
+            writer_batch_size=writer_batch_size,
+        )
+
+        return {"train": train_split, "test": test_split}
```
mariamabarham, line 1070: awesome

src/nlp/arrow_dataset.py
```diff
@@ -688,3 +691,380 @@ def map_function(batch, *args):
...
+
+        return self.select(
+            indices=permutation,
+            keep_in_memory=keep_in_memory,
+            load_from_cache_file=load_from_cache_file,
+            cache_file_name=cache_file_name,
+            writer_batch_size=writer_batch_size,
+        )
+
+    def train_test_split(
```
joeddav, line 891: Not this PR, but eventually a `stratify` argument like sklearn's might be useful as well.

src/nlp/arrow_dataset.py
```diff
@@ -688,3 +691,380 @@ def map_function(batch, *args):
...
+        self,
+        test_size: Union[float, int, None] = None,
+        train_size: Union[float, int, None] = None,
+        shuffle: bool = True,
+        seed: Optional[int] = None,
```
joeddav, line 896: Would it be worth raising a value error if both seed and generator are passed?

src/nlp/arrow_dataset.py
```diff
@@ -688,3 +691,380 @@ def map_function(batch, *args):
...
+        column: str,
+        kind: str = None,
+        keep_in_memory: bool = False,
+        load_from_cache_file: bool = True,
+        cache_file_name: Optional[str] = None,
+        writer_batch_size: Optional[int] = 1000,
+    ):
+        """ Sort the table according to a column.
+
+            Currently sorting according to a column name uses numpy sorting algorithm under the hood.
```
joeddav, line 768: Might be good to specify in the docstring for both `sort` and `shuffle` that it returns a copy rather than changing the ordering in place.

lhoestq:
> Might be a solution to put 0.25 and 0.75 as default values for respectively `test_size` and `train_size`. WDYT?

According to sklearn documentation, it is indeed set to 0.25 and 0.75 if both `test_size` and `train_size` are None.
Let me add it.

lhoestq:
I think we're good to go now :) @joeddav @thomwolf @jplu

PR #271, opened 2020-06-13
Fix allociné dataset configuration

TheophileBlard:
This is a patch for #244. According to the [live nlp viewer](url), the Allociné dataset must be loaded with :
```python
dataset = load_dataset('allocine', 'allocine')
```
This is redundant, as there is only one "dataset configuration", and should only be:
```python
dataset = load_dataset('allocine')
```

This is my mistake, because the code for [`allocine.py`](https://github.com/huggingface/nlp/blob/master/datasets/allocine/allocine.py) was inspired by [`imdb.py`](https://github.com/huggingface/nlp/blob/master/datasets/imdb/imdb.py), which also force the user to specify the "dataset configuration" (even if there is only one).

I believe this PR should solve this issue, making the Allociné dataset more convenient to use.


lhoestq:
Actually when there is only one configuration, then you don't need to specify the configuration in `load_dataset`. You can run:
```python
dataset = load_dataset('allocine')
```
and it works.

Maybe we should take that into account in the nlp viewer @srush ?

srush:
@lhoestq Just to understand the exact semantics. Are you suggesting that if there is exactly 1 configuration I should not show the configuration menu and just treat it as if there were 0 configurations?

lhoestq:
The configuration menu is fine imo.
It was more about the code snippet presented in the viewer.
For example for Allociné it currently shows this snippet to load the dataset:
```python
!pip install nlp
from nlp import load_dataset
dataset = load_dataset('allocine', 'allocine')
```
However for datasets with one or zero configurations, the second argument in `load_dataset` is optional. For Allociné, that has one configuration, we can expect to show instead:
```python
!pip install nlp
from nlp import load_dataset
dataset = load_dataset('allocine')
```

TheophileBlard:
> Actually when there is only one configuration, then you don't need to specify the configuration in `load_dataset`. You can run:
> 
> ```python
> dataset = load_dataset('allocine')
> ```
> 
> and it works.
> 
> Maybe we should take that into account in the nlp viewer @srush ?

Oh ok, I didn't expect it would work! 

Anyway, I think it's intrinsically better to simply remove the optional parameter. 
The dummy data folder architecture seems also more logical this way.

srush:
Fixed in the viewer. Checked that allocine works.

lhoestq:
Awesome thanks :)

Closing this.

PR #276, opened 2020-06-16
Fix metric compute (original_instructions missing)

### LINKED ISSUE #269: Error in metric.compute: missing `original_instructions` argument
I'm running into an error using metrics for computation in the latest master as well as version 0.2.1. Here is a minimal example:

```python
import nlp
rte_metric = nlp.load_metric('glue', name="rte")
rte_metric.compute(
    [0, 0, 1, 1],
    [0, 1, 0, 1],
)
```

```
    181             # Read the predictions and references
    182             reader = ArrowReader(path=self.data_dir, info=None)
--> 183             self.data = reader.read_files(node_files)
    184 
    185             # Release all of our locks

TypeError: read_files() missing 1 required positional argument: 'original_instructions'
```

I believe this might have been introduced with cc8d2508b75f7ba0e5438d0686ee02dcec43c7f4, which added the `original_instructions` argument. Elsewhere, an empty-string default is provided--perhaps that could be done here too?

lhoestq:
When loading arrow data we added in cc8d250 a way to specify the instructions that were used to store them with the loaded dataset.
However metrics load data the same way but don't need instructions (we use one single file).

In this PR I just make `original_instructions` optional when reading files to load a `Dataset` object.

This should fix #269

### DIFF
--- a/src/nlp/arrow_reader.py
+++ b/src/nlp/arrow_reader.py
@@ -22,7 +22,7 @@ import math
 import os
 import re
 from dataclasses import dataclass
-from typing import List, Optional, Any
+from typing import List, Optional
 
 import pyarrow as pa
 import pyarrow.parquet
--- a/src/nlp/arrow_reader.py
+++ b/src/nlp/arrow_reader.py
@@ -22,7 +22,7 @@ import math
 import os
 import re
 from dataclasses import dataclass
-from typing import List, Optional
+from typing import List, Optional, Any
 
 import pyarrow as pa
 import pyarrow.parquet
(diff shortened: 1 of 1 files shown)


stefan-it:
Awesome! This is working now:

```python
import nlp 
seqeval = nlp.load_metric("seqeval") 
y_true = [['O', 'O', 'O', 'B-MISC', 'I-MISC', 'I-MISC', 'O'], ['B-PER', 'I-PER', 'O']] 
y_pred = [['O', 'O', 'B-MISC', 'I-MISC', 'I-MISC', 'I-MISC', 'O'], ['B-PER', 'I-PER', 'O']]                                                                                                                                                                                    

results = seqeval.compute(y_true, y_pred)
```

I heavily need this fix for an upcoming `nlp` integration PR for Transformers (token classification example)  😅

lhoestq:
Haha nice ! We'll ship this fix with the next release that will probably come out on thursday :)

PR #289, opened 2020-06-19
update xsum

mariamabarham:
This PR makes the following update to the xsum dataset:

- Manual download is not required anymore

- dataset can be loaded as follow: `nlp.load_dataset('xsum`)


patrickvonplaten:
Looks cool!
@mariamabarham can you add a detailed description here what exactly is changed and how the user can load xsum now?

patrickvonplaten:
And a rebase should solve the conflicts

patrickvonplaten:
This is a super useful PR :-) @sshleifer - maybe you can take a look at the updated version of xsum if you can use it for your use case. Now, one should be able to just load it with:

```python 
nlp.load_datasets("xsum", ....)  # no manual dir required anymore
```

PR #291, opened 2020-06-22
break statement not required

mayurnewase:


mayurnewase:
I guess,test failing due to connection error?

lhoestq:
We just fixed the other dataset on master. Could you rebase from master and push to rerun the CI ?

lhoestq:
If I'm not wrong this function returns None if no main class was found.
I think it makes things less clear not to have a return at the end of the function.
I guess we can have one return in the for loop instead of the break statement, AND one return at the end to explicitly return None.
What do you think ?

mayur741:
I guess current logic makes it more readable.
you can close this pr.

PR #298, opened 2020-06-23
Add searchable datasets

lhoestq:
# Better support for Numpy format + Add Indexed Datasets

I was working on adding Indexed Datasets but in the meantime I had to also add more support for Numpy arrays in the lib.

## Better support for Numpy format

New features:
- New fast method to convert Numpy arrays from Arrow structure (up to x100 speed up) using Pandas.
- Allow to output Numpy arrays in batched `.map`, which was the only missing part to fully support Numpy arrays.

Pandas offers fast zero-copy Numpy arrays conversion from Arrow structures.
Using it we can speed up the reading of memory-mapped Numpy array stored in Arrow format.

With these changes you can easily compute embeddings of texts using `.map()`. For example:
```python
def embed(text):
    tokenized_example = tokenizer.encode(text, return_tensors="pt")
    embeddings = bert_encoder(tokenized_examples).numpy()
    return embeddings
dset_with_embeddings = dset.map(lambda example: {"embeddings": embed(example["text])})
```
And then reading the embeddings from the arrow format is be very fast.

PS1: Note that right now only 1d arrays are supported.
PS2: It seems possible to do without pandas but it will require more _trickery_.
PS3: I did a simple benchmark with google colab that you can view here:
https://colab.research.google.com/drive/1QlLTR6LRwYOKGJ-hTHmHyolE3wJzvfFg?usp=sharing

## Add Indexed Datasets

For many retrieval tasks it is convenient to index a dataset to be able to run fast queries.
For example for models like DPR, REALM, RAG etc. that are models for Open Domain QA, the retrieval step is very important.

Therefore I added two ways to add an index to a column of a dataset:
1) You can index it using a Dense Index like Faiss. It is used to index vectors.
    Faiss is a library for efficient similarity search and clustering of dense vectors.
    It contains algorithms that search in sets of vectors of any size, up to ones that possibly do not fit in RAM.
2) You can index it using a Sparse Index like Elasticsearch. It is used to index text and run queries based on BM25 similarity.

Example of usage:

```python
ds = nlp.load_dataset('crime_and_punish', split='train')
ds_with_embeddings = ds.map(lambda example: {'embeddings': embed(example['line']}))  # `embed` outputs a `np.array`
ds_with_embeddings.add_vector_index(column='embeddings')
scores, retrieved_examples = ds_with_embeddings.get_nearest(column='embeddings', query=embed('my new query'), k=10)
```

```python
ds = nlp.load_dataset('crime_and_punish', split='train')
es_client = elasticsearch.Elasticsearch()
ds.add_text_index(column='line', es_client=es_client, index_name="my_es_index")
scores, retrieved_examples = ds.get_nearest(column='line', query='my new query', k=10)
```

PS4: Faiss allows to specify many options for the [index](https://github.com/facebookresearch/faiss/wiki/The-index-factory) and for [GPU settings](https://github.com/facebookresearch/faiss/wiki/Faiss-on-the-GPU). I made sure that the user has full control over those settings.

## Tests

I added tests for Faiss, Elasticsearch and indexed datasets.
I had to edit the CI config because all the test scripts were not being run by CircleCI.

------------------

I'd be really happy to have some feedbacks :)


src/nlp/arrow_writer.py
```diff
@@ -135,6 +136,7 @@ def write(self, example: Dict[str, Any], writer_batch_size: Optional[int] = None
         Args:
             example: the Example to add.
         """
+        example = convert_sequences_in_lists(example)
```
lhoestq, line 139: Pyarrow doesn't like multi-dimensional numpy arrays nor tuples

src/nlp/search.py
```diff
@@ -0,0 +1,366 @@
...
+                `indices` (`List[List[int]]`): The indices of the retrieved examples per query.
+        """
+        assert len(queries.shape) == 2
+        if not queries.flags.c_contiguous:
+            queries = np.asarray(queries, order="C")
+        scores, indices = self.faiss_index.search(queries, k)
+        return scores, indices.astype(int)
+
+
+class IndexableMixin:
```
lhoestq, line 241: This is the Mixin class that adds indexing features to `nlp.Dataset`

src/nlp/search.py
```diff
@@ -0,0 +1,366 @@
...
+    """Add indexing features to `nlp.Dataset`"""
+
+    def __init__(self):
+        self._indexes: Dict[str, BaseIndex] = {}
```
thomwolf, line 245: `Dict[str, Dict[str, BaseIndex]]` for `{column_name: {index_id: BaseIndex}}`?

with a default index_id like `faiss_index_1` i.e. `camel_case(BaseIndex.__class__.__name__) + f"_{len(self._indexes[column_name])}"` and no need to give the name of the index if you only have one?

We can also keep that for later and wait for people to ask for several indexes on the same columns in the issues (not sure it's really useful at the moment).

src/nlp/search.py
```diff
@@ -0,0 +1,366 @@
...
+            )
+
+    def list_indexes(self) -> List[str]:
+        """List the columns/identifiers of all the attached indexes."""
+        return list(self._indexes)
+
+    def add_vector_index(
+        self,
+        column: str,
+        vectors,
```
thomwolf, line 266: Shall we give this option to supply external vectors?

Not sure I see how this integrates with our usual "all is in the table" approach.

src/nlp/search.py
```diff
@@ -0,0 +1,366 @@
...
+        """ Add a dense index using Faiss for fast retrieval.
+
+            Args:
+                `column` (`str`): The column/identifier of the index. This is the column that is used to call `.get_nearest` or `.search`.
+                `vectors` (`Union[np.ndarray, nlp.Dataset]`): The vectors to index. It can be a `nlp.Dataset` if the `dataset` is well formated
+                    using `dataset.set_format("numpy")` for example.
+                `device` (Optional `int`): If not None, this is the index of the GPU to use. By default it uses the CPU.
+                `string_factory` (Optional `str`): This is passed to the index factory Faiss to create the index. Default index class is IndexFlatIP.
+                `faiss_gpu_options` (Optional `FaissGpuOptions`): Options to configure the GPU resources of Faiss.
+                `column` (Optional `str`): In the case of a `nlp.Dataset` input, which column to index.
```
thomwolf, line 280: duplicate

src/nlp/search.py
```diff
@@ -0,0 +1,366 @@
...
+
+    def add_text_index(self, column: str, texts, es_client, index_name):
+        """ Add a text index using ElasticSearch for fast retrieval.
+
+            Args:
+                `column` (`str`): The column/identifier of the index. This is the column that is used to call `.get_nearest` or `.search`.
+                `texts` (`Union[List[str], nlp.Dataset]`): The texts to index. It can be a `nlp.Dataset`.
+                `es_client` (`elasticsearch.Elasticsearch`): The elasticsearch client used to create the index.
+                `index_name` (Optional `str`): The elasticsearch index name used to create the index.
+                `column` (Optional `str`): In the case of a `nlp.Dataset` input, which column to index.
```
thomwolf, line 293: duplicate

src/nlp/search.py
```diff
@@ -0,0 +1,366 @@
...
+        """
+        self._indexes[column] = DenseIndex(device, string_factory, faiss_gpu_options)
+        self._indexes[column].add_vectors(vectors, column=column)
+
+    def add_text_index(self, column: str, texts, es_client, index_name):
+        """ Add a text index using ElasticSearch for fast retrieval.
+
+            Args:
+                `column` (`str`): The column/identifier of the index. This is the column that is used to call `.get_nearest` or `.search`.
+                `texts` (`Union[List[str], nlp.Dataset]`): The texts to index. It can be a `nlp.Dataset`.
```
thomwolf, line 290: here as well, I'm not sure I understand how the option to add new texts works with our philosophy of having all in the Arrow table.

src/nlp/search.py
```diff
@@ -0,0 +1,366 @@
...
+            Args:
+                `column` (`str`): The column/identifier of the index.
+                `query` (`Union[str, np.ndarray]`): The query as a string if `column` is a text index or as a numpy array if `column` is a vector index.
+                `k` (`int`): The number of examples to retrieve.
+
+            Ouput:
+                `scores` (`List[List[float]`): The retrieval scores of the retrieved examples.
+                `examples` (`List[List[dict]]`): The retrieved examples.
+        """
+        self._check_index_is_initialized(column)
```
thomwolf, line 348: (nit-pick) you're checking this two times here

src/nlp/search.py
```diff
@@ -0,0 +1,366 @@
...
+            Args:
+                `column` (`str`): The column/identifier of the index.
+                `queries` (`Union[List[str], np.ndarray]`): The queries as a list of strings if `column` is a text index or as a numpy array if `column` is a vector index.
+                `k` (`int`): The number of examples to retrieve per query.
+
+            Ouput:
+                `scores` (`List[List[float]`): The retrieval scores of the retrieved examples per query.
+                `examples` (`List[List[dict]]`): The retrieved examples per query.
+        """
+        self._check_index_is_initialized(column)
```
thomwolf, line 364: (nit-pick) you're checking this two times here

src/nlp/arrow_dataset.py
```diff
@@ -118,12 +125,14 @@ def __init__(
         info: Optional[Any] = None,
         split: Optional[Any] = None,
     ):
-        super().__init__(info=info, split=split)
+        DatasetInfoMixin.__init__(self, info=info, split=split)
+        IndexableMixin.__init__(self)
```
thomwolf, line 129: I'm a tiny bit worried about this multiple inheritance scenario but ok, let's see how it goes for now.

src/nlp/arrow_dataset.py
```diff
@@ -247,11 +261,46 @@ def __repr__(self):
...
+    ):
+        """ To be used in a `with` statement. Set __getitem__ return format (type and columns)
+
+            Args:
+                type (Optional ``str``): output type selected in [None, 'numpy', 'torch', 'tensorflow', 'pandas']
+                    None means __getitem__ returns python objects (default)
+                columns (Optional ``List[str]``): columns to format in the output
+                    None means __getitem__ returns all columns (default)
+                output_all_columns (``bool`` default to False): keep un-formated columns as well in the output (as python objects)
+                format_kwargs ()
```
thomwolf, line 285: Document a bit `format_kwargs`

src/nlp/arrow_dataset.py
```diff
@@ -606,15 +691,22 @@ def apply_function_on_filtered_inputs(inputs, indices):
...
                 indices = list(range(*(slice(i, i + batch_size).indices(self._data.num_rows))))  # Something simpler?
-                batch = apply_function_on_filtered_inputs(batch, indices)
+                try:
+                    batch = apply_function_on_filtered_inputs(
+                        batch, indices, check_same_num_examples=len(self.list_indexes()) > 0
+                    )
+                except NumExamplesMismatch:
+                    raise DatasetTransformationNotAllowedError(
+                        "Using `.map` in batched mode on a dataset with attached indexes is allowed only if it doesn't create or remove existing examples."
+                    )
```
thomwolf, line 709: Clean!

Maybe you can add that the user can `drop_index` before doing so and recreate the index afterward (and same for all the other places where you forbid operating on a dataset with an index)

src/nlp/arrow_dataset.py
```diff
@@ -1075,3 +1190,54 @@ def train_test_split(
...
+            ```
+
+            Args:
+                `column` (`str`): The column of vectors to index.
+                `device` (Optional `int`): If not None, this is the index of the GPU to use. By default it uses the CPU.
+                `string_factory` (Optional `str`): This is passed to the index factory Faiss to create the index. Default index class is IndexFlatIP.
+                `faiss_gpu_options` (Optional `FaissGpuOptions`): Options to configure the GPU resources of Faiss.
+                `dtype` (data-type): The dtype of the numpy arrays that are indexed. Default is np.float32.
+        """
+        with self.formated_as(type="numpy", columns=[column], dtype=dtype):
```
thomwolf, line 1220: Clean!

src/nlp/arrow_dataset.py
```diff
@@ -1075,3 +1190,54 @@ def train_test_split(
         )
 
         return {"train": train_split, "test": test_split}
+
+    def add_vector_index(
```
thomwolf, line 1194: I think I would call it `add_faiss_index`.

In general, we may want to be compatible with a diversity of ann-engines. We'll probably have a way to register ann engine at some point and call a `dataset.add_index(registered_engine, **kwargs)`. But probably for now we can just have a simple API focusing on FAISS and ES and leave this for future work after we've had some feedback from the community.

src/nlp/arrow_dataset.py
```diff
@@ -1075,3 +1190,54 @@ def train_test_split(
...
+                `device` (Optional `int`): If not None, this is the index of the GPU to use. By default it uses the CPU.
+                `string_factory` (Optional `str`): This is passed to the index factory Faiss to create the index. Default index class is IndexFlatIP.
+                `faiss_gpu_options` (Optional `FaissGpuOptions`): Options to configure the GPU resources of Faiss.
+                `dtype` (data-type): The dtype of the numpy arrays that are indexed. Default is np.float32.
+        """
+        with self.formated_as(type="numpy", columns=[column], dtype=dtype):
+            super().add_vector_index(column, self, device, string_factory, faiss_gpu_options)
+        return self
+
+    def add_text_index(self, column: str, es_client, index_name):
```
thomwolf, line 1224: here I would have `add_elasticsearch_index`

src/nlp/search.py
```diff
@@ -0,0 +1,366 @@
...
+
+
+class MissingIndex(Exception):
+    pass
+
+
+class BaseIndex:
+    """Base class for indexing"""
+
+    def search(self, query, k: int = 10) -> Tuple[List[float], List[int]]:
```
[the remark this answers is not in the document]
lhoestq: In this particular case the signature of `search` is a convention widely used so I'm not sure it will change.
However I agree that using NamedTuples can make it more readable as a documentation.

src/nlp/search.py
```diff
@@ -0,0 +1,366 @@
...
+    An Elasticsearch server needs to be accessible, and a python client is declared with
+    ```
+    es_client = Elasticsearch([{'host': 'localhost', 'port': '9200'}])
+    ```
+    for example.
+    """
+
+    def __init__(self, es_client, index_name: str):
+        self.es_client = es_client
+        self.index_name = index_name
```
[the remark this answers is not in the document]
lhoestq: Yep I'll try to find something that is specific for the library. Maybe something like `huggingface_nlp_{cache_file_name}_{column}`

src/nlp/search.py
```diff
@@ -0,0 +1,366 @@
...
+        assert (
+            _has_elasticsearch
+        ), "You must install ElasticSearch to use SparseIndex. To do so you can run `pip install elasticsearch`"
+
+    def add_texts(self, texts: Union[List[str], "Dataset"], column: Optional[str] = None):
```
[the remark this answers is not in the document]
lhoestq: `index_name` can already be specified when the user adds the index.
Why not add allow to set a custom index_config too

src/nlp/search.py
```diff
@@ -0,0 +1,366 @@
...
+        """
+        response = self.es_client.search(
+            index=self.index_name,
+            body={"query": {"multi_match": {"query": query, "fields": ["text"], "type": "cross_fields"}}, "size": k},
+        )
+        hits = response["hits"]["hits"]
+        return [hit["_score"] for hit in hits], [hit["_id"] for hit in hits]
+
+
+class FaissGpuOptions:
```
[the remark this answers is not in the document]
lhoestq: Right 👍 
Also the typings of the attributes will be `Any` in that case, as they are many different faiss objects that could be given.

src/nlp/search.py
```diff
@@ -0,0 +1,366 @@
...
+    More info at https://github.com/facebookresearch/faiss/wiki/Faiss-on-the-GPU
+    """
+
+    def __init__(self, resource_vec, device_vec, cloner_options):
+        self.resource_vec = resource_vec
+        self.device_vec = device_vec
+        self.cloner_options = cloner_options
+
+
+class DenseIndex(BaseIndex):
```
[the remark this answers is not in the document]
lhoestq: Ok with FaissIndex 👍

src/nlp/search.py
lhoestq: Good catch on the typing :)
Not sure if we want to have farfetched names as it can be confusing when trying to query the right index.
I like the idea of not having to provide the name if we only have one index

src/nlp/search.py
lhoestq: I guess we can add it for now as it can be quite convenient, but we shouldn't expect to add more support for external things as it can end up out of scope pretty quickly.

src/nlp/search.py
lhoestq: For vectors I can understand that there can be use cases where you'd want to use your own vectors/embeddings but not sure about text...

src/nlp/search.py
lhoestq: Better twice than none x)

src/nlp/search.py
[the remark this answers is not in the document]
patrickvonplaten: Agree, NamedTuples have 0 disadvantages IMO

src/nlp/utils/py_utils.py
```diff
@@ -37,14 +38,15 @@
 memoize = functools.lru_cache
 
 
-def convert_tuples_in_lists(data_struct):
+def convert_sequences_in_lists(data_struct):
```
patrickvonplaten, line 41: (nit) `convert_sequences_to_lists`

patrickvonplaten: Ah, never mind maybe better `map_all_sequences_to_lists`

src/nlp/search.py
```diff
@@ -0,0 +1,366 @@
...
+                `k` (`int`): The number of examples to retrieve per query.
+
+            Ouput:
+                `scores` (`List[List[float]`): The retrieval scores of the retrieved examples per query.
+                `indices` (`List[List[int]]`): The indices of the retrieved examples per query.
+        """
+        self._check_index_is_initialized(column)
+        return self._indexes[column].search_batch(queries, k)
+
+    def get_nearest(self, column: str, query, k: int = 10) -> Tuple[List[float], List[dict]]:
```
patrickvonplaten, line 336: (nit) `get_nearest_example` is better IMO.

patrickvonplaten: not a fan of single letter variables `k` -> why not type checking for `query`?

patrickvonplaten:
Looks very cool! Only looked at it superficially though

src/nlp/search.py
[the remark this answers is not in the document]
lhoestq: actually I'll do something like `huggingface_nlp_tmpixfihib0` if nothing is specified (using tempfile to get the hash)

src/nlp/search.py
lhoestq: I think I'd rather allow the user to specify a `custom_index_name` (defaults to `column_name`) rather than having this `column_name>index_id` logic actually

src/nlp/search.py
lhoestq: `_check_index_is_initialized ` is actually used to factorize the `MissingIndex` error more than to check twice that the index has been initialized.

src/nlp/search.py
lhoestq: I like `get_nearest_examples ` too.
Let me type check query too.

About the name of `k`, it follows some naming convention widely used (cf 'k nearest neighbors').
Examples:
- sklearn.neighbors.KDTree.query uses `k`
- faiss.IndexFlat.search uses `k`
- annoy. AnnoyIndex.get_nns_by_vector uses `search_k `
- dlib/graphtools find_approximate_k_nearest_neighbors uses `k`

Sometimes it is called something different (`nn` for flann, `final_num_examples` for ScaNN...)

src/nlp/search.py
```diff
@@ -0,0 +1,435 @@
...
+        self,
+        host: Optional[str] = None,
+        port: Optional[int] = None,
+        es_client: Optional[Elasticsearch] = None,
+        index_name: Optional[str] = None,
+        index_config: Optional[dict] = None,
+    ):
+        assert (
+            _has_elasticsearch
+        ), "You must install ElasticSearch to use ElasticSearchIndex. To do so you can run `pip install elasticsearch`"
```
jplu, line 99: Maybe, to be more safe, we should propose a version of ES with which we know it works because I know by experience with ES that you can have big breaking changes across versions, even sometime across minor versions. And add it as well in the documentation.

This is really nice!! LGTM :)

lhoestq:
Alright I think I've checked all your comments, thanks :)

Moreover I just added a way to serialize faiss indexes.
This is important because for big datasets the index construction can take some time.

Examples:

```python
ds = nlp.load_dataset('crime_and_punish', split='train')
ds_with_embeddings = ds.map(lambda example: {'embeddings': embed(example['line']}))
ds_with_embeddings.add_faiss_index(column='embeddings')
# query
scores, retrieved_examples = ds_with_embeddings.get_nearest('embeddings', embed('my new query'), k=10)
# save index
ds_with_embeddings.get_index('embeddings').save('my_index.faiss')
```

```python
ds = nlp.load_dataset('crime_and_punish', split='train')
# load index
faiss_index = nlp.search.FaissIndex.load('my_index.faiss')
ds.add_faiss_index('embeddings', faiss_index=faiss_index)
# query
scores, retrieved_examples = ds_with_embeddings.get_nearest('embeddings', embed('my new query'), k=10)
```

Let me know what you think

thomwolf:
Nice!

Here are a few comments:

I think it would be good to separate (1) the name of the column we use for indexing and (2) the name of the index itself, at least in our head. As I understand it, once the index is created, the column we used to create it is irrelevant so the column name will only be relevant in the `add_faiss_index` and we should be able to supply a different index name, e.g. `my_faiss_index`. When we reload an index, we don't really care about the column that was used to create it, right? so it's maybe better to have an `index_name` (which default to the column name for a simple user experience but it can also be something else and this should be clear in our head when we define the API).

I'm wondering if we should not have a triple of methods for each retrieval engine: `add_xxx_index`, `save_xxx_index` and `load_xxx_index` when `xxx` can be `faiss` or `elasticsearch`. I'm not a fan of exposing `nlp.search.FaissIndex` unless you think there is a strong reason to have the user learn this abstraction.

Last but not least, I think we should already think about hosting index on our S3. I would maybe go for something like this: host the index serialized with the cached dataset on user-provided namespaces:
```python
wiki_indexed = load_dataset('thom/wiki_indexed_with_dpr_faiss')
```

lhoestq:
I agree, I just change to using `index_name` and having add/save/load methods

lhoestq:
To summarize:


```python
ds = nlp.load_dataset('crime_and_punish', split='train')
ds_with_embeddings = ds.map(lambda example: {'embeddings': embed(example['line']}))
ds_with_embeddings.add_faiss_index(column='embeddings')
# query
scores, retrieved_examples = ds_with_embeddings.get_nearest_examples('embeddings', embed('my new query'), k=10)
# save index
ds_with_embeddings.save_faiss_index('embeddings', 'my_index.faiss')
```

```python
ds = nlp.load_dataset('crime_and_punish', split='train')
# load index
ds.load_faiss_index('embeddings', 'my_index.faiss')
# query
scores, retrieved_examples = ds.get_nearest_examples('embeddings', embed('my new query'), k=10)
```

thomwolf:
Good to me. I understand that for now there is no check that the index matches the dataset on loading.
Maybe just add a basic test on the number of examples?

lhoestq:
Ok I think this one is ready now

lhoestq:
Looks like the CI is having troubles to pass because of `tests/test_dataset_common.py::AWSDatasetTest::test_builder_configs_{<insert_rando_dataset_name_here>}`, `requests.exceptions.ConnectionError` :/

Issue #238, opened 2020-06-05
[Metric] Bertscore : Warning : Empty candidate sentence; Setting recall to be 0.


Colanim:
When running BERT-Score, I'm meeting this warning :

> Warning: Empty candidate sentence; Setting recall to be 0.

Code :

```
import nlp
metric = nlp.load_metric("bertscore")
scores = metric.compute(["swag", "swags"], ["swags", "totally something different"], lang="en", device=0)
```

---

**What am I doing wrong / How can I hide this warning ?**

lhoestq:
This print statement comes from the official implementation of bert_score (see [here](https://github.com/Tiiiger/bert_score/blob/master/bert_score/utils.py#L343)). The warning shows up only if the attention mask outputs no candidate.
Right now we want to only use official code for metrics to have fair evaluations, so I'm not sure we can do anything about it. Maybe you can try to create an issue on their [repo](https://github.com/Tiiiger/bert_score) ?

PR #300, opened 2020-06-23
Fix bertscore references

### LINKED ISSUE #295: Improve input warning for evaluation metrics
Hi, 

I am the author of `bert_score`. Recently, we received [ an issue ](https://github.com/Tiiiger/bert_score/issues/62) reporting a problem in using `bert_score` from the `nlp` package (also see #238 in this repo).  After looking into this, I realized that the problem arises from the format `nlp.Metric` takes input. 

Here is a minimal example:
```python
import nlp

scorer = nlp.load_metric("bertscore")
with open("pred.txt") as p, open("ref.txt") as g:
    for lp, lg in zip(p, g):
        scorer.add(lp, lg)
score = scorer.compute(lang="en")
```

The problem in the above code is that `scorer.add()` expects a list of strings as input for the references. As a result, the `scorer` here would take a list of characters in `lg` to be the references. The correct implementation would be calling
```python
scorer.add(lp, [lg])
```

I just want to raise this issue to you to prevent future user errors of a similar kind. I assume some simple type checking can prevent this from happening?

Thanks!

lhoestq:
I added some type checking for metrics. There was an issue where a metric could interpret a string a a list. A `ValueError` is raised if a string is given instead of a list.

Moreover I added support for both strings and lists of strings for `references` in `bertscore`, as it is the case in the original code.

Both ways work:
```
import nlp

scorer = nlp.load_metric("bertscore")
with open("pred.txt") as p, open("ref.txt") as g:
    for lp, lg in zip(p, g):
        scorer.add(lp, [lg])
score = scorer.compute(lang="en")
```

```
import nlp

scorer = nlp.load_metric("bertscore")
with open("pred.txt") as p, open("ref.txt") as g:
    for lp, lg in zip(p, g):
        scorer.add(lp, lg)
score = scorer.compute(lang="en")
```

This should fix #295 and #238

### DIFF
--- a/src/nlp/metric.py
+++ b/src/nlp/metric.py
@@ -212,6 +212,7 @@ class Metric(object):
         """ Add one prediction and reference for the metric's stack.
         """
         example = {"predictions": prediction, "references": reference}
+        example = self.info.features.encode_example(example)
         if self.writer is None:
             self._init_writer()
         self.writer.write(example)
--- a/datasets/anli/anli.py
+++ b/datasets/anli/anli.py
@@ -18,8 +18,8 @@
 
 from __future__ import absolute_import, division, print_function
 
-import os
 import json
+import os
 
 import nlp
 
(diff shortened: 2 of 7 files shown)


src/nlp/metric.py
```diff
@@ -212,6 +212,7 @@ def add(self, prediction=None, reference=None, **kwargs):
         """ Add one prediction and reference for the metric's stack.
         """
         example = {"predictions": prediction, "references": reference}
+        example = self.info.features.encode_example(example)
```
thomwolf, line 215: Probably we'll want to move the info attributes one-level down in `Metrics` metrics like we did for `Dataset`.

PR #306, opened 2020-06-23
add pg19 dataset

lucidrains:
https://github.com/huggingface/nlp/issues/274

Add functioning PG19 dataset with dummy data

`cos_e.py` was just auto-linted by `make style`


datasets/pg19/pg19.py
```diff
@@ -0,0 +1,155 @@
...
+            citation=_CITATION,
+        )
+
+    def _split_generators(self, dl_manager):
+        """Returns SplitGenerators."""
+        # TODO(pg19): Downloads the data and defines the splits
+        # dl_manager is a nlp.download.DownloadManager that can be used to
+        # download and extract URLs
+
+        def get_filename(path):
```
mariamabarham, line 74: maybe this function can be defined outside of  `_split_generator` ? I mean outside the `class PG19`

datasets/pg19/pg19.py
```diff
@@ -0,0 +1,155 @@
...
+            return os.path.splitext(os.path.basename(path))[0]
+
+        def download_listdir(url, local_filepath):
```
mariamabarham, line 77: same as here?

datasets/pg19/pg19.py
```diff
@@ -0,0 +1,155 @@
...
+            results = tf.io.gfile.listdir(url)
+            with open(local_filepath, "w") as f:
+                f.write(json.dumps(results))
+            return local_filepath
+
+        def filepath_to_json(path):
```
mariamabarham, line 83: and here?

datasets/pg19/pg19.py
```diff
@@ -0,0 +1,155 @@
...
+        """Returns SplitGenerators."""
+        # TODO(pg19): Downloads the data and defines the splits
+        # dl_manager is a nlp.download.DownloadManager that can be used to
+        # download and extract URLs
+
+        def get_filename(path):
+            return os.path.splitext(os.path.basename(path))[0]
+
+        def download_listdir(url, local_filepath):
+            results = tf.io.gfile.listdir(url)
```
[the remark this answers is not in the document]
patrickvonplaten: It would be great if we can make this dataset `tf` independent. Would it be possible to just go with `glob.glob(url)` here?

datasets/pg19/pg19.py
patrickvonplaten: I actually think it's good with leaving the function in here since it's only going to be used inside the function.

datasets/pg19/pg19.py
patrickvonplaten: I actually think it's good with leaving the function in here since it's only going to be used inside the function.

datasets/pg19/pg19.py
patrickvonplaten: I actually think it's good with leaving the function in here since it's only going to be used inside the function.

patrickvonplaten:
@lucidrains - Thanks a lot for making the PR - PG19 is a super important dataset! Thanks for making it. Many people are asking for PG-19, so it would be great to have that in the library as soon as possible @thomwolf .

datasets/pg19/pg19.py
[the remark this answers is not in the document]
thomwolf: maybe @lhoestq has some idea on this?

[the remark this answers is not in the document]
lhoestq: TF is used here because if allows to browse the google storage as a filesystem to glob and download files.
This is the same for `boolq` if I'm not wrong.

To make it TF-independant we have to 
1) use public urls like `https://storage.googleapis.com/deepmind-gutenberg/train/{i}.txt` instead of `gs://deepmind-gutenberg/train/{i}.txt`, so that we we can use `dl_manager.download_and_extract`.
2) replace the `tf.io.gfile.listdir(url)` with something like `[get_url(i) for i in range(total_number_of_files)]`

src/nlp/utils/download_manager.py
```diff
@@ -20,6 +20,8 @@
 import logging
 import os
 
+from tqdm import tqdm
```
lhoestq, line 23:
```suggestion
from tqdm.auto import tqdm
```
for notebooks

lucidrains:
@mariamabarham yup! around 11GB!

lucidrains:
I'm looking forward to our first deep learning written novel already lol. It's definitely happening

datasets/pg19/pg19.py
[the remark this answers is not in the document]
lucidrains: @lhoestq for #2, we don't know which `i` falls into one of `/train`, `/test`, or `/validation`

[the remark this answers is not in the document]
lhoestq: Ok I see.
You can have the list by calling `https://storage.googleapis.com/storage/v1/b/deepmind-gutenberg/o?prefix=train` (you can use the `requests` package to call it), and take a look at the `name` attribute of each element.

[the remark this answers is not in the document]
lucidrains: @lhoestq very nice! I'll make the change

patrickvonplaten:
Good to merge IMO.

datasets/pg19/pg19.py
```diff
@@ -0,0 +1,179 @@
+"""TODO(pg19): Add a description here."""
```
thomwolf, line 1: Can you add a quick description here?

datasets/pg19/pg19.py
```diff
@@ -0,0 +1,179 @@
...
+_ASSET_ROOT_URL = "https://storage.googleapis.com/deepmind-gutenberg/"
+_STORAGE_API_ROOT_URL = "https://storage.googleapis.com/storage/v1/b/deepmind-gutenberg/o"
+
+_METADATA_URL = os.path.join(_ASSET_ROOT_URL, "metadata.csv")
+
+flat_map = lambda fn, arr: [el for sub_arr in map(fn, arr) for el in sub_arr]
+
+
+class Pg19(nlp.GeneratorBasedBuilder):
+    """TODO(pg19): Short description of my dataset."""
```
thomwolf, line 44: And a quick description here as well?

datasets/pg19/pg19.py
```diff
@@ -0,0 +1,179 @@
...
+  year = {2019},
+}
+
+"""
+
+# TODO(pg19):
+_DESCRIPTION = """\
+This repository contains the PG-19 language modeling benchmark.
+It includes a set of books extracted from the Project Gutenberg books library, that were published before 1919.
+It also contains metadata of book titles and publication dates.
```
thomwolf, line 32: Also, do you think you can add a word on the preprocessing of the dataset? I think you can add most of the information from: https://github.com/deepmind/pg19 actually.

All these elements on biases and limitations are important for all the users of the dataset.

```
PG-19 is over double the size of the Billion Word benchmark [2] and contains documents that are 20X longer, on average, than the WikiText long-range language modelling benchmark [3].

Books are partitioned into a train, validation, and test set. Book metadata is stored in metadata.csv which contains (book_id, short_book_title, publication_date).

Unlike prior benchmarks, we do not constrain the vocabulary size --- i.e. mapping rare words to an UNK token --- but instead release the data as an open-vocabulary benchmark. The only processing of the text that has been applied is the removal of boilerplate license text, and the mapping of offensive discriminatory words as specified by Ofcom [4] to placeholder tokens. Users are free to model the data at the character-level, subword-level, or via any mechanism that can model an arbitrary string of text.

To compare models we propose to continue measuring the word-level perplexity, by calculating the total likelihood of the dataset (via any chosen subword vocabulary or character-based scheme) divided by the number of tokens --- specified below in the dataset statistics table.

One could use this dataset for benchmarking long-range language models, or use it to pre-train for other natural language processing tasks which require long-range reasoning, such as LAMBADA [5] or NarrativeQA [6]. We would not recommend using this dataset to train a general-purpose language model, e.g. for applications to a production-system dialogue agent, due to the dated linguistic style of old texts and the inherent biases present in historical writing.
```

datasets/pg19/pg19.py
lucidrains: will do!

datasets/pg19/pg19.py
lucidrains: yup

lhoestq:
Oh I just noticed but as we changed the urls to download the files, we have to update `dataset_infos.json`.
Could you re-rurn `nlp-cli test ./datasets/pg19 --save_infos` ?

lucidrains:
@lhoestq on it!

lucidrains:
should be good!

patrickvonplaten:
@lhoestq - I think it's good to merge no?

lhoestq:
`dataset_infos.json` is still not up to date with the new urls (we can see that there are urls like `gs://deepmind-gutenberg/train/*` instead of `https://storage.googleapis.com/deepmind-gutenberg/train/*` in the json file)

Can you check that you re-ran the command to update the json file, and that you pushed the changes @lucidrains ?

lucidrains:
@lhoestq ohhh, I made the change in this commit https://github.com/lucidrains/nlp/commit/f3e23d823ad9942031be80b7c4e4212c592cd90c , that's interesting that the pull request didn't pick it up. maybe it's because I did it on another machine, let me check and get back to you!

lucidrains:
@lhoestq wrong branch 😅 thanks for catching!

lhoestq:
Awesome thanks 🎉

PR #309, opened 2020-06-24
Add narrative qa

Varal7:
Test cases for dummy data don't pass

Only contains data for summaries (not whole story)


ghomasHudson:
Does it make sense to download the full stories? I remember attempting to implement this dataset a while ago and ended up with something like:
```python

        dl_dir = dl_manager.download_and_extract(_DOWNLOAD_URL)
        data_dir = os.path.join(dl_dir, "narrativeqa-master")

        urls = {"test":{}, "train": {},"valid":{}}
        with open(os.path.join(data_dir,"documents.csv")) as f_in:
            csv_reader = csv.reader(f_in)
            next(csv_reader) # discard header row
            for i,row in enumerate(csv_reader):
                if i > 1572:
                    break
                if row != []:
                    urls[row[1]][row[0]] = row[3]

        url_files = {}
        for key in urls.keys():
            url_files[key] = dl_manager.download_and_extract(urls[key])

        return [
            nlp.SplitGenerator(
                name=nlp.Split.TRAIN,
                gen_kwargs={
                    "data_dir":data_dir,
                    "split":"train",
                    "doc_id_to_path":url_files["train"]
                    }
            ),
        ....
```
It does end up cluttering your huggingface cache dir though.

ghomasHudson:
Also since answer_1 and answer_2 are basically the same it might make sense to combine them:
```python
"answers": nlp.features.Sequence({
    "text": nlp.Value("string"),
    "tokenized": nlp.features.Sequence(nlp.Value("string"))
})
```
(the tokenized features should also probably be lists of strings not strings)

Again, this is a personal preference thing, but it might be useful to combine the document-related features:
```python
{
    "document": {
        "id": nlp.Value("string"),
        "kind": nlp.Value("string"),
        "url": nlp.Value("string"),
        "file_size": nlp.Value("int32"),
        "word_count": nlp.Value("int32"),
        "start": nlp.Value("string"),
        "end": nlp.Value("string"),
        "wiki_url": nlp.Value("string"),
        "wiki_title": nlp.Value("string"),
        "summary": nlp.features.Sequence({
            "text": nlp.Value("string"),
            "tokenized": nlp.features.Sequence(nlp.Value("string"))
        }),
        "text": nlp.Value("string"),
    },
    "question": nlp.features.Sequence({
        "text": nlp.Value("string"),
        "tokenized": nlp.features.Sequence(nlp.Value("string"))
    }),
    "answers": nlp.features.Sequence({
        "text": nlp.Value("string"),
        "tokenized": nlp.features.Sequence(nlp.Value("string"))
    })
}
```

lhoestq:
Did you manage to fix the dummy data @Varal7

ghomasHudson:
@lhoestq do you think it's acceptable for the `dl_manager` to go grab all the individual stories from project gutenburg? I've got a working version of that but it does clutter up your huggingface cache somewhat.

The real value (and original purpose) of this dataset is doing question answering on the full text.

lhoestq:
> @lhoestq do you think it's acceptable for the `dl_manager` to go grab all the individual stories from project gutenburg? I've got a working version of that but it does clutter up your huggingface cache somewhat.
> 
> The real value (and original purpose) of this dataset is doing question answering on the full text.

What's the problem exactly with the cache ?

ghomasHudson:
Nothing, just that because each story is a separate download it gets a bit messy as all 1573 files are under `~/.cache/hugginface/datasets` rather than organized under a subdir.

Probably doesn't matter to the end user though.

lhoestq:
Yea I agree it's a mess. I just created #393 to make things easier.

lhoestq:
I got the PR merged to have a cleaner the cache directory (everything is downloaded inside the 'downloads' sub-directory).
Feel free to download all the stories then @ghomasHudson @Varal7  x)
If you have the possibility of downloading a compressed file with most of the stories at once it would be better though.

ghomasHudson:
Good stuff. The problem I'm having at the moment is that stories from project Gutenberg occasionally fail. All books are out of copyright so we should be able to host them. 

Here's a zip file of the full text if we have anywhere to put them: https://drive.google.com/file/d/17jOR7NqvzDwSlPXrlHaYV-PGI8JG-KY5/view?usp=sharing

lhoestq:
I put the zip file here @ghomasHudson 
https://storage.googleapis.com/huggingface-nlp/datasets/narrative_qa/narrativeqa_full_text.zip

Sorry for the delay

thomwolf:
Closing in favor of #499

PR #316, opened 2020-06-26
add AG News dataset

jxmorris12:
adds support for the AG-News topic classification dataset


datasets/ag_news/ag_news.py
```diff
@@ -0,0 +1,107 @@
...
+  booktitle={NIPS},
+  year={2015}
+}
+"""
+
+_TRAIN_DOWNLOAD_URL = "https://raw.githubusercontent.com/mhjabreel/CharCnn_Keras/master/data/ag_news_csv/train.csv"
+_TEST_DOWNLOAD_URL = "https://raw.githubusercontent.com/mhjabreel/CharCnn_Keras/master/data/ag_news_csv/test.csv"
+
+
+class AGNewsConfig(nlp.BuilderConfig):
```
mariamabarham, line 56: I think this class can be removed see that it's not used? Or you can refer to the [squad](https://github.com/huggingface/nlp/blob/master/datasets/squad/squad.py) dataset on how you can use the `config class`

datasets/ag_news/ag_news.py
```diff
@@ -0,0 +1,107 @@
...
+        return nlp.DatasetInfo(
+            description=_DESCRIPTION,
+            features=nlp.Features(
+                {"text": nlp.Value("string"), "label": nlp.features.ClassLabel(names=["World", "Sports", "Business", "Sci/Tech"])}
+            ),
+            homepage="http://groups.di.unipi.it/~gulli/AG_corpus_of_news_articles.html",
+            citation=_CITATION,
+        )
+
+    def _vocab_text_gen(self, train_file):
```
mariamabarham, line 83: This function is not used also I think?

mariamabarham:
Thanks @jxmorris12 for adding this adding. 
Can you please add a small description of the PR?

datasets/ag_news/ag_news.py
jxmorris12: thanks, I removed it!

datasets/ag_news/ag_news.py
jxmorris12: I removed this too! I thought I needed it, sorry

datasets/ag_news/ag_news.py
mariamabarham: great, thanks

datasets/rotten_tomatoes/rotten_tomatoes.py
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
```
ghomasHudson, line 14: Did you mean to include rotten_tomatoes in here as well as AG News?

PR #333, opened 2020-07-01
fix variable name typo

stas00:


lhoestq:
Good catch :)
I think there is another occurence that needs to be fixed in the second gist (line 4924 of the notebook file):
```python
bleu = nlp.load_metric(...)
```

lhoestq:
Was fixed in e16f79b5f7fc12a6a30c777722be46897a272e6f
Closing it.

PR #340, opened 2020-07-03
Update cfq.py

brainshawn:
Make the dataset name consistent with in the paper:  Compositional Freebase Question => Compositional Freebase Questions.


mariamabarham:
Thanks @brainshawn  for this update

Issue #336, opened 2020-07-02
[Dataset requests] New datasets for Open Question Answering


thomwolf:
We are still a few datasets missing for Open-Question Answering which is currently a field in strong development.

Namely, it would be really nice to add:
- WebQuestions (Berant et al., 2013) [done]
- CuratedTrec (Baudis et al. 2015) [not open-source]
- MS-MARCO (NGuyen et al. 2016) [done]
- SearchQA (Dunn et al. 2017) [done]
- FEVER (Thorne et al. 2018) - [ done]

 

All these datasets are cited in http://arxiv.org/abs/2005.11401

PR #341, opened 2020-07-03
add fever dataset

mariamabarham:
This PR add the FEVER dataset https://fever.ai/ used in with the paper: FEVER: a large-scale dataset for Fact Extraction and VERification (https://arxiv.org/pdf/1803.05355.pdf).
#336


datasets/fever/fever.py
```diff
@@ -0,0 +1,224 @@
...
+
+        Args:
+          **kwargs: keyword arguments forwarded to super.
+        """
+        super(FeverConfig, self).__init__(**kwargs)
+        
+class Fever(nlp.GeneratorBasedBuilder):
+    """Fact Extraction and VERification Dataset."""
+
+    # Version 1.1.0 removes web contents.
```
lhoestq, line 57: What about v1.1.0 then ?

PR #344, opened 2020-07-06
Search qa

mariamabarham:
This PR adds the Search QA dataset used in **SearchQA: A New Q&A Dataset Augmented with Context from a Search Engine**. The dataset has the following config name:

- raw_jeopardy: raw data

- train_test_val:  which is the splitted version

#336


lhoestq:
Could you rebase from master just to make sure we won't break anything for `fever` pls @mariamabarham /

PR #354, opened 2020-07-08
More faiss control

lhoestq:
Allow users to specify a faiss index they created themselves, as sometimes indexes can be composite for examples


lhoestq:
> Ok, so we're getting rid of the `FaissGpuOptions`?

We support `device=...` because it's simple, but faiss GPU options can be used in so many ways (you can set different gpu options for the different parts of your index for example) that it's probably better to let the user create and configure its index and then use `custom_index=...`

PR #357, opened 2020-07-08
Add hashes to cnn_dailymail

jbragg:
The URL hashes are helpful for comparing results from other sources.


lhoestq:
Looks you to me :)

Could you also update the json file that goes with the dataset script by doing 
```
nlp-cli test ./datasets/cnn_dailymail --save_infos --all_configs
```
It will update the features metadata and the size of the dataset with your changes.

jbragg:
@lhoestq I ran that command.

Thanks for the helpful repository!

PR #370, opened 2020-07-09
Allow indexing Dataset via np.ndarray

jarednielsen:


jarednielsen:
Looks like a flaky CI, failed download from S3.

src/nlp/arrow_dataset.py
```diff
@@ -523,7 +523,7 @@ def _getitem(
                     outputs = self._data[key].to_pandas(split_blocks=True).to_list()
             else:
                 outputs = self._data[key].to_pandas(split_blocks=True).to_list()
-        elif isinstance(key, list):
+        elif isinstance(key, list) or isinstance(key, np.ndarray):
             data_subset = pa.concat_tables(self._data.slice(i, 1) for i in key)
```
thomwolf, line 527: Here we can even extend to torch and tensorflow indexing with the following:
```python
from collections.abc import Iterable
...
        elif isinstance(key, Iterable):
            data_subset = pa.concat_tables(self._data.slice(int(i), 1) for i in key)
```
(the `int(i)` is important)

jarednielsen: Nice! If suppose someone could pass in a float, but that's such a blatant error it's not worth checking.

PR #372, opened 2020-07-10
Make the json script more flexible

thomwolf:
Fix https://github.com/huggingface/nlp/issues/359
Fix https://github.com/huggingface/nlp/issues/369

JSON script now can accept JSON files containing a single dict with the records as a list in one attribute to the dict (previously it only accepted JSON files containing records as rows of dicts in the file).

In this case, you should indicate using `field=XXX` the name of the field in the JSON structure which contains the records you want to load. The records can be a dict of lists or a list of dicts.

E.g. to load the SQuAD dataset JSON (without using the `squad` specific dataset loading script), in which the data rows are in the `data` field of the JSON dict, you can do:
```python
from nlp import load_dataset
dataset = load_dataset('json', data_files='/PATH/TO/JSON', field='data')
```

### DIFF
--- a/datasets/json/json.py
+++ b/datasets/json/json.py
@@ -1,13 +1,12 @@
 # coding=utf-8
 
+import json
 from dataclasses import dataclass
-from typing import Union, List
 from io import BytesIO
+from typing import List, Union
 
 import pyarrow as pa
 import pyarrow.json as paj
-
-import json
 
 import nlp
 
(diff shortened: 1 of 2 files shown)


datasets/json/json.py
```diff
@@ -12,23 +15,30 @@
 class JsonConfig(nlp.BuilderConfig):
     """BuilderConfig for JSON."""
 
-    read_options: paj.ReadOptions = paj.ReadOptions()
-    parse_options: paj.ParseOptions = paj.ParseOptions()
+    features: nlp.Features = None
+    field: str = None
+    use_threads: bool = True
+    block_size: int = None
+    newlines_in_values: bool = None
```
lhoestq, line 22: Do you think we could also add the `encoding` as a parameter ?

thomwolf: You mean when we default on python reading logic?
I'm tented to always have `utf-8` as default in all the scripts but maybe we could have that as a generic arguments 🤔

PR #386, opened 2020-07-13
Update dataset loading and features - Add TREC dataset

thomwolf:
This PR:
- add a template for a new dataset script
- update the caching structure so that the path to the cached data files is also a function of the dataset loading script hash. This way when you update a loading script the data will be automatically updated instead of falling back to the previous version (which is usually a outdated). This makes it in particular easier to iterate when writing a new dataset loading script.
- fix a bug in the `ClassLabel` feature and make it more flexible so that its methods `str2int` and `int2str` can also accept list, numpy arrays and PyTorch/TensorFlow tensors.
- add the TREC-6 dataset


src/nlp/builder.py
```diff
@@ -129,6 +129,7 @@ def __init__(
         """
```
patrickvonplaten, line 129: would be good to add a `hash` Args docstring here

src/nlp/builder.py
```diff
@@ -242,17 +243,21 @@ def cache_dir(self):
...
-        if not with_version:
-            return builder_data_dir
-
-        version = self.config.version
-        version_data_dir = os.path.join(builder_data_dir, str(version))
-        return version_data_dir
+        if with_version:
+            builder_data_dir = os.path.join(builder_data_dir, str(self.config.version))
+        if hash:
+            builder_data_dir = os.path.join(builder_data_dir, hash)
```
patrickvonplaten, line 259: maybe a check that `hash` is a string here would be useful

src/nlp/metric.py
```diff
@@ -50,6 +51,8 @@ def __init__(
                     (in particular non-additive metrics like F1).
                 data_dir (str): path to a directory in which temporary data will be stored.
                     This should be a shared file-system for distributed setups.
+                hash (str): can be used to define a hash specific to the metrics computation script
+                    This avoid overriding the data of one metric when the metric loading script is modified.
```
patrickvonplaten, line 55:
```suggestion
                    This prevents the metric's data to be overridden when the metric loading script is modified.
```

src/nlp/commands/test.py
```diff
@@ -70,14 +70,14 @@ def run(self):
             print("Both parameters `config` and `all_configs` can't be used at once.")
             exit(1)
         path, name = self._dataset, self._name
-        module_path = prepare_module(path)
+        module_path, hash = prepare_module(path)
         builder_cls = import_main_class(module_path)
         builders: List[DatasetBuilder] = []
         if self._all_configs and len(builder_cls.BUILDER_CONFIGS) > 0:
             for config in builder_cls.BUILDER_CONFIGS:
                 builders.append(builder_cls(name=config.name, data_dir=self._data_dir))
```
lhoestq, line 78:
```suggestion
                builders.append(builder_cls(name=config.name, hash=hash, data_dir=self._data_dir))
```

lhoestq:
I just copied the files that are on google storage to follow the new `_relative_data_dir ` format. It should be good to merge now :)

Well actually it seems there are some merge conflicts to fix first

Issue #342, opened 2020-07-06
Features should be updated when `map()` changes schema


thomwolf:
`dataset.map()` can change the schema and column names.

We should update the features in this case (with what is possible to infer).

lhoestq:
`dataset.column_names` are being updated but `dataset.features` aren't indeed...

PR #390, opened 2020-07-14
Concatenate datasets

jarednielsen:
I'm constructing the "WikiBooks" dataset, which is a concatenation of Wikipedia & BookCorpus. So I implemented the `Dataset.from_concat()` method, which concatenates two datasets with the same schema.

This would also be useful if someone wants to pretrain on a large generic dataset + their own custom dataset. Not in love with the method name, so would love to hear suggestions.

Usage:
```python
from nlp import Dataset, load_dataset

data1, data2 = {"id": [0, 1, 2]}, {"id": [3, 4, 5]}
dset1, dset2 = Dataset.from_dict(data1), Dataset.from_dict(data2)
dset_concat = Dataset.from_concat([dset1, dset2])
print(dset_concat)
# Dataset(schema: {'id': 'int64'}, num_rows: 6)
```


lhoestq:
Looks cool :)

I feel like 
```python
concatenated_dataset = dataset1.concatenate(dataset2)
```
could be more natural. What do you think ?

Also could you also concatenate the `nlp.Dataset._data_files` ?
```python
return cls(table, info=info, split=split, data_files=self._data_files + other_dataset._data_files)
```

lhoestq:
I feel like "WikiBooks" would be a multi task dataset that could fit in the #217 discussion.
Not sure concatenate should be the solution for a multi task dataset.

jarednielsen:
Thanks for the suggestion! `dset1.concatenate(dset2)` does feel more natural. Although this seems to be a different "class" of transformation function than map() or filter(), acting on two datasets rather than on one. I would prefer the function signature treat both datasets symmetrically.

Python lists have `list1 + list2` or `list1.extend(list2)`.
NumPy has `np.concatenate((arr1, arr2))`.
Pandas has `pd.join((df1, df2))`.
PyTorch has `ConcatDataset((dset1, dset2))`.

Given the symmetrical treatment and clear communication that this creates a new object, rather than a simple chaining on the first, my preference is now for `nlp.concatenate((dset1, dset2))`. Does that work?

jarednielsen:
The multi-task discussion is interesting, thanks for pointing me to that! I'll be focusing on T5 in a few weeks, so I'm sure I'll have many opinions then :). For now, I think a simple concatenate feature is important and orthogonal to that discussion. For example, a user may want to create a custom dataset that joins Wikipedia with their own custom text.

lhoestq:
> Given the symmetrical treatment and clear communication that this creates a new object, rather than a simple chaining on the first, my preference is now for `nlp.concatenate((dset1, dset2))`. This would place the function in the same API class as `nlp.load_dataset`. Does that work?

Yep I like this idea. Maybe `nlp.concatenate_datasets()` ?

> For now, I think a simple concatenate feature is important and orthogonal to that discussion. For example, a user may want to create a custom dataset that joins Wikipedia with their own custom text.

I agree :)

jarednielsen:
Great, just updated!

src/nlp/load.py
```diff
@@ -530,3 +532,23 @@ def load_dataset(
...
+        features (:obj:``nlp.Features``, `optional`, defaults to :obj:``None``): If specified, the features types of the dataset
+        info (:obj:``nlp.DatasetInfo``, `optional`, defaults to :obj:``None``): If specified, the dataset info containing info like
+            description, citation, etc.
+        split (:obj:``nlp.NamedSplit``, `optional`, defaults to :obj:``None``): If specified, the name of the dataset split.
+    """
+    if not all([dset.schema == dsets[0].schema for dset in dsets]):
+        raise ValueError("Schema must match for all datasets")
+    table = pa.concat_tables([dset._data for dset in dsets])
+    data_files = list(itertools.chain.from_iterable([dset._data_files for dset in dsets]))
+    return Dataset(table, info=info, split=split, data_files=data_files)
```
thomwolf, line 554: I think it would be nice to try keep some info by default instead of losing them by default if possible.
We try to care a lot about dataset metadata and citations in the library.

Maybe we could concatenate most of the dataset info, attributes by attributes?
E.g. the citations can be concatenated, the description probably as well.

jarednielsen: Seems like we want a `merge_dataset_infos()` method. I could take a crack, but you're probably better suited to tinker around with what format you like best.

jarednielsen: Done, let me know if that's what you had in mind.

src/nlp/info.py
```diff
@@ -137,6 +137,25 @@ def _dump_license(self, file):
...
 
+    @classmethod
+    def from_merge(cls, dataset_infos: List["DatasetInfo"]):
+        dataset_infos = [dset_info for dset_info in dataset_infos if dset_info is not None]
+        description = "\n\n".join([info.description for info in dataset_infos])
+        citation = "\n\n".join([info.citation for info in dataset_infos])
+        homepage = "\n\n".join([info.homepage for info in dataset_infos])
+        license = "\n\n".join([info.license for info in dataset_infos])
+        features = None
+        supervised_keys = None
```
lhoestq, line 148: For `features` and `supervised_keys` we should probably check that they're all the same, and then take the first one.

jarednielsen: What's the difference between a dataset schema and dataset features? I see that the schema is an underlying Arrow object, while features is a Python class, but they seem to contain the same information.

lhoestq: > What's the difference between a dataset schema and dataset features?

`schema` is for the logic on the arrow side, while `features` is the logic from our side.

`nlp.Features` are simpler to create that `schema`, and allow to encode example in the correct arrow format.
`pyarrow.Schema` is used to write the arrow files.

Currently the `schema` is a public attribute of `nlp.Dataset` but we plan to hide it and only show `features` to the user. We have `nlp.Features` methods that allow to convert `features` from and to pyarrow's `schema`.

> It looks like doing a map changes the schema but not the features

Indeed x) We have an issue about it: #342 

We have to do changes in the `features` logic:
- make `features` the source of truth (and make the logic `schema` logic appear only in the `ArrowWriter`)
- make all the dataset transforms update `features` accordingly (we can infer features from the output of the given function in map for example).

> Since concatenate_datasets() will often be used in a preprocessing pipeline with many dataset transformations, comparing features is not the right behavior. It seems like since schema is the source of truth, we should rely on that and ignore the potentially incorrect features

You're right. Right now `features` can be incorrect. Until we do the changes in the `features` logic, `schema` stays the source of truth.

Let's ignore the `features` and `supervised_keys ` for now in `concatenate_datasets` then.

PR #392, opened 2020-07-15
Style change detection

ghomasHudson:
Another [PAN task](https://pan.webis.de/clef20/pan20-web/style-change-detection.html). This time about identifying when the style/author changes in documents.

- There's the possibility of adding the [PAN19](https://zenodo.org/record/3577602) and PAN18 style change detection tasks too (these are datasets whose labels are a subset of PAN20's). These would probably make more sense as separate datasets (like wmt is now)
- I've converted the integer 0,1 values to a boolean
- Using manually downloaded data again. This might be changed at some point following the discussion in https://github.com/huggingface/nlp/pull/349.


datasets/style_change_detection/style_change_detection.py
```diff
@@ -0,0 +1,127 @@
...
+            features=nlp.Features(features),
+            homepage="https://pan.webis.de/clef20/pan20-web/style-change-detection.html",
+            citation=_CITATION,
+        )
+
+    def _split_generators(self, dl_manager):
+        """Returns SplitGenerators."""
+        data_dir = os.path.abspath(os.path.expanduser(dl_manager.manual_dir))
+        train_dir = os.path.join(data_dir, "train", "dataset-" + self.config.name)
+        val_dir = os.path.join(data_dir, "train", "dataset-" + self.config.name)
```
lhoestq, line 97: We could check that the files exist here, and raise the adequate error if they don't exist, as we do in `xtreme.py` for example.

Something like:
```
            if not os.path.exists(train_dir):
                raise FileNotFoundError(
                    "{} does not exist. Make sure you insert a manual dir via `nlp.load_dataset('style_change_detection', data_dir=...)` that includes {}. Manual download instructions: {}".format(
                        train_dir, train_dir, self.manual_download_instructions
                    )
                )
```

This makes me think we should add a check like this in `hyperpartisan_news_detection.py` too

ghomasHudson: I agree - would be nice to give a more readable error here.

datasets/style_change_detection/style_change_detection.py
```diff
@@ -0,0 +1,136 @@
...
+            homepage="https://pan.webis.de/clef20/pan20-web/style-change-detection.html",
+            citation=_CITATION,
+        )
+
+    def _split_generators(self, dl_manager):
+        """Returns SplitGenerators."""
+        data_dir = os.path.abspath(os.path.expanduser(dl_manager.manual_dir))
+
+        train_dir = os.path.join(data_dir, "train", "dataset-" + self.config.name)
+        val_dir = os.path.join(data_dir, "train", "dataset-" + self.config.name)
```
lhoestq, line 98: It looks like both the train and validation sets are the same ?

ghomasHudson: good spot - fixed now

PR #397, opened 2020-07-15
Add contiguous sharding

jarednielsen:
This makes dset.shard() play nice with nlp.concatenate_datasets(). When I originally wrote the shard() method, I was thinking about a distributed training scenario, but https://github.com/huggingface/nlp/pull/389 also uses it for splitting the dataset for distributed preprocessing.

Usage:
```
nlp.concatenate_datasets([dset.shard(n, i, contiguous=True) for i in range(n)])
```


src/nlp/arrow_dataset.py
```diff
@@ -1329,13 +1330,23 @@ def shard(
...
+            a dataset with the same order as the original.
+
+            Be sure to shard before using any randomizing operator (such as shuffle).
             It is best if the shard operator is used early in the dataset pipeline.
 
+
             Args:
                 `num_shards` (`int`): How many shards to split the dataset into.
                 `index` (`int`): Which shard to select and return.
+                `contiguous`: (`bool`): Whether to select contiguous blocks of indices for shards.
```
lhoestq, line 1349:
```suggestion
                `contiguous`: (`bool`, default: `False`): Whether to select contiguous blocks of indices for shards.
```

jarednielsen: Done

PR #405, opened 2020-07-16
Make select() faster by batching reads

mitchellgordon95:
Here's a benchmark:

```
dataset = nlp.load_dataset('bookcorpus', split='train')

start = time.time()
dataset.select(np.arange(1000), reader_batch_size=1, load_from_cache_file=False)
end = time.time()
print(f'{end - start}')

start = time.time()
dataset.select(np.arange(1000), reader_batch_size=1000, load_from_cache_file=False)
end = time.time()
print(f'{end - start}')
```

Without batching, select takes around 1.27 seconds. With batching, it takes around 0.01 seconds. The slowness was upsetting me because dataset.shuffle() was supposed to take ~27 hours for bookcorpus. Now with the fix it takes ~2.5 hours (which still is pretty slow, but I'll open a separate issue for that).

### DIFF
--- a/src/nlp/arrow_dataset.py
+++ b/src/nlp/arrow_dataset.py
@@ -903,6 +903,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
         load_from_cache_file: bool = True,
         cache_file_name: Optional[str] = None,
         writer_batch_size: Optional[int] = 1000,
+        reader_batch_size: Optional[int] = 1000,
         verbose: bool = True,
     ):
         """ Create a new dataset with rows selected following the list/array of indices.
@@ -916,6 +917,8 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
                     results of the computation instead of the automatically generated cache file name.
                 `writer_batch_size` (`int`, default: `1000`): Number of rows per write operation for the cache file writer.
                     Higher value gives smaller cache files, lower value consume less temporary memory while running `.map()`.
+                `reader_batch_size` (`int`, default: `1000`): Number of rows per __getitem__ operation when reading from disk.
+                    Higher values may make reading faster but will also consume more temporary memory and make the progress bar less responsive.
                 `verbose` (`bool`, default: `True`): Set to `False` to deactivate the tqdm progress bar and informations.
         """
         if len(self.list_indexes()) > 0:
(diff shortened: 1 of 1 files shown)


src/nlp/arrow_dataset.py
```diff
@@ -903,6 +903,7 @@ def select(
         load_from_cache_file: bool = True,
         cache_file_name: Optional[str] = None,
         writer_batch_size: Optional[int] = 1000,
+        reader_batch_size: Optional[int] = 1000,
```
lhoestq, line 906: Could you add it to the docstrings ?

lhoestq: Also I would rename it `batch_size` (like in `map`, `filter`, etc.)
What do you think ?

src/nlp/arrow_dataset.py
```diff
@@ -953,10 +954,15 @@ def select(
                 logger.info("Caching processed dataset at %s", cache_file_name)
             writer = ArrowWriter(schema=self.schema, path=cache_file_name, writer_batch_size=writer_batch_size)
 
-        # Loop over single examples or batches and write to buffer/file if examples are to be updated
-        for i in tqdm(indices, disable=not verbose):
-            example = self._getitem(key=int(i), format_type=None, format_columns=None, format_kwargs=None)
-            writer.write(example)
+        # Loop over batches and write to buffer/file if examples are to be updated
+        for i in tqdm(range(0, len(indices), reader_batch_size), disable=not verbose):
+            example = self._getitem(
```
lhoestq, line 959: you can rename `example` to `batch`

src/nlp/arrow_dataset.py
mitchellgordon95: I thought about that, but I didn't want this option to conflict with the `batch_size` present in `map` and `filter`. In those functions, the `batch_size` is something that changes the user-visible behavior of `map` (it determines how many examples are sent to my callback per call.)

This `reader_batch_size` is purely behind the scenes and doesn't influence the output in a way that the end-user can see. So I thought it would be less confusing if they had different name. If `select` had a `batch_size`, I might expect it to return items in a list of chunks or something like that.

No strong opinions though.

lhoestq: Yea that makes sense. Let's keep `reader_batch_size`.

Issue #353, opened 2020-07-08
[Dataset requests] New datasets for Text Classification


thomwolf:
We are missing a few datasets for Text Classification which is an important field.

Namely, it would be really nice to add:
- Yelp-5
- Movie review (Movie Review (MR) dataset [156])
- SST (Stanford Sentiment Treebank)
- Multi-Perspective Question Answering (MPQA) dataset
- Amazon. This is a popular corpus of product reviews collected from the Amazon website [159]. It contains labels for both binary classification and multi-class (5-class) classification
- 20 Newsgroups. The 20 Newsgroups dataset
- Sogou News dataset
- Reuters news. The Reuters-21578 dataset [165]
- DBpedia. The DBpedia dataset [170]
- Ohsumed. The Ohsumed collection [171] is a subset of the MEDLINE database
- EUR-Lex. The EUR-Lex dataset
- WOS. The Web Of Science (WOS) dataset 
- PubMed. PubMed [173]
- TREC-QA. TREC-QA
- Quora. The Quora dataset [180]

All these datasets are cited in https://arxiv.org/abs/2004.03705

thomwolf:
Pinging @mariamabarham as well

jxmorris12:
- I added MR! It's called `rotten_tomatoes`
- SST is part of GLUE, no?
- I also added `ag_news`

mariamabarham:
Thanks @jxmorris12 for pointing this out. 

In glue we only have SST-2 maybe we can add separately SST-1.

jxmorris12:
This is the homepage for the Amazon dataset: https://www.kaggle.com/datafiniti/consumer-reviews-of-amazon-products

Is there an easy way to download kaggle datasets programmatically? If so, I can add this one!

mariamabarham:
Hi @jxmorris12 for now I think our `dl_manager` does not download from Kaggle.
@thomwolf , @lhoestq

ghomasHudson:
Pretty sure the quora dataset is the same one I implemented here: https://github.com/huggingface/nlp/pull/366

moscow25:
Great list. Any idea if Amazon Reviews has been added?

- ~40 GB of text (sadly no emoji)
- popular MLM pre-training dataset before bigger datasets like WebText https://arxiv.org/abs/1808.01371
[...]

PR #410, opened 2020-07-17
20newsgroup

mariamabarham:
Add 20Newsgroup dataset.
#353


datasets/newsgroup/newsgroup.py
```diff
@@ -0,0 +1,171 @@
...
+
+from __future__ import absolute_import, division, print_function
+
+import csv
+import os
+
+import nlp
+
+
+_CITATION = """
```
lhoestq, line 27: Could you use the standardized way for citations, like in the other scripts ?

datasets/newsgroup/newsgroup.py
```diff
@@ -0,0 +1,171 @@
...
+    ,title = "Newsweeder: Learning to filter netnews"
+    ,year = 1995
+    ,booktitle = "Proceedings of the Twelfth International Conference on Machine Learning"
+    ,pages = "331-339"
+    }
+ """
+
+_DESCRIPTION = """
+The 20 Newsgroups data set is a collection of approximately 20,000 newsgroup documents, partitioned (nearly) evenly across 
+20 different newsgroups. To the best of my knowledge, it was originally collected by Ken Lang, probably for his Newsweeder:
```
lhoestq, line 39: whose "knowledge" ? ^^
Is there a description that is more factual ?

datasets/newsgroup/newsgroup.py
```diff
@@ -0,0 +1,171 @@
...
+        self.data_dir = data_dir
+
+
+class Newsgroups(nlp.GeneratorBasedBuilder):
+    
+    BUILDER_CONFIGS = [
+        NewsgroupConfig(
+            name=name,
+            description = _DESC[name.split('_')[0]],
+            data_dir=name.split('_')[1],
```
lhoestq, line 110: Maybe we can rename this variable ? We use `data_dir` as the directory for manually downloaded data.
Something like `sub_directory` for example ?

datasets/newsgroup/newsgroup.py
mariamabarham: You're right.

PR #423, opened 2020-07-21
Change features vs schema logic

lhoestq:
## New logic for `nlp.Features` in datasets

Previously, it was confusing to have `features` and pyarrow's `schema` in `nlp.Dataset`.
However `features` is supposed to be the front-facing object to define the different fields of a dataset, while `schema` is only used to write arrow files.

Changes:
- Remove `schema` field in `nlp.Dataset`
- Make `features` the source of truth to read/write examples
- `features` can no longer be `None` in `nlp.Dataset`
- Update `features` after each dataset transform such as `nlp.Dataset.map`

Todo: change the tests to take these changes into account

### DIFF
--- a/src/nlp/arrow_dataset.py
+++ b/src/nlp/arrow_dataset.py
@@ -315,6 +315,25 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
             raise ValueError(f"Column ({column}) not in table columns ({self._data.column_names}).")
         return self._data.column(column).unique().to_pylist()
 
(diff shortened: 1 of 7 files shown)


lhoestq:
I had to make `SplitDict` serializable to be able to copy `DatasetInfo` objects properly.
Serialization was also asked in #389

thomwolf:
One thing I forgot to say here, is that we also want to use the features arguments of `load_dataset` (which goes in the builder’s config) to override the default features of a dataset script.

PR #429, opened 2020-07-23
mlsum

RachelKer:
Hello, 

The tests for the load_real_data fail, as there is no default language subset to download it looks for a file that does not exist. This bug does not happen when using the load_dataset function, as it asks you to specify a language if you do not, so I submit this PR anyway. The dataset is avalaible on : https://gitlab.lip6.fr/scialom/mlsum_data


datasets/mlsum/mlsum.py
```diff
@@ -0,0 +1,139 @@
...
+
+class Mlsum(nlp.GeneratorBasedBuilder):
+
+    # TODO(mlqa): Set up version.
+    VERSION = nlp.Version("1.0.0")
+    BUILDER_CONFIGS = (
+        [
+            MlsumConfig(
+                name=lang,
+                data_url=_URL,
```
mariamabarham, line 49: maybe you remove `data_url` given that it's the same for all config and it's equal to `_URL`

mariamabarham:
I think the dummy_data structure does not also match. In the `_split_generator` you have something like `os.path.join(downloaded_files["validation"], lang+'_val.jsonl')` but in you dummy_data you have `os.path.join(downloaded_files["validation"], lang+"_val.zip",  lang+'_val.jsonl')`. I think ` jsonl` files should be directly  in the `dummy_data` folder without the sub-folder 

@lhoestq

datasets/mlsum/mlsum.py
mariamabarham, line 49: I think `data_url` can be removed, it's not used and it the same as `_URL` for all configs

datasets/mlsum/mlsum.py
```diff
@@ -0,0 +1,139 @@
...
+We report cross-lingual comparative analyses based on state-of-the-art systems. 
+These highlight existing biases which motivate the use of a multi-lingual dataset. 
+"""
+_URL = "https://gitlab.lip6.fr/scialom/mlsum_data/-/raw/master/MLSUM/"
+_LANG = ["de", "es", "fr", "ru", "tu"]
+
+
+class MlsumConfig(nlp.BuilderConfig):
+    def __init__(self, data_url, **kwargs):
+        """BuilderConfig for MLQA
```
mariamabarham, line 31: `MLQA` to `MLSUM` ?

RachelKer: Indeed I used that script as a base, I corrected it

datasets/mlsum/mlsum.py
RachelKer: Indeed, I removed it

lhoestq:
Hi @RachelKer :)
Thanks for adding MLSUM !

To fix the CI I think you just have to rebase from master

RachelKer:
Great, I think it is working now. Thanks :)

lhoestq:
It looks like your PR does tons of changes in other datasets. 
Maybe this is because of the merge from master ?

RachelKer:
Hmm, I see, sorry I messed up somewhere. Maybe it's easier if we close the pull request and I do another one ?

lhoestq:
Yea if it's easier for you feel free to re-open a PR

PR #430, opened 2020-07-23
add DatasetDict

lhoestq:
### Overview

When you call `load_dataset` it can return a dictionary of datasets if there are several splits (train/test for example).
If you wanted to apply dataset transforms you had to iterate over each split and apply the transform.

Instead of returning a dict, it now returns a `nlp.DatasetDict` object which inherits from dict and contains the same data as before, except that now users can call dataset transforms directly from the output, and they'll be applied on each split.

Before:
```python
from nlp import load_dataset

squad = load_dataset("squad")
print(squad.keys())
# dict_keys(['train', 'validation'])
squad = {
    split_name: dataset.map(my_func) for split_name, dataset in squad.items()
}
print(squad.keys())
# dict_keys(['train', 'validation'])
```

Now:
```python
from nlp import load_dataset

squad = load_dataset("squad")
print(squad.keys())
# dict_keys(['train', 'validation'])
squad = squad.map(my_func)
print(squad.keys())
# dict_keys(['train', 'validation'])
```

### Dataset transforms

`nlp.DatasetDict` implements the following dataset transforms:
- map
- filter
- sort
- shuffle

### Arguments

The arguments of the methods are the same except for split-specific arguments like `cache_file_name`.
For such arguments, the expected input is a dictionary `{split_name: argument_value}`
It concerns:
- `cache_file_name` in map, filter, sort, shuffle
- `seed` and `generator` in shuffle


src/nlp/arrow_dataset.py
```diff
@@ -1280,7 +1303,7 @@ def train_test_split(
             )
         # If the array is empty we do nothing
         if len(self) == 0:
-            return self
+            return DatasetDict({"train": self, "test": self})
```
thomwolf, line 1306: Good catch!

src/nlp/arrow_dataset.py
```diff
@@ -1241,7 +1264,7 @@ def train_test_split(
         test_cache_file_name: Optional[str] = None,
         writer_batch_size: Optional[int] = 1000,
         verbose: bool = True,
-    ):
+    ) -> DatasetDict:
```
thomwolf, line 1267: :)

src/nlp/load.py
```diff
@@ -506,7 +507,7 @@ def load_dataset(
         **config_kwargs (Optional ``dict``): keyword arguments to be passed to the ``nlp.BuilderConfig`` and used in the ``nlp.DatasetBuilder``.
 
     Returns:
-        ``nlp.Dataset`` or ``Dict[nlp.Split, nlp.Dataset]``
+        ``nlp.Dataset`` or ``nlp.DatasetDict``
             if `split` is not None: the dataset requested,
             if `split` is None, a `dict<key: nlp.Split, value: nlp.Dataset>` with each split.
```
thomwolf, line 512: Update this line as well `a `dict<key: nlp.Split, value: nlp.Dataset>` with each split` to mention nlp.DatasetDict maybe

src/nlp/dataset_dict.py
```diff
@@ -0,0 +1,236 @@
+from typing import Dict, List, Optional
+
+import numpy as np
+
+from .features import Features
+
+
+class DatasetDict(dict):
```
thomwolf, line 8: We could maybe override `__setitem__` to make sure all values are `Dataset` or maybe simpler and more flexible is to have an assert and a clean clean error message in all the processing methods (`map`, `filter` etc) for this case (otherwise it will fail with a not-super-explicit "No map method for ...").

lhoestq:
I did the changes in the docstrings and I added a type check in each `DatasetDict` method to make sure all values are of type `Dataset`

thomwolf:
Awesome, do you mind adding these in the doc as well?

lhoestq:
I added it to the docs (processing + main classes)

HugToDebug:
I'm trying to follow along with the following about datasets from the docs:

https://huggingface.co/nlp/loading_datasets.html
https://huggingface.co/nlp/processing.html

However the train_test_split method no longer works as it is expecting a dataset, rather than a datsetdict. How would I got about splitting a CSV into a train and test set? 

I'm trying to utilize the Trainer() class, but am having trouble converting my data from a csv into dataset objects to pass in.

PR #431, opened 2020-07-24
Specify split post processing + Add post processing resources downloading

lhoestq:
Previously if you tried to do

```python
from nlp import load_dataset
wiki = load_dataset("wiki_dpr", "psgs_w100_with_nq_embeddings", split="train[:100]", with_index=True)
```
Then you'd get an error `Index size should match Dataset size...`
This was because it was trying to use the full index (21M elements).

To fix that I made it so post processing resources can be named according to the split.

I'm going to add tests on post processing too.

Note that the CI will fail as I added a new argument in `_post_processing_resources`: the AWS version of wiki_dpr fails, and there's also an error telling that it is not synced (it'll be synced once it's merged):
```
FAILED tests/test_dataset_common.py::AWSDatasetTest::test_load_dataset_wiki_dpr
FAILED tests/test_hf_gcp.py::TestDatasetSynced::test_script_synced_with_s3_wiki_dpr
```

EDIT: I did a change to ignore the script hash to locate the arrow files on GCS, so I removed the sync test. It was there just because of the hash logic for files on GCS


lhoestq:
I was using a hack in `wiki_dpr` to download the index from GCS even for the configurations without the embeddings.
However as GCS is something internal, I changed the logic to add a download step for indexes directly in the dataset script, using the `DownloadManager`.

This change was directly linked to the changes I did to take into account the split name in the post processing, so I included this change in this PR too.

To summarize:

Dataset builders can now implement
- `_post_processing_resources(split)`: return a dict `resource_name -> resource_file_name`. It defines the additional resources such as indexes or arrow files that you need in post processing
- `_download_post_processing_resources(split, resource_name, dl_manager))`: if some resources can be downloaded, you can use the download_manager to download them
- `_post_process(dataset, resources_path)`: (main function for post processing) given a dataset, you can apply dataset transforms or add indexes. For resources that have been downloaded, you can load them. For the others, you can generate and save them. The paths to load/save resources are in `resources_path` which is a dictionary `resource_name -> resource_path`

About the CI:
```
FAILED tests/test_dataset_common.py::AWSDatasetTest::test_load_dataset_wiki_dpr
```
It fails because I changed the input of post processing functions (to include the split name)

datasets/wiki_dpr/wiki_dpr.py
```diff
@@ -29,42 +29,71 @@
...
     """BuilderConfig for WikiDprConfig."""
 
-    def __init__(self, with_embeddings=True, with_index=False, index_train_size=262144, dummy=False, **kwargs):
+    def __init__(
+        self,
+        with_embeddings=True,
+        with_index=False,
+        index_train_size=262144,
+        dummy=False,
+        index_type="compressed",
```
thomwolf, line 44: So maybe we call this `index_name` and:
- store the index with a simple name like `compressed` and `exact`
- add metadata to the index somewhere in an additional `index_name.json` file or in the `dataset.json` (sorry I know it's one additional file but I feel like we might need a better way to store index metadata than the file name in the future)

thomwolf: what do you think?

lhoestq: I was thinking of adding the post processing resources (like the indexes) metadata (resource name, resource file name, size) in the DatasetInfo.

However `wiki_dpr` may have an index or not depending on the configuration (`with_index` parameter) so it's not ideal to do that.

More precisely:

A `dataset_info.json` is inside cache directory. A cache directory corresponds to a "downloaded and prepared" dataset  for a given configuration name + version + script hash. It means that for one cache dir we could have one dataset with the exact index, a one with the compressed index, but only one `dataset_info.json` and that's probably not what we want.

One solution to fix that would be to take the `config_kwargs` into account when building the `_relative_cache_dir`.
Another solution would to have one configuration name for exact index, one for compressed index, and one for no index.

lhoestq: I'm changing wiki_dpr to have one configuration name for all the combinations of
- dummy (true/false)
- wiki split (only one possibility for now: psgs_w100)
- embeddings name (no embeddings or "nq")
- index name (no index or "compressed" or "exact")

Then we'll have one set of post processing resources per config and we'll be fine to add indexes metadata in DatasetInfo.

lhoestq:
I started to add metadata in the DatasetInfo.
Note that because there are new field, **all the dataset_info[s].json generated after these changes won't be loadable from older versions of the lib**

Right now it looks like this:
```json
        "post_processing_resources_checksums": {
            "train": {
                "embeddings_index": {
                    "num_bytes": 30720045,
                    "checksum": "b04fb4f4f3ab83b9d1b9f6f9eb236f1c04a9fd61bef7cee16b12df8ac911766a"
                }
            }
        },
        "post_processing_size": 30720045,
```

thomwolf:
Good point. Should we anticipate already that we may add other fields in the future and change the code to support the addition of new fields without breaking backward compatibility in the future?

lhoestq:
I added:
- post processing features (inside a PostProcessedInfo object)
- backward compatibility for dataset info
- post processing tests (as_dataset and download_and_prepare) for map (change features), select (change number of elements) and add_faiss_index (add indexes)
And I fixed a bug in `map` that I found thanks to the new tests

Now I just have to move `post_processing_resources_checksums` to  PostProcessedInfo as well and everything should be good :)

PR #434, opened 2020-07-25
Fixed check for pyarrow

n2ds:
Fix check for pyarrow in __init__.py.  Previously would raise an error for pyarrow >= 1.0.0


thomwolf:
Great, thanks!

PR #441, opened 2020-07-27
Add features parameter in load dataset

lhoestq:
Added `features` argument in `nlp.load_dataset`.
If they don't match the data type, it raises a `ValueError`.

It's a draft PR because #440 needs to be merged first.


lhoestq:
This one is ready for review now

src/nlp/load.py
```diff
@@ -526,6 +529,7 @@ def load_dataset(
         data_dir=data_dir,
         data_files=data_files,
         hash=hash,
+        info=DatasetInfo(features=features),
```
thomwolf, line 532: Why do we use the full `DatasetInfo` instead of just the features here?

src/nlp/builder.py
```diff
@@ -165,6 +164,10 @@ def __init__(
                 )
                 os.rmdir(self._cache_dir)
 
+        # update info with user specified infos
+        if info is not None:
+            self.info.update(info)
```
thomwolf, line 169: Doesn't this break because some values in `DatasetInfo` like `description`, `citation`, etc are empty strings by default and not `None`?

Do we need to give access to other attributes than `features`? Otherwise, I would favor a more robust approach of using directly features which would avoid potential side effect if we change the DatasetInfo default values later.

lhoestq: Yes all the values are are not None would overwrite the existing ones, which is not good.
I'm fine with using only features then.

lhoestq:
I change to using features only, instead of info.
Let mw know if it sounds good to you now @thomwolf

PR #450, opened 2020-07-28
add sogou_news

mariamabarham:
This PR adds the sogou news dataset
#353


datasets/sogou_news/sogou_news.py
```diff
@@ -0,0 +1,99 @@
...
+    """Sogou News dataset"""
+
+    def _info(self):
+        return nlp.DatasetInfo(
+            description=_DESCRIPTION,
+            features=nlp.Features(
+                {
+                    "title": nlp.Value("string"),
+                    "content": nlp.Value("string"),
+                    "class": nlp.Value("string")
```
lhoestq, line 62: Maybe we can use `ClassLabel` ?
Feel free to hardcode the classes instead of reading `classes.txt` to define the labels.

PR #451, opened 2020-07-28
Fix csv/json/txt cache dir

### LINKED ISSUE #444: Keep loading old file even I specify a new file in load_dataset
I used load a file called 'a.csv' by 
```
dataset = load_dataset('csv', data_file='./a.csv')
```
And after a while, I tried to load another csv called 'b.csv'
```
dataset = load_dataset('csv', data_file='./b.csv')
```
However, the new dataset seems to remain the old 'a.csv' and not loading new csv file.

Even worse, after I load a.csv, the load_dataset function keeps loading the 'a.csv' afterward. 

Is this a cache problem?

lhoestq:
The cache dir for csv/json/txt datasets was always the same. This is an issue because it should be different depending on the data files provided by the user.

To fix that, I added a line that use the hash of the data files provided by the user to define the cache dir.

This should fix #444

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,6 +67,8 @@ REQUIRED_PKGS = [
     "dataclasses;python_version<'3.7'",
     # filesystem locks e.g. to prevent parallel downloads
     "filelock",
+    # for fast hashing
+    "xxhash"
 ]
 
 TESTS_REQUIRE = [
--- a/src/nlp/builder.py
+++ b/src/nlp/builder.py
@@ -235,18 +235,22 @@ class DatasetBuilder:
         if builder_config.data_files is not None:
             m = xxhash.xxh64()
             if isinstance(builder_config.data_files, str):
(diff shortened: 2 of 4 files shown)


thomwolf:
I think this is the way to go but I’m afraid this might be a little slow. I was thinking that we could use a high quality very fast non crypto hash like xxhash for these stuff (hashing data files)

lhoestq:
Yep good idea, I'll take a look

lhoestq:
I tested the hashing speed [here](https://colab.research.google.com/drive/1hlhP84kLIHmOzMRQN1h8x10hKWpXXyud?usp=sharing).
I was able to get 8x speed with `xxhashlib` (42ms vs 345ms for 100MiB of data).
What do you think @thomwolf ?

lhoestq:
I added xxhashlib and some tests

PR #459, opened 2020-07-31
[Breaking] Update Dataset and DatasetDict API

thomwolf:
This PR contains a few breaking changes so it's probably good to keep it for the next (major) release:
- rename the `flatten`, `drop` and `dictionary_encode_column` methods in `flatten_`, `drop_` and `dictionary_encode_column_` to indicate that these methods have in-place effects as discussed in #166. From now on we should keep the convention of having a trailing underscore for methods which have an in-place effet. I also adopt the conversion of not returning the (self) dataset for these methods. This is different than what PyTorch does for instance (`model.to()` is in-place but return the self model) but I feel like it's a safer approach in terms of UX.
- remove the `dataset.columns` property which returns a low-level Apache Arrow object and should not be used by users. Similarly, remove `dataset. nbytes` which we don't really want to expose in this bare-bone format.
- add a few more properties and methods to `DatasetDict`


src/nlp/arrow_dataset.py
```diff
@@ -400,27 +370,31 @@ def cast_(self, features: Features):
...
 
-        del self._info.features[column_name]
+        for column_name in column_names:
+            del self._info.features[column_name]
 
-        self._data = self._data.remove_column(column_index)
+        self._data = self._data.drop(column_names)
 
     def rename_column_(self, original_column_name: str, new_column_name: str):
         """
```
lhoestq, line 400: Can you add a check that `new_column_name` doesn't exist already ?

lhoestq (approved):
Cool !

PR #460, opened 2020-07-31
Fix KeyboardInterrupt in map and bad indices in select

lhoestq:
If you interrupted a map function while it was writing, the cached file was not discarded.
Therefore the next time you called map, it was loading an incomplete arrow file.

We had the same issue with select if there was a bad indice at one point.

To fix that I used temporary files that are renamed once everything is finished.

### DIFF
--- a/src/nlp/arrow_dataset.py
+++ b/src/nlp/arrow_dataset.py
@@ -20,6 +20,8 @@ import contextlib
 import hashlib
 import logging
 import os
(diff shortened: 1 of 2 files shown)


lhoestq:
Thanks @TevenLeScao for finding this issue

TevenLeScao:
Thanks @lhoestq for catching this ❤️

PR #465, opened 2020-07-31
Keep features after transform

lhoestq:
When applying a transform like `map`, some features were lost (and inferred features were used).
It was the case for ClassLabel, Translation, etc.

To fix that, I did some modifications in the `ArrowWriter`:

- added the `update_features` parameter. When it's `True`, then the features specified by the user (if any) can be updated with inferred features if their type don't match. `map` transform sets `update_features=True` when writing to cache file or buffer. Features won't change by default in `map`.

- added the `with_metadata` parameter. If `True`, the `features` (after update) will be written inside the metadata of the schema in this format:
```
{
    "huggingface": {"features" : <serialized Features exactly like dataset_info.json>}
} 
```
Then, once a dataset is instantiated without info/features, these metadata are used to set the features of the dataset.

### DIFF
--- a/src/nlp/features.py
+++ b/src/nlp/features.py
@@ -378,7 +378,9 @@ def get_nested_type(schema: FeatureType) -> pa.DataType:
     """ Convert our Feature nested object in an Apache Arrow type """
     # Nested structures: we allow dict, list/tuples, sequences
     if isinstance(schema, dict):
-        return pa.struct({key: get_nested_type(schema[key]) for key in sorted(schema)})  # sort to make the type deterministic
+        return pa.struct(
+            {key: get_nested_type(schema[key]) for key in sorted(schema)}
+        )  # sort to make the type deterministic
     elif isinstance(schema, (list, tuple)):
         assert len(schema) == 1, "We defining list feature, you should just provide one example of the inner type"
         inner_type = get_nested_type(schema[0])
--- a/src/nlp/features.py
+++ b/src/nlp/features.py
@@ -460,7 +460,8 @@ def generate_from_arrow_type(pa_type: pa.DataType):
     elif isinstance(pa_type, pa.FixedSizeListType):
         return Sequence(feature=generate_from_arrow_type(pa_type.value_type), length=pa_type.list_size)
     elif isinstance(pa_type, pa.ListType):
-        return [generate_from_arrow_type(pa_type.value_type)]
+        # return [generate_from_arrow_type(pa_type.value_type)]
+        return Sequence(feature=generate_from_arrow_type(pa_type.value_type))
     elif isinstance(pa_type, pa.DictionaryType):
         raise NotImplementedError  # TODO(thom) this will need access to the dictionary as well (for labels). I.e. to the py_table
     elif isinstance(pa_type, pa.DataType):
--- a/src/nlp/features.py
+++ b/src/nlp/features.py
@@ -277,7 +277,7 @@ class Translation:
     _type: str = field(default="Translation", init=False, repr=False)
 
     def __call__(self):
-        return pa.struct({lang: pa.string() for lang in self.languages})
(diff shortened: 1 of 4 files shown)


src/nlp/features.py
```diff
@@ -277,7 +277,7 @@ class Translation:
     _type: str = field(default="Translation", init=False, repr=False)
 
     def __call__(self):
-        return pa.struct({lang: pa.string() for lang in self.languages})
+        return pa.struct({lang: pa.string() for lang in sorted(self.languages)})
```
lhoestq, line 280: I had to add `sorted` in some places to make features comparisons work (features.type needs to be deterministic)

tests/test_arrow_dataset.py
```diff
@@ -309,6 +323,82 @@ def test_filter(self):
...
+        dset = Dataset.from_dict({"tokens": [["foo"] * 5] * 10, "labels": [[1] * 5] * 10}, features=features)
+
+        def invert_labels(x):
+            return {"labels": [(1 - label) for label in x["labels"]]}
+
+        inverted_dset = dset.map(invert_labels, keep_in_memory=True)
+        self.assertEqual(inverted_dset.features.type, features.type)
+        self.assertDictEqual(inverted_dset.features, features)
+
+    def test_keep_features_after_loading_from_cache(self):
```
thomwolf, line 385: Can you add a last test about when you add a new column?
E.g. create a duplicate `labels2` of the `labels` column and check that it's assigned an inferred features (should be `Sequence(Value('int32'))` I guess?) while the original `labels` column still has a `Sequence(ClassLabel(names=["negative", "positive"]))` feature

lhoestq: int64 actually ^^

tests/test_arrow_dataset.py
```diff
@@ -309,6 +323,105 @@ def test_filter(self):
...
+                "labels": Sequence(ClassLabel(names=["negative", "positive"])),
+                "labels2": Sequence(Value("int64")),
+            }
+        )
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            tmp_file = os.path.join(tmp_dir, "test.arrow")
+            inverted_dset = dset.map(invert_labels, cache_file_name=tmp_file)
+            self.assertEqual(inverted_dset.features.type, expected_features.type)
+            self.assertDictEqual(inverted_dset.features, expected_features)
```
lhoestq, line 423: I just added this test with a new column to check that inference is fine @thomwolf

src/nlp/features.py
```diff
@@ -458,7 +460,10 @@ def generate_from_arrow_type(pa_type: pa.DataType):
     elif isinstance(pa_type, pa.FixedSizeListType):
         return Sequence(feature=generate_from_arrow_type(pa_type.value_type), length=pa_type.list_size)
     elif isinstance(pa_type, pa.ListType):
-        return [generate_from_arrow_type(pa_type.value_type)]
+        feature = generate_from_arrow_type(pa_type.value_type)
+        if isinstance(feature, (dict, tuple, list)):
+            return [feature]
+        return Sequence(feature=feature)
```
lhoestq, line 466: I had to fix that.
There were issues with arrow types like `list of list of struct`.
We have one dataset that has this format: `docred`

thomwolf: Great catch!

lhoestq:
One note on features inference:

if an arrow type is `struct of items` where each item is a `list`, then we return a `dict` in which each item is a `Sequence`.
It means that we don't use the Sequence <-> dict swap when we infer features.

It's fine because the swap is generally used in dataset scripts, in which features are defined (inferred features are discarded)

lhoestq:
If it's fine for you @thomwolf we can merge this one :)

thomwolf:
Yes this is fine I think!

PR #466, opened 2020-08-01
[METRICS] Various improvements on metrics

thomwolf:
- Disallow the use of positional arguments to avoid `predictions` vs `references` mistakes
- Allow to directly feed numpy/pytorch/tensorflow/pandas objects in metrics


lhoestq:
The cast function is now called inside `features.encode_example`.
I also added `encode_batch` that was missing.

Moreover I used the cast function in `Dataset.map` to support torch/tensorflow tensors or numpy arrays inputs.

There are tests for tensors inputs in metrics and in .map

src/nlp/metric.py
```diff
@@ -228,18 +231,20 @@ def compute(self, predictions=None, references=None, timeout=120, **metrics_kwar
         predictions = self.data["predictions"]
         references = self.data["references"]
         with temp_seed(self.seed):
-            output = self._compute(predictions=predictions, references=references, **metrics_kwargs)
+            output = self._compute(predictions=predictions, references=references, **kwargs)
```
thomwolf, line 234: Here we should be careful that `_compute` doesn't allow `**kwargs` otherwise we could have a silent error where for instance `predictions` is written as `prediction` and is not recognized and this goes undetected and the predictions are missed.

This could maybe be done with a test on signature of the metrics `_compute` method in the tests for the metrics scripts (or maybe there is another smart way to do that).

src/nlp/metric.py
```diff
@@ -228,18 +231,20 @@ def compute(self, predictions=None, references=None, timeout=120, **metrics_kwar
...
         return output
 
-    def add_batch(self, predictions=None, references=None, **kwargs):
-        """ Add a batch of predictions and references for the metric's stack.
+    def add_batch(self, *, predictions=None, references=None, **kwargs):
```
thomwolf, line 237: Here also, maybe we should disallow `**kwargs` to avoid silent errors. It's a bit of a design decision because then it means we can't override `add_batch` in a derived class but I think we are probably fine with it.

src/nlp/metric.py
```diff
@@ -228,18 +231,20 @@ def compute(self, predictions=None, references=None, timeout=120, **metrics_kwar
...
+        Add a batch of predictions and references for the metric's stack.
         """
         batch = {"predictions": predictions, "references": references}
+        batch = self.info.features.encode_batch(batch)
         if self.writer is None:
             self._init_writer()
         self.writer.write_batch(batch)
 
-    def add(self, prediction=None, reference=None, **kwargs):
+    def add(self, *, prediction=None, reference=None, **kwargs):
```
thomwolf, line 247: Same here about `**kwargs`

src/nlp/utils/file_utils.py
```diff
@@ -89,6 +90,38 @@
 INCOMPLETE_SUFFIX = ".incomplete"
 
 
+@contextmanager
+def temp_seed(seed: int, set_pytorch=False, set_tensorflow=False):
```
thomwolf, line 94: Here there is an error with the tensorflow seed setter that I didn't have time to debug before my time off, maybe you can take a look? You can see the error simply by running the colab at https://colab.research.google.com/drive/1I_B1mcX0cOzOskr0rJN8u8xh_0CIken-?usp=sharing and using `set_tensorflow=True` in its call of `temp_seed`.

lhoestq: I fixed tensorflow's rng for eager mode.
Not sure how to do it when eager mode is off. I tried a few things but it didn't work

src/nlp/metric.py
lhoestq: done

src/nlp/metric.py
lhoestq: done

src/nlp/metric.py
lhoestq: I added a test for the signature of `_compute` to make sure it doesn't have `**kwargs`

thomwolf:
I think we can merge

Issue #438, opened 2020-07-26
New Datasets: IWSLT15+, ITTB


sshleifer:
**Links:**
[iwslt](https://pytorchnlp.readthedocs.io/en/latest/_modules/torchnlp/datasets/iwslt.html)
Don't know if that link is up to date.

[ittb](http://www.cfilt.iitb.ac.in/iitb_parallel/)
**Motivation**: replicate mbart finetuning
![image](https://user-images.githubusercontent.com/6045025/88490093-0c1c8c00-cf67-11ea-960d-8dcaad2aa8eb.png)

For future readers, we already have the following language pairs in the wmt namespaces:

```
wmt14: ['cs-en', 'de-en', 'fr-en', 'hi-en', 'ru-en']
wmt15: ['cs-en', 'de-en', 'fi-en', 'fr-en', 'ru-en']
wmt16: ['cs-en', 'de-en', 'fi-en', 'ro-en', 'ru-en', 'tr-en']
wmt17: ['cs-en', 'de-en', 'fi-en', 'lv-en', 'ru-en', 'tr-en', 'zh-en']
wmt18: ['cs-en', 'de-en', 'et-en', 'fi-en', 'kk-en', 'ru-en', 'tr-en', 'zh-en']
wmt19: ['cs-en', 'de-en', 'fi-en', 'gu-en', 'kk-en', 'lt-en', 'ru-en', 'zh-en', 'fr-de']
```

thomwolf:
Thanks Sam, we now have a very detailed tutorial and template on how to add a new dataset to the library. It typically take 1-2 hours to add one. Do you want to give it a try ?
The tutorial on writing a new dataset loading script is here: https://huggingface.co/nlp/add_dataset.html
And the part on how to share a new dataset is here: https://huggingface.co/nlp/share_dataset.html

mariamabarham:
Hi @sshleifer, I'm trying to add IWSLT using the link you provided but the download urls are not working. Only `[en, de]` pair is working. For others language pairs it throws a `404` error.

jbbqqf:
Hi! Triaging older issues this week — I think this one can be closed as out-of-scope for the current `datasets` repo.

[...]

PR #470, opened 2020-08-03
Adding IWSLT 2017  dataset.

Narsil:
Created a [IWSLT 2017](https://sites.google.com/site/iwsltevaluation2017/TED-tasks) dataset script for the *multilingual data*.

```
Bilingual data:  {Arabic, German, French, Japanese, Korean, Chinese} <-> English
Multilingual data:   German, English, Italian, Dutch, Romanian. (Any pair)
```

I'm unsure how to handle bilingual vs multilingual. Given `nlp` architecture a Config option seems to be the way to go, however, it might be a bit confusing to have different language pairs with different option. Using just language pairs is not viable as English to German exists in both.

Any opinion on how that should be done ?
EDIT: I decided to just omit de-en from multilingual as it's only a subset of the bilingual one. That way only language pairs exist.
EDIT : Could be interesting for #438


datasets/iwslt2017.py
```diff
@@ -0,0 +1,227 @@
...
+
+    def _split_generators(self, dl_manager):
+        """Returns SplitGenerators."""
+        source, target = self.config.pair.split("-")
+        if self.config.is_multilingual:
+            dl_dir = dl_manager.download_and_extract(MULTI_URL)
+            data_dir = os.path.join(dl_dir, "DeEnItNlRo-DeEnItNlRo")
+            years = [2010]
+        else:
+            bi_url = f"https://wit3.fbk.eu/archive/2017-01-trnted//texts/{source}/{target}/{source}-{target}.tgz"
```
lhoestq, line 125:
```suggestion
            bi_url = f"https://wit3.fbk.eu/archive/2017-01-trnted/texts/{source}/{target}/{source}-{target}.tgz"
```

datasets/iwslt2017.py
```diff
@@ -0,0 +1,227 @@
...
+    if source != target
+]
+BI_PAIRS = [
+    f"{source}-{target}"
+    for source in BI_LANGUAGES
+    for target in BI_LANGUAGES
+    if source != target and (source == "en" or target == "en")
+]
+
+PAIRS = MULTI_PAIRS + BI_PAIRS
```
lhoestq, line 78: I think it's fine this way

datasets/iwslt2017.py
```diff
@@ -0,0 +1,227 @@
...
+    ]
+
+    def _info(self):
+        return nlp.DatasetInfo(
+            # This is the description that will appear on the datasets page.
+            description=_DESCRIPTION,
+            # nlp.features.FeatureConnectors
+            features=nlp.Features(
+                {"source": nlp.Value("string"), "target": nlp.Value("string"),}
+            ),
```
lhoestq, line 107: I would use the `Translation` feature type.

You can see an example of usage in the [FLoRes dataset script](https://github.com/huggingface/nlp/blob/master/datasets/flores/flores.py#L93)

datasets/iwslt2017.py
```diff
@@ -0,0 +1,227 @@
...
+
+                                source_row = source_row.strip()
+                                target_row = target_row.strip()
+                            else:
+                                continue
+
+                        yield id_, {
+                            "source": source_row,
+                            "target": target_row,
+                        }
```
lhoestq, line 226: If you use the `Translation` feature type defined this way

```python
features=nlp.Features({"translation": nlp.features.Translation(languages=self.config.pair.split("-"))})
```

Then the `yield` statement here would become:

```python
yield id_, {
    "translation": {
        source: source_row,
        target: target_row
    }
}
```

where source and target are defined by
```python
source, target = self.config.pair.split("-")
```

Narsil: Ok done.

Narsil:
Ok I tried to add the dummy dataset (I actually modified the dummy_data command to generate them for me because it was too painful to do that manually).

The dummy_data test seems to work:
```bash
RUN_SLOW=1 pytest tests/test_dataset_common.py::LocalDatasetTest::test_load_dataset_all_configs_iwslt2017
```

However the test on the full data fails, because the `**config_kwargs` don't include `pair, multilingual`.
I could add a default parameter for the Config (but that feels broken, how can one config be the "default" ?). If I do I still have errors, saying that something within the downloader is a directory so I'm not sure where that comes from.

I can share my auto_zip dummy data code if you want (I tried to keep it clean). The way it works is that it just keeps X line from the beginning of the original files, and Y lines at the end. It's good enough for my usage, but I guess it could work for most data files out there (as long as they're real text and not binary format)

lhoestq:
The slow test doesn't support dataset that require config parameters that don't have default values.

To improve that we can replace it by two tests:
- one test that loads the default config (it can simply be the first config of the config lists for example)
- one tests that iterate over all configs and load them all one by one

By using the configs inside the builder config lists, there is no need to instantiate new configs, so the missing parameter error doesn't happen.

Does that sound good to you ?

Narsil:
Seems fair.
However I'm unsure what I should do ?

Should I wait for #527 to pass and rebase and the command will be the same ?
Should I update something ?

lhoestq:
I think everything is fine on your side. Thanks for adding this dataset :)

I think it's better to wait for the slow test to be updated if you don't mind.

Narsil:
Sure ! :)

datasets/iwslt2017/iwslt2017.py
```diff
@@ -0,0 +1,190 @@
...
+        """
+
+        Args:
+            pair: the language pair to consider
+            is_multilingual: Is this pair in the multilingual dataset (download source is different)
+            **kwargs: keyword arguments forwarded to super.
+        """
+        self.pair = pair
+        self.is_multilingual = is_multilingual
+        super(IWSLT2017Config, self).__init__(**kwargs)
```
julien-c, line 60:
```suggestion
        super().__init__(**kwargs)
```

We're Python 3 only so you can omit this

datasets/iwslt2017/iwslt2017.py
```diff
@@ -166,8 +166,8 @@ def _generate_examples(self, source_files, target_files, split):
         id_ = 0
         source, target = self.config.pair.split("-")
         for source_file, target_file in zip(source_files, target_files):
-            with open(source_file) as sf:
-                with open(target_file) as tf:
+            with open(source_file, "r", encoding="utf-8") as sf:
+                with open(target_file, "w", encoding="utf-8") as tf:
```
julien-c, line 170:
```suggestion
                with open(target_file, "r", encoding="utf-8") as tf:
```

Narsil: Bitten by my own internal pattern matching (target files get written to)

src/nlp/arrow_writer.py
```diff
@@ -362,6 +362,7 @@ def finalize(self, metrics_query_result: dict):
                 that the filter keeps only the metrics for the considered split, under the namespace `split_name`.
         """
         import apache_beam as beam
+
```
lhoestq, line 365: It looks like an isort issue

src/nlp/builder.py
```diff
@@ -958,6 +958,7 @@ def _build_pcollection(pipeline, extracted_dir):
     def _download_and_prepare(self, dl_manager, verify_infos):
         # Create the Beam pipeline and forward it to _prepare_split
         import apache_beam as beam
+
```
lhoestq, line 961: same here

src/nlp/search.py
```diff
@@ -14,8 +14,8 @@
 
 try:
     import elasticsearch as es
-    from elasticsearch import Elasticsearch
     import elasticsearch.helpers
+    from elasticsearch import Elasticsearch
```
lhoestq, line 18: same here

lhoestq (approved):
Looks good :)
Could you remove the changes caused by `isort` ? (see my comments)
We haven't updated it yet

src/nlp/arrow_writer.py
Narsil: Okay, I removed them (automated tools are eager to do stuff sometimes :))

lhoestq:
Thanks for fixing the isort/black changes :)
Feel free to merge if it's good for you @Narsil

PR #473, opened 2020-08-03
add DoQA dataset (ACL 2020)

mariamabarham:
add DoQA dataset (ACL 2020) http://ixa.eus/node/12931


datasets/doqa/doqa.py
```diff
@@ -0,0 +1,184 @@
...
+
+        super(DoqaConfig, self).__init__(version=nlp.Version("2.1.0", "New split API (https://tensorflow.org/datasets/splits)"), **kwargs)
+
+
+class Doqa(nlp.GeneratorBasedBuilder):
+    
+    BUILDER_CONFIGS = [
+        DoqaConfig(
+            name="cooking",
+            description=""
```
lhoestq, line 72: You can simply remove the description in that case ?

datasets/doqa/doqa.py
```diff
@@ -0,0 +1,184 @@
...
+                    gen_kwargs={"filepath": os.path.join(path, "doqa-v2.1", "doqa_dataset",  "doqa-movies-test-v2.1.json")},
+                )
+            ]
+        elif self.config.name == "travel":
+            return [
+                nlp.SplitGenerator(
+                    name=nlp.Split.TEST,
+                    gen_kwargs={"filepath": os.path.join(path, "doqa-v2.1", "doqa_dataset", "doqa-travel-test-v2.1.json")},
+                )
+            ]
```
lhoestq, line 138: maybe raise an error if the config name is wrong ?

datasets/doqa/doqa.py
```diff
@@ -0,0 +1,184 @@
...
+                        question = qa["question"]
+                        answers = qa["answers"]
+                        id1 = qa["id"]
+                        yesno = qa["yesno"]
+                        followup = qa["followup"]
+                        answer_text = [answer["text"] for answer in answers]
+                        answer_start = [answer["answer_start"] for answer in answers]
+                        
+                        orig_answer_start = [qa["orig_answer"]["answer_start"]]
+                        orig_answer_text = [qa["orig_answer"]["text"]]
```
lhoestq, line 163: Those two fields are supposed to be `Sequence` according to the features description, but it looks like they always contain exactly one element. Is this intended ?

lhoestq:
We just did the change nlp -> datasets @jplu 

The tests won't pass as the dataset won't be able to import nlp

PR #475, opened 2020-08-04
misc. bugs and quality of life

joeddav:
A few misc. bugs and QOL improvements that I've come across in using the library. Let me know if you don't like any of them and I can adjust/remove them.

1. Printing datasets without a description field throws an error when formatting the `single_line_description`. This fixes that, and also adds some formatting to the repr to make it slightly more readable.
```
>>> print(list_datasets()[0])
nlp.ObjectInfo(
	id='aeslc',
	description='A collection of email messages of employees in the Enron Corporation.There are two features:  - email_body: email body text.  - subject_line: email subject text.',
	files=[nlp.S3Object('aeslc.py'), nlp.S3Object('dataset_infos.json'), nlp.S3Object('dummy/1.0.0/dummy_data-zip-extracted/dummy_data/AESLC-master/enron_subject_line/dev/allen-p_inbox_29.subject'), nlp.S3Object('dummy/1.0.0/dummy_data-zip-extracted/dummy_data/AESLC-master/enron_subject_line/test/allen-p_inbox_24.subject'), nlp.S3Object('dummy/1.0.0/dummy_data-zip-extracted/dummy_data/AESLC-master/enron_subject_line/train/allen-p_inbox_20.subject'), nlp.S3Object('dummy/1.0.0/dummy_data.zip'), nlp.S3Object('urls_checksums/checksums.txt')]
)
```

2. Add id-only option to `list_datasets` and `list_metrics` to allow the user to easily print out just the names of the datasets & metrics. I often found myself annoyed that this took so many strokes to do.

```python
[dataset.id for dataset in list_datasets()] # before
list_datasets(id_only=True) # after
```

3. Fix null-seed randomization caching. When using `train_test_split` and `shuffle`, the computation was being cached even without a seed or generator being passed. The result was that calling `.shuffle` more than once on the same dataset didn't do anything without passing a distinct seed or generator. Likewise with `train_test_split`.

4. Indexing by iterables of bool. I added support for passing an iterable of type bool to `_getitem` as a numpy/pandas-like indexing method. Let me know if you think it's redundant with `filter` (I know it's not optimal memory-wise), but I think it's nice to have as a lightweight alternative to do simple things without having to create a copy of the entire dataset, e.g.

```python
dataset[dataset['label'] == 0] # numpy-like bool indexing to look at instances with labels of 0
```

5. Add an `input_column` argument to `map` and `filter`, which allows you to filter/map on a particular column rather than passing the whole dict to the function. Also adds `fn_kwargs` to be passed to the function. I think these together make mapping much cleaner in many cases such as mono-column tokenization:

```python
# before
dataset = dataset.map(lambda batch: tokenizer(batch["text"])
# after
dataset = dataset.map(tokenizer, input_column="text")
dataset = dataset.map(tokenizer, input_column="text", fn_kwargs={"truncation": True, "padding": True})
```

### DIFF
--- a/src/nlp/arrow_dataset.py
+++ b/src/nlp/arrow_dataset.py
@@ -1120,7 +1120,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
             cache_file_name=cache_file_name,
             writer_batch_size=writer_batch_size,
             verbose=verbose,
-            fn_kwargs=fn_kwargs
+            fn_kwargs=fn_kwargs,
         )
 
     def select(
--- a/src/nlp/arrow_dataset.py
+++ b/src/nlp/arrow_dataset.py
@@ -702,7 +702,16 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
             else:
                 outputs = self._data[key].to_pylist()
         elif isinstance(key, Iterable):
-            data_subset = pa.concat_tables(self._data.slice(int(i), 1) for i in key)
+            if len(key) > 0 and isinstance(key[0], (bool, np.bool_)):
+                if len(key) != self.__len__():
+                    raise ValueError(
+                        f"Iterable with bool entries must be length of dataset ({self.__len__()}), "
+                        f"not {len(key)}"
+                    )
+                indices = [i for i, val in enumerate(key) if val]
+            else:
+                indices = key
+            data_subset = pa.concat_tables(self._data.slice(int(i), 1) for i in indices)
             if format_type is not None:
                 if format_type == "pandas":
                     outputs = data_subset.to_pandas(split_blocks=True)
--- a/src/nlp/arrow_dataset.py
+++ b/src/nlp/arrow_dataset.py
@@ -705,8 +705,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
             if len(key) > 0 and isinstance(key[0], (bool, np.bool_)):
                 if len(key) != self.__len__():
                     raise ValueError(
-                        f"Iterable with bool entries must be length of dataset ({self.__len__()}), "
-                        f"not {len(key)}"
+                        f"Iterable with bool entries must be length of dataset ({self.__len__()}), " f"not {len(key)}"
                     )
                 indices = [i for i, val in enumerate(key) if val]
             else:
@@ -802,7 +801,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
         features: Optional[Features] = None,
         disable_nullable: bool = True,
         verbose: bool = True,
-        **fn_kwargs
+        **fn_kwargs,
     ) -> "Dataset":
         """ Apply a function to all the elements in the table (individually or in batches)
             and update the table (if function does updated examples).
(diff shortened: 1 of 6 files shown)


src/nlp/arrow_dataset.py
```diff
@@ -1284,7 +1322,7 @@ def shuffle(
         ), "The provided generator must be an instance of numpy.random.Generator"
 
         # Check if we've already cached this computation (indexed by a hash)
-        if self._data_files:
+        if self._data_files and (seed is not None or generator is not None):
```
lhoestq, line 1325: I think it would be better to simply remove all this `if` clause completely.
Indeed we can ignore this caching part as it is already present inside `select` that is called right after.

It also fixes the issue you mentioned.
Indeed when seed is None and generator is None, then it's going to be `permutation` that is going to influence the cache file name (that's how `select` works). Therefore if you call shuffle twice then it's going to return different datasets, as `permutation` will be different

src/nlp/arrow_dataset.py
```diff
@@ -984,22 +1013,26 @@ def filter(
         self,
         function,
         with_indices=False,
+        input_column: Optional[str] = None,
```
lhoestq, line 1016: You need to add `input_column` to the `cache_kwargs` of `filter`.
Same thing for `fn_kwargs`

src/nlp/arrow_dataset.py
```diff
@@ -766,6 +774,7 @@ def map(
         self,
         function,
         with_indices: bool = False,
+        input_column: Optional[str] = None,
```
lhoestq, line 777: You need to add `input_column` to the `cache_kwargs` of `map`.
Same thing for `fn_kwargs`

src/nlp/arrow_dataset.py
lhoestq: Actually I can see a case where this if clause has its advantages. It can avoid call `generator. permutation` and speed up things for huge datasets.

Therefore what I would suggest instead is simply move 

```python
        if generator is None:
            generator = np.random.default_rng(seed)
```

right before this if clause. It should to the job

src/nlp/arrow_dataset.py
```diff
@@ -1522,7 +1560,7 @@ def train_test_split(
             )
 
         # Check if we've already cached this computation (indexed by a hash)
-        if self._data_files:
+        if self._data_files and (seed is not None or generator is not None):
```
lhoestq, line 1563: same here

instead of testing for seed/generator being None or not, simply move 
```python
            if generator is None:
                generator = np.random.default_rng(seed)
```
right before this line

joeddav: Great. I'm assuming it's ok to move this out of the if statement that it was in before? Ok to create the generator and just not use it when not shuffling?

src/nlp/arrow_dataset.py
joeddav: Good point. In the `filter` case, I had to add to add `input_column` to the `fn_kwargs` rather than pass it to `map` as `input_column`, otherwise `map_function` wouldn't be able access the full batches. Not quite as elegant but it works.

src/nlp/arrow_dataset.py
joeddav: Done. Also removed `seed` from the cache in this case, since generator will always be defined and seed may not be.

src/nlp/arrow_dataset.py
lhoestq: Yes it's fine

src/nlp/arrow_dataset.py
thomwolf: E.g. dataset.map(tokenizer, input_columns=['sentence1', 'sentence2'])

src/nlp/arrow_dataset.py
```diff
@@ -1299,13 +1349,15 @@ def shuffle(
             generator, np.random.Generator
         ), "The provided generator must be an instance of numpy.random.Generator"
 
+        if generator is None:
+            generator = np.random.default_rng(seed)
```
thomwolf, line 1353: Good catch!

src/nlp/arrow_dataset.py
joeddav: Good call. Added.

joeddav:
Cool, I made those changes. LMK if you think it's ready for merge.

thomwolf:
Ok to merge for me

Issue #347, opened 2020-07-07
'cp950' codec error from load_dataset('xtreme', 'tydiqa')


jerryIsHere:
![image](https://user-images.githubusercontent.com/50871412/86744744-67481680-c06c-11ea-8612-b77eba92a392.png)

I guess the error is related to python source encoding issue that my PC is trying to decode the source code with wrong encoding-decoding tools, perhaps :
https://www.python.org/dev/peps/pep-0263/

I guess the error was triggered by the code " module = importlib.import_module(module_path)" at line 57 in the source code:  nlp/src/nlp/load.py / (https://github.com/huggingface/nlp/blob/911d5596f9b500e39af8642fe3d1b891758999c7/src/nlp/load.py#L51)

Any ideas?

p.s. tried the same code on colab, that runs perfectly

patpizio:
This is probably a Windows issue, we need to specify the encoding when `load_dataset()` reads the original CSV file.
Try to find the `open()` statement called by `load_dataset()` and add an `encoding='utf-8'` parameter.
See issues #242 and #307

lhoestq:
It should be in `xtreme.py:L755`:
```python
        if self.config.name == "tydiqa" or self.config.name.startswith("MLQA") or self.config.name == "SQuAD":
            with open(filepath) as f:
                data = json.load(f)
```

Could you try to add the encoding parameter:
```python
open(filepath, encoding='utf-8')
```

lhoestq:
Hello @jerryIsHere :) Did it work ?
If so we may change the dataset script to force the utf-8 encoding

jerryIsHere:
@lhoestq sorry for being that late, I found 4 copy of xtreme.py. I did the changes as what has been told to all of them.
The problem is not solved

lhoestq:
Could you provide a better error message so that we can make sure it comes from the opening of the `tydiqa`'s json files ?

jerryIsHere:
@lhoestq 
The error message is same as before:
Exception has occurred: UnicodeDecodeError
'cp950' codec can't decode byte 0xe2 in position 111: illegal multibyte sequence
  File "D:\python\test\test.py", line 3, in <module>
    dataset = load_dataset('xtreme', 'tydiqa')

![image](https://user-images.githubusercontent.com/50871412/87748794-7c216880-c829-11ea-94f0-7caeacb4d865.png)

I said that I found 4 copy of xtreme.py and add the 「, encoding='utf-8'」 parameter to the open() function
[...]

PR #481, opened 2020-08-06
Apply utf-8 encoding to all datasets

### LINKED ISSUE #468: UnicodeDecodeError while loading PAN-X task of XTREME dataset
Hi 🤗  team!

## Description of the problem
I'm running into a `UnicodeDecodeError` while trying to load the PAN-X subset the XTREME dataset: 

```
UnicodeDecodeError                        Traceback (most recent call last)
<ipython-input-5-1d61f439b843> in <module>
----> 1 dataset = load_dataset("xtreme", "PAN-X.en", data_dir='./data')

/usr/local/lib/python3.6/dist-packages/nlp/load.py in load_dataset(path, name, version, data_dir, data_files, split, cache_dir, features, download_config, download_mode, ignore_verifications, save_infos, **config_kwargs)
    528     ignore_verifications = ignore_verifications or save_infos
    529     # Download/copy dataset processing script
--> 530     module_path, hash = prepare_module(path, download_config=download_config, dataset=True)
    531 
    532     # Get dataset builder class from the processing script

/usr/local/lib/python3.6/dist-packages/nlp/load.py in prepare_module(path, download_config, dataset, force_local_path, **download_kwargs)
    265 
    266     # Download external imports if needed
--> 267     imports = get_imports(local_path)
    268     local_imports = []
    269     library_imports = []

/usr/local/lib/python3.6/dist-packages/nlp/load.py in get_imports(file_path)
    156     lines = []
    157     with open(file_path, mode="r") as f:
--> 158         lines.extend(f.readlines())
    159 
    160     logger.info("Checking %s for additional imports.", file_path)

/usr/lib/python3.6/encodings/ascii.py in decode(self, input, final)
     24 class IncrementalDecoder(codecs.IncrementalDecoder):
     25     def decode(self, input, final=False):
---> 26         return codecs.ascii_decode(input, self.errors)[0]
     27 
     28 class StreamWriter(Codec,codecs.StreamWriter):

UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 111: ordinal not in range(128)
```

## Steps to reproduce
Install from nlp's master branch
```python
pip install git+https://github.com/huggingface/nlp.git
```
then run
```python
from nlp import load_dataset
# AmazonPhotos.zip is located in data/
dataset = load_dataset("xtreme", "PAN-X.en", data_dir='./data')
```

## OS / platform details

- `nlp` version: latest from master
- Platform: Linux-4.15.0-72-generic-x86_64-with-Ubuntu-18.04-bionic
- Python version: 3.6.9
- PyTorch version (GPU?): 1.4.0 (True)
- Tensorflow version (GPU?): 2.1.0 (True)
- Using GPU in script?: True
- Using distributed or parallel set-up in script?: False

## Proposed solution
Either change [line 762](https://github.com/huggingface/nlp/blob/7ada00b1d62f94eee22a7df38c6b01e3f27194b7/datasets/xtreme/xtreme.py#L762) in `xtreme.py` to include UTF-8 encoding:

```
# old
with open(filepath) as f
# new
with open(filepath, encoding='utf-8') as f
```

or raise a warning that suggests setting the locale explicitly, e.g.
```python
import locale
locale.setlocale(locale.LC_ALL, 'C.UTF-8')
```
I have a preference for the first solution. Let me know if you agree and I'll be happy to implement the simple fix!

lewtun:
## Description
This PR applies utf-8 encoding for all instances of `with open(...) as f` to all Python files in `datasets/`. As suggested by @thomwolf in #468 , we use regular expressions and the following function

```python
def apply_encoding_on_file_open(filepath: str):
    """Apply UTF-8 encoding for all instances where a non-binary file is opened."""
    
    with open(filepath, 'r', encoding='utf-8') as input_file:
        regexp = re.compile(r"(?!.*\b(?:encoding|rb|w|wb|w+|wb+|ab|ab+)\b)(?<=\s)(open)\((.*)\)")
        input_text = input_file.read()
        match = regexp.search(input_text)
        
        if match:
            output = regexp.sub(lambda m: m.group()[:-1]+', encoding="utf-8")', input_text)
            with open(filepath, 'w', encoding='utf-8') as output_file:
                output_file.write(output)
```

to perform the replacement. 

Note:

1. I excluded all _**binary files**_ from the search since it's possible some objects are opened for which the encoding doesn't make sense. Please correct me if I'm wrong and I'll tweak the regexp accordingly
2. There were two edge cases where the regexp failed (e.g. two `open` instances on a single line), but I decided to just fix these manually in the interest of time.
3. I only applied the replacement to files in `datasets/`. Let me know if this should be extended to other places like `metrics/`
4. I have implemented a unit test that should catch missing encodings in future CI runs

Closes #468  and possibly #347


lewtun:
Not sure why the AWS test is failing - perhaps I made too many concurrent CI builds 😢. Can someone please rerun the CI to check the error is not on my end?

lewtun:
I pushed an improved docstring and the unit tests now pass. For some reason the docs are failing to build, but does not seem related to my changes ...

lhoestq:
The build_doc fail has been fixed on master.
It was due to the latest update of sphinx that has some issues, so I pinned the previous version for now.

lhoestq:
I noticed that you also changed the Apache Beam `open` to also use utf-8. However it doesn't have an `encoding` parameter.
Therefore you should ignore lines like

```python
beam.io.filesystems.FileSystems.open(filepath, encoding="utf-8")
```

I guess you could add a rule to your regex to only include the `open` call that have a space right before it.

lewtun:
Good catch @lhoestq! Your suggestion to match on `open(...)` with a whitespace was a great idea - it allowed me to simplify the regexp considerably 😄.

I fixed the Apache Beam false positives and also caught a few problems in `json.load()`, e.g.
```python
relation_name_map = json.load(open(rel_info), encoding='utf-8')
```

I've tested that the new regexp doesn't reintroduce these false positives, so I think the PR is ready for another review.

tests/test_file_encoding.py
```diff
@@ -0,0 +1,33 @@
...
+
+    with open(filepath, "r", encoding="utf-8") as input_file:
+        regexp = re.compile(r"(?!.*\b(?:encoding|rb|w|wb|w+|wb+|ab|ab+)\b)(?<=\s)(open)\((.*)\)")
+        input_text = input_file.read()
+        match = regexp.search(input_text)
+
+    return match
+
+
+def test_no_encoding_on_file_open():
```
lhoestq, line 27: Can you make this function a `unittest.TestCase` ?

tests/test_file_encoding.py
```diff
@@ -0,0 +1,33 @@
...
+    (?!.*\b(?:encoding|rb|w|wb|w+|wb+|ab|ab+)\b): Lookahead and discard match if `encoding` or `rb` etc are
+    arguments of `open()`.
+
+    (?<=\s): Lookbehind and match if `open()` predeceded by one whitespace.
+
+    (open)\((.*)\): Capture everything in parentheses of `open()`.
+    """
+
+    with open(filepath, "r", encoding="utf-8") as input_file:
+        regexp = re.compile(r"(?!.*\b(?:encoding|rb|w|wb|w+|wb+|ab|ab+)\b)(?<=\s)(open)\((.*)\)")
```
lhoestq, line 20: Can you update the regex with your latest version ?

tests/test_file_encoding.py
lewtun: Done!

tests/test_file_encoding.py
lewtun: Sorry if I wasn't clear in my previous messages - the simplified regex (amended in 
commit 03f98c66e8d9b3cea79ca10f4b220dab9f7e9e99) _is_ the latest version. 

Your suggestion allowed me to replace ugly patterns like `(?<!csv.reader\()` and `(?<!\.)` with just a whitespace lookbehind `(?<=\s)`.

Please let me know if I misunderstood your comment!

lhoestq: Oh ok I see it's good then

thomwolf:
Ok to merge this @lhoestq ?

lhoestq (approved):
Looks good :)
I think it's ok to merge now !

PR #487, opened 2020-08-10
Fix elasticsearch result ids returning as strings

sai-prasanna:
I am using the latest elasticsearch binary and master of nlp. For me elasticsearch searches failed because the resultant "id_" returned for searches are strings, but our library assumes them to be integers.


lhoestq:
It looks like you need to rebase from master to fix the CI. Could you do that please ?

PR #494, opened 2020-08-11
Fix numpy stacking

lhoestq:
When getting items using a column name as a key, numpy arrays were not stacked.
I fixed that and added some tests.

There is another issue that still needs to be fixed though: when getting items using a column name as a key, pytorch tensors are not stacked (it outputs a list of tensors). This PR should help with the to fix this issue.


lhoestq:
This PR also fixed a bug where numpy arrays were returned instead of pytorch tensors when getting with a clumn as a key.

Issue #501, opened 2020-08-12
Caching doesn't work for map (non-deterministic)


wulu473:
The caching functionality doesn't work reliably when tokenizing a dataset. Here's a small example to reproduce it. 

```python
import nlp
import transformers

def main():
    ds = nlp.load_dataset("reddit", split="train[:500]")

    tokenizer = transformers.AutoTokenizer.from_pretrained("gpt2")

    def convert_to_features(example_batch):
        input_str = example_batch["body"]
        encodings = tokenizer(input_str, add_special_tokens=True, truncation=True)
        return encodings

    ds = ds.map(convert_to_features, batched=True)

if __name__ == "__main__":
    main()
```

Roughly 3/10 times, this example recomputes the tokenization.

Is this expected behaviour?

lhoestq:
Thanks for reporting !

To store the cache file, we compute a hash of the function given in `.map`, using our own hashing function.
The hash doesn't seem to stay the same over sessions for the tokenizer.
Apparently this is because of the regex at `tokenizer.pat` is not well supported by our hashing function.

I'm working on a fix

wulu473:
Thanks everyone. Works great now.

MaveriQ:
Hi. I believe the fix was for the nlp library. Is there a solution to handle compiled regex expressions in .map() with the caching. I want to run a simple regex pattern on a big dataset, but I am running into the issue of compiled expression not being cached. 

Instead of opening a new issue, I thought I would put my query here. Let me know if a new issue would be more suitable. Thanks

mariosasko:
Hi @MaveriQ! This fix is also included in the `datasets` library. Can you provide a reproducer?

PR #502, opened 2020-08-13
Fix tokenizers caching

lhoestq:
I've found some cases where the caching didn't work properly for tokenizers:

1. if a tokenizer has a regex pattern, then the caching would be inconsistent across sessions
2. if a tokenizer has a cache attribute that changes after some calls, the the caching would not work after cache updates
3. if a tokenizer is used inside a function, the caching of this function would result in the same cache file for different tokenizers
4. if `unique_no_split_tokens`'s attribute is not the same across sessions (after loading a tokenizer) then the caching could be inconsistent

To fix that, this is what I did:

1. register a specific `save_regex` function for pickle that makes regex dumps deterministic
2. ignore cache attribute of some tokenizers before dumping
3. enable recursive dump by default for all dumps
4. make `unique_no_split_tokens` deterministic in https://github.com/huggingface/transformers/pull/6461

I also added tests to make sure that tokenizers hashing works as expected.
In the future we should find a way to test if hashing also works across session (maybe using two CI jobs ? or by hardcoding a tokenizer's hash ?)

### DIFF
--- a/src/nlp/utils/py_utils.py
+++ b/src/nlp/utils/py_utils.py
@@ -29,7 +29,6 @@ from types import CodeType
 
 import dill
 import numpy as np
-import regex
 
 
 # NOTE: When used on an instance method, the cache is shared across all
@@ -383,13 +382,18 @@ def copyfunc(func):
     return types.FunctionType(func.__code__, func.__globals__, func.__name__, func.__defaults__, func.__closure__)
 
 
-@pklregister(type(regex.Regex("", 0)))
-def _save_regex(pickler, obj):
-    dill._dill.log.info("Re: %s" % obj)
-    args = (
-        obj.pattern,
-        obj.flags,
-    )
-    pickler.save_reduce(regex.compile, args, obj=obj)
-    dill._dill.log.info("# Re")
-    return
+try:
+    import regex
+
+    @pklregister(type(regex.Regex("", 0)))
+    def _save_regex(pickler, obj):
+        dill._dill.log.info("Re: %s" % obj)
+        args = (
+            obj.pattern,
+            obj.flags,
+        )
+        pickler.save_reduce(regex.compile, args, obj=obj)
+        dill._dill.log.info("# Re")
+        return
+except ImportError:
+    pass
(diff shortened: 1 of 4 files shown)


lhoestq:
This should fix #501 and also the issue you sent me on slack @sgugger .

src/nlp/utils/py_utils.py
```diff
@@ -268,14 +270,29 @@ class Pickler(dill.Pickler):
...
+    Pickler(file, recurse=True).dump(obj)
     return
 
 
+@contextlib.contextmanager
+def _no_cache_fields(obj):
+    if _transformers_available:
+        import transformers as tr
+
+        if isinstance(obj, (tr.CTRLTokenizer, tr.GPT2Tokenizer, tr.OpenAIGPTTokenizer, tr.XLMTokenizer)):
```
thomwolf, line 282: This list might be a bit cumbersome to maintain in the future.
Should we just do a check that the `cache` attribute exists and is a dict maybe?

src/nlp/utils/py_utils.py
```diff
@@ -297,9 +314,11 @@ def save_code(pickler, obj):
     dill._dill.log.info("Co: %s" % obj)
     # Filenames of functions created in notebooks or shells start with '<'
     # ex: <ipython-input-13-9ed2afe61d25> for ipython, and <stdin> for shell
+    # Moreover lambda functions have a special name: '<lambda>'
+    # ex: (lambda x: x).__code__.co_name == "<lambda>"  # True
     # Only those two lines are different from the original implementation:
-    co_filename = "" if obj.co_filename.startswith("<") else obj.co_filename
-    co_firstlineno = 1 if obj.co_filename.startswith("<") else obj.co_firstlineno
+    co_filename = "" if obj.co_filename.startswith("<") or obj.co_name == "<lambda>" else obj.co_filename
+    co_firstlineno = 1 if obj.co_filename.startswith("<") or obj.co_name == "<lambda>" else obj.co_firstlineno
```
thomwolf, line 321: Good!

src/nlp/utils/py_utils.py
lhoestq: I changed it so that it checks if `cache` exists and is a dict

thomwolf (approved):
Good to merge for me when you are happy with it :-)

Issue #474, opened 2020-08-03
test_load_real_dataset when config has BUILDER_CONFIGS that matter


marcotcr:
It a dataset has custom `BUILDER_CONFIGS` with non-keyword arguments (or keyword arguments with non default values), the config is not loaded during the test and causes an error.
I think the problem is that `test_load_real_dataset` calls `load_dataset` with `data_dir=temp_data_dir` ([here](https://github.com/huggingface/nlp/blob/master/tests/test_dataset_common.py#L200)). This causes [this line](https://github.com/huggingface/nlp/blob/master/src/nlp/builder.py#L201) to always be false because `config_kwargs` is not `None`. [This line](https://github.com/huggingface/nlp/blob/master/src/nlp/builder.py#L222) will be run instead, which doesn't use `BUILDER_CONFIGS`.

For an example, you can try running the test for lince:
` RUN_SLOW=1 pytest tests/test_dataset_common.py::LocalDatasetTest::test_load_real_dataset_lince`
which yields
> E           TypeError: __init__() missing 3 required positional arguments: 'colnames', 'classes', and 'label_column'

lhoestq:
The `data_dir` parameter has been removed. Now the error is `ValueError: Config name is missing`

As mentioned in #470 I think we can have one test with the first config of BUILDER_CONFIGS, and another test that runs all of the configs in BUILDER_CONFIGS

lhoestq:
This was fixed in #527 

Closing this one, but feel free to re-open if you have other questions

PR #504, opened 2020-08-13
Added downloading to Hyperpartisan news detection

ghomasHudson:
Following the discussion on Slack and #349, I've updated the hyperpartisan dataset to pull directly from Zenodo rather than manual install, which should make this dataset much more accessible. Many thanks to @johanneskiesel !

Currently doesn't pass `test_load_real_dataset` - I'm using `self.config.name` which is `default` in this test. Might be related to #474


johanneskiesel:
Thank you @ghomasHudson for making our dataset available! This is great!

lhoestq:
The tests passes since #527 :)

datasets/hyperpartisan_news_detection/hyperpartisan_news_detection.py
```diff
@@ -111,42 +100,44 @@ def _info(self):
...
-                    gen_kwargs={
-                        "articles_file": os.path.join(
-                            data_dir, "articles-validation-" + self.config.name + "-20181122.xml"
-                        ),
-                        "labels_file": os.path.join(
-                            data_dir, "ground-truth-validation-" + self.config.name + "-20181122.xml"
-                        ),
-                    },
+                    name=split,
+                    gen_kwargs = data_dir[split]
```
lhoestq, line 132: Could you run `make style` before we merge ?

lhoestq (approved):
Thanks for updating the link :)

PR #515, opened 2020-08-19
Fix batched map for formatted dataset

lhoestq:
If you had a dataset formatted as numpy for example, and tried to do a batched map, then it would crash because one of the elements from the inputs was missing for unchanged columns (ex: batch of length 999 instead of 1000).
The happened during the creation of the `pa.Table`, since columns had different lengths.


thomwolf (approved):
Good catch!

PR #516, opened 2020-08-19
[Breaking] Rename formated to formatted

lhoestq:
`formated` is not correct but `formatted` is


src/nlp/arrow_dataset.py
```diff
@@ -482,7 +482,7 @@ def format(self):
         }
 
     @contextlib.contextmanager
-    def formated_as(
+    def formatted_as(
```
lhoestq, line 485: this is breaking

thomwolf (approved):
Lol a breaking typo. Ok for that since we are renaming the in-place methods for `1.0.0` anyway :-)

PR #518, opened 2020-08-19
[METRICS, breaking] Refactor caching behavior, pickle/cloudpickle metrics and dataset, add tests on metrics

thomwolf:
Move the acquisition of the filelock at a later stage during metrics processing so it can be pickled/cloudpickled after instantiation.

Also add some tests on pickling, concurrent but separate metric instances and concurrent and distributed metric instances.

Changes significantly the caching behavior for the metrics:
- if the metric is used in a non-distributed setup (most common case) we try to find a free cache file using UUID instead of asking for an `experiment_id` if we can't lock the cache file this allows to use several instances of the same metrics in parallel.
- if the metrics is used in a distributed setup we ask for an `experiment_id` if we can't lock the cache file (because all the nodes need to have related cache file names for the final sync.
- after the computation, we free the locks and delete all the cache files.

Breaking: Some arguments for Metrics initialization have been removed for simplicity (`version`...) and some have been renamed for consistency with the rest of the library (`in_memory` => `keep_in_memory`).

Also remove the `_has_transformers` detection in utils to avoid importing transformers everytime during loading.

### DIFF
--- a/src/nlp/metric.py
+++ b/src/nlp/metric.py
@@ -19,7 +19,7 @@ import logging
 import os
 import types
 import uuid
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
 import numpy as np
 import pyarrow as pa
@@ -30,7 +30,7 @@ from .arrow_reader import ArrowReader
 from .arrow_writer import ArrowWriter
 from .info import MetricInfo
 from .naming import camelcase_to_snakecase
-from .utils import HF_METRICS_CACHE, Version, copyfunc, temp_seed
+from .utils import HF_METRICS_CACHE, copyfunc, temp_seed
 from .utils.download_manager import DownloadManager
 from .utils.file_utils import DownloadConfig
 
@@ -48,6 +48,7 @@ class Metric(object):
         seed: Optional[int] = None,
         config_name: Optional[str] = None,
         experiment_id: Optional[str] = None,
+        max_concurrent_cache_files: int = 10000,
         **kwargs,
     ):
         """ A Metrics is the base class and common API for all metrics.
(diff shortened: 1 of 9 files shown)


thomwolf:
(test failure is unrelated)

lhoestq (approved):
Awesome !
Bonus test you could add: check that it's not possible to use two concurrent metric instances

src/nlp/metric.py
```diff
@@ -104,108 +107,117 @@ def __init__(
...
+                self._data_dir_root/self.name/self.config_name/self.hash (if not none)/
             If any of these element is missing or if ``with_version=False`` the corresponding subfolders are dropped.
         """
-        builder_data_dir = os.path.join(self.name, self.config_name)
-        if with_version:
-            builder_data_dir = os.path.join(builder_data_dir, str(self._version))
-        if self.hash:
-            builder_data_dir = os.path.join(builder_data_dir, self.hash)
+        builder_data_dir = self._data_dir_root
+        builder_data_dir = os.path.join(builder_data_dir, self.name, self.config_name)
```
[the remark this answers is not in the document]
lhoestq: Ok ! Maybe you can update the docstring that mentions `self.hash` then

src/nlp/metric.py
```diff
@@ -104,108 +107,123 @@ def __init__(
...
-        return os.path.join(self.data_dir, f"{self.experiment_id}-{self.name}-{node_id}.arrow")
+    def _get_all_cache_files(self, timeout=100) -> Tuple[List[str], List[FileLock]]:
+        """ Get a lock on all the cache files in a distributed setup.
+            We wait for timeout second to let all the distributed node finish their tasks (default is 100 seconds).
+        """
+        if self.num_process == 1:
+            file_paths = [self.cache_file_name]
+        else:
+            file_paths = [
+                os.path.join(self.data_dir, f"{self.experiment_id}-{self.num_process}-{process_id}.arrow")
```
lhoestq, line 176: What about the files with `file_uuid` like this ?

```python
file_path = os.path.join(
    self.data_dir, f"{self.experiment_id}-{file_uuid}-{self.num_process}-{self.process_id}.arrow"
)
```

thomwolf: They are only used for single-process metrics (the previous branches of the if-else).
I'll add a comment.

sgugger:
As discussed with @thomwolf merging since the hyperparameter-search has been merged in transformers.

PR #520, opened 2020-08-20
Transform references for sacrebleu

jbragg:
Currently it is impossible to use sacrebleu when len(predictions) != the number of references per prediction (very uncommon), due to a strange format expected by sacrebleu. If one passes in the data to `nlp.metric.compute()` in sacrebleu format, `nlp` throws an error due to mismatching lengths between predictions and references. If one uses a more standard format where predictions and references are lists of the same length, sacrebleu throws an error.

This PR transforms reference data in a more standard format into the [unusual format](https://github.com/mjpost/sacreBLEU#using-sacrebleu-from-python) expected by sacrebleu.

### DIFF
--- a/metrics/sacrebleu/sacrebleu.py
+++ b/metrics/sacrebleu/sacrebleu.py
@@ -84,9 +84,17 @@ class Sacrebleu(nlp.Metric):
                 lowercase=False,
                 tokenize=scb.DEFAULT_TOKENIZER,
                 use_effective_order=False):
+        # Transform references into sacrebleu format
+        # NOTE: Only uses first k references for each prediction, where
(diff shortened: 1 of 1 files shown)


lhoestq (commented):
Thanks for reporting this !
Indeed sacrebleu expected this unusual format as input...

I think it would be better to check that all the references have the same length rather than cropping to the minimum length. If one length doesn't match the others, we should probably raise an error. What do you think ?

jbragg:
I think I agree @lhoestq so I pushed a change.
Thanks for your work on the library!

thomwolf (approved):
Very cool, thanks @jbragg

lhoestq (approved):
Nice thank you !

PR #521, opened 2020-08-20
Fix dictionnary (dictionary) typo

yonigottesman:
This error happens many times I'm thinking maybe its spelled like this on purpose?


thomwolf:
Hahah thanks Yonatan. It was not on purpose, we are just not very good at spelling :)

PR #523, opened 2020-08-20
Speed up Tokenization by optimizing cast_to_python_objects

lhoestq:
I changed how `cast_to_python_objects` works to make it faster.
It is used to cast numpy/pytorch/tensorflow/pandas objects to python lists, and it works recursively.

To avoid iterating over possibly long lists, it first checks if the first element that is not None has to be casted.
If the first element needs to be casted, then all the elements of the list will be casted, otherwise they'll stay the same.
This trick allows to cast objects that contain tokenizers outputs without iterating over every single token for example.

Speed improvement:

```python
import transformers
import nlp

tok = transformers.BertTokenizerFast.from_pretrained("bert-base-uncased")
txt = ["a " * 512] * 1000
dataset = nlp.Dataset.from_dict({"txt": txt})

# Tokenization using .map is now faster. Previously it was taking 3.5s
%time _ = dataset.map(lambda x: tok(x["txt"]), batched=True, load_from_cache_file=False)
# 450ms

# for comparison
%time _ = tok(txt)
# 280ms

```

### DIFF
--- a/src/nlp/features.py
+++ b/src/nlp/features.py
@@ -38,7 +38,7 @@ if _tf_available:
     import tensorflow as tf
 
 
-def string_to_arrow(type_str: str):
+def string_to_arrow(type_str: str) -> pa.DataType:
     if type_str not in pa.__dict__:
         if str(type_str + "_") not in pa.__dict__:
             raise ValueError(
@@ -53,7 +53,7 @@ def string_to_arrow(type_str: str):
     return pa.__dict__[arrow_data_type_str]()
 
 
-def _cast_to_python_objects(obj):
+def _cast_to_python_objects(obj: Any) -> Tuple[Any, bool]:
     """
     Cast numpy/pytorch/tensorflow/pandas objects to python lists.
     It works recursively.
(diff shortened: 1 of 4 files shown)


src/nlp/features.py
```diff
@@ -54,21 +54,72 @@ def string_to_arrow(type_str: str):
 
 
 def _cast_to_python_objects(obj):
```
thomwolf, line 56: ` -> Tuple[Any, bool]` ?

src/nlp/features.py
```diff
@@ -54,21 +54,72 @@ def string_to_arrow(type_str: str):
...
+            casted_first_elmt, has_changed_first_elmt = _cast_to_python_objects(first_elmt)
+            if has_changed_first_elmt:
+                return [_cast_to_python_objects(elmt)[0] for elmt in obj], True
+            else:
+                if isinstance(obj, list):
+                    return obj, False
+                else:
+                    return list(obj), True
+        else:
+            return [], True
```
thomwolf, line 102: Not sure if this one should be `True` or `False` actually

thomwolf: Maybe a `return [], bool(isinstance(obj, tuple))`?

thomwolf (approved):
Nice! Maybe we can add a few tests on these behaviors?

src/nlp/features.py
lhoestq: Yes `return [], bool(isinstance(obj, tuple))` makes more sense

lhoestq:
I took your comments into account and added tests for `cast_to_python_objects`

PR #526, opened 2020-08-24
Returning None instead of "python" if dataset is unformatted

TevenLeScao:
Following the discussion on Slack, this small fix ensures that calling `dataset.set_format(type=dataset.format["type"])` works properly. Slightly breaking as calling `dataset.format` when the dataset is unformatted will return `None` instead of `python`.

### DIFF
--- a/src/nlp/arrow_dataset.py
+++ b/src/nlp/arrow_dataset.py
@@ -477,7 +477,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
     @property
     def format(self):
         return {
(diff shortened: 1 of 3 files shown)


lhoestq (approved):
Good catch :)
I'm also in favor of using `type=None` for unformatted instead of `python`

lhoestq:
We have to change the tests to expect `None` instead of `python` then

thomwolf (approved):
Yes, that's better!

TevenLeScao:
Merging!

PR #528, opened 2020-08-24
fix missing variable names in docs

### LINKED ISSUE #524: Some docs are missing parameter names
See https://huggingface.co/nlp/master/package_reference/main_classes.html#nlp.Dataset.map. I believe this is because the parameter names are enclosed in backticks in the docstrings, maybe it's an old docstring format that doesn't work with the current Sphinx version.

lhoestq:
fix #524


lhoestq:
The problem came from `default: ` that is rendered differently and hides the parameter names. I changed `default: ...` to `defaults to ...`

thomwolf (approved):

PR #529, opened 2020-08-24
Add MLSUM

RachelKer:
Hello (again :) !), 

So, I started a new branch because of a [rebase issue](https://github.com/huggingface/nlp/pull/463), sorry for the mess. 

However, the command `pytest tests/test_dataset_common.py::LocalDatasetTest::test_load_real_dataset_mlsum` still fails because there is no default language dataset : the script throws an error as a specific config language is necessary. 

I think that setting a default language would be a bad workaround for this so I kept it as it is. Putting all the train files across languages together would also be a bad idea because of the size. 

Thanks for your help, 

Rachel


lhoestq:
Could you test to run the test using the changes in #527 and let me know if it fixes the issue ? If so I'll merge it and we'll be good to go :)

datasets/mlsum/mlsum.py
```diff
@@ -0,0 +1,133 @@
...
+                gen_kwargs={
+                    "filepath": os.path.join(downloaded_files["test"], lang+'_test.jsonl'),
+                    "lang": lang,
+                },
+            )
+        ]
+
+    def _generate_examples(self, filepath, lang):
+        """Yields examples."""
+        with open(filepath) as f:
```
lhoestq, line 121: To fix the CI you have to specify an encoding here (ex: `encoding="utf-8"`).
This is for windows users that may encounter issues with some codecs

RachelKer:
Hello, it does work on the fixing real dataset branch. Merci Quentin :)

lhoestq:
Nice, glad to hear that :)
de rien !

datasets/mlsum/mlsum.py
```diff
@@ -0,0 +1,133 @@
...
+        """
+        super(MlsumConfig, self).__init__(version=nlp.Version("1.0.0",), **kwargs)
+
+
+class Mlsum(nlp.GeneratorBasedBuilder):
+
+    VERSION = nlp.Version("1.0.0")
+    BUILDER_CONFIGS = (
+        [
+            MlsumConfig(
```
lhoestq, line 45: You can use `nlp.BuilderConfig` directly here and remove `MlsumConfig`.

lhoestq (commented):
Very nice ! 

For this dataset I think there's no need for a custom builder config. Could you use `nlp.BuilderConfig` directly please ?

lhoestq (approved):
Thank you !
Looks all good :)

PR #536, opened 2020-08-27
Fingerprint

lhoestq:
This PR is a continuation of #513 , in which many in-place functions were introduced or updated (cast_, flatten_) etc.
However the caching didn't handle these changes. Indeed the caching took into account only the previous cache file name of the table, and not the possible in-place transforms of the table.

To fix that, I added the concept of dataset fingerprint, that is updated after each transform (in place or not), and stored inside the table metadata.

When a dataset is created, an initial fingerprint is computed. If the dataset is memory-mapped, then the fingerprint generator doesn't read the table and only looks at the filename. However if the table is in-memory, then the fingerprint generator reads the content of the table using a batched non-crypto hashing.

I added a utility class to compute hashes of arbitrary python objects in `fingerprint.py` : `Hasher`. The API is close to standard hashing tools (`.update`, `.hexdigest`). It also supports custom hashing functions depending on object types using a registry like pickle. I added a custom hashing function to hash a `pa.Table` in a batched way, and also for `nlp.DatasetInfo` to leverage its json serialization feature.

Note about this PR:
This is a draft PR because #513 needs to be merged first.
The diff that is shown is for branches fingerprint -> indices (and not master, for now)


src/nlp/arrow_dataset.py
```diff
@@ -406,6 +443,7 @@ def flatten_(self, max_depth=16):
         logger.info(
             "Flattened dataset from depth {} to depth {}.".format(depth, 1 if depth + 1 < max_depth else "unknown")
         )
+        self._fingerprint = update_fingerprint(self._fingerprint, self.__class__.flatten_, {"max_depth": max_depth})
```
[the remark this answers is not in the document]
thomwolf: Couldn't this almost be a decorator wrapping the class methods?
So you just write `@fingerprint` in front of the methods updating the fingerprint and it uses the transform name and all the arguments?
(not sure about it but it would be easier to maintain)

src/nlp/arrow_writer.py
```diff
@@ -183,17 +185,23 @@ def _build_writer(self, inferred_schema: pa.Schema):
         if self.disable_nullable:
             self._schema = pa.schema(pa.field(field.name, field.type, nullable=False) for field in self._schema)
         if self.with_metadata:
-            self._schema = self._schema.with_metadata(self._build_metadata(DatasetInfo(features=self._features)))
+            self._schema = self._schema.with_metadata(
+                self._build_metadata(DatasetInfo(features=self._features), self.fingerprint)
```
thomwolf, line 189: In my early implementation I added the `fingerprint` field directly in the `DatasetInfo` structure.

What do you think?

src/nlp/fingerprint.py
```diff
@@ -0,0 +1,94 @@
+import json
+from dataclasses import asdict
+
+import pyarrow as pa
+import xxhash
```
thomwolf, line 5: Yeyy!

thomwolf (commented):
Ok this is really nice!

A few comments for (quick) brainstorms

src/nlp/arrow_writer.py
lhoestq: The fingerprint changes across machines as it sometimes hashes absolute paths.
I'm not sure this should be inside `DatasetInfo` that are meant to be shared

lhoestq:
I changed the way I implemented fingerprint updates to use decorator functions.

I also added a new attribute called `_inplace_history` that stores the in-place history of transforms (like cast_, rename_columns, etc.). This history is useful to replay the changes that were done in-place when unpickling a dataset that is memory mapped from a file.

Let me know what you think @thomwolf

src/nlp/fingerprint.py
```diff
@@ -0,0 +1,164 @@
...
+            # Call actual function
+
+            out = func(self, *args, **kwargs)
+
+            # Update fingerprint of in-place transforms + update in-place history of transforms
+
+            if inplace:  # update after calling func so that the fingerprint doesn't change if the function fails
+                self._fingerprint = update_fingerprint(self._fingerprint, func, kwargs_for_fingerprint)
+                for inplace_hist_per_file in self._inplace_history:
+                    inplace_hist_per_file["transforms"].append((func.__name__, args, kwargs))
```
thomwolf, line 157: Same as I mentioned earlier, I think I would update the fingerprint before calling the function in case some of the inner working in `func` do in-place changes on some of the args/kwargs then we might loose the original calling args. What do you think?

thomwolf (approved):
This is really cool! Just a quick remaining question

src/nlp/fingerprint.py
thomwolf: Or does the fact that we have already un updated fingerprint risk having some side effects in `func`?

lhoestq: Ok I see, I will compute the new fingerprint and new history before calling `func`, and then update `self._fingerprint` and `self._inplace_history`

lhoestq: I may need to `deepcopy` the args/kwargs that are saved in the history

lhoestq: Done. I ended up doing it using deep copy for the args/kwargs stored in the history

I'm merging the PR

Issue #537, opened 2020-08-27
[Dataset] RACE dataset Checksums error


abarbosa94:
Hi there, I just would like to use this awesome lib to perform a dataset fine-tuning on RACE dataset. I have performed the following steps:

```
dataset = nlp.load_dataset("race")
len(dataset["train"]), len(dataset["validation"])
```

But then I got the following error:

```
---------------------------------------------------------------------------
NonMatchingChecksumError                  Traceback (most recent call last)
<ipython-input-15-8bf7603ce0ed> in <module>
----> 1 dataset = nlp.load_dataset("race")
      2 len(dataset["train"]), len(dataset["validation"])

~/miniconda3/envs/masters/lib/python3.8/site-packages/nlp/load.py in load_dataset(path, name, version, data_dir, data_files, split, cache_dir, features, download_config, download_mode, ignore_verifications, save_infos, **config_kwargs)
    546 
    547     # Download and prepare data
--> 548     builder_instance.download_and_prepare(
    549         download_config=download_config, download_mode=download_mode, ignore_verifications=ignore_verifications,
    550     )

~/miniconda3/envs/masters/lib/python3.8/site-packages/nlp/builder.py in download_and_prepare(self, download_config, download_mode, ignore_verifications, try_from_hf_gcs, dl_manager, **download_and_prepare_kwargs)
    460                         logger.info("Dataset not on Hf google storage. Downloading and preparing it from source")
    461                 if not downloaded_from_gcs:
--> 462                     self._download_and_prepare(
    463                         dl_manager=dl_manager, verify_infos=verify_infos, **download_and_prepare_kwargs
    464                     )

~/miniconda3/envs/masters/lib/python3.8/site-packages/nlp/builder.py in _download_and_prepare(self, dl_manager, verify_infos, **prepare_split_kwargs)
    519         # Checksums verification
    520         if verify_infos:
--> 521             verify_checksums(
    522                 self.info.download_checksums, dl_manager.get_recorded_sizes_checksums(), "dataset source files"
    523             )

~/miniconda3/envs/masters/lib/python3.8/site-packages/nlp/utils/info_utils.py in verify_checksums(expected_checksums, recorded_checksums, verification_name)
     36     if len(bad_urls) > 0:
     37         error_msg = "Checksums didn't match" + for_verification_name + ":\n"
---> 38         raise NonMatchingChecksumError(error_msg + str(bad_urls))
     39     logger.info("All the checksums matched successfully" + for_verification_name)
     40 

[...]

PR #540, opened 2020-08-29
[BUGFIX] Fix Race Dataset Checksum bug

abarbosa94:
In #537 I noticed that there was a bug in checksum checking when I have tried to download the race dataset. The reason for this is that the current preprocessing was just considering the `high school` data and it was ignoring the `middle` one. This PR just fixes it :)

Moreover, I have added some descriptions.


lhoestq:
I'm not sure this would fix #537 .
However your point about the missing `middle` data is right and we probably want to include these data as well.
Do you think it would we worth having different configurations for this dataset for users who want to only load part of it (`high school` or `middle` or `all`) ?

abarbosa94:
This has fixed #537 at least on my machine hahaha.

Nice point! I think it would totally worth it :) What the best implementation approach would you suggest?

Would it be possible to have `high school`, `middle` and  `all` inside each portion of `train`, `validation` and `test`? Would this make sense?

lhoestq:
I think we could have one dataset configuration for `high school`, one for `middle` and one for `all`.
You just need to add

```python

    BUILDER_CONFIGS = [
        HansardsConfig(
            name="house",
            description="""\
          Alignment of debates in the House of the 36th Canadian Parliament: 1,070K sentence pairs.
          """,
        ),
        HansardsConfig(
            name="senate",
            description="""\
          Alignment of debates in the Senate of the 36th Canadian Parliament: 208K sentence pairs.
          """,
        ),
    ]

**Comment by abarbosa94** [comment]
Hi @lhoestq sorry for the delay in addressing your comments. Thanks for your assistance :)

You were correct as well, as I was using the script without the `datasets/race/dataset_infos.json` file, it did not verify the checksum. I already fix it as well :)

I managed to get everything running smoothly by now. Please let me know if you think that I could improve my solution

File: datasets/race/race.py [hunk 1]
```diff
@@ -61,43 +69,36 @@ def _info(self):
...
 
-    def _generate_examples(self, files, filespath):
+    def _generate_examples(self, train_test_or_eval):
         """Yields examples."""
-        # TODO(race): Yields (key, example) tuples from the dataset
-        for file in files:
-            filepath = os.path.join(filespath, file)
-            with open(filepath, encoding="utf-8") as f:
+        current_path = Path(train_test_or_eval)
+        files_in_dir = [str(f.absolute()) for f in current_path.glob("**/*.txt")]
```
lhoestq, line 99:
```suggestion
        files_in_dir = [str(f.absolute()) for f in sorted(current_path.glob("**/*.txt"))]
```
We use sorted to make the files order deterministic

datasets/race/dataset_infos.json
```diff
@@ -1 +1 @@
...
...,\n    year={2017}\n}\n", "homepage": "http://www.cs.cmu.edu/~glai1/data/race/", "license": "", "features": {"article": {"dtype": "string", "id": null, "_type": "Value"}, "answer": {"dtype": "string", "id": null, "_type": "Value"}, "question": {"dtype": "string", "id": null, "_type": "Value"}, "options": {"feature": {"dtype": "string", "id": null, "_type": "Value"}, "length": -1, "id": null, "_type": "Sequence"}}, "supervised_keys": null, "builder_name": "race", "config_name": "default", "version": {"version_str": "0.1.0", "description": null, "datasets_version_to_prepare": null, "major": 0, "minor": 1, "patch": 0}, "splits": {"test": {"name": "test", "num_bytes": 8775394, "num_examples": 4934, "dataset_name": "race"}, "train": {"name": "train", "num_bytes": 157308694, "num_examples": 87866, "dataset_name": "race"}, "validation": {"name": "validation", "num_bytes": 8647200, "num_examples": 4887, "dataset_name": "race"}}, "download_checksums": {"http://www.cs.cmu.edu/~glai1/data/race/RACE.tar.gz": {"num_bytes": 25443609, "checksum": "b2769cc9fdc5c546a693300eb9a966cec6870bd349fbc44ed5225f8ad33006e5"}}, "download_size": 25443609, "dataset_size": 139047912, "size_in_bytes": 227675112}}
```
lhoestq, line 1: Could you update the dataset_info please ? So that it includes the new configurations

To do so, please delete this file and regenerate it with 
```
datasets-cli test ./race --save_infos --all_configs
```

abarbosa94: sure! 👍

abarbosa94: Done in here d6fbac0 @lhoestq 

However, I was wondering the following: I just tested it locally and noticed that if I tried to use the lib as:
`datasets.load_dataset('race')` instead of `datasets.load_dataset('race', 'all')` it still throws the checksum error :( 

I was expecting some kind of warning/error logging that I should pass either `all`; `high` or `middle`. Have I done something wrong?

Could you please advise me? Then I think that I could fix it :)

lhoestq: `datasets.load_dataset('race')` loads the dataset script that is on master if you installed the project using git clone.
It you try `datasets.load_dataset("./datasets/race")` however it should run the local script that you changed

abarbosa94: cool! Thanks 👍 

Just check it again. Everything seems ok by now.

datasets/race/dataset_infos.json
```diff
@@ -1 +1 @@
...
...", "license": "", "features": {"example_id": {"dtype": "string", "id": null, "_type": "Value"}, "article": {"dtype": "string", "id": null, "_type": "Value"}, "answer": {"dtype": "string", "id": null, "_type": "Value"}, "question": {"dtype": "string", "id": null, "_type": "Value"}, "options": {"feature": {"dtype": "string", "id": null, "_type": "Value"}, "length": -1, "id": null, "_type": "Sequence"}}, "post_processed": null, "supervised_keys": null, "builder_name": "race", "config_name": "all", "version": {"version_str": "0.1.0", "description": null, "major": 0, "minor": 1, "patch": 0}, "splits": {"test": {"name": "test", "num_bytes": 8775394, "num_examples": 4934, "dataset_name": "race"}, "train": {"name": "train", "num_bytes": 157308694, "num_examples": 87866, "dataset_name": "race"}, "validation": {"name": "validation", "num_bytes": 8647200, "num_examples": 4887, "dataset_name": "race"}}, "download_checksums": {"http://www.cs.cmu.edu/~glai1/data/race/RACE.tar.gz": {"num_bytes": 25443609, "checksum": "b2769cc9fdc5c546a693300eb9a966cec6870bd349fbc44ed5225f8ad33006e5"}}, "download_size": 25443609, "post_processing_size": null, "dataset_size": 174731288, "size_in_bytes": 200174897}}
```
lhoestq, line 1: Thanks ofr adding the new configurations to the json file :) 

Could you remove the 'default' configuration from this json file ?
To do so you can delete the json file and regenerate it.
Or you can simply remove this field from the json file ^^

abarbosa94: 75d7278 done! :)

I removed the field by hand. I tried to regenerate the file but for some reason, the `default`section was being generated as well ): Do you have any idea why would this be occurring?

lhoestq: Thanks :)

Actually it was still there because it was reusing a cached version of the json file

lhoestq (approved):
Thank you for updating the checksums and adding the new configurations :)

PR #542, opened 2020-08-29
Add TensorFlow example

jplu:
Update the Quick Tour documentation in order to add the TensorFlow equivalent source code for the classification example. Now it is possible to select either the code in PyTorch or in TensorFlow in the Quick tour.


thomwolf (approved):
Very nice @jplu!

lhoestq (approved):
Cool thanks :)

PR #547, opened 2020-08-31
[Distributed] Making loading distributed datasets a bit safer

thomwolf:
Add some file-locks during dataset loading


lhoestq (approved):
filelocks everywhere :o 

Thanks for fixing the test !

PR #548, opened 2020-08-31
[Breaking] Switch text loading to multi-threaded PyArrow loading

### LINKED ISSUE #546: Very slow data loading on large dataset
I made a simple python script to check the NLP library speed, which loads 1.1 TB of textual data.
It has been 8 hours and still, it is on the loading steps.
It does work when the text dataset size is small about  1 GB, but it doesn't scale.
It also uses a single thread during the data loading step.

```
train_files = glob.glob("xxx/*.txt",recursive=True)
random.shuffle(train_files)

print(train_files)

dataset = nlp.load_dataset('text', 
                           data_files=train_files,
                           name="customDataset",
                           version="1.0.0",
                           cache_dir="xxx/nlp")
```

Is there something that I am missing ?

thomwolf:
Test if we can get better performances for large-scale text datasets by using multi-threaded text file loading based on Apache Arrow multi-threaded CSV loader.

If it works ok, it would fix #546.

**Breaking change**:
The text lines now do not include final line-breaks anymore.


lhoestq:
Awesome !
Also I was wondering if we should try to make the hashing of the `data_files` faster (it is used to build the cache directory of datasets like `text` or `json`). Right now it reads each file and hashes all of its data. We could simply hash the path and some metadata including the `last updated` tag no ? Apparently we can get this tag with `os.path.getmtime(path)`

lhoestq:
I just rebased from master to include the hashing changes from #573

thomwolf:
I think this is ready to merge, no?

lhoestq:
Indeed it's ready to merge :)

thomwolf:
Ok added the breaking change info and we can merge indeed.

PR #549, opened 2020-09-01
Fix bleurt logging import

jbragg:
Bleurt started throwing an error in some code we have.
This looks like the fix but...

It's also unnerving that even a prebuilt docker image with pinned versions can be working 1 day and then fail the next (especially for production systems).

Any way for us to pin your metrics code so that they are guaranteed not to to change and possibly fail on repository changes?

Thanks (and also for your continued work on the lib...)

### DIFF
--- a/metrics/bleurt/bleurt.py
+++ b/metrics/bleurt/bleurt.py
@@ -17,7 +17,7 @@
 import os
 
 import nlp
-from nlp.logging import get_logger
+from nlp.utils.logging import get_logger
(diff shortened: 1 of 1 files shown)


thomwolf:
That’s a good point that we started to discuss internally as well. We should pin the dataset en metrics code by default indeed.
Let’s update this in the coming release.

thomwolf:
Ok closed this with #567 and we are working on a more general solution to pin dataset version in #562 (should be in the coming release).

PR #556, opened 2020-09-01
Add DailyDialog

julien-c:
http://yanran.li/dailydialog.html

https://arxiv.org/pdf/1710.03957.pdf


datasets/daily_dialog/daily_dialog.py
```diff
@@ -0,0 +1,157 @@
...
+            for i, (line_f, line_act, line_emotion) in enumerate(zip(f, act, emotion)):
+                if len(line_f.strip()) == 0:
+                    break
+                dialog = line_f.split(self.__EOU__)[:-1]
+                act = line_act.split(" ")[:-1]
+                emotion = line_emotion.split(" ")[:-1]
+
+                assert len(dialog) == len(act) == len(emotion), "Different turns btw dialogue & emotion & action"
+
+                yield f"{split}-i", {
```
lhoestq, line 153:
```suggestion
                yield f"{split}-{i}", {
```

julien-c: Good catch 😱

Looks like there's no check of the unicity of the generated index?

lhoestq: We don't use the example ids in practice. This is (yet) another thing that comes from TFDS

PR #562, opened 2020-09-02
[Reproductibility] Allow to pin versions of datasets/metrics

thomwolf:
Repurpose the `version` attribute in datasets and metrics to let the user pin a specific version of datasets and metric scripts:
```
dataset = nlp.load_dataset('squad', version='1.0.0')
metric = nlp.load_metric('squad', version='1.0.0')
```

Notes:
- version number are the release version of the library
- currently only possible for canonical datasets/metrics, ie. integrated in the GitHub repo of the library

### DIFF
--- a/src/nlp/load.py
+++ b/src/nlp/load.py
@@ -42,7 +42,7 @@ from .splits import Split
 from .utils.download_manager import GenerateMode
 from .utils.file_utils import DownloadConfig, cached_path, hf_bucket_url
 from .utils.logging import get_logger
-
+from .utils.version import Version
 
 logger = get_logger(__name__)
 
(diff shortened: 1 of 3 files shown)


src/nlp/load.py
```diff
@@ -530,7 +535,7 @@ def load_dataset(
     """
     ignore_verifications = ignore_verifications or save_infos
     # Download/copy dataset processing script
-    module_path, hash = prepare_module(path, download_config=download_config, dataset=True)
+    module_path, hash = prepare_module(path, download_config=download_config, version=version, dataset=True)
```
lhoestq, line 538: I guess we could call `prepare_module` with the github url that is already well formatted, instead of passing `version` to `prepare_module` then `cached_path` then `get_from_cache` ? Not sure that `cached_path` should have the `version` logic

lhoestq:
Closing this one in favor of #584

PR #564, opened 2020-09-02
Wait for writing in distributed metrics

lhoestq:
There were CI bugs where a distributed metric would try to read all the files in process 0 while the other processes haven't started writing.

To fix that I added a custom locking mechanism that waits for the file to exist before trying to read it

### DIFF
--- a/src/nlp/metric.py
+++ b/src/nlp/metric.py
@@ -240,7 +240,7 @@ class Metric(object):
             nofilelock.acquire(timeout=self.timeout)
         except Timeout:
             raise ValueError(
-                f"Expeced to find locked file {expected_lock_file_name} from process {self.process_id} but it doesn't exist."
+                f"Expected to find locked file {expected_lock_file_name} from process {self.process_id} but it doesn't exist."
             )
         else:
             nofilelock.release()
--- a/src/nlp/metric.py
+++ b/src/nlp/metric.py
@@ -192,18 +192,7 @@ class Metric(object):
         # Let's acquire a lock on each process files to be sure they are finished writing
         filelocks = []
         for process_id, file_path in enumerate(file_paths):
-            nofile_lock = NoFileLock(file_path)
             filelock = FileLock(file_path + ".lock")
-
-            # Check that writing has started
-            try:
-                nofile_lock.acquire(timeout=timeout)
-            except Timeout:
-                raise ValueError(f"Expected to find file {file_path} but it doesn't for process {process_id}, .")
-            else:
-                nofile_lock.release()
-
-            # Check that writing has finished
             try:
                 filelock.acquire(timeout=timeout)
             except Timeout:
@@ -213,10 +202,28 @@ class Metric(object):
 
         return file_paths, filelocks
 
+    def _check_all_processes_locks(self, timeout=100):
+        expected_lock_file_names = [
+            os.path.join(self.data_dir, f"{self.experiment_id}-{self.num_process}-{process_id}.arrow.lock")
+            for process_id in range(self.num_process)
+        ]
+        for expected_lock_file_name in expected_lock_file_names:
+            nofilelock = NoFileLock(expected_lock_file_name)
+            try:
+                nofilelock.acquire(timeout=timeout)
+            except Timeout:
+                raise ValueError(
+                    f"Expeced to find locked file {expected_lock_file_name} from process {self.process_id} but it doesn't exist."
+                )
+            else:
+                nofilelock.release()
+
     def finalize(self, timeout=100):
         """Close all the writing process and load/gather the data
         from all the nodes if main node or all_process is True.
         """
+        if self.num_process > 1:
+            self._check_all_processes_locks(timeout=timeout)
         if self.writer is not None:
             self.writer.finalize()
         self.writer = None
--- a/src/nlp/metric.py
+++ b/src/nlp/metric.py
@@ -245,13 +245,13 @@ class Metric(object):
         else:
             nofilelock.release()
         lock_file_name = os.path.join(self.data_dir, f"{self.experiment_id}-{self.num_process}-rdv.lock")
-        renndez_vous_lock = FileLock(lock_file_name)
+        rendez_vous_lock = FileLock(lock_file_name)
         try:
-            renndez_vous_lock.acquire(timeout=self.timeout)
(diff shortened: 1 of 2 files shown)


thomwolf:
I agree this fix the problem for the CI where the files are always created in a new and clean temporary directory.

However, in a general setting of a succession of fast distributed operation, the file could already exist from previous metrics run but one process may still finish before another has even started in which case it would mix results for disparate operations.

I feel like the most robust way to solve this is to setup a rendez-vous in the first writing on files, in which each process will test and only finish its operations when it cannot acquire a lock on all the other processes.

What do you think?

lhoestq:
What do you think of this @thomwolf  ? I check all the locks before finalizing

src/nlp/metric.py
```diff
@@ -38,6 +38,17 @@
 logger = get_logger(__file__)
 
 
+class NoFileLock(BaseFileLock):
+    """Thread lock until file exists"""
+
+    def _acquire(self):
+        if os.path.exists(self._lock_file):
+            self._lock_file_fd = self._lock_file
```
thomwolf, line 46: The thing is that the `_lock_file` persiste after the lock is release (it's not deleted) so you have to make a test that you cannot acquire the lock. Maybe the easiest is to do a wrapper around a FileLock object.

lhoestq: Maybe I can replace `os.path.exists` with something that tries to open the file ?

thomwolf: That's what the various platform-specific classes inherited from BaseFileLock do

thomwolf:
Ok on my side @lhoestq (cannot add you as a reviewer)

lhoestq:
The test doesn't pass if I add:
```python
            import time
            if self.process_id == 1:
                time.sleep(0.5)
```
right before `self.add_batch` in `Metric.compute`.

I'm investigating why it doesn't work in that case

lhoestq:
It looks like the process 1 runs `_check_all_processes_locks` correctly and then finishes and releases its lock before process 0 even managed to to run `_check_all_processes_locks` correctly.

thomwolf:
Strange!

lhoestq:
I changed the way the rendez-vous is done @thomwolf , let me know what you think.
The idea is that the master process has an additional lock `rendez_vous_lock` to tell every other process to wait for everyone to be ready before starting to write

src/nlp/metric.py
```diff
@@ -114,7 +114,7 @@ def __init__(
         self._data_dir_root = os.path.expanduser(data_dir or HF_METRICS_CACHE)
         self.data_dir = self._build_data_dir()
         self.seed: int = seed or np.random.get_state()[1][0]
-        self.timeout: Union[int, float] = 100
+        self.timeout: Union[int, float] = timeout
```
lhoestq, line 117: it took me time to find why the test was so slow x)

src/nlp/metric.py
```diff
@@ -183,15 +209,51 @@ def _get_all_cache_files(self, timeout=100) -> Tuple[List[str], List[FileLock]]:
...
+            os.path.join(self.data_dir, f"{self.experiment_id}-{self.num_process}-{process_id}.arrow.lock")
+            for process_id in range(self.num_process)
+        ]
+        for expected_lock_file_name in expected_lock_file_names:
+            nofilelock = FileFreeLock(expected_lock_file_name)
+            try:
+                nofilelock.acquire(timeout=self.timeout)
+            except Timeout:
+                raise ValueError(
+                    f"Expeced to find locked file {expected_lock_file_name} from process {self.process_id} but it doesn't exist."
```
thomwolf, line 231:
```suggestion
                    f"Expected to find locked file {expected_lock_file_name} from process {self.process_id} but it doesn't exist."
```

src/nlp/metric.py
```diff
@@ -183,15 +209,51 @@ def _get_all_cache_files(self, timeout=100) -> Tuple[List[str], List[FileLock]]:
...
+        try:
+            nofilelock.acquire(timeout=self.timeout)
+        except Timeout:
+            raise ValueError(
+                f"Expeced to find locked file {expected_lock_file_name} from process {self.process_id} but it doesn't exist."
+            )
+        else:
+            nofilelock.release()
+        lock_file_name = os.path.join(self.data_dir, f"{self.experiment_id}-{self.num_process}-rdv.lock")
+        renndez_vous_lock = FileLock(lock_file_name)
```
thomwolf, line 248:
```suggestion
        rendez_vous_lock = FileLock(lock_file_name)
```

src/nlp/metric.py
```diff
@@ -183,15 +209,51 @@ def _get_all_cache_files(self, timeout=100) -> Tuple[List[str], List[FileLock]]:
...
+        try:
+            renndez_vous_lock.acquire(timeout=self.timeout)
```
thomwolf, line 250:
```suggestion
            rendez_vous_lock.acquire(timeout=self.timeout)
```

src/nlp/metric.py
```diff
@@ -183,15 +209,51 @@ def _get_all_cache_files(self, timeout=100) -> Tuple[List[str], List[FileLock]]:
...
+        except Timeout:
+            raise ValueError(f"Couldn't acquire lock on {lock_file_name} from process {self.process_id}.")
+        else:
+            renndez_vous_lock.release()
```
thomwolf, line 254:
```suggestion
            rendez_vous_lock.release()
```

src/nlp/metric.py
```diff
@@ -183,15 +209,51 @@ def _get_all_cache_files(self, timeout=100) -> Tuple[List[str], List[FileLock]]:
...
+                nofilelock.release()
+
+    def _check_rendez_vous(self):
+        expected_lock_file_name = os.path.join(self.data_dir, f"{self.experiment_id}-{self.num_process}-0.arrow.lock")
+        nofilelock = FileFreeLock(expected_lock_file_name)
+        try:
+            nofilelock.acquire(timeout=self.timeout)
+        except Timeout:
+            raise ValueError(
+                f"Expeced to find locked file {expected_lock_file_name} from process {self.process_id} but it doesn't exist."
```
thomwolf, line 243:
```suggestion
                f"Expected to find locked file {expected_lock_file_name} from process {self.process_id} but it doesn't exist."
```

thomwolf (approved):
Ok, very nice!

PR #566, opened 2020-09-02
Remove logger pickling to fix gg colab issues

lhoestq:
A `logger` objects are not picklable in google colab, contrary to `logger` objects in jupyter notebooks or in python shells.
It creates some issues in google colab right now.

Indeed by calling any `Dataset` method, the fingerprint update pickles the transform function, and as the logger comes with it, it results in an error (full stacktrace [here](http://pastebin.fr/64330)):

```python
/usr/local/lib/python3.6/dist-packages/zmq/backend/cython/socket.cpython-36m-x86_64-linux-gnu.so in zmq.backend.cython.socket.Socket.__reduce_cython__()

TypeError: no default __reduce__ due to non-trivial __cinit__
```

To fix that I no longer dump the transform (`_map_single`, `select`, etc.), but the full name only (`nlp.arrow_dataset.Dataset._map_single`, `nlp.arrow_dataset.Dataset.select`, etc.)

### DIFF
--- a/src/nlp/fingerprint.py
+++ b/src/nlp/fingerprint.py
@@ -142,15 +142,17 @@ def fingerprint(inplace, use_kwargs=None, ignore_kwargs=None, fingerprint_names=
                     kwargs_for_fingerprint["generator"] = np.random.default_rng(None)
 
             # compute new_fingerprint and add it to the args of not in-place transforms
-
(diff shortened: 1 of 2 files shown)


thomwolf (approved):
Yep!

PR #573, opened 2020-09-04
Faster caching for text dataset

lhoestq:
As mentioned in #546 and #548 , hashing `data_files` contents to get the cache directory name for a text dataset can take a long time.

To make it faster I changed the hashing so that it takes into account the `path` and the `last modified timestamp` of each data file, instead of iterating through the content of each file to get a hash.

### DIFF
--- a/src/nlp/builder.py
+++ b/src/nlp/builder.py
@@ -269,9 +269,8 @@ class DatasetBuilder:
             for key in sorted(data_files.keys()):
                 m.update(key.encode("utf-8"))
                 for data_file in data_files[key]:
-                    with open(data_file, "rb") as f:
-                        for chunk in iter(lambda: f.read(1 << 20), b""):
-                            m.update(chunk)
+                    m.update(os.path.abspath(data_file).encode("utf-8"))
+                    m.update(str(os.path.getmtime(data_file)).encode("utf-8"))
             builder_config.name += "-" + m.hexdigest()
         return builder_config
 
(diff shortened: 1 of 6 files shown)


src/nlp/builder.py
```diff
@@ -644,41 +643,46 @@ def _build_single_dataset(self, split: Union[str, Split], run_post_process: bool
...
+                ds._info.size_in_bytes = self.info.size_in_bytes
+                if self.info.post_processed.features is not None:
+                    if self.info.post_processed.features.type != ds.features.type:
+                        raise ValueError(
+                            "Post-processed features info don't match the dataset:\nGot\n{}\nbut expected something like\n{}".format(
+                                self.info.post_processed.features, ds.features
+                            )
+                        )
+                    else:
+                        ds.info.features = self.info.post_processed.features
```
lhoestq, line 685: When I added the unit test to check the good caching behavior of text datasets, I found that the `DatasetInfo` were not properly updated if there was a post processing, so I fixed that

tests/test_search.py
```diff
@@ -5,47 +5,15 @@
...
-        train_info = SplitInfo(name="train", num_examples=30)
-        test_info = SplitInfo(name="test", num_examples=30)
-        split_infos = [train_info, test_info]
-        split_dict = SplitDict()
-        split_dict.add(train_info)
-        split_dict.add(test_info)
-        info = DatasetInfo(splits=split_dict)
-        reader = ReaderTester("", info)
-        dset = Dataset(**reader.read(name, "train", split_infos))
+        dset = Dataset.from_dict({"filename": ["my_name-train" + "_" + str(x) for x in np.arange(30).tolist()]})
```
lhoestq, line 16: I simplified this test too

src/nlp/builder.py
thomwolf: Good catch!

src/nlp/fingerprint.py
```diff
@@ -89,6 +90,9 @@ def generate_fingerprint(dataset):
             continue
         hasher.update(key)
         hasher.update(state[key])
+    # hash data files last modification timestamps as well
+    for data_file in state.get("_data_files", []) + state.get("_indices_data_files", []):
+        hasher.update(os.path.getmtime(data_file["filename"]))
```
thomwolf, line 95: Nice indeed!

thomwolf (approved):
Awesome!

Issue #554, opened 2020-09-01
nlp downloads to its module path


danieldk:
I am trying to package `nlp` for Nix, because it is now an optional dependency for `transformers`. The problem that I encounter is that the `nlp` library downloads to the module path, which is typically not writable in most package management systems:

```>>> import nlp
>>> squad_dataset = nlp.load_dataset('squad')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/nix/store/2yhik0hhqayksmkkfb0ylqp8cf5wa5wp-python3-3.8.5-env/lib/python3.8/site-packages/nlp/load.py", line 530, in load_dataset
    module_path, hash = prepare_module(path, download_config=download_config, dataset=True)
  File "/nix/store/2yhik0hhqayksmkkfb0ylqp8cf5wa5wp-python3-3.8.5-env/lib/python3.8/site-packages/nlp/load.py", line 329, in prepare_module
    os.makedirs(main_folder_path, exist_ok=True)
  File "/nix/store/685kq8pyhrvajah1hdsfn4q7gm3j4yd4-python3-3.8.5/lib/python3.8/os.py", line 223, in makedirs
    mkdir(name, mode)
OSError: [Errno 30] Read-only file system: '/nix/store/2yhik0hhqayksmkkfb0ylqp8cf5wa5wp-python3-3.8.5-env/lib/python3.8/site-packages/nlp/datasets/squad'
```

Do you have any suggested workaround for this issue?

Perhaps overriding the default value for `force_local_path` of `prepare_module`?

thomwolf:
Indeed this is a know issue with the fact that we try to be compatible with cloupickle.

Does this also happen if you are installing in a virtual environment?

danieldk:
> Indeed this is a know issue with the fact that we try to be compatible with cloupickle.
> 
> Does this also happen if you are installing in a virtual environment?

Then it would work, because the package is in a writable path.

thomwolf:
If it's fine for you then this is the recommended way to solve this issue.

danieldk:
> If it's fine for you then this is the recommended way to solve this issue.

I don't want to use a virtual environment but Nix, because Nix is fully reproducible. And I am the maintainer of the `transformers` in nixpkgs, so sooner or later I will have to package `nlp` ;).

thomwolf:
[...]

PR #574, opened 2020-09-04
Add modules cache

lhoestq:
As discusses in #554 , we should use a module cache directory outside of the python packages directory since we may not have write permissions.

I added a new HF_MODULES_PATH directory that is added to the python path when doing `import nlp`.
In this directory, a module `nlp_modules` is created so that datasets can be added to `nlp_modules.datasets` and metrics to `nlp_modules.metrics`. `nlp_modules` doesn't exist on Pypi.

If someone using cloudpickle still wants to have the downloaded dataset/metrics scripts to be inside the nlp directory, it is still possible to change the environment variable HF_MODULES_PATH to be a path inside the nlp lib.

### DIFF
--- a/src/nlp/load.py
+++ b/src/nlp/load.py
@@ -37,17 +37,24 @@ from .info import DATASET_INFOS_DICT_FILE_NAME
 from .metric import Metric
 from .splits import Split
 from .utils.download_manager import GenerateMode
-from .utils.file_utils import DownloadConfig, cached_path, hf_bucket_url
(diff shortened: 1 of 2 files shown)


lhoestq:
All the tests pass on my side. Not sure if it is a cache issue or a circleci issue

thomwolf:
I think I fixed it (sorry didn't notice you were on it as well)

thomwolf (approved):
Perfect!

PR #578, opened 2020-09-07
Add CommonGen Dataset

JetRunner:
CC Authors:
@yuchenlin @MichaelZhouwang


datasets/common_gen/common_gen.py
```diff
@@ -0,0 +1,105 @@
...
+to explicitly test machines for the ability of generative commonsense reasoning. Given 
+a set of common concepts; the task is to generate a coherent sentence describing an 
+everyday scenario using these concepts.
+
+CommonGen is challenging because it inherently requires 1) relational reasoning using 
+background commonsense knowledge, and 2) compositional generalization ability to work 
+on unseen concept combinations. Our dataset, constructed through a combination of 
+crowd-sourcing from AMT and existing caption corpora, consists of 30k concept-sets and 
+50k sentences in total.
+"""
```
JetRunner, line 32: Feel free to edit this if you think it's necessary. @yuchenlin @MichaelZhouwang

yuchenlin: Thank you for the help!

datasets/common_gen/common_gen.py
```diff
@@ -0,0 +1,105 @@
+from __future__ import absolute_import, division, print_function
+
+import json
+import os
+import random
+
+import nlp
+
+random.seed(42)  # This is important, to ensure the same order for concept sets as the official script.
```
lhoestq, line 9: You can use `random.Random()` to create a local random numbers generator and avoid setting the seed globally.

In your case instead of doing 

```python
random.shuffle(rand_order)
```

you could do 

```python
rng = random.Random(42)
for idx, row in enumerate(f):
    ...
    rng.shuffle(rand_order)
    ...
```

lhoestq (approved):
Nice, good job !

I just one one comment about the random seed that should be set locally imo

datasets/common_gen/common_gen.py
JetRunner: Thanks for the suggestion but I'd prefer to keep it consistent with the original `formatter.py` (https://github.com/INK-USC/CommonGen/blob/master/dataset/final_data/formatter.py)

PR #585, opened 2020-09-07
Fix select for pyarrow < 1.0.0

### LINKED ISSUE #583: ArrowIndexError on Dataset.select
If the indices table consists in several chunks, then `dataset.select` results in an `ArrowIndexError` error for pyarrow < 1.0.0

Example:

```python
from nlp import load_dataset

mnli = load_dataset("glue", "mnli", split="train")
shuffled = mnli.shuffle(seed=42)
mnli.select(list(range(len(mnli))))
```

raises:
```python
ArrowIndexError                           Traceback (most recent call last)
<ipython-input-64-006a5d38d418> in <module>
----> 1 mnli.shuffle(seed=42).select(list(range(len(mnli))))

~/Desktop/hf/nlp/src/nlp/fingerprint.py in wrapper(*args, **kwargs)
    161             # Call actual function
    162 
--> 163             out = func(self, *args, **kwargs)
    164 
    165             # Update fingerprint of in-place transforms + update in-place history of transforms

~/Desktop/hf/nlp/src/nlp/arrow_dataset.py in select(self, indices, keep_in_memory, indices_cache_file_name, writer_batch_size, new_fingerprint)
   1653         if self._indices is not None:
   1654             if PYARROW_V0:
-> 1655                 indices_array = self._indices.column(0).chunk(0).take(indices_array)
   1656             else:
   1657                 indices_array = self._indices.column(0).take(indices_array)

~/.virtualenvs/hf-datasets/lib/python3.7/site-packages/pyarrow/array.pxi in pyarrow.lib.Array.take()

~/.virtualenvs/hf-datasets/lib/python3.7/site-packages/pyarrow/error.pxi in pyarrow.lib.check_status()

ArrowIndexError: take index out of bounds
```

This is because the `take` method is only done on the first chunk which only contains 1000 elements by default (mnli has ~400 000 elements).

Shall we change that to use 
```python
pa.concat_tables(self._indices._indices.slice(i, 1) for i in indices_array)
```
instead of `take` ? @thomwolf

lhoestq:
Fix #583

### DIFF
--- a/datasets/common_gen/common_gen.py
+++ b/datasets/common_gen/common_gen.py
@@ -6,6 +6,7 @@ import random
 
 import nlp
 
+
 random.seed(42)  # This is important, to ensure the same order for concept sets as the official script.
 
 _CITATION = """\
--- a/datasets/doqa/doqa.py
+++ b/datasets/doqa/doqa.py
@@ -17,9 +17,9 @@
 """CRD3  dataset"""
 
 from __future__ import absolute_import, division, print_function
-import logging
 
 import json
+import logging
 import os
 
 import nlp
@@ -177,7 +177,3 @@ class Doqa(nlp.GeneratorBasedBuilder):
                                 "answer_start": orig_answer_start,
                         },
                         }
-                        
-                        
-                
-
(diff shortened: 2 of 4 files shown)


thomwolf (approved):
Indeed, good catch!

PR #596, opened 2020-09-09
[style/quality] Moving to isort 5.0.0 + style/quality on datasets and metrics

thomwolf:
Move the repo to isort 5.0.0.

Also start testing style/quality on datasets and metrics.

Specific rule: we allow F401 (unused imports) in metrics to be able to add imports to detect early on missing dependencies.
Maybe we could add this in datasets but while cleaning this I've seen many example of really unused imports in dataset so maybe it's better to have it as a line-by-line nova instead of a general rule like in metrics.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,8 +34,7 @@ jobs:
         parallelism: 1
         steps:
             - checkout
-            # we need a version of isort with https://github.com/timothycrosley/isort/pull/1000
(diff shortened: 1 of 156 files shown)


thomwolf:
Ready for review @lhoestq, just updated a few 156 files here

lhoestq (approved):
Nice !

PR #599, opened 2020-09-10
Add MATINF dataset

JetRunner:
@lhoestq The command to create metadata failed. I guess it's because the zip is not downloaded from a remote address? How to solve that? Also the CI fails and I don't know how to fix that :(


lhoestq:
Hi ! sorry for the late response

Could you try to rebase from master ? We changed the named of the library last week so you have to include this change in your code.

Can you give me more details about the error you get when running the cli command ?

Note that in case of a manual download you have to specify the directory where you downloaded the data with `--datadir <path/to/the/directory>`

JetRunner:
I fucked up the Git rebase lol. Closing it.

PR #603, opened 2020-09-10
Set scripts version to master

lhoestq:
By default the scripts version is master, so that if the library is installed with 
```
pip install git+http://github.com/huggingface/nlp.git
```
or
```
git clone http://github.com/huggingface/nlp.git
pip install -e ./nlp
```

will use the latest scripts, and not the ones from the previous version.


thomwolf (approved):
Ok!

PR #605, opened 2020-09-10
[Datasets] Transmit format to children

thomwolf:
Transmit format to children obtained when processing a dataset.

Added a test.

When concatenating datasets, if the formats are disparate, the concatenated dataset has a format reset to defaults.


lhoestq:
Closing as #607 was merged

PR #607, opened 2020-09-10
Add transmit_format wrapper and tests

lhoestq:
Same as #605 but using a decorator on-top of dataset transforms that are not in place

### DIFF
--- a/src/nlp/arrow_dataset.py
+++ b/src/nlp/arrow_dataset.py
@@ -25,7 +25,7 @@ import tempfile
 from collections import defaultdict
 from collections.abc import Iterable, Mapping
 from dataclasses import asdict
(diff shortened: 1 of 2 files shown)


src/nlp/arrow_dataset.py
```diff
@@ -737,6 +775,7 @@ def set_format(
                 )
             )
 
+        format_kwargs.update(format_kwargs.pop("format_kwargs", {}))  # allow to use self.set_format(self.format)
```
lhoestq, line 778: I had to add this line to make `format_kwargs` work when doing `self.set_format(self.format)`

thomwolf (approved):
Awesome!

PR #618, opened 2020-09-11
sync logging utils with transformers

stas00:
sync the docs/code with the recent changes in transformers' `logging` utils:
1. change the default level to `WARNING`
2. add `DATASETS_VERBOSITY` env var
3. expand docs

### DIFF
--- a/docs/source/package_reference/logging_methods.rst
+++ b/docs/source/package_reference/logging_methods.rst
@@ -29,7 +29,7 @@ verbose to the most verbose), those levels (with their corresponding int values 
   critical errors.
 - :obj:`datasets.logging.ERROR` (int value, 40): only report errors.
 - :obj:`datasets.logging.WARNING` or :obj:`datasets.logging.WARN` (int value, 30): only reports error and
-  warnings. This the default level used by the library.
+  warnings.
 - :obj:`datasets.logging.INFO` (int value, 20): reports error, warnings and basic information.
 - :obj:`datasets.logging.DEBUG` (int value, 10): report all information.
 
--- a/docs/source/package_reference/logging_methods.rst
+++ b/docs/source/package_reference/logging_methods.rst
@@ -1,8 +1,38 @@
 Logging methods
 ----------------------------------------------------
 
-`datasets` tries to be very transparent and explicit about it's inner working bt this can be quite verbose at some time.
+`datasets` tries to be very transparent and explicit about its inner working, but this can be quite verbose at some times.
+
 A series of logging methods let you easily adjust the level of logging of the whole library.
+
+Currently the default verbosity of the library is ``WARNING``.
+
+To change the level of verbosity, just use one of the direct setters. For instance, here is how to change the verbosity to the INFO level.
+
+.. code-block:: python
+
+    import datasets
+    datasets.logging.set_verbosity_info()
+
+You can also use the environment variable ``DATASETS_VERBOSITY`` to override the default verbosity. You can set it to one of the following: ``debug``, ``info``, ``warning``, ``error``, ``critical``. For example:
+
+.. code-block:: bash
+               
+    DATASETS_VERBOSITY=error ./myprogram.py
+
+All the methods of this logging module are documented below, the main ones are
+:func:`datasets.logging.get_verbosity` to get the current level of verbosity in the logger and
+:func:`datasets.logging.set_verbosity` to set the verbosity to the level of your choice. In order (from the least
+verbose to the most verbose), those levels (with their corresponding int values in parenthesis) are:
+
+- :obj:`datasets.logging.CRITICAL` or :obj:`datasets.logging.FATAL` (int value, 50): only report the most
+  critical errors.
+- :obj:`datasets.logging.ERROR` (int value, 40): only report errors.
+- :obj:`datasets.logging.WARNING` or :obj:`datasets.logging.WARN` (int value, 30): only reports error and
+  warnings. This the default level used by the library.
+- :obj:`datasets.logging.INFO` (int value, 20): reports error, warnings and basic information.
+- :obj:`datasets.logging.DEBUG` (int value, 10): report all information.
+
 
 Functions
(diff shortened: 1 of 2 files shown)


lhoestq (commented):
Cool thanks :) 

Not sure we want to have the logging set to warning right now though.
Right now there are still some messages related to caching that we may want to show the users by default.
Let's discuss a bit what should be shown by default in warning first ?

IMO those messages should be set to warning:
- load from cache when doing `load_dataset`
- the progress bar in `.map` ?
- the progress bar when writing to cache when doing `load_dataset` ?

Those ones are already warnings iirc:
- load from cache when using a dataset transform (`.map`, etc.)
- write to cache when using a dataset transform (`.map`, etc.)
- write to cache when doing `load_dataset`

thomwolf:
Also, some downloads and dataset processing can be quite long for large datasets like wikipedia/pg19/etc. We probably don't want to user to think that the library is hanging. Happy to reorganize logging between DEBUG/INFO/WARNING to make it less verbose by default though.

stas00:
The problem is that `transformers` imports `datasets` and the latter starts logging on `import`: at least 3 info messages - apache beam/torch/tf available - so it injects noise whether one uses the library or not - i.e. no choice given to the user.

Would you be open for me to changing this PR, to keep the initial level at INFO, but to keep the `DATASETS_VERBOSITY` env var it introduces, to let the user control the verbosity?

stas00:
> Also, some downloads and dataset processing can be quite long for large datasets like wikipedia/pg19/etc. We probably don't want to user to think that the library is hanging.

If you're referring to tqdm, it's not affected by changing the logging levels.

lhoestq:
> The problem is that `transformers` imports `datasets` and the latter starts logging on `import`: at least 3 info messages - apache beam/torch/tf available - so it injects noise whether one uses the library or not - i.e. no choice given to the user.
> 
> Would you be open for me to changing this PR, to keep the initial level at INFO, but to keep the `DATASETS_VERBOSITY` env var it introduces, to let the user control the verbosity?

For now we can do that, then I'll change some messages to warnings and set the default verbosity at warning as well at that point. Does it sound good to you ?

> If you're referring to tqdm progress reports, it's not affected by changing the logging levels. It's not using logging.

Actually we configured some progress bars to be disabled depending on the logging level ^^'

stas00:
> For now we can do that, then I'll change some messages to warnings and set the default verbosity at warning as well at that point. Does it sound good to you ?

If it is logical then by all means. 

> > If you're referring to tqdm progress reports, it's not affected by changing the logging levels. It's not using logging.
> 
> Actually we configured some progress bars to be disabled depending on the logging level ^^'

This is very smart!

I reverted s/WARNINGS/INFO/.

Thank you!

thomwolf:
Note that it’s the same in `transformers` @stas00, tdqm are also controlled by the logging level there.

stas00:
> Note that it’s the same in `transformers` @stas00, tdqm are also controlled by the logging level there.

That's good to know, @thomwolf - thank you!

I see that it's controlled in `trainer.py`, but in `examples` it's not - since that's where I usually see the progressbars (and they are great!). But I suppose they aren't API, so `examples` can behave differently.

stas00:
BTW, this is what I'm talking about:
```
python -c "import transformers"
2020-09-14 21:00:58.032658: I tensorflow/stream_executor/platform/default/dso_loader.cc:48] Successfully opened dynamic library libcudart.so.10.1
PyTorch version 1.7.0.dev20200910 available.
TensorFlow version 2.3.0 available.
Apache Beam available.
```
why does the user need to see this? Especially, if they aren't even using `datasets` directly?

thomwolf:
Yes you are right, we should re-think the logging level of various elements.
I also think that the `set_format` messages are confusing when they are the results of our internal operations (as mentioned [here](https://discuss.huggingface.co/t/pipeline-with-custom-dataset-tokenizer-when-to-save-load-manually/1084/7?u=thomwolf))

lhoestq (approved):
I'm changing the logging messages so that we can change the default level to warning asap

docs/source/package_reference/logging_methods.rst
```diff
@@ -1,8 +1,38 @@
...
+All the methods of this logging module are documented below, the main ones are
+:func:`datasets.logging.get_verbosity` to get the current level of verbosity in the logger and
+:func:`datasets.logging.set_verbosity` to set the verbosity to the level of your choice. In order (from the least
+verbose to the most verbose), those levels (with their corresponding int values in parenthesis) are:
+
+- :obj:`datasets.logging.CRITICAL` or :obj:`datasets.logging.FATAL` (int value, 50): only report the most
+  critical errors.
+- :obj:`datasets.logging.ERROR` (int value, 40): only report errors.
+- :obj:`datasets.logging.WARNING` or :obj:`datasets.logging.WARN` (int value, 30): only reports error and
+  warnings. This the default level used by the library.
```
stefan-it, line 32: but line 8 mentions that the default level is `INFO` 🤔

lhoestq: Right !
Let's change to INFO for now as well in the docs @stas00

lhoestq:
Actually I continued this PR in #635 to set the level to warning and update the logging level of some messages.

Let me know if it sounds good to you

docs/source/package_reference/logging_methods.rst
stas00: Great catch, @stefan-it . I removed that note, since it's already written above it. So having one place is less error-prone.

lhoestq:
Closing this one sice #635 got merged

stas00:
Awesome! Thank you!

Any ideas how to eliminate this remaining log line from tensorflow:
```
python -c "import transformers"
2020-09-17 08:38:34.718410: I tensorflow/stream_executor/platform/default/dso_loader.cc:48] Successfully opened dynamic library libcudart.so.10.1
```

PR #640, opened 2020-09-18
Make shuffle compatible with temp_seed

lhoestq:
This code used to return different dataset at each run
```python
import dataset as ds

dataset = ...

with ds.temp_seed(42):
    shuffled = dataset.shuffle()
```

Now it returns the same one since the seed is set


thomwolf (approved):
Nice!

PR #645, opened 2020-09-18
Don't use take on dataset table in pyarrow 1.0.x

### LINKED ISSUE #615: Offset overflow when slicing a big dataset with an array of indices in Pyarrow >= 1.0.0
How to reproduce:

```python

wiki = load_dataset("wikipedia", "20200501.en", split="train")
wiki[[0]]

ArrowInvalid                              Traceback (most recent call last)
<ipython-input-13-381aedc9811b> in <module>
----> 1 wikipedia[[0]]

~/Desktop/hf/nlp/src/datasets/arrow_dataset.py in __getitem__(self, key)
   1069             format_columns=self._format_columns,
   1070             output_all_columns=self._output_all_columns,
-> 1071             format_kwargs=self._format_kwargs,
   1072         )
   1073 

~/Desktop/hf/nlp/src/datasets/arrow_dataset.py in _getitem(self, key, format_type, format_columns, output_all_columns, format_kwargs)
   1037                 )
   1038             else:
-> 1039                 data_subset = self._data.take(indices_array)
   1040 
   1041             if format_type is not None:

~/.virtualenvs/hf-datasets/lib/python3.7/site-packages/pyarrow/table.pxi in pyarrow.lib.Table.take()

~/.virtualenvs/hf-datasets/lib/python3.7/site-packages/pyarrow/compute.py in take(data, indices, boundscheck)
    266     """
    267     options = TakeOptions(boundscheck)
--> 268     return call_function('take', [data, indices], options)
    269 
    270 

~/.virtualenvs/hf-datasets/lib/python3.7/site-packages/pyarrow/_compute.pyx in pyarrow._compute.call_function()

~/.virtualenvs/hf-datasets/lib/python3.7/site-packages/pyarrow/_compute.pyx in pyarrow._compute.Function.call()

~/.virtualenvs/hf-datasets/lib/python3.7/site-packages/pyarrow/error.pxi in pyarrow.lib.pyarrow_internal_check_status()

~/.virtualenvs/hf-datasets/lib/python3.7/site-packages/pyarrow/error.pxi in pyarrow.lib.check_status()

ArrowInvalid: offset overflow while concatenating arrays
```

It seems to work fine with small datasets or with pyarrow 0.17.1

lhoestq:
Fix #615

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -964,12 +964,12 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
 
             # Get the subset of the table
             if indices_array is not None:
-                if PYARROW_V0:
-                    data_subset = pa.concat_tables(
-                        self._data.slice(indices_array[i].as_py(), 1) for i in range(len(indices_array))
-                    )
-                else:
-                    data_subset = self._data.take(indices_array)
+                # if PYARROW_V0:  # don't use take (see https://issues.apache.org/jira/browse/ARROW-9773)
+                data_subset = pa.concat_tables(
+                    self._data.slice(indices_array[i].as_py(), 1) for i in range(len(indices_array))
+                )
+                # else:
+                #     data_subset = self._data.take(indices_array)
             else:
                 data_subset = self._data.slice(key_indices[0], key_indices[1] - key_indices[0])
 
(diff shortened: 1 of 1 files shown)


thomwolf (approved):
Ok, `filter` seems to be quite strongly hit on the benchmarks. Probably because the default batch size of 1000 is not optimal anymore.

Maybe we can reduce it to a small value like 10 or even just 1? (we are not really using batching anyway in most cases of use for filter, it's just about switching to a variable number of examples)

lhoestq:
I tried lower batch sizes and it didn't accelerate filter (quite the opposite actually).
The slow-down also appears for pyarrow 0.17.1 for some reason, not sure it comes from these changes

lhoestq:
I just checked the benchmarks of other PRs and some of them had 300s (!!) for filter and I don't think it comes from changes in the code (it seems quite random to me)

lhoestq:
Merging this one since it's not the cause of the the slow down

AisingioroHao0:
@lhoestq What might be the reason for the slowdown? When I was training large batchsize, the slowdown was obvious, and my original assumption was to increase write_batch_size

PR #654, opened 2020-09-21
Allow empty inputs in metrics

lhoestq:
There was an arrow error when trying to compute a metric with empty inputs. The error was occurring when reading the arrow file, before calling metric._compute.


thomwolf (approved):
LGTM!

Ezequiel198 (commented):
```suggestion

```

PR #644, opened 2020-09-18
Better windows support

lhoestq:
There are a few differences in the behavior of python and pyarrow on windows.

For example there are restrictions when accessing/deleting files that are open

Fix #590

PR #656, opened 2020-09-21
Use multiprocess from pathos for multiprocessing

lhoestq:
[Multiprocess](https://github.com/uqfoundation/multiprocess) (from the [pathos](https://github.com/uqfoundation/pathos) project) allows to use lambda functions in multiprocessed map.
It was suggested to use it by @kandorm.

We're already using dill which is its only dependency.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -80,6 +80,8 @@ REQUIRED_PKGS = [
     "filelock",
     # for fast hashing
     "xxhash"
+    # for better multiprocessing
+    "pathos"
 ]
 
 BENCHMARKS_REQUIRE = [
--- a/setup.py
+++ b/setup.py
@@ -79,7 +79,7 @@ REQUIRED_PKGS = [
     # filesystem locks e.g. to prevent parallel downloads
     "filelock",
     # for fast hashing
-    "xxhash"
+    "xxhash",
     # for better multiprocessing
     "pathos"
 ]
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -79,7 +79,9 @@
     # filesystem locks e.g. to prevent parallel downloads
     "filelock",
     # for fast hashing
-    "xxhash"
+    "xxhash",
+    # for better multiprocessing
+    "pathos"
```
thomwolf, line 84: This is install the full pathos package and then we only use the multiprocessing part which override the default multiprocessing package, did I understand it well?

lhoestq:
We can just install multiprocess actually, I'll change that

BramVanroy:
Just an FYI: I remember that I wanted to try pathos a couple of years back and I ran into issues considering cross-platform; the code would just break on Windows. If I can verify this PR by running CPU tests on Windows, let me know!

lhoestq:
That's good to know thanks
I guess we can just wait for #644 to be merged first. I'm working on fixing the tests for windows

lhoestq:
Looks like all the CI jobs on windows passed !
I also tested locally on my windows and it works great :) 

I think this is ready to merge, let me know if you have any remarks @thomwolf @BramVanroy

thomwolf (approved):
Looks to me!

Issue #132, opened 2020-05-15
[Feature Request] Add the OpenWebText dataset


LysandreJik:
The OpenWebText dataset is an open clone of OpenAI's WebText dataset. It can be used to train ELECTRA as is specified in the [README](https://www.github.com/google-research/electra).

More information and the download link are available [here](https://skylion007.github.io/OpenWebTextCorpus/).

vanyacohen:
We're experimenting with hosting the OpenWebText corpus on Zenodo for easier downloading. https://zenodo.org/record/3834942#.Xs1w8i-z2J8

lhoestq:
Closing since it's been added in #660

PR #660, opened 2020-09-22
add openwebtext

richarddwang:
This adds [The OpenWebText Corpus](https://skylion007.github.io/OpenWebTextCorpus/), which is a clean and large text corpus for nlp pretraining. It is an open source effort to reproduce OpenAI’s WebText dataset used by GPT-2, and it is also needed to reproduce ELECTRA.

It solves #132 .

### Besides dataset building script, I made some changes to the library.

1. Extract large amount of compressed files with multi processing
I add a `num_proc` argument to `DownloadManager.extract` and pass this `num_proc` to `map_nested`. So I can decompress 20 thousands compressed files faster. `num_proc` I add is default to `None`, so it shouldn't break any other thing.

2. In `cached_path`, I change the order to deal with different kind of compressed files (zip, tar, gzip)
Because there is no way to 100% detect a file is a zip file (see [this](https://stackoverflow.com/questions/18194688/how-can-i-determine-if-a-file-is-a-zip-file)), I found it wrongly detect `'./datasets/downloads/extracted/58764bd6898fa339b25d92e7fbbc3d8dbf64fb504edff1a30a1d7d99d1561027/openwebtext/urlsf_subset13-630_data.xz'` as a zip and try decompress it with zip, sure it will get error. So I made it detect wheter the file is tar or gzip first and detect zip in the last.

3. `MockDownloadManager.extract`
Cuz I pass `num_proc` to `DownloadManager.extract`, I also have to make `MockDownloadManager.extract` to accept extra keywork arguments. So I make it `extract(path, *args, **kwargs)`, but just return the path as original implementation.

**Note**: If there is better way for points mentioned above, thought I would like to help, unless we can solve point4 (make dataset building fast), I may not be able to afford rebuild the dataset again because of change of the dataset script (Building the dataset cost me 4 days). 

### There is something I think we can improve

4. Long time to decompress compressed files
Even I decompress those 20 thousands compressed files with 12 process on my 16 core 3.x Ghz server. It still took about 3 ~ 4days to complete dataset building. Most of time spent on decompress those files.

### Info about the source data
The source data is an  tar.xz file with following structure, files/directory beyond compressed file is what can we get after decompress it.
```
openwebtext.tar.xz
  |__ openwebtext
         |__subset000.xz
         |     |__ ....txt
         |     |__ ....txt
         |     ...
         |__ subset001.xz
         |
         ....
```
And this the structure of dummy data, same as the original one.
```
dummy_data.zip
  |__ dummy_data
         |__ openwebtext
              |__fake_subset-1_data-dirxz # actually it is a directory
              |     |__ ....txt
              |     |__ ....txt
              |__ fake_subset-2_data-dirxz
                    |__ ....txt
                    |__ ....txt
```


datasets/openwebtext/openwebtext.py
```diff
@@ -0,0 +1,86 @@
...
+            features=datasets.Features({"text": datasets.Value("string")}),
+            homepage="https://skylion007.github.io/OpenWebTextCorpus/",
+            citation=_CITATION,
+        )
+
+    def _split_generators(self, dl_manager):
+        dl_dir = dl_manager.download_and_extract(_URL)
+        owt_dir = os.path.join(dl_dir, 'openwebtext')
+        subset_xzs = [
+            os.path.join(owt_dir, file_name) for file_name in os.listdir(owt_dir) if file_name.endswith('xz') # filter out ...xz.lock
```
lhoestq, line 67: I'm not sure that `os.listdir` always returns the files in the same order depending on the filesystem.
Maybe we should consider sorting the file_names ?

datasets/openwebtext/openwebtext.py
```diff
@@ -0,0 +1,86 @@
...
+        ]
+        ex_dirs = dl_manager.extract(subset_xzs, num_proc=round(os.cpu_count()*0.75))
+        nested_txt_files = [ 
+          [ 
+            os.path.join(ex_dir,txt_file_name) for txt_file_name in os.listdir(ex_dir) if txt_file_name.endswith('txt')
```
lhoestq, line 72: same here

src/datasets/utils/download_manager.py
```diff
@@ -213,7 +213,7 @@ def iter_archive(self, path):
                 with open(absolute_file_path, "rb") as file_obj:
                     yield (relative_file_path, file_obj)
 
-    def extract(self, path_or_paths):
+    def extract(self, path_or_paths, num_proc=None):
```
lhoestq, line 216: could you add it to the docstring ?

lhoestq (commented):
Awesome work !

I left a comment about os.listdir, let me know what you think.

datasets/openwebtext/openwebtext.py
richarddwang: I didn't be aware of whether the order of file names returned by `os.listdir`. I thought as long as it collect all files, it is ok. Is there anything we should care about the order of file names ?

src/datasets/utils/download_manager.py
richarddwang: Sure !

datasets/openwebtext/openwebtext.py
lhoestq: Ideally the dataset should be ordered the same way for everyone.
Since there's a loop on the files to generate examples we need the files to always have the same order

richarddwang: Yeah you're right ! It should be sorted.

richarddwang:
BTW, is there a one-line command to make our building scripts pass flake8 test? (included code quality test), I got like trailing space or mixed space and tab warning and error, and fixed them manually.

lhoestq:
> BTW, is there a one-line command to make our building scripts pass flake8 test? (included code quality test), I got like trailing space or mixed space and tab warning and error, and fixed them manually.

I don't think so.
We have a command for black and isort but not flake8 as far as I know.

lhoestq (approved):
Thank you so much for adding this one :) 
Good job !

richarddwang:
Thanks for your awesome work too.
Also a little reminder, this solves #132

PR #662, opened 2020-09-22
Created dataset card snli.md

mcmillanmajora:
First draft of a dataset card using the SNLI corpus as an example


mcmillanmajora:
Resubmitting on a new fork

PR #677, opened 2020-09-28
Move cache dir root creation in builder's init

### LINKED ISSUE #671: [BUG] No such file or directory
This happens when both
1. Huggingface datasets cache dir does not exist
2. Try to load a local dataset script

builder.py throws an error when trying to create a filelock in a directory (cache/datasets) that does not exist
https://github.com/huggingface/datasets/blob/master/src/datasets/builder.py#L177

Tested on v1.0.2

@lhoestq

lhoestq:
We use lock files in the builder initialization but sometimes the cache directory where they're supposed to be was not created. To fix that I moved the builder's cache dir root creation in the builder's init.

Fix #671


thomwolf (approved):
Good catch!

PR #680, opened 2020-09-28
Fix bug related to boolean in GAP dataset.

otakumesi:
### Why I did
The value in `row["A-coref"]` and `row["B-coref"]` is `'TRUE'` or `'FALSE'`.
This type is `string`, then `bool('FALSE')` is equal to `True` in Python.
So, both rows are transformed into `True` now.

So, I modified this problem.

### What I did
I modified `bool(row["A-coref"])` and `bool(row["B-coref"])` to `row["A-coref"] == "TRUE"` and `row["B-coref"] == "TRUE"`.

Thank you!


lhoestq:
Hi !

Good catch, thanks for creating this PR :)

Could you also regenerate the metadata for this dataset using 
```
datasets-cli test ./datasets/gap --save_infos --all_configs
```

That'b be awesome

otakumesi:
@lhoestq Thank you review!!!

I've performed it and have read CONTRIBUTING.md now!

lhoestq (approved):
Awesome thank you :)

PR #681, opened 2020-09-28
Adding missing @property (+2 small flake8 fixes).

### LINKED ISSUE #678: The download instructions for c4 datasets are not contained in the error message
The manual download instructions are not clear 
```The dataset c4 with config en requires manual data. 
 Please follow the manual download instructions: <bound method C4.manual_download_instructions of <datasets_modules.datasets.c4.830b0c218bd41fed439812c8dd19dbd4767d2a3faa385eb695cf8666c982b1b3.c4.C4 object at 0x7ff8c5969760>>. 
 Manual data can be loaded with `datasets.load_dataset(c4, data_dir='<path/to/manual/data>')
```

Either `@property`  could be added to C4.manual_download_instrcutions (or make it a real property), or the manual_download_instructions function needs to be called I think.

Let me know if you want a PR for this, but I'm not sure which possible fix is the correct one.

Narsil:
Fixes #678

### DIFF
--- a/datasets/c4/c4.py
+++ b/datasets/c4/c4.py
@@ -145,6 +145,7 @@ class C4(datasets.BeamBasedBuilder):
         ),
     ]
 
+    @property
     def manual_download_instructions(self):
         return """\
     For the WebText-like config, you must manually download 'OpenWebText.zip'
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
Thanks :)

PR #688, opened 2020-09-30
Disable tokenizers parallelism in multiprocessed map

lhoestq:
It was reported in #620 that using multiprocessing with a tokenizers shows this message:
```
The current process just got forked. Disabling parallelism to avoid deadlocks...
To disable this warning, please explicitly set TOKENIZERS_PARALLELISM=(true | false)
```
This message is shown when TOKENIZERS_PARALLELISM is unset.
Moreover if it is set to `true`, then the program just hangs.

To hide the message (if TOKENIZERS_PARALLELISM is unset) and avoid hanging (if TOKENIZERS_PARALLELISM is `true`), then I set TOKENIZERS_PARALLELISM to `false` when forking the process. After forking is gets back to its original value.

Also I added a warning if TOKENIZERS_PARALLELISM was `true` and is set to `false`:
```
Setting TOKENIZERS_PARALLELISM=false for forked processes.
```

cc @n1t0

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -1265,7 +1265,17 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
                 return cache_file_name
 
             prev_env = dict(os.environ)
-            if prev_env.get("TOKENIZERS_PARALLELISM", "false") != "false":
+            # check if parallelism if off
+            # from https://github.com/huggingface/tokenizers/blob/bb668bc439dc34389b71dbb8ce0c597f15707b53/tokenizers/src/utils/parallelism.rs#L22
+            if prev_env.get("TOKENIZERS_PARALLELISM", "false").lower() not in (
+                "",
+                "off",
+                "false",
+                "f",
+                "no",
+                "n",
+                "0",
+            ):
                 logger.warning("Setting TOKENIZERS_PARALLELISM=false for forked processes.")
             os.environ["TOKENIZERS_PARALLELISM"] = "false"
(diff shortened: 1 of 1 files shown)


src/datasets/arrow_dataset.py
```diff
@@ -1264,7 +1264,12 @@ def format_cache_file_name(cache_file_name, rank):
                 logger.info("Process #{} will write at {}".format(rank, cache_file_name))
                 return cache_file_name
 
+            prev_env = dict(os.environ)
+            if prev_env.get("TOKENIZERS_PARALLELISM", "false") != "false":
```
thomwolf, line 1268: Are you sure that checking `"false"` is enough?
In `transformers` we check various way to write env variable values: https://github.com/huggingface/transformers/blob/master/src/transformers/file_utils.py#L52-L55

lhoestq: Well `false` is different from anything in `("1", "ON", "YES", "AUTO")` :)
But for consistency I'll use the same thing instead of checking for `false`

lhoestq: Not sure about "AUTO" though

thomwolf: maybe you can convert to `lower` before checking for `"false"` and check for `0` as well?

lhoestq: I checked `tokenizers` source code to check for every way of saying "no" to parallelism.
It should be good now

n1t0: I was going to send you our list from `tokenizers`, you were faster!

thomwolf (approved):

Issue #617, opened 2020-09-11
Compare different Rouge implementations  


ibeltagy:
I used RougeL implementation provided in `datasets` [here](https://github.com/huggingface/datasets/blob/master/metrics/rouge/rouge.py) and it gives numbers that match those reported in the pegasus paper but very different from those reported in other papers, [this](https://arxiv.org/pdf/1909.03186.pdf) for example.
Can you make sure the google-research implementation you are using matches the official perl implementation? 
There are a couple of python wrappers around the perl implementation, [this](https://pypi.org/project/pyrouge/) has been commonly used, and [this](https://github.com/pltrdy/files2rouge) is used in fairseq). 
There's also a python reimplementation [here](https://github.com/pltrdy/rouge) but its RougeL numbers are way off.

ibeltagy:
Updates - the differences between the following three
(1) https://github.com/bheinzerling/pyrouge (previously popular. The one I trust the most)
(2) https://github.com/google-research/google-research/tree/master/rouge
(3) https://github.com/pltrdy/files2rouge (used in fairseq)
can be explained by two things, stemming and handling multiple sentences.

Stemming: 
(1), (2): default is no stemming. (3): default is with stemming ==> No stemming is the correct default as you did [here](https://github.com/huggingface/datasets/blob/master/metrics/rouge/rouge.py#L84)

Multiple sentences:
(1) `rougeL` splits text using `\n`
(2) `rougeL` ignores `\n`. 
(2) `rougeLsum` splits text using `\n`
(3) `rougeL` splits text using `.`

For (2), `rougeL` and `rougeLsum` are identical if the sequence doesn't contain `\n`. With `\n`, it is `rougeLsum` that matches (1) not `rougeL`. 

Overall, and as far as I understand, for your implementation here https://github.com/huggingface/datasets/blob/master/metrics/rouge/rouge.py#L65 to match the default, you only need to change `rougeL` [here](https://github.com/huggingface/datasets/blob/master/metrics/rouge/rouge.py#L86) to `rougeLsum` to correctly compute metrics for text with newlines.

lhoestq:
Thanks for the clarification !
We're adding Rouge Lsum in #701

sshleifer:
This is a real issue, sorry for missing the mention @ibeltagy

[...]

PR #700, opened 2020-10-02
Add rouge-2 in rouge_types for metric calculation

Shashi456:
The description of the ROUGE metric says, 
```
_KWARGS_DESCRIPTION = """
Calculates average rouge scores for a list of hypotheses and references
Args:
    predictions: list of predictions to score. Each predictions
        should be a string with tokens separated by spaces.
    references: list of reference for each prediction. Each
        reference should be a string with tokens separated by spaces.
Returns:
    rouge1: rouge_1 f1,
    rouge2: rouge_2 f1,
    rougeL: rouge_l f1,
    rougeLsum: rouge_l precision
"""
```

but the `rouge_types` argument defaults to  `rouge_types = ["rouge1", "rougeL"]`, this PR updates and add `rouge2` to the list so as to reflect the description card.

### DIFF
--- a/metrics/rouge/rouge.py
+++ b/metrics/rouge/rouge.py
@@ -83,7 +83,7 @@ class Rouge(datasets.Metric):
 
     def _compute(self, predictions, references, rouge_types=None, use_agregator=True, use_stemmer=False):
         if rouge_types is None:
-            rouge_types = ["rouge1", "rougeL"]
+            rouge_types = ["rouge1", "rouge2", "rougeL"]
 
         scorer = rouge_scorer.RougeScorer(rouge_types=rouge_types, use_stemmer=use_stemmer)
         if use_agregator:


lhoestq:
Indeed there's currently a mismatch between the description and what it rouge actually returns.
Thanks for proposing this fix :) 

I think it's better to return rouge 1-2-L.
Was there a reason to only include rouge 1 and rouge L @thomwolf ?

lhoestq:
rougeLsum is also missing, could you add it ?

thomwolf:
Adding `RougeLSum` would fix https://github.com/huggingface/datasets/issues/617

lhoestq:
I am opening a PR with both of them right now actually :)

Shashi456:
Also the format of the output isn't exactly ideal

Shashi456:
I'll close this since you seem to have already added it in another PR. Sorry for the delay in responding to you @lhoestq.

thomwolf:
What do you mean by "Formatting the output to reflect how ROUGE-1-5-5" @Shashi456 ?

lhoestq:
I like the idea of returning all the scores for two reason:
- Rouge's aggregator does sampling and therefore it returns "low" "mid" and "high" scores
- It is interesting to have the precision and recall to see how the F1 score was computed
But I understand your point that returning only the F1 score makes sense since it's the one that's always used

Shashi456:
@thomwolf the scores now returned look like this:
```
{'rouge1': AggregateScore(low=Score(precision=0.16620308156871524, recall=0.18219819615984395, fmeasure=0.16226017699359463), mid=Score(precision=0.17274338501705871, recall=0.1890957812369246, fmeasure=0.16823877588620403), high=Score(precision=0.17934569582981455, recall=0.1965626706042028, fmeasure=0.17491509794856058)), 
'rouge2': AggregateScore(low=Score(precision=0.12478835737689957, recall=0.1362113231755514, fmeasure=0.12055941950062395), mid=Score(precision=0.1303967602691664, recall=0.1423747229852964, fmeasure=0.1258363976151122), high=Score(precision=0.13654527560789362, recall=0.1488071465116122, fmeasure=0.13184989406704056)), 
'rougeL': AggregateScore(low=Score(precision=0.16568068818352072, recall=0.1811919016674486, fmeasure=0.1614784523482225), mid=Score(precision=0.17156684723552357, recall=0.1879777628247058, fmeasure=0.16720699286250762), high=Score(precision=0.17788847350584547, recall=0.1948899838530898, fmeasure=0.17316501523379826))}
```

While when computed through the rouge script, the output looks like the following:
```
[2020-07-30 18:13:38,556 INFO] Rouges at step 13000                                       
>> ROUGE-F(1/2/3/l): 43.43/20.42/39.78              
ROUGE-R(1/2/3/l): 53.91/25.34/49.32
```

The formatting allows for easy reading, and although "low", "mid", "high" make sense, this is more concise and effective.

Shashi456:
cc: @lhoestq

lhoestq:
Ok I see.
I think it's also important to follow one of the existing output format (there are already too many different formats, let's try not to add another different one)
I'd still stick with the current format and not transform the output of the python implementation of rouge since it's already widely used.
What do you think ?

thomwolf:
Maybe we could convert the dataclasses in dictionnaries, would that help @Shashi456 ?

Shashi456:
@thomwolf yeah I think that would help. I initially didn't understand the high low mid categories. Dataclasses could help in this case I guess.

PR #701, opened 2020-10-02
Add rouge 2 and rouge Lsum to rouge metric outputs

### LINKED ISSUE #617: Compare different Rouge implementations

lhoestq:
Continuation of #700 

Rouge 2 and Rouge Lsum were missing in Rouge's outputs.
Rouge Lsum is also useful to evaluate Rouge L for sentences with `\n`

Fix #617

### DIFF
--- a/metrics/rouge/rouge.py
+++ b/metrics/rouge/rouge.py
@@ -55,10 +55,10 @@ Args:
     references: list of reference for each prediction. Each
         reference should be a string with tokens separated by spaces.
 Returns:
-    rouge1: rouge_1 f1,
-    rouge2: rouge_2 f1,
-    rougeL: rouge_l f1,
-    rougeLsum: rouge_l precision
+    rouge1: rouge_1 (precision, recall, f1),
+    rouge2: rouge_2 (precision, recall, f1),
+    rougeL: rouge_l (precision, recall, f1),
+    rougeLsum: rouge_l (precision, recall, f1)
 """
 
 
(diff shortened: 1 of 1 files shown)


metrics/rouge/rouge.py
```diff
@@ -55,10 +55,10 @@
     references: list of reference for each prediction. Each
         reference should be a string with tokens separated by spaces.
 Returns:
-    rouge1: rouge_1 f1,
-    rouge2: rouge_2 f1,
-    rougeL: rouge_l f1,
-    rougeLsum: rouge_l precision
+    rouge1: rouge_1 (precision, recall, f1),
```
thomwolf, line 58: Do you mind adding this to the Args doc just above?
```
      rouge_types: A list of rouge types to calculate. Valide names: `"rouge{n}"` (e.g. `"rouge1"`, `"rouge2"`) where {n} is the n-gram based scoring, `"rougeL"`: Longest common subsequence based scoring. `"rougeLSum"`: rougeLsum splits text using `"\n"`. See details in https://github.com/huggingface/datasets/issues/617
      use_stemmer: Bool indicating whether Porter stemmer should be used to strip word suffixes.
      use_agregator: Return aggregates if this is set to True
```

thomwolf:
Oups too late, sorry

metrics/rouge/rouge.py
lhoestq: Yes indeed

PR #715, opened 2020-10-05
Use python read for text dataset

lhoestq:
As mentioned in #622 the pandas reader used for text dataset doesn't work properly when there are \r characters in the text file.

Instead I switched to pure python using `open` and `read`.
From my benchmark on a 100MB text file, it's the same speed as the previous pandas reader.


thomwolf (approved):
Ok :)

datasets/text/text.py
```diff
@@ -55,19 +55,17 @@ def _split_generators(self, dl_manager):
...
-                encoding=self.config.encoding,
-                sep="\n",
-            )
-            for j, df in enumerate(text_file_reader):
-                pa_table = pa.Table.from_pandas(df)
-                # Uncomment for debugging (will print the Arrow table size and elements)
-                # logger.warning(f"pa_table: {pa_table} num rows: {pa_table.num_rows}")
-                # logger.warning('\n'.join(str(pa_table.slice(i, 1).to_pydict()) for i in range(pa_table.num_rows)))
-                yield (i, j), pa_table
+            j = 0
```
BramVanroy, line 57: Not trying to be an *ss, but perhaps you could use more meaningful vars here? file_idx, batch_idx, something like that. So that others can read through the code more easily. Looking great though!

lhoestq: Sure !

thomwolf: Yes Bram is right, in particular since this script is expected to be used and read a lot!

Ezequiel198 (approved):
Recuperación de datos y privacidad

thomwolf:
One thing though, could we try to read the files in parallel?

datasets/text/text.py
Ezequiel198: > Not trying to be an *ss, but perhaps you could use more meaningful vars here? file_idx, batch_idx, something like that. So that others can read through the code more easily. Looking great though!

lhoestq:
We could but I'm not sure this would help a lot since the bottleneck is the drive IO if the files are big enough.
It could make sense for very small files.

lhoestq:
Looks like windows is not a big fan of this approach
I'm working on a fix

datasets/text/text.py
lhoestq: I changed to file_idx and batch_idx

thomwolf:
I remember issue https://github.com/huggingface/datasets/issues/546 where this was kinda requested (but maybe IO would bottleneck). What do you think?

lhoestq:
I think it's worth testing multiprocessing. It could also be something we add to our speed benchmarks

BramVanroy:
> I remember issue #546 where this was kinda requested (but maybe IO would bottleneck). What do you think?

It still would be interesting I think, especially in scenarios where IO is less of an issue (SSDs particularly) and where there are many smaller files. Wrapping this function in a `pool.map` is perhaps an easy thing to try.

lhoestq:
Merging this one for now for the patch release

Issue #302, opened 2020-06-23
Question - Sign Language Datasets


AmitMY:
An emerging field in NLP is SLP - sign language processing.

I was wondering about adding datasets here, specifically because it's shaping up to be large and easily usable.
The metrics for sign language to text translation are the same.

So, what do you think about (me, or others) adding datasets here?

An example dataset would be [RWTH-PHOENIX-Weather 2014 T](https://www-i6.informatik.rwth-aachen.de/~koller/RWTH-PHOENIX-2014-T/)
For every item in the dataset, the data object includes:
1. video_path - path to mp4 file
2. pose_path - a path to `.pose` file with human pose landmarks
3. openpose_path - a path to a `.json` file with human pose landmarks
4. gloss - string
5. text - string
6. video_metadata - height, width, frames, framerate

------

To make it a tad more complicated - what if sign language libraries add requirements to `nlp`? for example, sign language is commonly annotated using `ilex`, `eaf`, or `srt` files, which are all loadable as text, but there is no reason for the dataset to parse that file by itself, if libraries exist to do so.

AmitMY:
Even more complicating - 

As I see it, datasets can have "addons".
For example, the WebNLG dataset is a dataset for data-to-text. However, a work of mine and other works enriched this dataset with text plans / underlying text structures. In that case, I see a need to load the dataset "WebNLG" with "plans" addon.

Same for sign language - if there is a dataset of videos, one addon can be to run OpenPose, another to run ARKit4 pose estimation, and another to run PoseNet, or even just a video embedding addon. (which are expensive to run individually for everyone who wants to use these data)

This is something I dabbled with my own implementation to a [research datasets library](https://github.com/AmitMY/meta-scholar/) and I love to get the discussion going on these topics.

lhoestq:
This is a really cool idea !
The example for data objects you gave for the RWTH-PHOENIX-Weather 2014 T dataset can totally fit inside the library.

For your point about formats like `ilex`, `eaf`, or `srt`, it is possible to use any library in your dataset script.
However most user probably won't need these libraries, as most datasets don't need them, and therefore it's unlikely that we will have them in the minimum requirements to use `nlp` (we want to keep it as light-weight as possible). If a user wants to load your dataset and doesn't have the libraries you need, an error is raised asking the user to install them.

More generally, we plan to have something like a `requirements.txt` per dataset. This could also be a place for addons as you said. What do you think ?

AmitMY:
[...]

PR #722, opened 2020-10-10
datasets(RWTH-PHOENIX-Weather 2014 T): add initial loading script

AmitMY:
This is the first sign language dataset in this repo as far as I know.
Following an old issue I opened https://github.com/huggingface/datasets/issues/302.

I added the dataset official REAMDE file, but I see it's not very standard, so it can be removed.


julien-c:
This might be interesting to @kayoyin the author of https://github.com/kayoyin/transformer-slt – pinging you just in case :)

datasets/rwth_phoenix_weather_2014_t/dataset_infos.json
```diff
@@ -0,0 +1,81 @@
...
+      },
+      "validation": {
+        "name": "validation",
+        "num_bytes": 502,
+        "num_examples": 1,
+        "dataset_name": "rwth_phoenix_weather2014_t"
+      }
+    },
+    "download_checksums": {
+      "/home/nlp/amit/forks/datasets/datasets/rwth_phoenix_weather_2014_t/dummy/1.0.0/dummy_data.zip": {
```
lhoestq, line 71: Could you regenerate this file to update the url ?
```
datasets-cli test ./datasets/rwth_phoenix_weather_2014_t --all_configs --save_infos --ignore_verifications
```

src/datasets/utils/file_utils.py
```diff
@@ -359,6 +361,30 @@ def cached_path(
...
+        initial=resume_size,
+        desc="Downloading",
+        disable=not_verbose,
+    )
+
+    def progress_hook(block_num, block_size, total_size):
+        progress.update(block_size)
+
+    urlretrieve(url, temp_file.name, progress_hook)
+
```
lhoestq, line 386:
```suggestion
    progress.close()
```

src/datasets/utils/file_utils.py
```diff
@@ -472,7 +498,8 @@ def get_from_cache(
             )
         elif response is not None and response.status_code == 404:
             raise FileNotFoundError("Couldn't find file at {}".format(url))
-        raise ConnectionError("Couldn't reach {}".format(url))
+        elif not url.startswith("ftp"):
+            raise ConnectionError("Couldn't reach {}".format(url))
```
lhoestq, line 502: At this point it would be nice to throw a `ConnectionError` if ftp login fails, and a `FileNotFoundError` if the file doesn't exist (if the ftp.size() fails for example).

Only those two kinds of errors are handled to provide indications to the user when it crashes.

src/datasets/utils/file_utils.py
```diff
@@ -509,7 +536,11 @@ def _resumable_file_manager():
             logger.info("%s not found in cache or force_download set to True, downloading to %s", url, temp_file.name)
 
             # GET file object
-            http_get(url, temp_file, proxies=proxies, resume_size=resume_size, user_agent=user_agent, cookies=cookies)
+            if url.startswith("ftp"):
+                ftp_get(url, temp_file)
+            else:
+                http_get(url, temp_file, proxies=proxies, resume_size=resume_size, user_agent=user_agent,
+                         cookies=cookies)
```
lhoestq, line 543: Nice !

datasets/rwth_phoenix_weather_2014_t/rwth_phoenix_weather_2014_t.py
```diff
@@ -0,0 +1,112 @@
...
+        """ Yields examples. """
+
+        filepath = os.path.join(base_path, "annotations", "manual", "PHOENIX-2014-T." + split + ".corpus.csv")
+        images_path = os.path.join(base_path, "features", "fullFrame-210x260px", split)
+
+        with open(filepath) as f:
+            data = csv.DictReader(f, delimiter="|", quoting=csv.QUOTE_NONE)
+            for row in data:
+                yield row["name"], {
+                    "video": os.path.join(images_path, row["video"]),
```
lhoestq, line 106: Do you think it could be worth adding the images directly in the samples instead of having the path ?

lhoestq (commented):
Very nice !
Thanks for adding the support for FTP downloads.
The dataset looks cool as well.

I left a few comments.
My main concern is that we have to raise the appropriate errors if the ftp download fails, let me know if it makes sense to you.

datasets/rwth_phoenix_weather_2014_t/rwth_phoenix_weather_2014_t.py
AmitMY: That might be a good idea.
1. Could you point me to a dataset in this repo that you think does this well?
2. If the dataset contains images, would it still be able to load fast for people who only want to access `gloss` and `translation`, or would the images slow everything down drastically?

lhoestq: We added support for images recently for the LXMERT dataset but it hasn't been added yet. Which means that this. dataset could be the first one with images :)

I've made images support compatible with memory mapping so you can load them very quickly.
Moreover since Apache Arrow has a NumPy integration it's even faster to load them if you format the dataset for numpy. It also makes loading as pytorch or tf tensors faster.

If your images are multi-channel high res and if you have several of them per samples, it will slow down accessing the other columns though. I haven't tested to have a dataset with a sequence of images per sample.

Could you tell me more about the images resolutions / nb of channels / nb of images per sample ?

thomwolf: Pinging @sgugger here who know a lot about the tradeoffs here

AmitMY: The images are `210x260`, `RGB` channels, with up to (maximum) 400 images per sample.
It would be equivalent in space to a `2184x10000` high res image.

I guess I'll give that a go unless @sgugger knows why I shouldn't

sgugger: Images are usually stored as encoded jpegs (or other format) to avoid the dataset being too large. I don't know how many images this dataset contains but keep in mind storing them in raw format could easily blow up disk storage. But if it fits and is not a problem for the size of the download, it would certainly speed up the decoding part.
 
Of course that also means you should avoid loading everything in RAM but I think the datasets library is already doing this?

AmitMY: I couldn't find a feature that is specifically for images, in order to store it in a compressed format encoding (jpeg etc..)
Right now, I'm thinking on doing `Sequence(Array3D(shape=...))`
@lhoestq does this makes sense or is there a specific type for it?

lhoestq: > Of course that also means you should avoid loading everything in RAM but I think the datasets library is already doing this?

Yes indeed, since the dataset is memory mapped.



> I couldn't find a feature that is specifically for images, in order to store it in a compressed format encoding (jpeg etc..)
> Right now, I'm thinking on doing `datasets.features.Sequence(datasets.features.Array3D(shape=(210, 260, 3), dtype="uint8"))`
> @lhoestq does this makes sense or is there a specific type for it?

That's the right feature type for your case.

src/datasets/utils/file_utils.py
AmitMY: added resume option, added ftps suppot

src/datasets/utils/file_utils.py
AmitMY: added these errors

datasets/rwth_phoenix_weather_2014_t/rwth_phoenix_weather_2014_t.py
AmitMY: Thanks @lhoestq 

I now added this type, as well as am loading the images to numpy array in the correct shape, but this fails for some reason and I can't figure out why.
I am setting `video` as a list (sequential) of 3D numpy arrays (shape `260, 210, 3`, type `uint8`)

Mind taking a look? You can change the dataset path to the `dummy_data.zip` file, which includes 1 item, and would be very fast to load and process.

```
  File "/home/nlp/amit/forks/datasets/src/datasets/builder.py", line 434, in incomplete_dir
    yield tmp_dir
  File "/home/nlp/amit/forks/datasets/src/datasets/builder.py", line 476, in download_and_prepare
    dl_manager=dl_manager, verify_infos=verify_infos, **download_and_prepare_kwargs
  File "/home/nlp/amit/forks/datasets/src/datasets/builder.py", line 553, in _download_and_prepare
  File "/home/nlp/amit/forks/datasets/src/datasets/builder.py", line 846, in _prepare_split
  File "/home/nlp/amit/forks/datasets/src/datasets/arrow_writer.py", line 299, in finalize
    self.write_on_file()
  File "/home/nlp/amit/forks/datasets/src/datasets/arrow_writer.py", line 232, in write_on_file
  File "pyarrow/array.pxi", line 196, in pyarrow.lib.array
  File "pyarrow/array.pxi", line 107, in pyarrow.lib._handle_arrow_array_protocol
  File "/home/nlp/amit/forks/datasets/src/datasets/arrow_writer.py", line 99, in __arrow_array__
    out = pa.array(self.data, type=type)
  File "pyarrow/array.pxi", line 269, in pyarrow.lib.array
  File "pyarrow/array.pxi", line 38, in pyarrow.lib._sequence_to_array
  File "pyarrow/error.pxi", line 105, in pyarrow.lib.check_status
pyarrow.lib.ArrowNotImplementedError: MakeBuilder: cannot construct builder for type extension<arrow.py_extension_type>

During handling of the above exception, another exception occurred:

  File "/home/nlp/amit/forks/datasets/src/load_test.py", line 6, in <module>
    '/home/nlp/amit/forks/datasets/datasets/rwth_phoenix_weather_2014_t/rwth_phoenix_weather_2014_t.py')
  File "/home/nlp/amit/forks/datasets/src/datasets/load.py", line 611, in load_dataset
    ignore_verifications=ignore_verifications,
  File "/home/nlp/amit/forks/datasets/src/datasets/builder.py", line 483, in download_and_prepare
    self._save_info()
  File "/home/nlp/amit/anaconda2/envs/meta-scholar/lib/python3.7/contextlib.py", line 130, in __exit__
    self.gen.throw(type, value, traceback)
  File "/home/nlp/amit/forks/datasets/src/datasets/builder.py", line 440, in incomplete_dir
    shutil.rmtree(tmp_dir)
  File "/home/nlp/amit/anaconda2/envs/meta-scholar/lib/python3.7/shutil.py", line 498, in rmtree
    onerror(os.rmdir, path, sys.exc_info())
  File "/home/nlp/amit/anaconda2/envs/meta-scholar/lib/python3.7/shutil.py", line 496, in rmtree
    os.rmdir(path)
OSError: [Errno 39] Directory not empty: '/home/nlp/amit/.cache/huggingface/datasets/rwth_phoenix_weather2014_t/default/1.0.0/f94e5a4ba5905587046b092078275d655495cb484f281c43d3a53f85ac366743.incomplete'
```

datasets/rwth_phoenix_weather_2014_t/rwth_phoenix_weather_2014_t.py
```diff
@@ -0,0 +1,118 @@
...
+                frames_path = os.path.join(images_path, row["video"])[:-7]
+                np_frames = []
+                for frame_name in os.listdir(frames_path):
+                    frame_path = os.path.join(frames_path, frame_name)
+                    im = Image.open(frame_path)
+                    np_frames.append(np.asarray(im))
+                    im.close()
+
+                yield row["name"], {
+                    "video": np_frames,
```
AmitMY, line 114: Now that video is just a sequence of np arrays, loading is killed after 170 examples exactly.
I checked, and I can indeed load more than 170 examples at a time, so I suspect something to do with how images are saved in arrow? Not sure. @lhoestq sorry to bother you with this, but do you have any idea?

kayoyin:
Thanks Amit, this is a great idea! I'm thinking of porting the SLT models from my paper here as well, having this dataset would be perfect for that :)

albertvillanova:
Thanks for your contribution, @AmitMY. Are you still interested in adding this dataset?

PR #725, opened 2020-10-12
pretty print dataset objects

stas00:
Currently, if I do:
```
load_dataset("wikihow", 'all', data_dir="/hf/pegasus-datasets/wikihow/")
```
I get:
```

DatasetDict({'train': Dataset(features: {'text': Value(dtype='string', id=None),
'headline': Value(dtype='string', id=None), 'title': Value(dtype='string',
id=None)}, num_rows: 157252), 'validation': Dataset(features: {'text':
Value(dtype='string', id=None), 'headline': Value(dtype='string', id=None),
'title': Value(dtype='string', id=None)}, num_rows: 5599), 'test':
Dataset(features: {'text': Value(dtype='string', id=None), 'headline':
Value(dtype='string', id=None), 'title': Value(dtype='string', id=None)},
num_rows: 5577)})
```

This is not very readable. 

Can we either have a better `__repr__` or have a custom method to nicely pprint the dataset object? 

Here is my very simple attempt. With this PR, it produces:
```
DatasetDict({
  train:   Dataset({
    features: ['text', 'headline', 'title'],
    num_rows: 157252
  })
  validation:   Dataset({
    features: ['text', 'headline', 'title'],
    num_rows: 5599
  })
  test:   Dataset({
    features: ['text', 'headline', 'title'],
    num_rows: 5577
  })
})
```
I did omit the data types on purpose to make it more readable, but it shouldn't be too difficult to integrate those too.

note that this PR also fixes the inconsistency in output that in master misses enclosing `{}` for Dataset, but it is there for `DatasetDict` - or perhaps it was by design.

I'm totally not attached to this format, just wanting something more readable. One approach could be to serialize to `json.dumps` or something similar. It'd make the indentation simpler.

Thank you.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -702,7 +702,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
             )
 
     def __repr__(self):
-        return f"Dataset(features: {self.features}, num_rows: {self.num_rows})"
+        return f"  Dataset(\n    features: {self.features.keys()},\n    num_rows: {self.num_rows}\n  )"
 
     @property
     def format(self):
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -702,7 +702,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
             )
 
     def __repr__(self):
-        return f"  Dataset(\n    features: {self.features.keys()},\n    num_rows: {self.num_rows}\n  )"
+        return f"  Dataset({{\n    features: {list(self.features.keys())},\n    num_rows: {self.num_rows}\n  }})"
 
     @property
     def format(self):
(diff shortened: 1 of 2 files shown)


lhoestq (approved):
I like this format, this makes things way more readable.
Thanks !

stas00:
OK, then I improved the code a bit to automate indentation in the parent class, so that the child repr doesn't need to guess and works nicely on its own.

- do we want indent=4 or 2?
- do we want `{` ... `}` or w/o?

currently it's indent4 and w/ curly braces, so it looks:

```
DatasetDict({
    train: Dataset({
        features: ['text', 'headline', 'title'],
        num_rows: 157252
    })
    validation: Dataset({
        features: ['text', 'headline', 'title'],
        num_rows: 5599
    })
    test: Dataset({
        features: ['text', 'headline', 'title'],
        num_rows: 5577
    })
})
```

thomwolf (approved):

thomwolf:
Yes!  A lot better indeed!

lhoestq (approved):
I like it this way !
Thanks :)

PR #731, opened 2020-10-14
dataset(aslg_pc12): initial loading script

AmitMY:
This contains the only current public part of this corpus.

The rest of the corpus is not yet been made public, but this sample is still being used by researchers.


datasets/aslg_pc12/aslg_pc12.py
```diff
@@ -0,0 +1,85 @@
...
+
+from __future__ import absolute_import, division, print_function
+
+import csv
+import os
+
+import datasets
+
+_DESCRIPTION = """\
+Synthetic English-ASL Gloss Parallel Corpus 2012
```
lhoestq, line 25: Maybe you can provide more details in the description ?
That's what is showed in the dataset pages at https://huggingface.co/datasets

lhoestq (commented):
Thanks for adding this one :)

To fix the CI you'll need to add the encodings in the `open` calls, and also add the dummy data.

AmitMY:
Thanks @lhoestq 
Are there any guidelines for the dummy data?
In this particular case for example, the dataset fetches from two hardcoded URLs. 
Do I just `head -n 10` both files and zip them?

lhoestq:
> Thanks @lhoestq
> Are there any guidelines for the dummy data?
> In this particular case for example, the dataset fetches from two hardcoded URLs.
> Do I just `head -n 10` both files and zip them?

Yes the idea is just to have a few examples to properly test the script and make sure it keeps working in the long run.

And FYI there's a command to help you name the dummy data files correctly. More info in the documentation [here](https://huggingface.co/docs/datasets/share_dataset.html#adding-dummy-data)

AmitMY:
@lhoestq passes all tests

lhoestq (approved):
Awesome thanks !

PR #734, opened 2020-10-14
Fix GLUE metric description

sgugger:
Small typo: the description says translation instead of prediction.


lhoestq (approved):
Good catch, thanks

PR #736, opened 2020-10-15
Start community-provided dataset docs

sshleifer:
This is one I did to get the pseudo-labels updated. Not sure if it generalizes, but I figured I would write it down. It was pretty easy because all I had to do was make properly formatted directories and change URLs.

+ In slack @thomwolf  called it a `user-namespace` dataset, but the docs call it `community dataset`.
I think the first naming is clearer, but I didn't address that here.

+ I didn't add metadata, will try that.


datasets/wmt_t2t/wmt_utils.py
```diff
@@ -385,10 +385,18 @@ def get_path(self, src):
     SubDataset(
         name="uncorpus_v1",
         target="en",
-        sources={"ru", "zh"},
+        sources={"ru"},
```
sshleifer, line 388: accidental, will rm

sshleifer: actually the delta is just wrong.

julien-c:
can you also reference the `--organization` flag like in https://github.com/huggingface/transformers/blob/master/docs/source/model_sharing.rst#upload-your-model-with-the-cli ?

lhoestq:
Not sure if the changes in `datasets/wmt_t2t/wmt_utils.py` are intentional.
If you want to add more configs to wmt, could you do it in a serapate PR ?

sshleifer:
I don't think I changed wmt_utils I think github is wrong.

sshleifer:
Trying new fork.

PR #742, opened 2020-10-19
Add OCNLI, a new CLUE dataset

JetRunner:
OCNLI stands for Original Chinese Natural Language Inference. It is a corpus for
            Chinese Natural Language Inference, collected following closely the procedures of MNLI,
            but with enhanced strategies aiming for more challenging inference pairs. We want to
            emphasize we did not use human/machine translation in creating the dataset, and thus
            our Chinese texts are original and not translated.


lhoestq (approved):
Awesome thanks :) 

Before merging, could you update the dataset_infos.json file so that it contains the metadata of the download link you just changed ?

lhoestq:
Thanks :) merging it

PR #745, opened 2020-10-20
Fix emotion description

lewtun:
Fixes the description of the emotion dataset to reflect the class names observed in the data, not the ones described in the paper.

I also took the liberty to make use of `ClassLabel` for the emotion labels.


lhoestq (approved):
LGTM
Also thank you for switching to ClassLabel and updating the dataset_infos.json file

juliette-sch:
Hello, I probably have a silly question but the labels of the emotion dataset are in the form of numbers and not string, so I can not use the function classification_report because it mixes numbers and string (prediction). How can I access the label in the form of a string and not a number? 
Thank you in advance.

Issue #672, opened 2020-09-26
Questions about XSUM 


danyaljj:
Hi there ✋ 

I'm looking into your `xsum` dataset and I have several questions on that. 
So here is how you load it: 
```
>>> data = datasets.load_dataset('xsum', version='1.0.1')
>>> data['train']
Dataset(features: {'document': Value(dtype='string', id=None), 'summary': Value(dtype='string', id=None)}, num_rows: 204017)
>>> data['test']
Dataset(features: {'document': Value(dtype='string', id=None), 'summary': Value(dtype='string', id=None)}, num_rows: 11333)
```

The first issue is, the instance counts don’t match what I see on [the dataset's website](https://github.com/EdinburghNLP/XSum/tree/master/XSum-Dataset#what-builds-the-xsum-dataset):
```
 … training (90%, **204,045**), validation (5%, 11,332), and test (5%, **11,334**) set.
```
Any thoughts why? Perhaps @mariamabarham could help here, since she recently had a PR on this dataaset https://github.com/huggingface/datasets/pull/289  (reviewed by @patrickvonplaten) 

Another issue is that the instances don't seem to have IDs. The original datasets provides IDs for the instances: https://github.com/EdinburghNLP/XSum/blob/master/XSum-Dataset/XSum-TRAINING-DEV-TEST-SPLIT-90-5-5.json but to be able to use them, the dataset sizes need to match. 

CC @jbragg

lhoestq:
We should try to regenerate the data using the official script.
But iirc that's what we used in the first place, so not sure why it didn't match in the first place.

I'll let you know when the dataset is updated

danyaljj:
Thanks, looking forward to hearing your update on this thread. 

This is a blocking issue for us; would appreciate any progress on this front. We can also help with the fix, if you deem it appropriately.

lhoestq:
I just started the generation on my side, I'll let you know how it goes :)

lhoestq:
Hmm after a first run I'm still missing 136668/226711 urls.
I'll relaunch it tomorrow to try to get the remaining ones.

lhoestq:
Update: I'm missing 36/226711 urls but I haven't managed to download them yet

danyaljj:
Thanks! That sounds like a reasonable number!

lhoestq:
So I managed to download them all but when parsing only 226181/226711 worked.
Not sure if it's worth digging and debugging parsing at this point :/

lhoestq:
[...]

PR #754, opened 2020-10-23
Use full released xsum dataset

jbragg:
#672 Fix xsum to expand coverage and include IDs
Code based on parser from older version of `datasets/xsum/xsum.py`
@lhoestq


datasets/xsum/xsum.py
```diff
@@ -73,39 +96,59 @@ def _info(self):
     def _split_generators(self, dl_manager):
         """Returns SplitGenerators."""
 
-        dl_path = dl_manager.download_and_extract(_URL)
+        data_path = dl_manager.download_and_extract(_URL_DATA)
+        splits_path = dl_manager.download(_URL_SPLITS)
```
lhoestq, line 100: Could you change this to
```python
files_to_download = {
    "data": _URL_DATA,
    "splits": _URL_SPLITS
}
downloaded_files = dl_manager.download_and_extract(files_to_download)
```
since the download manager is able to do parallel downloads when a dict is passed.
Moreover it should fix the dummy data error in the CI

datasets/xsum/xsum.py
```diff
@@ -73,39 +96,59 @@ def _info(self):
...
+
+        with open(splits_path, "r") as f:
```
lhoestq, line 102: Could you specify the encoding each time you call `open` ? `encoding="utf-8"` for example.
This is for windows users.

lhoestq (commented):
Thank you for updating xsum to get the IDS ! Also it looks like it also includes samples that were missing.

I left a few comments that should help you make the CI pass.

Moreover could you increase the dataset `VERSION` field ?
And remove the `SUPPORTED_VERSIONS` one that is not supported anyways.

jbragg:
@lhoestq I took a shot at addressing your comments but the build scripts seem to be complaining about not being able to open dummy files. How do I resolve those errors without copying the full dataset into the dummy dir?

lhoestq:
Could you check that the names of the dummy data files are right ?
You can use 
```
datasets-cli dummy_data ./datasets/xum
```
to print the expected file names

jbragg:
I did what it suggested
```
- Please create the following dummy data files 'dummy_data/XSum-TRAINING-DEV-TEST-SPLIT-90-5-5.json' from the folder './datasets/xsum/dummy/1.2.0'

- For each of the splits 'train, validation, test', make sure that one or more of the dummy data files provide at least one example

- If the method `_generate_examples(...)` includes multiple `open()` statements, you might have to create other files in addition to 'dummy_data/XSum-TRAINING-DEV-TEST-SPLIT-90-5-5.json'. In this case please refer to the `_generate_examples(...)` method
```
but this last point wasn't clear to me

jbragg:
Ok @lhoestq looks like I got the tests to pass :)

lhoestq (approved):
Awesome thank you !

PR #756, opened 2020-10-23
Start community-provided dataset docs 

sshleifer:
Continuation of #736 with clean fork.

#### Old description
This is what I did to get the pseudo-labels updated. Not sure if it generalizes, but I figured I would write it down. It was pretty easy because all I had to do was make properly formatted directories and change URLs.

In slack @thomwolf called it a user-namespace dataset, but the docs call it community dataset.
I think the first naming is clearer, but I didn't address that here.

I didn't add metadata, will try that.

### DIFF
--- a/docs/source/share_dataset.rst
+++ b/docs/source/share_dataset.rst
@@ -102,8 +102,31 @@ Push the changes to your account using:
 
 Sharing a "community provided" dataset
 -----------------------------------------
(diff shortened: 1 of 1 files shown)


lhoestq (approved):

thomwolf:
Oh, really cool @sshleifer!

PR #776, opened 2020-10-29
Allow custom split names in text dataset

### LINKED ISSUE #735: Throw error when an unexpected key is used in data_files
I have found that only "train", "validation" and "test" are valid keys in the `data_files` argument. When you use any other ones, those attached files are silently ignored - leading to unexpected behaviour for the users.

So the following, unintuitively, returns only one key (namely `train`).

```python
datasets = load_dataset("text", data_files={"train": train_f, "valid": valid_f})
print(datasets.keys())
# dict_keys(['train'])
```

whereas using `validation` instead, does return the expected result:

```python
datasets = load_dataset("text", data_files={"train": train_f, "validation": valid_f})
print(datasets.keys())
# dict_keys(['train', 'validation'])
```

I would like to see more freedom in which keys one can use, but if that is not possible at least an error should be thrown when using an unexpected key.

lhoestq:
The `text` dataset used to return only splits like train, test and validation. Other splits were ignored.
Now any split name is allowed.

I did the same for `json`, `pandas` and `csv`

Fix #735

### DIFF
--- a/datasets/csv/csv.py
+++ b/datasets/csv/csv.py
@@ -67,12 +67,10 @@ class Csv(datasets.ArrowBasedBuilder):
                 files = [files]
             return [datasets.SplitGenerator(name=datasets.Split.TRAIN, gen_kwargs={"files": files})]
         splits = []
-        for split_name in [datasets.Split.TRAIN, datasets.Split.VALIDATION, datasets.Split.TEST]:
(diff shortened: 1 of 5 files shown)


BramVanroy:
Awesome! This will make the behaviour much more intuitive for some non-standard code.

Thanks!

PR #779, opened 2020-10-29
Feature/fidelity metrics from emnlp2020 evaluating and characterizing human rationales

rathoreanirudh:
This metric computes fidelity (Yu et al. 2019, DeYoung et al. 2019) and normalized fidelity (Carton et al. 2020).

### DIFF
--- a/metrics/fidelity/fidelity.py
+++ b/metrics/fidelity/fidelity.py
@@ -1,0 +1,145 @@
+import logging
+
+import numpy as np
+
+import metrics.fidelity.utility as utility
+import datasets
+
+
+logger = logging.getLogger(__name__)
+
+_CITATION = """
+citation is under construction 
+@misc{carton2020evaluating,
+      title={Evaluating and Characterizing Human Rationales}, 
+      author={Samuel Carton and Anirudh Rathore and Chenhao Tan},
+      year={2020},
+      eprint={2010.04736},
+      archivePrefix={arXiv},
+      primaryClass={cs.CL}
+}
+"""
+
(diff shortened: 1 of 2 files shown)


lhoestq:
Hi ! This looks interesting, thanks for adding it :) 

For metrics there should only be two features fields: references and predictions.
Both of them can be defined as you want using nested structures if you need to.
Also I'm not sure what goes into references and what goes into predictions, could you give more details please ?
All the other computations parameters (model etc.) are fine though. Maybe explain a bit more what they're used for

rathoreanirudh:
> Hi ! This looks interesting, thanks for adding it :)
> 
> For metrics there should only be two features fields: references and predictions.
> Both of them can be defined as you want using nested structures if you need to.
> Also I'm not sure what goes into references and what goes into predictions, could you give more details please ?
> All the other computations parameters (model etc.) are fine though. Maybe explain a bit more what they're used for

The `predictions` are the predicted labels by a model for a particular input. Do you mean making `prob_y_hat`, `prob_y_hat_alpha` and the `null_difference` a part of references? Also, I have added the description for other parameters in kwargs_description. I can expand it if that makes sense?

lhoestq:
I think every value that is generated by the model (so label, prob_y_hat, prob_y_hat_alpha etc.) should be in `predictions`.
Feel free to add more details in the kwargs_description, this is very useful for the end user.

rathoreanirudh:
Hi @lhoestq , I have updated the code according to your feedback. Please, let me know if it looks good and can be merged now.

albertvillanova:

PR #781, opened 2020-10-30
Add XNLI train set

lhoestq:
I added the train set that was built using the translated MNLI.
Now you can load the dataset specifying one language:
```python

xnli_en = load_dataset("xnli", "en")
print(xnli_en["train"][0])
# {'hypothesis': 'Product and geography are what make cream skimming work .', 'label': 1, 'premise': 'Conceptually cream skimming has two basic dimensions - product and geography .'}
print(xnli_en["test"][0])                                                                                                                   
# {'hypothesis': 'I havent spoken to him again.', 'label': 2, 'premise': "Well, I wasn't even thinking about that, but I was so frustrated, and, I ended up talking to him again."}
```

Cc @sgugger


EvanYangAB:
Hi! Thanks for adding the translated MNLI! Do you know what translations system / model you used when you created the datasets in the other languages?

lhoestq:
> The training data is not from translators.

What makes you think that ? The paper litteraly says

> we hire translators to translate the resulting sentences into 15 languages using the One Hour Translation platform.

EvanYangAB:
However the annotators only did test and validation sets, as this was what
in the paper: “we construct an evaluation set for XLU by extending the
development and test sets of the Multi-Genre Natural Language Inference
Corpus (MultiNLI) to 15 languages".

PR #785, opened 2020-10-31
feat(aslg_pc12): add dev and test data splits

AmitMY:
For reproducibility sake, it's best if there are defined dev and test splits.

The original paper author did not define splits for the entire dataset, not for the sample loaded via this library, so I decided to define:
- 5/7th for train
- 1/7th for dev
- 1/7th for test


lhoestq:
Hi ! I'm not sure we should make this split decision arbitrarily on our side. Users can split it afterwards to whatever they want using `dataset.train_test_split` for example.
Moreover it looks like there's already papers that use this dataset and propose their own splits ([here](http://xanthippi.ceid.upatras.gr/HealthSign/resources/Publications/sitis_paper_25_10.pdf) 80-20) 
What do you think ?

AmitMY:
I was not aware of the `train_test_split` method, thanks!
Soe ven though it contributes to reproducibility, no need to do this split then.

PR #791, opened 2020-11-02
add amazon reviews

joeddav:
Adds the Amazon US Reviews dataset as requested in #353. Converted from [TensorFlow Datasets](https://www.tensorflow.org/datasets/catalog/amazon_us_reviews). cc @clmnt @sshleifer


datasets/amazon_us_reviews/amazon_us_reviews.py
```diff
@@ -0,0 +1,199 @@
...
+    helpful_votes     - Number of helpful votes.
+    total_votes       - Number of total votes the review received.
+    vine              - Review was written as part of the Vine program.
+    verified_purchase - The review is on a verified purchase.
+    review_headline   - The title of the review.
+    review_body       - The review text.
+    review_date       - The date the review was written.
+"""
+
+_DATA_OPTIONS_V1_00 = [
```
patrickvonplaten, line 51: Do we need `_DATA_OPTIONS_V1_00`, ... really? Since it seems that all of _00, _01, _02, is concatenated anyways and never used on its own, I'd think that it's better to just have one `_PRODUCTS` list

datasets/amazon_us_reviews/amazon_us_reviews.py
```diff
@@ -0,0 +1,199 @@
...
+_DL_URLS = {
+    name: "https://s3.amazonaws.com/amazon-reviews-pds/tsv/amazon_reviews_us_" + name + ".tsv.gz"
+    for name in _DATA_OPTIONS
+}
+
+
+class AmazonUSReviewsConfig(datasets.BuilderConfig):
+    """BuilderConfig for AmazonUSReviews."""
+
+    def __init__(self, *, data=None, **kwargs):
```
patrickvonplaten, line 121: why do we have the `*` in the init?

datasets/amazon_us_reviews/amazon_us_reviews.py
```diff
@@ -0,0 +1,199 @@
...
+            homepage="https://s3.amazonaws.com/amazon-reviews-pds/readme.html",
+            citation=_CITATION,
+        )
+
+    def _split_generators(self, dl_manager):
+        url = _DL_URLS[self.config.name]
+        path = dl_manager.download_and_extract(url)
+
+        # There is no predefined train/val/test split for this dataset.
+        return [
```
patrickvonplaten, line 182: I'm fine with this! I did the splitting myself for other datasets, just (90%, 5%, 5%), but not at all sure what the better approach is here. Think this way here is actually better because if there is no official split we should not do it and the user can easily make use of `select` or `filter` to do the split her/himself.

patrickvonplaten (approved):
PR looks good to me! All the slow tests pass. 

I added some suggestions, but I don't think there are minor things. 
Still curious though why there is a `*` in the `__init__` of the config.

datasets/amazon_us_reviews/amazon_us_reviews.py
joeddav: Yeah my feeling is if there's no official split that it's best to let the user create their own, but I guess that's a general question that we don't have a consensus around.

datasets/amazon_us_reviews/amazon_us_reviews.py
thomwolf: Usually this is to prevent the use of positional args (we used that for the metrics in the lib to prevent user-mistakes) but I'll let @joeddav confirm

joeddav: This is just an artifact I missed from the original TFDS code. Maybe they used it for that reason but I don't think it's needed.

joeddav:
@patrickvonplaten Yeah this is adapted from tfds so a lot is just how they wrote the code. Addressed your comments and also simplified the weird `AmazonUSReviewsConfig` definition. Will merge once tests pass.

lhoestq:
Thanks for checking this one :) 
Looks good to me 

Just one question : is there a particular reason to use `names=["Y", "N"]` in this order ? Usually the positive label is at index 1 and the negative one at index 0 for binary classification

joeddav:
> is there a particular reason to use `names=["Y", "N"]` in this order ? Usually the positive label is at index 1 and the negative one at index 0 for binary classification

Hmm that's a good point. I'll submit a quick fix.

PR #810, opened 2020-11-06
Fix seqeval metric

sgugger:
The current seqeval metric returns the following error when computed:
```
~/.cache/huggingface/modules/datasets_modules/metrics/seqeval/78a944d83252b5a16c9a2e49f057f4c6e02f18cc03349257025a8c9aea6524d8/seqeval.py in _compute(self, predictions, references, suffix)
    102         scores = {}
    103         for type_name, score in report.items():
--> 104             scores[type_name]["precision"] = score["precision"]
    105             scores[type_name]["recall"] = score["recall"]
    106             scores[type_name]["f1"] = score["f1-score"]

KeyError: 'LOC'
```
This is because the current code basically tries to do:
```
scores = {}
scores["LOC"]["precision"] = some_value
```
which does not work in python. This PR fixes that while keeping the previous nested structure of results, with the same keys.

### DIFF
--- a/metrics/seqeval/seqeval.py
+++ b/metrics/seqeval/seqeval.py
@@ -99,13 +99,15 @@ class Seqeval(datasets.Metric):
         report.pop("weighted avg")
         overall_score = report.pop("micro avg")
 
-        scores = {}
-        for type_name, score in report.items():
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
Good catch thanks !

PR #819, opened 2020-11-09
Make save function use deterministic global vars order

### LINKED ISSUE #816: [Caching] Dill globalvars() output order is not deterministic and can cause cache issues.
Dill uses `dill.detect.globalvars` to get the globals used by a function in a recursive dump. `globalvars` returns a dictionary of all the globals that a dumped function needs. However the order of the keys in this dict is not deterministic and can cause caching issues.

To fix that one could register an implementation of dill's `save_function` in the `datasets` pickler that sorts the globals keys before dumping a function.

lhoestq:
The `dumps` function need to be deterministic for the caching mechanism.
However in #816 I noticed that one of dill's method to recursively check the globals of a function may return the globals in different orders each time it's used. To fix that I sort the globals by key in the `globs` dictionary.
I had to add a rectified `save_function` to the saving functions registry of the Pickler to make it work.

This should fix #816

### DIFF
--- a/src/datasets/utils/py_utils.py
+++ b/src/datasets/utils/py_utils.py
@@ -524,7 +524,7 @@ def save_function(pickler, obj):
     """
     if not dill._dill._locate_function(obj):
         dill._dill.log.info("F1: %s" % obj)
-        if getattr(pickler, '_recurse', False):
+        if getattr(pickler, "_recurse", False):
             # recurse to get all globals referred to by obj
             globalvars = dill.detect.globalvars
             globs = globalvars(obj, recurse=True, builtin=True)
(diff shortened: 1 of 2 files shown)


thomwolf (approved):
Impressive catch!

davips:
Sorry, asking for help here, but the dill thread stop around 2013. Is it possible to use dill deterministically? I tried to monkeypatch the solution presented here into dill, but I suppose it requires forking their project.

lhoestq:
Hi ! What we did was to subclass `dill`'s Pickler to fix the non-deterministic behaviors, and it's been working fine. A fork should also do the job

PR #825, opened 2020-11-10
Add accuracy, precision, recall and F1 metrics

jplu:
This PR adds several single metrics, namely:

- Accuracy
- Precision
- Recall
- F1

They all uses under the hood the sklearn metrics of the same name. They allow different useful features when training a multilabel/multiclass model:
- have a macro/micro/per label/weighted/binary/per sample score
- score only the selected labels (usually what we call the positive labels) and ignore the negative ones. For example in case of a Named Entity Recognition task, positive labels are (`PERSON`, `LOCATION` or `ORGANIZATION`) and the negative one is `O`.

### DIFF
--- a/metrics/accuracy/accuracy.py
+++ b/metrics/accuracy/accuracy.py
@@ -14,8 +14,9 @@
 # limitations under the License.
 """Accuracy metric."""
 
-import datasets
 from sklearn.metrics import accuracy_score
+
+import datasets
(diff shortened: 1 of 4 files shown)


lhoestq (approved):
LGTM thanks :)

thomwolf (approved):
Very nice!

PR #840, opened 2020-11-11
Update squad_v2.py

Javier-Jimenez99:
Change lines 100 and 102 to prevent overwriting ```predictions``` variable.


Javier-Jimenez99:
With this change all the checks are passed.

ijaz314:
Good

lhoestq (approved):
Good catch thanks !

PR #858, opened 2020-11-16
Add SemEval-2010 task 8

JoelNiklaus:
Hi,
I don't know how to add dummy data, since I create the validation set out of the last 1000 examples of the train set. If you have a suggestion, I am happy to implement it.
Cheers,
Joel


JoelNiklaus:
Added dummy data and encoding to open(). Now everything should be fine, hopefully :)

lhoestq (approved):
Looks all good thank you !

PR #868, opened 2020-11-18
Consistent metric outputs

lhoestq:
To automate the use of metrics, they should return consistent outputs.
In particular I'm working on adding a conversion of metrics to keras metrics.
To achieve this we need two things:
- have each metric return dictionaries of string -> floats since each keras metrics should return one float
- define in the metric info the different fields of the output dictionary

In this PR I'm adding these two features.
I also fixed a few bugs in some metrics

#867 needs to be merged first

### DIFF
--- a/metrics/accuracy/accuracy.py
+++ b/metrics/accuracy/accuracy.py
@@ -39,16 +39,31 @@ Returns:
     accuracy: Accuracy score.
 """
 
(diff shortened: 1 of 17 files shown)


lhoestq:
I keep this PR in stand-by for next week's datasets sprint

PR #882, opened 2020-11-24
Update README.md

vaibhavad:
"no label" is "-" in the original dataset but "-1" in Huggingface distribution.


lhoestq (approved):
Right ! thank you :)

PR #884, opened 2020-11-24
Auto generate dummy data

lhoestq:
When adding a new dataset to the library, dummy data creation can take some time.
To make things easier I added a command line tool that automatically generates dummy data when possible.

The tool only supports certain data files types: txt, csv, tsv, jsonl, json and xml.

Here are some examples:
```
python datasets-cli dummy_data ./datasets/snli --auto_generate
python datasets-cli dummy_data ./datasets/squad --auto_generate --json_field data
python datasets-cli dummy_data ./datasets/iwslt2017 --auto_generate --xml_tag seg --match_text_files "train*" --n_lines 15
# --xml_tag seg => each sample corresponds to a "seg" tag in the xml tree
# --match_text_files "train*" =>  also match text files that don't have a proper text file extension (no suffix like ".txt" for example)
# --n_lines 15 => some text files have headers so we have to use at least 15 lines
```

and here is the command usage:

```
usage: datasets-cli <command> [<args>] dummy_data [-h] [--auto_generate]
                                  [--n_lines N_LINES]
                                  [--json_field JSON_FIELD]
                                  [--xml_tag XML_TAG]
                                  [--match_text_files MATCH_TEXT_FILES]
                                  [--keep_uncompressed]
                                  [--cache_dir CACHE_DIR]
                                  path_to_dataset

positional arguments:
path_to_dataset       Path to the dataset (example: ./datasets/squad)

optional arguments:
-h, --help            show this help message and exit
--auto_generate       Try to automatically generate dummy data
--n_lines N_LINES     Number of lines or samples to keep when auto-
                                                generating dummy data
--json_field JSON_FIELD
                                                Optional, json field to read the data from when auto-
                                                generating dummy data. In the json data files, this
                                                field must point to a list of samples as json objects
                                                (ex: the 'data' field for squad-like files)
--xml_tag XML_TAG     Optional, xml tag name of the samples inside the xml
                                                files when auto-generating dummy data.
--match_text_files MATCH_TEXT_FILES
                                                Optional, a comma separated list of file patterns that
                                                looks for line-by-line text files other than *.txt or
                                                *.csv. Example: --match_text_files *.label
--keep_uncompressed   Don't compress the dummy data folders when auto-
                                                generating dummy data. Useful for debugging for to do
                                                manual adjustements before compressing.
--cache_dir CACHE_DIR
                                                Cache directory to download and cache files when auto-
                                                generating dummy data
```

The command generates all the necessary `dummy_data.zip` files (one per config).

How it works:
- it runs the split_generators() method of the dataset script to download the original data files
- when downloading it records a mapping between the downloaded files and the corresponding expected dummy data files paths
- then for each data file it creates the dummy data file keeping only the first samples (the strategy depends on the type of file)
- finally it compresses the dummy data folders into dummy_zip files ready for dataset tests

Let me know if that makes sense or if you have ideas to improve this tool !

I also added a unit test.


src/datasets/commands/dummy_data.py
```diff
@@ -49,84 +280,125 @@ def run(self):
...
+            dl_manager.compress_autogenerated_dummy_data(path_do_dataset)
+        else:
+            generated_dummy_data_dir = os.path.join(self._path_to_dataset, mock_dl_manager.dummy_data_folder)
+            logger.info(
+                f"Dummy data generated in directory '{generated_dummy_data_dir}' but kept uncompressed. "
+                "Please compress this directory into a zip file to use it for dummy data tests."
+            )
+        return dl_manager
 
-            files_to_create = set()
```
lhoestq, line 69: I moved all this part in the function below without changes.
Sorry github makes it kinda hard to read

src/datasets/commands/dummy_data.py
```diff
@@ -13,28 +24,248 @@
...
+    def download_and_extract(self, url_or_urls):
+        output = super().extract(super().download(url_or_urls))
+        dummy_output = self.mock_download_manager.download(url_or_urls)
+        map_nested(self.downloaded_paths.append, output, map_tuple=True)
+        map_nested(self.expected_dummy_paths.append, dummy_output, map_tuple=True)
+        return output
+
+    def auto_generate_dummy_data_folder(
+        self,
+        n_lines=5,
```
patrickvonplaten, line 60:
```suggestion
        n_lines: int =5,
```

src/datasets/commands/dummy_data.py
```diff
@@ -13,28 +24,248 @@
...
+
+
 class DummyDataCommand(BaseTransformersCLICommand):
     @staticmethod
     def register_subcommand(parser: ArgumentParser):
         test_parser = parser.add_parser("dummy_data")
-        test_parser.add_argument("--requires_manual", action="store_true", help="Dataset requires manual data")
-        test_parser.add_argument("path_to_dataset", type=str, help="Name of the dataset to download")
+        test_parser.add_argument(
+            "--auto_generate", action="store_true", help="Try to automatically generate dummy data"
```
patrickvonplaten, line 206:
```suggestion
            "--auto_generate", action="store_true", help="Automatically generate dummy data"
```

src/datasets/commands/dummy_data.py
```diff
@@ -13,28 +24,248 @@
...
+        test_parser.add_argument(
+            "--match_text_files",
+            type=str,
+            default=None,
+            help="Optional, a comma separated list of file patterns that looks for line-by-line text files other than *.txt or *.csv. Example: --match_text_files *.label",
+        )
+        test_parser.add_argument(
+            "--keep_uncompressed",
+            action="store_true",
+            help="Don't compress the dummy data folders when auto-generating dummy data. Useful for debugging for to do manual adjustements before compressing.",
```
patrickvonplaten, line 232:
```suggestion
            help="Whether to leave the dummy data folders uncompressed when auto-generating dummy data. Useful for debugging for to do manual adjustements before compressing.",
```

patrickvonplaten (approved):
This looks amazing! Can't wait to try it out

docs/source/share_dataset.rst
```diff
@@ -246,6 +246,64 @@ This command will output instructions specifically tailored to your dataset and
 	- Make sure you have created the file 'dummy_data.zip' in './datasets/my-dataset/dummy/1.1.0' 
 	===================================================================================
 
+There is a tool that automatically generates dummy data for you but it only works with some types of data files such as txt, csv, tsv, jsonl, json and xml.
```
thomwolf, line 249:
```suggestion
There is a tool that automatically generates dummy data for you. At the moment it supports data files in the following format: txt, csv, tsv, jsonl, json, xml.

If your data files are in another format you will have to generate dummy data yourself by creating small data files with only a few examples from the full data files.

```

src/datasets/commands/dummy_data.py
```diff
@@ -13,28 +24,248 @@
...
+                with open(src_path, "r", encoding="utf-8") as src_file:
+                    json_data = json.load(src_file)
+                    if json_field is not None:
+                        json_data = json_data[json_field]
+                    if isinstance(json_data, dict):
+                        if not all(isinstance(v, list) for v in json_data.values()):
+                            raise ValueError(
+                                f"Couldn't parse columns {list(json_data.keys())}. "
+                                "Maybe specify which json field must be used "
+                                "to read the data with --json_field <my_field>."
```
thomwolf, line 131: Very nice!

thomwolf (approved):
Look really great to me! Looking forward to try it!

Should we run a test of the dataset at the end (if the auto-generation worked) to immediately show to the user that the process worked well?

lhoestq:
I took your comments into account.
Also now after compressing the dummy_data.zip file it runs a dummy data test (=make sure each split has at least 1 example using the dummy data)

lhoestq:
I just tested the tool with some datasets and found out that it's not working for datasets that download files using `download_and_extract(file_url)` (where file_url is a `str`). That's because in that case the dummy_data.zip is not a folder but a single zipped file.

I think we have to fix that or we can have unexpected behavior when a scripts calls `download_and_extract(file_url)` several times, since it would always point to the same dummy data file.

So I decided to change that to have a folder containing the dummy files instead but it breaks around 90 tests so I need to update 90 dummy data files to follow this scheme. I'll probably fix them tomorrow morning.

What do you guys think ? Also cc @patrickvonplaten to make sure I understand things correctly

lhoestq:
Ok I changed to use the dummy_data.zip content to be a folder even for single url calls to `dl_manager.download_and_extract`. Therefore the automatic dummy data generation tool works for most datasets now.

To avoid having to change all the old dummy_data.zip files I added backward compatiblity. 

The only test failing is `tests/test_dataset_common.py::RemoteDatasetTest::test_load_dataset_xcopa`
It is expected to fail since I had modify its dummy data structure that was wrong. It was causing issue with backward compatibility

PR #889, opened 2020-11-25
Optional per-dataset default config name

joeddav:
This PR adds a `DEFAULT_CONFIG_NAME` class attribute to `DatasetBuilder`. This allows a dataset to have a specified default config name when a dataset has more than one config but the user does not specify it. For example, after defining `DEFAULT_CONFIG_NAME = "combined"` in PolyglotNER, a user can now do the following:

```python
ds = load_dataset("polyglot_ner")
```
which is equivalent to,
```python
ds = load_dataset("polyglot_ner", "combined")
```
In effect (for this particular dataset configuration), this means that if the user doesn't specify a language, they are given the combined dataset including all languages.

Since it doesn't always make sense to have a default config, this feature is opt-in. If `DEFAULT_CONFIG_NAME` is not defined and a user does not pass a config for a dataset with multiple configs available, a ValueError is raised like usual.

Let me know what you think about this approach @lhoestq @thomwolf and I'll add some documentation and define a default for some of our existing datasets.


lhoestq:
I like the idea ! And the approach is right imo

Note that by changing this we will have to add a way for users to get the config lists of a dataset. In the current user workflow, the user could see the list of the config when the missing config error is raised but now it won't be the case because of the default config.

lhoestq:
Maybe let's add a test in the test_builder.py test script ?

src/datasets/inspect.py
```diff
@@ -95,3 +95,29 @@ def inspect_metric(path: str, local_path: str, download_config: Optional[Downloa
...
+    """Get the list of available config names for a particular dataset.
+
+    Args:
+        path (``str``): path to the dataset processing script with the dataset builder. Can be either:
+            - a local path to processing script or the directory containing the script (if the script has the same name as the directory),
+                e.g. ``'./dataset/squad'`` or ``'./dataset/squad/squad.py'``
+            - a dataset identifier on HuggingFace AWS bucket (list all available datasets and ids with ``datasets.list_datasets()``)
+                e.g. ``'squad'``, ``'glue'`` or ``'openai/webtext'``
+    """
+    return list(get_dataset_infos(path).keys())
```
lhoestq, line 123: maybe use `builder_cls.builder_configs.keys()` here ? So that the function works even if the dataset infos has not been exported yet

joeddav:
@lhoestq Okay great, I added a test as well as two new inspect functions: `get_dataset_config_names` and `get_dataset_infos` (the latter is something I've been wanting anyway). As a quick hack, you can also just pass a random config name (e.g. an empty string) to `load_dataset` to get the config names as before. Also added a couple paragraphs to the adding new datasets doc.

I'll send a separate PR incorporating this in existing datasets so we can get this merged before our sprint on Monday.

Any ideas on the failing tests? I'm having trouble making sense of it.

tests/test_builder.py
```diff
@@ -573,3 +584,17 @@ def test_custom_writer_batch_size(self):
             dataset3 = dummy_builder3.as_dataset("train")
             self.assertEqual(len(dataset3._data[0].chunks), 10)
             del dataset1, dataset2, dataset3
+
+    def test_config_names(self):
```
lhoestq, line 588: thanks for adding this test !

src/datasets/inspect.py
```diff
@@ -95,3 +95,33 @@ def inspect_metric(path: str, local_path: str, download_config: Optional[Downloa
...
+    Args:
+        path (``str``): path to the dataset processing script with the dataset builder. Can be either:
+            - a local path to processing script or the directory containing the script (if the script has the same name as the directory),
+                e.g. ``'./dataset/squad'`` or ``'./dataset/squad/squad.py'``
+            - a dataset identifier on HuggingFace AWS bucket (list all available datasets and ids with ``datasets.list_datasets()``)
+                e.g. ``'squad'``, ``'glue'`` or ``'openai/webtext'``
+    """
+    module_path, _ = prepare_module(path)
+    builder_cls = import_main_class(module_path, dataset=True)
+    return list(builder_cls.builder_configs.keys())
```
lhoestq, line 127: love it

src/datasets/builder.py
```diff
@@ -210,20 +213,24 @@ def get_exported_dataset_info(self) -> DatasetInfo:
...
-                raise ValueError(
-                    "Config name is missing."
-                    "\nPlease pick one among the available configs: %s" % list(self.builder_configs.keys())
-                    + "\nExample of usage:\n\t`{}`".format(example_of_usage)
-                )
-            builder_config = self.BUILDER_CONFIGS[0]
-            logger.info("No config specified, defaulting to first: %s/%s", self.name, builder_config.name)
+            if self.DEFAULT_CONFIG_NAME is not None:
+                builder_config = self.builder_configs.get(self.DEFAULT_CONFIG_NAME)
+                logger.info("No config specified, defaulting to: %s/%s", self.name, builder_config.name)
```
lhoestq: As we said offline, this should be warning indeed.

The other one on the other hand `No config specified...` should stay at info level since there is no ambiguity in which config to load.

docs/source/add_dataset.rst
```diff
@@ -312,6 +312,15 @@ While the configuration attributes are used in this case to control the reading/
...
 
+Specifying a default dataset configuration
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When a user loads a dataset with more than one configuration, they must specify a configuration name or else a ValueError is raised. With some datasets, it may be preferable to specify a default configuration that will be loaded if a user does not specify one.
+
+This can be done with the :attr:`datasets.DatasetBuilder.DEFAULT_CONFIG_NAME` attribute. By setting this attribute equal to the name of one of the dataset configurations, that config will be loaded in the case that the user does not specify a config name.
+
+This feature is opt-in and should only be used where a default configuration makes sense for the dataset. For example, many cross-lingual datasets have a different configuration for each language. In this case, it may make sense to create an aggregate configiration which can serve as the default. This would, in effect, load all languages of the dataset by default unless the user specifies a particular language. See the `Polyglot NER loading script <https://github.com/huggingface/datasets/blob/master/datasets/polyglot_ner/polyglot_ner.py>`__ for an example.
```
lhoestq, line 322:
```suggestion
This feature is opt-in and should only be used where a default configuration makes sense for the dataset. For example, many cross-lingual datasets have a different configuration for each language. In this case, it may make sense to create an aggregate configuration which can serve as the default. This would, in effect, load all languages of the dataset by default unless the user specifies a particular language. See the `Polyglot NER loading script <https://github.com/huggingface/datasets/blob/master/datasets/polyglot_ner/polyglot_ner.py>`__ for an example.
```

lhoestq (approved):
Thanks a lot ! LGTM :)

PR #892, opened 2020-11-26
Add a few datasets of reference in the documentation

lhoestq:
I started making a small list of various datasets of reference in the documentation.
Since many datasets share a lot in common I think it's good to have a list of datasets scripts to get some inspiration from.

Let me know what you think, and if you have ideas of other datasets that we may add to this list, please let me know.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -75,6 +75,13 @@
 	```bash
 	python datasets-cli dummy_data datasets/<your-dataset-folder>
 	```
+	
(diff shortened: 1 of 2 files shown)


thomwolf:
Looks good to me. Do we also support TSV in this helper (explain if it should be text or CSV) and in the dummy-data creator?

lhoestq:
snli is basically based on tsv files (but named as .txt) and it is in the list of datasets of reference.
The dummy data creator supports tsv

lhoestq:
merging this one.
If you think of other datasets of reference to add we can still add them later

PR #894, opened 2020-11-26
Allow several tags sets

lhoestq:
Hi !

Currently we have three dataset cards : snli, cnn_dailymail and allocine.
For each one of those datasets a set of tag is defined. The set of tags contains fields like `multilinguality`, `task_ids`, `licenses` etc.

For certain datasets like `glue` for example, there exist several configurations: `sst2`, `mnli` etc. Therefore we should define one set of tags per configuration. However the current format used for tags only supports one set of tags per dataset.

In this PR I propose a simple change in the yaml format used for tags to allow for several sets of tags.

Let me know what you think, especially @julien-c let me know if it's good for you since it's going to be parsed by moon-landing

### DIFF
--- a/datasets/allocine/README.md
+++ b/datasets/allocine/README.md
@@ -1,21 +1,25 @@
-annotations_creators:
-- no-annotation
-language_creators:
-- found
-languages:
-- fr
-licenses:
-- mit
-multilinguality:
-- monolingual
-size_categories:
-- 100K<n<1M
-source_datasets:
-- original
-task_categories:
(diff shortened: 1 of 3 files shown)


julien-c (commented):
discussed w/ @lhoestq offline, I'll do a backward compatible parser on the website side so we can keep things simple in the most common cases

lhoestq:
Closing since we don't need to update the tags of those three datasets (for each one of them there is only one tag set)

PR #896, opened 2020-11-26
Add template and documentation for dataset card

yjernite:
This PR adds a template for dataset cards, as well as a guide to filling out the template and a completed example for the ELI5 dataset, building on the work of @mcmillanmajora 

New pull requests adding datasets should now have a README.md file which serves both to hold the tags we will have to index the datasets and as a data statement.

The template is designed to be pretty extensive. The idea is that the person who uploads the dataset should put in all the basic information (at least the Dataset Description section) and whatever else they feel comfortable adding and leave the `[More Information Needed]` annotation everywhere else as a placeholder.

We will then work with @mcmillanmajora to involve the data authors more directly in filling out the remaining information.

Direct links to:
- [Documentation](https://github.com/yjernite/datasets/blob/add_dataset_card_doc/templates/README_guide.md)
- [Empty template](https://github.com/yjernite/datasets/blob/add_dataset_card_doc/templates/README.md)
- [ELI5 example](https://github.com/yjernite/datasets/blob/add_dataset_card_doc/datasets/eli5/README.md)


templates/README.md
```diff
@@ -0,0 +1,116 @@
+---
+YAML tags:
+- copy-paste the tags obtained with the tagging app: http://34.68.228.168:8501/
```
lhoestq, line 3: This tagger only has existing datasets.
When someone adds a dataset, it won't be listed in there yet (until it's merged on master).

templates/README_guide.md
```diff
@@ -0,0 +1,170 @@
...
+
+Provide a brief overview of the languages represented in the dataset. Describe relevant details about specifics of the language such as whether it is social media text, African American English,...
+
+When relevant, please provide [BCP-47 codes](https://tools.ietf.org/html/bcp47), which consist of a [primary language subtag](https://tools.ietf.org/html/bcp47#section-2.2.1), with a [script subtag](https://tools.ietf.org/html/bcp47#section-2.2.3) and/or [region subtag](https://tools.ietf.org/html/bcp47#section-2.2.4) if available.
+
+## Dataset Structure
+
+### Data Instances
+
+Provide an example and brief description of a typical instance in the dataset. If available, provide a link to further examples.
```
lhoestq, line 57: Maybe suggest to just copy paste the python dictionary of one example ?

templates/README.md
thomwolf: Indeed. @yjernite do you think you could share the code of the streamlit app so that people can run it on their own setup?

templates/README_guide.md
```diff
@@ -0,0 +1,170 @@
...
+
+Describe other people represented or mentioned in the data. Where possible, link to references for the information.
+
+### Annotations
+
+If the dataset contains annotations which are not part of the initial data collection, describe them in the following paragraphs.
+
+#### Annotation process
+
+If applicable, describe the annotation process and any tools used, or state otherwise. Describe the amount of data annotated, if not all. Describe or reference annotation guidelines provided to the annotators. If available, provide interannotator statistics. Describe any annotation validation processes.
```
thomwolf, line 114: What should people do if they can't fill in some paragraph? Add `N/A`?

thomwolf (approved):
This will be really neat!

templates/README_guide.md
yjernite: They should put `N/A` if there is no information in the paragraph or leave `[Needs More Information]` if they simply don't know, in that case we'll eventually want to reach out to the authors or explain why the information is missing.

templates/README_guide.md
yjernite: Good point, I added the suggestion

templates/README.md
yjernite: I made a repo (private for now): https://github.com/huggingface/datasets-tagging

@lhoestq that's right. People can still use it to make yaml tags for any dataset without loading them in the menu for now, I'll look into loading a local dataset first thing Monday before I start working on my list.

julien-c: maybe by default it shouldn't even display a dataset when you open the Streamlit web app (blank dataset) but still let you write tags?

julien-c (commented):
nice!

templates/README.md
yjernite: Yup! I improved the app a bit, and it now loads an empty config. It can also load a local dataset once the CLI has created the `dataset_infos.json` file

PR #901, opened 2020-11-27
Addition of Nl2Bash Dataset

reshinthadithyan:
## Overview
The NL2Bash data contains over 10,000 instances of linux shell commands and their corresponding natural language descriptions provided by experts, from the Tellina system. The dataset features 100+ commonly used shell utilities.
## Footnotes
The following dataset marks the first ML on source code related Dataset in datasets module. It'll be really useful as a lot of the research direction involves Transformer Based Model.
Thanks.
### Reference Links

> Paper Link = https://arxiv.org/pdf/1802.08979.pdf 
> Github Link = https://github.com/TellinaTool/nl2bash

### DIFF
--- a/datasets/nl2bash/nl2bash.py
+++ b/datasets/nl2bash/nl2bash.py
@@ -1,9 +1,23 @@
+# coding=utf-8
+# Copyright 2020 The TensorFlow Datasets Authors and the HuggingFace Datasets Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Lint as: python3
 """Natural Language and Bash Command Dataset"""
 from __future__ import absolute_import, division, print_function
-
 import json
 import os
-
 import datasets
 
 
--- a/datasets/nl2bash/nl2bash.py
+++ b/datasets/nl2bash/nl2bash.py
@@ -23,44 +23,27 @@ Dataset features includes:
   - cmd: Command corresponding to the nl_query.
 This dataset can be downloaded upon requests
 """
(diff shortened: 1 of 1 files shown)


datasets/nl2bash/nl2bash.py
```diff
@@ -0,0 +1,89 @@
...
+        if split_key == "test":
+            return (test_samples)
+        else:
+            raise ValueError(f"Invalid split key {split_key}")
+    def _generate_examples(self,split,data_dir):
+        """Given a split("train","validation","test") gets nl-cmd pair"""
+        data = self._get_examples_from_split(split,data_dir)
+        output_dict = {"nl_query" : [i["invocation"] for i in data],
+                "cmd":[i["cmd"] for i in data]}
+        yield output_dict
```
lhoestq, line 89: here you need to yield a tuple that consists in `(id, example_dict)`

datasets/nl2bash/nl2bash.py
```diff
@@ -0,0 +1,89 @@
...
+        file_dir = os.path.join(data_dir, "nl2bash-data.json")
+        with open(file_dir, 'r') as f:
+            data = json.load(f)
+
+        data_len = len(data)
+
+        # 80/10/10 split
+        i1 = int(data_len * 0.8)
+        i2 = int(data_len * 0.9)
+        i3 = int(data_len*1.0)
```
lhoestq, line 71: Is this way of splitting conventional for this dataset ? i.e. defined by the dataset authors ?
If not, we should probably only provide a train set and let users split it as they want using `dataset.train_test_split`

lhoestq (commented):
Thanks for working on adding this one :)

To make the CI pass you need to generate the json metadata file `dataset_infos.json` and also provide dummy data to make your script properly tested :)

You can find more infos in https://github.com/huggingface/datasets/blob/master/CONTRIBUTING.md#how-to-add-a-dataset

reshinthadithyan:
Hello, thanks. I'm fixing these, will get back. Although I have a question, What should _id_ be in the return statement? Should that be something like a start index (or) the type of split will do? Thanks.

TathagataChakraborti:
@reshinthadithyan we should hold off on this for a couple of weeks till NeurIPS concludes. The [NLC2CMD](http://nlc2cmd.us-east.mybluemix.net/) data will be out then; which includes a cleaner version of this NL2Bash data. The older data is sort of obsolete now.

TathagataChakraborti:
Ah nvm you already commented 😆

PR #903, opened 2020-11-27
Fix URL with backslash in Windows

albertvillanova:
In Windows, `os.path.join` generates URLs containing backslashes, when the first "path" does not end with a slash. 

In general, `os.path.join` should be avoided to generate URLs.

### DIFF
--- a/datasets/boolq/boolq.py
+++ b/datasets/boolq/boolq.py
@@ -3,7 +3,6 @@
 from __future__ import absolute_import, division, print_function
 
 import json
-import os
 
 import datasets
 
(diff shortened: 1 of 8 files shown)


lhoestq (commented):
Good catch indeed ! Thank you so much

We should add a test to make sure that it doesn't happen again.
I'll open a separate PR to add a test if you don't mind

albertvillanova:
@lhoestq I was indeed working on that... to make another commit on this feature branch...

albertvillanova:
But as you prefer... nevermind! :)

lhoestq (approved):
Thanks again ! I just ran the slow tests of all those datasets and they all passed.

lhoestq:
Ah what do you have in mind for the tests ? I was thinking of adding a check in the MockDownloadManager used for tests based on dummy data. I'm creating a PR right now, I'd be happy to have your opinion

albertvillanova:
Indeed I was thinking of something similar: monckeypatching the HTTP request...

albertvillanova:
Therefore, if you agree, I am removing all the rest of `os.path.join`, both from the code and the docs...

lhoestq:
If you spot other `os.path.join` for urls in dataset scripts or metrics scripts feel free to fix them.
In the library itself (/src/datasets) it should be fine since there are tests and a windows CI, but if you have doubts of some usage of `os.path.join` somewhere, let me know.

lhoestq:
Alright create the test in #905 .
The windows CI is failing for all the datasets that have bad usage of `os.path.join` for urls.
There are of course the ones you fixed in this PR (thanks again !) but I found others as well such as pg19.
You can check the full list by looking at the CI failures

lhoestq:
I am merging this one as well as #906 that should fix all of the datasets.
Then I'll rebase #905 which adds the test that checks for bad urls and make sure it' all green now

PR #905, opened 2020-11-27
Disallow backslash in urls

lhoestq:
Following #903 @albertvillanova noticed that there are sometimes bad usage of `os.path.join` in datasets scripts to create URLS. However this should be avoided since it doesn't work on windows.

I'm suggesting a test to make sure we that all the urls don't have backslashes in them in the datasets scripts.
The tests works by adding a callback feature to the MockDownloadManager used to test the dataset scripts. In a download callback I just make sure that the url is valid.

### DIFF
--- a/src/datasets/utils/mock_download_manager.py
+++ b/src/datasets/utils/mock_download_manager.py
@@ -19,9 +19,11 @@
 import os
 import urllib.parse
 from pathlib import Path
+from typing import Callable, List, Optional, Union
 
 from .file_utils import cached_path, hf_github_url
 from .logging import get_logger
+from .version import Version
 
 
 logger = get_logger(__name__)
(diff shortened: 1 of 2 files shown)


lhoestq:
Looks like the test doesn't detect all the problems fixed by #907 , I'll fix that

lhoestq:
Ok found why it doesn't detect the problems fixed by #907 . That's because for all those datasets the urls are actually fine (no backslash) on windows, even if it uses `os.path.join`.

This is because of the behavior of `os.path.join` on windows when the first path ends with a slash : 

```python
import os
os.path.join("https://test.com/foo", "bar.txt")
# 'https://test.com/foo\\bar.txt'
os.path.join("https://test.com/foo/", "bar.txt")
# 'https://test.com/foo/bar.txt'
```

However even though the urls are correct, this is definitely bad practice and we should never use `os.path.join` for urls

PR #907, opened 2020-11-27
Remove os.path.join from all URLs

albertvillanova:
Remove `os.path.join` from all URLs in dataset scripts.


lhoestq (approved):
Thank you very much for fixing all these ones

PR #908, opened 2020-11-27
Add dependency on black for tests

albertvillanova:
Add package 'black' as an installation requirement for tests.


albertvillanova:
Sorry, I have just seen that it was already in `QUALITY_REQUIRE`.

For some reason it did not get installed on my virtual environment...

PR #914, opened 2020-11-29
Add list_github_datasets api for retrieving dataset name list in github repo

zhuzilin:
Thank you for your great effort on unifying data processing for NLP!

This pr is trying to add a new api `list_github_datasets` in the `inspect` module. The reason for it is that the current `list_datasets` api need to access https://huggingface.co/api/datasets to get a large json. However, this connection can be really slow... (I was visiting from China) and from my own experience, most of the time `requests.get` failed to download the whole json after a long wait and will trigger fault in `r.json()`.
I also noticed that the current implementation will first try to download from github, which makes me be able to smoothly run `load_dataset('squad')` in the example.
Therefore, I think it would be better if we can have an api to get the list of datasets that are available on github, and it will also improve newcomers' experience (it is a little frustrating if one cannot successfully run the first function in the README example.) before we have faster source for huggingface.co.

As for the implementation, I've added a `dataset_infos.json` file under the `datasets` folder, and it has the following structure:
```json
    {
        "id": "aeslc",
        "folder": "datasets/aeslc",
        "dataset_infos": "datasets/aeslc/dataset_infos.json"
    },
    ...
    {
        "id": "json",
        "folder": "datasets/json"
    },
   ...
```
The script I used to get this file is:
```python
import json
import os

DATASETS_BASE_DIR = "/root/datasets" 
DATASET_INFOS_JSON = "dataset_infos.json"

datasets = []
for item in os.listdir(os.path.join(DATASETS_BASE_DIR, "datasets")):
    if os.path.isdir(os.path.join(DATASETS_BASE_DIR, "datasets", item)):
        datasets.append(item)

datasets.sort()

total_ds_info = []
for ds in datasets:
    ds_dir = os.path.join("datasets", ds)
    ds_info_dir = os.path.join(ds_dir, DATASET_INFOS_JSON)
    if os.path.isfile(os.path.join(DATASETS_BASE_DIR, ds_info_dir)):
        total_ds_info.append({"id": ds,
                              "folder": ds_dir,
                              "dataset_infos": ds_info_dir})
    else:
        total_ds_info.append({"id": ds,
                              "folder": ds_dir})

with open(DATASET_INFOS_JSON, "w") as f:
    json.dump(total_ds_info, f)
```
The new `dataset_infos.json` was saved as a formated json so that it will be easy to add new dataset.

When calling `list_github_datasets`, the user will get the list of dataset names in this github repo and if `with_details` is set to be `True`, they can get the url of specific dataset info.

Thank you for your time on reviewing this pr :).

### DIFF
--- a/datasets/dataset_infos.json
+++ b/datasets/dataset_infos.json
@@ -1,0 +1,859 @@
+[
+    {
+        "id": "aeslc",
+        "folder": "datasets/aeslc",
+        "dataset_infos": "datasets/aeslc/dataset_infos.json"
+    },
+    {
+        "id": "ag_news",
+        "folder": "datasets/ag_news",
+        "dataset_infos": "datasets/ag_news/dataset_infos.json"
+    },
+    {
+        "id": "ai2_arc",
+        "folder": "datasets/ai2_arc",
+        "dataset_infos": "datasets/ai2_arc/dataset_infos.json"
+    },
+    {
+        "id": "allocine",
+        "folder": "datasets/allocine",
+        "dataset_infos": "datasets/allocine/dataset_infos.json"
+    },
+    {
+        "id": "amazon_us_reviews",
+        "folder": "datasets/amazon_us_reviews",
+        "dataset_infos": "datasets/amazon_us_reviews/dataset_infos.json"
+    },
+    {
+        "id": "anli",
+        "folder": "datasets/anli",
+        "dataset_infos": "datasets/anli/dataset_infos.json"
+    },
+    {
+        "id": "arcd",
+        "folder": "datasets/arcd",
+        "dataset_infos": "datasets/arcd/dataset_infos.json"
+    },
+    {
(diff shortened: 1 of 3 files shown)


julien-c:
We can look into removing some of the attributes from `GET /api/datasets` to make it smaller/faster, what do you think @lhoestq?

lhoestq:
> We can look into removing some of the attributes from `GET /api/datasets` to make it smaller/faster, what do you think @lhoestq?

yes at least remove all the `dummy_data.zip`

julien-c:
`GET /api/datasets` should now be much faster. @zhuzilin can you check if `list_datasets` is now faster for you?

zhuzilin:
> `GET /api/datasets` should now be much faster. @zhuzilin can you check if `list_datasets` is now faster for you?

Yes, much faster! Thank you!

PR #916, opened 2020-11-30
Add Swedish NER Corpus

abhishekkrthakur:


datasets/swedish_ner_corpus/swedish_ner_corpus.py
```diff
@@ -0,0 +1,74 @@
...
+                        sentence_counter,
+                        {
+                            "id": str(sentence_counter),
+                            "words": current_words,
+                            "labels": current_labels,
+                        },
+                    )
+                    sentence_counter += 1
+                    current_words = []
+                    current_labels = []
```
patrickvonplaten, line 67: I'd set a breakpoint here, just to be sure the code actually reaches `yield`

datasets/swedish_ner_corpus/swedish_ner_corpus.py
```diff
@@ -0,0 +1,74 @@
...
+                row = row.rstrip()
+                row_split = row.split("\t")
+                if len(row_split) > 1:
+                    token, label = row_split
+                    current_words.append(token)
+                    current_labels.append(label)
+                else:
+                    if not current_words:
+                        continue
+                    assert len(current_words) == len(current_labels)
```
patrickvonplaten, line 56: assert message would be nice

datasets/swedish_ner_corpus/swedish_ner_corpus.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def _generate_examples(self, datapath):
+        sentence_counter = 0
+        with open(datapath, encoding="utf-8") as f:
+            current_words = []
+            current_labels = []
+            for row in f:
+                row = row.rstrip()
+                row_split = row.split("\t")
+                if len(row_split) > 1:
```
patrickvonplaten, line 49: Think it's better to replace this with `if len(row_split) == 2` and then add a nice error message if `if len(row_split)` has an unexpected length

datasets/swedish_ner_corpus/swedish_ner_corpus.py
```diff
@@ -0,0 +1,101 @@
...
+
+_DESCRIPTION = """\
+Webbnyheter 2012 from Spraakbanken, semi-manually annotated and adapted for CoreNLP Swedish NER. Semi-manually defined in this case as: Bootstrapped from Swedish Gazetters then manually correcte/reviewed by two independent native speaking swedish annotators. No annotator agreement calculated.
+"""
+_HOMEPAGE_URL = "https://github.com/klintan/swedish-ner-corpus"
+_TRAIN_URL = "https://raw.githubusercontent.com/klintan/swedish-ner-corpus/master/train_corpus.txt"
+_TEST_URL = "https://raw.githubusercontent.com/klintan/swedish-ner-corpus/master/test_corpus.txt"
+
+
+class SwedishNERCorpusDataset(datasets.GeneratorBasedBuilder):
```
lhoestq, line 28: Could you rename it `SwedishNERCorpus` ?
The dataset class must be the camel case equivalent of `swedish_ner_corpus` which is the script name.

datasets/swedish_ner_corpus/swedish_ner_corpus.py
```diff
@@ -0,0 +1,101 @@
...
+    def _info(self):
+        return datasets.DatasetInfo(
+            description=_DESCRIPTION,
+            features=datasets.Features(
+                {
+                    "id": datasets.Value("string"),
+                    "tokens": datasets.Sequence(datasets.Value("string")),
+                    "ner_tags": datasets.Sequence(
+                        datasets.features.ClassLabel(names=["0", "LOC", "MISC", "ORG", "PER"])
+                    ),
```
lhoestq, line 40: perfect !

datasets/swedish_ner_corpus/swedish_ner_corpus.py
```diff
@@ -0,0 +1,101 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Lint as: python3
+import datasets
+
+
```
lhoestq, line 19: Could you add a `_CITATION` as well ?

lhoestq (commented):
Really cool ! Thanks for adding it :)

Could you also add a dataset card ? see the template guide https://github.com/huggingface/datasets/blob/master/templates/README_guide.md

patrickvonplaten (approved):
LGTM! @lhoestq - is it still ok if datasets have no config?

lhoestq:
Yes the use of configs is optional

datasets/swedish_ner_corpus/swedish_ner_corpus.py
abhishekkrthakur: there is none for this. so ill add none.

lhoestq: Oh I didn't know. Leave it that way then

yjernite:
@abhishekkrthakur we want to keep track of the information that is and isn't in the dataset cards so we're asking everyone to use the full template :) If there is some information in there that you really can't find or don't feel qualified to add, you can just leave the `[More Information Needed]` text

PR #920, opened 2020-11-30
add dream dataset

patil-suraj:
Adding Dream: a Dataset and for Dialogue-Based Reading Comprehension

More details:
https://dataset.org/dream/
https://github.com/nlpdata/dream


lhoestq (commented):
Awesome good job !

Could you also add a dataset card using the template guide here : https://github.com/huggingface/datasets/blob/master/templates/README_guide.md
If you can't fill some fields then just leave `[N/A]`

yjernite:
> Awesome good job !
> 
> Could you also add a dataset card using the template guide here : https://github.com/huggingface/datasets/blob/master/templates/README_guide.md
> If you can't fill some fields then just leave `[N/A]`

Quick amendment: `[N/A]` is for fields that are not relevant: if you can't find the information just leave `[More Information Needed]`

patil-suraj:
@lhoestq since datset cards are optional for this sprint I'll add those later. Good for merge.

lhoestq:
Indeed we only require the tags to be added now (the yaml part at the top of the dataset card).
Could you add them please ?
You can find more infos here : https://github.com/huggingface/datasets/blob/master/ADD_NEW_DATASET.md#tag-the-dataset-and-write-the-dataset-card

patil-suraj:
@lhoestq added tags, I'll fill rest of the info after current sprint :)

patil-suraj:
The tests are failing tests for other datasets, not this one.

patil-suraj:
@lhoestq  could you tell me why these tests are failing, they don't seem related to this PR.

datasets/dream/dream.py
```diff
@@ -0,0 +1,119 @@
...
+                {
+                    "id": datasets.Value("string"),
+                    "dialogue": datasets.features.Sequence(datasets.Value("string")),
+                    "questions": [
+                        {
+                            "question": datasets.Value("string"),
+                            "choice": datasets.features.Sequence(datasets.Value("string")),
+                            "answer": datasets.Value("string"),
+                        },
+                    ],
```
lhoestq, line 89: We can make this more practical to use by flattening the nested structure (like other datasets like coqa or com_qa)
```
question: datasets.Value("string"),
choices: datasets.Sequence(datasets.Value("string")),
answer: datasets.Value("string"),
```
this way we have one question per example in the dataset, which is convenient to train models.

What do you think ?

lhoestq (commented):
Clean !

I have one suggestion about the feature types. I think we can make them more practical and consistent with the other datasets.

datasets/dream/dream.py
patil-suraj: Makes sense !

lhoestq (approved):
Thanks !

PR #923, opened 2020-11-30
Add CC-100 dataset

### LINKED ISSUE #773: Adding CC-100: Monolingual Datasets from Web Crawl Data
## Adding a Dataset
- **Name:** CC-100: Monolingual Datasets from Web Crawl Data
- **Description:** https://twitter.com/alex_conneau/status/1321507120848625665
- **Paper:** https://arxiv.org/abs/1911.02116
- **Data:** http://data.statmt.org/cc-100/
- **Motivation:** A large scale multi-lingual language modeling dataset. Text is de-duplicated and filtered by how "Wikipedia-like" it is, hopefully helping avoid some of the worst parts of the common crawl.

Instructions to add a new dataset can be found [here](https://huggingface.co/docs/datasets/share_dataset.html).

albertvillanova:
Add CC-100.

Close #773


albertvillanova:
Hello @lhoestq, I would like just to ask you if it is OK that I include this feature 9f32ba1 in this PR or you would prefer to have it in a separate one.

lhoestq:
Hi ! Sure that would be valuable to support .xz files. Feel free to open a separate PR for this.
And feel free to create the first test case for extracting compressed files if you have some inspiration (maybe create test_file_utils.py ?). We can still spend more time on tests next week when the sprint is over though so don't spend too much time on it.

albertvillanova:
@lhoestq, DONE! ;) See PR #950.

lhoestq:
Thanks for adding support for `.xz` files :)

Feel free to rebase from master to include it in your PR

albertvillanova:
@lhoestq DONE; I have merged instead, to avoid changing the history of my public PR ;)

albertvillanova:
Hi @lhoestq, I would need that you generate the dataset_infos.json and the dummy data for this dataset with a bigger computer. Sorry, but my laptop did not succeed...

lhoestq:
Thanks for your work @albertvillanova 
We'll definitely look into it after this sprint :)

lhoestq:
Looks like #1456 added CC100 already.
The difference with your approach is that this implementation uses the `BuilderConfig` parameters to allow the creation of custom configs for all the languages, without having to specify them in the `BUILDER_CONFIGS` class attribute.
For example even if the dataset doesn't have a config for english already, you can still load the english CC100 with
```python

load_dataset("cc100", lang="en")
```

albertvillanova:
@lhoestq, oops!! I remember having assigned this dataset to me in the Google sheet, besides having mentioned the corresponding issue in the Pull Request... Nevermind! :)

lhoestq:
Yes indeed I can see that...
Sorry for noticing that only now 

The code of the other PR ended up being pretty close to yours though
If you want to add more details to the cc100 dataset card or in the script feel to do so, any addition is welcome

PR #926, opened 2020-11-30
add inquisitive

patil-suraj:
Adding inquisitive qg dataset

More info: https://github.com/wjko2/INQUISITIVE


datasets/inquisitive/inquisitive.py
```diff
@@ -0,0 +1,166 @@
...
+This type of question reflects a real-world scenario: if one has questions during reading, some of them are answered by the text later on, \
+the rest are not, but any of them would help further the reader’s understanding at the particular point when they asked it. \
+This resource could enable question generation models to simulate human-like curiosity and cognitive processing, which may open up a new realm of applications.
+"""
+
+_ARTICLES_URL = "https://github.com/wjko2/INQUISITIVE/raw/master/articles.tgz"
+_QUESTIONS_URL = "https://github.com/wjko2/INQUISITIVE/raw/master/questions.txt"
+
+DEV_ARTICLE_IDS = list(itertools.chain(range(1, 101), range(1051, 1101)))
+TEST_ARTICLE_IDS = list(itertools.chain(range(101, 151), range(501, 551), range(1101, 1151)))
```
patil-suraj, line 51: Here dev and test data are not provided as separate files, instead the repo lists these ids. So hardcoding them here.
Is this the right way ?

patil-suraj:
`dummy_data` right now contains all article files, keeping only the required articles for dummy data fails the dummy data test.
Any idea ?

datasets/inquisitive/inquisitive.py
lhoestq: yes it's fine :)

lhoestq:
> `dummy_data` right now contains all article files, keeping only the required articles for dummy data fails the dummy data test.
> Any idea ?

We should definitely find a way to make it work with only a few articles.

If it doesn't work right now for dummy data, I guess it's because it tries to load every single article file ?

If so, then maybe you can use `os.listdir` method to first check all the data files available in the path where the `articles.tgz` file is extracted. Then you can simply iter through the data files and depending on their ID, include them in the train or test set. With this method you should be able to have only a few articles files per split in the dummy data. Does that make sense ?

patil-suraj:
fixed! so the issue was, I was preparing `articles_ids` based on the number of files in articles dir, so for dummy data questions it was not able to load some articles due to incorrect ids and the test was failing

datasets/inquisitive_qg/inquisitive_qg.py
```diff
@@ -0,0 +1,161 @@
...
+        ]
+
+    def _generate_examples(self, articles_dir, questions_file, article_ids):
+        with open(questions_file, encoding="utf-8") as f:
+            questions_counter = 0
+            rows = f.readlines()
+            for i, row in enumerate(rows):
+                if i == 0:
+                    continue  # skip header line
+                row = row.replace("\n", "").strip()
```
patrickvonplaten, line 136: (curious) Are there multiple `"\n"` in `row`? Isn't `row.strip()` enough?

datasets/inquisitive_qg/inquisitive_qg.py
```diff
@@ -0,0 +1,161 @@
...
+    BUILDER_CONFIGS = [
+        InquisitiveQgConfig(name="plain_text", version=datasets.Version("1.0.0", ""), description="plain_text"),
+    ]
+
+    def _info(self):
+        return datasets.DatasetInfo(
+            description=_DESCRIPTION,
+            features=datasets.Features(
+                {
+                    "id": datasets.Value("string"),
```
patrickvonplaten, line 81: I'd put this as well as `"int32"` since it's an int anyways...

patrickvonplaten (approved):

aa0532525555 (approved):

aa0532525555 (approved):

aa0532525555 (approved):

aa0532525555 (approved):

datasets/inquisitive_qg/inquisitive_qg.py
```diff
@@ -0,0 +1,161 @@
...
+  title     = {Inquisitive Question Generation for High Level Text Comprehension},
+  booktitle = {Proceedings of EMNLP},
+  year      = {2020},
+}
+"""
+
+_DESCRIPTION = """\
+A dataset of about 20k questions that are elicited from readers as they naturally read through a document sentence by sentence. \
+Compared to existing datasets, INQUISITIVE questions target more towards high-level (semantic and discourse) comprehension of text. \
+Because these questions are generated while the readers are pro-cessing the information, the questions directly communicate gaps between \
```
lhoestq, line 39:
```suggestion
Because these questions are generated while the readers are processing the information, the questions directly communicate gaps between \
```

lhoestq (approved):
Nice ! Thank you for adding this one :)

PR #928, opened 2020-11-30
Add the Multilingual Amazon Reviews Corpus

joeddav:
- **Name:** Multilingual Amazon Reviews Corpus* (`amazon_reviews_multi`)
- **Description:** A collection of Amazon reviews in English, Japanese, German, French, Spanish and Chinese.
- **Paper:** https://arxiv.org/abs/2010.02573

### Checkbox

- [x] Create the dataset script `/datasets/my_dataset/my_dataset.py` using the template
- [x] Fill the `_DESCRIPTION` and `_CITATION` variables
- [x] Implement `_infos()`, `_split_generators()` and `_generate_examples()`
- [x] Make sure that the `BUILDER_CONFIGS` class attribute is filled with the different configurations of the dataset and that the `BUILDER_CONFIG_CLASS` is specified if there is a custom config class.
- [x] Generate the metadata file `dataset_infos.json` for all configurations
- [x] Generate the dummy data `dummy_data.zip` files to have the dataset script tested and that they don't weigh too much (<50KB)
- [x] Add the dataset card `README.md` using the template : fill the tags and the various paragraphs
- [x] Both tests for the real data and the dummy data pass.


datasets/amazon_reviews_multi/README.md
```diff
@@ -0,0 +1,235 @@
...
+  - en
+  es:
+  - es
+  fr:
+  - fr
+  ja:
+  - ja
+  zh:
+  - zh
+licenses: []
```
lhoestq, line 26: Maybe add an `other-amazon-conditions` license here ?
(it starts with `other` since it's not part of the licenses list we have)

datasets/amazon_reviews_multi/amazon_reviews_multi.py
```diff
@@ -0,0 +1,132 @@
...
+        )
+
+    def _split_generators(self, dl_manager):
+        train_urls = [_DOWNLOAD_URL.format(split="train", lang=lang) for lang in self.config.languages]
+        dev_urls = [_DOWNLOAD_URL.format(split="dev", lang=lang) for lang in self.config.languages]
+        test_urls = [_DOWNLOAD_URL.format(split="test", lang=lang) for lang in self.config.languages]
+
+        train_paths = [dl_manager.download_and_extract(url) for url in train_urls]
+        dev_paths = [dl_manager.download_and_extract(url) for url in dev_urls]
+        test_paths = [dl_manager.download_and_extract(url) for url in test_urls]
```
lhoestq, line 118: Just for information you can use 
```python
train_urls = [_DOWNLOAD_URL.format(split="train", lang=lang) for lang in self.config.languages]
train_paths = dl_manager.download_and_extract(train_urls)
```
directly and ouput a list of paths.
It can be quite convenient sometimes and it allows parallel downloads if there are many many files that need to be downloaded.

(here you can leave it this way though)

lhoestq (approved):
Awesome ! good job

feel free to merge if there's no more changes to be added

datasets/amazon_reviews_multi/amazon_reviews_multi.py
joeddav: Oh no that's way better, thanks

PR #930, opened 2020-11-30
Lambada

VictorSanh:
Added LAMBADA dataset.

A couple of points of attention (mostly because I am not sure)
- The training data are compressed in a .tar file inside the main tar.gz file. I had to manually un-tar the training file to access the examples.
- The dev and test splits don't have the `category` field so I put `None` by default.

Happy to make changes if it doesn't respect the guidelines!

Victor


datasets/lambada/lambada.py
```diff
@@ -0,0 +1,134 @@
...
+                gen_kwargs={
+                    "filepath": os.path.join(data_dir, "lambada_development_plain_text.txt"),
+                    "split": "dev",
+                },
+            ),
+        ]
+
+    def _generate_examples(self, filepath, split):
+        """ Yields examples. """
+        if split == "train":
```
patrickvonplaten, line 119: I think we usually want to extract in the `_split_generators` function. I think it's better to just hand over the `filepath` to `_generate_examples()` and then start with 

```python
if split == "train":
            recursion_pattern = f"{os.path.dirname(filepath)}/train-novels/*/*.txt"
            for idx, novel_path in enumerate(glob.iglob(recursion_pattern, recursive=True)):
```

VictorSanh: ok, let me move the un-tar in the `_split_generators`

patrickvonplaten (approved):

datasets/lambada/README.md
```diff
@@ -0,0 +1,181 @@
...
+- expert-generated
+source_datasets:
+- extended|bookcorpus
+size_categories:
+- 10K<n<100K
+licenses:
+- cc-by-4.0
+---
+
+# Dataset Card for [Dataset Name]
```
yjernite, line 22:
```suggestion
# Dataset Card for LAMBADA
```

yjernite (commented):
Looking good!

PR #933, opened 2020-11-30
Add NumerSense

joeddav:
Adds the NumerSense dataset
- Webpage/leaderboard: https://inklab.usc.edu/NumerSense/
- Paper: https://arxiv.org/abs/2005.00683
- Description: NumerSense is a new numerical commonsense reasoning probing task, with a diagnostic dataset consisting of 3,145 masked-word-prediction probes. Basically, it's a benchmark to see whether your MLM can figure out the right number in a fill-in-the-blank task based on commonsense knowledge (a bird has **two** legs)


datasets/numer_sense/README.md
```diff
@@ -0,0 +1,182 @@
...
+- crowdsourced
+languages:
+- en
+licenses:
+- mit
+multilinguality:
+- monolingual
+size_categories:
+- 10K<n<100K
+source_datasets: []
```
lhoestq, line 14: If the dataset use original data please use

```suggestion
source_datasets:
- original
```

datasets/numer_sense/README.md
```diff
@@ -0,0 +1,182 @@
...
+Each value of the training set consists of (1) a sentence with a number masked out with the `<mask>` token, and (2) a
+ground truth target value (both strings). The test sets include the masked sentences but without the ground truth
+values.
+
+### Data Splits
+
+The dataset includes the following pre-defined data splits:
+
+- A train set with >10K labeled examples (i.e. containing a ground truth value)
+- A core test set (`test_core`) with 1,132 examples (no ground truth provided)
```
lhoestq, line 94: could you make it explicit that we use the empty string `""` for no label for this split ?

lhoestq (approved):
LGTM !

Feel free to merge if you have more more changes to do

datasets/numer_sense/README.md
joeddav: It doesn't use original data, but the source dataset isn't in the library so @yjernite suggested leaving the tag blank and linking to the source dataset in the card below.

yjernite: Sorry that was a miscommunication: this should be `extended|other`

If you get pull the latest tagger, you can also enter the name for the other

PR #941, opened 2020-12-01
Add People's Daily NER dataset

JetRunner:


datasets/peoples_daily_ner/peoples_daily_ner.py
```diff
@@ -0,0 +1,129 @@
...
+# limitations under the License.
+
+# Lint as: python3
+"""Introduction to People's Daily Dataset"""
+
+import logging
+
+import datasets
+
+
```
lhoestq, line 23: is there a citation available somewhere for this dataset ?

lhoestq (commented):
LGTM thanks :)

Before we merge, could you add a dataset card ? see here for more info: https://github.com/huggingface/datasets/blob/master/ADD_NEW_DATASET.md#tag-the-dataset-and-write-the-dataset-card

Note that only the tags at the top of the dataset card are mandatory, if you feel like it's going to take too much time writing the rest to fill it all you can just skip the paragraphs

JetRunner:
> LGTM thanks :)
> 
> 
> 
> Before we merge, could you add a dataset card ? see here for more info: https://github.com/huggingface/datasets/blob/master/ADD_NEW_DATASET.md#tag-the-dataset-and-write-the-dataset-card
> 
> 
> 
> Note that only the tags at the top of the dataset card are mandatory, if you feel like it's going to take too much time writing the rest to fill it all you can just skip the paragraphs

Nope. I don't think there is a citation. Also, can I do the dataset card later (maybe in bulk)?

lhoestq:
We're doing one PR = one dataset to keep track of things. Feel free to add the tags later in this PR if you want to.
Also only the tags are required now, because we don't want people spending too much time on the cards

JetRunner:
added @lhoestq

lhoestq (approved):
LGTM thank you !

lhoestq:
Merging since the CI is fixed on master

PR #946, opened 2020-12-01
add PEC dataset

zhongpeixiang:
A persona-based empathetic conversation dataset published at EMNLP 2020.


datasets/pec/pec.py
```diff
@@ -0,0 +1,205 @@
...
+                                "context": context,
+                                "response_speaker": response_speaker,
+                                "response": response,
+                            }
+                            context_speakers = []
+                            context = []
+                            response_speaker = ""
+                            response = ""
+                            example_id += 1
+                    except (IndexError, KeyError):
```
lhoestq, line 202: Could you remove that if it was used for debugging purposes ?

datasets/pec/pec.py
```diff
@@ -0,0 +1,205 @@
...
+        return persona
+
+    def _split_generators(self, dl_manager):
+        """Returns SplitGenerators."""
+        # TODO: Downloads the data and defines the splits
+        # dl_manager is a datasets.download.DownloadManager that can be used to
+        # download and extract URLs
+        dl_dir = dl_manager.download_and_extract(_URL)
+        data_dir = os.path.join(dl_dir, "hf_pec")
+        print("The downloaded data has been saved to ", data_dir)
```
lhoestq, line 114:
```suggestion

```

datasets/pec/pec.py
```diff
@@ -0,0 +1,205 @@
...
+        else:
+            filepath = [filepath]
+
+        # create persona
+        if not hasattr(self, "persona_" + self.config.domain):
+            persona_paths = [os.path.join(os.path.dirname(fpath), "persona.txt") for fpath in filepath]
+            setattr(self, "persona_" + self.config.domain, self._load_persona(persona_paths))
+
+        persona = getattr(self, "persona_" + self.config.domain)
+        # print("Domain {0}: number of unique personas: {1}".format(self.config.domain, len(persona)))
```
lhoestq, line 171: Why not simply use this instead : 
```
persona_paths = [os.path.join(os.path.dirname(fpath), "persona.txt") for fpath in filepath]
persona = self._load_persona(persona_paths)
```
?

(or maybe I'm missing something)

lhoestq (commented):
Nice job ! This is super clean thanks

I left a few minor comments and I think after they're resolved we'll be ready to merge :)

datasets/pec/pec.py
zhongpeixiang: I dont want to re-create the `persona` for every split of the dataset since it is shared by all splits.

zhongpeixiang:
The checks failed again even if I didn't make any changes.

lhoestq:
you just need to rebase from master to fix the CI :)

datasets/pec/pec.py
lhoestq: Then feel free to generate the persona once in the `_split_generator` method (it's called once for all splits), and pass the persona object through the `gen_kwargs` that are passed to `_generate_examples `
This way the persona will be created only once and reused in all the splits.
Does that make sense ?

zhongpeixiang:
Sorry for the mess, I'm confused by the rebase and thus created a new branch.

PR #950, opened 2020-12-01
Support .xz file format

albertvillanova:
Add support to extract/uncompress files in .xz format.


lhoestq (approved):
Very clean thank you :)

I also like the use of fixtures. That's something we'll use more in the future imo

PR #952, opened 2020-12-01
Add orange sum

moussaKam:
Add OrangeSum a french abstractive summarization dataset. 

Paper: [BARThez: a Skilled Pretrained French Sequence-to-Sequence Model](https://arxiv.org/abs/2010.12321)


datasets/orange_sum/README.md
```diff
@@ -0,0 +1,171 @@
...
+  "source_datasets": [
+    "original"
+  ],
+  "size_categories": [
+    "10K<n<100K"
+  ],
+  "licenses": [
+    "unknown"
+  ]
+}
```
lhoestq, line 30: We actually expect yaml here, sorry if it was clear (even though the tagger allows to show both json and yaml)


```suggestion
annotations_creators:
- found
language_creators:
- found
languages:
- fr
licenses:
- unknown
multilinguality:
- monolingual
size_categories:
- 10K<n<100K
source_datasets:
- original
task_categories:
- conditional-text-generation
task_ids:
- summarization
```

lhoestq (approved):
Very nice job thank you !

we can merge after replacing json with yaml for the tags

PR #953, opened 2020-12-01
added health_fact dataset 

bhavitvyamalik:
Added dataset Explainable Fact-Checking for Public Health Claims (dataset_id: health_fact)


lhoestq (commented):
Cool thank you for adding this one !

I left a few comments about the usage of the `-1` value as a placeholder when labels are missing.

Also could you add the dataset card please ? you can find more info here : https://github.com/huggingface/datasets/blob/master/ADD_NEW_DATASET.md#tag-the-dataset-and-write-the-dataset-card
Only the yaml tags at the top are mandatory.
If you feel like adding the rest of the rest of the paragraphs take too much time you can just leave the fields as `[More Information Needed]`

bhavitvyamalik:
Hi @lhoestq,
Initially I tried int(-1) only in place of nan labels and missing values but I kept on getting this error ```pyarrow.lib.ArrowTypeError: Expected bytes, got a 'int' object``` maybe because I'm sending int values (-1) to objects which are string type

bhavitvyamalik:
Or should I stick to "-1" only wherever it sees nan labels or missing values in dataset and just before yielding I can put a check if label is "-1" then convert it to int(-1)? It'll ensure my other fields are string values ("-1") and labels are int values int(-1). Just checked and it's working fine. Let me know if it's correct, I'll push the code. Finishing my dataset card in another hour. Thanks!

lhoestq (approved):
Nice thank you very much !

PR #957, opened 2020-12-01
Isixhosa ner corpus

yvonnegitau:


lhoestq (approved):
Perfect !
thanks for adding these two datasets, african languages datasets are very valuable

PR #963, opened 2020-12-01
add CODAH dataset

patil-suraj:
Adding CODAH dataset.

More info:
https://github.com/Websail-NU/CODAH


datasets/codah/codah.py
```diff
@@ -0,0 +1,140 @@
...
+        ),
+        CodahConfig(
+            name="fold_2", version=datasets.Version("1.0.0"), description="Official CV split (fold_2)", fold="fold_2"
+        ),
+        CodahConfig(
+            name="fold_3", version=datasets.Version("1.0.0"), description="Official CV split (fold_3)", fold="fold_3"
+        ),
+        CodahConfig(
+            name="fold_4", version=datasets.Version("1.0.0"), description="Official CV split (fold_4)", fold="fold_4"
+        ),
```
patil-suraj, line 90: The dataset contains official CV folds, included those here along with full dataset

datasets/codah/codah.py
```diff
@@ -0,0 +1,140 @@
...
+    ]
+
+    def _info(self):
+        return datasets.DatasetInfo(
+            description=_DESCRIPTION,
+            features=datasets.Features(
+                {
+                    "id": datasets.Value("int32"),
+                    "question_category": datasets.Value("string"),
```
patil-suraj, line 99: since the goal of this dataset is not to classify question category , using type `string`. Wdyt @yjernite ?

lhoestq: I would still use ClassLabel but no strong opinion on that.

lhoestq (commented):
Looks pretty good !

Don't forget to add at least the yaml tags of the dataset card 
You can find more info here : https://github.com/huggingface/datasets/blob/master/ADD_NEW_DATASET.md#tag-the-dataset-and-write-the-dataset-card
Only the yaml tags are mandatory, if you feel like the rest of the paragraphs take too much time to add you can leave them as `[More Information Needed]`

datasets/codah/codah.py
yjernite: I like ClassLAbel if it's not too much of a pain to collect the label set: it's more self-explanatory.

lhoestq (approved):
LGTM ! thanks

PR #964, opened 2020-12-01
Adding the WebNLG dataset

yjernite:
This PR adds data from the WebNLG challenge, with one configuration per release and challenge iteration.

More information can be found [here](https://webnlg-challenge.loria.fr/)

Unfortunately, the data itself comes from a pretty large number of small XML files, so the dummy data ends up being quite large (8.4 MB even keeping only one example per file).


yjernite:
This is task is part of the GEM suite so will actually need a more complete dataset card. I'm taking a break for now though and will get back to it before merging :)

datasets/web_nlg/web_nlg.py
```diff
@@ -0,0 +1,247 @@
...
+            )
+            for spl, dir_suffix_list in _FILE_PATHS[self.config.name].items()
+        ]
+
+    def _generate_examples(self, filedirs):
+        """ Yields examples. """
+
+        id_ = 0
+        for xml_location in filedirs:
+            for xml_file in glob(pjoin(xml_location, "*.xml")):
```
lhoestq, line 242: Could you sort the files using `sorted` for examples ?
Since we want the dataset to always have the examples in the same order, then we want the script to always iter through file list in the same order, no matter the environment.

lhoestq (commented):
Nice ! good job on parsing those files

About the dummy data we can probably remove some unused dummy files.

For example since we're using `glob`, then we can just remove most of the files and only keep one per task no ? as long as there is at least one `xml_file ` that yields a few examples it's fine.

Let me know if that does the trick. I would like to keep the repo lightweight, or it will take ages to clone in the future for those with slow connections

PR #976, opened 2020-12-02
Arabic pos dialect

mcmillanmajora:
A README.md and loading script for the Arabic POS Dialect dataset. The README is missing the sections on personal information, biases, and limitations, as it would probably be better for those to be filled by someone who can read the contents of the dataset and is familiar with Arabic NLP.


datasets/arabic_pos_dialect/arabic_pos_dialect.py
```diff
@@ -0,0 +1,129 @@
...
+        super(ArabicPosDialectConfig, self).__init__(**kwargs)
+        assert dialect in _DIALECTS, ("Invalid dialect code: %s", dialect)
+        self.dialect = dialect
+
+
+class ArabicPosDialect(datasets.GeneratorBasedBuilder):
+    """POS-tagged Arabic tweets in four major dialects."""
+
+    VERSION = datasets.Version("1.1.0")
+
```
lhoestq, line 67:
```suggestion
    BUILDER_CONFIG_CLASS = ArabicPosDialectConfig
```
we need that to specify that we use a custom config

datasets/arabic_pos_dialect/arabic_pos_dialect.py
```diff
@@ -0,0 +1,129 @@
...
+            description=_DESCRIPTION,
+            features=datasets.Features(
+                {
+                    "Fold": datasets.Value("int32"),
+                    "SubFold": datasets.Value("string"),
+                    "SentID": datasets.Value("int32"),
+                    "Order": datasets.Value("int32"),
+                    "Word": datasets.Value("string"),
+                    "Segmentation": datasets.Value("string"),
+                    "POS": datasets.Value("string"),
```
lhoestq, line 90: We can make the features more practical to use IMO if use the sentences instead of the words.
For example we can use these features:
```python
features=datasets.Features(
    {
        "Fold": datasets.Value("int32"),
        "SubFold": datasets.Value("string"),
        "tokens": datasets.Sequence(datasets.Value("string")),
        "ner_tags": datasets.Sequence(datasets.ClassLabel(names=ner_labels_list)),
        "pos_tags": datasets.Sequence(datasets.ClassLabel(names=pos_labels_list)),
    }
)
```
This way it's easier to use to train a model and it is also consistent with the format of the other NER and POS datasets

What do you think ?

lhoestq (commented):
Thanks for adding this one !

I added a suggestion about the feature types and about grouping the words in sentences.
Let me know what you think

datasets/arabic_pos_dialect/arabic_pos_dialect.py
mcmillanmajora: `Segmentation` doesn't refer to NER tags. It refers to morphemes. So I could see the sentence version being:
```
features=datasets.Features(
    {
        "Fold": datasets.Value("int32"),
        "SubFold": datasets.Value("string"),
        "tokens": datasets.Sequence(datasets.Value("string")),
        "segments": datasets.Sequence(datasets.Value("string")),
        "pos_tags": datasets.Sequence(datasets.ClassLabel(names=pos_labels_list)),
    }
)
```
The POS tags also correspond to the segments rather than the words, so a single word can have multiple POS tags connected with `+` like `PROG_PART+V+PRON`. We could collect all the POS tag combinations in the dataset to make the ClassLabel list, but I think decoupling the tags would result in a loss of information as you wouldn't have the word boundaries anymore.

lhoestq:
looks like this PR includes changes about many other files than the oens for Araboc POS Dialect

Can you create a another branch and another PR please ?

mcmillanmajora:
Sorry! I'm not sure how I managed to do that. I'll make a new branch.

PR #995, opened 2020-12-02
added dataset circa

bhavitvyamalik:
Dataset Circa added. Only README.md and dataset card left


datasets/circa/circa.py
```diff
@@ -0,0 +1,154 @@
...
+    VERSION = datasets.Version("1.1.0")
+
+    def _info(self):
+        features = datasets.Features(
+            {
+                "context": datasets.Value("string"),
+                "question-X": datasets.Value("string"),
+                "canquestion-X": datasets.Value("string"),
+                "answer-Y": datasets.Value("string"),
+                "judgements": datasets.Value("string"),
```
lhoestq, line 63: We could make `judgements` a sequence of labels by splitting on the `#` if I understand correctly. What do you think ?

lhoestq (commented):
Thanks for adding this one !

I left one comment.

Also could you add the tags in the dataset card ? it's the yaml part at the top of the dataset card. You can find more info here : https://github.com/huggingface/datasets/blob/master/ADD_NEW_DATASET.md#tag-the-dataset-and-write-the-dataset-card
You can leave the paragraphs with `[More Information Needed]`

datasets/circa/circa.py
bhavitvyamalik: In their paper and in the dataset, ```judgement``` is a combination of results which 5 annotators marked separated by```#```. They use majority judgement as the gold standard, subject to at least three annotators making that judgement. I reckon they'll eventually need gold standard for the question-answer-pair-classification task. Hence, I marked its value as string and not sequence of labels.

lhoestq (approved):
Thanks for adding this one !

k125-ak (commented):
I will be taking action on all the spying going on on my phone

datasets/circa/circa.py
k125-ak: I will be taking action agony all of this spying that has been requested my a company that does not exists. AKA boyfriend. Which you should be checking into

k125-ak: > I will be taking action agony all of this spying that has been requested my a company that does not exists. AKA boyfriend. Which you should be checking into

thomwolf:
Blocked @k125-ak ;) Bye-bye

PR #1000, opened 2020-12-02
UM005: Urdu <> English Translation Dataset

abhishekkrthakur:
Adds Urdu-English dataset for machine translation: http://ufal.ms.mff.cuni.cz/umc/005-en-ur/


datasets/um005/um005.py
```diff
@@ -0,0 +1,137 @@
...
+_DESCRIPTION = """\
+UMC005 English-Urdu is a parallel corpus of texts in English and Urdu language with sentence alignments. The corpus can be used for experiments with statistical machine translation.
+
+The texts come from four different sources:
+- Quran
+- Bible
+- Penn Treebank (Wall Street Journal)
+- Emille corpus
+
+We provide the religious texts of Quran and Bible for direct download. Because of licensing reasons, Penn and Emille texts cannot be redistributed freely. However, if you already hold a license for the original corpora, we are able to provide scripts that will recreate our data on your disk. Our modifications include but are not limited to the following:
```
lhoestq, line 31:
```suggestion
The authors provide the religious texts of Quran and Bible for direct download. Because of licensing reasons, Penn and Emille texts cannot be redistributed freely. However, if you already hold a license for the original corpora, we are able to provide scripts that will recreate our data on your disk. Our modifications include but are not limited to the following:
```

datasets/um005/um005.py
```diff
@@ -0,0 +1,137 @@
...
+    DEFAULT_CONFIG_NAME = _ALL
+
+    def _info(self):
+        return datasets.DatasetInfo(
+            description=_DESCRIPTION,
+            features=datasets.Features(
+                {
+                    "id": datasets.Value("string"),
+                    "urdu": datasets.Value("string"),
+                    "english": datasets.Value("string"),
```
lhoestq, line 91: Could you use the `Translation` feature type ?
see [flores](https://github.com/huggingface/datasets/blob/master/datasets/flores/flores.py) for an example of usage

lhoestq (commented):
Looks good thanks !

I left a comment about changing the feature type to `Translation`

datasets/um005/um005.py
```diff
@@ -0,0 +1,137 @@
...
+        self.sources = sources
+
+    @property
+    def filepaths(self):
+        return {source: _SOURCES_FILEPATHS[source] for source in self.sources}
+
+
+class UM005(datasets.GeneratorBasedBuilder):
+    BUILDER_CONFIGS = [
+        UM005Config(name=source, sources=[source], description=f"Source. {source}.") for source in _SOURCES
```
patrickvonplaten, line 79: nit - I'd just write out the three configs here. So 
```python
BUILDER_CONFIGS = [
        UM005Config(name="bible", sources=["bible'], description=f"Source. bible"), 
...
]
```

patrickvonplaten: And I guess the description could be a bit more specific for each config

datasets/um005/um005.py
```diff
@@ -0,0 +1,137 @@
...
+            homepage=_HOMEPAGE_URL,
+            citation=_CITATION,
+        )
+
+    def _split_generators(self, dl_manager):
+        path = dl_manager.download_and_extract(_URL)
+        return [
+            datasets.SplitGenerator(
+                name=datasets.Split.TRAIN,
+                gen_kwargs={"datapath": path, "datatype": "train"},
```
patrickvonplaten, line 104: This can be improved IMO. If I understand correctly each config has the same file names and each split has exactly two filepaths. I'd just pass the two filepaths here directly

datasets/um005/um005.py
```diff
@@ -0,0 +1,137 @@
...
+institution = {Univerzita Karlova},
+address     = {Praha, Czechia},
+issn        = {0032-6585},
+}
+"""
+
+_ALL = "all"
+_VERSION = "1.0.0"
+_SOURCES = ["bible", "quran"]
+_SOURCES_FILEPATHS = {
```
patrickvonplaten, line 57: If the source filepaths are all the same - I think there is no need for a dict with three keys that all have the same values no?

datasets/um005/um005.py
```diff
@@ -0,0 +1,137 @@
...
+            ),
+        ]
+
+    def _generate_examples(self, datapath, datatype):
+        sentence_counter = 0
+        for source, filepath in self.config.filepaths.items():
+            filepath = filepath[datatype]
+            urdu_file = filepath["urdu"]
+            english_file = filepath["english"]
+            urdu_path = os.path.join(datapath, source, urdu_file)
```
patrickvonplaten, line 122: would be nice if the "path" creation could be done in `_split_generators`. I'd just pass the two filepaths or two lists of filepaths if '"all"' is needed.

datasets/um005/um005.py
```diff
@@ -0,0 +1,137 @@
...
+year        = {2011},
+journal     = {The Prague Bulletin of Mathematical Linguistics},
+number      = {95},
+institution = {Univerzita Karlova},
+address     = {Praha, Czechia},
+issn        = {0032-6585},
+}
+"""
+
+_ALL = "all"
```
patrickvonplaten, line 54: Do we have to have a config that concatenates all other datasets? Don't we want to let the user do this in case he want's both configs? Not sure what the rule is here

patrickvonplaten (approved):
Left a couple of nits. Think more path logic could be moved to the `_split_generators`, but it's not that important I think

datasets/um005/um005.py
lhoestq: It depends on the dataset. If it is a common usage of the dataset to concatenate everything then we should have a config for it. Otherwise we should just keep the separate configs

datasets/um005/um005.py
lhoestq: what do you think ?

abhishekkrthakur: Will fix at some point today. Got distracted by more interesting datasets :)

lhoestq (approved):
Thanks LGTM :)

PR #1007, opened 2020-12-02
Include license file in source distribution

synapticarbors:
It would be helpful to include the license file in the source distribution.


lhoestq (approved):
right !

PR #1008, opened 2020-12-02
Adding C3 dataset: the first free-form multiple-Choice Chinese machine reading Comprehension dataset.      https://github.com/nlpdata/c3 https://arxiv.org/abs/1904.09679

Narsil:


Narsil:
Dupe of #1009

PR #1017, opened 2020-12-02
Specify file encoding

albertvillanova:
If not specified, Python uses system default, which for Windows is not "utf-8".


yjernite:
Thanks!

PR #1022, opened 2020-12-02
add MRQA

VictorSanh:
MRQA (shared task 2019)
out of distribution generalization
Framed as extractive question answering
Dataset is the concatenation (of subsets) of existing QA datasets processed to match the SQuAD format


lhoestq (commented):
Nice ! good job :)
Looks good to me

Could you just regenerate the dummy data but with the `--n_lines 1` flags ?
This will make them lighter since they're quire heavy right now (almost 200KB).
It will keep one example per data file.

VictorSanh:
THanks!
Done!

PR #1033, opened 2020-12-03
Add support for ".txm" format

albertvillanova:
In dummy data generation, add support for XML-like ".txm" file format.

Also support filenames with additional compression extension: ".txm.gz".


src/datasets/commands/dummy_data.py
```diff
@@ -112,8 +112,9 @@ def _create_dummy_data(
         encoding = encoding or DEFAULT_ENCODING
         if os.path.isfile(src_path):
             logger.debug(f"Trying to generate dummy data file {dst_path}")
+            dst_path_extension = get_main_extension(dst_path)
```
lhoestq, line 115: I think you can simply use pathlib here
```python
import pathlib

print(pathlib.Path('yourPath.example').suffix) # '.example'
```

lhoestq (commented):
Thank you for adding support for tmx files :)

I think we can simply use pathlib instead of adding yet another util function. Could you change that ?

src/datasets/commands/dummy_data.py
albertvillanova: @lhoestq yes, you are right!! Not exactly .suffix, but .suffixes... I am changing this. ;)

albertvillanova: @lhoestq note that in the case "file.xml.gz" I want to consider the first extension ".xml" and not the last one.

src/datasets/commands/dummy_data.py
```diff
@@ -112,8 +112,9 @@ def _create_dummy_data(
         encoding = encoding or DEFAULT_ENCODING
         if os.path.isfile(src_path):
             logger.debug(f"Trying to generate dummy data file {dst_path}")
+            dst_path_extension = Path(dst_path).suffixes[0] if Path(dst_path).suffixes else ""
             line_by_line_extensions = [".txt", ".csv", ".jsonl", ".tsv"]
-            is_line_by_line_text_file = any(dst_path.endswith(extension) for extension in line_by_line_extensions)
+            is_line_by_line_text_file = dst_path_extension in line_by_line_extensions
```
yjernite, line 117: Not sure exactly how `Path` gets the suffixes, but this feels like it might fail for people who use dots in their file names (e.g. would `train.en.jsonl.tar.gs` give you `en` or 'jsonl'?)

Thoughts @albertvillanova? Any reasons not to check every suffix?

albertvillanova: @yjernite you are right: `train.en.jsonl.tar.gz` would give `en`, which is wrong. Therefore, I am implementing the check for every suffix.

yjernite:
Neat! Looks like you need a rebase and then should be good to go :)

albertvillanova:
Done, @yjernite, @lhoestq.

albertvillanova:
If you agree, we could merge this.

lhoestq:
Hi ! yes sure :) can you just merge master into this branch before we merge ?

albertvillanova:
Done @lhoestq

lhoestq (approved):
Thanks !

PR #1035, opened 2020-12-03
add wiki_hop

patil-suraj:
This PR adds the WikiHop dataset from the QAngaroo multi hop reading comprehension datasets

More info:
http://qangaroo.cs.ucl.ac.uk/index.html


datasets/wiki_hop/wiki_hop.py
```diff
@@ -0,0 +1,118 @@
...
+            name="original",
+            version=datasets.Version("1.0.0"),
+            description="The un-maksed WikiHop dataset",
+            masked=False,
+        ),
+        WikiHopConifg(
+            name="masked", version=datasets.Version("1.0.0"), description="Masked WikiHop dataset", masked=True
+        ),
+    ]
+
```
lhoestq, line 73:
```suggestion
    BUILDER_CONFIG_CLASS = WikiHopConfig
```

datasets/wiki_hop/wiki_hop.py
```diff
@@ -0,0 +1,118 @@
...
+    VERSION = datasets.Version("1.0.0")
+    BUILDER_CONFIGS = [
+        WikiHopConifg(
+            name="original",
+            version=datasets.Version("1.0.0"),
+            description="The un-maksed WikiHop dataset",
+            masked=False,
+        ),
+        WikiHopConifg(
```
lhoestq, line 69:
```suggestion
        WikiHopConfig(
```

datasets/wiki_hop/wiki_hop.py
```diff
@@ -0,0 +1,118 @@
...
+        super(WikiHopConifg, self).__init__(**kwargs)
+        self.masked = masked
+
+
+class WikiHop(datasets.GeneratorBasedBuilder):
+    """WikiHop: Reading Comprehension with Multiple Hops"""
+
+    VERSION = datasets.Version("1.0.0")
+    BUILDER_CONFIGS = [
+        WikiHopConifg(
```
lhoestq, line 63:
```suggestion
        WikiHopConfig(
```

datasets/wiki_hop/wiki_hop.py
```diff
@@ -0,0 +1,118 @@
...
+
+_DESCRIPTION = """\
+WikiHop is open-domain and based on Wikipedia articles; the goal is to recover Wikidata information by hopping through documents. \
+The goal is to answer text understanding queries by combining multiple facts that are spread across different documents.
+"""
+
+_URL = "https://drive.google.com/uc?export=download&id=1ytVZ4AhubFDOEL7o7XrIRIyhU8g9wvKA"
+
+
+class WikiHopConifg(datasets.BuilderConfig):
```
lhoestq, line 44:
```suggestion
class WikiHopConfig(datasets.BuilderConfig):
```

datasets/wiki_hop/wiki_hop.py
```diff
@@ -0,0 +1,118 @@
...
+                gen_kwargs={"filepath": os.path.join(wikihop_path, dev_file), "split": "dev"},
+            ),
+        ]
+
+    def _generate_examples(self, filepath, split):
+        with open(filepath, encoding="utf-8") as f:
+            examples = json.load(f)
+            for i, example in enumerate(examples):
+                if split == "train":
+                    example["annotations"] = []
```
lhoestq, line 117: can you explain with a comment why you need to do this ?

datasets/wiki_hop/wiki_hop.py
```diff
@@ -0,0 +1,118 @@
...
+
+    DEFAULT_CONFIG_NAME = "original"
+
+    def _info(self):
+        return datasets.DatasetInfo(
+            description=_DESCRIPTION,
+            features=datasets.Features(
+                {
+                    "id": datasets.Value("string"),
+                    "query": datasets.Value("string"),
```
lhoestq, line 82: could this be named `question` instead for consistency with other QA datasets ?

lhoestq (commented):
Awesome thank you !

I left a few comments

lhoestq:
Also the dummy data files are quite big (500KB)
If you could reduce that that would be nice (just look at the files inside and remove unecessary chunks of texts)
in general dummy data are just a few KB and we suggest to not get higher than 50KB

Having light dummy data makes the repo faster to clone

patil-suraj:
> Also the dummy data files are quite big (500KB)
> If you could reduce that that would be nice (just look at the files inside and remove unecessary chunks of texts)
> in general dummy data are just a few KB and we suggest to not get higher than 50KB
> 
> Having light dummy data makes the repo faster to clone

@lhoestq  Actually the downloaded file contains both the WikiHop and MedHop dataset, which is why the dummy data is big. We could only keep the required dataset to lower the size and delete the other one.

datasets/wiki_hop/wiki_hop.py
patil-suraj: `query` seems better suited here since these are not exactly questions but are small queries. But no strong feeling about that.

lhoestq: what do you think ?

patil-suraj: Will name it `question`, wont' be an issue.

lhoestq (approved):
Awesome thanks !

PR #1037, opened 2020-12-03
Fix docs indentation issues

albertvillanova:
Replace tabs with spaces.


lhoestq:
is this an issue ?

albertvillanova:
Yes @lhoestq, look at the docs site. For example, in https://huggingface.co/docs/datasets/add_dataset.html, look at the indentation in the code block under the sentence:
> Here are the features of the SQuAD dataset for instance, which is taken from the squad dataset loading script:

lhoestq (approved):
Indeed, thanks

PR #1040, opened 2020-12-03
Add UN Universal Declaration of Human Rights (UDHR)

joeddav:
Universal declaration of human rights with translations in 464 languages and dialects.

- UN page: https://www.ohchr.org/EN/UDHR/Pages/UDHRIndex.aspx
- Raw data source: https://unicode.org/udhr/index.html

Each instance of the dataset corresponds to one translation of the document. Since there's only one instance per language (and because there are 500 languages so the dummy data would be messy), I opted to just include them all under the same single config. I wasn't able to find any kind of license so I just copied the copyright notice.

I was pretty careful careful generating the language tags so they _should_ all be correct & consistent BCP-47 codes per the docs.


datasets/udhr/udhr.py
```diff
@@ -0,0 +1,555 @@
...
+    def _split_generators(self, dl_manager):
+        xml_dir = dl_manager.download_and_extract(_XML_DOWNLOAD_URL)
+        txt_dir = dl_manager.download_and_extract(_TXT_DOWNLOAD_URL)
+
+        return [
+            datasets.SplitGenerator(name=datasets.Split.TRAIN, gen_kwargs={"xml_dir": xml_dir, "txt_dir": txt_dir}),
+        ]
+
+    def _generate_examples(self, xml_dir, txt_dir):
+        for ilang, lang in enumerate(_LANGUAGES):
```
lhoestq, line 540: There's a trick we can use here to reduce the size of the dummy data.
If we use `glob.glob` or `os.listdir` here instead of iterating through all the languages we can make this script compatible with dummy data that only contain 1 file.

note : when using glob or listdir make sure to always apply `sorted` to make sure the order is deterministic no matter the environment

lhoestq (commented):
Impressive job !
You did right to have one config for the whole dataset.

I suggested a trick to have the dummy data lighter. Let me know what you think

datasets/udhr/udhr.py
joeddav: Oh ok that's a good idea. In that case I don't actually even need the massive `_LANGUAGES` list in the loading script at all. Do you think it's still worth having there for documentation purposes or just remove it?

PR #1042, opened 2020-12-03
Add Big Patent dataset

mattbui:
- More info on the dataset: https://evasharma.github.io/bigpatent/
- There's another raw version of the dataset available from tfds. However, they're quite large so I don't have the resources to fully test all the configs for that version yet. We'll try to add it later.
- ~Currently, there are no dummy data for this dataset yet as I'm facing some problems with generating them. I'm trying to add them later.~


lhoestq:
Looks like this PR include changes about many other files than the ones related to big patent.
Could you create another branch and another PR ?

mattbui:
@lhoestq Just created a new PR here: https://github.com/huggingface/datasets/pull/1087

PR #1043, opened 2020-12-03
Add TSAC: Tunisian Sentiment Analysis Corpus

abhishekkrthakur:
github: https://github.com/fbougares/TSAC

paper: https://www.aclweb.org/anthology/W17-1307/


datasets/tsac/tsac.py
```diff
@@ -0,0 +1,114 @@
...
+            supervised_keys=None,
+            homepage=_HOMEPAGE_URL,
+            citation=_CITATION,
+        )
+
+    def _split_generators(self, dl_manager):
+        train_pos_path = dl_manager.download_and_extract(_TRAIN_POS_URL)
+        train_neg_path = dl_manager.download_and_extract(_TRAIN_NEG_URL)
+        test_pos_path = dl_manager.download_and_extract(_TEST_POS_URL)
+        test_neg_path = dl_manager.download_and_extract(_TEST_NEG_URL)
```
lhoestq, line 73: just for info you can use a dict like 
```python
{
    "train_pos": _TRAIN_POS_URL,
    "train_neg": _TRAIN_NEG_URL,
    "test_pos": _TEST_POS_URL,
    "test_neg": _TEST_NEG_URL
}
```
and pass it to the `dl_manager.download_and_extract` function, which returns the dictionary of paths as expected.

This is useful when you have multiple downloads since they can be run in parallel

lhoestq (approved):
LGTM !

Feel free to merge unless you want to change the downloads as proposed

datasets/tsac/tsac.py
abhishekkrthakur: Will keep it this way since the dataset is small and take care of this in future for large datasets. merging.

lhoestq: sounds good to me

PR #1053, opened 2020-12-03
Fix dataset URL and file names, and add column name in "Social Bias Frames" dataset

otakumesi:
# Why I did
When I use "social_bias_frames" datasets in this library, I got 404 Errors.
So, I fixed this error and another some problems that I faced to use the dataset.

# What I did
* Modify this dataset URL
* Modify this dataset file names
* Add a "dataSource" column

Thank you!


yjernite:
Thanks a lot, looks good!

PR #1058, opened 2020-12-03
added paws-x dataset

bhavitvyamalik:
Added paws-x dataset. Updating README and tags in the dataset card in a while


datasets/paws-x/paws-x.py
```diff
@@ -0,0 +1,175 @@
...
+        for config_name in _DATA_OPTIONS
+    ]
+
+    DEFAULT_CONFIG_NAME = "en"  # It's not mandatory to have a default configuration. Just use one if it make sense.
+
+    def _info(self):
+        features = datasets.Features(
+            {
+                "id": datasets.Value("int32"),
+                "sentences": datasets.features.Sequence({"text": datasets.Value("string")}),
```
lhoestq, line 99: could you use two columns `sentence1` and `sentence2` ? 
we're already doing this for other paws-x

datasets/paws-x/paws-x.py
```diff
@@ -0,0 +1,175 @@
...
+
+    BUILDER_CONFIGS = [
+        PAWSXConfig(
+            name=config_name,
+            description=(f"This config contains samples in {config_name}."),
+        )
+        for config_name in _DATA_OPTIONS
+    ]
+
+    DEFAULT_CONFIG_NAME = "en"  # It's not mandatory to have a default configuration. Just use one if it make sense.
```
lhoestq, line 93:
```suggestion
```
I'm not sure it makes sense to put one language forward compared to the others

lhoestq (commented):
Nice ! Thank you for adding this dataset :) 

I left a few comments

Also could you make the dummy_data.zip files a bit lighter ? 
To do so you can for example take a look inside the dummy_data.zip file of `en` and remove all the files that don't belong to the `en` configuration

datasets/paws-x/paws-x.py
bhavitvyamalik: Something like this you mean?
```
features = datasets.Features(
            {
                "id": datasets.Value("int32"),
                "sentence1": datasets.Value("string"),
                "sentence2": datasets.Value("string"),
                "label": datasets.features.ClassLabel(names=["0", "1"]),
            }
        )
```

datasets/paws-x/paws-x.py
bhavitvyamalik: Sure, I'll make that change. Let the user decide what they want

datasets/paws-x/README.md
```diff
@@ -0,0 +1,206 @@
...
+languages:
+- de
+- en
+- es
+- fr
+- ja
+- ko
+- zh
+licenses:
+- other-The dataset may be freely used for any purpose, although acknowledgement of
```
yjernite, line 17: This should only be `other`, the free text description needs to be in the paragraph at the end of the file

bhavitvyamalik: Hi @yjernite,
I made that change in README and when I pushed the code few tests failed. Can you please look into it? Thanks!

lhoestq (approved):
Thanks ! LGTM :)

PR #1068, opened 2020-12-03
Add Pubmed (citation + abstract) dataset (2020).

Narsil:


yjernite:
LGTM! ftp addition looks fine but maybe have a look @thomwolf ?

Narsil:
It's not finished yet, I need to run the tests on the full dataset (it was running this weekend, there is an error somewhere deep)

Narsil:
@yjernite Ready for review !
@thomwolf 

So I tried to follow closely the original format that means I still had to drop information (namely tags on elements are simply discarded for now but they don't seem to carry critical information).
Some elements are also discarded they tend to not come up often:
 - The most notable is Author affiliation, which seems to be all over the place in terms of what it look meaning it's hard to actually get a consistent format.
 - Journal is the same, all the elements in there can be wildly different so I decided to drop it for now instead of trying to figure out a way to have a common presentation. (the DOI and medline ID are present so it can be recovered).

I think this PR could go as it but we probably should add a way to get easier information to use with a config.
For instance `{"title": "string", "abstract": "string", "authors": List[str], "substances": List[str]}` maybe ? (substances for instance is a tricky one, some substances only have an international identifier, some have simply a common name, some both)

It's relatively easy to do I think it's mostly discarding other fields and renaming some deep structure into a flat one.

thomwolf:
Look ok to me but @lhoestq is the master on the Download Manager side

src/datasets/utils/file_utils.py
```diff
@@ -466,6 +474,10 @@ def get_from_cache(
 
     # We don't have the file locally or we need an eTag
     if not local_files_only:
+        if url.startswith("ftp://"):
+            with closing(urllib.request.urlopen(url)) as r:
```
lhoestq, line 478: Would it be possible to raise a `ConnectionError` here if the `urlopen` fails ?
Currently a bad ftp url raises raises an `URLError`

Can you also add a timeout ? (ex: 2secs)

src/datasets/utils/file_utils.py
```diff
@@ -386,6 +387,13 @@ def get_datasets_user_agent(user_agent: Optional[Union[str, dict]] = None) -> st
     return ua
 
 
+def ftp_get(url, temp_file, proxies=None, resume_size=0, user_agent=None, cookies=None):
+    logger.info(f"Getting through FTP {url} into {temp_file.name}")
+    print(f"Getting through FTP {url} into {temp_file.name}")
+    with closing(urllib.request.urlopen(url)) as r:
```
lhoestq, line 393: same here

src/datasets/utils/file_utils.py
```diff
@@ -386,6 +387,13 @@ def get_datasets_user_agent(user_agent: Optional[Union[str, dict]] = None) -> st
     return ua
 
 
+def ftp_get(url, temp_file, proxies=None, resume_size=0, user_agent=None, cookies=None):
+    logger.info(f"Getting through FTP {url} into {temp_file.name}")
+    print(f"Getting through FTP {url} into {temp_file.name}")
```
lhoestq, line 392: can you remove this print ?

datasets/pubmed/README.md
```diff
@@ -0,0 +1,197 @@
...
+            "MeshHeading": [],
+        },
+    },
+    "PubmedData": {
+        "ArticleIdList": {"ArticleId": "10.1002/bjs.1800650203"},
+        "PublicationStatus": "ppublish",
+        "History": {"PubMedPubDate": [{"Year": 0, "Month": 0, "Day": 0}]},
+    },
+}
+```
```
lhoestq, line 125: it doesn't look like a valid structure here, can you fix that ?

In particular we can see a closing parenthesis alone at `{"PublicationType": [])` and a dictionary with brackets at `[
            "RegistryNumber": "XX",
            "NameOfSubstance": "Methanol"
        ]`

datasets/pubmed/pubmed.py
```diff
@@ -0,0 +1,399 @@
...
+"""
+
+_HOMEPAGE = "https://www.nlm.nih.gov/databases/download/pubmed_medline.html"
+
+_LICENSE = ""
+
+# TODO: Add link to the official dataset URLs here
+# The HuggingFace dataset library don't host the datasets but only point to the original files
+# This can be an arbitrary nested dict/list of URLs (see below in `_split_generators` method)
+_URLs = [f"ftp://ftp.ncbi.nlm.nih.gov/pubmed/baseline/pubmed21n{i:04d}.xml.gz" for i in range(1, 1063)]
```
lhoestq, line 45: Because there are 1000+ urls here, then currently 1000+ files are expected to be in the dummy_data.zip file as well. This is a bit overkill imo and it makes the zip file unnecessarily big (1MB+).

There is a way to only have 1 file in the dummy zip file used for tests when urls are formatted in a certain way.
Currently it only works for shards formatted using the tfrecords convention (for example `data.txt.0001-of-0300`) but feel free to extend it so that it also supports the urls formats that you have here.

You can see the current regex here: https://github.com/huggingface/datasets/blob/master/src/datasets/utils/mock_download_manager.py#L173-L175

Basically if the input of the dl_manager is a list of url that matches a regex, then during the dummy data tests the output is going to be a list of path to one file (instead of 1000+ different files).

datasets/pubmed/pubmed.py
```diff
@@ -0,0 +1,399 @@
...
+                        except Exception:
+                            continue
+                        citations.append(citation)
+        article["PubmedData"]["ReferenceList"] = citations
+
+    def _generate_examples(self, filenames):
+        """ Yields examples. """
+        id_ = 0
+        for filename in filenames:
+            print(filename)
```
lhoestq, line 374: can you remove this print ?

datasets/pubmed/pubmed.py
```diff
@@ -0,0 +1,399 @@
...
+    """Deep update target dict with src
+    For each k,v in src: if k doesn't exist in target, it is deep copied from
+    src to target. Otherwise, if v is a list, target[k] is extended with
+    src[k]. If v is a set, target[k] is updated with v, If v is a dict,
+    recursively deep-update it.
+
+    Examples:
+    >>> t = {'name': 'Ferry', 'hobbies': ['programming', 'sci-fi']}
+    >>> deepupdate(t, {'hobbies': ['gaming']})
+    >>> print t
```
lhoestq, line 60: oh no python 2

datasets/pubmed/pubmed.py
```diff
@@ -0,0 +1,399 @@
...
+                continue
+
+            for article in xmldict["PubmedArticleSet"]["PubmedArticle"]:
+                self.update_citation(article)
+                new_article = default_article()
+
+                try:
+                    deepupdate(new_article, article)
+                except Exception:
+                    logger.error(f"Ignoring article {article}, it is malformed")
```
lhoestq, line 390: Maybe use warning as well here ?

Showing an error should make the user take an action to fix it but it looks like there's not must we can do.

On the other side warnings are messages that the user may want to know about because it can affect the behavior of the program, while not requiring any action to be taken if the user is fine with it.

lhoestq (commented):
Really cool thank you !
Good job with the parsing and also for generating all the metadata.

Looks like there are some conflicts with some recent changes regarding file_utils.py, can you resolve them ?

I also left a few comments

src/datasets/utils/file_utils.py
Narsil: Perfect. I extracted this as a function (like `ftp_get`).

src/datasets/utils/file_utils.py
Narsil: Done.

src/datasets/utils/file_utils.py
Narsil: Done. ftp_head returns a boolean whereas  ftp_get raises an Exception, I'm trying to follow the common code with Http as much as possible which I think uses the same path.

datasets/pubmed/README.md
Narsil: It's supposed to be a list of dictionaries.

datasets/pubmed/pubmed.py
Narsil: Done

datasets/pubmed/pubmed.py
Narsil: Edited for Py3 version

datasets/pubmed/pubmed.py
Narsil: Good point !

datasets/pubmed/pubmed.py
Narsil: Ok.
I have done it with a similar approach though it might seem fair to have a different hook if this becomes regular :)

lhoestq (approved):
Thanks !

PR #1087, opened 2020-12-04
Add Big Patent dataset

mattbui:
* More info on the dataset: https://evasharma.github.io/bigpatent/
* There's another raw version of the dataset available from tfds. However, they're quite large so I don't have the resources to fully test all the configs for that version yet. We'll try to add it later.


lhoestq (commented):
Very cool !

Could you also add the tags of the dataset card ? It's the yaml part at the top of the dataset card. More info here : https://github.com/huggingface/datasets/blob/master/ADD_NEW_DATASET.md#tag-the-dataset-and-write-the-dataset-card

Also it looks like the dummy data files are very heavy (50MB)
Could you try to make them just a few KB ?

To do so feel free to take a look at the files inside and remove unnecessary files or unnecessary chunks of texts

mattbui:
@lhoestq reduced the dummy data size to around 19MB in total and added the dataset card.

datasets/big_patent/big_patent.py
```diff
@@ -0,0 +1,156 @@
...
+                name=datasets.Split.TEST,
+                gen_kwargs={"path": extract_paths["test"]},
+            ),
+        ]
+
+    def _generate_examples(self, path=None):
+        """Yields examples."""
+        for cpc_code in self.config.cpc_codes:
+            filenames = glob.glob(os.path.join(path, cpc_code, "*"))
+            for filename in filenames:
```
lhoestq, line 148:
```suggestion
            for filename in sorted(filenames):
```
we need to sort here to make the order deterministic across environments

lhoestq (commented):
Thanks !

19MB is still too much though, we aim for only a few KB

Since you're using `glob` you cam remove many files and only keep 1 file per split.
You can also make sure that the remaining files are not too heavy by removing unnecessary chunks of texts

mattbui:
@lhoestq so I ended up removing all the nested JSON objects in the gz datafile and keep only one object with minimal content: `{"publication_number": "US-8230922-B2", "abstract": "dummy abstract", "application_number": "US-201113163519-A", "description": "dummy description"}`. 

They're reduced to 35KB in total (2KB per domain and 17KB for all domain), hopefully, it's small enough.

lhoestq (approved):
Good job on reducing the size of the dummy data files 
LGTM :)

PR #1095, opened 2020-12-04
Add TupleInf Open IE Dataset

mattbui:
For more information: https://allenai.org/data/tuple-ie


datasets/tuple_ie/tuple_ie.py
```diff
@@ -0,0 +1,161 @@
...
+            domains=list(_DOMAIN_FILES.keys()),
+            description="collected using training questions from 4th and 8th grade as queries.",
+        )
+    ] + [
+        TupleIEConfig(
+            name=name, domains=[name], description=f"collected using training questions from {name} as queries."
+        )
+        for name in _DOMAIN_FILES.keys()
+    ]
+
```
lhoestq, line 72:
```suggestion
    BUILDER_CONFIG_CLASS = TupleIEConfig
```
we need that to specify that we're using a custom config here

lhoestq (commented):
Thanks for adding this one :)

Don't forget to add the tags of the dataset card. It the yaml part at the top of the dataset card
more info here : https://github.com/huggingface/datasets/blob/master/ADD_NEW_DATASET.md#tag-the-dataset-and-write-the-dataset-card

Also if you could describe a bit the different fields of the dataset in the `Data Fields` paragraph of the dataset card it would be useful. You can leave the other paragraphs as `[More Information Needed]`

lhoestq:
Errors are in the CI are not related to this PR (RemoteDatasetError)
the CI is fixed on master so it's fine

mattbui:
@lhoestq Added the dataset card. Please let me know if more information needs to be added.

lhoestq (approved):
Thanks for adding the dataset card :)

PR #1099, opened 2020-12-04
Add tamilmixsentiment data

jamespaultg:


lhoestq (commented):
Looks all good :) 

Could you add the tags of the dataset card ? It's the yaml part at the top of the dataset card.
You can find more info here : https://github.com/huggingface/datasets/blob/master/ADD_NEW_DATASET.md#tag-the-dataset-and-write-the-dataset-card

After that we'll be able to merge this one, thank you for adding this one !

lhoestq (approved):
Thanks for adding the tags !
Also good job on filling the dataset card as well

PR #1113, opened 2020-12-04
add qed

patil-suraj:
adding QED: Dataset for Explanations in Question Answering
https://github.com/google-research-datasets/QED
https://arxiv.org/abs/2009.06354


datasets/qed/qed.py
```diff
@@ -0,0 +1,137 @@
...
+            description=_DESCRIPTION,
+            features=datasets.Features(
+                {
+                    "example_id": datasets.Value("int64"),
+                    "title_text": datasets.Value("string"),
+                    "url": datasets.Value("string"),
+                    "question": datasets.Value("string"),
+                    "paragraph_text": datasets.Value("string"),
+                    "sentence_starts": datasets.Sequence(datasets.Value("int32")),
+                    "original_nq_answers": [[span_features]],
```
lhoestq, line 81: why is it nested as lis of list of spans ?

datasets/qed/qed.py
```diff
@@ -0,0 +1,137 @@
...
+                        ],
+                        "answer": [
+                            {
+                                "sentence_reference": reference_features,
+                                "paragraph_reference": span_features,
+                            }
+                        ],
+                        "explanation_type": datasets.Value("string"),
+                        "selected_sentence": span_features,
+                    },
```
lhoestq, line 97: could we simplify this using less nested structures ?

for example
```
"sentence_answers": datasets.Sequence(span_features),
"paragraph_answers": datasets.Sequence(span_features),
"sentence_referential_equalities": datasets.Sequence(reference_features),
"paragraph_referential_equalities": datasets.Sequence(reference_features),
"explanation_type": datasets.Value("string"),
"selected_sentence": span_features,
```
let me know what you think

lhoestq (commented):
Really nice thanks !

I left a few comments

datasets/qed/qed.py
patil-suraj: I went with the original structure, but looked at the data again, and yes inner list is unnecessary

datasets/qed/qed.py
patil-suraj: I think it would be better to stick with author's conventions. The nested structure seems conceptually easier to understand.  Also pinging @yjernite  about this.

datasets/qed/qed.py
yjernite: Let's simplify that one then

datasets/qed/qed.py
yjernite: We can keep the original structure to stay coherent

lhoestq (approved):
LGTM :)

also thanks for completing the sharc dataset card

PR #1119, opened 2020-12-04
Add Google Great Code Dataset

abhishekkrthakur:


src/datasets/utils/mock_download_manager.py
```diff
@@ -169,6 +170,9 @@ def create_dummy_data_dict(self, path_to_dummy_data, data_url):
 
     def create_dummy_data_list(self, path_to_dummy_data, data_url):
         dummy_data_list = []
+        # trick: if there are many shards named like `data.txt-000001-of-00300`, only use the first one
+        if all(bool(re.findall("[0-9]{3,}-of-[0-9]{3,}", url)) for url in data_url):
+            data_url = [data_url[0]] * len(data_url)
```
lhoestq, line 175: added this trick to be able to have only 1 file per split instead of 300 @abhishekkrthakur

abhishekkrthakur: nice!

abhishekkrthakur: file is much smaller now.

lhoestq (approved):
LGTM thanks :)

PR #945, opened 2020-12-01
Adding Babi dataset - English version

thomwolf:
Adding the English version of bAbI.

Samples are taken from ParlAI for consistency with the main users at the moment.

PR #1126, opened 2020-12-04
Adding babi dataset

thomwolf:
Adding the English version of bAbI.

Samples are taken from ParlAI for consistency with the main users at the moment.

Supersede #945 (problem with the rebase) and adresses the issues mentioned in the review (dummy data are smaller now and code comments are fixed).


lhoestq (commented):
Very cool thank you !

Looks like the dummy_data.zip files are quite heavy (>150KB) per file.
Could you try to reduce their sizes ?

For example you can take a look inside the dummy_data.zip file of `task_10_with_10k_examples` and remove the all the unnecessary files from the other configurations.

thomwolf:
I already reduced as much as possible 😬

thomwolf:
This is ok now @lhoestq

I've included the tweak to `dummy_data` to only use the data transmitted to `_generate_examples` by default (it only do that if it can find at least one path to an existing file in the `gen_kwargs` and this can be unactivated with a flag).

Should I extract it in another PR or is it ok like this?

lhoestq:
Nice !
Could you add the dummy data generation trick in another PR ?
I think we can also extend it to make it work not only with data files paths but also with data directories (sometimes it's one of the parent directory that is passed to gen_kwargs, not the actual path to the file).
This will help a lot to make the dummy data lighter !

gchhablani:
This PR can be closed in lieu of #2053 @lhoestq

gchhablani:
This PR can be closed due to #2053 @lhoestq

PR #1127, opened 2020-12-04
Add wikiqaar dataset

zaidalyafeai:
Arabic Wiki Question Answering Corpus.


datasets/wiki_qa_ar/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Dataset Structure
+
+### Data Instances
+
+Each data point contains the question and whether the answer is a valid or not.  
+
+### Data Fields
+
+[More Information Needed]
```
lhoestq, line 75: Could you detail a bit the fields ? Especially the `label` one since the name is not explicit

datasets/wiki_qa_ar/wiki_qa_ar.py
```diff
@@ -0,0 +1,119 @@
...
+        return datasets.DatasetInfo(
+            description=_DESCRIPTION,
+            features=datasets.Features(
+                {
+                    "question_id": datasets.Value("string"),
+                    "question": datasets.Value("string"),
+                    "document_id": datasets.Value("string"),
+                    "answer_id": datasets.Value("string"),
+                    "answer": datasets.Value("string"),
+                    "label": datasets.features.ClassLabel(num_classes=2),
```
lhoestq, line 82: do the labels have names that we can use here ?

lhoestq (commented):
Looks good, thanks for adding this one :)
I left a few comments

datasets/wiki_qa_ar/wiki_qa_ar.py
zaidalyafeai: Not that I am aware of. I think it is 1 if the sentence is an answer to the question or 0 otherwise.

datasets/wiki_qa_ar/README.md
zaidalyafeai: Fixed, let me now if that's ok.

datasets/wiki_qa_ar/wiki_qa_ar.py
yjernite: from the Github repo: "We make this dataset publicly available for the community to explore models such as cross-language question answering and answer triggering across languages; i.e. deciding if an answer exists in the same language as the question and ---if not--- searching in the other language."

Can you update the `Data Fields` in the README to reflect that?

zaidalyafeai: The dataset was translated from English and it can be used for bi-lingual tasks. I didn't include the English data because I think it is already added in `wiki_qa` https://huggingface.co/datasets/wiki_qa.

yjernite: Yes! My understanding was that the `label` indicates whether the answer can be found in Arabic Wikipedia or not: if that is indeed the case, we should say it in the README.md :)

zaidalyafeai: I already made a commit with the updated README.md.

lhoestq (approved):
Thanks ! all good now :)

PR #1137, opened 2020-12-04
add wmt mlqe 2020 shared task

VictorSanh:
First commit for Shared task 1 (wmt_mlqw_task1) of WMT20 MLQE (quality estimation of machine translation)
Note that I copied the tags in the README for only one (of the 7 configurations): `en-de`.
There is one configuration for each pair of languages.


VictorSanh:
re-created in #1218 because this was too messy

PR #1139, opened 2020-12-04
Add ReFreSD dataset

mpariente:
This PR adds the **ReFreSD dataset**. 
The original data is hosted [on this github repo](https://github.com/Elbria/xling-SemDiv) and we use the `REFreSD_rationale` to expose all the data. 

Need feedback on:
- I couldn't generate the dummy data. The file we download is a tsv file, but without extension, I suppose this is the problem. I'm sure there is a simple trick to make this work. 
- The feature names. 
  - I don't know if it's better to stick to the classic `sentence1`, `sentence2` or to `sentence_en`, `sentence_fr` to be more explicit. 
  - There is a binary label (called `label`, no problem here), and a 3-class label called `#3_labels` in the original tsv. I changed it to `all_labels` but I'm sure there is better. 
- The rationales are lists of integers, extracted as a string at first. I wonder what's the best way to treat them, any idea? Also, I couldn't manage to make a `Sequence` of `int8` but I'm sure I've missed something simple. 

Thanks in advance


datasets/refresd/refresd.py
```diff
@@ -0,0 +1,109 @@
...
+
+_DESCRIPTION = """\
+The Rationalized English-French Semantic Divergences (REFreSD) dataset consists of 1,039 
+English-French sentence-pairs annotated with sentence-level divergence judgments and token-level 
+rationales. For any questions, write to ebriakou@cs.umd.edu.
+"""
+
+_HOMEPAGE = "https://github.com/Elbria/xling-SemDiv/tree/master/REFreSD"
+
+_LICENSE = ""
```
mpariente, line 46: No license in the repo, I could open a PR with MIT license, I guess it would get merged.

datasets/refresd/refresd.py
```diff
@@ -0,0 +1,109 @@
...
+    def _split_generators(self, dl_manager):
+        """Returns SplitGenerators."""
+        my_urls = _URL
+        data_file_path = dl_manager.download_and_extract(my_urls)
+        return [
+            datasets.SplitGenerator(
+                name=datasets.Split.TRAIN,
+                gen_kwargs={
+                    "filepath": data_file_path,
+                    "split": "train",
```
mpariente, line 91: Realize now that this can be removed.

yjernite:
Cool dataset! Replying in-line:

> This PR adds the **ReFreSD dataset**.
> The original data is hosted [on this github repo](https://github.com/Elbria/xling-SemDiv) and we use the `REFreSD_rationale` to expose all the data.
> 
> Need feedback on:
> 
> * I couldn't generate the dummy data. The file we download is a tsv file, but without extension, I suppose this is the problem. I'm sure there is a simple trick to make this work.

you can use `--match_text_files` in the dummy data generation:
`python datasets-cli dummy_data datasets/refresd --auto_generate --match_text_files "REFreSD_rationale"`

> * The feature names.
>   
>   * I don't know if it's better to stick to the classic `sentence1`, `sentence2` or to `sentence_en`, `sentence_fr` to be more explicit.

It would actually be even better to use the `Translation` feature here to replace best:
`"sentence_pair": datasets.Translation(languages=['en', 'fr']),`

Then during `_generate_examples` this filed should look like"
`{"sentence_pair": {"fr": french, "en": english}}`

>   * There is a binary label (called `label`, no problem here), and a 3-class label called `#3_labels` in the original tsv. I changed it to `all_labels` but I'm sure there is better.
Looks good!

> * The rationales are lists of integers, extracted as a string at first. I wonder what's the best way to treat them, any idea? Also, I couldn't manage to make a `Sequence` of `int8` but I'm sure I've missed something simple.

Having the feature declared as `"rationale_en": datasets.Sequence(datasets.Value("int32"))` should work

> 
> Thanks in advance

Hope that helps you out! Don't forget to `make style`, rebase from master, and run all the tests before pushing again! You will also need to add a `README.md` as described in the guide:

mpariente:
Thanks a lot for the answer, that does help a lot !
I opened a PR for a License in the original repo so I was waiting for that for the model card. If there is no news on Monday, I'll add it without License.

yjernite:
Looks good! It looks like it might need a rebase to pass the tests. Once you do that, should be good to go!

datasets/refresd/refresd.py
Elbria: Hi Manuel :) I just merged your request.

mpariente: Hey, thanks a lot for your message :upside_down_face: 
I'll get back to it, probably tomorrow.

lhoestq (approved):
Nice thank you !

PR #1142, opened 2020-12-04
Fix PerSenT

jeromeku:
New PR for dataset PerSenT


datasets/per_sent/per_sent.py
```diff
@@ -0,0 +1,145 @@
...
+}
+
+
+class PerSenT(datasets.GeneratorBasedBuilder):
+    """Person SenTiment (PerSenT) is a crowd-sourced dataset that captures the sentiment of an author towards the main entity in a news article. This dataset contains annotation for 5.3k documents and 38k paragraphs covering 3.2k unique entities."""
+
+    VERSION = datasets.Version("1.1.0")
+
+    def _info(self):
+        label = datasets.features.ClassLabel(names=["", "Negative", "Neutral", "Positive"])
```
lhoestq, line 66:
```suggestion
        label = datasets.features.ClassLabel(names=["Negative", "Neutral", "Positive"])
```
Usually we only use the valid class label names here, and when a label is missing we use the value `-1` in the example dictionary that is yielded

Could you change that ?
You'll need to re-generate the dataset_infos.json after this change

datasets/per_sent/per_sent.py
```diff
@@ -0,0 +1,145 @@
...
+
+_URLs = {
+    "train": "https://raw.githubusercontent.com/MHDBST/PerSenT/main/train.csv",
+    "dev": "https://raw.githubusercontent.com/MHDBST/PerSenT/main/dev.csv",
+    "test_random": "https://raw.githubusercontent.com/MHDBST/PerSenT/main/random_test.csv",
+    "test_fixed": "https://raw.githubusercontent.com/MHDBST/PerSenT/main/fixed_test.csv",
+}
+
+
+class PerSenT(datasets.GeneratorBasedBuilder):
```
lhoestq, line 60:
```suggestion
class PerSent(datasets.GeneratorBasedBuilder):
```
this mus be the camel case version of the dataset script name `per_sent`

datasets/per_sent/README.md
```diff
@@ -0,0 +1,321 @@
...
+### Languages
+
+English
+
+## Dataset Structure
+
+### Data Instances
+
+```json
+  {
```
lhoestq, line 79: this is the json representation of the feature types right ?

in this paragraph we expect to see one example of the data instead.
For example you can get the first example with `dataset["train"][0]`

could you change that please ?

datasets/per_sent/README.md
```diff
@@ -0,0 +1,321 @@
...
+      "class_names": [
+        "",
+        "Negative",
+        "Neutral",
+        "Positive"
+      ]
+    }
+  }
+```
+
```
lhoestq, line 272: can you re-add the missing Data Fields section please ?
in there you can explain what each field means and also mention that missing labels are replaced with -1

lhoestq (commented):
Thanks for the changes !
I left the final comments :)
The rest looks all good to me

datasets/per_sent/README.md
jeromeku: Fixed

datasets/per_sent/README.md
```diff
@@ -0,0 +1,163 @@
...
+  - [Dataset Creation](#dataset-creation)
+    - [Curation Rationale](#curation-rationale)
+    - [Source Data](#source-data)
+      - [Initial Data Collection and Normalization](#initial-data-collection-and-normalization)
+      - [Who are the source language producers?](#who-are-the-source-language-producers)
+    - [Annotations](#annotations)
+      - [Annotation process](#annotation-process)
+      - [Who are the annotators?](#who-are-the-annotators)
+    - [Personal and Sensitive Information](#personal-and-sensitive-information)
+  - [Considerations for Using the Data](#considerations-for-using-the-data)
```
lhoestq, line 44: Looks like this section is missing.

Can you re-add all the sections that were removed ? You can leave their content with `[More Information Needed]`

jeromeku: Done!

lhoestq (approved):
Looks all good thanks !

PR #1143, opened 2020-12-04
Add the Winograd Schema Challenge

joeddav:
Adds the Winograd Schema Challenge, including configs for the more canonical wsc273 as well as wsc285 with 12 new examples.

- https://cs.nyu.edu/faculty/davise/papers/WinogradSchemas/WS.html

The data format was a bit of a nightmare but I think I got it to a workable format.


datasets/winograd_wsc/winograd_wsc.py
```diff
@@ -0,0 +1,134 @@
...
+            description=_DESCRIPTION,
+            features=datasets.Features(
+                {
+                    "text": datasets.Value("string"),
+                    "pronoun": datasets.Value("string"),
+                    "pronoun_loc": datasets.Value("int32"),
+                    "quote": datasets.Value("string"),
+                    "quote_loc": datasets.Value("int32"),
+                    "options": datasets.Sequence(datasets.Value("string")),
+                    "label": datasets.Value("int32"),
```
lhoestq, line 86: Can you use the ClassLabel feature type here ?

lhoestq (commented):
Thanks for adding this one !
Good dataset card as well

I just left one comment about the feature type of the label fields, let me know what you think

datasets/winograd_wsc/winograd_wsc.py
joeddav: Sure yeah, as long as it doesn't matter that the meaning of the label changes from instance to instance (so having a label name doesn't make sense)

lhoestq (approved):
Thanks !

PR #1150, opened 2020-12-05
adding dyk dataset

abecadel:


datasets/dyk/README.md
```diff
@@ -0,0 +1,142 @@
+---
+YAML tags:
```
lhoestq, line 2:
```suggestion
```

datasets/dyk/README.md
```diff
@@ -0,0 +1,142 @@
...
+### Data Instances
+
+[More Information Needed]
+
+### Data Fields
+
+- q_id: question id
+- question: question sentence
+- answer: answer sentence
+- target:
```
lhoestq, line 80:
```suggestion
- target: 1 if the answer is correct, 0 otherwise. Note that the test split doesn't have target values so -1 is used instead
```

lhoestq (approved):
Nice thanks for adding this one !

PR #1154, opened 2020-12-05
Opus sardware

spatil6:
Added Opus sardware dataset for machine translation English to Sardinian.
for more info : http://opus.nlpl.eu/sardware.php


datasets/opus_sardware/opus_sardware.py
```diff
@@ -0,0 +1,98 @@
...
+            len(target_sentences),
+            source_file,
+            target_file,
+        )
+
+        source, target = tuple(self.config.name.split("-"))
+        for idx, (l1, l2) in enumerate(zip(source_sentences, target_sentences)):
+            result = {"translation": {source: l1, target: l2}}
+            if all(result.values()):
+                yield idx, result
```
lhoestq, line 98:
```suggestion
            yield idx, result
```
the if statement is always true here

lhoestq (approved):
Very clean thanks !

datasets/opus_sardware/opus_sardware.py
lhoestq: let me know if you had something in mind for this, otherwise we can commit this change and merge

PR #1160, opened 2020-12-05
adding TabFact dataset

patil-suraj:
Adding TabFact: A Large-scale Dataset for Table-based Fact Verification.

https://github.com/wenhuchen/Table-Fact-Checking

- The tables are stored as individual csv files, so need to download 16,573 🤯  csv files. As a result the `datasets_infos.json` file is huge (6.62 MB).
- Original dataset has nested structure where, where table is one example and each table has multiple statements,
flattening the structure here so that each statement is one example.


lhoestq (commented):
Really cool thanks !

I noticed that because there are many downloaded files, the dataset_infos.json file gets super heavy (6MB), do you think it is possible to download all of them at once ? I think it's ok to download the github repository containing the data. Ideally the link that downloads the repo is fixed using a commit hash, so that the downloaded archive always have the same content

lhoestq (approved):
Thanks !

julien-c:
FYI you guys are on GitHub's homepage 😍

Screenshot 2020-12-09 at 12 34 28

patil-suraj:
Yeayy 🔥

PR #1161, opened 2020-12-05
Linguisticprobing

sileod:
Adding Linguistic probing datasets from
What you can cram into a single $&!#* vector: Probing sentence embeddings for linguistic properties
 https://www.aclweb.org/anthology/P18-1198/


datasets/linguisticprobing/README.md
```diff
@@ -0,0 +1,32 @@
...
+  top_constituents:
+  - 10K<n<100K
+  tree_depth:
+  - 100K<n<1M
+  word_content:
+  - 100K<n<1M
+source_datasets: []
+task_categories:
+- text-classification
+task_ids: []
```
lhoestq, line 32: Looks like the dataset card is missing the content summary as well as the section titles.
More info here : https://github.com/huggingface/datasets/blob/master/ADD_NEW_DATASET.md#tag-the-dataset-and-write-the-dataset-card

You can leave the section contents with `[More Information Needed]`

lhoestq (commented):
Looking good so far thanks !

Could you add the missing sections in the dataset card ? you can just have the titles and leave the content with `[More Information Needed]`

also it looks like there are also changes about the `discovery` dataset, could you include in this PR only changes for the `linguisticprobing` dataset and add `discoevry` in a separate one please ?

datasets/linguisticprobing/README.md
```diff
@@ -0,0 +1,60 @@
...
+  sentence_length:
+  - text-classification-other-sentence length prediction
+  subj_number:
+  - text-classification-other-subjects number prediction
+  top_constituents:
+  - text-classification-other-constituents prediction
+  tree_depth:
+  - text-classification-other-syntactic tree depth prediction
+  word_content:
+  - text-classification-other-words prediction
```
lhoestq, line 60: thanks for adding all these tags !

Can you also add the other sections of the dataset card ?
You can leave their content with `[More Information Needed]`
You can see the template here : https://github.com/huggingface/datasets/blob/master/templates/README.md

lhoestq (commented):
Really cool thanks :) 

Before we merge, can you also add the other sections of the dataset card ?

Also can you try to reduce the sizes of the dummy_data.zip files (they're currently 17KB each).
To do so feel free to take a look inside the dummy_data.zip file for the `bigram_shift` config for example and remove all the unnecessary files about the other configs.

lhoestq (commented):
Thanks ! did you get a chance to take a look at the dummy data files to try to reduce their sizes ?

albertvillanova:
Thanks for your contribution, @sileod.

As you already created this dataset under your organization namespace (https://huggingface.co/datasets/metaeval/linguisticprobing), I think we can safely close this PR.

We would suggest you add a dataset card with the YAML tags, to make it searchable and discoverable.

Please, feel free to tell us if you need some help.

PR #1162, opened 2020-12-05
Add Mocha dataset

mattbui:
More information: https://allennlp.org/mocha


lhoestq (approved):
Thanks for adding this one !
LGTM
and good dataset card as well

PR #1170, opened 2020-12-05
Fix path handling for Windows

edugp:


edugp:
@lhoestq here's the fix!

lhoestq (approved):
thanks !

PR #1176, opened 2020-12-05
Add OpenPI Dataset

Bharat123rox:
Add the OpenPI Dataset by AI2 (AllenAI)


yjernite:
Hi @Bharat123rox ! It looks like some of the dummy data is broken or missing. Did you auto-generate it? Does the local test pass for you?

Bharat123rox:
@yjernite requesting you to have a look as to why the tests are failing only on Windows, there seems to be a backslash error somewhere, could it be the result of `os.path.join` and what should be the fix for this?

Bharat123rox:
This is the black output locally:
```
(datasets_env) datasets (openpi) > black --check --line-length 119 --target-version py36 datasets/openpi/
All done! ✨ 🍰 ✨
1 file would be left unchanged.
```

yjernite:
Can you check your version of black (should be `20.8b1`) and run `make style again`? (And don't forget to rebase before pushing ;) )

The other test was a time-out error so should be good on the next commit

Bharat123rox:
Thanks @yjernite the CI tests finally passed!!

datasets/openpi/README.md
```diff
@@ -0,0 +1,318 @@
...
+multilinguality:
+- monolingual
+size_categories:
+- 1K<n<10K
+source_datasets:
+- original
+task_categories:
+- named-entity-recognition
+task_ids:
+- structure-prediction
```
lhoestq, line 19:
```suggestion
task_categories:
- structure-prediction
task_ids:
- named-entity-recognition
```

datasets/openpi/openpi.py
```diff
@@ -0,0 +1,247 @@
...
+            description="Provides the answer sentences",
+        ),
+        OpenpiConfig(
+            name="answers-metadata",
+            mode="answers",
+            type_="metadata",
+            version=VERSION,
+            description="Provides the answer metadata",
+        ),
+    ]
```
lhoestq, line 109: Do you think it would be useful to have only one config with all the data in it ?
Not sure if it makes sense to have a config with only the question and no metadata nor answers + answers metadata

datasets/openpi/openpi.py
```diff
@@ -0,0 +1,247 @@
...
+        with open(filepath, encoding="utf-8") as f:
+            for id_, row in enumerate(f):
+                data = json.loads(row)
+                if "question" in data:
+                    result = {"id": data["id"], "question": data["question"]}
+                elif "question_metadata" in data:
+                    result = {
+                        "id": data["id"],
+                        "question_metadata": {
+                            "url": data["question_metadata"].get("url", ""),
```
lhoestq, line 224: Why use get here ? Are some urls fields missing ?
Same for the other fields

If some of them are missing you can keep `get` but maybe you can mention that you replace missing values with empty strings in the dataset card

lhoestq (commented):
Nice thanks for adding this one !

I left a few comments

datasets/openpi/openpi.py
Bharat123rox: The Data are split across 4 files tied together only with a single ID field and the files are split that way (Given in https://github.com/Bharat123rox/datasets/blob/6a1b6dd23269c62eadfb4a53ffb0b8ba83a149c7/datasets/openpi/openpi.py#L188-L193)
(Metadata provide additional information different from the raw questions/answers, and all of them can be "joined" or tied only with the ID of the question, so I felt that the user could fetch each of these aspects on demand, hence the 4-way config)

If not, how do you visualize the ideal data model? Also, if there is a way where we could specify some sort of joining key, I'm definitely open to making these changes!

datasets/openpi/openpi.py
Bharat123rox: Which section in the card should this go under? I didn't get it from the `README_guide.md`

Bharat123rox: Included in the dataset card

lhoestq:
Hi @Bharat123rox did you manage to join the different config into one using the IDs ?

Feel free to ping me when you're ready for the next review :)

Bharat123rox:
> Hi @Bharat123rox did you manage to join the different config into one using the IDs ?
> 
> Feel free to ping me when you're ready for the next review :) 

Not yet @lhoestq still working on this! Meanwhile please review #1507 where I added the SelQA dataset :)

lhoestq:
Ok ! Let me review SelQA then :) 
Thanks for your help !

nikett:
Apologies for the very late response. Here is the openpi dataset file with a single file per partition after merging `id_answers, answers.jsonl, question.jsonl , question_metadata.jsonl`

https://github.com/allenai/openpi-dataset/blob/main/data/gold-v1.1/dev.jsonl

lhoestq:
Nice thank you @nikett !

nikett:
Hi @Bharat123rox , when you get a chance, please feel free to use the dataset from the repo ( [Link](https://github.com/allenai/openpi-dataset/blob/main/data/gold-v1.1/dev.jsonl) ) . Please let me know if any file is missing! Thank you

nikett:
Hi @Bharat123rox are you working on this?

Bharat123rox:
@nikett Sorry I'm no longer working on this as I'm out of time for it, please feel free to raise a new PR for this

PR #1179, opened 2020-12-05
Small update to the doc: add flatten_indices in doc

thomwolf:
Small update to the doc: add flatten_indices in doc


lhoestq (approved):
thanks !

PR #1180, opened 2020-12-05
Add  KorQuAD v2 Dataset

cceyda:
# The Korean Question Answering Dataset v2
Adding the [KorQuAD](https://korquad.github.io/) v2 dataset as part of the sprint 🎉 
This dataset is very similar to SQuAD and is an extension of [squad_kor_v1](https://github.com/huggingface/datasets/pull/1178) which is why I added it as `squad_kor_v2`. 

- Crowd generated questions and answer (1-answer per question) for Wikipedia articles. Differently from V1 it includes the html structure and markup, which makes it a different enough dataset. (doesn't share ids between v1 and v2 either)

- [x] All tests passed
- [x] Added dummy data

Edit: 🤦 looks like squad_kor_v1 commit sneaked in here too


lhoestq:
looks like this PR also includes the changes for the V1
Could you only include the files of the V2 ?

datasets/squad_kor_v2/squad_kor_v2.py
```diff
@@ -0,0 +1,149 @@
...
+                name=datasets.Split.VALIDATION, gen_kwargs={"dirs": downloaded_files["validation"]}
+            ),
+        ]
+
+    def _generate_examples(self, dirs):
+        """Yields examples."""
+
+        for d in dirs:
+            filepaths = os.scandir(d)
+            for filepath in filepaths:
```
lhoestq, line 120:
```suggestion
            for filepath in sorted(filepaths):
```
we need the order here to be deterministic

lhoestq (commented):
Looking pretty good so far !

I noticed that the dummy_data.zip file was quite big (6MB)
Could you try to make it smaller ? Usually they're only a few KB

To do so feel free to take a look inside it and inside the `KorQuAD_2.1_train_*` directories and remove all the unnecessary files to keep only 1.
Also feel free to remove unnecessary chunks of text in the json files

cceyda:
hmm I have made the dummy data lighter retested on local and it passed not sure why it fails here?

lhoestq (approved):
Thanks ! LGTM

lhoestq:

PR #1184, opened 2020-12-05
Add Adversarial SQuAD dataset

cceyda:
# Adversarial SQuAD

Adding the Adversarial [SQuAD](https://github.com/robinjia/adversarial-squad) dataset as part of the sprint 🎉 
This dataset adds adversarial sentences to a subset of the SQuAD dataset's dev examples. How to get the original squad example id is explained in readme->Data Instances. The whole data is intended for use in evaluation. (Which could of course be also used for training if one wants). So there is no classical train/val/test split, but a split based on the number of adversaries added.

There are 2 splits of this dataset:

- AddSent: Has up to five candidate adversarial sentences that don't answer the question, but have a lot of words in common with the question. This adversary is does not query the model in any way.
- AddOneSent: Similar to AddSent, but just one candidate sentences was picked at random. This adversary is does not query the model in any way.

(The AddAny and AddCommon datasets mentioned in the paper are dynamically generated based on model's output distribution thus are not included here)

The failing test look like some unrelated timeout thing, will probably clear if rerun.
- [x] All tests passed
- [x] Added dummy data


lhoestq:
the CI error was just a connection error due to all the activity on the repo this week ^^'
I re-ran it so it should be good now

datasets/squad_adversarial/README.md
```diff
@@ -0,0 +1,171 @@
...
+`id` field is formed like: [original_squad_id]-[annotator_id]
+
+### Data Fields
+```py
+{'id': Value(dtype='string', id=None),
+ 'title': Value(dtype='string', id=None),
+ 'context': Value(dtype='string', id=None),
+ 'question': Value(dtype='string', id=None),
+ 'answers': Sequence(feature={'text': Value(dtype='string', id=None), 'answer_start': Value(dtype='int32', id=None)}, length=-1, id=None)}
+```
```
lhoestq, line 89: here we expect a description in natural language of each field, can you change this please ?

lhoestq (commented):
Thanks for adding this one !

I just left one comment about the Data Fields section of the dataset card

Moreover it looks like the dummy_data.zip file is quite heavy (700KB), could you try to reduce it ?
To do so feel free to take a look inside it and remove unnecessary files and unnecessary chunks of texts

datasets/squad_adversarial/squad_adversarial.py
```diff
@@ -0,0 +1,125 @@
...
+AddSent: Generates up to five candidate adversarial sentences that don't answer the question, but have a lot of words in common with the question. Picks the one that most confuses the model.
+AddOneSent: Similar to AddSent, but just picks one of the candidate sentences at random. This adversary is does not query the model in any way.
+"""
+
+_HOMEPAGE = "https://worksheets.codalab.org/worksheets/0xc86d3ebe69a3427d91f9aaa63f7d1e7d/"
+_LICENSE = "MIT License"
+
+_URLS = {
+    "AddSent": "https://worksheets.codalab.org/rest/bundles/0xb765680b60c64d088f5daccac08b3905/contents/blob/",
+    "AddOneSent": "https://worksheets.codalab.org/rest/bundles/0x3ac9349d16ba4e7bb9b5920e3b1af393/contents/blob/",
```
lhoestq, line 54: I think it would be better to have two configurations for this dataset, one for each adversaries.
What do you think ?

cceyda: I thought about that too, but decided to name it as splits (although unconventional). because AddSent is kind of like a subset of AddOneSent explained in the README. My thinking was that they both have the same schema so they are not different configurations.
But if this is a bit too unconventional I can easily change it to be configurations

datasets/squad_adversarial/README.md
cceyda: The fields had felt pretty self-explanatory following standard SQuAD format. (I mentioned differences in data Instances)
But I added some more explanations now just in case

cceyda:
I hadn't realized the problem with the dummies since it had passed without errors.
Suggestion: maybe we can show the user a warning based on the generated dummy size.

datasets/squad_adversarial/squad_adversarial.py
lhoestq: Ok I see. We should go for two configurations then.

Configurations are used when the processing is different or the data are different, while splits are used to define which parts are used for the different steps of evaluation of a model.

Could you do this change please if you don't mind ?

cceyda: changed from splits to configs.

lhoestq:
Thanks for changing to configs ! Looks all good now :) 

Before we merge, can you re-lighten the dummy data please if you don't mind ? The idea is to have them weigh only a few KB (currently it's 50KB each). Feel free to remove any unnecessary files or chunk of text

lhoestq:
(also you can ignore the `RemoteDatasetTest ` CI errors, they're fixed on master )

lhoestq (approved):
Awesome thank you !

lhoestq:

PR #1199, opened 2020-12-06
Turkish NER dataset, script works fine, couldn't generate dummy data

merveenoyan:
I've written the script (Turkish_NER.py) that includes dataset. The dataset is a zip inside another zip, and it's extracted as .DUMP file. However, after preprocessing I only get .arrow file. After I ran the script with no error messages, I get .arrow file of dataset, LICENSE and dataset_info.json.


datasets/Turkish_NER/Turkish_NER.py
```diff
@@ -0,0 +1,196 @@
...
+                    "split": "train",
+                },
+            ),
+        ]
+
+    def _generate_examples(self, filepath, split):
+        """ Yields examples. """
+        global data_dir
+        logging.info("⏳ Generating examples from = %s", filepath)
+        os.system(f"""unzip "{filepath}" -d {data_dir}""")
```
cceyda, line 156: I suggest you use `dl_manager.extract` function instead of os.system calls to get the zips inside the zip. (in the `_split_generators` function)

cceyda:
the .DUMP file looks like a txt with one example per line so adding `--match_text_files *.DUMP --n_lines 50`  to the dummy generation command might work .

datasets/Turkish_NER/Turkish_NER.py
merveenoyan: I just did 👍 But this still didn't extract the file as a .DUMP. when I run load_dataset I get a .lock file of the dataset in the "extracted" in folder in cache, and I still get .arrow file in the dataset's folder (in cache) instead of the .DUMP. So when I tried to generate a dummy file with the script you've sent it can't find any matches for it (obviously). I don't know in which step it decides to unzip it that way or if it's unsuccessful in unzipping, tried unit tests but couldn't find. Thanks for the help though.

cceyda: try clearing the cache folder and lock files then try again, I had a similar problem before.

datasets/Turkish_NER/Turkish_NER.py
```diff
@@ -0,0 +1,196 @@
...
+                        tokens.append(splits[2])
+            # last example
+            yield guid, {
+                "id": str(guid),
+                "domains": domains,
+                "tokens": tokens,
+                "ner_tags": ner_tags,
+            }
+
+
```
cceyda, line 192: You should also remove these lines ⬇️

datasets/Turkish_NER/Turkish_NER.py
```diff
@@ -0,0 +1,196 @@
...
+                gen_kwargs={
+                    "filepath": os.path.join(data_dir, _FILE_NAME_ZIP),
+                    "split": "train",
+                },
+            ),
+        ]
+
+    def _generate_examples(self, filepath, split):
+        """ Yields examples. """
+        global data_dir
```
lhoestq, line 154: Please don't use global vars. Instead you can pass variables from _split_generators to _generate_examples using the `gen_kwargs` dictionary

datasets/Turkish_NER/Turkish_NER.py
lhoestq, line 156: Can you use `dl_manager.extract` in the _split_generators method to unzip all the files ?
Then you can pass the files paths to _generate_examples using the `gen_kwargs`

datasets/Turkish_NER/Turkish_NER.py
```diff
@@ -0,0 +1,196 @@
+# coding=utf-8
```
lhoestq, line 1: can you move your dataset from `Turkish_NER` to `turkish_ner` ?
We're using the snake case convention

Can you also rename the dataset script accordingly  ?

lhoestq (commented):
This is a very good start ! good job

I left a few comments that could help you improve the extraction part of the script.

Then the next steps will be to generate the metadata file, the dummy data for tests and add a dataset card. You can find all the information there : https://github.com/huggingface/datasets/blob/master/ADD_NEW_DATASET.md

lhoestq:
We can close this PR since a new PR was open at #1268

PR #1211, opened 2020-12-06
Add large spanish corpus

lewtun:
Adds a collection of Spanish corpora that can be useful for pretraining language models. 

Following a nice suggestion from @yjernite we provide the user with three main ways to preprocess / load either 

* the whole corpus (17GB!)
* one specific sub-corpus
* the whole corpus, but return a single split. this is useful if you want to cache the whole preprocessing step once and interact with individual sub-corpora

See the dataset card for more details.

Ready for review!


datasets/large_spanish_corpus/large_spanish_corpus.py
```diff
@@ -0,0 +1,124 @@
...
+            license=_LICENSE,
+            citation=_CITATION,
+        )
+
+    def _split_generators(self, dl_manager):
+        data_dir = dl_manager.download_and_extract(_URL)
+        return [
+            datasets.SplitGenerator(name=corpus, gen_kwargs={"filepath": os.path.join(data_dir, filepath)})
+            for corpus, filepath in zip(self.config.corpora, self.config.filepaths)
+        ]
```
lhoestq, line 117: Splits are supposed to be used for train-dev-test splits for example depending of the model evaluations you want to make.
Here if I understand correctly the different corpora are just defined by their sources.
For different sources we usually use dataset configurations.

So I would keep all the configurations you defined, but I would change the splits.
Since the data is not split, please use one split `name=datasets.Split.TRAIN` for all of them.
For the combined config, the different corpora can simply be concatenated.

Does that make sense ?

datasets/large_spanish_corpus/large_spanish_corpus.py
```diff
@@ -0,0 +1,124 @@
...
+    "Europarl",
+    "NewsCommentary11",
+    "UN",
+    "EUBookShop",
+    "ParaCrawl",
+    "OpenSubtitles2018",
+    "DGT",
+]
+
+_CORPORA_FILEPATHS = {corpus: os.path.join("spanish-corpora/raw", f"{corpus}.txt") for corpus in _CORPORA}
```
lhoestq, line 67:
```suggestion
_CORPORA_FILEPATHS = {corpus: os.path.join("spanish-corpora", "raw", f"{corpus}.txt") for corpus in _CORPORA}
```

lhoestq (commented):
Awesome ! Thanks for adding this dataset.
This one is very exciting :) 

I left a comment about the organization of the splits, let me know what you think

datasets/large_spanish_corpus/large_spanish_corpus.py
lewtun: Thanks for the feedback!

Sorry for abusing the splits - I'll fix it as you suggest. When you say "the different corpora can simply be concatenated" are you saying I should change the existing config for combined? If yes, would you mind elaborating on what you'd like to see changed?

lewtun: OK I've updated the logic along the lines you suggested and am currently generating the `dataset_infos.json` and dummy data (takes a looong time with this corpus 😅). So the tests will fail until then ...

Is this what you had in mind?

lhoestq: Yes that's exactly what I had in mind, thank you !

datasets/large_spanish_corpus/large_spanish_corpus.py
```diff
@@ -0,0 +1,125 @@
...
+
+    BUILDER_CONFIGS = [
+        LargeSpanishCorpusConfig(name=corpus, corpora=[corpus], description=f"Spanish examples in corpus {corpus}.")
+        for corpus in _CORPORA
+    ] + [
+        LargeSpanishCorpusConfig(
+            name=_COMBINED, corpora=_CORPORA, description=f"Complete Spanish dataset with all corpora."
+        )
+    ]
+
```
lhoestq, line 95:
```suggestion
    BUILDER_CONFIG_CLASS = LargeSpanishCorpusConfig
```
minor change, doesn't affect the generation of the dataset_infos.json or the dummy data

datasets/large_spanish_corpus/large_spanish_corpus.py
lewtun: Cool! Will let you know when the PR is ready for a final pass :)

lewtun: Looks like the tests are passing with the new changes / files - ready for final review!

lhoestq (approved):
Awesome thank you so much !
LGTM

PR #1215, opened 2020-12-06
Add irc disentanglement

dhruvjoshi1998:
added files for irc disentanglement dataset
was unable to test dummy data as a result of vpn/proxy issues


datasets/irc_disentangle/irc_disentangle.py
```diff
@@ -0,0 +1,270 @@
...
+            annotations = dict()
+            for row in range(last_id, last_id + len(raw_sentences)):
+                annotations[row] = set()
+
+            for (a, b) in annotation_pairs:
+                # required for dummy data creation
+                if last_id + a not in annotations:
+                    annotations[last_id + a] = set()
+                if last_id + b not in annotations:
+                    annotations[last_id + b] = set()
```
lhoestq, line 246: are we sure there's no collision here ?

lhoestq (commented):
Thanks for adding this one ! :) 

I left a few comments.

Could you also add a few comments in the _generate_examples method to explain a bit more what's going on ?

lhoestq (commented):
Also the dummy_data.zip files are quite big (almost 400KB)

Could you try to make them smaller ?
For example you can take a look inside the dummy_data.zip file for the `ubuntu` config and remove unnecessary files related to the `channel_two` config. You can also remove unnecessary chunks for text inside the .txt files.
The idea is to have dummy_data.zip files that weigh only a few KB

Let me know if you have any question :)

datasets/irc_disentangle/irc_disentangle.py
```diff
@@ -0,0 +1,270 @@
...
+            description="This part of the dataset is the annotated conversations from the Ubuntu channel",
+        ),
+        datasets.BuilderConfig(
+            name="channel_two",
+            version=VERSION,
+            description="This part of the dataset is the annotated conversations from the Channel Two",
+        ),
+    ]
+
+    DEFAULT_CONFIG_NAME = "ubuntu"
```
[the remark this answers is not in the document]
dhruvjoshi1998: The ubuntu dataset was introduced in this paper, where as the channel-two dataset is for re-analysis of previous work.

from [the github](https://github.com/jkkummerfeld/irc-disentanglement):

> This work:
> 1. Introduces a new dataset, with disentanglement for 77,563 messages of IRC.
> 2. Introduces a new model, which achieves significantly higher results than prior work.
> 3. Re-analyses prior work, identifying issues with data and assumptions in models.

datasets/irc_disentangle/irc_disentangle.py
dhruvjoshi1998: I am not sure I understand fully.

The issue for the dummy data was that the number of lines being picked up was not complete, and hence the annotation lines referenced annotation sets that were not created. These lines create sets for these lines, however these sets are not yielded since the for loop with the yield does not reach these values.

lhoestq:
looks like this PR includes changes about many files other than the ones for irc_disentanglement

Could you please create a new branch and create another PR please ?

lhoestq:
closing in favor of #1586

PR #1221, opened 2020-12-06
Add HKCanCor

j-chim:
This PR adds the [Hong Kong Cantonese Corpus](http://compling.hss.ntu.edu.sg/hkcancor/), by [Luke and Wong 2015](http://compling.hss.ntu.edu.sg/hkcancor/data/LukeWong_Hong-Kong-Cantonese-Corpus.pdf). 

The dummy data included here was manually created, as the original dataset uses a xml-like format (see a copy hosted [here](https://github.com/fcbond/hkcancor/blob/master/sample/d1_v.txt) for example) that requires a few processing steps.


datasets/hkcancor/hkcancor.py
```diff
@@ -0,0 +1,161 @@
...
+"""
+
+_HOMEPAGE = "http://compling.hss.ntu.edu.sg/hkcancor/"
+
+_LICENSE = "CC BY 4.0"
+
+_URL = "http://compling.hss.ntu.edu.sg/hkcancor/data/hkcancor-utf8.zip"
+
+
+class HkCanCor(datasets.GeneratorBasedBuilder):
```
lhoestq, line 57:
```suggestion
class Hkcancor(datasets.GeneratorBasedBuilder):
```
this must be the camel case version of the dataset script name `hkcancor`

datasets/hkcancor/hkcancor.py
```diff
@@ -0,0 +1,161 @@
...
+
+    VERSION = datasets.Version("1.0.0")
+
+    def _info(self):
+        features = datasets.Features(
+            {
+                "conversation_id": datasets.Value("string"),
+                "speaker": datasets.Value("string"),
+                "turn_number": datasets.Value("int16"),
+                "words": datasets.Sequence(datasets.Value("string")),
```
lhoestq, line 68: can you rename this column `tokens` for consistency with other datasets ?

datasets/hkcancor/hkcancor.py
```diff
@@ -0,0 +1,161 @@
...
+                "transcriptions": datasets.Sequence(datasets.Value("string")),
+                "pos_tags": datasets.Sequence(datasets.Value("string")),
```
lhoestq, line 70: can you use the ClassLabel feature type here ?
You can find an example of usage in [conll2003](https://github.com/huggingface/datasets/blob/master/datasets/conll2003/conll2003.py) for example

lhoestq (commented):
Looks really cool ! thanks for adding this one

I left a few comments

datasets/hkcancor/hkcancor.py
j-chim: Done, thanks for the feedback @lhoestq!

I've also suggested some adjustments to the tagset. 

Basically, the original corpus included fine-grained tags that are (in my opinion) suited for qualitative linguistic analyses but challenging to use for NLP. To improve usability, I've left the original tags as-is, but also **added a mapping** that compresses the original set (n>100) to the **Universal Dependency 2.0** standard & into a **smaller set (n=16)**. This mapping is adapted from code from the PyCantonese library, which I have explicitly cited in the code and datasheet (both in the citation sections and where the dict was introduced). 

I know for this sprint the goal is to retain the properties of the original corpus with minimal additions, but I think it would be prohibitive to only have the original tags. Please let me know what you think. Thanks!

datasets/hkcancor/hkcancor.py
```diff
@@ -0,0 +1,315 @@
...
+            features=features,
+            supervised_keys=None,
+            homepage=_HOMEPAGE,
+            license=_LICENSE,
+            citation=_CITATION,
+        )
+
+    def _split_generators(self, dl_manager):
+        """Returns SplitGenerators."""
+        data_dir = dl_manager.download_and_extract(_URL) + "/utf8/"
```
lhoestq, line 233:
```suggestion
        data_dir = os.path.join(dl_manager.download_and_extract(_URL), "utf8")
```

datasets/hkcancor/hkcancor.py
```diff
@@ -0,0 +1,315 @@
...
+                    "data_dir": data_dir,
+                    "split": "train",
+                },
+            )
+        ]
+
+    def _generate_examples(self, data_dir, split):
+        """ Yields examples. """
+
+        downloaded_files = [data_dir + fn for fn in os.listdir(data_dir)]
```
lhoestq, line 248:
```suggestion
        downloaded_files = [data_dir + fn for fn in sorted(os.listdir(data_dir))]
```
we need to make the order deterministic

lhoestq (approved):
Thank you for adding the conversion to the Universal Dependencies 2.0 scheme, that's a great addition :)

PR #1226, opened 2020-12-06
Add menyo_20k_mt dataset

yvonnegitau:
Add menyo_20k_mt dataset


datasets/menyo_20k_mt/menyo_20k_mt.py
```diff
@@ -0,0 +1,111 @@
...
+    def _info(self):
+
+        return datasets.DatasetInfo(
+            # This is the description that will appear on the datasets page.
+            description=_DESCRIPTION,
+            # This defines the different columns of the dataset and their types
+            features=datasets.Features(
+                {
+                    "en": datasets.Value("string"),
+                    "yor": datasets.Value("string"),
```
lhoestq, line 84: can you use the `Translation` feature type here ?
you can find an example of usage in [flores](https://github.com/huggingface/datasets/blob/master/datasets/flores/flores.py) for example

lhoestq (commented):
Looking pretty good ! Thanks for adding this one

I left a comment about the feature type used for translation

Also we require the dataset class name `Menyo20kMt` to be the camel case version of the dataset script name and directory. Currently they're named `menyo_20k_mt` but we actually expect them to be named `menyo20k_mt`. Can you rename the directory and the daatset script name ?

You may need to regenerate the dataset_infos.json file after this change

lhoestq:
looks like your PR includes changes about many other files than the ones for menyo 20k mt
Can you create another branch and another PR please ?

yvonnegitau:
Yes, I will

PR #1228, opened 2020-12-06
add opus_100 dataset

VasudevGupta7:
This PR will add [opus100 dataset](http://opus.nlpl.eu/opus-100.php).


datasets/opus_100/README.md
```diff
@@ -0,0 +1,251 @@
+---
+task_categories:
+- sequence-modeling
+multilinguality:
+- translation
+task_ids:
+- language-modeling
+languages:
```
lhoestq, line 8: The languages depend on the configuration that is used. Therefore can you group the languages by config name here ?

For example:
```
languages:
  en-es:
  - en
  - es
  en-hi:
  - en
  - hi
```

Could you change this please ?

lhoestq (commented):
Looking pretty good so far !
Thank you so much for adding this one :) 

I left a comment about the languages tags

Also we need the dataset class name `Opus100` to be the camel case version of the dataset script name and directory. However they're named `opus_100` while we expect the snake case version to be `opus100`. Can you rename the directory and the dataset script `opus100` please ?

VasudevGupta7:
done.

lhoestq (approved):
Awesome thank you !

PR #1233, opened 2020-12-07
Add Curiosity Dialogs Dataset

vineeths96:
Add Facebook [Curiosity Dialogs](https://github.com/facebookresearch/curiosity) Dataset.


datasets/curiosity_dialogs/curiosity_dialogs.py
```diff
@@ -0,0 +1,220 @@
...
+                    "created_time": datasets.Value("int64"),
+                    "aspects": datasets.Sequence(datasets.Value("string")),
+                    "first_aspect": datasets.Value("string"),
+                    "second_aspect": datasets.Value("string"),
+                    "shuffle_facts": datasets.ClassLabel(names=["True", "False"]),
+                    "related_entities": datasets.Sequence(datasets.Value("string")),
+                    "tag": datasets.Value("string"),
+                    "user_id": datasets.Value("int32"),
+                    "assistant_id": datasets.Value("int32"),
+                    "is_annotated": datasets.ClassLabel(names=["True", "False"]),
```
lhoestq, line 112: usually we use the label at index 0 for False, and the one at index 1 for True

can you change that please ? and for the other boolean fields as well ?

you'll need to regenerate the dataset_infos.json after this change

lhoestq (commented):
Looking pretty good so far !
Thanks for adding this one

I left a comment about the boolean fields

Also it looks like the dummy_data.zip file is quite big (5MB). Could you try to reduce its size please ? Ideally it should only be a few KB

To do so feel free to take a look inside the zip file and remove and the unnecessary chunks of text inside the json files to only keep a few sentences

vineeths96:
@lhoestq I tried manually creating the dummy files. But unfortunately it was raising an error during testing the dummy data (regarding JSON parsing). Do you have any suggestions?

vineeths96:
@lhoestq I have made all the changes you mentioned.

lhoestq (approved):
Awesome thanks !

PR #1235, opened 2020-12-07
Wino bias

akshayb7:
The PR will fail circleCi tests because of the requirement of manual loading of data. Fresh PR because of messed up history of the previous one.


datasets/wino_bias/wino_bias.py
```diff
@@ -0,0 +1,312 @@
...
+            version=VERSION,
+            description="WinoBias: Winograd-schema dataset for detecting gender bias",
+        ),
+    ]
+
+    @property
+    def manual_download_instructions(self):
+        return """\
+             You need to manually download the 'anonymized.augmented.train.english.v4_auto_conll' file from Google Drive
+             (https://drive.google.com/uc?export=download&confirm=yLNb&id=14Im3BnNl-d2fYETYmiH5yq6eFGLVC3g0). The folder containing the saved file
```
lhoestq, line 85: You don't have to use manual data here

You can pass the url `https://drive.google.com/uc?export=download&id=14Im3BnNl-d2fYETYmiH5yq6eFGLVC3g0` directly to the `dl_manager` and it will download the file.

lhoestq (commented):

The url you're using works to download the data file.
Can you use the `dl_manager` instead of asking for manual download ?

You'll need to regenerate the dummy data and the dataset_infos.json after this change

datasets/wino_bias/wino_bias.py
akshayb7: Hi, @lhoestq , I did try the download_manager, but unfortunately it was not working. It points to the page of the download rather than the link to download the file itself. After discussion with @yjernite I re-wrote to download manually. 
Even now when I use the link provided by you, it still points to the download page itself.

lhoestq: The url should work with the download manager

For example for me when I run 
```python
from datasets import DownloadManager

DownloadManager().download("https://drive.google.com/uc?export=download&id=14Im3BnNl-d2fYETYmiH5yq6eFGLVC3g0")
```

it downloads the file.

What issue do you have with this link ?

akshayb7: Hi @lhoestq, thank you for persisting with me. When I try the code given by you (as is), it does download the file. But when I implement it in my code, I get the following error.

`>>> data = load_dataset('./datasets/wino_bias')
Downloading and preparing dataset wino_bias/wino_bias (download: Unknown size, generated: 82.91 MiB, post-processed: Unknown size, total: 82.91 MiB) to C:\Users\aksha\.cache\huggingface\datasets\wino_bias\wino_bias\4.0.0\81b712b8f7b261b5f1239cd953aa3dc38716fc03b99ff295acc50c12bda231c0...
Downloading: 269MB [01:01, 4.40MB/s] 
  File "d:\personal\huggingface\datasets\src\datasets\load.py", line 610, in load_dataset
    ignore_verifications=ignore_verifications,
  File "d:\personal\huggingface\datasets\src\datasets\builder.py", line 527, in download_and_prepare
    dl_manager=dl_manager, verify_infos=verify_infos, **download_and_prepare_kwargs
  File "d:\personal\huggingface\datasets\src\datasets\builder.py", line 587, in _download_and_prepare
    self.info.download_checksums, dl_manager.get_recorded_sizes_checksums(), "dataset source files"
  File "d:\personal\huggingface\datasets\src\datasets\utils\info_utils.py", line 34, in verify_checksums
    raise UnexpectedDownloadedFile(str(set(recorded_checksums) - set(expected_checksums)))
datasets.utils.info_utils.UnexpectedDownloadedFile: {'https://drive.google.com/uc?export=download&id=14Im3BnNl-d2fYETYmiH5yq6eFGLVC3g0'}`

I thought it is because of bad cache or me trying to redownload the data, but it persists even after clearing the cache. Coincidentally it always fails after downloading about 269 MBs of data while the original file downloaded manually is about 256 MBs.

akshayb7:
Closing this PR because of messed up history and opening another one after discussion with Quentin Lhoest.

PR #1236, opened 2020-12-07
Opus finlex dataset of language pair  Finnish and Swedish

spatil6:
Added Opus_finlex dataset of language pair  Finnish and Swedish
More info : http://opus.nlpl.eu/Finlex.php


lhoestq (approved):
Awesome thanks for adding this one !

PR #1238, opened 2020-12-07
adding poem_sentiment

patil-suraj:
Adding poem_sentiment dataset.
https://github.com/google-research-datasets/poem-sentiment


datasets/poem_sentiment/poem_sentiment.py
```diff
@@ -0,0 +1,85 @@
...
+        ]
+
+    def _generate_examples(self, filepath):
+        with open(filepath, encoding="utf-8") as f:
+            lines = f.readlines()
+            for line in lines:
+                fields = line.strip().split("\t")
+                idx, verse_text, label = fields
+                label = int(label)
+                label = label + 1 if label == 1 or label == -1 else label
```
lhoestq, line 84: is it possible to end up with label=1 for positive this way ??
label = -1 -> 0
label = 0 -> 0
label = 1 -> 2

lhoestq (commented):
Nice :) 

I left one comment about the labels, let me know what you think

datasets/poem_sentiment/poem_sentiment.py
patil-suraj: My bad. And thanks Quentin, while fixing this, I also realised that there is another label "mixed" which I missed to include

lhoestq (approved):
Nice thank you !

PR #1243, opened 2020-12-07
Add Google Noun Verb Dataset

abhishekkrthakur:


datasets/noun_verb/README.md
```diff
@@ -0,0 +1,142 @@
+---
+annotations_creators:
+- expert-generated
+language_creators:
+- expert-generated
+languages:
+- en
+licenses: [other-google-llc]
```
lhoestq, line 8:
```suggestion
licenses: 
- other-google-llc
```

datasets/noun_verb/noun_verb.py
```diff
@@ -0,0 +1,124 @@
...
+        return datasets.DatasetInfo(
+            description=_DESCRIPTION,
+            features=datasets.Features(
+                {
+                    "id": datasets.Value("int32"),
+                    "token_ids": datasets.Sequence(datasets.Value("int32")),
+                    "tokens": datasets.Sequence(datasets.Value("string")),
+                    "lemmas": datasets.Sequence(datasets.Value("string")),
+                    "upos_tags": datasets.Sequence(datasets.Value("string")),
+                    "xpos_tags": datasets.Sequence(datasets.Value("string")),
```
lhoestq, line 64: Can you use the ClassLabel feature type for those two fields please ?
You can take a look at an example of usage in [conll2003](https://github.com/huggingface/datasets/blob/master/datasets/conll2003/conll2003.py)

lhoestq (commented):
Really cool thanks !

I left a comment about the feature type

Could you do this change please ?
After this change you'll need to regenerate the dataset_infos.json

datasets/noun_verb/noun_verb.py
lhoestq: Hi @abhishekkrthakur what do you think about using class label ?

PR #1245, opened 2020-12-07
Add Google Turkish Treebank Dataset

abhishekkrthakur:


datasets/turkish_treebank/turkish_treebank.py
```diff
@@ -0,0 +1,123 @@
...
+        return datasets.DatasetInfo(
+            description=_DESCRIPTION,
+            features=datasets.Features(
+                {
+                    "id": datasets.Value("int32"),
+                    "token_ids": datasets.Sequence(datasets.Value("int32")),
+                    "tokens": datasets.Sequence(datasets.Value("string")),
+                    "lemmas": datasets.Sequence(datasets.Value("string")),
+                    "upos_tags": datasets.Sequence(datasets.Value("string")),
+                    "xpos_tags": datasets.Sequence(datasets.Value("string")),
```
lhoestq, line 79: can you use ClassLabel here ?
you can find an example in [conll2003](https://github.com/huggingface/datasets/blob/master/datasets/conll2003/conll2003.py)

lhoestq (commented):
Nice thanks ! one comment about the feature type.

also please regenerate the dataset_infos.json file after this change :)

datasets/turkish_treebank/turkish_treebank.py
lhoestq: (up)

What do you think about using Classlabel here ?

sorry for the spam ^^

PR #1258, opened 2020-12-07
arXiv dataset added

tanmoyio:


tanmoyio:
Need help

datasets/arxiv_dataset/arxiv_dataset.py
```diff
@@ -0,0 +1,138 @@
...
+                    path_to_manual_file, _FILENAME, self.manual_download_instructions
+                )
+            )
+        return [datasets.SplitGenerator(name=datasets.Split.TRAIN, gen_kwargs={"path": path_to_manual_file})]
+
+    def _generate_examples(self, path=None, title_set=None):
+        """ Yields examples. """
+
+        data = []
+        for line in open(path, "r"):
```
cceyda, line 120: it is complaining about not having  `encoding="utf-8"` defined change line to -> `for line in open(path,  encoding="utf-8"):`

cceyda: btw why are you loading json line by line. Look at the example here -> https://github.com/huggingface/datasets/blob/master/templates/new_dataset_script.py#L162

tanmoyio: @cceyda I have fixed the code and still it is showing this error
Screenshot (21)

PR #1206, opened 2020-12-06
Adding Enriched WebNLG dataset

TevenLeScao:
This pull requests adds the `en` and `de` versions of the [Enriched WebNLG](https://github.com/ThiagoCF05/webnlg) dataset

PR #1264, opened 2020-12-07
enriched webnlg dataset rebase

TevenLeScao:
Rebase of #1206 !


datasets/enriched_web_nlg/README.md
```diff
@@ -0,0 +1,193 @@
+---
+---
```
lhoestq, line 2: can you add the tags please ?

datasets/enriched_web_nlg/enriched_web_nlg.py
```diff
@@ -0,0 +1,189 @@
...
+
+_DESCRIPTION = """\
+WebNLG is a valuable resource and benchmark for the Natural Language Generation (NLG) community. However, as other NLG benchmarks, it only consists of a collection of parallel raw representations and their corresponding textual realizations. This work aimed to provide intermediate representations of the data for the development and evaluation of popular tasks in the NLG pipeline architecture (Reiter and Dale, 2000), such as Discourse Ordering, Lexicalization, Aggregation and Referring Expression Generation.
+"""
+
+_HOMEPAGE = "https://github.com/ThiagoCF05/webnlg"
+
+_LICENSE = "CC Attribution-Noncommercial-Share Alike 4.0 International"
+
+_URL = "https://github.com/ThiagoCF05/webnlg/zipball/master/"
```
lhoestq, line 48: You need to set the commit sha in the url, otherwise the data may be not deterministic

datasets/enriched_web_nlg/enriched_web_nlg.py
```diff
@@ -0,0 +1,189 @@
...
+        "test": [f"data/v1.5/en/test/{i}triples/" for i in range(1, 8)],
+    },
+    "de": {
+        "train": [f"data/v1.5/de/train/{i}triples/" for i in range(1, 8)],
+        "dev": [f"data/v1.5/de/dev/{i}triples/" for i in range(1, 8)],
+    },
+}
+
+
+def et_to_dict(tree):
```
lhoestq, line 63: can you add a docstring and some comments in this function ?

datasets/enriched_web_nlg/enriched_web_nlg.py
```diff
@@ -0,0 +1,189 @@
...
+        text = tree.text.strip()
+        if children or tree.attrib:
+            if text:
+                dct[tree.tag]["text"] = text
+        else:
+            dct[tree.tag] = text
+    return dct
+
+
+def parse_entry(entry):
```
lhoestq, line 84: same here

datasets/enriched_web_nlg/enriched_web_nlg.py
```diff
@@ -0,0 +1,189 @@
...
+            citation=_CITATION,
+            license=_LICENSE,
+        )
+
+    def _split_generators(self, dl_manager):
+        """Returns SplitGenerators."""
+        data_dir = dl_manager.download_and_extract(_URL)
+        # Downloading the repo adds the current commit sha to the directory, so we can't specify the directory
+        # name in advance.
+        revision_name = os.listdir(data_dir)[0]
```
lhoestq, line 163: ideally we should try to avoid doing that

If you force the use of a specific commit it should be fine noe ?

datasets/enriched_web_nlg/enriched_web_nlg.py
```diff
@@ -0,0 +1,189 @@
...
+                gen_kwargs={"filedirs": filedirs},
+            )
+            for split, filedirs in split_files.items()
+        ]
+
+    def _generate_examples(self, filedirs):
+        """ Yields examples. """
+
+        id_ = 0
+        print(filedirs)
```
lhoestq, line 181: can you remove the prints ?

lhoestq (commented):
Looks really good so far thanks :D 

I added some comments.
My may concern is the download url which may return different data in the future.
It can be fixed by specifying the sha of a commit in the url

lhoestq (commented):
Nice ! thanks for updating the url and adding the dummy data :) 

Also before we merge, can you try to reduce the size of the dummy data ? They're currently 50KB and we usually expect them to be only a few KB.
To do so feel free to take a look inside the dummy_data.zip files and remove unnecessary files and chunks of texts. For example for the dummy_data.zip file of the `en` config you can remove the unnecessary files related to the `de` config

TevenLeScao:
I've removed the `en` within `de` and reciprocally; but I don't think I will be able to thin it more than this.

lhoestq (approved):
Thanks !

PR #1292, opened 2020-12-08
arXiv dataset added

tanmoyio:


lhoestq (commented):
We need to have a dummy_data.zip file located in `./datasets/arxiv_dataset/dummy/default/1.1.0`

To build the dummy_data.zip file, here are the steps:
- go into `./datasets/arxiv_dataset/dummy/default/1.1.0`
- create a `dummy_data` folder
- inside this folder create a file `arxiv-metadata-oai-snapshot.json`
- inside the file write dummy data, i.e. the exact same json structure as the real data but with only 4-5 examples
- zip the dummy_data folder into dummy_data.zip
- run the test to make sure everything works

```
RUN_SLOW=1 pytest tests/test_dataset_common.py::LocalDatasetTest::test_load_dataset_all_configs_arxiv_dataset
```

if everything worked you can remove the dummy_data directory and only keep the zip file

let me know how it goes :)

PR #1294, opened 2020-12-08
adding opus_euconst

patil-suraj:
Adding EUconst, a parallel corpus collected from the European Constitution.
21 languages, 210 bitexts


lhoestq (approved):
Awesome !

Also thank you for making the dummy data files as small as possible. Since there are many configs I was afraid of how big it could be ^^

PR #1296, opened 2020-12-08
The Snips Built In Intents 2016 dataset.

bduvenhage:
This PR proposes to add the Snips.ai built in intents dataset. The first configuration added is for the intent labels only, but the dataset includes entity slots that may in future be added as alternate configurations.


bduvenhage:
It is not clear how to automatically add the dummy data if the source data is a more complex json format. Should I manually take a fraction of the source data and include it as dummy data?

bduvenhage:
Will tag the dataset and update the dataset card.

PR #1312, opened 2020-12-08
Jigsaw toxicity pred

Tigrex161:
Requires manually downloading data from Kaggle.


datasets/jigsaw_toxicity_pred/jigsaw_toxicity_pred.py
```diff
@@ -0,0 +1,128 @@
...
+            # This defines the different columns of the dataset and their types
+            features=datasets.Features(
+                {
+                    "comment_text": datasets.Value("string"),
+                    "toxic": datasets.Value("float32"),
+                    "severe_toxic": datasets.Value("float32"),
+                    "obscene": datasets.Value("float32"),
+                    "threat": datasets.Value("float32"),
+                    "insult": datasets.Value("float32"),
+                    "identity_hate": datasets.Value("float32"),
```
lhoestq, line 63: can you use the ClassLabel feature type for those fields ?
for example
```
"toxic": datasets.ClassLabel(names=["false", "true"])
```

to make it work you'll have to cast to `int` instead of `float` in _generate_examples

datasets/jigsaw_toxicity_pred/jigsaw_toxicity_pred.py
```diff
@@ -0,0 +1,128 @@
...
+            )
+
+        test_csv_path = os.path.join(data_dir, "test.csv")
+        test_labels_path = os.path.join(data_dir, "test_labels.csv")
+        output_path = os.path.join(data_dir, "test_expanded.csv")
+
+        df1 = pd.read_csv(test_csv_path)
+        df2 = pd.read_csv(test_labels_path)
+        df3 = df1.merge(df2)
+        df3[df3["toxic"] != -1].to_csv(output_path)
```
lhoestq, line 97: please don't create new files.

You can load df1 and df2 to build df3 inside _generate_examples, and without creating files.
To get the paths to the csv files in _generate_examples, you can pass the directory where the files for df1 and df2 in `gen_kwargs`.
for example
```
gen_kwargs={"dara_dir": data_dir}
```

`gen_kwargs` is passed as keyword arguments to _generate_examples

this way in _generate examples you can load df1 and df2 since you have their data_dir, or you can load the train data since again it's in the same data_dir

lhoestq (commented):

I left a few comments

PR #1313, opened 2020-12-08
Add HateSpeech Corpus for Polish

kacperlukawski:
This PR adds a HateSpeech Corpus for Polish, containing offensive language examples.

- **Homepage:** http://zil.ipipan.waw.pl/HateSpeech
- **Paper:** http://www.qualitativesociologyreview.org/PL/Volume38/PSJ_13_2_Troszynski_Wawer.pdf


datasets/hate_speech_pl/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+- `id`: unique identifier of the entry
+- `text_id`: text identifier, useful when a single text is rated several times by different annotators
+- `annotator_id`: identifier of the person who annotated the text
+- `minority_id`: the internal identifier of the minority described in the text
+- `negative_emotions`: boolean indicator of the presence of negative emotions in the text
+- `call_to_action`: boolean indicator set to true, if the text calls the audience to perform any action, typically with a negative emotions
+- `source_of_knowledge`: categorical variable, describing the source of knowledge for the post rating - 0, 1 or 2 (direct, lexical or contextual, but the description of the meaning for different values couldn't be found)
+- `irony_sarcasm`: boolean indicator of the present of irony or sarcasm
+- `topic`: internal identifier of the topic the text is about
```
lhoestq, line 110: what are the available topics ?

datasets/hate_speech_pl/hate_speech_pl.py
```diff
@@ -0,0 +1,112 @@
...
+                {
+                    "id": datasets.Value("uint16"),
+                    "text_id": datasets.Value("uint32"),
+                    "annotator_id": datasets.Value("uint8"),
+                    "minority_id": datasets.Value("uint8"),
+                    "negative_emotions": datasets.Value("bool"),
+                    "call_to_action": datasets.Value("bool"),
+                    "source_of_knowledge": datasets.Value("uint8"),
+                    "irony_sarcasm": datasets.Value("bool"),
+                    "topic": datasets.Value("uint8"),
```
lhoestq, line 71: Can you use the ClassLabel feature type for topic please ?
for example
```
"topic": datasets.ClassLabel(names=[topic_name_for_identifier_0, topic_name_for_identifier_1, etc.])
```

you'll have to regenerate the dataset_infos.json file after this change

lhoestq (commented):
Really cool thanks ! I left a few comments

datasets/hate_speech_pl/README.md
kacperlukawski: Unfortunately the paper doesn't clarify the meaning of the labels. The topic is just some sort of identifier, but its meaning is not described in anywhere. I may put a note on that in README though.

datasets/hate_speech_pl/hate_speech_pl.py
kacperlukawski: Sure, I was wondering if the ClassLabel should be used here. Actually it could have been applied to many different columns as well.

kacperlukawski:
@lhoestq Do you think using the ClassLabel is correct if we don't know the meaning of them?

lhoestq:
Once we find out the meanings we can still add them to the dataset card

lhoestq:
Feel free to ping me when the PR is ready for the final review

lhoestq (approved):
I also did some research and couln't find the labels :/ 

It's ok to leave it this way

Thanks for adding this one

PR #1314, opened 2020-12-08
Add snips built in intents 2016 12

bduvenhage:
This PR proposes to add the Snips.ai built in intents dataset. The first configuration added is for the intent labels only, but the dataset includes entity slots that may in future be added as alternate configurations.


bduvenhage:
It is not clear how to automatically add the dummy data if the source data is a more complex json format. Should I manually take a fraction of the source data and include it as dummy data?

bduvenhage:
Added a fraction of the real data as dummy data.

datasets/snips_built_in_intents/snips_built_in_intents.py
```diff
@@ -0,0 +1,142 @@
...
+
+                        # Option 1 (Unshuffled): Yield samples here if un-shuffled samples ok.
+                        # yield num_examples, {"text": query_text, "label": label}
+
+                        num_examples += 1  # Explicitly keep track of the number of examples.
+
+        # Option 2 (Shuffled): Yield samples here if shuffled samples are required.
+        random.Random(12345).shuffle(
+            samples_txt_lbl
+        )  # Use a new instance of Random to fix shuffle pattern without affecting global rand.
```
lhoestq, line 139: could yo not shuffle the data ?

If the user wants to shuffle they can do `dataset.shuffle()`

lhoestq (commented):
Nice thanks ! good job with the dummy data as well

I just left one comment about shuffling

datasets/snips_built_in_intents/snips_built_in_intents.py
bduvenhage: Hi @lhoestq . I was hoping to have this discussion :-)
Yes, I don't need to shuffle. There is an 'option 1' a little higher up in the code that would then be the way to go. I'll uncomment that option and ping you again.

bduvenhage: @lhoestq You can have a look. If you are happy with the current version then we can remove all of the commented code.

lhoestq (approved):
Looks all good !

Thanks for adding this one :)

lhoestq:

PR #1322, opened 2020-12-08
add indonlu benchmark datasets

yasirabd:
The IndoNLU benchmark is a collection of resources for training, evaluating, and analyzing natural language understanding systems for the Indonesian language. There are 12 datasets in IndoNLU.


datasets/indonlu/README.md
```diff
@@ -0,0 +1,594 @@
+|    | dataset | annotations_creators | language_creators | languages | licenses | multilinguality | size_categories | source_datasets | task_categories      | task_ids                                                  |
```
lhoestq, line 1: The tags must be in YAML, not markdown

Can you fix that please ?

You can use the tagger app to generate the yaml. More info here : https://github.com/huggingface/datasets/blob/master/ADD_NEW_DATASET.md#tag-the-dataset-and-write-the-dataset-card

lhoestq (commented):
That's really cool thanks !

For POS tagging and NER we always use the same column names: 
- `tokens` and `pos_tags` for POS tagging
- `tokens` and `ner_tags` for NER

Can you rename the columns `sentence` and `seq_label` accordingly please ?

You'll need to regenerate the dataset_infos.json file after this change

datasets/indonlu/README.md
yasirabd: Thanks for the review @lhoestq.
I have already made some changes to your review. But, I ended up creating a new PR in here #1530, kindly check for reviews, thank you 🤗

PR #1328, opened 2020-12-08
Added the NewsPH Raw dataset and corresponding dataset card

jcblaisecruz02:
This PR adds the original NewsPH dataset which is used to autogenerate the NewsPH-NLI dataset. Reopened a new PR as the previous one had problems.

Paper: https://arxiv.org/abs/2010.11574
Repo: https://github.com/jcblaisecruz02/Filipino-Text-Benchmarks


lhoestq (approved):
Awesome thanks !

PR #1331, opened 2020-12-08
First version of the new dataset hausa_voa_topics

michael-aloys:
Contains loading script as well as dataset card including YAML tags.


lhoestq (approved):
Nice thank you !

PR #1336, opened 2020-12-08
Add dataset Yoruba BBC Topic Classification

michael-aloys:
Added new dataset Yoruba BBC Topic Classification

Contains loading script as well as dataset card including YAML tags.


lhoestq (approved):
Nice thanks !

PR #1339, opened 2020-12-08
hate_speech_18 initial commit

czabo:


datasets/hate_speech_18/README.md
```diff
@@ -0,0 +1,156 @@
...
+multilinguality:
+- monolingual
+size_categories:
+- 10K<n<100K
+source_datasets:
+- original
+task_categories:
+- text-classification
+task_ids:
+- intent-classification
```
lhoestq, line 18:
```suggestion
---
annotations_creators:
- found
language_creators:
- found
languages:
- en
licenses:
- cc-by-sa-3.0
multilinguality:
- monolingual
size_categories:
- 10K<n<100K
source_datasets:
- original
task_categories:
- text-classification
task_ids:
- intent-classification
---
```

lhoestq (commented):
Nice thanks !

Can you rename the dataset folder and the dataset script name `hate_speech18` instead of `hate_speech_18` to follow the snake case convention we're using ?

Also it looks like the dummy_data.zip file is quite big (almost 4MB).
Can you try to reduce its size ?

To do so feel free to take a look inside it and remove all the unnecessary files or chunks of texts. The idea is to only keep a few examples

czabo:
> Nice thanks !
> 
> Can you rename the dataset folder and the dataset script name `hate_speech18` instead of `hate_speech_18` to follow the snake case convention we're using ?
> 
> Also it looks like the dummy_data.zip file is quite big (almost 4MB).
> Can you try to reduce its size ?
> 
> To do so feel free to take a look inside it and remove all the unnecessary files or chunks of texts. The idea is to only keep a few examples

Done, thanks!

yjernite:
Re-opened in https://github.com/huggingface/datasets/pull/1486

PR #1341, opened 2020-12-08
added references to only data card creator to all guides

yjernite:
We can now use the wonderful online form for dataset cards created by @evrardts


thomwolf (approved):
LGTM

PR #1343, opened 2020-12-08
Add LiveQA

j-chim:
This PR adds LiveQA, the Chinese real-time/timeline-based QA task by [Liu et al., 2020](https://arxiv.org/pdf/2010.00526.pdf).


datasets/liveqa/liveqa.py
```diff
@@ -0,0 +1,128 @@
...
+                gen_kwargs={"filepaths": data_dir, "split": "train"},
+            )
+        ]
+
+    def _generate_examples(self, filepaths, split):
+        """ Yields examples. """
+
+        data_raw = ""
+        for filepath in filepaths:
+            with open(filepath, "r") as f:
```
lhoestq, line 101: To fix the CI, please specify the encoding here, for example `encoding="utf-8"`

lhoestq (commented):
Nice thanks :) 

left one comment

datasets/liveqa/liveqa.py
j-chim: Ah can't believe I missed that! Many thanks for the pointer @lhoestq, fixed now.

lhoestq (approved):
Awesome thanks !

PR #1347, opened 2020-12-08
Add spanish billion words corpus

mariagrandury:
Add an unannotated Spanish corpus of nearly 1.5 billion words, compiled from different resources from the web.


lhoestq (commented):
Awesome thanks !

It looks like the dummy_data.zip file is quite big (70KB), could you try to reduce its size ? To do so feel free to take a look inside it and remove unnecessary files. In particular in the `spanish_billion_words` subfolder you can keep 1 or 2 files with just a few lines of text

mariagrandury:
Thank you for your feedback! I've reduced the dummy data size to 2KB.

I had to rebase to fix `RemoteDatasetTest` fails, sorry about the 80 commits. 
I could create a new clean PR if you prefer.

mariagrandury:
I have seen that in similar cases you have suggested to other contributors to create another branch and another PR, so I will do that.

lhoestq:
Yes thank you !

mariagrandury:
The new PR is #1476 :)

PR #1352, opened 2020-12-09
change url for prachathai67k to internet archive

cstorm125:
`prachathai67k` is currently downloaded from git-lfs of PyThaiNLP github. Since the size is quite large (~250MB), I moved the URL to archive.org in order to prevent rate limit issues.


lhoestq (approved):
thanks :)

PR #1356, opened 2020-12-09
Add StackOverflow StackSample dataset

ncoop57:
This PR adds the StackOverflow StackSample dataset from Kaggle: https://www.kaggle.com/stackoverflow/stacksample

Ran through all of the steps. However, since my dataset requires manually downloading the data, I was unable to run the pytest on the real dataset (the dummy data pytest passed).


datasets/so_stacksample/so_stacksample.py
```diff
@@ -0,0 +1,173 @@
...
+            supervised_keys=None,
+            homepage=_HOMEPAGE,
+            license=_LICENSE,
+        )
+
+    def _split_generators(self, dl_manager):
+        """Returns SplitGenerators."""
+        path_to_manual_file = os.path.join(
+            os.path.abspath(os.path.expanduser(dl_manager.manual_dir)), self.config.name + ".csv"
+        )
```
lhoestq, line 122: can you raise an error here if the file doesn't exist ?

You can find an example here : https://github.com/huggingface/datasets/blob/master/datasets/arxiv_dataset/arxiv_dataset.py#L104-L109

datasets/so_stacksample/README.md
```diff
@@ -0,0 +1,230 @@
...
+### Data Instances
+
+For Answers:
+```json
+{
+  "Id": { # Unique ID given to the Answer post
+    "feature_type": "Value",
+    "dtype": "int32"
+  },
+  "OwnerUserId": { # The UserID of the person who generated the Answer on StackOverflow. -99 means NA
```
lhoestq, line 86: Can you use -1 instead of -99 for missing id ? That's what we use in general

lhoestq (commented):

I left a few comments

Also it looks like the dummy_data.zip files are missing. You should have three of them (one per config). For example `datasets/so_stacksample/dummy/Answers/1.1.0/dummy_data.zip` is missing

Can you add the dummy_data.zip files please ?

ncoop57:
@lhoestq Thanks for the review and suggestions! I've added your comments and pushed the changes. I'm having issues with the dummy data still. When I run the dummy data test

```bash
RUN_SLOW=1 pytest tests/test_dataset_common.py::LocalDatasetTest::test_load_dataset_all_configs_so_stacksample
```
to get the recommended steps needed to add the dataset, but I keep getting this error: 

```
___________________________________________ LocalDatasetTest.test_load_dataset_all_configs_so_stacksample ____________________________________________

self = <tests.test_dataset_common.LocalDatasetTest testMethod=test_load_dataset_all_configs_so_stacksample>, dataset_name = 'so_stacksample'

    @slow
    def test_load_dataset_all_configs(self, dataset_name):
        configs = self.dataset_tester.load_all_configs(dataset_name, is_local=True)
>       self.dataset_tester.check_load_dataset(dataset_name, configs, is_local=True)

tests/test_dataset_common.py:237: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/test_dataset_common.py:198: in check_load_dataset
E   AssertionError: False is not true

FAILED tests/test_dataset_common.py::LocalDatasetTest::test_load_dataset_all_configs_so_stacksample - AssertionError: False is not true
```

I tried formatting the data similar to other datasets, but I think I don't have my csv's in the zip folder with the proper name. I also ran the command that's supposed to outline the exact steps I need to perform to get them into the correct format, but I followed them and they don't seem to be working still :/. Any help would be greatly appreciated!

lhoestq:
Ok I found the issue with the dummy data.
It's currently failing because it's not generating a single example using the dummy csv file.
That's because there's only only line in the dummy csv file, and this line is skipped using the `next()` call used to ignore the headers of the csv.

To fix the dummy data you must add headers to the dummy csv files.

lhoestq:
Also can you make sure that all the original CSV files have headers ?

ncoop57:
> Ok I found the issue with the dummy data.
> It's currently failing because it's not generating a single example using the dummy csv file.
> That's because there's only only line in the dummy csv file, and this line is skipped using the `next()` call used to ignore the headers of the csv.
> 
> To fix the dummy data you must add headers to the dummy csv files.

Oh man, I bamboozled myself! Thank you for catching that! I've updated the dummy csv's to include headers and also confirmed that they all have headers, so I am not throwing away any information with that `next()` call. When I run the test locally for the dummy data it passes, so hopefully it is good to go :D

lhoestq (approved):
Nice thank you !

lhoestq:
merging since the Ci is fixed on master

PR #1364, opened 2020-12-09
Narrative QA (Manual Download Stories) Dataset

rsanjaykamath:
Narrative QA with manual download for stories.


lhoestq:
Hi ! Maybe we can rename it `narrativeqa_manual` to make it explicit that this one requires manual download contrary to `narrativeqa` ?
It's important to have this one as well, in case the `narrativeqa` one suffers from download issues (checksums or dead links for example).

You can also copy the dataset card from `narrativeqa` and add the dummy data as well

rsanjaykamath:
Thanks @lhoestq will do all this and submit a request in the coming days. 😊

rsanjaykamath:
Closing this as another pull request is already done.

PR #1368, opened 2020-12-09
Re-adding narrativeqa dataset

ghomasHudson:
An update of #309.


ghomasHudson:
@lhoestq I think I've fixed the dummy data - it finally passes!

ghomasHudson:
@lhoestq - pretty happy with it now

lhoestq (commented):
Awesome thank you !

Could you try to reduce the size of the dummy_data.zip file before we merge ? (it's 300KB right now)

To do so feel free to take a look inside it and remove all the unnecessary files and chunks of text, to only keep a few examples. The idea is to have a zip file that is only a few KB

ghomasHudson:
> Awesome thank you !
> 
> Could you try to reduce the size of the dummy_data.zip file before we merge ? (it's 300KB right now)
> 
> To do so feel free to take a look inside it and remove all the unnecessary files and chunks of text, to only keep a few examples. The idea is to have a zip file that is only a few KB

Ah, it only contains 1 example for each split. I think the problem is that I include an entire story (like in the full dataset). We can probably get away with a summarised version.

lhoestq (commented):
Nice thank you, can you make it even lighter if possible ? Something round 10KB would be awesone
We try to keep the repo light so that it doesn't take ages to clone. So we have to make sure the dummy data are as small as possible for every single dataset.

ghomasHudson:
> Nice thank you, can you make it even lighter if possible ? Something round 10KB would be awesone
> We try to keep the repo light so that it doesn't take ages to clone. So we have to make sure the dummy data are as small as possible for every single dataset.

Have trimmed a little more out of each example now.

PR #1369, opened 2020-12-09
Use passed --cache_dir for modules cache

albertvillanova:
When passed `--cache_dir` arg:
```shell
python datasets-cli test datasets/<my-dataset-folder> --save_infos --all_configs --cache_dir <my-cache-dir>
```
it is not used for caching the modules, which are cached in the default location at `.cache/huggingface/modules`.

With this fix, the modules will be cached at `<my-cache-dir>/modules`.

### DIFF
--- a/src/datasets/commands/test.py
+++ b/src/datasets/commands/test.py
@@ -93,7 +93,7 @@ class TestCommand(BaseTransformersCLICommand):
             print("Both parameters `config` and `all_configs` can't be used at once.")
             exit(1)
         path, name = self._dataset, self._name
-        module_path, hash = prepare_module(path)
+        module_path, hash = prepare_module(path, cache_dir=self._cache_dir)
         builder_cls = import_main_class(module_path)
 
         if self._all_configs and len(builder_cls.BUILDER_CONFIGS) > 0:
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -35,7 +35,15 @@ from .info import DATASET_INFOS_DICT_FILE_NAME
 from .metric import Metric
 from .splits import Split
 from .utils.download_manager import GenerateMode
-from .utils.file_utils import HF_MODULES_CACHE, DownloadConfig, cached_path, head_hf_s3, hf_bucket_url, hf_github_url
+from .utils.file_utils import (
+    HF_MODULES_CACHE,
+    DownloadConfig,
+    cached_path,
+    create_modules_cache,
+    head_hf_s3,
+    hf_bucket_url,
+    hf_github_url,
+)
 from .utils.filelock import FileLock
 from .utils.logging import get_logger
 from .utils.version import Version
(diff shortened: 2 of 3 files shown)


albertvillanova:
I have a question: why not using a tmp dir instead, like the DummyDataGeneratorDownloadManager does?

src/datasets/commands/test.py
```diff
@@ -93,7 +93,7 @@ def run(self):
             print("Both parameters `config` and `all_configs` can't be used at once.")
             exit(1)
         path, name = self._dataset, self._name
-        module_path, hash = prepare_module(path)
+        module_path, hash = prepare_module(path, cache_dir=self._cache_dir)
```
lhoestq, line 96: The datasets library uses three caches: one to store the datasets data, one to store the metrics data and one to store the modules (either datasets or metrics python scripts).

By default each one of these caches is a directory in the huggingface cache home directory (by default `~/.cache/huggingface`). The three directories are named `datasets`, `metrics` and `modules`. We try to have them separate so that if users ever want to clear the data cache but keeping the modules for example they can do that easily.

In the `test.py` script the `--cache_dir` refers to the `datasets` cache dir since it is passed as `cache_dir` to the dataset builders.
However here `prepare_module` expects the `modules` cache dir.


Maybe in `test.py` we can change `--cache_dir` to be the huggingface cache home directory and then pass `cache_dir / "datasets"` to the builders and `cache_dir / "modules"` to `prepare_modules` ?

lhoestq: One more thing: `prepare_module` also expects the cache dir to be the one for downloaded files (default is `~/.cache/huggingface/datasets/downloads`).

Maybe to make things clearer we should keep `cache_dir` as the cache used to download files and store datasets data files, and instead add a `dynamic_module_path` parameter that defaults to `DYNAMIC_MODULES_PATH`.
I'd like to keep them separate in terms or variable name to avoiding making things confusing.

albertvillanova:
Hi @lhoestq, I am trying to understand better the logic...

Why do we have a `dynamic_module_path` besides the modules cache path?
```python
DYNAMIC_MODULES_PATH = os.path.join(HF_MODULES_CACHE, "datasets_modules")
```
Moreover, 2 subdirectories (for datasets and for metrics) were created inside it:
```python
DATASETS_PATH = os.path.join(DYNAMIC_MODULES_PATH, "datasets")
METRICS_PATH = os.path.join(DYNAMIC_MODULES_PATH, "metrics")
```

lhoestq:
Hi :) 
The modules cache path is the path added to `sys.path`.
Therefore inside we need to have a folder that is going to be a package: `datasets_modules`.
This package will contain dynamic modules, i.e. datasets and metrics modules added on-the-fly.
Then we have two sub-modules `datasets_modules.datasets` and `datasets_modules.metrics`.

Maybe we can make things more explicit in the code with some comments explaining the structure, and maybe better variable naming as well..

Also I wanted to say that I started to work on offline loading of modules in #1726 and actually it lead to do similar changes to what you did to control the path where modules are stored.

albertvillanova:
Hi @lhoestq, I see...

Indeed I was also creating a draft for test_load, to clarify the expected behavior... ;)

So, for the command line:
```sh
python datasets-cli test datasets/<my-dataset-folder> --save_infos --all_configs --cache_dir <my-cache-dir>
```
the `cache_dir` argument refers to dataset cache dir. We do not have control over the modules cache dir, but we would like to have. And if I understand well, you suggest adding another argument `dynamic_module_path`. Am I right?

lhoestq:
> So, for the command line:
> 
> ```shell
> python datasets-cli test datasets/<my-dataset-folder> --save_infos --all_configs --cache_dir <my-cache-dir>
> ```
> 
> the `cache_dir` argument refers to dataset cache dir. We do not have control over the modules cache dir, but we would like to have. And if I understand well, you suggest adding another argument `dynamic_module_path`. Am I right?

Yes the cache_dir is used to download files and also so save the dataset arrow files.
This is indeed different from the path for dynamic modules.

I suggested to have `dynamic_module_path` as a parameter but actually this is the parent directory `hf_modules_cache` that we would need (it's the one that is passed to `init_dynamic_modules ` that we need to add to `sys.path`).

Currently it's already possible to override it using the env variable `HF_MODULES_CACHE` but we can imagine having it as a parameter as well.

This way the user controls both the `cache_dir` and the `hf_modules_cache` which are the two places used by the library to read/write stuff.

lhoestq:
I think #1726 is going to be merged pretty soon. Maybe can work on this as soon as it's merged to avoid doing the same things twice and to avoid conflicts ?

albertvillanova:
I agree. Indeed I took some of your code in one of my last commit, to try to implement the logic you described.

tests/data/datasets/dummy_dataset/dummy_dataset.py
```diff
@@ -0,0 +1,3 @@
+"""Dummy dataset."""
```
lhoestq, line 1: this file is not used anywhere ?
maybe you can remove it

lhoestq (commented):
This does will download the dataset script inside the cache dir, but note that the modules cache isn't changed (still uses the default value)
The module directory is controlled by the `dynamic_modules_path` argument. 

Did you want to change the modules cache dir or simply make the download cache be at `cache_dir` ?

PR #1375, opened 2020-12-09
Add OPUS EMEA Dataset

abhishekkrthakur:


datasets/emea/emea.py
```diff
@@ -0,0 +1,120 @@
...
+                x = x.strip()
+                y = y.strip()
+                result = (
+                    sentence_counter,
+                    {
+                        "id": str(sentence_counter),
+                        "translation": {l1: x, l2: y},
+                    },
+                )
+                sentence_counter += 1
```
lhoestq, line 119:
```suggestion
```

lhoestq (approved):
Thanks ! LGTM

datasets/emea/emea.py
sick78: i hope that i get paid to image

sick78: image

PR #1385, opened 2020-12-09
add best2009

cstorm125:
`best2009` is a Thai word-tokenization dataset from encyclopedia, novels, news and articles by [NECTEC](https://www.nectec.or.th/) (148,995/2,252 lines of train/test). It was created for [BEST 2010: Word Tokenization Competition](https://thailang.nectec.or.th/archive/indexa290.html?q=node/10). The test set answers are not provided publicly.


datasets/best/best.py
```diff
@@ -0,0 +1,140 @@
...
+                gen_kwargs={"filepath": os.path.join(data_dir, self._TRAIN_FOLDER), "split": "train"},
+            ),
+            datasets.SplitGenerator(
+                name=datasets.Split.TEST,
+                gen_kwargs={"filepath": os.path.join(data_dir, self._TEST_FOLDER), "split": "train"},
+            ),
+        ]
+
+    def _generate_examples(self, filepath, split):
+        for fname in Path(filepath).rglob("*.txt"):
```
lhoestq, line 119:
```suggestion
        for fname in sorted(Path(filepath).rglob("*.txt")):
```
we need to make the order deterministic

lhoestq (commented):
Thanks for adding his one !

Can we use a less generic name for the dataset ? Maybe best2010 ?
We would need to rename the folder and the dataset script with the new name.

The dataset class name must also be the camel case version of the dataset script name si it would be Best2010.

Also it looks like the dummy_data.zip file is quite big ( 300KB), could you try to reduce its size ?

To do so feel free to take a look inside and remove all the unnecessary files to only keep 1 or 2 files with just a few examples in them

datasets/best/best.py
cstorm125: Thanks for the review. I think I'll go with `best2009` to conform with the citations. Will reduce size of dummy_data.zip and do sorted file names as well.

lhoestq (approved):
Thank you !

PR #1393, opened 2020-12-09
Add script_version suggestion when dataset/metric not found

joeddav:
Adds a helpful prompt to the error message when a dataset/metric is not found, suggesting the user might need to pass `script_version="master"` if the dataset was added recently. The whole error looks like:

> Couldn't find file locally at blah/blah.py, or remotely at https://raw.githubusercontent.com/huggingface/datasets/1.1/metrics/blah/blah.py or https://s3.amazonaws.com/datasets.huggingface.co/datasets/met
rics/blah/blah.py.
If the dataset was added recently, you may need to to pass script_version="master" to find the loading script on the master branch.

### DIFF
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -278,9 +278,9 @@ def prepare_module(
                 local_path = cached_path(file_path, download_config=download_config)
             except FileNotFoundError:
                 raise FileNotFoundError(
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
Thanks !

PR #1399, opened 2020-12-09
Add HoVer Dataset

abhishekkrthakur:
HoVer: A Dataset for Many-Hop Fact Extraction And Claim Verification
https://arxiv.org/abs/2011.03088


datasets/hover/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Languages
+
+[More Information Needed]
+
+## Dataset Structure
+
+### Data Instances
+
+```
```
lhoestq, line 72: We expect here to see an example of a sample from the dataset.

Can you move the sample training set you provided below in this section please ?

The other section Data Fields expects just a brief description of the different fields of the data (claim, supporting_facts, etc.)

datasets/hover/hover.py
```diff
@@ -0,0 +1,111 @@
...
+                    "id": datasets.Value("int32"),
+                    "uid": datasets.Value("string"),
+                    "claim": datasets.Value("string"),
+                    "supporting_facts": [
+                        {
+                            "key": datasets.Value("string"),
+                            "value": datasets.Value("int32"),
+                        }
+                    ],
+                    "label": datasets.Value("string"),
```
lhoestq, line 57: can you use the ClassLabel feature type here ?

for example
```
"label": datasets.ClassLabel(names=["NOT_SUPPORTED", "SUPPORTED])
```
when the label like in the test set, you need to replace the value "None" with -1

you'll have to regenerate the dataset_infos.json file after this change

datasets/hover/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Data Fields
+
+A sample training set is provided below
+
+```
+{'id': 14856, 'uid': 'a0cf45ea-b5cd-4c4e-9ffa-73b39ebd78ce', 'claim': 'The park at which Tivolis Koncertsal is located opened on 15 August 1843.', 'supporting_facts': [{'key': 'Tivolis Koncertsal', 'value': 0}, {'key': 'Tivoli Gardens', 'value': 1}], 'label': 'SUPPORTED', 'num_hops': 2, 'hpqa_id': '5abca1a55542993a06baf937'}
+```
+
+Please note that in test set sentence only id, uid and claim are available.
```
lhoestq, line 97: Can you mention that you replace the missing labels in the test set with a default value ? (probably -1 if you switch to ClassLabel)

lhoestq (commented):
This one is really cool thanks !

I left a few comments:

abhishekkrthakur:
@lhoestq all comments addressed :)

datasets/hover/hover.py
```diff
@@ -0,0 +1,111 @@
...
+                    "label": d["label"],
+                    "num_hops": d["num_hops"],
+                    "hpqa_id": d["hpqa_id"],
+                }
+            else:
+                resp = {
+                    "id": sentence_counter,
+                    "uid": d["uid"],
+                    "claim": d["claim"],
+                    "supporting_facts": [{"key": "None", "value": -1}],
```
lhoestq, line 106:
```suggestion
                    "supporting_facts": [],
```
an empty list does the job

lhoestq (approved):
Thanks !

lhoestq:

PR #1405, opened 2020-12-09
Adding TaPaCo Dataset with README.md

pacman100:


lhoestq (commented):
Really cool thanks ! Looks all good to me :) 

Before we merge can you try to reduce the size of the dummy_data.zip files ? There are many of them and they're all around 20KB. I you could make each one of them less than 1 or 2 KB it would be perfect

To do that feel free to take a look inside them and remove all the unnecessary files. For example for the dummy_data.zip file for the config `en` you can remove all the text files of the other configurations `es`, `fr` etc.
(I would also suggest to use some code to automate that).

For the `all_configs` dummy_data.zip file you can keep it this way

lhoestq:
We want to keep the repo as light as possible so that it doesn't take ages to clone, that's why we ask for small dummy data files (especially when there are many of them). Let me know if you have questions or if we can help you on this

pacman100:
Hello @lhoestq , made the changes as you suggested and pushed, please review. By default, the dummy data was generated the way it was by the dummy data auto generate command. Thank you.

lhoestq (approved):
Thank you !
Looks all good now :)

PR #1410, opened 2020-12-09
Add penn treebank dataset

harshalmittal4:


datasets/penn_treebank/README.md
```diff
@@ -0,0 +1,150 @@
...
+multilinguality:
+- monolingual
+size_categories:
+- n<1K
+source_datasets:
+- original
+task_categories:
+- structure-prediction
+task_ids:
+- structure-prediction-other-POS Tagging
```
lhoestq, line 18:
```suggestion
---
annotations_creators:
- expert-generated
language_creators:
- found
languages:
- en
licenses:
- other-LDC User Agreement for Non-Members
multilinguality:
- monolingual
size_categories:
- n<1K
source_datasets:
- original
task_categories:
- structure-prediction
task_ids:
- structure-prediction-other-POS Tagging
---
```

datasets/penn_treebank/penn_treebank.py
```diff
@@ -0,0 +1,152 @@
...
+    BUILDER_CONFIGS = [
+        PennTreebankConfig(
+            name="penn_treebank",
+            version=VERSION,
+            description="Load the Penn Treebank dataset",
+        ),
+    ]
+
+    def _info(self):
+        features = datasets.Features({"tokens": datasets.Value("string")})
```
lhoestq, line 101: this dataset is supposed to be for POS tagging but I see no POS tagging features.

Is this expected ?

lhoestq (commented):
Nice thanks !

I left a comment about the features

datasets/penn_treebank/penn_treebank.py
harshalmittal4: Hello @lhoestq , currently it outputs the word level tokens from raw text. 
For the pos-tagging task, I was not able to find the pos-tags for this dataset. In practice, the dataset is also used for word level language modelling task. So is this fine?

lhoestq: Ok I see !
Can you mention that in the dataset card in this case ? i.e. say that currently the POS tags are missing and it can only be used for word level language modeling

You can also add language modeling to the task field of the YAML tags of the dataset card

harshalmittal4: I will add the feature details of the dataset in the README to make this explicit if it's fine.

harshalmittal4: Thanks, will do these both soon and update the PR! I will also add some more details in the card.

datasets/penn_treebank/README.md
yjernite, line 18: this should be sequence modeling and language modeling since there aren't any POS tags

datasets/penn_treebank/penn_treebank.py
```diff
@@ -0,0 +1,152 @@
...
+        # It is in charge of opening the given file and yielding (key, example) tuples from the dataset
+        # The key is not important, it's more here for legacy reason (legacy from tfds)
+        eos_token = "</s>"
+        with open(filepath, encoding="utf-8") as f:
+            for line in f:
+                a = line.split()
+                a.append(eos_token)
+                for id_, word in enumerate(a):
+                    feat = dict()
+                    feat["tokens"] = word
```
yjernite, line 151:
```suggestion
                    feat["tokens"] = feat.get("tokens", []) + [word]
```
Make sure to add all the words in the line instead of just the last one

yjernite (changes requested):
I added a couple of comments to fix some bugs in the code.

More importantly, this should be renamed to `ptb_text_only` (and `PtbTextOnly`)

Having the text can definitely be useful, but we should warn people up front that this version does not have the POS tags or the dependency parses.

Also, please mention whether this version has the full vocabulary or the rare words already replaces with UNK

datasets/penn_treebank/README.md
harshalmittal4: updated

datasets/penn_treebank/penn_treebank.py
harshalmittal4: done

harshalmittal4:
@yjernite I have updated the PR to be language modeling task specific. Please review!

datasets/ptb_text_only/ptb_text_only.py
```diff
@@ -0,0 +1,152 @@
...
+        # The key is not important, it's more here for legacy reason (legacy from tfds)
+        eos_token = "</s>"
+        with open(filepath, encoding="utf-8") as f:
+            for line in f:
+                a = line.split()
+                a.append(eos_token)
+                feat = dict()
+                for id_, word in enumerate(a):
+                    feat["tokens"] = feat.get("tokens", []) + [word]
+                    yield id_, feat
```
lhoestq, line 152: Actually, do we really want to process the dataset this way ?
I think the processing depends on the task you want to do with this dataset. It's to the user to define and apply its processing.
We can just have one example = one line (as a string) instead of yielding partial sentences for each word.

harshalmittal4: Since the dataset can mainly be used for word level language modelling (as the pos-tags are missing currently), for example if the line is "This is the PTB dataset", the datapoints will be: 
1. This
2. This is 
3. This is the
4. This is the PTB
.. and so on for each line in dataset.
Must I change it?

yjernite: The input isn't the same for all models, and we try to be application-agnostic in the preprocessing, so yes we do want one line per item as requested by @lhoestq :)

harshalmittal4: Sure, I will update the PR

harshalmittal4: Sure, I have updated the PR

yjernite (approved):
Does one line really correspond to a sentence or should the field just be called `line`?

Aside from that, good with the changes.

lhoestq (approved):
Awesome thanks !

PR #1413, opened 2020-12-09
Add OffComBR

hugoabonizio:
Add [OffComBR](https://github.com/rogersdepelle/OffComBR) from [Offensive Comments in the Brazilian Web: a dataset and baseline results](https://sol.sbc.org.br/index.php/brasnam/article/view/3260/3222) paper.

But I'm having a hard time generating dummy data since the original dataset extion is `.arff` and the [_create_dummy_data function](https://github.com/huggingface/datasets/blob/a4aeaf911240057286a01bff1b1d75a89aedd57b/src/datasets/commands/dummy_data.py#L185) doesn't allow it.


SBrandeis:
Hello @hugoabonizio, thanks for the contribution.
Regarding the fake data, you can generate it manually.
Running the `python datasets-cli dummy_data datasets/offcombr` should give you instructions on how to manually create the dummy data.
For reference, here is a spec for `.arff` files : https://www.cs.waikato.ac.nz/ml/weka/arff.html

datasets/offcombr/offcombr.py
```diff
@@ -0,0 +1,110 @@
...
+_LICENSE = "Unknown"
+
+_URLs = {
+    "offcombr-2": "https://github.com/rogersdepelle/OffComBR/raw/master/OffComBR2.arff",
+    "offcombr-3": "https://github.com/rogersdepelle/OffComBR/raw/master/OffComBR3.arff",
+}
+
+
+# TODO: Name of the dataset usually match the script name with CamelCase instead of snake_case
+class NewDataset(datasets.GeneratorBasedBuilder):
```
lhoestq, line 46:
```suggestion
class Offcombr(datasets.GeneratorBasedBuilder):
```

datasets/offcombr/offcombr.py
```diff
@@ -0,0 +1,110 @@
...
+    DEFAULT_CONFIG_NAME = (
+        "offcombr-2"  # It's not mandatory to have a default configuration. Just use one if it make sense.
+    )
+
+    def _info(self):
+        return datasets.DatasetInfo(
+            description=_DESCRIPTION,
+            features=datasets.Features(
+                {
+                    "label": datasets.Value("string"),
```
lhoestq, line 73: Can you use the ClassLabel feature type for the label field please ?
For example
```
"label": datasets.ClassLabel(names=["no", "yes"])
```

datasets/offcombr/offcombr.py
```diff
@@ -0,0 +1,110 @@
...
+        datasets.BuilderConfig(
+            name="offcombr-3",
+            version=VERSION,
+            description="OffComBR-3 only contains instances for which the class has been agreed by all three judges",
+        ),
+    ]
+
+    DEFAULT_CONFIG_NAME = (
+        "offcombr-2"  # It's not mandatory to have a default configuration. Just use one if it make sense.
+    )
```
lhoestq, line 66: why have this one as default ?

lhoestq (commented):
Really cool thanks !

I left a few comments.
After the change of class name and feature type you'll need to regenerate the dataset_infos.json file.

Also it looks like the dummy data test is not passing, can you take a look at what's wrong ?
As mentioned by Simon you can show some instructions about the creation of the dummy data using the CLI.

Let us know if you have any question or if we can help you with the dummy data

datasets/offcombr/offcombr.py
hugoabonizio: `offcombr-3` is more strict than `offcombr-2`, thus it has less samples. `offcombr-2` is presented first on the original paper, but there is no strict default version to be honest.

hugoabonizio:
@lhoestq again the failing tests doesn't seem to be related

datasets/offcombr/offcombr.py
```diff
@@ -0,0 +1,107 @@
...
+            description="OffComBR-2 contains samples with Fleiss Kappa measure between annotators of 0.71",
+        ),
+        datasets.BuilderConfig(
+            name="offcombr-3",
+            version=VERSION,
+            description="OffComBR-3 only contains instances for which the class has been agreed by all three judges",
+        ),
+    ]
+
+    DEFAULT_CONFIG_NAME = "offcombr-2"
```
lhoestq, line 63:
```suggestion
```
Let's keep no default config here and let the user choose

lhoestq (approved):
Thanks !

lhoestq:

PR #1429, opened 2020-12-09
extract rar files

zaidalyafeai:
Unfortunately, I didn't find any native python libraries for extracting rar files. The user has to manually install `sudo apt-get install unrar`. Discussion with @yjernite is in the slack channel.

### DIFF
--- a/src/datasets/utils/file_utils.py
+++ b/src/datasets/utils/file_utils.py
@@ -368,12 +368,9 @@ def cached_path(
                     with open(output_path_extracted, "wb") as extracted_file:
                         shutil.copyfileobj(compressed_file, extracted_file)
             elif rarfile.is_rarfile(output_path):
-                try:
-                    rf = rarfile.RarFile(output_path)
-                    rf.extractall(output_path_extracted)
-                    rf.close()
-                except:
-                    raise EnvironmentError("Please apt-install unrar")
+                rf = rarfile.RarFile(output_path)
+                rf.extractall(output_path_extracted)
+                rf.close()
             else:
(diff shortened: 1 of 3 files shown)


src/datasets/utils/file_utils.py
```diff
@@ -22,6 +22,7 @@
 from zipfile import ZipFile, is_zipfile
 
 import numpy as np
+import rarfile
```
lhoestq, line 25: This import would make the library impossible to import for people that don't have `rarfile` installed.

To make it an optional import you can do the same as what we did for pytorch at the top of the `file_utils.py` file.

The idea is to have a boolean `_rarfile_available` that is True when `rarfile` is installed. Then before any call to `rarfile` you first have to check if it is avaibale using `_rarfile_available`.

lhoestq (commented):
Thanks !

I just left a comment about making rarfile an optional dependency

src/datasets/utils/file_utils.py
zaidalyafeai: Thanks @lhoestq , I also added `rarfile` to `setup.py` file, does that make sense or I should remove it from there ?

zaidalyafeai: Refactored everything. I tested with and without `rarfile` and it seems to work, also tested on a dataset that I am adding with rar files. I also added an exception if the file is `rar` but the user doesn't have `rarfile` installed.

lhoestq: Nice thanks !

And yes we should have rarfile in the `tests` requirements, thanks for adding it

lhoestq (approved):
Awesome thank you !

PR #1434, opened 2020-12-10
add_sofc_materials_articles

ZacharySBrown:
adding [SOFC-Exp Corpus](https://arxiv.org/abs/2006.03039)


datasets/sofc_materials_articles/README.md
```diff
@@ -0,0 +1,206 @@
+
```
lhoestq, line 1:
```suggestion
```

lhoestq (commented):
Impressive work on this one, good job !

Can you add comments in the _generate_examples methods to explain a bit each step of the processing of the .txt files ?

Also it looks like the dummy_data.zip file is quite big (170KB), could you try to reduce its size ?
To do so feel free to only keep 1 or 2 text files and remove unnecessary chunks of texts

ZacharySBrown:
Hey @lhoestq , thanks for the feedback on this! I updated the `_generate_examples` with some comments on the process, and reduced the `dummy_data.zip` down quite a bit as well. 

For the dummy data, I reduced the text to only three sentences, and aligned the corresponding entity/token/sentence annotations to that (reduced accordingly). The frames file is a strange combined format for the annotations and I found if I reduced that that would break the parser no matter what I did, so I left that as is. The difference between a reduced frames and non-reduced frames file in the compressed dummy data was only about ~4kb, so hopefully leaving this as is will be ok!

datasets/sofc_materials_articles/sofc_materials_articles.py
```diff
@@ -0,0 +1,470 @@
...
+            "begin_char_offset",
+            "end_char_offset",
+            "entity_label",
+            "slot_label",
+        ]
+
+        # Start the processing loop
+        # For each text file, we'll load all of the
+        # associated annotation files
+        for id_, name in enumerate(names):
```
lhoestq, line 311:
```suggestion
        for id_, name in enumerate(sorted(names)):
```
we need to make the order deterministic here

lhoestq (approved):
Awesome thanks !

PR #1439, opened 2020-12-10
Update README.md

tuner007:
1k-10k -> 1k-1M

3 separate configs are available with min. 1K and max. 211.3k examples


lhoestq (approved):
thanks !

PR #1442, opened 2020-12-10
Create XML dummy data without loading all dataset in memory

albertvillanova:
While I was adding one XML dataset, I noticed that all the dataset was loaded in memory during the dummy data generation process (using nearly all my laptop RAM).

Looking at the code, I have found that the origin is the use of `ET.parse()`. This method loads **all the file content in memory**.

In order to fix this, I have refactorized the code and use `ET.iterparse()` instead, which **parses the file content incrementally**.

I have also implemented a test.


thomwolf (approved):
Nice!

lhoestq (approved):
Really nice thanks !

PR #1451, opened 2020-12-10
Add European Center for Disease Control and Preventions's (ECDC) Translation Memory dataset

SBrandeis:
ECDC-TM homepage: https://ec.europa.eu/jrc/en/language-technologies/ecdc-translation-memory


lhoestq (approved):
Awesome thank you !

PR #1463, opened 2020-12-10
Adding enriched_web_nlg features + handling xml bugs

TevenLeScao:
This PR adds features of the enriched_web_nlg dataset that were not present yet (most notably sorted rdf triplet sets), and deals with some xml issues that led to returning no data in cases where surgery could be performed to salvage it.


lhoestq (approved):
Thanks !

PR #1471, opened 2020-12-11
Adding the HAREM dataset

jonatasgrosman:
Adding the HAREM dataset, a Portuguese language dataset for NER tasks


datasets/harem/harem.py
```diff
@@ -0,0 +1,279 @@
...
+# limitations under the License.
+"""HAREM dataset"""
+
+from __future__ import absolute_import, division, print_function
+
+import json
+import logging
+from typing import List, Tuple
+
+from transformers.tokenization_utils import _is_punctuation, _is_whitespace
```
lhoestq, line 23: Maybe we can avoid the dependency on transformers just for those two functions ?

Feel free to copy paste them in the script from
https://github.com/huggingface/transformers/blob/master/src/transformers/tokenization_utils.py#L77-L89
and
https://github.com/huggingface/transformers/blob/master/src/transformers/tokenization_utils.py#L53-L62

lhoestq (commented):
Awesome !
You did an incredible job with the tokenization here to match the NER dataset format.

Can you also add one config for the raw default and raw selective please ?
This is just in case some users want to use the data in their raw format

datasets/harem/harem.py
jonatasgrosman: Hi @lhoestq, I've just made the adjustments you've proposed

lhoestq:
Thanks for the changes !

Sorry if I wasn't clear about the suggestion of adding the `raw` dataset as well.
By `raw` I meant the dataset with its original features, i.e. not tokenized to follow the conll format for NER.
The `raw` dataset has data fields `doc_text`, `doc_id` and `entities`.

jonatasgrosman:
Alright, @lhoestq, now I understand your suggestion, but the JSON files the script downloads aren't actually the "raw" HAREM dataset, the [real raw version](https://www.linguateca.pt/primeiroHAREM/harem/ColeccaoDouradaHAREM.zip) of it is on XML format that needs a lot of preprocessing. Those JSON are just pre-processed versions of the dataset.

I can make the config of the raw version of the pre-processed dataset, but I'll leave a comment on the dataset summary about those details.

lhoestq:
Oh I see ! Sorry I thought there really were the raw (i.e. original/unprocessed) files

lhoestq:
The original xml format doesn't seem very practical to use :/ 
We can ignore that and only keep the default and selective configs then.
Can you revert to the old configs ?

Thanks again for adding this dataset !

jonatasgrosman:
Hi @lhoestq, I've reverted the changes and put more details on README about the dataset.

lhoestq (approved):
Awesome thank you !

PR #1474, opened 2020-12-11
Create JSON dummy data without loading all dataset in memory

albertvillanova:
See #1442.

The statement `json.load()` loads **all the file content in memory**.

In order to avoid this, file content should be parsed **iteratively**, by using the library `ijson` e.g.

I have refactorized the code into a function `_create_json_dummy_data` and I have added some tests.


setup.py
```diff
@@ -62,6 +62,8 @@
 DOCLINES = __doc__.split('\n')
 
 REQUIRED_PKGS = [
+    # For iterative parsing JSON files
+    "ijson",
```
thomwolf, line 66: We are very careful about adding new dependencies. I think this one could be in the `tests` or `dev` deps only.

thomwolf (commented):
This is nice and useful for very large datasets, and thanks for adding tests as well!

Just a comment about the dependencies.

setup.py
albertvillanova: @thomwolf, done! I totally agree that one have to be cautious about adding new dependencies. The ideal option would have been that the Python Standard Library `json` package has an `iterload()` method (just as `xml.etree.ElemntTree` has `iterparse()`). But this is not the case, and thus the need to add an external dependency.

However, I would like to point out that when I decided to add it as a core dependency, I was thinking that maybe it should (or may) be used by dataset scripts (in `_generate_examples()` method) when dealing with large JSON files... Just keep this in mind! :)

src/datasets/commands/dummy_data.py
```diff
@@ -8,6 +8,8 @@
 from pathlib import Path
 from typing import Optional
 
+import ijson
```
lhoestq, line 11: ijson is an optional dependency so this import would crash for the users that don't have ijson

Can you move the import inside the `_create_json_dummy_data` method and raise an error if it's not installed ?
The error should also tell the user to install ijson with `pip install ijson` to be able to create dummy json files.

lhoestq (commented):
Nice thanks :) 

I left a comment about the ijson import

src/datasets/commands/dummy_data.py
albertvillanova: @lhoestq done. I have also fixed the conflicts it had with the master branch.

PR #1475, opened 2020-12-11
Fix XML iterparse in opus_dogc dataset

albertvillanova:
I forgot to add `elem.clear()` to clear the element from memory.


lhoestq (approved):
Thanks !

PR #1476, opened 2020-12-11
Add Spanish Billion Words Corpus

mariagrandury:
Add an unannotated Spanish corpus of nearly 1.5 billion words, compiled from different resources from the web.


lhoestq (approved):
Awesome thanks !

PR #1482, opened 2020-12-11
Adding medical database chinese and english

vrindaprabhu:
Error in creating dummy dataset


lhoestq (commented):
We need to have a dummy_data.zip file in ./datasets/medical_dialog/dummy/en/1.0.0/ and also one in ./datasets/medical_dialog/dummy/zh/1.0.0/

To create the one in ./datasets/medical_dialog/dummy/en/1.0.0/, first go in this folder and do as follows:
- create a folder named `dummy_data`
- create one dummy text file inside the `dummy_data` folder containing a few examples (generally 4-5 examples). The examples must be formatted exactly like in the real text files.
- zip the `dummy_data` folder into dummy_data.zip

Then you can do the same for ./datasets/medical_dialog/dummy/zh/1.0.0/

Finally to make sure everything works as expected you can run 
```
RUN_SLOW=1 pytest tests/test_dataset_common.py::LocalDatasetTest::test_load_dataset_all_configs_medical_dialog
```

lhoestq:
Let me know it that helps !
Also feel free to ping me if you have other questions or if I can help you.

vrindaprabhu:
Now I am getting an Assertion Error!
image

vrindaprabhu:
All tests have passed. However, PyTest is still failing with the `AssertionError` as before. Also _datasets_info.json_ actually does not seem to provide much info. Please review and let me know what has to be improved.

Thanks!

datasets/medical_dialog/medical_dialog.py
```diff
@@ -0,0 +1,280 @@
...
+                                        conv_flag = True
+
+                                # Continues till the next ID is parsed
+                                if conv_flag:
+                                    sen = line.rstrip()
+                                    if sen == "":
+                                        continue
+
+                                    if sen[-1] not in ".。？，,?!！~～":
+                                        sen += "."
```
lhoestq, line 272: I don't think we should re-add the dots here.
The processing of the dataset is left to the user, here we're just supposed to parse the data files

datasets/medical_dialog/medical_dialog.py
```diff
@@ -0,0 +1,280 @@
...
+                                if data_lang == "en":
+                                    last_user = "Patient"
+                                    sen = line.rstrip()
+                                    des_flag = True
+
+                                if des_flag:
+                                    if sen == "":
+                                        continue
+                                    if sen[-1] not in ".。？，,?!！~～":
+                                        sen += "."
```
lhoestq, line 210: same here

lhoestq (commented):
That's really cool !

Good job with all the parsing.
I just left a comment about additional processing you did

datasets/medical_dialog/medical_dialog.py
vrindaprabhu: I removed the two pre-processing lines but now the circleci tests are failing.

vrindaprabhu:
[PR-1503](https://github.com/huggingface/datasets/pull/1503) on the COVID dialog dataset from the same University has similar features. I kept it separate because it is only on COVID qa. Kindly let me know if it has to be added as two more configurations to this existing dataset itself.

lhoestq:
It's ok to have them separate

lhoestq (approved):
Thank you !

Looks all good now :)

PR #1485, opened 2020-12-11
Re-added wiki_movies dataset due to previous PR having changes from m…

aclifton314:
…any other unassociated files.


lhoestq (approved):
Thanks !

PR #1496, opened 2020-12-12
Add Multi-Dimensional Gender Bias classification data

yjernite:
https://parl.ai/projects/md_gender/

Mostly has the ABOUT dimension since the others are inferred from other datasets in most cases.

I tried to keep the dummy data small but one of the configs has 140 splits ( > 56KB data)


lhoestq (approved):
Really cool one ! thanks :)

PR #1517, opened 2020-12-12
Kd conv smangrul

pacman100:


ijaz314:
Hii please follow me

lhoestq (commented):
Really cool thanks !

Before we merge, can you try to reduce the size of the dummy_data.zip files ?
Ideally they should all be below 10KB

To do so feel free to take a look inside them and remove unnecessary files and chunks of text.
Especially if there are long dialogues you can keep only the few sentences. Same for knowledge bases: only keep less than 10 entries if possible

lhoestq (approved):
Awesome thank you !

lhoestq:

PR #1521, opened 2020-12-12
Atomic

ontocord:
This is the ATOMIC common sense dataset. More info can be found here:

* README.md still to be created.


ontocord:
I had to create a new PR to fix git errors. See: https://github.com/huggingface/datasets/pull/1525

I'm closing this PR.

PR #1524, opened 2020-12-12
ADD: swahili dataset for language modeling

akshayb7:
Add a corpus for Swahili language modelling. All tests passed locally. README updated with all information available.


lhoestq (approved):
Awesome thank you !

PR #1462, opened 2020-12-10
Added conv ai 2 (Again)

rkc007:
The original PR -> https://github.com/huggingface/datasets/pull/1383

Reason for creating again - 

The reason I had to create the PR again was due to the master rebasing issue. After rebasing the changes, all the previous commits got added to the branch.

PR #1383, opened 2020-12-09
added conv ai 2

rkc007:
Dataset : https://github.com/DeepPavlov/convai/tree/master/2018

PR #1527, opened 2020-12-13
Add : Conv AI 2 (Messed up original PR)

rkc007:
@lhoestq Sorry I messed up the previous 2 PR's -> https://github.com/huggingface/datasets/pull/1462 -> https://github.com/huggingface/datasets/pull/1383. So created a new one. Also, everything is fixed in this PR. Can you please review it ?
Thanks in advance.


datasets/conv_ai_2/conv_ai_2.py
```diff
@@ -0,0 +1,86 @@
...
+        ]
+
+    def _generate_examples(self, filepath):
+        with open(filepath, encoding="utf-8") as f:
+            examples = json.load(f)
+            for example in examples:
+                example["id"] = example["dialog_id"]
+                if example["eval_score"] is None:
+                    example["eval_score"] = -1  # missing eval_score is replaced with -1
+                if not example["profile_match"]:
```
lhoestq, line 84:
```suggestion
                if example["profile_match"] == "":
```

Please use more robust checks.

if `example["profile_match"]` is 0, then `bool(not example["profile_match"])` is True and therefore all zeros are replaced with -1, which is not what we want.

lhoestq (approved):
Really cool thanks !

PR #1537, opened 2020-12-13
added ohsumed 

skyprince999:
UPDATE2: PR passed all tests. Now waiting for review.

UPDATE: pushed  a new version. cross fingers that it should complete all the tests! :) 
               If it passes all tests then it's not a draft version. 

This is a draft version


datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,167 @@
...
+
+# TODO: Add link to the official dataset URLs here
+# The HuggingFace dataset library don't host the datasets but only point to the original files
+# This can be an arbitrary nested dict/list of URLs (see below in `_split_generators` method)
+_URLs = {"ohsumed": "https://storage.googleapis.com/corona-tweet/ohsumed.zip"}
+
+
+# TODO: Name of the dataset usually match the script name with CamelCase instead of snake_case
+class Ohsumed(datasets.GeneratorBasedBuilder):
+    """TODO: Short description of my dataset."""
```
SBrandeis, line 70: Missing short description of the dataset here.

You can reuse the one l.15

datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,167 @@
...
+    # If you need to make complex sub-parts in the datasets with configurable options
+    # You can create your own builder configuration class to store attribute, inheriting from datasets.BuilderConfig
+    # BUILDER_CONFIG_CLASS = MyBuilderConfig
+
+    # You will be able to load one or the other configurations in the following list with
+    # data = datasets.load_dataset('my_dataset', 'first_domain')
+    # data = datasets.load_dataset('my_dataset', 'second_domain')
+    BUILDER_CONFIGS = [
+        datasets.BuilderConfig(
+            name="ohsumed", version=VERSION, description="This part of my dataset covers a first domain"
```
SBrandeis, line 87: Don't forget to update the descritpion here too

datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,167 @@
...
+        )
+    ]
+
+    def _info(self):
+        # TODO: This method specifies the datasets.DatasetInfo object which contains informations and typings for the dataset
+        features = datasets.Features(
+            {
+                "seq_id": datasets.Value("string"),
+                "medline_ui": datasets.Value("string"),
```
SBrandeis, line 96: Are these two field numbers ? If yes, it would be better to use `datasets.Value("int64")`

datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,167 @@
...
+                "mesh_terms": datasets.Value("string"),
```
SBrandeis, line 97: Can this be a `ClassLabel` feature ?

datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,167 @@
...
+                "title": datasets.Value("string"),
+                "publication_type": datasets.Value("string"),
```
SBrandeis, line 99: Can this be a `ClassLabel` feature ?

datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,167 @@
...
+                gen_kwargs={"filepath": os.path.join(data_dir, "test.jsonl"), "split": "test"},
+            ),
+        ]
+
+    def _generate_examples(self, filepath, split):
+        """ Yields examples. """
+        # TODO: This method will receive as arguments the `gen_kwargs` defined in the previous `_split_generators` method.
+        # It is in charge of opening the given file and yielding (key, example) tuples from the dataset
+        # The key is not important, it's more here for legacy reason (legacy from tfds)
+        print(filepath)
```
SBrandeis, line 153: I'd suggest removing this print
```suggestion
```

datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,167 @@
...
+        with open(filepath, encoding="utf-8") as f:
+            abstracts = json.load(f)
+            for id_, data in enumerate(abstracts):
```
SBrandeis, line 156: Data seems to be in `jsonl` format, I would recommend using the following syntax to avoid load all the file in memory at once and take advantage of the generator pattern:
```suggestion
        with open(filepath, encoding="utf-8") as f:
            for id_, line in enumerate(abstracts):
                data = json.loads(line)
```

datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,167 @@
...
+                # data = json.load(row)
```
SBrandeis, line 157:
```suggestion
```

SBrandeis (commented):
Thanks @skyprince999 for the contribution.

I have a few questions and suggestions 👇

datasets/ohsumed/ohsumed.py
skyprince999: @SBrandeis I tried with an `int32` earlier, but some of the values were being read as a string. So changed it to `string`

datasets/ohsumed/ohsumed.py
skyprince999: the number of `mesh_terms` is very high, won't be practical to add it as a class label.

datasets/ohsumed/ohsumed.py
skyprince999: `publication_type` is one of the meta fields and not the target variable, so its not clear what is expected.

datasets/ohsumed/ohsumed.py
skyprince999: this is more of a `json-list` rather than a` json-lines` file! 
To give you some background, the original dataset is in some weird format. I had to donwload the dataset and then processed it to a `jsonl` format.  You can check the pre-processing code in my repo here - https://github.com/skyprince999/ohsumed-preprocessed

Also refer to my discussion with @yjernite  on the slack channel 
https://huggingface.slack.com/archives/C01FSCW6JLU/p1607713806319200

The original format is something similar to this- 
"""
Row 1  Col 1
Row 1
Col 2
Row 1
Col 3
.
.
Row 2 Col 1
Row 2
Col 2 
Row 2
Col 3

"""

datasets/ohsumed/ohsumed.py
SBrandeis: Do you mean that some values cannot be parsed to an integer ?
Otherwise you can cast the read values as integers at line 161 & 162, for example:
` "medline_ui": int(data["medline_ui"]),`

datasets/ohsumed/ohsumed.py
SBrandeis: Wow, thanks for taking the trouble to preprocess the data !
Would you mind refactoring to proper json-line files though ? I think it would be worth the hassle

skyprince999: I loved the part that this is a 20 year old dataset (from pre 2000's era ...that's like before Y2k ! ) 

Done, dusted with the refactoring!

datasets/ohsumed/ohsumed.py
skyprince999: done! this was giving me an error (yesterday)! 
both `seq_id `& `medline_ui `are now defined as `int64`.

datasets/ohsumed/ohsumed.py
skyprince999: currently the processed dataset is uploaded on my private gcp bucket. 
Is it possible to upload it to HF datahub ?

datasets/ohsumed/README.md
```diff
@@ -0,0 +1,205 @@
...
+
+### Languages
+
+The text is primarily in English. The BCP 47 code is `en`
+
+## Dataset Structure
+
+### Data Instances
+
+[More Information Needed]
```
SBrandeis, line 87: It would be nice to have an actual sample here, in JSON format 😄 
Something like: 
```json
{
    "seq_id": <seq_id>,
    "medline_ui": <medline_ui>,
    "mesh_terms": "terms",
    "title": "Title",
    "publication_type": "pub_type",
    "abstract": "Abstract of the document",
    "author": "John Doe",
    "source": "source of the document",
}
```

datasets/ohsumed/README.md
```diff
@@ -0,0 +1,205 @@
...
+multilinguality:
+- monolingual
+size_categories:
+- 100k< n<500K
+source_datasets:
+- original
+task_categories:
+- text-classification
+task_ids:
+- multi-label-classification
```
SBrandeis, line 19: What are the classification input/target?

datasets/ohsumed/ohsumed.py
SBrandeis: Don't forget to remove this print :)

datasets/ohsumed/ohsumed.py
```diff
@@ -0,0 +1,168 @@
...
+        )
+        return datasets.DatasetInfo(
+            # This is the description that will appear on the datasets page.
+            description=_DESCRIPTION,
+            # This defines the different columns of the dataset and their types
+            features=features,  # Here we define them above because they are different between the two configurations
+            # If there's a common (input, target) tuple from the features,
+            # specify them here. They'll be used if as_supervised=True in
+            # builder.as_dataset.
+            supervised_keys=None,
```
SBrandeis, line 115: WOuld be nice to set this since it's a classification task
```python
supervised_keys=(input_feature_name, label_feature_name)
```

SBrandeis (changes requested):
Thanks, @skyprince999, apart from a dangling print and some adjustment in the README we're almost there!

datasets/ohsumed/ohsumed.py
lhoestq: Since it's not something we can to predict with a model we can keep it as a string, it's fine

datasets/ohsumed/ohsumed.py
lhoestq: Hi @skyprince999 !

Thank you so much for processing all the data - the format is indeed kinda weird.
Do you think it could be possible to include the processing directly inside the dataset script ?
I checked your notebook and this is typically the kind of processing we use in dataset scripts when formats are conventional.
This way the data can directly be downloaded from the trec website.

What do you think ?

datasets/ohsumed/ohsumed.py
skyprince999: So I am not certain how to respond to this issue. 
To paraphrase information that's available in the readme files & a preview of the original paper. Link to the published paper is given below. Since its behind a paywall, I could just read the preview. 

https://link.springer.com/chapter/10.1007%2F978-1-4471-2099-5_20

The purpose of the exercise was creating a dataset which could be used for search. Physicians enter the searchterm and some relevant information regarding the patient and then rank the results from the MEDLINE system via their relevance scores (1- possibly relevant; 2- definitely relevant; not relevant topics were not included in the dataset)

The system needs to learn to differentiate between _possibly relevant_ & _definitely relevant_ topics. 

_A document is considered "relevant" to a MeSH "topic" if the MeSH concept name is listed in the MeSH term field (.M) of the document. Please note that the MeSH concepts form a hierarchy. It is common practice to index a document *only* by the most specific MeSH concept that is relevant._

There are 5000+ MeSH terms. Which are similar to something like this- 

"""
Adolescence; Artificial Limbs/EC/UT; Child; Child Health Services/EC/UT; Community Health Services/EC/*UT; Female; Financing, Government; Handicapped/*; Health Services Research; Human; Male; Orthotic Devices/EC/UT; Social Support; Support, Non-U.S. Gov't; Walkers/EC/UT.
"""

topic statement is a query defined by a physician - 

"""
<top>
<num> Number: OHSU1
<title> 60 year old menopausal woman without hormone replacement therapy
<desc> Description:
Are there adverse effects on lipids when progesterone is given with estrogen replacement therapy
</top>
"""


I had deliberately kept it open, by not defining any supervised tasks. But if you want to suggest a different approach, would love to hear about it.

lhoestq: If there's not an explicit and favored task that aims at predicting one field given other fields, then we should keep it to None. Here the logic looks a bit different so I don't mind keeping it as None

datasets/ohsumed/README.md
skyprince999: Pls check my note on the supervised task below.

datasets/ohsumed/ohsumed.py
skyprince999: I believe this will be inside the `generate_examples` function...? do you have a sample which I can use as a reference?

skyprince999: reviewing what will be required:
- integrate the data transformation logic 
- recreate the `dataset_info.json `file
- update the correct path to the extracted files
- update the dummy files similar to the extracted files

lhoestq: > I believe this will be inside the generate_examples function...? do you have a sample which I can use as a reference?

Yes exactly. You can iterate over the lines and yield one example once its parsing is completed.

For example that's what the [conll2003](https://github.com/huggingface/datasets/blob/master/datasets/conll2003/conll2003.py) script does but for NER

skyprince999: updated the new commit with the above steps. waiting for the tests to complete. 
 
It's quite late over here. so need to shut down for the night! :) 
will check in the morning.

SBrandeis (approved):
Looks good to me! Thanks for including the preprocessing in the script.
I'll let @lhoestq double check on this before we merge. 🙂 

Here is some feedback to make future reviews smoother:
- Try to avoid resolving conversations when you're done fixing them. It saves us an extra click when reviewing the requested changes 😉 
- Try to avoid altering the git history when it's already pushed on your remote, as it breaks GitHub's "changes since the last review" feature.  This means: prefer `merge` over `rebase` when syncing your branch with master, and avoid the `git push --force` command.

Thanks again for your work!

datasets/ohsumed/README.md
```diff
@@ -0,0 +1,216 @@
...
+
```
lhoestq, line 20:
```suggestion
```

datasets/ohsumed/README.md
```diff
@@ -0,0 +1,216 @@
...
+{'seq_id': 7770,
+  'medline_ui': 87120420,
+  'mesh_terms': 'Adult; Aged; Aneurysm/CO; Arteriovenous Fistula/*TH; Carotid Arteries; Case Report; Female; Human; Jugular Veins; Male; Methods; Middle Age; Neck/*BS; Vertebral Artery.',
+  'title': 'Arteriovenous fistulas of the large vessels of the neck: nonsurgical percutaneous occlusion.',
+  'publication_type': 'JOURNAL ARTICLE.',
+  'abstract': 'We describe the nonsurgical treatment of arteriovenous fistulas of the large vessels in the neck using three different means of endovascular occlusion of these large lesions, which are surgically difficult to approach and treat.',
+  'author': 'Vitek JJ; Keller FS.',
+  'source': 'South Med J 8705; 80(2):196-200'}
+
+"""
```
lhoestq, line 97:
````suggestion
```
{'seq_id': 7770,
  'medline_ui': 87120420,
  'mesh_terms': 'Adult; Aged; Aneurysm/CO; Arteriovenous Fistula/*TH; Carotid Arteries; Case Report; Female; Human; Jugular Veins; Male; Methods; Middle Age; Neck/*BS; Vertebral Artery.',
  'title': 'Arteriovenous fistulas of the large vessels of the neck: nonsurgical percutaneous occlusion.',
  'publication_type': 'JOURNAL ARTICLE.',
  'abstract': 'We describe the nonsurgical treatment of arteriovenous fistulas of the large vessels in the neck using three different means of endovascular occlusion of these large lesions, which are surgically difficult to approach and treat.',
  'author': 'Vitek JJ; Keller FS.',
  'source': 'South Med J 8705; 80(2):196-200'}

```
````

File: datasets/ohsumed/README.md [hunk 14]
```diff
@@ -0,0 +1,216 @@
...
+ .I      sequential identifier 
+   (important note: documents should be processed in this order)
+ .U      MEDLINE identifier (UI) 
+   (<DOCNO> used for relevance judgements)
+ .M      Human-assigned MeSH terms (MH)
+ .T      Title (TI)
+ .P      Publication type (PT)
+ .W      Abstract (AB)
+ .A      Author (AU)
+ .S      Source (SO)
```
lhoestq, line 113:
```suggestion
- seg_id: sequential identifier 
   (important note: documents should be processed in this order)
- medline_ui: MEDLINE identifier (UI) 
   (<DOCNO> used for relevance judgements)
 - mesh_terms: Human-assigned MeSH terms (MH)
- title: Title (TI)
- publication_type : Publication type (PT)
- abstract: Abstract (AB)
- author: Author (AU)
- source: Source (SO)
```

lhoestq (approved):
Awesome thank you !
Looks all good :) 

Good job with the parsing and with the dataset card

PR #1539, opened 2020-12-13
Added Wiki Asp dataset

katnoria:
Hello,

I have added Wiki Asp dataset. Please review the PR.


datasets/wiki_asp/wiki_asp.py
```diff
@@ -0,0 +1,212 @@
...
+        datasets.BuilderConfig(name="town", version=VERSION, description="A subset of dataset from the town domain"),
+        datasets.BuilderConfig(
+            name="written_work", version=VERSION, description="A subset of dataset from the written work domain"
+        ),
+    ]
+
+    # Using Plant as default because of its small size (286MB)
+    # The full dataset is 28GB
+    DEFAULT_CONFIG_NAME = "plant"
+
```
lhoestq, line 150:
```suggestion
```
not sure we should have a default config here

lhoestq (commented):
Awesome thank you !

I just left one comment.

Also it looks like the dummy_data.zip files are quite big (around 500KB each)
Can you try to reduce their sizes please ? Ideally they should be <20KB each

To do so feel free to take a look inside them and in the jsonl files only keep 1 or 2 samples instead of 5 and also remove big chunks of text to only keep a few passages.

katnoria:
> Awesome thank you !
> 
> I just left one comment.
> 
> Also it looks like the dummy_data.zip files are quite big (around 500KB each)
> Can you try to reduce their sizes please ? Ideally they should be <20KB each
> 
> To do so feel free to take a look inside them and in the jsonl files only keep 1 or 2 samples instead of 5 and also remove big chunks of text to only keep a few passages.

Thanks, I have updated the dummy data to keep each domain <20/30KB.

datasets/wiki_asp/wiki_asp.py
katnoria: So we do not want to have a default for this dataset? This is multi-domain dataset where I chose the smallest domain as default. Happy to remove if you think we can do without default here.

katnoria:
> > Awesome thank you !
> > I just left one comment.
> > Also it looks like the dummy_data.zip files are quite big (around 500KB each)
> > Can you try to reduce their sizes please ? Ideally they should be <20KB each
> > To do so feel free to take a look inside them and in the jsonl files only keep 1 or 2 samples instead of 5 and also remove big chunks of text to only keep a few passages.
> 
> Thanks, I have updated the dummy data to keep each domain <20/30KB.

Looks like this branch has other commits. I will open a new PR with suggested changes.

katnoria:
opened a new PR #1612

PR #1330, opened 2020-12-08
added un_ga dataset

param087:
Hi :hugs:, This is a PR for [United nations general assembly resolutions: A six-language parallel corpus](http://opus.nlpl.eu/UN.php) dataset

PR #1569, opened 2020-12-14
added un_ga dataset

param087:
Hi :hugs:, This is a PR for [United nations general assembly resolutions: A six-language parallel corpus](http://opus.nlpl.eu/UN.php) dataset.
With suggested changes in #1330


lhoestq (approved):
Awesome thanks !

PR #1577, opened 2020-12-15
Add comet metric

ricardorei:
Hey! I decided to add our new Crosslingual Optimized Metric for Evaluation of Translation (COMET) to the list of the available metrics.

COMET was [presented at EMNLP20](https://www.aclweb.org/anthology/2020.emnlp-main.213/) and it is the highest performing metric, so far, on the WMT19 benchmark.

We also participated in the [WMT20 Metrics shared task ](http://www.statmt.org/wmt20/pdf/2020.wmt-1.101.pdf) where once again COMET was validated as a top-performing metric. 

I hope that this metric will help researcher's and industry workers to better validate their MT systems in the future 🤗 !

Cheers,
Ricardo

### DIFF
--- a/metrics/comet/comet.py
+++ b/metrics/comet/comet.py
@@ -1,0 +1,133 @@
+# coding=utf-8
+# Copyright 2020 The HuggingFace Datasets Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+""" COMET metric. 
+
+Usage:
+
+from datasets import load_metric
+comet_metric = load_metric('metrics/comet/comet.py')
+#comet_metric = load_metric('comet')
+#comet_metric = load_metric('comet', 'wmt-large-hter-estimator')
+
+
+source = ["Dem Feuer konnte Einhalt geboten werden", "Schulen und Kindergärten wurden eröffnet."]
+hypothesis = ["The fire could be stopped", "Schools and kindergartens were open"]
+reference = ["They were able to control the fire.", "Schools and kindergartens opened"]
+
+predictions = comet_metric.compute(predictions=hypothesis, references=reference, sources=source)
+predictions['scores']
+"""
+
+import os
+from logging import getLogger
+
+from comet.models import download_model, load_checkpoint
+
+import datasets
+
+
+logger = getLogger(__name__)
+
+_CITATION1 = """\
+@inproceedings{rei-etal-2020-comet,
+   title = "{COMET}: A Neural Framework for {MT} Evaluation",
+   author = "Rei, Ricardo  and
+      Stewart, Craig  and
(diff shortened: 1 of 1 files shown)


metrics/comet/comet.py
```diff
@@ -0,0 +1,138 @@
...
+`references` (list of str): reference translations
+`cuda` (bool): If set to True, runs COMET using GPU
+`show_progress` (bool): Shows progress
+
+
+`model`: COMET model to be used. Will default to `wmt-large-da-estimator-1719` if None.
+
+Returns:
+    `predictions`: List of dictionaries with `src`, `mt`, `ref` and `score`.
+    'scores': List of scores.
```
SBrandeis, line 102:
```suggestion
    `scores`: List of scores.
```

metrics/comet/comet.py
```diff
@@ -0,0 +1,138 @@
...
+`predictions` (list of str): candidate translations
+`references` (list of str): reference translations
+`cuda` (bool): If set to True, runs COMET using GPU
+`show_progress` (bool): Shows progress
+
+
+`model`: COMET model to be used. Will default to `wmt-large-da-estimator-1719` if None.
+
+Returns:
+    `predictions`: List of dictionaries with `src`, `mt`, `ref` and `score`.
```
SBrandeis, line 101: Following the output of the `_compute` method I believe `predictions` should be renamed to `samples`:

```suggestion
    `samples`: List of dictionaries with `src`, `mt`, `ref` and `score`.
```

SBrandeis (commented):
Super cool, thansk a lot !
I have spotted a few typos in the docstrings.

lhoestq (commented):
Really cool :) 
Thank you for adding it and sorry for the delay.

I just left two comments.

I like the general API of this metric when we cant to compute the score using
```
comet_metric.compute(predictions=hypothesis, references=reference, sources=source)
```

But I would also like to highlight the fact that because `sources` is not part of the features (either in `references` or in `predictions`), then `sources` will always be required when calling `compute` when adding `references` and `predictions` **by batch**.

Indeed if `references` and `predictions` are passed by batch with `comet_metric.add_batch(references=references, predictions=predictions)` then we need to call `comet_metric.compute(sources=sources)`.

IMO we can keep things as they are right now.

An alternative would be to make the `references` feature a dictionary of both MT references and sources but this would change the API this way
```
comet_metric.compute(predictions=hypothesis, references={
    "mt_references": mt_references,
    "sources": sources}
})
```
but I'm not sure this is better.

Let me know what you think.
If you think it's fine this way we can keep it as it is

metrics/comet/comet.py
```diff
@@ -0,0 +1,138 @@
...
+#comet_metric = load_metric('comet', 'wmt-large-hter-estimator')
+
+
+source = ["Dem Feuer konnte Einhalt geboten werden", "Schulen und Kindergärten wurden eröffnet."]
+hypothesis = ["The fire could be stopped", "Schools and kindergartens were open"]
+reference = ["They were able to control the fire.", "Schools and kindergartens opened"]
+
+predictions = comet_metric.compute(predictions=hypothesis, references=reference, sources=source)
+predictions['scores']
+```
```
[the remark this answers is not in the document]
ricardorei: Yes, no problem. I just didn't do it before to be consistent with other metrics. Can you check it now?

ricardorei:
I also thought a bit about the fact that "sources" can't be added to the batch.. but changing that would require a lot more changes. And I agree that the idea of adding them as part of the references is not ideal. Conceptually they are not references.

I would keep it like this for now.. And in the future, work on a more consistent batch interface.

lhoestq (approved):
Thank you. Looks all good now :)

PR #1589, opened 2020-12-16
Update doc2dial.py

songfeng:
Added data loader for machine reading comprehension tasks proposed in the Doc2Dial EMNLP 2020 paper.


lhoestq:
Thanks for adding the `doc2dial_rc` config :) 

It looks like you're missing the dummy data for this config though. Could you add them please ?
Also to fix the CI you'll need to format the code with `make style`

PR #1599, opened 2020-12-17
add Korean Sarcasm Dataset

stevhliu:


datasets/kor_sarcasm/kor_sarcasm.py
```diff
@@ -0,0 +1,73 @@
...
+    """Korean Sarcasm Detection Dataset"""
+
+    VERSION = datasets.Version("1.1.0")
+
+    def _info(self):
+
+        return datasets.DatasetInfo(
+            description=_DESCRIPTION,
+            features=datasets.Features(
+                {"tokens": datasets.Value("string"), "label": datasets.features.ClassLabel(names=["0", "1"])}
```
lhoestq, line 47: Could you use more explicit label class names please ?
For example
```python
names=["no-sarcasm", "sarcasm"]
```

you'll have to regenerate the dataset_infos.json after this change

lhoestq (commented):
Nice ! thanks for adding this one

I only have one comment, all the rest is all good to me :)

datasets/kor_sarcasm/kor_sarcasm.py
stevhliu: i get a `KeyError: '1'` when I try to rename the class names more explicitly. is there a way I can change this without modifying the original class labels in the dataset?

datasets/kor_sarcasm/kor_sarcasm.py
```diff
@@ -0,0 +1,73 @@
...
+    def _generate_examples(self, filepath):
+        """Generate Korean sarcasm examples"""
+
+        with open(filepath, encoding="utf-8") as csv_file:
+            data = csv.reader(csv_file, delimiter=",")
+            next(data, None)
+            for id_, row in enumerate(data):
+                row = row[1:3]
+                tokens, label = row
+                yield id_, {"tokens": tokens, "label": label}
```
lhoestq, line 73:
```suggestion
                yield id_, {"tokens": tokens, "label": int(label)}
```
Yes you can fix the KeyError by converting to integer here to get 0 or 1.

lhoestq (approved):
Thanks !

Looks all good now :)

PR #1601, opened 2020-12-18
second update of the id_newspapers_2018

cahya-wirawan:
The feature "url" is currently set wrongly to data["date"], this PR fix it to data["url"].
I added also an additional POC.


cahya-wirawan:
I close this PR, since it based on 1 week old repo. And I will create a new one

PR #1602, opened 2020-12-18
second update of id_newspapers_2018

cahya-wirawan:
The feature "url" is currently set wrongly to data["date"], this PR fix it to data["url"].
I added also an additional POC.


lhoestq (approved):
good catch !

Thanks

PR #1603, opened 2020-12-18
Add retries to HTTP requests

### LINKED ISSUE #1102: Add retries to download manager

SBrandeis:
## What does this PR do ?

Adding retries to HTTP GET & HEAD requests, when they fail with a `ConnectTimeout` exception.

The "canonical" way to do this is to use [urllib's Retry class](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.Retry) and wrap it in a [HttpAdapter](https://requests.readthedocs.io/en/master/api/#requests.adapters.HTTPAdapter). Seems a bit overkill to me, plus it forces us to use the `requests.Session` object. I prefer this simpler implementation. I'm open to remarks and suggestions @lhoestq @yjernite 

Fixes #1102

### DIFF
--- a/src/datasets/utils/file_utils.py
+++ b/src/datasets/utils/file_utils.py
@@ -393,9 +393,19 @@ def get_datasets_user_agent(user_agent: Optional[Union[str, dict]] = None) -> st
 
 
 def _request_with_retry(
-    verb: str, url: str, max_retries: int = 0, time_between_tries: int = 1, **params
+    verb: str, url: str, max_retries: int = 0, base_wait_time: float = 0.5, max_wait_time: float = 2, **params
 ) -> requests.Response:
-    """Wrapper around requests to retry in case it fails with a ConnectTimeout"""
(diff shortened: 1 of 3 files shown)


src/datasets/utils/file_utils.py
```diff
@@ -386,11 +392,32 @@ def get_datasets_user_agent(user_agent: Optional[Union[str, dict]] = None) -> st
...
+        tries += 1
+        try:
+            response = requests.request(verb.upper(), url, **params)
+            success = True
+        except requests.exceptions.ConnectTimeout as err:
+            if tries > max_retries:
+                raise err
+            else:
+                logger.info(f"{verb} request to {url} timed out, retrying... [{tries/max_retries}]")
+                time.sleep(time_between_tries)
```
julien-c, line 410: should it be an exponential backoff?

lhoestq (approved):
Nice thanks !

I think the default max_retries to 1 is fine

lhoestq:
merging this one then :)

PR #1608, opened 2020-12-19
adding  ted_talks_iwslt

skyprince999:
UPDATE2: (2nd Jan) Wrote a long writeup on the slack channel. I don't think this approach is correct. Basically this created language pairs (109*108) 
Running the `pytest `went for more than 40+ hours and it was still running!  
So working on a different approach, such that the number of configs = number of languages. Will make a new pull request with that. 

UPDATE: This requires manual download dataset

This is a draft version


skyprince999:
Closing this with reference to the new approach #1676

PR #1612, opened 2020-12-20
Adding wiki asp dataset as new PR

katnoria:
Hi @lhoestq, Adding wiki asp as new branch because #1539 has other commits. This version has dummy data for each domain <20/30KB.


datasets/wiki_asp/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Languages
+
+[More Information Needed]
+
+## Dataset Structure
+
+### Data Instances
+
+```
```
lhoestq, line 69:
````suggestion
An example from the "plant" configuration:
```
````

**Review by lhoestq — APPROVED** [review]
Thank you !
LGTM

File: datasets/wiki_asp/wiki_asp.py [hunk 2]
```diff
@@ -0,0 +1,212 @@
...
+        ),
+        datasets.BuilderConfig(name="town", version=VERSION, description="A subset of dataset from the town domain"),
+        datasets.BuilderConfig(
+            name="written_work", version=VERSION, description="A subset of dataset from the written work domain"
+        ),
+    ]
+
+    # Using Plant as default because of its small size (286MB)
+    # The full dataset is 28GB
+    DEFAULT_CONFIG_NAME = "plant"
```
lhoestq, line 149:
```suggestion

    # Using Plant as default because of its small size (286MB)
    # The full dataset is 28GB
    DEFAULT_CONFIG_NAME = "plant"
```
As mentioned earlier we want the users to specify which configuration they want to load for this dataset.
It can be confusing to get only the plant config when nothing is specified

PR #1619, opened 2020-12-21
data loader for reading comprehension task

songfeng:
added doc2dial data loader and dummy data for reading comprehension task.

### DIFF
--- a/datasets/doc2dial/doc2dial.py
+++ b/datasets/doc2dial/doc2dial.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 # Lint as: python3
-"""Doc2dial: A Goal-Oriented Document-Grounded Dialogue Dataset v0.9"""
+"""Doc2dial: A Goal-Oriented Document-Grounded Dialogue Dataset v0.9.0"""
 
 from __future__ import absolute_import, division, print_function
 
@@ -93,7 +93,12 @@ class Doc2dial(datasets.GeneratorBasedBuilder):
                             "turn_id": datasets.Value("int32"),
                             "role": datasets.Value("string"),
                             "da": datasets.Value("string"),
-                            "reference": [{"keys": datasets.Value("string"), "values": datasets.Value("string"),}],
+                            "reference": [
+                                {
+                                    "keys": datasets.Value("string"),
+                                    "values": datasets.Value("string"),
+                                }
+                            ],
                             "utterance": datasets.Value("string"),
                         }
                     ],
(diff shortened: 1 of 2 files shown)


lhoestq (commented):
Nice ! 

It looks like the dummy data zip file is quite big though (almost 8MB).
Could you try to reduce its size please ?
Ideally it should be under 20KB

songfeng:
Thank you for all the feedback! I have updated the dummy data with a zip under 30KB, which needs to include at least one data instance from both document domain and dialog domain. Please let me know if it is still too big. Thanks!

datasets/doc2dial/doc2dial.py
```diff
@@ -163,6 +212,133 @@ def _split_generators(self, dl_manager):
...
+                                turn_to_predict["reference"], doc["spans"], doc["doc_text"]
+                            )
+                        if not turn_to_predict["references"]:
+                            qa["is_impossible"] = True
+                            qa["answers"] = []
+                        else:
+                            qa["is_impossible"] = False
+                            qa["answers"] = turn_to_predict["references"]
+                            if len(qa["answers"]) < 1:
+                                assert len((qa["answers"])) >= 1
```
lhoestq, line 331: What is this used for ?

datasets/doc2dial/doc2dial.py
```diff
@@ -163,6 +212,133 @@ def _split_generators(self, dl_manager):
...
+
+    def _load_doc_data_rc(self, filepath):
+        doc_filepath = os.path.join(os.path.dirname(filepath), "..", "doc2dial_doc.json")
+        with open(doc_filepath, encoding="utf-8") as f:
+            data = json.load(f)["doc_data"]
+        return data
+
+    def _get_start_end_candidates_rc(self, tss):
+        start_candidates, end_candidates = [], []
+        for ts_id, ts in tss.items():
```
lhoestq, line 248: What do `ts` and `tss` mean here ?

lhoestq (commented):
Thanks for reducing the size of the dummy data :) 

Can you add some comments and docstrings to explain your processing please ?

Also can you add an example from the dataset in the `Data Instances` section of the readme of this dataset ?
You can get one by using `load_dataset` and taking the first sample `dataset["train"][0]` for example.
And can you briefly describe the data fields in the `Data Fields` section ?

songfeng:
Thank you again for the feedback! I am not too sure what the preferable style for data instance in readme, but still added my edits. Thanks!

datasets/doc2dial/doc2dial.py
```diff
@@ -411,9 +341,62 @@ def _generate_examples(self, filepath):
...
+                                        turn_to_predict["reference"], doc["spans"], doc["doc_text"]
+                                    )
+                                if not turn_to_predict["references"]:
+                                    qa["is_impossible"] = True
+                                else:
+                                    qa["is_impossible"] = False
+                                    qa["answers"] = turn_to_predict["references"]
+                                    """ Ensure the answers are not empty if the question is answerable."""
+                                    if len(qa["answers"]) < 1:
+                                        assert len((qa["answers"])) >= 1
```
lhoestq, line 401:
```suggestion
                                    assert (
                                        len((qa["answers"])) >= 1
                                    ), "Ensure the answers are not empty if the question is answerable"
```

lhoestq (approved):
Thank you !

PR #1321, opened 2020-12-08
added dutch_social

skyprince999:
The Dutch social media tweets dataset. Which has a total of more than 210k tweets in dutch language. These tweets have been machine annotated with sentiment scores (`label` feature) and `industry` and `hisco_codes`

It can be used for sentiment analysis, multi-label classification and entity tagging

PR #1621, opened 2020-12-22
updated dutch_social.py for loading jsonl (lines instead of list) files

skyprince999:
the data_loader  is modified to load files on the fly. Earlier it was reading the entire file and then processing the records

Pls refer to previous PR #1321


lhoestq (approved):
Nice thanks !

PR #1629, opened 2020-12-24
add wongnai_reviews test set labels

cstorm125:
- add test set labels provided by @ekapolc
- refactor `star_rating` to a `datasets.features.ClassLabel` field


lhoestq (approved):
Thank you for adding the test set @cstorm125 @ekapolc :)

PR #1640, opened 2020-12-26
Fix "'BertTokenizerFast' object has no attribute 'max_len'"

mflis:
Tensorflow 2.3.0 gives:
 FutureWarning: The `max_len` attribute has been deprecated and will be removed in a future version, use `model_max_length` instead.

Tensorflow 2.4.0 gives:
AttributeError 'BertTokenizerFast' object has no attribute 'max_len'


lhoestq (approved):
Indeed ! thank you

PR #1663, opened 2020-12-29
update saving and loading methods for faiss index so to accept path l…

tslott:
- Update saving and loading methods for faiss index so to accept path like objects from pathlib

The current code only supports using a string type to save and load a faiss index. This change makes it possible to use a string type OR a Path from [pathlib](https://docs.python.org/3/library/pathlib.html). The codes becomes a more intuitive this way I think.

### DIFF
--- a/src/datasets/search.py
+++ b/src/datasets/search.py
@@ -1,5 +1,6 @@
 import os
 import tempfile
+from pathlib import PurePath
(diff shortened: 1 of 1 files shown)


thomwolf:
Seems ok for me, what do you think @lhoestq ?

lhoestq (approved):
Indeed thanks :)

PR #1676, opened 2021-01-02
new version of Ted Talks IWSLT (WIT3)

skyprince999:
In the previous iteration #1608 I had used language pairs. Which created 21,582 configs (109*108) !!! 

Now, TED talks in _each language_ is a separate config. So it's more cleaner with _just 109 configs_  (one for each language). Dummy files were created manually. 

Locally  I was able to clear the `python datasets-cli test datasets/......` . Which created the `dataset_info.json`  file . The test for the dummy files was also cleared. However couldn't figure out how to specify the local data folder for the real dataset

**Note: that this requires manual download of the dataset.** 
**Note2: The high number of _Files changed (112)_ is because of the large number of dummy files/configs!**


datasets/ted_talks_iwslt_2/ted_talks_iwslt_2.py
```diff
@@ -0,0 +1,417 @@
...
+
+        if os.path.exists(source):
+            with zipfile.ZipFile(source) as zf_source:
+                try:
+                    source_talks = get_talks(zf_source)
+                    # source_ids = [talk.get("head")[0].get("talkid")[0] for talk in source_talks]
+
+                except Exception as pe:
+                    print(f"ERROR: {pe}")
+                    print(f"Which likely means that you have a malformed XML file:: {source}\n")
```
lhoestq, line 384: You can use `logging.warning` instead of `print`

lhoestq (commented):
Nice thank you ! Actually as it is a translation dataset we should probably have one configuration = one language pair no ?

For this kind of datasets where there are many configurations possible (here for any pair of languages), we don't construct all the configuration but only a subset of them (in general <10 configurations).
Then all the configurations that are not loaded can be constructed on the fly. For example it is the case for the `opus_gnome` dataset in which you can load any language pair using for example
```python
dataset = load_dataset("opus_gnome", lang1="it", lang2="pl")
```
When this code is run, a configuration is created on the fly for this language pair even if it's not already in the `BUILDER_CONFIGS` list. This is possible since the dataset configuration for this dataset has the parameters `lang1` and `lang2`. You can check it out [here](https://github.com/huggingface/datasets/blob/master/datasets/opus_gnome/opus_gnome.py#L65-L73). 

This way we don't have to define hundreds of configs (and therefore hundreds of dummy data files), but only a few of them only for the configurations that exist in `BUILDER_CONFIGS`. For example for `opus_gnome` only 10 configurations (=10 language pairs) are defined.

Could you use the same trick for this dataset ?

datasets/ted_talks_iwslt_2/README.md
```diff
@@ -0,0 +1,321 @@
...
+| `<date>` | the issue date of the talk |
+| `<content>` |  talk subtitles |
+
+
+### Data Splits
+
+Dat has been split based on the year
+2014 -> Train
+2015 -> Dev
+2016 -> Test
```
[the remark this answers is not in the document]
skyprince999: the paper doesn't suggest any splits. So did my own _arbit_ splits. Any suggestions?

datasets/ted_talks_iwslt_2/ted_talks_iwslt_2.py
```diff
@@ -0,0 +1,417 @@
...
+# TODO: Name of the dataset usually match the script name with CamelCase instead of snake_case
+class TedTalksIWSLT2(datasets.GeneratorBasedBuilder):
+    """TODO: Short description of my dataset."""
+
+    VERSION = datasets.Version("1.1.0")
+
+    @property
+    def manual_download_instructions(self):
+        return """\
+        \nThe file can be downloaded from: https://drive.google.com/file/d/1Cz1Un9p8Xn9IpEMMrg2kXSDt0dnjxc4z/view?usp=sharing
```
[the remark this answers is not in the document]
skyprince999: Oh... I tried that. 
It doesn't work, since its a large file ( ~1.6GB) Google prompts you for download

datasets/ted_talks_iwslt_2/ted_talks_iwslt_2.py
skyprince999: ha ha ...okay!

datasets/ted_talks_iwslt_2/ted_talks_iwslt_2.py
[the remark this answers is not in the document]
lhoestq: Yep but the dl_manager handles the prompt ;)

datasets/ted_talks_iwslt_2/README.md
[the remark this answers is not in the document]
lhoestq: Maybe the year can be a parameter as well ?
```
load_dataset("ted_talks_iwslt_2", lang1="fr", lang2="en", year="2015")
```
If `year` is not specified it can download all the years

skyprince999:
> Nice thank you ! Actually as it is a translation dataset we should probably have one configuration = one language pair no ?
> 
> Could you use the same trick for this dataset ?

I was looking for this input, infact I had written a long post on the Slack channel,...(_but unfortunately due to the holidays prob didn;t get a respones_). Initially I had tried with language pairs and then with specific language configs.

datasets/ted_talks_iwslt_2/README.md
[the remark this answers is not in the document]
skyprince999: okay... I guess then there are no train/test/dev splits ?

lhoestq:
Oh sorry I must have missed your message then :/
I was off a few days during the holidays

Hopefully this trick can enable the use of any language pair (+ year ?) combination and also simplify a lot the dummy data creation since it will only require a few configs.

datasets/ted_talks_iwslt_2/README.md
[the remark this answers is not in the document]
lhoestq: Yes in general we let the users define their splits if there's no split explicitly given by the authors.

For example users can use the `.train_test_split()` method

skyprince999:
Updated it as per the comments. But couldn't figure out why the dummy tests are failing >> 
```

tests/test_dataset_common.py:198: in check_load_dataset
E   AssertionError: False is not true
```

datasets/ted_talks_iwslt/README.md
```diff
@@ -0,0 +1,328 @@
...
+size_categories:
+- 100K< n<500k
+source_datasets:
+- original
+task_categories:
+- conditional-text-generation
+- other
+task_ids:
+- machine-translation
+- other-other-machine-translation
```
lhoestq, line 130: Don't need the `other` tags here

```suggestion
task_categories:
- conditional-text-generation
task_ids:
- machine-translation
```

datasets/ted_talks_iwslt/README.md
```diff
@@ -0,0 +1,328 @@
...
+  'nl': 'Toen Ebola in maart 2014 uitbrak, zijn Pardis Sabeti en haar team aan het werk gegaan om het genoom in kaart te brengen. Zo ontdekten ze hoe het virus zich verspreidde en muteerde. Sabeti zette direct haar onderzoek op het internet, zodat wereldwijd virus-jagers en wetenschappers mee konden werken aan de strijd. In deze talk laat ze zien hoe die openheid geholpen heeft bij het stoppen van het virus en hoe het kan helpen bij de strijd tegen het volgende virus. "We moesten transparant werken, delen en samenwerken". Sabeti zegt:"Laat de wereld niet ten onder gaan aan een virus, maar verlicht worden door miljoenen harten en geesten die samenwerken."'}}
+```
+
+The original XML file has the following data instances:
```
lhoestq, line 189:
```suggestion
The original XML files are formatted like this example:
```

datasets/ted_talks_iwslt/README.md
```diff
@@ -0,0 +1,328 @@
...
+machine learning task, language modeling and generation
+
+### Languages
+
+Ted talks are mostly held in English (`en`). Almost all of the talks have been translated, by volunteers, into Arabic, Bulgarian, Chinese (simplified), French, Italian, Korean, Portuguese (Brazil) and Spanish. For about 70 other languages, the number of translated talks ranges from several hundreds (e.g. such as other Dutch, German, Hebrew, Romanian) to one (e.g. Hausa, Hupa, Bislama, Ingush, Maltese)
+
+## Dataset Structure
+
+### Data Instances
+
```
lhoestq, line 183:
```suggestion

One example from the dataset is:

```

datasets/ted_talks_iwslt/README.md
```diff
@@ -0,0 +1,328 @@
...
+    <date>2014/01/12</date>
+    <wordnum>885</wordnum>
+    <charnum>5051</charnum>
+  </head>
+  <content>(Music) For any of you who have visited or lived in New York City, these shots might start to look familiar. This is Central Park, ............new ways for people to interact with and experience music. Thank you. (Applause)</content>
+</file>
+```
+
+### Data Fields
+
```
lhoestq, line 229:
```suggestion
The fields of the dataset are:
- translation:
  - <lang1>: text in <lang1>
  - <lang2>L translated text in <lang2> 
```

datasets/ted_talks_iwslt/README.md
```diff
@@ -0,0 +1,328 @@
...
+
```
lhoestq, line 230:
```suggestion

Information about the original data files:

```

lhoestq (approved):
Looks all good to me know !
Thank you :)

datasets/ted_talks_iwslt/README.md
```diff
@@ -0,0 +1,328 @@
...
+- **Paper:** https://www.aclweb.org/anthology/2012.eamt-1.60.pdf
+- **Leaderboard:** [Needs More Information]
+- **Point of Contact:** [Mauro Cettolo](mailto:cettolo@fbk.eu)
+[Roldano Cattoni](mailto:cattoni@fbk.eu)
+
+
+### Dataset Summary
+
+The Web Inventory Talk is a collection of the original Ted talks and their translated version. The translations are available in more than 109+ languages, though the distribution is not uniform. 
+
```
lhoestq, line 171:
```suggestion

To load a language pair which isn't part of the config, all you need to do is specify the language code as pairs.
You can find the valid pairs in Homepage section of Dataset Description: http://opus.nlpl.eu/GNOME.php
E.g.

`dataset = load_dataset("ted_talks_iwslt", language_pair=("it", "pl"), year="2014")`

The full list of languages is: 'af', 'am', 'ar', 'arq', 'art-x-bork', 'as', 'ast', 'az', 'be', 'bg', 'bi', 'bn', 'bo', 'bs', 'ca', 'ceb', 'cnh', 'cs', 'da', 'de', 'el', 'en', 'eo', 'es', 'et', 'eu', 'fa', 'fi', 'fil', 'fr', 'fr-ca', 'ga', 'gl', 'gu', 'ha', 'he', 'hi', 'hr', 'ht', 'hu', 'hup', 'hy', 'id', 'ig', 'inh', 'is', 'it', 'ja', 'ka', 'kk', 'km', 'kn', 'ko', 'ku', 'ky', 'la', 'lb', 'lo', 'lt', 'ltg', 'lv', 'mg', 'mk', 'ml', 'mn', 'mr', 'ms', 'mt', 'my', 'nb', 'ne', 'nl', 'nn', 'oc', 'pa', 'pl', 'ps', 'pt', 'pt-br', 'ro', 'ru', 'rup', 'sh', 'si', 'sk', 'sl', 'so', 'sq', 'sr', 'srp', 'sv', 'sw', 'szl', 'ta', 'te', 'tg', 'th', 'tl', 'tlh', 'tr', 'tt', 'ug', 'uk', 'ur', 'uz', 'vi', 'zh', 'zh-cn', 'zh-tw'.

The full list of years is: '2014', '2015', '2016'.

```
we can also add one usage example here showing that we can load any pair

datasets/ted_talks_iwslt/README.md
```diff
@@ -0,0 +1,343 @@
...
+To load a language pair which isn't part of the config, all you need to do is specify the language code as pairs.
+You can find the valid pairs in Homepage section of Dataset Description: http://opus.nlpl.eu/GNOME.php
```
lhoestq, line 171:
```suggestion
```
wrong sentence

PR #1684, opened 2021-01-04
Add CANER Corpus

KMFODA:
What does this PR do?

Adds the following dataset:

https://github.com/RamziSalah/Classical-Arabic-Named-Entity-Recognition-Corpus

Who can review?

@lhoestq


datasets/caner/README.md
```diff
@@ -0,0 +1,172 @@
...
+- 100K<n<1M
+source_datasets:
+- original
+task_categories:
+- structure-prediction
+task_ids:
+- named-entity-recognition
+---
+
+# Dataset Card for [Persian NER]
```
lhoestq, line 21:
```suggestion
# Dataset Card for CANER
```

datasets/caner/README.md
```diff
@@ -0,0 +1,172 @@
...
+- [Additional Information](#additional-information)
+  - [Dataset Curators](#dataset-curators)
+  - [Licensing Information](#licensing-information)
+  - [Citation Information](#citation-information)
+
+## Dataset Description
+
+- **Homepage:** 
+- **Repository:** https://github.com/RamziSalah Classical-Arabic-Named-Entity-Recognition-Corpus
+- **Paper:** https://www.researchgate.net/publication/330075080_BUILDING_THE_CLASSICAL_ARABIC_NAMED_ENTITY_RECOGNITION_CORPUS_CANERCORPUS
```
lhoestq, line 50:
```suggestion
- **Repository:** [Classical-Arabic-Named-Entity-Recognition-Corpus](https://github.com/RamziSalah)
- **Paper:** [Researchgate](https://www.researchgate.net/publication/330075080_BUILDING_THE_CLASSICAL_ARABIC_NAMED_ENTITY_RECOGNITION_CORPUS_CANERCORPUS)
```

datasets/caner/README.md
```diff
@@ -0,0 +1,172 @@
...
+- **Leaderboard:**
+- **Point of Contact:** 
+
+### Dataset Summary
+
+The dataset contrains a cclassical Arabic Named Entity Recognition corpus as a new corpus of tagged data that can be useful for handling the issues in recognition of Arabic named entities.
```
lhoestq, line 56:
```suggestion
The Classical Arabic Named Entity Recognition corpus is a new corpus of tagged data that can be useful for handling the issues in recognition of Arabic named entities.
```

datasets/caner/README.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Supported Tasks and Leaderboards
+
+[More Information Needed]
```
lhoestq, line 60:
```suggestion
- Named Entity Recognition
```

datasets/caner/README.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Languages
+
+[More Information Needed]
```
lhoestq, line 64:
```suggestion
Classical Arabic
```

datasets/caner/caner.py
```diff
@@ -0,0 +1,133 @@
...
+class Caner(datasets.GeneratorBasedBuilder):
+    """Classical Arabic Named Entity Recognition corpus as a new corpus of tagged data that can be useful for handling the issues in recognition of Arabic named entities"""
+
+    VERSION = datasets.Version("1.1.0")
+
+    BUILDER_CONFIGS = [
+        datasets.BuilderConfig(name="first_domain", version=VERSION, description="This covers the training dataset"),
+    ]
+
+    DEFAULT_CONFIG_NAME = "first_domain"
```
lhoestq, line 58:
```suggestion
```
You don't need all this if there's only one config

After this change you'll need to replace all the occurrences of "first_domain" by "default" in the dataset_infos.json file. You'll also need to move the dummy_data.zip file from `datasets/caner/dummy/first_domain/1.1.0` to `datasets/caner/dummy/default/1.1.0`.

datasets/caner/caner.py
```diff
@@ -0,0 +1,133 @@
...
+        my_urls = _URL
+        data_dir = dl_manager.download_and_extract(my_urls)
+
+        return [
+            datasets.SplitGenerator(
+                name=datasets.Split.TRAIN,
+                # These kwargs will be passed to _generate_examples
+                gen_kwargs={
+                    "filepath": os.path.join(
+                        data_dir, "Classical-Arabic-Named-Entity-Recognition-Corpus-master/CANERCorpus.csv"
```
lhoestq, line 114: I noticed that in the CANER github repository there're also many txt files.
What's the difference between the txt files and this csv ?

I'm asking because in this csv there's one row per token but it seems that we can't reconstruct the sentences. In general for NER datasets we try to have sentences instead of individual tokens (see [conll2003](https://github.com/huggingface/datasets/blob/master/datasets/conll2003/conll2003.py) for example, in which each sample contains several tokens)

If there's no way to get the sentences we should stick with the individual tokens though, as it is right now.

lhoestq (commented):
Nice thank you !

I left a few comments

Also it looks like the dummy_data.zip file is quite big (600KB). Could you try to reduce its size ?
To do so feel free to take a look inside it and in the csv files only keep the first 5 lines.

datasets/caner/caner.py
KMFODA: Thanks for the comments, very helpful. I put the dummy_data zip in `datasets/caner/dummy/default/1.1.0` however the dummy_data tests failed. Putting it in `datasets/caner/dummy/1.1.0` passed the tests. Not sure if this behaviour is unexpected?

datasets/caner/caner.py
KMFODA: Good spot, although I had realised that we try to have sentences instead of individual tokens but when I read through the paper and looked through all the datasets (in txt and csv versions) none had any indications as to when a sentence starts or ends unfortunately.

KMFODA: Also reduced the dummy_data.zip file as requested.

datasets/caner/caner.py
lhoestq: Thanks !
Yes indeed `datasets/caner/dummy/1.1.0` is the right path, my bad ;)

datasets/caner/README.md
```diff
@@ -0,0 +1,172 @@
...
+- Named Entity Recognition
+
+### Languages
+
+Classical Arabic
+
+## Dataset Structure
+
+### Data Instances
+
```
lhoestq, line 69: We can add an example from the dataset here:

````suggestion

An example from the dataset:
```
{'ner_tags': 1, 'tokens': 'الجامع'}
```
Where 1 stands for "Book"
````

File: datasets/caner/caner.py [hunk 9]
```diff
@@ -0,0 +1,127 @@
...
+    """Classical Arabic Named Entity Recognition corpus as a new corpus of tagged data that can be useful for handling the issues in recognition of Arabic named entities"""
+
+    VERSION = datasets.Version("1.1.0")
+
+    def _info(self):
+
+        features = datasets.Features(
+            {
+                "tokens": datasets.Value("string"),
+                "ner_tags": datasets.ClassLabel(
```
lhoestq, line 59: Since there's only one token per example maybe we should rename the fields `token` and `ner_tag`. What do you think ?

lhoestq (commented):
Thanks for the changes !

datasets/caner/caner.py
KMFODA: yes that makes sense since it's just one token and tag. Will do that now :)

lhoestq (approved):
Thanks for the changes !

Looks all good now :)

PR #1690, opened 2021-01-05
Fast start up

lhoestq:
Currently if optional dependencies such as tensorflow, torch, apache_beam, faiss and elasticsearch are installed, then it takes a long time to do `import datasets` since it imports all of these heavy dependencies.

To make a fast start up for `datasets` I changed that so that they are not imported when `datasets` is being imported. On my side it changed the import time of `datasets` from 5sec to 0.5sec, which is enjoyable.

To be able to check if optional dependencies are available without importing them I'm using `importlib_metadata`, which is part of the standard lib in python>=3.8 and was backported. The difference with `importlib` is that it also enables to get the versions of the libraries without importing them.
I added this dependency in `setup.py`.

### DIFF
--- a/src/datasets/search.py
+++ b/src/datasets/search.py
@@ -1,3 +1,4 @@
+import importlib
 import os
 import tempfile
 from typing import TYPE_CHECKING, Dict, List, NamedTuple, Optional, Union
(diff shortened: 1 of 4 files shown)


src/datasets/search.py
```diff
@@ -1,3 +1,4 @@
+import importlib
```
sgugger, line 1: FYI, to use `importlib.util` you should import it as `import importlib.util`.
I have seen errors on some setups (don't know what causes them exactly, maybe a python version problem?) when not importing the submodule explicitly.

thomwolf (approved):
Yes very cool!

and @AmitMY will be happy as well :-)

src/datasets/search.py
lhoestq: thanks, I changed the import to `import importlib.util`

PR #1693, opened 2021-01-06
Fix reuters metadata parsing errors

jbragg:
Was missing the last entry in each metadata category


lhoestq (approved):
Good catch ! Thank you for fixing it

PR #1695, opened 2021-01-07
fix ner_tag bugs

cstorm125:
fix bug that results in `ner_tag` always equal to 'O'.


lhoestq (commented):
Thanks :)

Apparently the dummy_data.zip got removed. Is this expected ?
Also can you remove the `data-pos.conll` file that you added ?

cstorm125:
> Thanks :)
> 
> Apparently the dummy_data.zip got removed. Is this expected ?
> Also can you remove the `data-pos.conll` file that you added ?

Not expected. I forgot to remove the `dummy_data` folder used to create `dummy_data.zip`. 
Changed to only `dummy_data.zip`.

lhoestq (approved):
Thank you :)

PR #1702, opened 2021-01-07
Fix importlib metdata import in py38

lhoestq:
In Python 3.8 there's no need to install `importlib_metadata` since it already exists as `importlib.metadata` in the standard lib.

PR #1703, opened 2021-01-07
Improvements regarding caching and fingerprinting

lhoestq:
This PR adds these features:
- Enable/disable caching
    If disabled, the library will no longer reload cached datasets files when applying transforms to the datasets.
    It is equivalent to setting `load_from_cache` to `False` in dataset transforms.
```python
from datasets import set_caching_enabled

set_caching_enabled(False)
```
- Allow unpicklable functions in `map`
    If an unpicklable function is used, then it's not possible to hash it to update the dataset fingerprint that is used to name cache files. To workaround that, a random fingerprint is generated instead and a warning is raised.
```python
logger.warning(
    f"Transform {transform} couldn't be hashed properly, a random hash was used instead. "
    "Make sure your transforms and parameters are serializable with pickle or dill for the dataset fingerprinting and caching to work. "
    "If you reuse this transform, the caching mechanism will consider it to be different from the previous calls and recompute everything."
)
```

and also (open to discussion, EDIT: actually NOT included):
- Enable/disable fingerprinting
    Fingerprinting allows to have one deterministic fingerprint per dataset state.
    A dataset fingerprint is updated after each transform.
    Re-running the same transforms on a dataset in a different session results in the same fingerprint.
    Disabling the fingerprinting mechanism makes all the fingerprints random.
    Since the caching mechanism uses fingerprints to name the cache files, then cache file names will be different.
    Therefore disabling fingerprinting will prevent the caching mechanism from reloading datasets files that have already been computed.
    Disabling fingerprinting may speed up the lib for users that don't care about this feature and don't want to use caching.
```python
from datasets import set_fingerprinting_enabled

set_fingerprinting_enabled(False)
```

Other details:
- I renamed the `fingerprint` decorator to `fingerprint_transform` since the name was clearly not explicit. This decorator is used on dataset transform functions to allow them to update fingerprints.
- I added some `ignore_kwargs` when decorating transforms with `fingerprint_transform`, to make the fingerprint update not sensible to kwargs like `load_from_cache` or `cache_file_name`.

Todo: tests for set_fingerprinting_enabled + documentation for all the above features


src/datasets/fingerprint.py
```diff
@@ -34,29 +56,31 @@ def __init__(self):
...
             return cls.hash_default(value)
 
-    def update(self, value):
-        self.m.update(f"=={type(value)}==".encode("utf8"))
-        self.m.update(self.hash(value).encode("utf-8"))
+    def update(self, value: Any) -> None:
+        header_for_update = f"=={type(value)}=="
+        value_for_update = self.hash(value)
+        self.m.update(header_for_update.encode("utf8"))
+        self.m.update(value_for_update.encode("utf-8"))
```
lhoestq, line 81: I changed the order here to not update the hash in case `self.hash` fails (for unpicklable objects for example).

thomwolf:
I few comments here for discussion:
- I'm not convinced yet the end user should really have to understand the difference between "caching" and 'fingerprinting", what do you think? I think fingerprinting should probably stay as an internal thing. Is there a case where we want cahing without fingerprinting or vice-versa?
- while I think the random fingerprint mechanism is smart, I have one question: when we disable caching or fingerprinting we also probably don't want the disk usage to grow so we should then try to keep only one cache file. Is it the case currently?
- the warning should be emitted only once per session if possible (we have a mechanism to do that in transformers, you should ask Lysandre/Sylvain)

src/datasets/fingerprint.py
gaganmastana13: @lhoestq

gaganmastana13:
[__](https://github.com/huggingface/datasets/pull/1702)

gaganmastana13:
[__](https://github.com/huggingface/datasets/pull/1702)

lhoestq:
About your points:
- Yes I agree, I just wanted to bring the discussion on this point. Until now fingerprinting hasn't been blocking for user experience. I'll probably remove the enable/disable fingerprinting function to keep things simple from the user's perspective.
- Right now every time a not in-place transform (i.e. map, filter) is applied, a new cache file is created. It is the case even if caching is disabled since disabling it only means that the cache file won't be reloaded. Therefore you're right that it might end up filling the disk with files that won't be reused. I like the idea of keeping only one cache file. Currently all the cache files are kept on disk until the user clears the cache. To be able to keep only one, we need to know if a dataset that has been transformed is still loaded or not. For example
```python
# case 1 - keep both cache files (dataset1 and dataset2)
dataset2 = dataset1.map(...)
# case 2 - keep only the new cache file
dataset1 = dataset1.map(...)
```
In python it doesn't seem trivial to detect such changes. One thing that we can actually do on the other hand is store the cache files in a temporary directory that is cleared when the session closes. I think that's a good a simple solution for this problem.
- Yes good idea ! I don't like spam either :)

thomwolf:
> * To be able to keep only one, we need to know if a dataset that has been transformed is still loaded or not. For example
> 
> ```python
> # case 1 - keep both cache files (dataset1 and dataset2)
> dataset2 = dataset1.map(...)
> # case 2 - keep only the new cache file
> dataset1 = dataset1.map(...)
> ```

I see what you mean. It's a tricky question. One option would be that if caching is deactivated we have a single memory mapped file and have copy act as a copy by reference instead of a copy by value. We will then probably want a `copy()` or `deepcopy()` functionality. Maybe we should think a little bit about it though.

aballigier (approved):

yjernite:
- I like the idea of using a temporary directory per session!
- My intuition is that the default behavior when caching is disabled should probably always leave the original arrow file untouched, and make a second one that can be changed. That way the memory usage is at most twice the original, but we don't force people to re-download every time there's a mistake. Depends on whether we think that disk space or download/processing is more costly to users.
- So we already have a keyword argument for `dataset1 = dataset1.map(..., in_place=True)`?

lhoestq:
> * If the default behavior when caching is disabled is to re-use the same file, I'm a little worried about people making mistakes and having to re-download and process from scratch.

We should distinguish between the caching from load_dataset (base dataset cache files) and the caching after dataset transforms such as map or filter (transformed dataset cache files). When disabling caching only the second type (for map and filter) doesn't reload from cache files.
Therefore nothing is re-downloaded. To re-download the dataset entirely the argument `download_mode="force_redownload"` must be used in `load_dataset`.
Do we have to think more about the naming to make things less confusing in your opinion ?

> * So we already have a keyword argument for `dataset1 = dataset1.map(..., in_place=True)`?

There's no such `in_place` parameter in map, what do you mean exactly ?

lhoestq:
I updated the PR:
- I removed the enable/disable fingerprinting function
- if caching is disabled arrow files are written in a temporary directory that is deleted when session closes
- the warning that is showed when hashing a transform fails is only showed once
- I added the `set_caching_enabled` function to the docs and explained the caching mechanism and its relation with fingerprinting

I would love to have some feedback :)

yjernite:
> > * So we already have a keyword argument for `dataset1 = dataset1.map(..., in_place=True)`?
> 
> There's no such `in_place` parameter in map, what do you mean exactly ?

Sorry, that wasn't clear at all. I was responding to your previous comment about case 1 / case 2. I don't think the behavior should depend on the command, but we could have:

```
# case 1 - keep both cache files (dataset1 and dataset2)
dataset2 = dataset1.map(...)
# case 2 - keep only the new cache file
dataset1 = dataset1.map(..., in_place=True)
```

Case 1 returns a new reference using the new cache file, case 2 returns the same reference

docs/source/processing.rst
```diff
@@ -491,4 +491,51 @@ Both :obj:`datasets.Dataset` and :obj:`datasets.DatasetDict` objects can be save
...
+Locally you can prevent the library from reloading a cached file by using ``load_from_cache=False`` in transforms like :func:`datasets.Dataset.map` for example.
+You can also specify the name of path where the cache file will be written using the parameter ``cache_file_name``.
+
+It is also possible to disable caching globally with :func:`datasets.set_caching_enabled`:
+
+.. code-block::
+
+    >>> from datasets import set_caching_enabled
+    >>> set_caching_enabled(False)
+
```
yjernite, line 533: I would put the code block after the description of what disabling caching entails, people might miss some of the effects otherwise.

lhoestq:
> Sorry, that wasn't clear at all. I was responding to your previous comment about case 1 / case 2. I don't think the behavior should depend on the command, but we could have:
> 
> ```
> # case 1 - keep both cache files (dataset1 and dataset2)
> dataset2 = dataset1.map(...)
> # case 2 - keep only the new cache file
> dataset1 = dataset1.map(..., in_place=True)
> ```
> 
> Case 1 returns a new reference using the new cache file, case 2 returns the same reference

Ok I see !
`in_place` is a parameter that is used in general to designate a transform so I would name that differently (maybe `overwrite` or something like that).
Not sure if it's possible to update an already existing arrow file that is memory-mapped, let me check real quick.
Also it's possible to call `dataset2.cleanup_cache_files()` to delete the other cache files.

docs/source/processing.rst
```diff
@@ -491,4 +491,53 @@ Both :obj:`datasets.Dataset` and :obj:`datasets.DatasetDict` objects can be save
...
+When applying transforms on a dataset, the data are stored in cache files.
+The caching mechanism allows to reload an existing cache file if it's already been computed.
+
+Reloading a dataset is possible since the cache files are named using the dataset fingerprint, which is updated after each transform.
+
+
+Fingerprinting
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The fingerprint of a dataset in a given state is an internal value computed by combining the fingerprint of the previous state and a hash of the latest transform that was applied.
```
thomwolf, line 503:
```suggestion
The fingerprint of a dataset in a given state is an internal value computed by combining the fingerprint of the previous state and a hash of the latest transform that was applied. (Transform are all the processing method for transforming a dataset that we listed in this chapter (:func:`datasets.Dataset.map`, :func:`datasets.Dataset.shuffle`, etc)
```

thomwolf, line 503:
```suggestion
The fingerprint of a dataset in a given state is an internal value computed by combining the fingerprint of the previous state and a hash of the latest transform that was applied. (Transform are all the processing method for transforming a dataset that we listed in this chapter (:func:`datasets.Dataset.map`, :func:`datasets.Dataset.shuffle`, etc)
```

docs/source/processing.rst
```diff
@@ -491,4 +491,53 @@ Both :obj:`datasets.Dataset` and :obj:`datasets.DatasetDict` objects can be save
...
+
+.. code-block::
+
+    >>> from datasets import Dataset
+    >>> dataset1 = Dataset.from_dict({"a": [0, 1, 2]})
+    >>> dataset2 = dataset1.map(lambda x: {"a": x["a"] + 1})
+    >>> print(dataset1._fingerprint, dataset2._fingerprint)
+    d19493523d95e2dc 5b86abacd4b42434
+
+The new fingerprint is a combination of the previous fingerprint and the hash of the given transform. For a transform to be hashable, it needs to be picklable using dill or pickle.
```
thomwolf, line 516:
```suggestion
The new fingerprint is a combination of the previous fingerprint and the hash of the given transform. For a transform to be hashable, it needs to be picklable using dill or pickle. In particular for :func:`datasets.Dataset.map`, you need to provide a picklable processing method to apply on the dataset so that a determinist fingerprint can be computed by hashing the full state of the provided method (the fingerprint is computed taking into account all the dependencies of the method you provide). 
```

docs/source/processing.rst
```diff
@@ -491,4 +491,53 @@ Both :obj:`datasets.Dataset` and :obj:`datasets.DatasetDict` objects can be save
...
+- use :func:`datasets.Dataset.save_to_disk` to save a transformed dataset or it will be deleted when session closes
+- caching doesn't affect :func:`datasets.load_dataset`. If you want to regenerate a dataset from scratch you should use
+the ``download_mode`` parameter in :func:`datasets.load_dataset`.
+
+To disable caching you can run:
+
+.. code-block::
+
+    >>> from datasets import set_caching_enabled
+    >>> set_caching_enabled(False)
```
thomwolf, line 543: You can also query the current status of the caching with :func:`datasets.is_caching_enabled`.

docs/source/package_reference/main_classes.rst
```diff
@@ -31,6 +31,8 @@ The base class :class:`datasets.Dataset` implements a Dataset backed by an Apach
 
 .. autofunction:: datasets.concatenate_datasets
 
+.. autofunction:: datasets.set_caching_enabled
```
thomwolf, line 34: .. autofunction:: datasets.is_caching_enabled

docs/source/processing.rst
```diff
@@ -491,4 +491,53 @@ Both :obj:`datasets.Dataset` and :obj:`datasets.DatasetDict` objects can be save
 Controling the cache behavior
 -----------------------------------
 
-[UNDER CONSTRUCTION]
+When applying transforms on a dataset, the data are stored in cache files.
+The caching mechanism allows to reload an existing cache file if it's already been computed.
+
+Reloading a dataset is possible since the cache files are named using the dataset fingerprint, which is updated after each transform.
```
thomwolf, line 497: Note that the caching extend beyond sessions. Re-running the very same dataset processing methods (in the same order and on the same data files) in a different session will load from the same cache files.

This is possible thanks to a custom hashing function that works with most python objects (see fingerprinting section below).

src/datasets/fingerprint.py
thomwolf: Good point!

thomwolf (approved):
Looks good to me, added a couple of comments on the doc

yjernite (approved):

Issue #878, opened 2020-11-23
Loading Data From S3 Path in Sagemaker


mahesh1amour:
In Sagemaker Im tring to load the data set from S3 path as follows

`train_path = 's3://xxxxxxxxxx/xxxxxxxxxx/train.csv'
    valid_path = 's3://xxxxxxxxxx/xxxxxxxxxx/validation.csv'
    test_path = 's3://xxxxxxxxxx/xxxxxxxxxx/test.csv'
    
    data_files = {}
    data_files["train"] = train_path
    data_files["validation"] = valid_path
    data_files["test"] = test_path
    extension = train_path.split(".")[-1]
    datasets = load_dataset(extension, data_files=data_files, s3_enabled=True)
    print(datasets)`

I getting an error of

`algo-1-7plil_1  |   File "main.py", line 21, in <module>
algo-1-7plil_1  |     datasets = load_dataset(extension, data_files=data_files)
algo-1-7plil_1  |   File "/opt/conda/lib/python3.6/site-packages/datasets/load.py", line 603, in load_dataset
algo-1-7plil_1  |     **config_kwargs,
algo-1-7plil_1  |   File "/opt/conda/lib/python3.6/site-packages/datasets/builder.py", line 155, in __init__
algo-1-7plil_1  |     **config_kwargs,
algo-1-7plil_1  |   File "/opt/conda/lib/python3.6/site-packages/datasets/builder.py", line 305, in _create_builder_config
algo-1-7plil_1  |     m.update(str(os.path.getmtime(data_file)))
algo-1-7plil_1  |   File "/opt/conda/lib/python3.6/genericpath.py", line 55, in getmtime
algo-1-7plil_1  |     return os.stat(filename).st_mtime
algo-1-7plil_1  | FileNotFoundError: [Errno 2] No such file or directory: 's3://lsmv-sagemaker/pubmedbert/test.csv`

But when im trying with pandas , it is able to load from S3

Does the datasets library support S3 path to load

julien-c:
This would be a neat feature

mahesh1amour:
> neat feature

I dint get these clearly, can you please elaborate like how to work on these

thomwolf:
It could maybe work almost out of the box just by using `cached_path` in the text/csv/json scripts, no?

mahesh1amour:
Thanks thomwolf and julien-c

I'm still confusion on what you guys said, 

I have solved the problem as follows:

1. read the csv file using pandas from s3 
[...]

PR #1723, opened 2021-01-12
ADD S3 support for downloading and uploading processed datasets

philschmid:
# What does this PR do?

This PR adds the functionality to load and save `datasets` from and to s3. 
You can save `datasets` with either `Dataset.save_to_disk()` or `DatasetDict.save_to_disk`. 
You can load `datasets` with either `load_from_disk` or `Dataset.load_from_disk()`, `DatasetDict.load_from_disk()`. 

Loading `csv` or `json` datasets from s3 is not implemented. 

To save/load datasets to s3 you either need to provide an `aws_profile`, which is set up on your machine, per default it uses the `default` profile or you have to pass an `aws_access_key_id` and `aws_secret_access_key`. 

The implementation was done with the `fsspec` and `boto3`.

### Example `aws_profile` :

<details>

```python
dataset.save_to_disk("s3://moto-mock-s3-bucket/datasets/sdk", aws_profile="hf-sm")

load_from_disk("s3://moto-mock-s3-bucket/datasets/sdk", aws_profile="hf-sm")
```

</details>

### Example `aws_access_key_id` and `aws_secret_access_key` :

<details>

```python
dataset.save_to_disk("s3://moto-mock-s3-bucket/datasets/sdk",
                            aws_access_key_id="fake_access_key", 
                            aws_secret_access_key="fake_secret_key"
                           )

load_from_disk("s3://moto-mock-s3-bucket/datasets/sdk",
                            aws_access_key_id="fake_access_key", 
                            aws_secret_access_key="fake_secret_key"
                           )
```

</details>

If you want to load a dataset from a public s3 bucket you can pass `anon=True` 

### Example  `anon=True` :

<details>

```python
dataset.save_to_disk("s3://moto-mock-s3-bucket/datasets/sdk", aws_profile="hf-sm")

load_from_disk("s3://moto-mock-s3-bucketdatasets/sdk",anon=True)
```

</details>

### Full Example

```python
import datasets

dataset = datasets.load_dataset("imdb")
print(f"DatasetDict contains {len(dataset)} datasets")
print(f"train Dataset has the size of: {len(dataset['train'])}")

dataset.save_to_disk("s3://moto-mock-s3-bucket/datasets/sdk", aws_profile="hf-sm")

remote_dataset = datasets.load_from_disk("s3://moto-mock-s3-bucket/datasets/sdk", aws_profile="hf-sm")
print(f"DatasetDict contains {len(remote_dataset)} datasets")
print(f"train Dataset has the size of: {len(remote_dataset['train'])}")
```

Related to #878 

I would also adjust the documentation after the code would be reviewed, as long as I leave the PR in "draft" status. Something that we can consider is renaming the functions and changing the `_disk` maybe to `_filesystem`


src/datasets/utils/filesystem_utils.py
```diff
@@ -0,0 +1,49 @@
...
+        elif anon is True:
+            # The connection can be anonymous - in which case only publicly-available, read-only buckets are accessible
+            fs = fsspec.filesystem("s3", anon=True)
+        else:
+
+            # Credentials are refreshable, so accessing your access key / secret key
+            # separately can lead to a race condition. Use this to get an actual matched
+            # set.
+            credentials = Session(profile_name=aws_profile).get_credentials().get_frozen_credentials()
+            fs = fsspec.filesystem("s3", anon=False, key=credentials.access_key, secret=credentials.secret_key)
```
lhoestq, line 32: Maybe users can define their filesystem themselves and pass `fs=` to `save_to_disk` ? This would simplify things while allowing more control to the user regarding authentication for s3 but also other filesystems.

Let me know what you think.

tests/test_arrow_dataset.py
```diff
@@ -179,6 +181,41 @@ def test_dummy_dataset_serialize(self, in_memory):
             self.assertEqual(dset["filename"][0], "my_name-train_0")
             del dset
 
+    @mock_s3
+    def test_dummy_dataset_serialize_s3(self, in_memory):
```
lhoestq, line 185: thank you for the tests !!

src/datasets/arrow_dataset.py
```diff
@@ -472,19 +482,38 @@ def save_to_disk(self, dataset_path: str):
...
+            aws_secret_access_key (:obj:`str`,  `optional`, defaults to :obj:``None``): the aws secret access key used to create the `boto_session` for downloading the data to s3
+            anon (:obj:`boolean`,  `optional`, defaults to :obj:``False``): The connection can be anonymous - in which case only publicly-available, read-only buckets are accessible, for anonymous connection use `anon=True`
         """
+        # copies file from filesystem if it is s3 to local filesystem and modifies dataset_path to temp directory containing local copies
+        if is_remote_filesystem(dataset_path):
+            # gets filesystem from dataset, either s3:// or file://
+            fs, proc_dataset_path = get_filesystem_from_dataset_path(
+                dataset_path, aws_profile, aws_access_key_id, aws_secret_access_key, anon
+            )
+            tmp_dir = tempfile.TemporaryDirectory()
```
lhoestq, line 513: good idea to use a temp dir

lhoestq (commented):
Thank you for your help on this :) 

I left a few comments.
In particular I think we can make things even simpler by adding a `fs` parameter to `save_to_disk`/`load_from_disk` so that the user can have full control on authentication but also on the type of filesystem they want to use.

I'm not too familiar with `moto` unfortunately but if you think I can help fixing the tests let me know.

src/datasets/utils/filesystem_utils.py
philschmid: I have thought of something similar. I already looked into the other known implementations of `fsspec` then i realized that they all have different authentication/authorization `tokens`, `keys`. 
After that, I thought okay, it would be great in the future to be flexible with the filesystem. But now it is maybe enough to hardcode s3 and then learn from user feedback. 

My idea was to extract the filesystem from the uri. `s3:// == 's3'`, `gcs:// == 'gcs` ....

lhoestq: Yea `fsspec` can infer the filesystem from the URI. But this can only be used if `fs=None` for example.

setup.py
```diff
@@ -82,38 +82,44 @@
     "xxhash",
     # for better multiprocessing
     "multiprocess",
+    # for saving datsets to local or s3
+    "fsspec",
+    "fsspec[s3]",
+    # for getting credentials from aws_profile
+    "boto3",
```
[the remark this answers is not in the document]
philschmid: this means to use datasets with s3 you have to install it using `pip install datasets[s3]`?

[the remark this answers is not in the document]
philschmid: how would you include them in the `TEST_REQUIRE` ?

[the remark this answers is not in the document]
lhoestq: > this means to use datasets with s3 you have to install it using `pip install datasets[s3]`?

yes exactly

> how would you include them in the `TEST_REQUIRE` ?

You can include `'fsspec[s3]'` and `'boto3'` in the TEST_REQUIRE

setup.py
```diff
@@ -171,5 +185,5 @@
         "Programming Language :: Python :: 3.7",
         "Topic :: Scientific/Engineering :: Artificial Intelligence",
     ],
-    keywords='datasets machine learning datasets metrics',
-)
+    keywords="datasets machine learning datasets metrics",
+)
```
julien-c, line 189: `black` misconfiguration?

src/datasets/arrow_dataset.py
```diff
@@ -425,17 +425,26 @@ def __setstate__(self, state):
...
         """
-        Save the dataset in a dataset directory
+        Save the dataset in a dataset directory or to a s3 bucket
 
         Args:
-            dataset_path (``str``): path of the dataset directory where the dataset will be saved to
+            dataset_path (``str``): path or s3 uri of the dataset directory where the dataset will be saved to
+            aws_profile (:obj:`str`,  `optional`, defaults to :obj:``default``): the aws profile used to create the `boto_session` for uploading the data to s3
+            aws_access_key_id (:obj:`str`,  `optional`, defaults to :obj:``None``): the aws access key id used to create the `boto_session` for uploading the data to s3
+            aws_secret_access_key (:obj:`str`,  `optional`, defaults to :obj:``None``): the aws secret access key used to create the `boto_session` for uploading the data to s3
```
julien-c, line 438: if we have `boto3` as an optional dependency, maybe there's a way to not use those (verbose) params and use like a profile name or something instead?

setup.py
philschmid: I have the `black` version `20.8b1` installed.

When i `run black --line-length 119 --target-version py36 ./setup.py`
```
All done! ✨ 🍰 ✨
1 file left unchanged.
```

Are the double quotes wrong? If yes I have to investigate why this happens.

src/datasets/arrow_dataset.py
n1t0: As @lhoestq mentioned, we can probably avoid using these params by letting the user provide a custom `fs` directly. I think this has several advantages (avoid having too many params, let's remove code specific to s3, ..)

philschmid: so you would suggest something like that? 
```python
import fsspec

s3 = fsspec.filesystem("s3", anon=False, key=aws_access_key_id, secret=aws_secret_access_key)

dataset.save_to_disk('s3://my-s3-bucket-with-region/dataset/train',fs=s3)
```

What I don't like about that is the customer experience. `fsspec` is not that well documented for the remote filesystems, e.g. when you want to know which "credentials" you need you to have to go to the `s3fs` documentation.

what do you think if we remove the named arguments `aws_profile`... and handle it how `fsspec` does with an `storage_options` dict. 
```python
dataset.save_to_disk('s3://my-s3-bucket-with-region/dataset/train',
                                      storage_options={
                                             'aws_access_key_id': 123,
                                             'aws_secret_access_key': 123
                                      })
```

setup.py
julien-c: I have no idea, @lhoestq is the reference here, it's "his" codebase ;)

src/datasets/arrow_dataset.py
lhoestq: Indeed the docstring of `fsspec.filesystem` is not ideal:
```
Signature: fsspec.filesystem(protocol, **storage_options)
Docstring:
Instantiate filesystems for given protocol and arguments

``storage_options`` are specific to the protocol being chosen, and are
passed directly to the class.
```
Maybe we can have a better documentation on our side instead using a wrapper:
```python
import datasets

fs = datasets.filesystem("s3", anon=False, key=aws_access_key_id, secret=aws_secret_access_key)
```
Where the docstring of `datasets.filesystem` is more complete and includes examples for popular filesystems like s3

setup.py
lhoestq: Copy from our conversation on slack:

Yes if it changes quotes to double quotes it means that your black is not configured like the ones used by the other contributors. We try to avoid this since it can lead to conflicts
(this is not a big deal to change the quote in itself, this is more about the risk of future conflicts)
Maybe you can try to revert the quotes to single quote and try to see why black is changing the quotes to fix it

src/datasets/arrow_dataset.py
lhoestq: Another option would be to make available the filesystems classes easily:
```python
>>> from datasets.filesystems import S3FileSystem  # S3FileSystem si simply the class from s3fs
>>> S3FileSystem?  # show the docstring
```
shows
```
Init signature: s3fs.core.S3FileSystem(*args, **kwargs)
Docstring:     
Access S3 as if it were a file system.

This exposes a filesystem-like API (ls, cp, open, etc.) on top of S3
storage.

Provide credentials either explicitly (``key=``, ``secret=``) or depend
on boto's credential methods. See botocore documentation for more
information. If no credentials are available, use ``anon=True``.

Parameters
----------
anon : bool (False)
    Whether to use anonymous connection (public buckets only). If False,
    uses the key/secret given, or boto's credential resolver (client_kwargs,
    environment, variables, config files, EC2 IAM server, in that order)
key : string (None)
    If not anonymous, use this access key ID, if specified
secret : string (None)
    If not anonymous, use this secret access key, if specified
token : string (None)
    If not anonymous, use this security token, if specified
etc.
```

n1t0: - If we add the various params separately, the more new filesystems we support, the more complicated it becomes. For the user it becomes difficult to know which params to look at, and for us to document everything. It seems like a good option to test things out, but having to change it down the road will require breaking changes, and we usually try to avoid these as much as possible.
- Using some kind of `storage_options` just like `fsspec.filesystem` might be better but it seems difficult to document also. I think the same argument applies to having a `datasets.filesystem` helper.

I think I have a preference for your second option @lhoestq 

- It seems easy to show all the available filesystems to the user, with each of them having a meaningful documentation
- We can probably add tests for those we support explicitly
- If all of them share the same interface, then power users can probably use anything they want from `fsspec` without explicit support from us?

What do you guys think?

lhoestq: Yes I think the second option is interesting and I totally agree with your three points.
Maybe let's start by having `S3FileSystem ` in `datasets.filesystems` and we can add the other ones later.

In the documentation of `save_to_disk`/`load_from_disk` we can then say that any filesystem from `datasets.filesystems` or `fsspec` can be used.

setup.py
philschmid: I double-checked my `black` configuration even reinstalled it. Still making double-quotes. I also checked if there is an option for `--single-quotes` and found this [github issue](https://github.com/psf/black/issues/594), which seems like they are not even supporting `single-quotes`. 

I am going to fix it by hand but for me, it seems not to be a `black` issue. Would be nice if you could run `black --line-length 119 --target-version py36 ./setup.py`  

e.g. https://github.com/aws/sagemaker-python-sdk/blob/master/setup.py is having double quotes too. 
Could it be possible that no one has ever run `black` for `setpu.py` since it is not in `make style`? 

If not would I adjust the `"` to `'` to be hands.

src/datasets/arrow_dataset.py
philschmid: I have rebuilt everything so that you can now pass in a `fsspec` like filesystem.

```python 
from datasets import S3Filesystem, load_from_disk

s3 = datasets.S3FileSystem(key=aws_access_key_id, secret=aws_secret_access_key)  # doctest: +SKIP

dataset = load_from_disk('s3://my-private-datasets/imdb/train',fs=s3)  # doctest: +SKIP

print(len(dataset))
# 25000
```

I also created a "draft" documentation version with a few examples for `S3Fileystem`. It would be nice if take look at everything. Afterwards, I would adjust the documentation of `save_to_dist`/`load_from_disk` 

image

setup.py
lhoestq: Indeed you're right: it's just that black has never been run on `setup.py`.
Therefore your local installation of black is actually fine.
We can keep the setup.py this way then (i.e. with the " instead of the ')

setup.py
```diff
@@ -83,37 +83,43 @@
     # for better multiprocessing
     "multiprocess",
     # to get metadata of optional dependencies such as torch or tensorflow for Python versions that don't have it
-    "importlib_metadata;python_version<'3.8'"
+    "importlib_metadata;python_version<'3.8'",
+    # for saving datsets to local
+    "fsspec",
```
lhoestq, line 88: We need fsspec in the requirements just for the local filesystem right ?
Maybe we can simply have our own LocalFileSystem from [their implementation](https://filesystem-spec.readthedocs.io/en/latest/_modules/fsspec/implementations/local.html) instead in `datasets.filesystem. LocalFileSystem`

src/datasets/arrow_dataset.py
```diff
@@ -425,17 +427,25 @@ def __setstate__(self, state):
...
+            dataset_path (``str``): path (e.g. ``dataset/train``) or remote uri (e.g. ``s3://my-bucket/dataset/train``) of the dataset directory where the dataset will be saved to
+            fs (Optional[:class:`datasets.filesystem.S3FileSystem`,``fsspec.spec.AbstractFileSystem``],  `optional`, defaults ``None``): instance of :class:`datasets.filesystem.S3FileSystem` or ``fsspec.spec.AbstractFileSystem`` used to download the files from remote filesystem.
         """
         assert (
             not self.list_indexes()
         ), "please remove all the indexes using `dataset.drop_index` before saving a dataset"
         self = pickle.loads(pickle.dumps(self))
+
+        if is_remote_filesystem(fs):
+            dataset_path = preproc_dataset_path(dataset_path)
```
lhoestq, line 444: In `save_to_disk`, when the dataset or the indices are in memory (as opposed to being loaded from arrow files on disk), then currently the arrow files are created in `dataset_path` using the ArrowWriter [here](https://github.com/huggingface/datasets/blob/85f02970f1872e9c394b76c37e7eaf2e0fb95bee/src/datasets/arrow_dataset.py#L453) for the dataset and [here](https://github.com/huggingface/datasets/blob/85f02970f1872e9c394b76c37e7eaf2e0fb95bee/src/datasets/arrow_dataset.py#L460) for the indices before being uploaded. However the ArrowWriter expects a local path. Currently if I understand correctly the arrow file is written in a local path, then copied to remote fs and then the local parent dir is deleted when you call 
```
        if "file" not in fs.protocol:
            shutil.rmtree(list(src.parents)[-2])
```

This can be improved on two points:
1. The local path that is used to create the ArrowWriter is currently the s3 path (see the 2 links to the code above). For example uploading a dataset that is in memory to `s3://foo` first creates an arrow file locally at `foo/xxxx.arrow`, then uploads it, and then the local `foo` is deleted. Maybe we can use a temp dir instead ?
2. Currently the local parent dir of the arrow file is always deleted (for remote fs). If the dataset is not in memory (i.e. is loaded from cached arrow files) then we don't want it to be deleted or the users might lose their data. First, this directory contains the arrow file of the current dataset and we don't want it to be deleted, and second this dir may also contain many other arrow files from other processing steps of this dataset or just other splits (train, test, val), as well as metadata files (json).

src/datasets/arrow_dataset.py
```diff
@@ -472,30 +483,54 @@ def save_to_disk(self, dataset_path: str):
...
+            Path(dataset_path).joinpath("dataset_info.json").as_posix(), "r", encoding="utf-8"
+        ) as dataset_info_file:
             dataset_info = json.load(dataset_info_file)
         state["_info"] = json.dumps(dataset_info)
         dataset = Dataset.from_dict({})
         state = {k: state[k] for k in dataset.__dict__.keys()}  # in case we add new fields
         # Change path to absolute path
         for data_file in state.get("_data_files", []) + state.get("_indices_data_files", []):
-            data_file["filename"] = os.path.join(dataset_path, data_file["filename"])
+            data_file["filename"] = Path(dataset_path).joinpath(data_file["filename"]).as_posix()
```
lhoestq, line 529: Why this change here ?
Windows is probably not going to like having posix paths in this field.

src/datasets/dataset_dict.py
```diff
@@ -478,31 +481,51 @@ def shuffle(
...
+            fs (Optional[:class:`datasets.filesystem.S3FileSystem`,``fsspec.spec.AbstractFileSystem``],  `optional`, defaults ``None``): instance of :class:`datasets.filesystem.S3FileSystem` or ``fsspec.spec.AbstractFileSystem`` used to download the files from remote filesystem.
         """
-        os.makedirs(dataset_dict_path, exist_ok=True)
+        if is_remote_filesystem(fs):
+            proc_dataset_dict_path = preproc_dataset_path(dataset_dict_path)
+        else:
+            fs = fsspec.filesystem("file")
+            proc_dataset_dict_path = dataset_dict_path
+
+        os.makedirs(proc_dataset_dict_path, exist_ok=True)
```
lhoestq, line 498: We only need to create it is the destination is a local path.
For remote path we should use a temp dir for the upload as in my previous comment.

src/datasets/dataset_dict.py
```diff
@@ -478,31 +481,51 @@ def shuffle(
...
+        else:
+            fs = fsspec.filesystem("file")
+            proc_dataset_dict_path = dataset_dict_path
+        for k in json.load(
+            fs.open(Path(proc_dataset_dict_path).joinpath("dataset_dict.json").as_posix(), "r", encoding="utf-8")
+        )["splits"]:
+            dataset_dict_split_path = (
+                dataset_dict_path.split("://")[0] + "://" + Path(proc_dataset_dict_path).joinpath(k).as_posix()
+                if is_remote_filesystem(fs)
+                else Path(proc_dataset_dict_path).joinpath(k).as_posix()
```
lhoestq, line 528: Why use posix path for a local path ? I'm not sure Windows will like that

lhoestq (commented):
Nice thanks for updating the API :) 

I left a few comments, mainly regarding how files are handled before being uploaded on S3

docs/source/package_reference/main_classes.rst
```diff
@@ -1,6 +1,16 @@
 Main classes
 ----------------------------------------------------
 
+``Filesystems``
+~~~~~~~~~~~~~~~~~~~~~
```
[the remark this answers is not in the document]
philschmid: that was only for demonstration purpose. So you would move them before `Features`?

setup.py
philschmid: Yes, we need `fsspec` for the local filesystem. So would copy their implementation instead of installing `fsspec` i can give it a try later.

docs/source/package_reference/main_classes.rst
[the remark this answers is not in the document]
lhoestq: Actually `Features` is also a fundamental notion when using the library, but either after Features or MetricInfo is fine IMO

src/datasets/arrow_dataset.py
philschmid: We had to adjust everything where `datasets` is saving/loading something from disk, because `os.path.join` would create, e.g. in Windows something like `C://my-data/tmp\train.arrow`. 

With `Path().as_posix()` it creates always  the correct path for both windows and unix no matter what happens in it

src/datasets/filesystem.py
```diff
@@ -0,0 +1,88 @@
+import fsspec
+import s3fs
```
lhoestq, line 2: Since those are optional dependencies we should check if they actually exist in the users' env before loading them.
This can be done using
```python
import importlib

_has_s3fs = importlib.util.find_spec("s3fs") is not None
```

Then to be able to have the S3FileSystem imported only if s3fs is installed I would replace the filesystem.py by a `filesystem` module with an `__init__.py` that looks like
```python
import importlib

_has_s3fs = importlib.util.find_spec("s3fs") is not None

if _has_s3fs:
    from .s3filesystem import S3FileSystem
```

and `filesystem/s3filesystem.py` that can import `s3fs` since it's imported only if `_has_s3fs` is true.

Let me know what you think.

In the future this can be extended by checking other dependencies like gcsfs or fsspec itself.

src/datasets/arrow_dataset.py
philschmid: 1. So you would suggest creating a `temp_dir` as we have done it for `load_from_disk` [here](https://github.com/huggingface/datasets/blob/85f02970f1872e9c394b76c37e7eaf2e0fb95bee/src/datasets/arrow_dataset.py#L514)?

2. Yes it is only deleted if you save `files` to the remote filesystem. If I am correct this should be solved to if implement a `temp_dir` if the `filesystem` is "remote" ?

src/datasets/dataset_dict.py
philschmid: it is only for windows :/ that's how it works.

src/datasets/arrow_dataset.py
lhoestq: Ok I see ! wasn't aware of that :)

setup.py
philschmid: sadly not possible since `Localfilesystem` is having `AbstractFileSystem` as dependencies, which has several more dependencies. 
https://filesystem-spec.readthedocs.io/en/latest/_modules/fsspec/spec.html#AbstractFileSystem

src/datasets/filesystem.py
philschmid: we have implemented this within our session together

setup.py
lhoestq: We've taken a look with @philschmid and it's not just a class to copy so we'll need to have fsspec as a hard dependency

src/datasets/arrow_dataset.py
philschmid: 1. I have looked into using `temp_dir` instead of the "local" version of the path. The problem is when we are serializing the state the "local" version of the path is used an `fs` determines if its uploaded to s3 or not. If want to work with a `temp_dir` we have to adjust this with conditional too. 
```python
with fs.open(Path(dataset_path).joinpath("state.json").as_posix(), "w", encoding="utf-8") as state_file:
    json.dump(state, state_file, indent=2, sort_keys=True)
with fs.open(
    Path(dataset_path).joinpath("dataset_info.json").as_posix(), "w", encoding="utf-8"
) as dataset_info_file:
    json.dump(dataset_info, dataset_info_file, indent=2, sort_keys=True)
``` 

2. I think what you explain is not a problem since we are saving the dataset to remote and before that, it is "saved" independently in a local representation `my-bucket/datasets/test` of the s3 uri `s3://my-bucket/datasets/test`. So i don´t think there is any risk in data loss

lhoestq: About your two last points:
1. You can write the json files directly in the destination path using the filesystem. The local temp dir is only needed to write the dataset and the indices since the ArrowWriter works only for local paths.
2. Most of the time you don't need to write the dataset and the indices since they're already on disk. In this case it's the `data_files` attributes that tells you where the arrow files are located. By default the arrow files are in some cache directory of the lib along with many other arrow files. We have to make sure that we don't remove them because of `shutil.rmtree(list(src.parents)[-2])` so I would remove this line.

philschmid: 1. but the problem is that `dataset_path` is used for both the ` Arrow_writer` and `json` files so need to add some sort of condition configuration for the `json`.  Sadly I cannot create a link to the code but it is in `arrow_dataset.py` line 482

lhoestq: The temporary dir path can be used instead for the arrow_writer instead of the dataset_path. And dataset_path can be used for the json.

philschmid: yeah but not for `save_to_disk('local/path')` -> there is no `temp_dir`

lhoestq: Indeed in the case of a local path this is the dest path that is used for the arrow_writer.

Another option would be to change that and always pass a temp dir path to the arrow_writer and then use `fs.put`, exactly like in the remote case.

philschmid: i have implemented `save_to_disk()`  using a `tmp_dir` no matter if it is a remote filesystem or a local one.

philschmid:
I created the documentation the loading and saving datasets to filesystem with an example of using `datasets.filesystem.S3Filesystem`. I added a note on the `Saving a processed dataset on disk and reload` saying that it is also possible to use other filesystems and cloud storages such as S3 with a link to the newly created documentation page from me. 
I Attach a screenshot of it here. 
screencapture-localhost-5500-docs-build-html-filesystems-html-2021-01-19-17_16_10

docs/source/filesystems.rst
```diff
@@ -0,0 +1,154 @@
+FileSystems Integration for cloud storages
+====================================================================
+
+Supported Filesystems
+---------------------
+
+Currenlty ``datasets`` offers an s3 filesystem implementation with :class:`datasets.filesystems.S3FileSystem`. ``S3FileSystem`` is a subclass of `s3fs.S3FileSystem <https://s3fs.readthedocs.io/en/latest/api.html>`_, which is a known implemenetation of ``fsspec``.
```
lhoestq, line 7:
```suggestion
Currenlty ``datasets`` offers an s3 filesystem implementation with :class:`datasets.filesystems.S3FileSystem`. ``S3FileSystem`` is a subclass of `s3fs.S3FileSystem <https://s3fs.readthedocs.io/en/latest/api.html>`_, which is a known implementation of ``fsspec``.
```

docs/source/filesystems.rst
```diff
@@ -0,0 +1,154 @@
...
+Furthermore ``datasets`` supports all ``fsspec`` implementations. Currently Known Implementations these are: 
+
+- `s3fs <https://s3fs.readthedocs.io/en/latest/>`_  for Amazon S3 and other compatible stores
+- `gcsfs <https://gcsfs.readthedocs.io/en/latest/>`_ for Google Cloud Storage
+- `adl <https://github.com/dask/adlfs>`_ for Azure DataLake storage
+- `abfs <https://github.com/dask/adlfs>`_ for Azure Blob service
+- `dropbox <https://github.com/MarineChap/dropboxdrivefs>`_ for access to dropbox shares
+- `gdrive <https://github.com/intake/gdrivefs>`_ to access Google Drive and shares (experimental)
+
+These know implementations are going to be nativly added in the near future within ``datasets``.
```
lhoestq, line 18:
```suggestion
These know implementations are going to be natively added in the near future within ``datasets``.
```

docs/source/filesystems.rst
```diff
@@ -0,0 +1,154 @@
...
+      >>> s3 = datasets.filesystems.S3FileSystem(anon=True)  # doctest: +SKIP
+      >>> s3.ls('public-datasets/imdb/train')  # doctest: +SKIP
+      ['dataset_info.json.json','dataset.arrow','state.json']
+
+Listing files from private s3 bucket using ``aws_access_key_id`` and ``aws_secret_access_key``.
+
+.. code-block::
+
+      >>> import datasets
+      >>> s3 = datasets.filesystems..S3FileSystem(key=aws_access_key_id, secret=aws_secret_access_key)  # doctest: +SKIP
```
lhoestq, line 43:
```suggestion
      >>> s3 = datasets.filesystems.S3FileSystem(key=aws_access_key_id, secret=aws_secret_access_key)  # doctest: +SKIP
```

src/datasets/arrow_dataset.py
```diff
@@ -432,77 +434,110 @@ def __setstate__(self, state):
         if self._indices is None and self._indices_data_files:
             self._indices = reader._read_files(self._indices_data_files)
 
-    def save_to_disk(self, dataset_path: str):
+    def save_to_disk(self, dataset_path: str, fs=None):
         """
-        Save the dataset in a dataset directory
+        Saves a dataset to a filesystem using either :class:`datasets.filesystem.S3FileSystem` or ``fsspec.spec.AbstractFileSystem``.
```
lhoestq, line 439:
```suggestion
        Saves a dataset to a dataset directory, or in a filesystem using either :class:`datasets.filesystem.S3FileSystem` or any implementation of ``fsspec.spec.AbstractFileSystem``.
```

src/datasets/arrow_dataset.py
```diff
@@ -432,77 +434,110 @@ def __setstate__(self, state):
...
+            ) as dataset_info_file:
+                json.dump(dataset_info, dataset_info_file, indent=2, sort_keys=True)
+            logger.info("Dataset saved in {}".format(dataset_path))
 
     @staticmethod
-    def load_from_disk(dataset_path: str) -> "Dataset":
-        """Load the dataset from a dataset directory
+    def load_from_disk(dataset_path: str, fs=None) -> "Dataset":
+        """
+        Loads a dataset that was previously saved using ``dataset.save_to_disk(dataset_path)`` from a filesystem using either :class:`datasets.filesystem.S3FileSystem` or ``fsspec.spec.AbstractFileSystem``.
```
lhoestq, line 507:
```suggestion
        Loads a dataset that was previously saved using ``dataset.save_to_disk(dataset_path)`` from a dataset directory, or from a filesystem using either :class:`datasets.filesystem.S3FileSystem` or any implementation of ``fsspec.spec.AbstractFileSystem``.
```

src/datasets/arrow_dataset.py
```diff
@@ -432,77 +434,110 @@ def __setstate__(self, state):
...
+            ``datasets.Dataset`` or ``datasets.DatasetDict``
+                if `dataset_path` is a path of a dataset directory: the dataset requested,
+                if `dataset_path` is a path of a dataset dict directory: a ``datasets.DatasetDict`` with each split.
         """
-        with open(os.path.join(dataset_path, "state.json"), "r", encoding="utf-8") as state_file:
+        # copies file from filesystem if it is remote filesystem to local filesystem and modifies dataset_path to temp directory containing local copies
+        if is_remote_filesystem(fs):
+            src_dataset_path = extract_path_from_uri(dataset_path)
+            tmp_dir = tempfile.TemporaryDirectory()
+            dataset_path = Path(tmp_dir.name).joinpath(src_dataset_path)
```
lhoestq, line 522:
```suggestion
            dataset_path = Path(tmp_dir).joinpath(src_dataset_path)
```
Here you want to use the tmp_dir no ?
Using tmp_dir.name may point to a relative directory

src/datasets/dataset_dict.py
```diff
@@ -478,31 +481,51 @@ def shuffle(
...
+            fs = fsspec.filesystem("file")
+            dest_dataset_dict_path = dataset_dict_path
+        for k in json.load(
+            fs.open(Path(dest_dataset_dict_path).joinpath("dataset_dict.json").as_posix(), "r", encoding="utf-8")
+        )["splits"]:
+            dataset_dict_split_path = (
+                dataset_dict_path.split("://")[0] + "://" + Path(dest_dataset_dict_path).joinpath(k).as_posix()
+                if is_remote_filesystem(fs)
+                else Path(dest_dataset_dict_path).joinpath(k).as_posix()
+            )
```
lhoestq, line 529: Could this way be simpler and do the same thing ?

```suggestion
            dataset_dict_split_path = Path(dataset_dict_path).joinpath(k).as_posix()
```

src/datasets/filesystems/__init__.py
```diff
@@ -0,0 +1,34 @@
+import importlib
+
+import fsspec
+
+
+_has_s3fs = importlib.util.find_spec("s3fs")
```
lhoestq, line 6:
```suggestion
_has_s3fs = importlib.util.find_spec("s3fs") is not None
```
(nit) to have a boolean

src/datasets/filesystems/s3filesystem.py
```diff
@@ -0,0 +1,64 @@
+import s3fs
+
+
+class S3FileSystem(s3fs.S3FileSystem):
+    """
+    ``datasets.filesystems.S3FileSystem`` is a subclass of `s3fs.S3FileSystem <https://s3fs.readthedocs.io/en/latest/api.html>`_, which is a known
+    implemenetation of ``fsspec``. `Filesystem Spec (FSSPEC) <https://filesystem-spec.readthedocs.io/en/latest/?badge=latest>`_  is a project to
```
lhoestq, line 7:
```suggestion
    implementation of ``fsspec``. `Filesystem Spec (FSSPEC) <https://filesystem-spec.readthedocs.io/en/latest/?badge=latest>`_  is a project to
```

src/datasets/filesystems/s3filesystem.py
```diff
@@ -0,0 +1,64 @@
...
+
+      >>> import datasets
+      >>> s3 = datasets.filesystems.S3FileSystem(anon=True)  # doctest: +SKIP
+      >>> s3.ls('public-datasets/imdb/train')  # doctest: +SKIP
+      ['dataset_info.json.json','dataset.arrow','state.json']
+
+      Listing files from private s3 bucket using ``aws_access_key_id`` and ``aws_secret_access_key``.
+
+      >>> import datasets
+      >>> s3 = datasets.filesystems..S3FileSystem(key=aws_access_key_id, secret=aws_secret_access_key)  # doctest: +SKIP
```
lhoestq, line 22:
```suggestion
      >>> s3 = datasets.filesystems.S3FileSystem(key=aws_access_key_id, secret=aws_secret_access_key)  # doctest: +SKIP
```

src/datasets/load.py
```diff
@@ -673,24 +676,32 @@ def load_dataset(
     return ds
 
 
-def load_from_disk(dataset_path: str) -> Union[Dataset, DatasetDict]:
+def load_from_disk(dataset_path: str, fs=None) -> Union[Dataset, DatasetDict]:
     """
-    Load a dataset that was previously saved using ``dataset.save_to_disk(dataset_path)``.
+    Loads a dataset that was previously saved using ``dataset.save_to_disk(dataset_path)`` from a filesystem using either :class:`datasets.filesystem.S3FileSystem` or ``fsspec.spec.AbstractFileSystem``.
```
lhoestq, line 681:
```suggestion
    Loads a dataset that was previously saved using ``dataset.save_to_disk(dataset_path)`` from a dataset directory, or from a filesystem using either :class:`datasets.filesystem.S3FileSystem` or any implementation of ``fsspec.spec.AbstractFileSystem``.
```

tests/test_filesystem.py
```diff
@@ -0,0 +1,54 @@
...
+    os.environ["AWS_SESSION_TOKEN"] = "fake_session_token"
+
+
+@pytest.fixture(scope="function")
+def s3(aws_credentials):
+    with mock_s3():
+        yield boto3.client("s3", region_name="us-east-1")
+
+
+@mock_s3
```
lhoestq, line 26: If I understand correctly you already mock s3 in the `s3` fixture right ?
So maybe you don't need to add `@mock_s3` here ?

lhoestq (commented):
Thanks for all the changes regarding the temp directories !
I added a few comments but it's mostly typos

docs/source/filesystems.rst
```diff
@@ -0,0 +1,154 @@
...
+
+- ``arrow files.arrow``: they contain your dataset's data
+- ``dataset_info.json``: contains the description, citations, etc. of the dataset
+- ``state.json``: contains the list of the arrow files and other informations like the dataset format type, if any (torch or tensorflow for example)
+
+Saving ``encoded_dataset`` to a private s3 bucket using ``aws_access_key_id`` and ``aws_secret_access_key``.
+
+.. code-block::
+
+      >>> from datasets import filesystems
```
lhoestq, line 73:
```suggestion
      >>> from datasets.filesystems import S3FileSystem
```

tests/test_filesystem.py
philschmid: True.

thomwolf (approved):

lhoestq (approved):
Thanks for adding this @philschmid :)

I tried to find a better name for save_to_disk/load_from_disk but actually I like it this way

docs/source/filesystems.rst
```diff
@@ -0,0 +1,154 @@
...
+
+Using ``S3Filesystem`` with ``botocore.session.Session`` and custom ``aws_profile``.
+
+.. code-block::
+
+      >>> import botocore 
+      >>> from datasets.filesystems import S3Filesystem
+      >>> s3_session = botocore.session.Session(profile_name='my_profile_name')
+      >>>
+      >>> s3 = S3FileSystem(sessions=s3_session)  # doctest: +SKIP
```
julien-c, line 55:
```suggestion
      >>> s3 = S3FileSystem(session=s3_session)  # doctest: +SKIP
```

src/datasets/filesystems/s3filesystem.py
```diff
@@ -0,0 +1,64 @@
...
+      >>> s3.ls('my-private-datasets/imdb/train')  # doctest: +SKIP
+      ['dataset_info.json.json','dataset.arrow','state.json']
+
+      Using ``S3Filesystem`` with ``botocore.session.Session`` and custom ``aws_profile``.
+
+      >>> import botocore
+      >>> from datasets.filesystems import S3Filesystem
+      >>> s3_session = botocore.session.Session(profile_name='my_profile_name')
+      >>>
+      >>> s3 = S3FileSystem(sessions=s3_session)  # doctest: +SKIP
```
julien-c, line 32:
```suggestion
      >>> s3 = S3FileSystem(session=s3_session)  # doctest: +SKIP
```

docs/source/processing.rst
```diff
@@ -488,6 +488,10 @@ Saving a dataset creates a directory with various files:
 
 Both :obj:`datasets.Dataset` and :obj:`datasets.DatasetDict` objects can be saved on disk, by using respectively :func:`datasets.Dataset.save_to_disk` and :func:`datasets.DatasetDict.save_to_disk`.
 
+Furthermore it is also possible to save :obj:`datasets.Dataset` and :obj:`datasets.DatasetDict` to other filesystems and cloud storages such as S3 by  by using respectively :func:`datasets.Dataset.save_to_disk`
```
julien-c, line 491:
```suggestion
Furthermore it is also possible to save :obj:`datasets.Dataset` and :obj:`datasets.DatasetDict` to other filesystems and cloud storages such as S3 by using respectively :func:`datasets.Dataset.save_to_disk` 
```

julien-c (approved):

PR #1732, opened 2021-01-13
[GEM Dataset] Added TurkCorpus, an evaluation dataset for sentence simplification.

mounicam:
We want to use TurkCorpus for validation and testing of the sentence simplification task.


datasets/turk/README.md
```diff
@@ -0,0 +1,142 @@
+---
+title: 'TURKCorpus'
+type: Simplification
+motivation: TURKCorpus is a high-quality simplification dataset where each source sentence is associated with 8 human-written simplifications.
```
lhoestq, line 4: Can you provide valid dataset tags here ?
You can find more information here:

datasets/turk/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Languages
+
+TURKCorpus contains English text only (BCP-47: `en`).
+
+## Dataset Structure
+
+### Data Instances
+
+- `simplification` configuration: an instance consists of an original sentence and 8 possible reference simplifications that focus on lexical paraphrasing.
```
lhoestq, line 57: Here we expect an example from the actual dataset object.
You can get one by using `load_dataset` and then taking the first sample `dataset["train"][0]` for example.

datasets/turk/turk.py
```diff
@@ -0,0 +1,124 @@
...
+
+    BUILDER_CONFIGS = [
+        datasets.BuilderConfig(
+            name="simplification",
+            version=VERSION,
+            description="A set of original sentences aligned with 8 possible simplifications for each.",
+        )
+    ]
+
+    DEFAULT_CONFIG_NAME = "simplification"
```
lhoestq, line 78:
```suggestion
```
you don't need that since there's only one config

datasets/turk/turk.py
```diff
@@ -0,0 +1,124 @@
...
+        f"https://raw.githubusercontent.com/cocoxu/simplification/master/data/turkcorpus/{spl}.8turkers.tok.turk.{i}",
+    )
+    for spl in ["tune", "test"]
+    for i in range(8)
+]
+
+_URLs = dict(_URL_LIST)
+
+
+class TurkCorpus(datasets.GeneratorBasedBuilder):
```
lhoestq, line 66:
```suggestion
class Turk(datasets.GeneratorBasedBuilder):
```
This must be the camel case version of the dataset script name `turk`

You'll need to regenerate the dataset_infos.json file after this change

datasets/turk/turk.py
```diff
@@ -0,0 +1,124 @@
...
+            features=features,
+            supervised_keys=None,
+            homepage=_HOMEPAGE,
+            license=_LICENSE,
+            citation=_CITATION,
+        )
+
+    def _split_generators(self, dl_manager):
+        data_dir = dl_manager.download_and_extract(_URLs)
+        if self.config.name == "simplification":
```
lhoestq, line 100: you don't need to check the config name here since there's only one config

datasets/turk/turk.py
```diff
@@ -0,0 +1,124 @@
...
+                ),
+                datasets.SplitGenerator(
+                    name=datasets.Split.TEST,
+                    gen_kwargs={"filepaths": data_dir, "split": "test"},
+                ),
+            ]
+
+    def _generate_examples(self, filepaths, split):
+        """ Yields examples. """
+        if self.config.name == "simplification":
```
lhoestq, line 117: same here

lhoestq (commented):
Really cool thank you :)

I left a few comments.

Also it looks like the dummy_data.zip file is quite big (1MB). Can you try to reduce its size to something around 10KB ?
To do so feel free to take a look at each text file inside dummy_data.zip to only keep 4 or 5 lines (currently there are hundred of lines).

mounicam:
Thank you for the feedback! I updated the code.

lhoestq (approved):
Thank you !
Looks all good now :)

PR #1734, opened 2021-01-13
Fix empty token bug for `thainer` and `lst20`

cstorm125:
add a condition to check if tokens exist before yielding in `thainer` and `lst20`


SBrandeis (approved):
Thanks for updating this!

lhoestq (approved):
Thanks ! Looks all good

Also for information: 

This also updated the number of examples in the dataset_infos.json
For thainer it removed the last example that was empty. This is thanks to your fix :) 
For lst20 on the other hand it looks like the raw data files got updated and some examples got removed:
- 67104 -> 63310 examples in train split
- 6094 -> 5620 examples in validation split
- 5733 -> 5250 examples in test split

The new numbers of examples are the right ones given the info in the dataset card:


|                      | train     | eval        | test        | all       |
|----------------------|-----------|-------------|-------------|-----------|
| ...              | ...   | ...      | ...      | ...   |
| sentences            | 63,310    | 5,620       | 5,250       | 74,180    |
| ...              | ...   | ...      | ...      | ...   |

PR #1736, opened 2021-01-13
Adjust BrWaC dataset features name

jonatasgrosman:
I added this dataset some days ago, and today I used it to train some models and realized that the names of the features aren't so good.

Looking at the current features hierarchy, we have "paragraphs" with a list of "sentences" with a list of "sentences?!". But the actual hierarchy is a "text" with a list of "paragraphs" with a list of "sentences".

I confused myself trying to use the dataset with these names. So I think it's better to change it.


SBrandeis (approved):
Thanks for updating this!

lhoestq (approved):
Right ! thanks for making the names clearer and updating the dataset card and the dataset_infos.json accordingly

PR #1738, opened 2021-01-14
Conda support

LysandreJik:
Will push a new version on anaconda cloud every time a tag starting with `v` is pushed (like `v1.2.2`).

Will appear here: https://anaconda.org/huggingface/datasets

Depends on `conda-forge` for now, so the following is required for installation:

```
conda install -c huggingface -c conda-forge datasets
```

### DIFF
--- a/.github/workflows/release-conda.yml
+++ b/.github/workflows/release-conda.yml
@@ -3,7 +3,7 @@ name: Release - Conda
 on:
   push:
     tags:
-      - '*'
+      - [0-9]+.[0-9]+.[0-9]+*
 
 env:
   ANACONDA_API_TOKEN: ${{ secrets.ANACONDA_API_TOKEN }}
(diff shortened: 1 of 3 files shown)


lhoestq:
Nice thanks :) 
Note that in `datasets` the tags are simply the version without the `v`. For example `1.2.1`.

LysandreJik:
Do you push tags only for versions?

lhoestq:
Yes I've always used tags only for versions

.github/workflows/release-conda.yml
```diff
@@ -3,7 +3,7 @@ name: Release - Conda
 on:
   push:
     tags:
-      - v*
+      - '*'
```
LysandreJik: Let's try it like this then! Should trigger on all tags.

lhoestq: What about `[0-9]+.[0-9]+.[0-9]+*` ?

LysandreJik: Sure, just pushed it!

lhoestq (approved):
Thank you :)

PR #1752, opened 2021-01-20
COMET metric citation

ricardorei:
In my last pull request to add COMET metric, the citations where not following the usual "format". Because of that they where not correctly displayed on the website: 

Screenshot 2021-01-20 at 09 48 44

This pull request is only intended to fix that.


ricardorei:
I think its better to create a new branch with this fix. I forgot I was still using the old branch.

PR #1760, opened 2021-01-21
More tags

lhoestq:
Since the hub v2 is going to be released soon I figured it would be great to add the missing tags at least for some of the datasets of reference listed [here](https://github.com/huggingface/datasets/blob/master/ADD_NEW_DATASET.md#write-the-loadingprocessing-code)

### DIFF
--- a/datasets/ag_news/README.md
+++ b/datasets/ag_news/README.md
@@ -1,4 +1,22 @@
 ---
+annotations_creators:
+- found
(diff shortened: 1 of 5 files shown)


yjernite:
Conll has `multilingual` but is only tagged as `en`

lhoestq:
good catch, that was a bad copy paste x)

PR #1763, opened 2021-01-21
PAWS-X: Fix csv Dictreader splitting data on quotes

gowtham1997:
```python
# load english paws-x dataset 
datasets = load_dataset('paws-x', 'en')
print(len(datasets['train']))                     # outputs 49202 but official dataset has 49401 pairs
print(datasets['train'].unique('label'))     # outputs [1, 0, -1] but labels are binary [0,1]
```

changed `data = csv.DictReader(f, delimiter="\t")` to `data = csv.DictReader(f, delimiter="\t", quoting=csv.QUOTE_NONE)` in the dataloader to make csv module not split by quotes.

The results are as expected for all languages after the change.


lhoestq (approved):
Good catch ! And thank you for the fix :)

I also removed the code that could make -1 labels, and I updated the dataset_infos.json file as well as the readme

PR #1767, opened 2021-01-22
Add Librispeech ASR

patrickvonplaten:
This PR adds the librispeech asr dataset: https://www.tensorflow.org/datasets/catalog/librispeech

There are 2 configs: "clean" and "other" whereas there are two "train" datasets for "clean", hence the name "train.100" and "train.360".

As suggested by @lhoestq, due to the enormous size of the dataset in `.arrow` format, the speech files are not directly prepared to a float32-array, but instead just the path to the array file is stored.


datasets/librispeech_asr/librispeech_asr.py
```diff
@@ -0,0 +1,153 @@
...
+  organization={IEEE}
+}
+"""
+
+_DESCRIPTION = """\
+LibriSpeech is a corpus of approximately 1000 hours of read English speech with sampling rate of 16 kHz,
+prepared by Vassil Panayotov with the assistance of Daniel Povey. The data is derived from read
+audiobooks from the LibriVox project, and has been carefully segmented and aligned.87
+
+Note that in order to limit the required storage for preparing this dataset, the audio
```
patrickvonplaten, line 43: As discussed offline only processing data samples to a path to the `.flac` file. Added this statement to the config so people know how to retrieve the fp32 array.

src/datasets/load.py
```diff
@@ -161,7 +161,20 @@ def get_imports(file_path: str):
 
     logger.info("Checking %s for additional imports.", file_path)
     imports = []
+    is_in_docstring = False
```
patrickvonplaten, line 164: @lhoestq, would be nice if you can take a look! I think this is pretty save for now. The regex only hits if there is exactly one occurrence of `"""` to switch "in_docstring" mode to True

lhoestq: yes this looks good :)

datasets/librispeech_asr/README.md
```diff
@@ -0,0 +1,164 @@
...
+  - [Licensing Information](#licensing-information)
+  - [Citation Information](#citation-information)
+
+## Dataset Description
+
+- **Homepage:** http://www.openslr.org/12
+- **Repository:** [Needs More Information]
+- **Paper:** https://www.danielpovey.com/files/2015_icassp_librispeech.pdf
+- **Leaderboard:** https://paperswithcode.com/sota/speech-recognition-on-librispeech-test-other
+- **Point of Contact:** dpovey@gmail.com
```
lhoestq, line 54:
```suggestion
- **Homepage:** [LibriSpeech ASR corpus](http://www.openslr.org/12)
- **Repository:** [Needs More Information]
- **Paper:** [LibriSpeech: An ASR Corpus Based On Public Domain Audio Books](https://www.danielpovey.com/files/2015_icassp_librispeech.pdf)
- **Leaderboard:** [Paperswithcode Leaderboard](https://paperswithcode.com/sota/speech-recognition-on-librispeech-test-other)
- **Point of Contact:** [Daniel Povey](mailto:dpovey@gmail.com)
```
add hyperlinks

datasets/librispeech_asr/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Data Instances
+
+A typical data point comprises the path to the audio file, usually called `file` and its transcription, called `text`. Some additional information about the speaker and the passage which contains the transcription is provided.
+
+{'chapter_id': 141231,
+ 'file': '/home/patrick/.cache/huggingface/datasets/downloads/extracted/b7ded9969e09942ab65313e691e6fc2e12066192ee8527e21d634aca128afbe2/dev_clean/1272/141231/1272-141231-0000.flac',
+ 'id': '1272-141231-0000',
+ 'speaker_id': 1272,
+ 'text': 'A MAN SAID TO THE UNIVERSE SIR I EXIST'}
```
lhoestq, line 81:
````suggestion
```
{'chapter_id': 141231,
 'file': '/home/patrick/.cache/huggingface/datasets/downloads/extracted/b7ded9969e09942ab65313e691e6fc2e12066192ee8527e21d634aca128afbe2/dev_clean/1272/141231/1272-141231-0000.flac',
 'id': '1272-141231-0000',
 'speaker_id': 1272,
 'text': 'A MAN SAID TO THE UNIVERSE SIR I EXIST'}
```
````
use code block

File: datasets/librispeech_asr/librispeech_asr.py [hunk 5]
```diff
@@ -0,0 +1,153 @@
...
+```python
+import soundfile as sf
+
+def map_to_array(batch):
+    speech_array, _ = sf.read(batch["file"])
+    batch["speech"] = speech_array
+    return batch
+
+dataset = dataset.map(map_to_array, remove_columns=["file"])
+"""
```
lhoestq, line 57:
````suggestion
```
"""
````
close code block

File: datasets/librispeech_asr/librispeech_asr.py [hunk 6]
```diff
@@ -0,0 +1,153 @@
...
+            features=datasets.Features(
+                {
+                    "file": datasets.Value("string"),
+                    "text": datasets.Value("string"),
+                    "speaker_id": datasets.Value("int64"),
+                    "chapter_id": datasets.Value("int64"),
+                    "id": datasets.Value("string"),
+                }
+            ),
+            supervised_keys=("speech", "text"),
```
lhoestq, line 112: "speech" is not a field of the dataset (even though users can add it) so I don't think we should have it in the supervised_keys here

datasets/librispeech_asr/librispeech_asr.py
```diff
@@ -0,0 +1,153 @@
...
+            citation=_CITATION,
+        )
+
+    def _split_generators(self, dl_manager):
+        archive_path = dl_manager.download_and_extract(_DL_URLS[self.config.name])
+
+        if self.config.name == "clean":
+            train_splits = [
+                datasets.SplitGenerator(name="train.100", gen_kwargs={"archive_path": archive_path["train.100"]}),
+                datasets.SplitGenerator(name="train.360", gen_kwargs={"archive_path": archive_path["train.360"]}),
```
lhoestq, line 123: What's the difference between train.100 and train.360 ?

datasets/librispeech_asr/librispeech_asr.py
```diff
@@ -0,0 +1,153 @@
...
+
+        return train_splits + [
+            datasets.SplitGenerator(name=datasets.Split.VALIDATION, gen_kwargs={"archive_path": archive_path["dev"]}),
+            datasets.SplitGenerator(name=datasets.Split.TEST, gen_kwargs={"archive_path": archive_path["test"]}),
+        ]
+
+    def _generate_examples(self, archive_path):
+        """Generate examples from a Librispeech archive_path."""
+        transcripts_glob = os.path.join(archive_path, "LibriSpeech", "*/*/*/*.txt")
+        for transcript_file in glob.glob(transcripts_glob):
```
lhoestq, line 138:
```suggestion
        for transcript_file in sorted(glob.glob(transcripts_glob)):
```
we need to make the order deterministic here

lhoestq (commented):
Nice thank you !

I added a few comments

datasets/librispeech_asr/librispeech_asr.py
patrickvonplaten: Great catch!

datasets/librispeech_asr/librispeech_asr.py
patrickvonplaten: From the paper:

> The size of the corpus makes it impractical, or at least inconvenient
> for some users, to distribute it as a single large archive. Thus the
> training portion of the corpus is split into three subsets, with approximate size 100, 360 and 500 hours respectively. A simple automatic
> procedure was used to select the audio in the first two sets to be, on
> average, of higher recording quality and with accents closer to US
> English. An acoustic model was trained on WSJ’s si-84 data subset
> and was used to recognize the audio in the corpus, using a bigram
> LM estimated on the text of the respective books. We computed the
> Word Error Rate (WER) of this automatic transcript relative to our
> reference transcripts obtained from the book texts
> ...

> The rest of the audio in the “clean” pool was randomly split into
> two training sets with approximate size 100 and 360 hours respectively. For each speaker in these training sets the amount of speech
> was limited to 25 minutes, in order to avoid major imbalances in
> per-speaker audio duration

> ...

> The “other” pool was similarly split into test and development
> sets, and a single training set of approximately 500 hours.

I'll update the README.md to make this clearer

lhoestq (commented):
Awesome thank you !

The dummy data are quite big but it was expected given that the raw files are flac files.
Given that the script doesn't even read the flac files I think we can remove them. Or maybe use empty flac files (see [here](https://hydrogenaud.io/index.php?topic=118685.0) for example). What do you think ?

We'll find a better solution to be able to have bigger dummy_data (max 1MB instead of a few KB, maybe using git LFS.

patrickvonplaten:
> Awesome thank you !
> 
> The dummy data are quite big but it was expected given that the raw files are flac files.
> Given that the script doesn't even read the flac files I think we can remove them. Or maybe use empty flac files (see [here](https://hydrogenaud.io/index.php?topic=118685.0) for example). What do you think ?
> 
> We'll find a better solution to be able to have bigger dummy_data (max 1MB instead of a few KB, maybe using git LFS.

Hmm, I already made the dummy data as small as possible (a single flac filie per split only). I'd like to keep them at least to have complete dummy data and don't think 500KB over all is a problem (the long-range summarization datasets are similarly heavy). The moment we allow dummy data to be loaded directly for testing, we need the flac files IMO.

PR #1778, opened 2021-01-25
Narrative QA Manual

rsanjaykamath:
Submitting the manual version of Narrative QA script which requires a manual download from the original repository


rsanjaykamath:
@lhoestq sorry I opened a new pull request because of some issues with the previous code base. This pull request is originally from #1364

datasets/narrativeqa_manual/README.md
```diff
@@ -0,0 +1,194 @@
...
+- **Leaderboard:**
+- **Point of Contact:** [Tomáš Kočiský](mailto:tkocisky@google.com) [Jonathan Schwarz](mailto:schwarzjn@google.com) [Phil Blunsom](pblunsom@google.com) [Chris Dyer](cdyer@google.com) [Karl Moritz Hermann](mailto:kmh@google.com) [Gábor Melis](mailto:melisgl@google.com) [Edward Grefenstette](mailto:etg@google.com)
+
+### Dataset Summary
+
+NarrativeQA Manual is an English-language dataset of stories and corresponding questions designed to test reading comprehension, especially on long documents. THIS DATASET REQUIRES A MANUALLY DOWNLOADED FILE! Because of a script in the original repository which downloads the stories from original URLs everytime, the links are sometimes broken or invalid. Therefore, you need to manually download the stories for this dataset using the script provided by the authors (https://github.com/deepmind/narrativeqa/blob/master/download_stories.sh). Running the shell script creates a folder named "tmp" in the root directory and downloads the stories there. This folder containing the stories can be used to load the dataset via `datasets.load_dataset("narrativeqa_manual", data_dir="<path/to/folder>").
```
lhoestq, line 57:
```suggestion
NarrativeQA Manual is an English-language dataset of stories and corresponding questions designed to test reading comprehension, especially on long documents. THIS DATASET REQUIRES A MANUALLY DOWNLOADED FILE! Because of a script in the original repository which downloads the stories from original URLs everytime, the links are sometimes broken or invalid. Therefore, you need to manually download the stories for this dataset using the script provided by the authors (https://github.com/deepmind/narrativeqa/blob/master/download_stories.sh). Running the shell script creates a folder named "tmp" in the root directory and downloads the stories there. This folder containing the stories can be used to load the dataset via `datasets.load_dataset("narrativeqa_manual", data_dir="<path/to/folder>")`.
```
close backtick

datasets/narrativeqa_manual/narrativeqa_manual.py
```diff
@@ -0,0 +1,342 @@
...
+# This can be an arbitrary nested dict/list of URLs (see below in `_split_generators` method)
+_URL = "https://github.com/deepmind/narrativeqa"
+_URLS = {
+    "documents": "https://raw.githubusercontent.com/deepmind/narrativeqa/master/documents.csv",
+    "summaries": "https://raw.githubusercontent.com/deepmind/narrativeqa/master/third_party/wikipedia/summaries.csv",
+    "qaps": "https://raw.githubusercontent.com/deepmind/narrativeqa/master/qaps.csv",
+}
+
+
+class NarrativeQaManual(datasets.GeneratorBasedBuilder):
```
lhoestq, line 71:
```suggestion
class NarrativeqaManual(datasets.GeneratorBasedBuilder):
```
This needs to be the camel case version of the dataset script name `narrativeqa_manual`

datasets/narrativeqa_manual/narrativeqa_manual.py
```diff
@@ -0,0 +1,342 @@
...
+        """Returns SplitGenerators."""
+        data_dir = dl_manager.download_and_extract(_URLS)
+        documents = data_dir["documents"]
+        summaries = data_dir["summaries"]
+        qaps = data_dir["qaps"]
+
+        train_dict = {}
+        test_dict = {}
+        valid_dict = {}
+        with open(documents, encoding="utf-8") as csv_file:
```
lhoestq, line 129: `_split_generators` is not supposed to read the data files. Its role is to define the splits and provide them with the necessary resources (i.e. downloaded data files).

Can you move all this processing to `_generate_examples ` instead ?
In the `gen_kwargs` defined for each split you can simply pass the directories:
```python
return [
            datasets.SplitGenerator(
                name=datasets.Split.TRAIN,
                gen_kwargs={
                    "data_dir": data_dir,
                    "manual_dir": manual_dir,
                    "split": "train",
                },
            ),
            datasets.SplitGenerator(
                name=datasets.Split.TEST,
                gen_kwargs={
                    "data_dir": data_dir,
                    "manual_dir": manual_dir,
                    "split": "test",
                },
            ),
            datasets.SplitGenerator(
                name=datasets.Split.VALIDATION,
                gen_kwargs={
                    "data_dir": data_dir,
                    "manual_dir": manual_dir,
                    "split": "valid",
                },
            ),
        ]
```
And then when generating the examples you can ignore the examples that don't belong to the split that is being generated
```python
            for id_, row in enumerate(csv_reader):
                if row:
                    if row[1] != split:
                        continue
```

datasets/narrativeqa_manual/narrativeqa_manual.py
```diff
@@ -0,0 +1,342 @@
...
+                        train_dict[row[0]] = {
+                            "document_id": row[0],
+                            "kind": row[2],
+                            "story_url": row[3],
+                            "story_file_size": row[4],
+                            "wiki_url": row[5],
+                            "wiki_title": row[6],
+                            "story_word_count": row[7],
+                            "story_start": row[8],
+                            "story_end": row[9],
```
lhoestq, line 146: Maybe you can define the `question-answers` field as an empty list here ?
This way you don't have to use try/except on KeyError when iterating on `qaps `

datasets/narrativeqa_manual/narrativeqa_manual.py
```diff
@@ -0,0 +1,342 @@
...
+
+            for id_, row in enumerate(csv_reader):
+                if row:
+                    if row[1] == "train":
+                        train_dict[row[0]]["summary"] = row[2]
+                        train_dict[row[0]]["summary_tokenized"] = row[3]
+                        try:
+                            train_dict[row[0]]["story_text"] = str(story_texts[row[0]])
+                        except KeyError:
+                            train_dict[row[0]]["story_text"] = ""
```
lhoestq, line 206:
```suggestion
                        train_dict[row[0]]["story_text"] = str(story_texts.get(row[0], ""))
```
I think this is simpler this way

lhoestq (commented):
Really cool thanks ! And good dataset card as well :) 

I left a few comments

rsanjaykamath:
Excellent comments. Thanks for those valuable suggestions. I changed everything as you have pointed out :)

datasets/narrativeqa_manual/narrativeqa_manual.py
```diff
@@ -0,0 +1,248 @@
...
+            homepage=_HOMEPAGE,
+            license=_LICENSE,
+            citation=_CITATION,
+        )
+
+    def _split_generators(self, dl_manager):
+        """Returns SplitGenerators."""
+        data_dir = dl_manager.download_and_extract(_URLS)
+        manual_dir = os.path.abspath(os.path.expanduser(dl_manager.manual_dir))
+
```
lhoestq, line 122:
```suggestion
        if not os.path.exists(manual_dir):
            raise FileNotFoundError(
                "{} does not exist. Make sure you insert a manual dir via `datasets.load_dataset('narrativeqa_manual', data_dir=...)` that includes the stories downloaded from the original repository. Manual download instructions: {}".format(
                    manual_dir, self.manual_download_instructions
                )
            )
```

Could you move the check on the `manual_dir` here please ? This way it's triggered earlier before trying to compute anything

datasets/narrativeqa_manual/narrativeqa_manual.py
```diff
@@ -0,0 +1,248 @@
...
+                            "question": datasets.Value("string"),
+                            "answer1": datasets.Value("string"),
+                            "answer2": datasets.Value("string"),
+                            "question_tokenized": datasets.Value("string"),
+                            "answer1_tokenized": datasets.Value("string"),
+                            "answer2_tokenized": datasets.Value("string"),
+                        }
+                    ),
+                }
+            ),
```
lhoestq, line 111: For consistency I think it could be nice to have the same features here as in narrativeqa, i.e.
```

            features=datasets.Features(
                {
                    "document": {
                        "id": datasets.Value("string"),
                        "kind": datasets.Value("string"),
                        "url": datasets.Value("string"),
                        "file_size": datasets.Value("int32"),
                        "word_count": datasets.Value("int32"),
                        "start": datasets.Value("string"),
                        "end": datasets.Value("string"),
                        "summary": {
                            "tokens": datasets.features.Sequence(datasets.Value("string")),
                            "url": datasets.Value("string"),
                            "title": datasets.Value("string"),
                        },
                    },
                    "question": {
                        "tokens": datasets.features.Sequence(datasets.Value("string")),
                    },
                    "answers": [
                        {
                            "tokens": datasets.features.Sequence(datasets.Value("string")),
                        }
                    ],
                }
            ),
```
What do you think ?

datasets/narrativeqa_manual/README.md
```diff
@@ -0,0 +1,194 @@
...
+    },
+    "question": {
+        "text": "Where does Joe Bloggs live?",
+        "tokens": ["Where", "does", "Joe", "Bloggs", "live", "?"],
+    },
+    "answers": [
+        {"text": "At home", "tokens": ["At", "home"]},
+        {"text": "His house", "tokens": ["His", "house"]}
+    ]
+}
```
lhoestq, line 100: Here we expect an actual example from the dataset object, but the format of this example is not the one used in the `narrativeqa_manual` script at the moment.

However if we switch to using the same `features` as in the `narrativeqa` script then this would be a valid example. And on the contrary if we stick with the current features of `narrativeqa_manual`, we would need to replace this example by a valid one.

datasets/narrativeqa_manual/README.md
```diff
@@ -0,0 +1,194 @@
...
+- `document.file_size` - File size (in bytes) of the story.
+- `document.word_count` - Number of tokens in the story.
+- `document.start` - First 3 tokens of the story. Used for verifying the story hasn't been modified.
+- `document.end` - Last 3 tokens of the story. Used for verifying the story hasn't been modified.
+- `document.summary.text` - Text of the wikipedia summary of the story.
+- `document.summary.tokens` - Tokenized version of `document.summary.text`.
+- `document.summary.url` - Wikipedia URL of the summary.
+- `document.summary.title` - Wikipedia Title of the summary.
+- `question` - `{"text":"...", "tokens":[...]}` for the question about the story.
+- `answers` - List of `{"text":"...", "tokens":[...]}` for valid answers for the question.
```
lhoestq, line 117: Same for the fields described here.

lhoestq (commented):
Thanks for moving all the processing to _generate_examples !

I was wondering whether we should use the same features format as in the `narrative_qa` dataset script (same column names, same structure) for consistency. Let me know what you think

datasets/narrativeqa_manual/narrativeqa_manual.py
rsanjaykamath: Ok I've done this.

datasets/narrativeqa_manual/narrativeqa_manual.py
rsanjaykamath: Ok I've copied the same template for it now.

rsanjaykamath:
I've copied the same template as NarrativeQA now. Please let me know if this is fine.

lhoestq (commented):
Awesome thank you !!
This looks all good :) 

Just before we merge, I was wondering if you knew why the number of examples in the train set went from 1102 to 32747 in your last commit ? I can't see why the changes in the code would cause such a big difference

rsanjaykamath:
> Awesome thank you !!
> This looks all good :)
> 
> Just before we merge, I was wondering if you knew why the number of examples in the train set went from 1102 to 32747 in your last commit ? I can't see why the changes in the code would cause such a big difference

Ok the change was the way I presented the data. 
In my previous code, I presented a story with a list of questions-answers related to the story per sample. So the total 1102 was the number of stories (not questions). 

In the case of `NarrativeQA`, the code presented each sample data with one single question. So the story gets replicated as many times based on number of questions per story. I felt this was not really memory efficient so I had coded the way I did earlier. 

But since this would be inconsistent as you pointed out, I modified my code to suit the `NarrativeQA` approach. Hope it's clear now :)

lhoestq:
Ok I see ! that makes sense

lhoestq (approved):
Thanks for adding this one :)

rsanjaykamath:
Thanks for your time and helping me with all this :) Really appreciate the hardwork you guys do.

PR #1789, opened 2021-01-28
[BUG FIX] typo in the import path for metrics

yjernite:
This tiny PR fixes a typo introduced in https://github.com/huggingface/datasets/pull/1726 which prevents loading new metrics


lhoestq (approved):
Thanks for fixing this !
It was not caught by the tests because the metrics are still missing certain tests. But adding these tests is in scope for 2.0, I'll add some soon

PR #1791, opened 2021-01-29
Small fix with corrected logging of train vectors

TezRomacH:
Now you can set `train_size` to the whole dataset size via `train_size = -1` and login writes not `Training the index with the first -1 vectors` but (for example) `Training the index with the first 16123 vectors`. And maybe more than dataset length. Logging will be correct


lhoestq (approved):
Indeed ! thank you

Note that using `train_size=-1` doesn't actually use all the vectors since it ignores the last one

PR #1792, opened 2021-01-29
Allow loading dataset in-memory

### LINKED ISSUE #708: Datasets performance slow? - 6.4x slower than in memory dataset
I've been very excited about this amazing datasets project. However, I've noticed that the performance can be substantially slower than using an in-memory dataset.

Now, this is expected I guess, due to memory mapping data using arrow files, and you don't get anything for free. But I was surprised at how much slower.

For example, in the `yelp_polarity` dataset (560000 datapoints, or 17500 batches of 32), it was taking me 3:31 to just get process the data and get it on the GPU (no model involved). Whereas, the equivalent in-memory dataset would finish in just 0:33.

Is this expected? Given that one of the goals of this project is also accelerate dataset processing, this seems a bit slower than I would expect. I understand the advantages of being able to work on datasets that exceed memory, and that's very exciting to me, but thought I'd open this issue to discuss.

For reference I'm running a AMD Ryzen Threadripper 1900X 8-Core Processor CPU, with 128 GB of RAM and an NVME SSD Samsung 960 EVO. I'm running with an RTX Titan 24GB GPU.

I can see with `iotop` that the dataset gets quickly loaded into the system read buffers, and thus doesn't incur any additional IO reads. Thus in theory, all the data *should* be in RAM, but in my benchmark code below it's still 6.4 times slower.

What am I doing wrong? And is there a way to force the datasets to completely load into memory instead of being memory mapped in cases where you want maximum performance?

At 3:31 for 17500 batches, that's 12ms per batch. Does this 12ms just become insignificant as a proportion of forward and backward passes in practice, and thus it's not worth worrying about this in practice?

In any case, here's my code `benchmark.py`. If you run it with an argument of `memory` it will copy the data into memory before executing the same test.

``` py
import sys
from transformers import DataCollatorWithPadding, BertTokenizerFast
from tqdm import tqdm

if __name__ == '__main__':
    tokenizer = BertTokenizerFast.from_pretrained('bert-base-cased')
    collate_fn = DataCollatorWithPadding(tokenizer, padding=True)

    ds = load_dataset('yelp_polarity')

    def do_tokenize(x):
        return tokenizer(x['text'], truncation=True)

    ds = ds.map(do_tokenize, batched=True)
    ds.set_format('torch', ['input_ids', 'token_type_ids', 'attention_mask'])

    if len(sys.argv) == 2 and sys.argv[1] == 'memory':
        # copy to memory - probably a faster way to do this - but demonstrates the point
        # approximately 530 batches per second - 17500 batches in 0:33
        print('using memory')
        _ds = [data for data in tqdm(ds['train'])]
    else:
        # approximately 83 batches per second - 17500 batches in 3:31
        print('using datasets')
        _ds = ds['train']

    dl = DataLoader(_ds, shuffle=True, collate_fn=collate_fn, batch_size=32, num_workers=4)

    for data in tqdm(dl):
        for k, v in data.items():
            data[k] = v.to('cuda')
```

For reference, my conda environment is [here](https://gist.github.com/05b6101518ff70ed42a858b302a0405d)

Once again, I'm very excited about this library, and how easy it is to load datasets, and to do so without worrying about system memory constraints.

Thanks for all your great work.

albertvillanova:
Allow loading datasets either from:
- memory-mapped file (current implementation)
- from file descriptor, copying data to physical memory

Close #708


albertvillanova:
I am wondering how to test their difference...

src/datasets/arrow_dataset.py
```diff
@@ -262,10 +262,13 @@ def from_file(
...
     ) -> "Dataset":
         """ Instantiate a Dataset backed by an Arrow table at filename """
-        mmap = pa.memory_map(filename)
-        f = pa.ipc.open_stream(mmap)
+        # Stream backed by memory-mapped file / file descriptor
+        stream_from = pa.memory_map if not in_memory else pa.input_stream
+        stream = stream_from(filename)
+        f = pa.ipc.open_stream(stream)
         pa_table = f.read_all()
         data_files = [{"filename": filename}]
```
lhoestq, line 273:
```suggestion
        data_files = [{"filename": filename}] if not in_memory else None
```
We differentiate in-memory datasets from on-disk datasets by looking at `data_files`. If it's empty, then the dataset is in-memory.

We might change that at some point though since it's not flexible in my opinion.

lhoestq:
> ring how to test their difference...

Hmm I don't think pyarrow exposes an API to check if a Table comes from a file that is memory-mapped. In particular since all the buffer/memory logic is in the C++ part of pyarrow.

Otherwise we can still check the difference of RAM used when loading a big chunk of data.

albertvillanova:
> Hmm I don't think pyarrow exposes an API to check if a Table comes from a file that is memory-mapped. In particular since all the buffer/memory logic is in the C++ part of pyarrow.
> 
> Otherwise we can still check the difference of RAM used when loading a big chunk of data.

@lhoestq I think I found the way: `pa.total_allocated_bytes()` :smirk:

lhoestq (commented):
Good job for the in-memory test !

src/datasets/builder.py
```diff
@@ -882,18 +888,16 @@ class GeneratorBasedBuilder(DatasetBuilder):
...
-
-    def __init__(self, *args, **kwargs):
-        self._writer_batch_size = kwargs.pop("writer_batch_size", self._writer_batch_size)
+    def __init__(self, *args, writer_batch_size=None, **kwargs):
         super(GeneratorBasedBuilder, self).__init__(*args, **kwargs)
+        # Batch size used by the ArrowWriter
+        # It defines the number of samples that are kept in memory before writing them
+        # and also the length of the arrow chunks
+        # None means that the ArrowWriter will use its default value
+        self._writer_batch_size = writer_batch_size
```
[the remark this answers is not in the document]
albertvillanova: Sure! I agree it will be more useful with a default class attribute value. :)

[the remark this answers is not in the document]
albertvillanova: @lhoestq, BTW, is there any plan to migrate tests from unittest to pytest?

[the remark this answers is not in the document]
lhoestq: We're already using pytest (even though some parts of the test code uses unittest). In what part of the tests would you like to see such changes ?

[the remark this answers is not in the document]
albertvillanova: I just mentioned it because the test for this part of the code uses unittest, e.g.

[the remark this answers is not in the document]
lhoestq: What would be the necessary changes ? I guess it's about not using the TestCase class and use plain test functions with fixtures instead. Or are you thinking of other changes ? Also I'm curious about the advantages of such a migration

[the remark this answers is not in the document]
lhoestq: I see !
Moreover since we're starting to have several cases where we have to test many cases through the same code this is something we should put forward in terms of best practice, or we'll end up with the issues that come from code duplication in the test code.
I agree for having that as a background task for now. Thanks for the suggestion !

src/datasets/arrow_reader.py
```diff
@@ -319,14 +341,14 @@ def __init__(self, path: str, info: Optional["DatasetInfo"]):
...
-    def _get_dataset_from_filename(self, filename_skip_take):
+    def _get_dataset_from_filename(self, filename_skip_take, in_memory=False):
         """Returns a Dataset instance from given (filename, skip, take)."""
         filename, skip, take = (
             filename_skip_take["filename"],
             filename_skip_take["skip"] if "skip" in filename_skip_take else None,
             filename_skip_take["take"] if "take" in filename_skip_take else None,
         )
-        pa_table = pa.parquet.read_table(filename, memory_map=True)
+        pa_table = pa.parquet.read_table(filename, memory_map=not in_memory)
```
lhoestq, line 351:
```suggestion
        pa_table = pa.parquet.read_table(filename, memory_map=True)
```
The `memory_map` of the parquet reader of pyarrow is only to improve the performance of the parquet file loading, but in the end it will still be loaded in memory. See more details in the doc of [pyarrow.parquet.read_table](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.read_table.html)

src/datasets/builder.py
```diff
@@ -208,6 +208,7 @@ def __init__(
         name=None,
         hash=None,
         features=None,
+        keep_in_memory=False,
```
lhoestq, line 211: The role of a dataset builder is to write the dataset in the cache using `download_and_prepare`, and then return the loaded dataset from disk using `as_dataset`.

I think it would make more sense to have the `keep_in_memory` parameter directly in `as_dataset`.

What do you think ?

tests/test_builder.py
```diff
@@ -612,3 +546,76 @@ def test_config_names(self):
 
             dummy_builder = DummyBuilderWithDefaultConfig(cache_dir=tmp_dir)
             self.assertEqual(dummy_builder.config.name, "a")
+
```
lhoestq, line 549: Thanks a lot for converting to pytest !!

lhoestq (commented):
This is awesome thank you so much !

I left a few comments, let me know what you think.

Also feel free to merge master (again) into your branch since the PR is almost ready to merge :)

src/datasets/arrow_reader.py
albertvillanova: Thanks @lhoestq for pointing this out. Now you make me doubt... Let me check this! ;)

src/datasets/builder.py
albertvillanova: @lhoestq well seen! I totally agree with you. This will eventually help for further decoupling... ;)

src/datasets/arrow_reader.py
albertvillanova: Good point, @lhoestq! I have tested it and you are right: the total_allocated_bytes always increase whether memory_map is True or False.

lhoestq (approved):
Thanks for the change ! this is awesome :)

PR #1801, opened 2021-01-31
[GEM] Updated the source link of the data to update correct tokenized version.

mounicam:


yjernite:
@mounicam we'll keep the original version in the Turk dataset proper, and use the updated file in the GEM aggregated dataset which I'll add later today

@lhoestq do not merge, I'll close when I've submitted the GEM dataset PR :)

yjernite:
Vlosed by https://github.com/huggingface/datasets/pull/1807

PR #1807, opened 2021-02-02
Adding an aggregated dataset for the GEM benchmark

yjernite:
This dataset gathers modified versions of several other conditional text generation datasets which together make up the shared task for the Generation Evaluation and Metrics workshop (think GLUE for text generation)

The changes from the original datasets are detailed in the Dataset Cards on the GEM website, which are linked to in this dataset card.

cc @sebastianGehrmann


lhoestq:
Nice !

PR #1826, opened 2021-02-05
Print error message with filename when malformed CSV

### LINKED ISSUE #1821: Provide better exception message when one of many files results in an exception
I find when I process many files, i.e.

```
train_files = glob.glob('rain*.csv')
validation_files = glob.glob(validation*.csv')
datasets = load_dataset("csv", data_files=dict(train=train_files, validation=validation_files))
```

I sometimes encounter an error due to one of the files being misformed (i.e. no data, or a comma in a field that isn't quoted, etc).

For example, this is the tail of an exception which I suspect is due to a stray comma.

>   File "pandas/_libs/parsers.pyx", line 756, in pandas._libs.parsers.TextReader.read
>   File "pandas/_libs/parsers.pyx", line 783, in pandas._libs.parsers.TextReader._read_low_memory
>   File "pandas/_libs/parsers.pyx", line 827, in pandas._libs.parsers.TextReader._read_rows
>   File "pandas/_libs/parsers.pyx", line 814, in pandas._libs.parsers.TextReader._tokenize_rows
>   File "pandas/_libs/parsers.pyx", line 1951, in pandas._libs.parsers.raise_parser_error
> pandas.errors.ParserError: Error tokenizing data. C error: Expected 2 fields in line 559, saw 3

It would be nice if the exception trace contained the name of the file being processed (I have 250 separate files!)

albertvillanova:
Print error message specifying filename when malformed CSV file.

Close #1821


src/datasets/packaged_modules/csv/csv.py
```diff
@@ -126,9 +126,15 @@ def _generate_tables(self, files):
...
+
+            try:
+                for batch_idx, df in enumerate(csv_file_reader):
+                    pa_table = pa.Table.from_pandas(df, schema=schema)
+                    # Uncomment for debugging (will print the Arrow table size and elements)
+                    # logger.warning(f"pa_table: {pa_table} num rows: {pa_table.num_rows}")
+                    # logger.warning('\n'.join(str(pa_table.slice(i, 1).to_pydict()) for i in range(pa_table.num_rows)))
+                    yield (file_idx, batch_idx), pa_table
+            except ValueError as e:
+                # logger.error(f"Failed to read file '{csv_file_reader.f}' with error {type(e)}: {e}")
```
albertvillanova, line 138: I am wondering why HF sets logging propagation to False...

lhoestq: What's the issue exactly here ? Why use print instead of the logging ?

lhoestq (commented):
Thanks ! This is really useful :)

and thanks for the tests as well

src/datasets/packaged_modules/csv/csv.py
albertvillanova: Yes. I first logged, but nothing appeared. Then I discovered that logging propagation is disabled. And I would like to know why you decided that. :)

lhoestq: We're doing the same thing on `transformers`, I'm not 100% sure why though. The `logging` doc says
> Note If you attach a handler to a logger and one or more of its ancestors, it may emit the same record multiple times. In general, you should not need to attach a handler to more than one logger - if you just attach it to the appropriate logger which is highest in the logger hierarchy, then it will see all events logged by all descendant loggers, provided that their propagate setting is left set to True. A common scenario is to attach handlers only to the root logger, and to let propagation take care of the rest.

However having propagation disabled makes the `caplog`/`capsys` unable to get what's logged.

In `transformers` we're using some utilities to test the logging outputs. It can be used like this:
```python
from transformers import logging
from transformers.testing_utils import CaptureLogger

msg = "Testing 1, 2, 3"
logging.set_verbosity_info()
logger = logging.get_logger("transformers.models.bart.tokenization_bart")
with CaptureLogger(logger) as cl:
    logger.info(msg)
assert cl.out, msg + "\n"
```

Maybe we could have the same here ? The testing utils are in there: 

https://github.com/huggingface/transformers/blob/master/src/transformers/testing_utils.py#L443

We can have the same code in `datasets` for the tests

Also cc @LysandreJik @sgugger who may know more

albertvillanova: Thanks for your detailed answer @lhoestq. 

That is one of the reasons why I would like to know why you disabled propagation: current implementation does not allow any capture by `caplog`/`capsys`. If keeping propagation disabled, I agree with you that we should implement this alternative log-capturing mechanism for tests.

But there is another reason: end user. The typical end user normally does not custom configure HF logging in advance. Thus, currently, they do not have access to (error/warning) logs to the default sys.stderr. That is why in HF code `print`s are used instead: to make sure the user receives HF error/warning messages.

On the other hand, if propagation is enabled:
- our tests can normally use `caplog`/`capsys`
- the typical end user (who does not configure logging) sees the error/warning logs in the terminal. 

  This is so because of the "default" handler: Python logging package has a default handler (only used if the end user has not previously set any custom handler) which sends error/warning logs to the sys.stderr.
- if an advanced end user (or an application using HF) configures logging with custom handlers, then those are used instead of the "default" handler.
- no need of prints for error/warnings

lhoestq: Ok I found why we had propagation disabled in the first place. This was related to TF 1.14 and an issue with absl (more info [here](https://github.com/tensorflow/tensorflow/issues/26691))

So now we should clearly re-enable propagation since the issue is fixed with absl>0.8.0

Moreover I'm not a fan of adding a handler by default (see logging configuration advise [here](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library)) so I would also remove the handler that was added when we disabled propagation.

I'm creating a PR right away to make these changes

lhoestq: PR is here: https://github.com/huggingface/datasets/pull/1845

albertvillanova: Thank you @lhoestq. Much clear logging defaults now, following best practices!!!

I am going to adapt this PR, once yours is merged.

lhoestq: PR is merged :) fee free to merge master into your branch

lhoestq (approved):
Nice thank you !

PR #1834, opened 2021-02-08
Fixes base_url of limit dataset

Paethon:
`test.json` is not available in the master branch of the repository anymore. Linking to a specific commit.


Paethon:
OK, apparently it is a lot more complicated than simply changing the URL? Going to make an issue.

PR #1846, opened 2021-02-08
Make DownloadManager downloaded/extracted paths accessible

### LINKED ISSUE #1831: Some question about raw dataset download info in the project .
Hi , i review the code in 
https://github.com/huggingface/datasets/blob/master/datasets/conll2003/conll2003.py
in the _split_generators function is the truly logic of download raw datasets with dl_manager
and use Conll2003 cls by use import_main_class in load_dataset function
My question is that , with this logic it seems that i can not have the raw dataset download location
in variable in downloaded_files in _split_generators.
If someone also want use huggingface datasets as raw dataset downloader,
how can he retrieve the raw dataset download path from attributes in 
datasets.dataset_dict.DatasetDict ?

albertvillanova:
Make accessible the file paths downloaded/extracted by DownloadManager.

Close #1831.

The approach:
- I set these paths as DownloadManager attributes: these are DownloadManager's concerns
- To access to these from DatasetBuilder, I set the DownloadManager instance as DatasetBuilder attribute: object composition


src/datasets/utils/download_manager.py
```diff
@@ -184,6 +186,7 @@ def download(self, url_or_urls):
         )
         duration = datetime.now() - start_time
         logger.info("Downloading took {} min".format(duration.total_seconds() // 60))
+        self.downloaded_paths = downloaded_path_or_paths
```
lhoestq, line 189: Sometimes users call `.download` several times.
This happens for example if you want to first download and read some kind of manifest that tells you which other files you need to download.

Maybe this could be a dict or a list that is updated each time `.download` is called ?

src/datasets/utils/download_manager.py
```diff
@@ -233,11 +236,13 @@ def extract(self, path_or_paths, num_proc=None):
         download_config = self._download_config.copy()
         download_config.extract_compressed_file = True
         download_config.force_extract = False
-        return map_nested(
+        extracted_paths = map_nested(
             partial(cached_path, download_config=download_config),
             path_or_paths,
             num_proc=num_proc,
         )
+        self.extracted_paths = extracted_paths
```
lhoestq, line 244: same here

lhoestq (commented):
Nice thanks :) 
And thanks for the test as well. I didn't know about the `monkeypatch` fixture of pytest, I'm used to the unittest patch function but I like this one too

I left a comment about how to handle multiple calls to `.download` in a row

src/datasets/utils/download_manager.py
albertvillanova: Well seen! Thanks for pointing this out, @lhoestq: one, it can be called several times; two, it can return a list or a dict (or a tuple, or a ndarray,...). I am refactoring to take this into account.

lhoestq (commented):
Nice thank you :)
So right now if I understand correctly the `.downloaded_paths` is a list of Union[str, list, dict] right ? I was thinking of having a more user-friendly structure like a flattended list of str, or maybe a dict url->path. What do you think ?

It's possible to have a flattened list of paths with
```python
self.downloaded_paths.extend(flatten_nested(downloaded_path_or_paths))
```
And if you think a dict would be more appropriate feel free to take a look at the `_record_sizes_checksums` method which does something really similar (but storing a dict url -> checksum_dict)

albertvillanova:
First I was thinking of the dict, which makes sense for .download, mapping URL to downloaded path. However does this make sense for .extract, mapping the downloaded path to the extracted path? I ask this because the user did not chose the downloaded path, so this is completely unknown for them...

lhoestq:
There could be several situations:
- download a file with no extraction
- download a file and extract it
- download a file, extract it and then inside the output folder extract some more files
- extract a local file (for datasets with data that are manually downloaded for example)
- extract a local file, and then inside the output folder extract some more files

So I think it's ok to have `downloaded_paths` as a dict url -> downloaded_path and `extracted_paths` as a dict local_path -> extracted_path.

albertvillanova:
OK. I am refactoring this. I have opened #1879...

lhoestq (approved):
Nice thank you !

PR #1853, opened 2021-02-09
Configure library root logger at the module level

albertvillanova:
Configure library root logger at the datasets.logging module level (singleton-like).

By doing it this way:
- we are sure configuration is done only once: module level code is only runned once
- no need of global variable
- no need of threading lock


lhoestq (approved):
Thanks :)

PR #1870, opened 2021-02-12
Implement Dataset add_item

### LINKED ISSUE #1854: Feature Request: Dataset.add_item
I'm trying to integrate `huggingface/datasets` functionality into `fairseq`, which requires (afaict) being able to build a dataset through an `add_item` method, such as https://github.com/pytorch/fairseq/blob/master/fairseq/data/indexed_dataset.py#L318, as opposed to loading all the text into arrow, and then `dataset.map(binarizer)`.
Is this possible at the moment? Is there an example? I'm happy to use raw `pa.Table` but not sure whether it will support uneven length entries.

### Desired API

```python
import numpy as np
tokenized: List[np.NDArray[np.int64]] = [np.array([4,4,2]), np.array([8,6,5,5,2]), np.array([3,3,31,5])

def build_dataset_from_tokenized(tokenized: List[np.NDArray[int]]) -> Dataset:
   """FIXME"""
   dataset = EmptyDataset()
   for t in tokenized: dataset.append(t)
   return dataset
ds = build_dataset_from_tokenized(tokenized)
assert (ds[0] == np.array([4,4,2])).all()
```

### What I tried
grep, google for "add one entry at a time", "datasets.append"

### Current Code
This code achieves the same result but doesn't fit into the `add_item` abstraction.

```python
    dataset = load_dataset('text', data_files={'train': 'train.txt'})
    tokenizer = RobertaTokenizerFast.from_pretrained('roberta-base', max_length=4096)
    def tokenize_function(examples):
        ids = tokenizer(examples['text'], return_attention_mask=False)['input_ids']
        return {'input_ids': [x[1:] for x in ids]}
    ds = dataset.map(tokenize_function, batched=True, num_proc=4, remove_columns=['text'], load_from_cache_file=not overwrite_cache)
	print(ds['train'][0]) => np array
```

Thanks in advance!

albertvillanova:
Implement `Dataset.add_item`.

Close #1854.


lhoestq (commented):
Nice !

The pyarrow table created from the dict may have different feature types that the current dataset so you may want to cast the features of the table before concatenating.
To do so please use
```python
# get the pyarrow type corresponding to the features
type = self.features.type
# build the schema of the table
schema = pa.schema({col_name: type[col_name].type for col_name in self._data.column_names})
# cast the table
table = table.cast(schema)
```

Also adding examples this way breaks some assumptions regarding `__get_state__` for pickling.
In particular one assumption is that the dataset is either fully in memory (dataset._data_files is empty), or the dataset can be reloaded from disk (using the dataset._data_files).
This assumption was convenient to handle both in-memory and on-disk dataset differently:
- in-memory dataset can just be pickled/unpickled in-memory
- on-disk dataset could be unloaded to only keep the filepaths when pickling, and then reloaded from the disk when unpickling

So I think we'll need to refactor these things first (this is a mandatory thing to do anyway in my opinion).

Maybe we can have a design that allows a Dataset to have a Table that can be rebuilt from heterogenous sources like in-memory tables or on-disk tables ? This could also be further extended in the future

albertvillanova:
Thanks @lhoestq for your remarks. Yes, I agree there are still many issues to be tackled... This PR is just a starting point, so that we can discuss how Dataset should be generalized.

lhoestq:
Sure ! I opened an issue #1877 so we can discuss this specific aspect :)

src/datasets/arrow_dataset.py
```diff
@@ -2653,6 +2653,20 @@ def add_elasticsearch_index(
...
+        Args:
+            item (dict): Item data to be added.
+        """
+        table = InMemoryTable.from_pydict({k: [v] for k, v in item.items()})
+        # Cast item
+        type = self.features.type
+        schema = pa.schema({col_name: type[col_name].type for col_name in self._data.column_names})
+        table = table.cast(schema)
+        # Concatenate tables
+        self._data = concat_tables([self._data, table])
```
albertvillanova, line 2668: Maybe here I could use `ConcatenationTable.from_tables` instead.

albertvillanova: Or even better, `ConcatenationTable.from_blocks`.

lhoestq: Either `concat_tables` or `ConcatenationTable.from_tables` are fine :)

But `ConcatenationTable.from_blocks` only takes `InMemoryTable` or `MemoryMappedTable` objects as input so it may fail if `self._data` is already a `ConcatenationTable`

src/datasets/arrow_dataset.py
```diff
@@ -2653,6 +2653,20 @@ def add_elasticsearch_index(
             )
         return self
 
+    def add_item(self, item: dict):
```
lhoestq, line 2656: I think we should make this method return a new Dataset object for consistency with all the other methods.
WDYT ?

lhoestq (commented):
Cool !

Also that means that if someone call `add_item` 100 times, then we end up with 100 InMemoryTable objects.

Maybe we can have a consolidation step ?
For example we can merge successive InMemoryTable objects into one InMemoryTable in a ConcatenationTable ?
This will help speed up subsequent `ConcatenationTable.slice` calls for example, since it iterates on the `table.blocks`.
This should also speed up `Dataset.__getitem__`.

src/datasets/arrow_dataset.py
albertvillanova: Good. I leave as it is then.

albertvillanova:
I am going to implement this consolidation step in #2151.

lhoestq:
Sounds good !

src/datasets/arrow_dataset.py
albertvillanova: Right! Done @lhoestq.

src/datasets/arrow_dataset.py
```diff
@@ -2848,6 +2848,26 @@ def add_elasticsearch_index(
...
+        Args:
+            item (dict): Item data to be added.
+
+        Returns:
+            :class:`Dataset`
+        """
+        item_table = InMemoryTable.from_pydict({k: [v] for k, v in item.items()})
+        # Cast item
+        type = self.features.type
+        schema = pa.schema({col_name: type[col_name].type for col_name in self._data.column_names})
```
lhoestq, line 2865:
```suggestion
        schema = pa.schema(self.features.type)
```
It's maybe simpler to do it this way ?

lhoestq (approved):
thank you !

src/datasets/arrow_dataset.py
albertvillanova: Sure!!! 😅

PR #1873, opened 2021-02-13
add iapp_wiki_qa_squad

cstorm125:
`iapp_wiki_qa_squad` is an extractive question answering dataset from Thai Wikipedia articles.
It is adapted from [the original iapp-wiki-qa-dataset](https://github.com/iapp-technology/iapp-wiki-qa-dataset)
to [SQuAD](https://rajpurkar.github.io/SQuAD-explorer/) format, resulting in
5761/742/739 questions from 1529/191/192 articles.


datasets/iapp_wiki_qa_squad/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+### Curation Rationale
+
+[More Information Needed]
+
+### Source Data
+
+#### Initial Data Collection and Normalization
+
+From the original `iapp-wiki-qa-dataset`, [@cstorm125](https://github.com/cstorm125/) applied the following processing:
```
lhoestq, line 136: Do you think it can be worth having this dataset as well ?

lhoestq (approved):
Awesome thank you !

datasets/iapp_wiki_qa_squad/README.md
cstorm125: Not at this stage since the dataset is very raw. I am in contact with the dataset provider @kobkrit and we think it might be better to include SQuAD version first.

PR #1875, opened 2021-02-14
Adding sari metric

ddhruvkr:
Adding SARI metric that is used in evaluation of text simplification. This is required as part of the GEM benchmark.

### DIFF
--- a/metrics/sari/sari.py
+++ b/metrics/sari/sari.py
@@ -274,12 +274,11 @@ class Sari(datasets.Metric):
         )
 
     def _compute(self, sources, predictions, references):
-        srcs = [normalize(sent) for sent in sources]
-        preds = [normalize(sent) for sent in predictions]
-        refs = [[normalize(sent) for sent in ref_sents] for ref_sents in references]
 
-        sari_score = []
-        for i in range(len(srcs)):
-            sari_score.append(SARIsent(srcs[i], preds[i], refs[i]))
-
-        return {"sari": 100.0 * (sum(sari_score) / len(sari_score))}
+        if not (len(sources) == len(predictions) == len(references)):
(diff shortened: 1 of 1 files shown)


metrics/sari/sari.py
```diff
@@ -0,0 +1,285 @@
...
+        )
+
+    def _compute(self, sources, predictions, references):
+        srcs = [normalize(sent) for sent in sources]
+        preds = [normalize(sent) for sent in predictions]
+        refs = [[normalize(sent) for sent in ref_sents] for ref_sents in references]
+
+        sari_score = []
+        for i in range(len(srcs)):
+            sari_score.append(SARIsent(srcs[i], preds[i], refs[i]))
```
lhoestq, line 283: Maybe we could have something that doesn't store the normalized sentences to save memory ?
Also we can add a check that there are as many sources as predictions/references.

Maybe something like that ?

```suggestion
        if not (len(sources) == len(predictions) == len(references)):
            raise ValueError("Sources length must match predictions and references lengths.")

        sari_score = 0
        for src, pred, refs in zip(sources, predictions, references):
            sari_score += SARIsent(
                normalize(src), normalize(pred), [normalize(sent) for sent in refs]
            ))
        sari_score = sari_score / len(predictions)
```

lhoestq (commented):
Nice ! Thanks for adding this metric :)

I added a suggestion to make it more memory efficient, let me know what you think

metrics/sari/sari.py
ddhruvkr: Thanks. I have modified the code based on the suggestions.

lhoestq (approved):
Nice thanks !

PR #1879, opened 2021-02-15
Replace flatten_nested

albertvillanova:
Replace `flatten_nested` with `NestedDataStructure.flatten`.

This is a first step towards having all NestedDataStructure logic as a separated concern, independent of the caller/user of the data structure.

Eventually, all checks (whether the underlying data is list, dict, etc.) will be only inside this class.

I have also generalized the flattening, and now it handles multiple levels of nesting.


src/datasets/utils/download_manager.py
```diff
@@ -122,9 +122,7 @@ def upload(local_file_path):
 
     def _record_sizes_checksums(self, url_or_urls, downloaded_path_or_paths):
```
lhoestq, line 123: Can we start adding type hints here ? Especially since the variable names are the same as in `.download` but the type is different: here they're NestedDataStructure objects

lhoestq (commented):
Nice ! Good first step in separating the logic of nested data structures :)

src/datasets/utils/download_manager.py
albertvillanova: Sure!

albertvillanova:
Hi @lhoestq. If you agree to merge this, I will start separating the logic for NestedDataStructure.map ;)

lhoestq (approved):
Thanks !

PR #1881, opened 2021-02-15
`list_datasets()` returns a list of strings, not objects

pminervini:
Here and there in the docs there is still stuff like this:

```python
>>> datasets_list = list_datasets()
>>> print(', '.join(dataset.id for dataset in datasets_list))
```

However, my understanding is that `list_datasets()` returns a list of strings rather than a list of objects.


lhoestq (approved):
Indeed. Thank you !

PR #1882, opened 2021-02-15
Create Remote Manager

albertvillanova:
Refactoring to separate the concern of remote (HTTP/FTP requests) management.

### DIFF
--- a/src/datasets/utils/file_utils.py
+++ b/src/datasets/utils/file_utils.py
@@ -142,7 +142,7 @@ def head_hf_s3(
     identifier: str, filename: str, use_cdn=False, dataset=True, max_retries=0
 ) -> Union[requests.Response, Exception]:
     try:
-        return http_head(
+        return RemoteManager.http_head(
             hf_bucket_url(identifier=identifier, filename=filename, use_cdn=use_cdn, dataset=dataset),
             max_retries=max_retries,
         )
@@ -434,24 +434,6 @@ def http_get(url, temp_file, proxies=None, resume_size=0, headers=None, cookies=
     progress.close()
 
 
-def http_head(
-    url, proxies=None, headers=None, cookies=None, allow_redirects=True, timeout=10, max_retries=0
-) -> requests.Response:
-    headers = copy.deepcopy(headers) or {}
-    headers["user-agent"] = get_datasets_user_agent(user_agent=headers.get("user-agent"))
-    response = RemoteManager.request_with_retry(
-        verb="HEAD",
-        url=url,
-        proxies=proxies,
-        headers=headers,
-        cookies=cookies,
-        allow_redirects=allow_redirects,
-        timeout=timeout,
-        max_retries=max_retries,
-    )
-    return response
-
-
 def get_from_cache(
     url,
     cache_dir=None,
@@ -509,7 +491,7 @@ def get_from_cache(
         if url.startswith("ftp://"):
             connected = ftp_head(url)
         try:
-            response = http_head(
+            response = RemoteManager.http_head(
                 url,
                 allow_redirects=True,
                 proxies=proxies,
--- a/src/datasets/utils/file_utils.py
+++ b/src/datasets/utils/file_utils.py
@@ -14,7 +14,6 @@ import shutil
 import sys
 import tarfile
 import tempfile
-import time
 import urllib
 from contextlib import closing, contextmanager
 from dataclasses import dataclass
@@ -34,6 +33,7 @@ from tqdm.auto import tqdm
 from .. import __version__, config
 from .filelock import FileLock
 from .logging import WARNING, get_logger
+from .remote_utils import RemoteManager
 
 
 logger = get_logger(__name__)  # pylint: disable=invalid-name
@@ -388,36 +388,6 @@ def get_authentication_headers_for_url(url: str, use_auth_token: Optional[str] =
     return headers
 
 
-def _request_with_retry(
-    verb: str, url: str, max_retries: int = 0, base_wait_time: float = 0.5, max_wait_time: float = 2, **params
-) -> requests.Response:
-    """Wrapper around requests to retry in case it fails with a ConnectTimeout, with exponential backoff
-
-    Args:
-        verb (str): HTTP verb, such as 'GET' or 'HEAD'
-        url (str): The URL of the ressource to fetch
-        max_retries (int): Maximum number of retries, defaults to 0 (no retries)
-        base_wait_time (float): Duration (in seconds) to wait before retrying the first time. Wait time between
-            retries then grows exponentially, capped by max_wait_time.
-        max_wait_time (float): Maximum amount of time between two retries, in seconds
-        **params: Params to pass to `requests.request`
-    """
-    tries, success = 0, False
-    while not success:
-        tries += 1
-        try:
-            response = requests.request(verb.upper(), url, **params)
-            success = True
-        except requests.exceptions.ConnectTimeout as err:
-            if tries > max_retries:
-                raise err
-            else:
-                logger.info(f"{verb} request to {url} timed out, retrying... [{tries/max_retries}]")
-                sleep_time = max(max_wait_time, base_wait_time * 2 ** (tries - 1))  # Exponential backoff
-                time.sleep(sleep_time)
-    return response
-
-
 def ftp_head(url, timeout=2.0):
     try:
         with closing(urllib.request.urlopen(url, timeout=timeout)) as r:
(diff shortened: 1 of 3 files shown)


tests/test_load.py
```diff
@@ -207,14 +208,14 @@ def test_loading_from_the_datasets_hub():
...
+            return original(url, *args, headers=headers, **kwargs)
 
-    def assert_auth(url, *args, headers, **kwargs):
-        assert headers["authorization"] == "Bearer foo"
-        return http_head(url, *args, headers=headers, **kwargs)
+        return wrapped_method
 
-    with patch("datasets.utils.file_utils.http_head") as mock_head:
-        mock_head.side_effect = assert_auth
+    with patch.object(RemoteManager, "http_head", wraps=assert_auth(RemoteManager.http_head)) as mock_head:
```
lhoestq, line 218: Nice, this will make things more convenient for tests :)

albertvillanova: Exactly!!! I just thought of this when I previously monkeypatched the requests library in PR #1846 ;)

src/datasets/utils/remote_utils.py
```diff
@@ -0,0 +1,177 @@
...
+                unit_scale=True,
+                total=total,
+                initial=resume_size,
+                desc="Downloading",
+                disable=not_verbose,
+            )
+            for chunk in response.iter_content(chunk_size=1024):
+                if chunk:  # filter out keep-alive new chunks
+                    progress.update(len(chunk))
+                    callback(chunk)
```
lhoestq, line 135: Why use callbacks here ? It used to write to the file directly right ?

I can understand that writing to a file is out of scope for a method named "get" but in this case `get_from_cache` can just take the iterator returned by `get` and take care of iterating through the chunks, update the progress bar and write ?

This way the one who's writing files is only `get_from_cache`, not HttpClient nor FtpClient

src/datasets/utils/remote_utils.py
```diff
@@ -0,0 +1,177 @@
...
+            return False
+        return True
+
+    @staticmethod
+    def get(url, timeout=2.0, callback=None):
+        try:
+            logger.info(f"Getting through FTP {url}")  # into {temp_file.name}")
+            with urllib.request.urlopen(url, timeout=timeout) as r:
+                if callback:
+                    callback(r)
```
lhoestq, line 173: same here

lhoestq (commented):
Nice ! This is pretty good :) Just have one question about the callbacks:

src/datasets/utils/remote_utils.py
albertvillanova: Yes, you are right. I first implemented callback to keep the parallelism: after the the call, data was written. Afterwards I thought something similar to what you mention, once I realized that with callback=None I was returning a generator and iterating through it could be done outside!

I factorize this! ;)

src/datasets/utils/remote_utils.py
albertvillanova: For this case, however, I need the file open while copying it by chunks, thus within the context manager scope. ;)

src/datasets/utils/remote_utils.py
albertvillanova: The same as below applies here: I must keep the connection open while reading the data. I am going to make a small refactoring to make this clear.

src/datasets/utils/file_utils.py
```diff
@@ -620,16 +507,16 @@ def _resumable_file_manager():
 
             # GET file object
             if url.startswith("ftp://"):
-                ftp_get(url, temp_file, proxies=proxies, resume_size=resume_size, headers=headers, cookies=cookies)
+                FtpClient.get(url, callback=lambda r: shutil.copyfileobj(r, temp_file))
             else:
-                http_get(
+                HttpClient.get(
```
lhoestq, line 512: I feel like the callback logic is not explicit, it makes the code harder to understand imho. This is not javascript.

I would move here all this logic from the HttpClient.get:
```python
            if response.status_code == 416:  # Range not satisfiable
                return
            if callback:
                content_length = response.headers.get("Content-Length")
                total = resume_size + int(content_length) if content_length is not None else None
                not_verbose = bool(logger.getEffectiveLevel() > WARNING)
                progress = tqdm(
                    unit="B",
                    unit_scale=True,
                    total=total,
                    initial=resume_size,
                    desc="Downloading",
                    disable=not_verbose,
                )
                for chunk in response.iter_content(chunk_size=1024):
                    if chunk:  # filter out keep-alive new chunks
                        progress.update(len(chunk))
                        callback(chunk)
```
So that it's the _resumable_file_manager that is responsible for the progress bar and the status codes handling.

I'd be happy to discuss that though.

I think a HttpClient should be quite bare-bone and not do additional magic.
Simply taking care of the header, handling retries and returning the response is enough.

Another possible option would be to give pass as a parameter `local_filename` where to write the data. This option would keep the for loop inside the HttpClient.get and would remove the callback logic. This also has the advantage that we can do the same thing for the FtpClient to have consistent behaviors. Let me know what option sounds best for you.

src/datasets/utils/remote_utils.py
```diff
@@ -0,0 +1,198 @@
...
+            headers["Range"] = "bytes=%d-" % (resume_size,)
+        with cls.request_with_retry(
+            verb="GET",
+            url=url,
+            stream=True,
+            proxies=proxies,
+            headers=headers,
+            cookies=cookies,
+            max_retries=max_retries,
+        ) as response:
```
lhoestq, line 117: Do you really need to use a `with` statement here ?

src/datasets/utils/remote_utils.py
lhoestq: I would remove the callback logic here too.
It's ok if the ftp get must be used with `with` to properly close the file.

lhoestq (commented):
Hi ! I added a comment about the callback logic and some suggestions

src/datasets/utils/file_utils.py
albertvillanova: Hi @lhoestq . Thanks a lot for your critical review: that is exactly the way we will further improve Hugginface Datasets!!! :)

I agree with you: I do not like this "callback" logic either. It is not Pythonic and it is confusing (as you said right, adding hidden logic under the hood). 

I was somehow induced to implement it because of the original call hierarchy:
- temp local file opening
  - remote http/ftp call
     - use again temp local file logic but from within the remote caller scope
As you see, "calback" was a way to be able to pass logic from the outer scope to the inner one...

However, I was not happy with the resulting implementation. "There must be a better way...". And thinking about this during the weekend, I had a new idea: let's change the call hierarchy!!!

I am refactoring all this and once done, let me know your opinion... :)

src/datasets/utils/remote_utils.py
albertvillanova: That is the key point! I use the `with` statement to make explicit that the connection is NOT closed (after calling get) when you pass the parameter `stream=True`.

albertvillanova: The explanation is here: https://requests.readthedocs.io/en/master/user/advanced/#body-content-workflow

> If you set `stream` to True when making a request, Requests cannot release the connection back to the pool unless you consume all the data or call Response.close. This can lead to inefficiency with connections. If you find yourself partially reading request bodies (or not reading them at all) while using `stream=True`, you should make the request within a `with` statement to ensure it's always closed:

    with requests.get('https://httpbin.org/get', stream=True) as r:
        # Do things with the response here.

lhoestq: But `request_with_retry` is not a contextmanager right ? So using `with` doesn't change anything ? Or did I misunderstand something ?

src/datasets/utils/file_utils.py
lhoestq: Changing the hierarchy order sounds good ! :)

src/datasets/utils/remote_utils.py
albertvillanova: `request_with_retry` returns a `requests.Response` and this can be used as a context manager. When using it or not as a context manager? It depends whether the request connection has been kept open or has been closed. When you pass the parameter `stream=True`, the connection keeps open.

albertvillanova:
@lhoestq I have refactorized the logic. Instead of the previous hierarchy call (local temp file opening -> remote call -> use again temp local file logic but from within the remote caller scope), now it is flattened. Schematically:
```python
with src.open() as src_file, dst.open() as dst_file:
    src_file.fetch(dst_file)
```

I have created `RemotePath` (analogue to Path) with method `.open()` that returns `FtpFile`/`HttpFile` (analogue to file-like).

Now I am going to implement `RemotePath.exists()` method (analogue to the Path's method) to check if remote resource is accessible, using `Ftp/Http.head()`.

src/datasets/utils/file_utils.py
```diff
@@ -422,37 +422,38 @@ def get_from_cache(
     if not local_files_only:
         if url.startswith("ftp://"):
             connected = FtpClient.head(url)
```
albertvillanova: This is a bug, isn't it?

lhoestq:
Quick update on this one:
we discussed offline with @albertvillanova on this PR and I think using `fsspec` can help a lot, since it already implements many parts of the abstraction we need to have nice download tools for both http and ftp (and others !)

PR #1883, opened 2021-02-15
Add not-in-place implementations for several dataset transforms

SBrandeis:
Should we deprecate in-place versions of such methods?


src/datasets/arrow_dataset.py
```diff
@@ -650,13 +647,24 @@ def flatten_(self, max_depth=16):
...
+        """Flattens the table.
+        Each column with a struct type is flattened into one column per struct field.
+        Other columns are left unchanged.
+
+        Returns:
+            A copy of the dataset with flattened columns
+        """
+        dataset = copy.deepcopy(self)
+        dataset._fingerprint = new_fingerprint
+        dataset.flatten_(max_depth=max_depth)
```
lhoestq, line 661: flatten_ already updates the fingerprint so we're updating the fingerprint twice here.
We could simply copy paste the code from flatten_. I think it's ok since we may deprecate flatten_ at one point, and the code is short and straightforward.

This also applies for the other transforms. Let me know what you think

lhoestq (commented):
Nice thanks !
I think it would be bool to deprecate the in-place functions indeed for consistency.

Also I think there is also dictionary_encode_column_ that is in-place

src/datasets/arrow_dataset.py
SBrandeis: I also added a `@deprecated` decorator that emits a DeprecationWarning when calling the methods

SBrandeis:
@lhoestq I am not sure how to test `dictionary_encode_column` (in-place version was not tested before)

src/datasets/arrow_dataset.py
lhoestq: Nice thanks !
Although usually deprecation warnings are only emitted once, maybe we can just give an id parameter to the deprecated decorator so that the second time a deprecated function is called we can say that the warning has already been emitted and therefore keep it silent ? We can use a dictionary to keep track of already emitted deprecation warning as in `transformers`.

lhoestq:
I can take a look at dictionary_encode_column tomorrow.
Although it's likely that it doesn't work then. It was added at the beginning of the lib and never tested nor used afaik.

src/datasets/arrow_dataset.py
SBrandeis: Sounds good!

src/datasets/utils/deprecation_utils.py
```diff
@@ -1,13 +1,25 @@
+import logging
 import warnings
 from typing import Callable
 
 
+_emitted_deprecation_warnings = set()
+logger = logging.getLogger(__name__)
```
lhoestq, line 7: can you use the logging module of `datasets` to get the logger ?
```python
from .logging import get_logger

logger = get_logger(__name__)
```

src/datasets/utils/deprecation_utils.py
```diff
@@ -1,13 +1,25 @@
...
+
+
 def deprecated(deprecated_function: Callable):
```
lhoestq, line 10: Could we allow a to add a message to help users remove the warning ?
Users must know that they can replace `dataset.rename_columns_` by `dataset = dataset.rename_column` for example

src/datasets/utils/deprecation_utils.py
```diff
@@ -1,13 +1,25 @@
...
-            category=DeprecationWarning,
-        )
+        func_hash = hash(deprecated_function)
+        if func_hash not in _emitted_deprecation_warnings:
+            warning_msg = (
+                f"{deprecated_function.__name__} is deprecated and will be removed "
+                "in the next major version of datasets."
+            )
+            warnings.warn(message=warning_msg, category=DeprecationWarning)
+            logger.warning(warning_msg)
```
lhoestq, line 21: doesn't this warn twice ?

lhoestq (commented):
Thanks for making the warning emit only once !

lhoestq (approved):
That looks all good now !
I added some tests (especially about pickling) and they're all passing :)

Thank you so much !

lhoestq:
Now let's update the documentation to use the new methods x)

PR #1885, opened 2021-02-15
add missing info on how to add large files

stas00:
Thanks to @lhoestq's instructions I was able to add data files to a custom dataset repo. This PR is attempting to tell others how to do the same if they need to.

@lhoestq

### DIFF
--- a/docs/source/share_dataset.rst
+++ b/docs/source/share_dataset.rst
@@ -220,29 +220,47 @@ Other files can safely be deleted.
 Uploading your files
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Once the repo is cloned, you can add the dataset script and dataset infos.
-
-You can add these files to the staging environment and verify that they have been correctly staged with the ``git
-status`` command:
+Once the repo is cloned, add the files containing the data.
 
 .. code-block:: bash
 
+    cp /somewhere/data/*json .
+    git lfs track *.json
+    git add .gitattributes
+    git commit -m "track json files"
+    git add *json
+
+It's crucial that ``git lfs track`` gets run on the large data files before ``git add``. If later during ``git push`` you get the error:
+
+.. code-block:: bash
+
+  remote: Your push was rejected because it contains files larger than 10M.
+  remote: Please use https://git-lfs.github.com/ to store larger files.
+
+it means you ``git add``ed the data files before telling ``lfs`` to track those.
+
+Now you can add the dataset script and `dataset_infos.json` file:
+
+.. code-block:: bash
+
+    cp /somewhere/data/dataset_infos.json .
+    cp /somewhere/data/load_script.py .
     git add --all
+
+Quickly verify that they have been correctly staged with:
+
+.. code-block:: bash
+
     git status
 
-Finally, the files should be committed:
+Finally, the files are ready to be committed and pushed to the remote:
 
 .. code-block:: bash
 
     git commit -m "First version of the your_dataset_name dataset."
-
-And pushed to the remote:
-
-.. code-block:: bash
-
     git push
 
-This will upload the folder containing the dataset script and dataset infos that we have just prepared.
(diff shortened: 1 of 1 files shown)


docs/source/share_dataset.rst
```diff
@@ -220,29 +220,48 @@ Other files can safely be deleted.
...
 
-You can add these files to the staging environment and verify that they have been correctly staged with the ``git
-status`` command:
+.. code-block:: bash
+
+    cp /somewhere/data/*json .
+    git lfs track *.json
+    git add .gitattributes
+    git commit -m "track json files"
+    git add *json
```
lhoestq, line 232:
```suggestion
    cp /somewhere/data/*.json .
    git lfs track *.json
    git add .gitattributes
    git commit -m "track json files"
    git add *.json
```
*.json instead of *json

lhoestq (approved):
Thanks this is much better !

docs/source/share_dataset.rst
```diff
@@ -220,29 +220,48 @@ Other files can safely be deleted.
...
 
-You can add these files to the staging environment and verify that they have been correctly staged with the ``git
-status`` command:
+.. code-block:: bash
+
+    cp /somewhere/data/*.json .
+    git lfs track *.json
+    git add .gitattributes
+    git commit -m "track json files"
+    git add *.json
```
julien-c, line 232: wouldn't you reverse those two lines? if you want to actually commit your json files?

lhoestq: Yes indeed. Or you would need a second commit after adding the json files

Let's reverse those two line and have one commit

lhoestq: Actually from the git-lfs github page they recommend to commit after each git track command:
https://github.com/git-lfs/git-lfs
> After any invocation of git-lfs-track(1) or git-lfs-untrack(1), you must commit changes to your .gitattributes file.

julien-c: hmm, I don't think they're saying that you have to commit after each change to .gitattributes. Just that you have to commit the file to version control it.

julien-c: .gitattributes is just a text file (like .gitignore etc) there's nothing special about it

lhoestq: Yes right.
What do you think about having two commits ? It's one more command but I feel it makes it clear that your first need to track and add your .gitattributes before actually adding your data files

julien-c: I would personally do just one commit, but no strong opinion

I do not believe you need to `add` your .gitattributes before adding your data files (just track)

lhoestq: Actually I think you do need to add your .gitattributes first.
Let me check

lhoestq: I just did
```
git clone https://huggingface.co/datasets/lhoestq/testt
cd testt
touch dummy.yo
git lfs track *.yo
git add .
git commit -m "test"
git push
```
and dummy.yo doesn't seem to be on lfs
```
❯ curl -I https://huggingface.co/datasets/lhoestq/testt/resolve/main/dummy.yo
HTTP/1.1 200 OK
Server: nginx/1.18.0 (Ubuntu)
Date: Tue, 16 Feb 2021 11:07:28 GMT
Content-Type: application/octet-stream
Content-Length: 0
Connection: keep-alive
X-Powered-By: Express
ETag: "e69de29bb2d1d6434b8b29ae775ad8c2e48c5391"
```

julien-c: there might be a special case for empty files (their sha256 is... non-existent)

Can you try the same thing with a non-empty file?

lhoestq: Indeed it worked for an non-empty file:
```
echo "Hello world" > dummy.foo
git lfs track *.foo
git add .
git commit -m "test2"
git push
```
```
❯ curl -I https://huggingface.co/datasets/lhoestq/testt/resolve/main/dummy.fooHTTP/1.1 302 Found
Server: nginx/1.18.0 (Ubuntu)
Date: Tue, 16 Feb 2021 11:18:06 GMT
Content-Type: text/plain; charset=utf-8
Content-Length: 140
Connection: keep-alive
X-Powered-By: Express
X-Linked-Size: 12
X-Linked-ETag: "1894a19...183ea2afc7c48f"
Location: https://cdn-lfs.huggingface.co/datasets/lhoestq/testt/1894a19...83ea2afc7c48f
Vary: Accept
```

PR #1887, opened 2021-02-16
Implement to_csv for Dataset

SBrandeis:
cc @thomwolf 

`to_csv` supports passing either a file path or a *binary* file object
The writing is batched to avoid loading the whole table in memory


src/datasets/arrow_dataset.py
```diff
@@ -2175,6 +2178,50 @@ def generator():
         logger.info(f"Finished writing TFRecord to {filename}")
         self = None  # delete the dataset reference used by tf_dataset
 
+    def _write_csv(self, file_obj: BinaryIO, batch_size: int, **to_csv_kwargs) -> int:
+        """
+        Writes the pyarrow table as CSV to a binary file handle.
+        Caller is responsible for opening and closing the handle.
+        """
+        written = 0
+        to_csv_kwargs.pop("header", None)
```
lhoestq, line 2187: I think we should allow the header parameter.
something like `head=header if offset == 0 else False`

src/datasets/arrow_dataset.py
```diff
@@ -2175,6 +2178,50 @@ def generator():
...
+        to_csv_kwargs.pop("path_or_buf", None)
+        to_csv_kwargs.pop("encoding", None)
```
lhoestq, line 2189: We should also allow the `encoding` parameter

lhoestq (commented):
Nice thanks !

src/datasets/arrow_dataset.py
```diff
@@ -2175,6 +2178,50 @@ def generator():
...
+
+        for offset in range(0, self._data.num_rows, batch_size):
+            batch = self._data.slice(offset, offset + batch_size)
+            csv_str = batch.to_pandas().to_csv(
```
lhoestq, line 2193: Also note that we have custom features Array2D, Array3D etc. defined in features.py

They correspond to the custom pyarrow types: Array2DExtensionType, Array3DExtensionType etc.
We also have their corresponding custom pandas type PandasArrayExtensionDType

The conversion using to_pandas() for these types is supported since we have the associated pandas type, but I don't know how pandas supports custom types when converting to csv.

I just tested and this code crashes unfortutanely:
```python

features = Features({"a": Array2D(dtype="int32", shape=(2, 2))})
ds = Dataset.from_dict({"a": [[[0, 0], [0, 0]]]}, features=features)
df = ds._data.to_pandas()
df.head()
#                a
# 0 [[0 0]
#  [0 0]]
df["a"].dtype
# <datasets.features.PandasArrayExtensionDtype at 0x17a2ed080>
df.to_csv("a.csv")
```
raises:
```
IndexError                                Traceback (most recent call last)
<ipython-input-7-df968361d1f9> in <module>
----> 1 df.to_csv("a.csv")

~/.virtualenvs/hf-datasets/lib/python3.7/site-packages/pandas/core/generic.py in to_csv(self, path_or_buf, sep, na_rep, float_format, columns, header, index, index_label, mode, encoding, compression, quoting, quotechar, line_terminator, chunksize, date_format, doublequote, escapechar, decimal)
   3202             decimal=decimal,
   3203         )
-> 3204         formatter.save()
   3205 
   3206         if path_or_buf is None:

~/.virtualenvs/hf-datasets/lib/python3.7/site-packages/pandas/io/formats/csvs.py in save(self)
    202             )
    203 
--> 204             self._save()
    205 
    206         finally:

~/.virtualenvs/hf-datasets/lib/python3.7/site-packages/pandas/io/formats/csvs.py in _save(self)
    321                 break
    322 
--> 323             self._save_chunk(start_i, end_i)
    324 
    325     def _save_chunk(self, start_i: int, end_i: int) -> None:

~/.virtualenvs/hf-datasets/lib/python3.7/site-packages/pandas/io/formats/csvs.py in _save_chunk(self, start_i, end_i)
    336                 decimal=self.decimal,
    337                 date_format=self.date_format,
--> 338                 quoting=self.quoting,
    339             )
    340 

~/.virtualenvs/hf-datasets/lib/python3.7/site-packages/pandas/core/internals/blocks.py in to_native_types(self, slicer, na_rep, quoting, **kwargs)
   1819 
   1820         values = np.asarray(values.astype(object))
-> 1821         values[mask] = na_rep
   1822 
   1823         # we are expected to return a 2-d ndarray

IndexError: arrays used as indices must be of integer (or boolean) type
```

It looks like the PandasArrayExtensionArray doesn't implement `isna` and so the mask can't be applied.

lhoestq: I can open a PR to fix that

SBrandeis: Good to know, thanks!
I'll look into it

lhoestq: I merged a fix for the custom features, exporting to csv for them now works as expected :)

src/datasets/arrow_dataset.py
```diff
@@ -2175,6 +2178,50 @@ def generator():
...
+        Writes the pyarrow table as CSV to a binary file handle.
+        Caller is responsible for opening and closing the handle.
+        """
+        written = 0
+        header = to_csv_kwargs.pop("header", True)
+        encoding = to_csv_kwargs.pop("encoding", "utf-8")
+        to_csv_kwargs.pop("path_or_buf", None)
+
+        for offset in range(0, self._data.num_rows, batch_size):
+            batch = self._data.slice(offset, offset + batch_size)
```
lhoestq, line 2192:
```suggestion
            batch = query_table(
                self._data, slice(offset, offset + batch_size), indices=self._indices.column(0) if self._indices is not None else None
            )
```
This should add support for shuffled dataset as well (or any dataset after a shuffle/shard/select transform)
`self._indices` is the mapping between the indices of the dataset the indices of the table.

src/datasets/arrow_dataset.py
```diff
@@ -2175,6 +2178,50 @@ def generator():
...
+        """
+        Writes the pyarrow table as CSV to a binary file handle.
+        Caller is responsible for opening and closing the handle.
+        """
+        written = 0
+        header = to_csv_kwargs.pop("header", True)
+        encoding = to_csv_kwargs.pop("encoding", "utf-8")
+        to_csv_kwargs.pop("path_or_buf", None)
+
+        for offset in range(0, self._data.num_rows, batch_size):
```
lhoestq, line 2191:
```suggestion
        for offset in range(0, len(self), batch_size):
```

SBrandeis: Good to know, thanks!

SBrandeis:
@lhoestq I stumbled upon an interesting failure when adding tests for CSV serialization of `ArrayXD` features (see the failing unit tests in the CI)

It's due to the fact that booleans cannot be converted from arrow format to numpy with copy: https://arrow.apache.org/docs/python/generated/pyarrow.Array.html#pyarrow.Array.to_numpy

lhoestq:
Good catch ! I must be able to fix that one by allowing copies for this kind of arrays.
This is the kind of surprise you get sometimes when playing with arrow x)

lhoestq:
Raising this error for booleans was introduced in https://issues.apache.org/jira/browse/ARROW-2871?jql=text%20~%20%22boolean%20to_numpy%22 without much explanations unfortunately.
So "no copy" only works for primitive types - except booleans.
This is confirmed in the source code at https://github.com/wesm/arrow/blob/c07b9b48cf3e0bbbab493992a492ae47e5b04cad/python/pyarrow/array.pxi#L621

I'm opening a PR to allow copies for booleans...

lhoestq:
I just merged the fix for boolean ArrayXD, feel free to merge from master to see if it fixes the ci :)

SBrandeis:
@lhoestq unfirtunately, arrays of strings (or any other non-primitive type) require a copy too

A list of primitive types can be found here: https://github.com/wesm/arrow/blob/c07b9b48cf3e0bbbab493992a492ae47e5b04cad/python/pyarrow/types.pxi#L821

pyarrow provides a `is_primitive` function to check whether a type is primitive , I used it to set `zero_copy_only`

Also, `PandasArrayExtensionArray.isna` was using `numpy.isnan` which fails for arrays of strings. I replaced it with `pandas.isna`. Let me know what you think! :)

lhoestq (approved):
Awesome !
and thanks for fixing the PandasExtensionArray

PR #1889, opened 2021-02-16
Implement to_dict and to_pandas for Dataset

SBrandeis:
With options to return a generator or the full dataset

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -38,6 +38,7 @@ import pyarrow as pa
 from multiprocess import Pool, RLock
 from tqdm.auto import tqdm
 
+from . import config
 from .arrow_reader import ArrowReader
 from .arrow_writer import ArrowWriter, TypedSequence
 from .features import Features, Value, cast_to_python_objects
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -2208,7 +2208,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
             )
 
     def to_pandas(
-        self, batch_size: Optional[int] = None, batched: bool = True
+        self, batch_size: Optional[int] = None, batched: bool = False
     ) -> Union[pd.DataFrame, Iterator[pd.DataFrame]]:
         """Returns the dataset as a :class:`pandas.DataFrame`. Can also return a generator for large datasets.
 
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -2195,7 +2195,11 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
             `dict` or `Iterator[dict]`
         """
         if not batched:
-            return self._data.to_pydict()
+            return query_table(
+                pa_table=self._data,
+                key=slice(0, len(self)),
+                indices=self._indices.column(0) if self._indices is not None else None,
+            ).to_pydict()
         else:
             batch_size = batch_size if batch_size else config.DEFAULT_MAX_BATCH_SIZE
             return (
@@ -2204,7 +2208,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
                     key=slice(offset, offset + batch_size),
                     indices=self._indices.column(0) if self._indices is not None else None,
                 ).to_pydict()
-                for offset in range(0, self._data.num_rows, batch_size)
+                for offset in range(0, len(self), batch_size)
             )
 
     def to_pandas(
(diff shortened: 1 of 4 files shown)


lhoestq (commented):
Nice thanks !

Also can you use `query_table` with self._indices (same change as I suggested in the to_csv PR) ? This way this will support shuffled/sharded datasets

src/datasets/arrow_dataset.py
```diff
@@ -2181,6 +2182,58 @@ def generator():
...
+            return self._data.to_pydict()
+        else:
+            batch_size = batch_size if batch_size else config.DEFAULT_MAX_BATCH_SIZE
+            return (
+                query_table(
+                    pa_table=self._data,
+                    key=slice(offset, offset + batch_size),
+                    indices=self._indices.column(0) if self._indices is not None else None,
+                ).to_pydict()
+                for offset in range(0, self._data.num_rows, batch_size)
```
lhoestq, line 2207:
```suggestion
                for offset in range(0, len(self), batch_size)
```
if the dataset has an indices mapping, then probably the size of the dataset is not the size of the underlying table, but actually the size of the indices mapping. 
So you can use `len(self)` here, that returns the size of the dataset in both cases.

src/datasets/arrow_dataset.py
```diff
@@ -2181,6 +2182,58 @@ def generator():
...
+            batched (`bool`): Set to :obj:`True` to return a generator that yields the dataset as batches of
+                :param:`batch_size` rows. Defaults to :obj:`False` (returns the whole datasetas once)
+            bacth_size (`Optional[int]`): The size (number of rows) of the batches if :param:`batched` is `True`.
+                Defaults to :obj:`.config.DEFAULT_MAX_BATCH_SIZE`.
+
+        Returns:
+            `dict` or `Iterator[dict]`
+        """
+        if not batched:
+            return self._data.to_pydict()
```
lhoestq, line 2198: You will need to use query_table here as well to take the indices mapping into account (`indices=self._indices.column(0) if self._indices is not None else None`).

Feel free to use the key `slice(0, len(self))` to query the full table after applying the indices mapping.

src/datasets/arrow_dataset.py
```diff
@@ -2181,6 +2182,58 @@ def generator():
...
+            return self._data.to_pandas()
+        else:
+            batch_size = batch_size if batch_size else config.DEFAULT_MAX_BATCH_SIZE
+            return (
+                query_table(
+                    pa_table=self._data,
+                    key=slice(offset, offset + batch_size),
+                    indices=self._indices.column(0) if self._indices is not None else None,
+                ).to_pandas()
+                for offset in range(0, self._data.num_rows, batch_size)
```
lhoestq, line 2234: same: 
```suggestion
                for offset in range(0, len(self), batch_size)
```

src/datasets/arrow_dataset.py
```diff
@@ -2181,6 +2182,58 @@ def generator():
...
+            batched (`bool`): Set to :obj:`True` to return a generator that yields the dataset as batches of
+                :param:`batch_size` rows. Defaults to :obj:`False` (returns the whole datasetas once)
+            bacth_size (`Optional[int]`): The size (number of rows) of the batches if :param:`batched` is `True`.
+                Defaults to :obj:`.config.DEFAULT_MAX_BATCH_SIZE`.
+
+        Returns:
+            `pandas.DataFrame` or `Iterator[pandas.DataFrame]`
+        """
+        if not batched:
+            return self._data.to_pandas()
```
lhoestq, line 2225: same: use query_table

tests/test_arrow_dataset.py
```diff
@@ -1222,6 +1222,52 @@ def test_export(self, in_memory):
...
+                    self.assertLessEqual(len(batch[col_name]), bacth_size)
+
+            # Full
+            dset_to_dict = dset.to_dict()
+            self.assertIsInstance(dset_to_dict, dict)
+            self.assertListEqual(sorted(dset_to_dict.keys()), sorted(dset.column_names))
+            for col_name in dset.column_names:
+                self.assertIsInstance(dset_to_dict[col_name], list)
+                self.assertLessEqual(len(dset_to_dict[col_name]), dset.num_rows)
+
```
lhoestq, line 1246: feel free to add a test where there is an indices mapping
```
dset_with_indices_mapping = dset.select([1, 0])
```
for example.
The new `dset_with_indices_mapping` dataset is of size 2 and its first element is at the position 1 in the pyarrow table, and the second element is at the position 0 in the pyrrow table.

lhoestq (commented):
Nice ! Thanks for adding the support for datasets with an indices mapping.
I added some more comments about query_table and the batched=False case, as well as suggestions to use `len(self)` instead of `self._data.num_rows` in query_table

src/datasets/arrow_dataset.py
SBrandeis: Woops

lhoestq (approved):
Nice thank you !

lhoestq:
Next step is going to add these two in the documentation ^^

Issue #1895, opened 2021-02-16
Bug Report: timestamp[ns] not recognized


justin-yan:
Repro:

```
from datasets import Dataset
import pandas as pd
import pyarrow

df = pd.DataFrame(pd.date_range("2018-01-01", periods=3, freq="H"))
pyarrow.Table.from_pandas(df)
Dataset.from_pandas(df)
# Throws ValueError: Neither timestamp[ns] nor timestamp[ns]_ seems to be a pyarrow data type.
```

The factory function seems to be just "timestamp": https://arrow.apache.org/docs/python/generated/pyarrow.timestamp.html#pyarrow.timestamp

It seems like https://github.com/huggingface/datasets/blob/master/src/datasets/features.py#L36-L43 could have a little bit of additional structure for handling these cases?  I'd be happy to take a shot at opening a PR if I could receive some guidance on whether parsing something like `timestamp[ns]` and resolving it to timestamp('ns') is the goal of this method.

Alternatively, if I'm using this incorrectly (e.g. is the expectation that we always provide a schema when timestamps are involved?), that would be very helpful to know as well!

```
$ pip list  # only the relevant libraries/versions
datasets                      1.2.1
pandas                        1.0.3
pyarrow                       3.0.0
```

lhoestq:
Thanks for reporting !

You're right, `string_to_arrow` should be able to take `"timestamp[ns]"` as input and return the right pyarrow timestamp type.
Feel free to suggest a fix for `string_to_arrow` and open a PR if you want to contribute ! This would be very appreciated :)

To give you more context:

As you may know we define the features types of a dataset using the `Features` object in combination with feature types like `Value`. For example
```python
features = Features({
    "age": Value("int32")
})
```
However under the hood we are actually using pyarrow to store the data, and so we have a mapping between the feature types of `datasets` and the types of pyarrow.

For example, the `Value` feature types are created from a pyarrow type with `Value(str(pa_type))`.
However it looks like the conversion back to a pyarrow type doesn't work with `"timestamp[ns]"`.
This is the `string_to_arrow` function you highlighted that does this conversion, so we should fix that.

justin-yan:
Thanks for the clarification @lhoestq !

This may be a little bit of a stupid question, but I wanted to clarify one more thing before I took a stab at this:

When the features get inferred, I believe they already have a pyarrow schema (https://github.com/huggingface/datasets/blob/master/src/datasets/arrow_dataset.py#L234).

[...]

PR #1900, opened 2021-02-17
Issue #1895: Bugfix for string_to_arrow timestamp[ns] support

justin-yan:
Should resolve https://github.com/huggingface/datasets/issues/1895

The main part of this PR adds additional parsing in `string_to_arrow` to convert the timestamp dtypes that result from `str(pa_type)` back into the pa.DataType TimestampType.

While adding unit-testing, I noticed that support for the double/float types also don't invert correctly, so I added them, which I believe would hypothetically make this section of `Value` redundant:

```
    def __post_init__(self):
        if self.dtype == "double":  # fix inferred type
            self.dtype = "float64"
        if self.dtype == "float":  # fix inferred type
            self.dtype = "float32"
```

However, since I think Value.dtype is part of the public interface, removing that would result in a backward-incompatible change, so I didn't muck with that.

The rest of the PR consists of docstrings that I added while developing locally so I could keep track of which functions were supposed to be inverses of each other, and thought I'd include them initially in case you want to keep them around, but I'm happy to delete or remove any of them at your request!

### DIFF
--- a/src/datasets/features.py
+++ b/src/datasets/features.py
@@ -15,6 +15,7 @@
 
 # Lint as: python3
 """ This class handle features definition in datasets and some utilities to display table type."""
+import re
 from collections.abc import Iterable
 from dataclasses import dataclass, field, fields
 from typing import Any, ClassVar, Dict, List, Optional, Sequence, Tuple, Union
@@ -33,18 +34,60 @@ logger = get_logger(__name__)
 
 
 def string_to_arrow(type_str: str) -> pa.DataType:
-    if type_str not in pa.__dict__:
+    """
+    string_to_arrow takes a string pyarrow dtype and converts it to a pyarrow.DataType.
+
+    In effect, `dt == string_to_arrow(str(dt))`
+
+    This is necessary because the datasets.Value() primitive type is constructed using a string dtype, oftentimes
+
+    Value(dtype=str(pa_type))
+
+    But Features.type (via `get_nested_type()` expects to resolve Features into a pyarrow Schema,
+        which means that each Value() must resolve into its corresponding pyarrow.DataType, which is the purpose
+        of this function.
+    """
+    timestamp_regex = re.compile("^timestamp\[(.*)\]$")
+    timestamp_matches = timestamp_regex.search(type_str)
+    if timestamp_matches:
+        """
+        >>> import pyarrow as pa
+        >>> str(pa.timestamp('us'))
+        'timestamp[us]'
+        >>> str(pa.timestamp('us', tz='America/New_York'))
+        'timestamp[us, tz=America/New_York]'
+        """
+        timestamp_internals = timestamp_matches.group(1)
+        internals_regex = re.compile("^(s|ms|us|ns),\s*tz=([a-zA-Z0-9/_+:]*)$")
+        internals_matches = internals_regex.search(timestamp_internals)
+        if timestamp_internals in ["s", "ms", "us", "ns"]:
+            return pa.timestamp(timestamp_internals)
+        elif internals_matches:
+            return pa.timestamp(internals_matches.group(1), internals_matches.group(2))
+        else:
+            raise ValueError(
+                f"{type_str} is not a validly formatted string representation of a pyarrow timestamp."
+                f"Examples include timestamp[us] or timestamp[us, tz=America/New_York]"
+                f"See: https://arrow.apache.org/docs/python/generated/pyarrow.timestamp.html#pyarrow.timestamp"
+            )
+    elif type_str == "double":
+        return pa.float64()
+    elif type_str == "float":
+        return pa.float32()
+    elif type_str == "halffloat":
+        return pa.float16()
+    elif type_str not in pa.__dict__:
         if str(type_str + "_") not in pa.__dict__:
             raise ValueError(
                 f"Neither {type_str} nor {type_str + '_'} seems to be a pyarrow data type. "
                 f"Please make sure to use a correct data type, see: "
                 f"https://arrow.apache.org/docs/python/api/datatypes.html#factory-functions"
             )
-        arrow_data_type_str = str(type_str + "_")
+        arrow_data_factory_function_name = str(type_str + "_")
     else:
-        arrow_data_type_str = type_str
+        arrow_data_factory_function_name = type_str
 
-    return pa.__dict__[arrow_data_type_str]()
+    return pa.__dict__[arrow_data_factory_function_name]()
 
 
 def _cast_to_python_objects(obj: Any) -> Tuple[Any, bool]:
@@ -129,7 +172,7 @@ def cast_to_python_objects(obj: Any) -> Any:
 class Value:
     """Encapsulate an Arrow datatype for easy serialization."""
 
-    dtype: str
+    dtype: str  # The string representation of a pyarrow datatype: str(pyarrow.DataType)
     id: Optional[str] = None
     # Automatically constructed
     pa_type: ClassVar[Any] = None
@@ -207,7 +250,6 @@ class Array5D(_ArrayXD):
 
 
 class _ArrayXDExtensionType(pa.PyExtensionType):
-
     ndims: int = None
 
     def __init__(self, shape: tuple, dtype: str):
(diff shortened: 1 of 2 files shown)


src/datasets/features.py
```diff
@@ -33,18 +34,60 @@
...
+            return pa.timestamp(timestamp_internals)
+        elif internals_matches:
+            return pa.timestamp(internals_matches.group(1), internals_matches.group(2))
+        else:
+            raise ValueError(
+                f"{type_str} is not a validly formatted string representation of a pyarrow timestamp."
+                f"Examples include timestamp[us] or timestamp[us, tz=America/New_York]"
+                f"See: https://arrow.apache.org/docs/python/generated/pyarrow.timestamp.html#pyarrow.timestamp"
+            )
+    elif type_str == "double":
```
justin-yan, line 73: this handles the fact that `str(pa.float64()) == "double"` by inverting it, which is redundant with https://github.com/huggingface/datasets/pull/1900/files#diff-54081ede051fd0a7ef65748c481cc06f90209f01bb89968747089d13a2ca052bR182

src/datasets/features.py
```diff
@@ -129,7 +172,7 @@ def cast_to_python_objects(obj: Any) -> Any:
 class Value:
     """Encapsulate an Arrow datatype for easy serialization."""
 
-    dtype: str
+    dtype: str  # The string representation of a pyarrow datatype: str(pyarrow.DataType)
```
justin-yan, line 175: From the documentation here: https://huggingface.co/docs/datasets/features.html

I think it would be helpful to clarify whether this `dtype` should be the name of a pyarrow function, or if it should be the string representation of a pyarrow DataType:

i.e. `float64` or `double`?

My intuition is that if you want to accept `dtype`s such as `timestamp[ns]`, then this dtype value should correspond to `str(pa.DataType)` instead of `pyarrow.__dict__[<factory function name>]` - otherwise, you have one style of dtype for certain primitives, and another altogether for timestamps, decimal, etc.:

```
>>> str(pyarrow.decimal128(1))
'decimal(1, 0)'
```

I realize that double->float64 and float->float32 are already on the public interface so I wouldn't propose changing those since that would be a backward-incompatible change, but it would be useful for documenting going forward, I think.

Happy to accept a decision either way and to remove this comment, but wanted to highlight it so that the maintainers can provide explicit guidance.  Either way you choose, a comment and some additional documentation on the https://huggingface.co/docs/datasets/features.html page would probably be helpful, and I'd be happy to help with that.

src/datasets/features.py
```diff
@@ -696,7 +738,13 @@ class Sequence:
 
 
 def get_nested_type(schema: FeatureType) -> pa.DataType:
-    """ Convert our Feature nested object in an Apache Arrow type """
+    """
```
justin-yan, line 741: docstring I left while I was working locally - happy to remove or keep based on guidance from the maintainers!

src/datasets/features.py
```diff
@@ -759,6 +807,12 @@ def encode_nested_example(schema, obj):
 def generate_from_dict(obj: Any):
     """Regenerate the nested feature object from a serialized dict.
     We use the '_type' fields to get the dataclass name to load.
+
+    generate_from_dict is the recursive helper for Features.from_dict, and allows for a convenient constructor syntax
```
justin-yan, line 811: docstring I left while I was working locally - happy to remove or keep based on guidance from the maintainers!

src/datasets/features.py
```diff
@@ -775,7 +829,16 @@ def generate_from_dict(obj: Any):
     return class_type(**{k: v for k, v in obj.items() if k in field_names})
 
 
-def generate_from_arrow_type(pa_type: pa.DataType):
+def generate_from_arrow_type(pa_type: pa.DataType) -> FeatureType:
+    """
+    generate_from_arrow_type accepts an arrow DataType and returns a datasets FeatureType to be used as the type for
```
justin-yan, line 834: docstring I left while I was working locally - happy to remove or keep based on guidance from the maintainers!

src/datasets/features.py
```diff
@@ -33,18 +34,60 @@
...
+        >>> str(pa.timestamp('us'))
+        'timestamp[us]'
+        >>> str(pa.timestamp('us', tz='America/New_York'))
+        'timestamp[us, tz=America/New_York]'
+        """
+        timestamp_internals = timestamp_matches.group(1)
+        internals_regex = re.compile(r"^(s|ms|us|ns),\s*tz=([a-zA-Z0-9/_+:]*)$")
+        internals_matches = internals_regex.search(timestamp_internals)
+        if timestamp_internals in ["s", "ms", "us", "ns"]:
+            return pa.timestamp(timestamp_internals)
```
justin-yan, line 64: I opted for multiple returns, but if you don't like this style, I'm happy to refactor this into assigning a `return_datatype` value and having a single return at the end.

src/datasets/features.py
```diff
@@ -33,18 +34,60 @@
 
 
 def string_to_arrow(type_str: str) -> pa.DataType:
-    if type_str not in pa.__dict__:
+    """
+    string_to_arrow takes a string pyarrow dtype and converts it to a pyarrow.DataType.
+
+    In effect, `dt == string_to_arrow(str(dt))`
```
lhoestq, line 40: This is true only for types supported by Value. For instance decimal is not supported afaik

src/datasets/features.py
lhoestq: I'm fine with this, thanks :)

src/datasets/features.py
lhoestq: I'm ok for having this so that the `dt == string_to_arrow(str(dt))` rule holds.
Not that in practice since Value resolves "double" to "float64", then this case shouldn't happen to get the pyarrow type of a Value type.

lhoestq (commented):
Really cool thank you ! 

Timestamp handling looks good to me

I agree with you that we should explain in the documentation the link between the `datasets` feature types and the internal pyarrow types. I added a few comments.

Also thanks for the tests and the additional docstrings, this is helpful.

src/datasets/features.py
```diff
@@ -759,6 +807,12 @@ def encode_nested_example(schema, obj):
 def generate_from_dict(obj: Any):
     """Regenerate the nested feature object from a serialized dict.
     We use the '_type' fields to get the dataclass name to load.
+
+    generate_from_dict is the recursive helper for Features.from_dict, and allows for a convenient constructor syntax
+        for users to define new datasets and provide their own Features.  This acts as an analogue to
```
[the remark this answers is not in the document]
justin-yan: Interesting, I didn't realize this.  When users provide their own custom Features to override automatic inference, doesn't it go through Features.from_dict() which goes through this?

src/datasets/features.py
justin-yan: If you want to maintain your own dtypes separate from pyarrow, then I think that makes a lot of sense.

In that world, perhaps there should actually be a separate function `arrow_to_string` that would act as an inverse for `string_to_arrow` that you could specifically control vs. relying on whatever `str(pa_type)` happens to do?

I'm happy to clean up my PR with this in mind, and I can open a follow-up PR that will incorporate this suggestion and add some documentation if you'd like!

justin-yan:
OK! Thank you for the review - I will follow up with a separate PR for the comments here (https://github.com/huggingface/datasets/pull/1900#discussion_r578319725)!

src/datasets/features.py
lhoestq: It makes sense to have `arrow_to_string` indeed, thanks for the suggestion and for the help :)
And sure feel free to add some documentation, this is greatly appreciated

We try to have our own dtypes so that we can provide alternative backends instead of pyarrow, which can be a bit overkill for data that easily fit in memory IMO. This is something we're exploring

src/datasets/features.py
```diff
@@ -33,18 +34,53 @@
 
 
 def string_to_arrow(type_str: str) -> pa.DataType:
-    if type_str not in pa.__dict__:
+    """
+    string_to_arrow takes a datasets string dtype and converts it to a pyarrow.DataType.
+
+    In effect, `dt == string_to_arrow(arrow_to_datasets_dtype(dt))`
```
lhoestq, line 40:
```suggestion
    In effect, `dt == string_to_arrow(str(dt))`
```
`arrow_to_datasets_dtype` doesn't exist yet ^^

lhoestq (approved):
Thanks a lot !
Let me apply the minor change to the docstring and remove the mention of `arrow_to_datasets_dtype` for now since it doesn't exist yet :)

PR #1904, opened 2021-02-18
Fix to_pandas for boolean ArrayXD

lhoestq:
As noticed in #1887 the conversion of a dataset with a boolean ArrayXD feature types fails because of the underlying ListArray conversion to numpy requires `zero_copy_only=False`.

zero copy is available for all primitive types except booleans
see https://arrow.apache.org/docs/python/generated/pyarrow.Array.html#pyarrow.Array.to_numpy
and https://issues.apache.org/jira/browse/ARROW-2871?jql=text%20~%20%22boolean%20to_numpy%22

cc @SBrandeis

### DIFF
--- a/src/datasets/features.py
+++ b/src/datasets/features.py
@@ -265,12 +265,16 @@ class ArrayExtensionArray(pa.ExtensionArray):
         return self.storage[i]
 
     def to_numpy(self):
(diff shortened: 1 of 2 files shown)


SBrandeis:
Thanks!

PR #1905, opened 2021-02-18
Standardizing datasets.dtypes

justin-yan:
This PR was further branched off of jdy-str-to-pyarrow-parsing, so it depends on https://github.com/huggingface/datasets/pull/1900 going first for the diff to be up-to-date (I'm not sure if there's a way for me to use jdy-str-to-pyarrow-parsing as a base branch while having it appear in the pull requests here).

This moves away from `str(pyarrow.DataType)` as the method of choice for creating dtypes, favoring an explicit mapping to a list of supported Value dtypes.

I believe in practice this should be backward compatible, since anyone previously using Value() would only have been able to use dtypes that had an identically named pyarrow factory function, which are all explicitly supported here.

### DIFF
--- a/src/datasets/features.py
+++ b/src/datasets/features.py
@@ -15,6 +15,7 @@
 
 # Lint as: python3
 """ This class handle features definition in datasets and some utilities to display table type."""
+import re
 from collections.abc import Iterable
 from dataclasses import dataclass, field, fields
 from typing import Any, ClassVar, Dict, List, Optional, Sequence, Tuple, Union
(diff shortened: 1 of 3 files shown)


justin-yan:
Also - I took a stab at updating the docs, but I'm not sure how to actually check the outputs to see if it's formatted properly.

src/datasets/features.py
```diff
@@ -32,19 +35,154 @@
...
+        arrow_data_factory_function_name = str(datasets_dtype + "_")
     else:
-        arrow_data_type_str = type_str
+        arrow_data_factory_function_name = datasets_dtype
+
+    return pa.__dict__[arrow_data_factory_function_name]()
+
+
+@dataclass
+class Value:
```
justin-yan, line 136: There is no behavior change, I just added the docstring below and moved the class up the file to be near the two conversion functions above!

Issue #870, opened 2020-11-19
[Feature Request] Add optional parameter in text loading script to preserve linebreaks


jncasey:
I'm working on a project about rhyming verse using phonetic poetry and song lyrics, and line breaks are a vital part of the data. 

I recently switched over to use the datasets library when my various corpora grew larger than my computer's memory. And so far, it is SO great. 

But the first time I processed all of my data into a dataset, I hadn't realized the text loader script was processing the source files line-by-line and stripping off the newlines. 

Once I caught the issue, I made my own data loader by modifying one line in the default text loader (changing `batch = batch.splitlines()` to `batch = batch.splitlines(True)` inside `_generate_tables`). And so I'm all set as far as my project is concerned.

But if my use case is more general, it seems like it'd be pretty trivial to add a kwarg to the default text loader called keeplinebreaks or something, which would default to False and get passed to `splitlines()`.

lhoestq:
Hi ! Thanks for your message.
Indeed it's a free feature we can add and that can be useful.
If you want to contribute, feel free to open a PR to add it to the text dataset script :)

mariosasko:
Resolved via #1913.

PR #1913, opened 2021-02-19
Add keep_linebreaks parameter to text loader

lhoestq:
As asked in #870 and https://github.com/huggingface/transformers/issues/10269 there should be a parameter to keep the linebreaks when loading a text dataset.
cc @sgugger @jncasey


sgugger:
Just so I understand how it can be used in practice, do you have an example showing how to load a text dataset with this option?

lhoestq:
Sure ! Here is an example:
```python

load_dataset("text", keep_linebreaks=True, data_files=...)
```

I'll update the documentation to explain this

sgugger:
Perfect!

PR #1914, opened 2021-02-19
Fix logging imports and make all datasets use library logger

albertvillanova:
Fix library relative logging imports and make all datasets use library logger.


lhoestq (approved):
Thanks !

Also note that the logger names are pretty long.
For example
```
datasets_modules.datasets.turkish_product_reviews.5d8320105c3f5236ec90844fe92c9532f4f43574d14c746adb2cf04150fbfd65.turkish_product_reviews
```
Maybe we can crop it to
```
datasets_modules.datasets.turkish_product_reviews
```
but if someone uses several versions of the same dataset then logs would collide.
Let's keep it this way with the long logger names for now.

src/datasets/utils/py_utils.py
```diff
@@ -146,21 +147,21 @@ def __get__(self, obj, objtype=None):
...
-            return self
-        if self.fget is None:
-            raise AttributeError("unreadable attribute")
-        attr = "__cached_" + self.fget.__name__
-        cached = getattr(obj, attr, None)
-        if cached is None:
-            cached = self.fget(obj)
-            setattr(obj, attr, cached)
-        return cached
+# class memoized_property(property):  # pylint: disable=invalid-name
```
lhoestq, line 150: I just noticed that you also commented these.
Can you uncomment them ?
We can discuss removing them in another PR

albertvillanova: Done @lhoestq

lhoestq: Thanks !

PR #1918, opened 2021-02-20
Fix QA4MRE download URLs

M-Salti:
The URLs in the `dataset_infos` and `README` are correct, only the ones in the download script needed updating.


lhoestq (approved):
Good catch thanks !
This is probably an artefact from the renaming of the library of `nlp` to `datasets`. Maybe a bad regex did this x)

PR #1920, opened 2021-02-20
Fix save_to_disk issue

### LINKED ISSUE #1919: Failure to save with save_to_disk
When I try to save a dataset locally using the `save_to_disk` method I get the error:

```bash
FileNotFoundError: [Errno 2] No such file or directory: '/content/squad/train/squad-train.arrow'
```

To replicate:

1. Install `datasets` from master
2. Run this code:

    ```python
    squad = load_dataset("squad")   # or any other dataset
    squad.save_to_disk("squad")     # error here
    ```

The problem is that the method is not creating a directory with the name `dataset_path` for saving the dataset in (i.e. it's not creating the *train* and *validation* directories in this case). After creating the directory the problem resolves.
I'll open a PR soon doing that and linking this issue.

M-Salti:
Fixes #1919

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -468,6 +468,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
         with tempfile.TemporaryDirectory() as tmp_dir:
             temp_dataset_path = Path(tmp_dir).joinpath(dataset_path)
             os.makedirs(temp_dataset_path, exist_ok=True)
+            os.makedirs(dataset_path, exist_ok=True)
 
             # Write indices if needed
             if self._indices is not None:


M-Salti:
So I was curious why the issue reported at #1919 wasn't caught in [this test](https://github.com/huggingface/datasets/blob/248104c4bdb2e01c036b7578867199191fbff181/tests/test_arrow_dataset.py#L209), so I did some digging.
I tried to save to a temporary directory (just like in the test), like this:
```python
with tempfile.TemporaryDirectory() as requested_tempdir:
    squad.save_to_disk(requested_tempdir)     # no error
```
and it executes succesfuly without problems.
So why does it work, but this doesn't?
```python
squad.save_to_disk("./squad")   # error
```
It's because `save_to_disk` also creates a temporary directory (let's call it `tempdir`), and since `tempdir` and `requested_tempdir` share the same parents, the `Path.joinpath` method [(here)](https://github.com/huggingface/datasets/blob/248104c4bdb2e01c036b7578867199191fbff181/src/datasets/arrow_dataset.py#L469) will keep `requested_tempdir` as it is and the *train* directory will be created under `requested_tempdir` and hence no errors will arise.

But in the second case (where we are saving to a local dir), the *train* directory is created under *squad* which in turn is created under `tempdir`, not under `.` (current dir).

So, all of this probably doesn't help solving the issue but it might help creating a better test, and it also makes me wonder why are we saving to a temporary dir in `save_to_disk` anyway? I mean, won't it be removed with all its contents upon execution completion? what's the point then?

src/datasets/arrow_dataset.py
```diff
@@ -468,6 +468,7 @@ def save_to_disk(self, dataset_path: str, fs=None):
         with tempfile.TemporaryDirectory() as tmp_dir:
             temp_dataset_path = Path(tmp_dir).joinpath(dataset_path)
             os.makedirs(temp_dataset_path, exist_ok=True)
+            os.makedirs(dataset_path, exist_ok=True)
```
lhoestq, line 471:
```suggestion
            fs.makedirs(dataset_path, exist_ok=True)
```
dataset_path could be a remote filesystem

lhoestq (commented):
Hi ! Thanks for reporting the issue.

Indeed the issue comes from the line you mentioned. `dataset_path` shouldn't be used to define the temporary directory.
This used to hide the actual issue that the 

The temporary directory is used as an intermediary step to write on disk the Dataset data that are in memory (self._data or self._indices). Since save_to_disk must support writing to a remote filesystem and that the ArrowWriter only supports writing locally (for now), we first write locally in a temp dir before moving the file to the destination location.

I opened #1923 to apply your fix (but using the `fs.makedirs` for remote filesystems) and remove the bad `joinpath`. I added a test for relative directories as well.

Thanks again for reporting and investigating ! This is really helpful

lhoestq:
CLosing in favor of #1923

PR #1921, opened 2021-02-20
Standardizing datasets dtypes

justin-yan:
This PR follows up on discussion in #1900 to have an explicit set of basic dtypes for datasets.

This moves away from str(pyarrow.DataType) as the method of choice for creating dtypes, favoring an explicit mapping to a list of supported Value dtypes.

I believe in practice this should be backward compatible, since anyone previously using Value() would only have been able to use dtypes that had an identically named pyarrow factory function, which are all explicitly supported here, with `float32` and `float64` acting as the official datasets dtypes, which resolves the tension between `double` being the pyarrow dtype and `float64` being the pyarrow type factory function.

### DIFF
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -1,12 +1,37 @@
 Dataset features
-===========================
+================
 
-:class:`datasets.Features` define the internal structure and typings for each example in the dataset. Features are used to specify the underlying serailization format but also contain high-level informations regarding the fields, e.g. conversion methods from names to integer values for a class label field.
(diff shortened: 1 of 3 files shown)


justin-yan:
@lhoestq - apologies for the multiple PRs, my previous one (#1905) got mangled due to some merge conflicts that I had trouble resolving so I just cherry-picked my changes onto a fresh branch here.

lhoestq (approved):
Nice thank you !

PR #1929, opened 2021-02-22
Improve typing and style and fix some inconsistencies

mariosasko:
This PR:
* improves typing (mostly more consistent use of `typing.Optional`)
* `DatasetDict.cleanup_cache_files` now correctly returns a dict 
* replaces `dict()` with the corresponding literal
* uses `dict_to_copy.copy()` instead of `dict(dict_to_copy)` for shallow copying


src/datasets/features.py
```diff
@@ -471,15 +473,15 @@ class ClassLabel:
             classes, one per line.
     """
 
-    num_classes: int = None
-    names: List[str] = None
-    names_file: str = None
+    num_classes: Optional[int] = None
+    names: Optional[List[str]] = None
+    names_file: Optional[str] = None
```
lhoestq, line 478: Currently it's true that these can stay None.
We should disallow that though by doing some changes in the post_init.
This is out of scope for this PR though, so let's keep the Optional for now

src/datasets/features.py
```diff
@@ -471,15 +473,15 @@ class ClassLabel:
...
     id: Optional[str] = None
     # Automatically constructed
     dtype: ClassVar[str] = "int64"
     pa_type: ClassVar[Any] = pa.int64()
-    _str2int: ClassVar[Dict[str, int]] = None
-    _int2str: ClassVar[Dict[int, int]] = None
+    _str2int: ClassVar[Optional[Dict[str, int]]] = None
+    _int2str: ClassVar[Optional[Dict[int, int]]] = None
```
lhoestq, line 484: same here but out of scope for this PR

src/datasets/features.py
```diff
@@ -667,8 +670,8 @@ class TranslationVariableLanguages:
         }
     """
 
-    languages: List = None
-    num_languages: int = None
+    languages: Optional[List] = None
+    num_languages: Optional[int] = None
```
lhoestq, line 674: same here but out of scope for this PR

src/datasets/utils/version.py
```diff
@@ -35,10 +36,10 @@ class Version:
...
 
     version_str: str
-    description: str = None
-    major: str = None
-    minor: str = None
-    patch: str = None
+    description: Optional[str] = None
+    major: Optional[str] = None
+    minor: Optional[str] = None
+    patch: Optional[str] = None
```
lhoestq, line 42:
```suggestion
    major: str = None
    minor: str = None
    patch: str = None
```
These three fields are strings. They're set during post_init

lhoestq (commented):
Nice thank you.
This looks all good to me.
I just have one concern about the Version fields major, minor and patch: they're actually always strings so no need to make them optional

src/datasets/utils/version.py
mariosasko: Checked the souce and it seems these three fields are ints (post_init calls _str_to_version which returns the 3-tuple of ints; the allow_wildcard param of _str_to_version is False by default so '*' is not allowed). I think the correct typing would look like this:

```python
major: int = field(init=False)
minor: int = field(init=False)
patch: int = field(init=False)
```

field(init=False) excludes them from the `__init__` (only version_str is required due to the steps performed in the post_init that are described above).

Another improvement would be to annotate the version_str field with `InitVar[str]`. This would make post_init's implementation cleaner:
```python
def __post_init__(self, version_str):
    self.major, self.minor, self.patch = _str_to_version(version_str) # no longer requires self.version_str
``` 
but with this change dataclasses.fields(cls) can no longer be used in Version.from_dict since it doesn't include info about the class InitVars. So implementation of Version.from_dict would be as follows:
```python
@classmethod
def from_dict(cls, dic):
    field_names = set(f.name for f in cls.__dataclass_fields__.values() if f.init)
    return cls(**{k: v for k, v in dic.items() if k in field_names})
```

mariosasko:
@lhoestq Thanks for the quick review.

src/datasets/utils/version.py
lhoestq: Thanks for the details.
Let's keep it this way and discuss this in another PR

lhoestq:
I merged master to this branch to re-run the CI before merging :)

lhoestq (approved):
Thanks :)

PR #1938, opened 2021-02-24
Disallow ClassLabel with no names

lhoestq:
It was possible to create a ClassLabel without specifying the names or the number of classes.
This was causing silent issues as in #1936 and breaking the conversion methods str2int and int2str.

cc @justin-yan

### DIFF
--- a/src/datasets/features.py
+++ b/src/datasets/features.py
@@ -549,52 +549,41 @@ class ClassLabel:
             classes, one per line.
     """
 
-    num_classes: Optional[int] = None
-    names: Optional[List[str]] = None
+    num_classes: int = None
+    names: List[str] = None
     names_file: Optional[str] = None
     id: Optional[str] = None
     # Automatically constructed
     dtype: ClassVar[str] = "int64"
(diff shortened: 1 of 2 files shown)


src/datasets/features.py
```diff
@@ -549,53 +549,42 @@ class ClassLabel:
...
-            if self.names is None:
+            elif self.num_classes is not None:
                 self.names = [str(i) for i in range(self.num_classes)]
-            elif len(self.names) != self.num_classes:
-                raise ValueError(
-                    "ClassLabel number of names do not match the defined num_classes. "
-                    "Got {} names VS {} num_classes".format(len(self.names), self.num_classes)
-                )
-
+            else:
```
justin-yan, line 572: wonder if it might be easier to do something like `assert any([num_classes, names, names_file])` at the top of the method to enforce the check that one of the three should exist.  (I guess it would have to be refactored a bit to ensure that it's exactly one of the three(?))

lhoestq: I think it's fine this way
Plus it allows to have both names and num_classes as long as num_classes == len(names)

PR #1943, opened 2021-02-25
Implement Dataset from JSON and JSON Lines

albertvillanova:
Implement `Dataset.from_jsonl`.


lhoestq (commented):
You can use the Json dataset builder instead (from `datasets.packaged_modules.json.json`)

This dataset builder supports both json and jsonl. This is what's used when you do `load_dataset("json", data_files=...)` for example.

Moreover since it's a dataset builder, the pyarrow table can be written in memory or on disk to use memory mapping.

albertvillanova:
Thanks @lhoestq. I was trying to follow @thomwolf suggestion about integrating that script but as `from_json` method...
> Note that I don't think this is necessary a breaking change, we can still keep the old scripts around

Do you think there is a better way of doing it?

I was trying to implement more or less the same logic as in the script, but I confess I assumed the target was in-memory only...

albertvillanova:
Basically, I was trying to reimplement `Json(datasets.ArrowBasedBuilder)._generate_tables`, and no writing to arrow file (I assumed only in-memory usage). I started with the first "else" clause...

albertvillanova:
@lhoestq I am wondering why `keep_in_memory` has no effect for JSON...

lhoestq:
What's the issue exactly ? Apparently it's correctly passed to as_dataset so I don't find the issue

src/datasets/packaged_modules/json/json.py
```diff
@@ -80,8 +80,11 @@ def _generate_tables(self, files):
                     pa_table = paj.read_json(
                         file,
                         read_options=self.config.pa_read_options,
-                        parse_options=self.config.pa_parse_options,
+                        # parse_options=self.config.pa_parse_options,
```
lhoestq, line 83: why not use the parse options ?

lhoestq: To avoid breaking changes I'd still keep the parse options but only for the `newlines_in_values` parameter, since the schema is now used for casting the table to allow int<->str conversions.

albertvillanova: That is what I finally did!!! ;)

lhoestq:
merging master into this branch should fix the CI issue :)

src/datasets/arrow_dataset.py
```diff
@@ -41,6 +41,8 @@
 from . import config
 from .arrow_reader import ArrowReader
 from .arrow_writer import ArrowWriter, TypedSequence
+
+# from .dataset_reader import JsonlDatasetReader
```
lhoestq, line 45:
```suggestion
```

src/datasets/arrow_dataset.py
```diff
@@ -461,6 +463,33 @@ def from_csv(
...
+    @staticmethod
+    def from_json(
+        path_or_paths: Union[PathLike, List[PathLike]],
+        split: Optional[NamedSplit] = None,
+        features: Optional[Features] = None,
+        cache_dir: str = None,
+        keep_in_memory: bool = False,
+        **kwargs,
+    ):
+        """Create Dataset from JSON Lines file(s).
```
lhoestq, line 475: It works for jsonl files but also for json !
You just need to add the `field` parameter to `from_json` and `JsonDatasetReader` pass it to the `Json` dataset builder at initialization.

It allows to load squad-like json files for example.
The json looks like this:
```
{
    "data": [...],  # the actual data are in the "data" 
    ...
}
```

lhoestq (commented):
Nice :) !
Can you also add the `field` parameter for regular json files if yo don't mind ?

src/datasets/arrow_dataset.py
albertvillanova: The `field` parameter was implicitly passed in `kwargs`, but I can make it explicit. Also in the docstring ;)

What about the other JsonConfig parameters: `use_threads`, `block_size`, `newlines_in_values`?

lhoestq: Yes right ! I didn't noticed that it was actually passed with the kwargs.
I think it's better to have it as a named argument directly since it's a very important one.

For the other parameters I think it's fine to have them in the kwargs

src/datasets/arrow_dataset.py
```diff
@@ -461,6 +461,41 @@ def from_csv(
...
+    def from_json(
+        path_or_paths: Union[PathLike, List[PathLike]],
+        split: Optional[NamedSplit] = None,
+        features: Optional[Features] = None,
+        cache_dir: str = None,
+        keep_in_memory: bool = False,
+        field: Optional[str] = None,
+        **kwargs,
+    ):
+        """Create Dataset from JSON (Lines) file(s).
```
lhoestq, line 474:
```suggestion
        """Create Dataset from JSON or JSON Lines file(s).
```
(nit) this is clearer this way imo

lhoestq (commented):
Thanks for adding the field to the arguments :)

Finally, can you just edit this part of the error message in src/datasets/packaged_modules/json/json.py from
```
Select the correct one and provide it as `field='XXX'` to the `load_dataset` method.
```
to
```
Select the correct one and provide it as `field='XXX'` to the dataset loading method.
```
This error is raised when you forget to set `field=`.

Previously it was an indication that the user needs to set `field=` in `load_dataset` but now it could also be from `from_csv`. Therefore a more general error message as I suggested above is welcome.

(I wanted to edit the file directly in the PR but I couldn't for some reason)

lhoestq:
Good job ! I think we can merge after the last changes regarding the error message and the docstring above :)

src/datasets/arrow_dataset.py
albertvillanova: Done

albertvillanova:
@lhoestq Done! And I have also added some tests for the `field` parameter.

albertvillanova:
Let me add some more tests for dict of lists JSON file, please.

albertvillanova:
@lhoestq done! ;)

tests/test_arrow_dataset.py
```diff
@@ -1933,6 +1933,62 @@ def test_dataset_from_file(in_memory, dataset, arrow_file):
...
+    [
+        None,
+        {"col_1": "string", "col_2": "int64", "col_3": "float64"},
+        {"col_1": "string", "col_2": "string", "col_3": "string"},
+        {"col_1": "int32", "col_2": "int32", "col_3": "int32"},
+        {"col_1": "float32", "col_2": "float32", "col_3": "float32"},
+    ],
+)
+@pytest.mark.parametrize("split", [None, NamedSplit("train"), "train", "test"])
+@pytest.mark.parametrize("path_type", [str, list])
```
lhoestq, line 1949: All these combinations create 100+ tests. I think it added ~5min of time for the CI to run.
Could we make this faster ?

albertvillanova: I am trying to find out why they are so slow...

albertvillanova: Tests per-se are not slow. But I think there is some kind of race condition due to `utils\filelock.py::acquire`:
```python
time.sleep(poll_intervall)
```

albertvillanova: Not a big performance gain after my last commits... :(

albertvillanova: I have finally removed tests.

lhoestq: That did the job to reduce testing time, thanks !

Maybe we can unit-test the json builder for the supported format in test_packaged_modules.py (as for the csv builder) ? (in another PR)

lhoestq (approved):
Looks all good !

Do you have more changes to do ? Otherwise we can merge :)

albertvillanova:
We can merge. Additional work will be done in another PR. ;)

PR #1944, opened 2021-02-25
Add Turkish News Category Dataset (270K - Lite Version)

yavuzKomecoglu:
This PR adds the Turkish News Categories Dataset (270K - Lite Version) dataset which is a text classification dataset by me,  @basakbuluz and @serdarakyol. 
This dataset contains the same news from the current [interpress_news_category_tr dataset](https://huggingface.co/datasets/interpress_news_category_tr) but contains less information, OCR errors are reduced, can be easily separated, and can be divided into 10 classes ("kültürsanat", "ekonomi", "siyaset", "eğitim", "dünya", "spor", "teknoloji", "magazin", "sağlık", "gündem") were rearranged.

@SBrandeis @lhoestq, can you please review this PR?


datasets/interpress_news_category_tr_lite/interpress_news_category_tr_lite.py
```diff
@@ -0,0 +1,100 @@
...
+        ),
+    ]
+
+    def _info(self):
+        return datasets.DatasetInfo(
+            description=_DESCRIPTION,
+            features=datasets.Features(
+                {
+                    "content": datasets.Value("string"),
+                    "category": datasets.features.ClassLabel(num_classes=10),
```
lhoestq, line 72: Can you provide the class names here please ?
("kültürsanat", "ekonomi", "siyaset", "eğitim", "dünya", "spor", "teknoloji", "magazin", "sağlık", "gündem")

They must be in the correct order so that the label id 0 corresponds to the first name of this list, the label id 1 corresponds to the second name etc.

You'll have to regenerate the dataset_infos.json after this change

datasets/interpress_news_category_tr_lite/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Languages
+
+The dataset is based on Turkish.
+
+## Dataset Structure
+
+### Data Instances
+
+A text classification dataset with 10 different news category.
```
lhoestq, line 64: Can you add one example from the dataset here please ?

You can get one by doing `dataset = load_dataset("./datasets/interpress_news_category_tr_lite", split="train")` and taking the first sample `dataset[0]`

datasets/interpress_news_category_tr_lite/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Dataset Structure
+
+### Data Instances
+
+A text classification dataset with 10 different news category. 
+
+### Data Fields
+
+[More Information Needed]
```
lhoestq, line 68: Can you add a brief description of each field `content` and `category` here please ?

lhoestq (commented):
Nice ! Thanks for adding this one :)

added a few comments:

datasets/interpress_news_category_tr_lite/interpress_news_category_tr_lite.py
serdarakyol: The values of the label/category field in the dataset are type of int, not string. so I can't edit it as you wanted.

The error occur in "output.append(self._str2int[str(value)])"

lhoestq: This error means that the value passed in `yield` are strings, not int
Can you try to cast the label to int ?
```
"category": int(row["label"])
```

yavuzKomecoglu:
I updated your suggestions. Thank you very much for your support. @lhoestq

datasets/interpress_news_category_tr_lite/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+The dataset is based on Turkish.
+
+## Dataset Structure
+
+### Data Instances
+
+A text classification dataset with 10 different news category. 
+
+<<<<<<< HEAD
```
lhoestq, line 66: Looks like there are some git conflicts to resolve

lhoestq (commented):
Thanks for changing to ClassLabel :)
This is all good now !

However I can see changes in other files than the ones for interpress_news_category_tr_lite, can you please fix that ?
To do so you can create another branch and another PR to only include the interpress_news_category_tr_lite files.

Maybe this happened because of a git rebase ? Once you've already pushed your code, please use git merge instead of rebase in order to avoid this.

yavuzKomecoglu:
> Thanks for changing to ClassLabel :)
> This is all good now !
> 
> However I can see changes in other files than the ones for interpress_news_category_tr_lite, can you please fix that ?
> To do so you can create another branch and another PR to only include the interpress_news_category_tr_lite files.
> 
> Maybe this happened because of a git rebase ? Once you've already pushed your code, please use git merge instead of rebase in order to avoid this.

Thanks for the feedback.
New PR https://github.com/huggingface/datasets/pull/1967

PR #1946, opened 2021-02-25
Implement Dataset from CSV

albertvillanova:
Implement `Dataset.from_csv`.

Analogue to #1943.

If finally, the scripts should be used instead, at least we can reuse the tests here.


src/datasets/io/csv.py
```diff
@@ -0,0 +1,77 @@
...
+        table = pa.Table.from_pandas(df, schema=schema)
+        return table
+
+
+class CsvDatasetBuilder:
+    def __init__(
+        self,
+        path,
+        name=None,
+        data_dir=None,
```
albertvillanova, line 27: @lhoestq do we need `data_dir` for `Dataset.from_csv()`?

src/datasets/io/abc.py
```diff
@@ -0,0 +1,26 @@
+from abc import ABC, abstractmethod
+from typing import Optional
+
+from .. import Features, NamedSplit
+from ..arrow_dataset import Dataset
+from ..utils.typing import PathLike
+
+
+class AbstractDatasetBuilder(ABC):
```
lhoestq, line 9: We've already a DatasetBuilder abstraction.
Maybe we can find another name ?

Or maybe we can simply add `build` to the existing DatasetBuilder

albertvillanova: Yes, @lhoestq, I totally agree with you. I was thinking of this since I pushed the change: it is confusing to call it Builder, because we have already a builder classes. I am going to rename it back to `DatasetReader.read`.

I prefer having a specific class to deal with `from_csv` method: it is basically an adapter with a single method, which hides all the complexity underneath. I would prefer not adding another method to DatasetBuilder: my long-term objective is in the opposite direction, by reducing that class (which is already too large) by composing it of specialized classes...

Thanks again for your critical view!! :heart:

lhoestq: Makes sense !
I agree that DatasetBuilder is already quite big, let's have a DatasetReader then :)

albertvillanova: Done @lhoestq

src/datasets/io/csv.py
```diff
@@ -0,0 +1,53 @@
...
+        cache_dir: str = None,
+        keep_in_memory: bool = False,
+        **kwargs,
+    ):
+        super().__init__(
+            path, split=split, features=features, cache_dir=cache_dir, keep_in_memory=keep_in_memory, **kwargs
+        )
+        self.builder = Csv(
+            cache_dir=cache_dir,
+            data_files=path,
```
lhoestq, line 24: When data_files is just one path, then the builder generates one dataset that is considered to be the train split.
In order to use the `split` parameter of the Reader, you can change this by using a dict, like this:
```python
data_files={split: path}
```
The split defined in `data_files` are the ones you can get when using `as_dataset`

You may have to set a default value for `split` though (the builder does require a valid split name, not None):
```python
split = split if split is not None else "train"
```

lhoestq (commented):
Awesome :)

src/datasets/io/csv.py
albertvillanova: Indeed `path` can be one path, or a dict or a list... as it is the case for `load_dataset`...

albertvillanova: However, this behavior is different from the one in `load_dataset`
> split (`datasets.Split` or `str`): which split of the data to load.
> - If None, will return a `dict` with all splits (typically `datasets.Split.TRAIN` and `datasets.Split.TEST`).
> - If given, will return a single Dataset.

lhoestq: It can make sense if Dataset.from_csv always returns a Dataset, not a DatasetDict with several splits.
If users want to have several splits maybe we can have DatasetDict.from_csv

Or we can allow a dictionary of paths but in this case it's better if it's consistent with load_dataset

albertvillanova: I see the problem... Yes, it makes sense that Dataset.from_csv should return a Dataset (and not a DatasetDict)... 

I guess we must make a decision with this trade-off between consistency and user-experience: is the user responsible for using Dataset/DatasetDict .from_csv or does the user expect (and does not mind) either one of them and the user?

lhoestq: I have a small preference for having both Dataset/DatasetDict.from_csv since it's simple and clear.
Moreover the `split` argument of load_dataset is different from the one used in from_xxx:
- The split argument in load_dataset is used to select the splits you want among the available splits. Also when loading a csv with a single file as you did, only a train split is available by default.
- The split argument in Dataset.from_xxx is used to set the split name of the resulting Dataset object

Because of that I'm more in favor of implementing Dataset/DatasetDict.from_csv separately and have a different behavior than load_dataset.

src/datasets/io/csv.py
```diff
@@ -0,0 +1,53 @@
...
+            base_path=base_path,
+            use_auth_token=use_auth_token,
+        )
+
+        # Build dataset for splits
+        ds = self.builder.as_dataset(
+            split=self.split, ignore_verifications=ignore_verifications, in_memory=self.keep_in_memory
+        )
+        if save_infos:
+            self.builder._save_infos()
```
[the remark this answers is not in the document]
albertvillanova: Done.

src/datasets/arrow_dataset.py
```diff
@@ -435,6 +434,32 @@ def from_dict(
...
+        keep_in_memory: bool = False,
+        **kwargs,
+    ):
+        """Create Dataset from CSV file(s).
+        Args:
+            path_or_paths (path-like or list of path-like): Path(s) of the CSV file(s).
+            split (NamedSplit, optional): Split name to be assigned to the dataset.
+            features (Features, optional): Dataset features.
+            cache_dir (str, optional, default="~/datasets"): Directory to cache data.
+            keep_in_memory (bool, default=False): Whether to copy the data in-memory.
```
lhoestq, line 452: Could you mention that the kwargs are passed to pandas.read_csv ?

lhoestq (commented):
Nice ! 🔥 

I think the only thing left to do is just add tests for Dataset.from_csv and DatasetDict.from_csv in test_arrrow_dataset.py and test_dataset_dict.py
Good job !

src/datasets/arrow_dataset.py
albertvillanova: Sure!

src/datasets/io/csv.py
albertvillanova: Done!

albertvillanova:
@lhoestq question about public API: `keep_in_memory` or just `in_memory`?

lhoestq:
For consistence I'd say `keep_in_memory`, but no strong opinion.

lhoestq (approved):
Looks all good ! Thanks for adding this 🎉

Feel free to also add `from_csv` to the list of documented methods in the documentation in `main_classes.rst`.

PR #1951, opened 2021-02-26
Add cross-platform support for datasets-cli

mariosasko:
One thing I've noticed while going through the codebase is the usage of `scripts` in `setup.py`. This [answer](https://stackoverflow.com/a/28119736/14095927) on SO explains it nicely why it's better to use `entry_points` instead of `scripts`. To add cross-platform support to the CLI, this PR replaces `scripts` with `entry_points` in `setup.py` and moves datasets-cli to src/datasets/commands/datasets_cli.py. All *.md and *.rst files are updated accordingly. The same changes were made in the transformers repo to add cross-platform ([link to PR](https://github.com/huggingface/transformers/pull/4131)).

### DIFF
--- a/ADD_NEW_DATASET.md
+++ b/ADD_NEW_DATASET.md
@@ -144,12 +144,12 @@ Sometimes you need to use several *configurations* and/or *splits* (usually at l
 **Last step:** To check that your dataset works correctly and to create its `dataset_infos.json` file run the command:
 
 ```bash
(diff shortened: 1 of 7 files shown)


lhoestq (approved):
That's good to know :)

Thanks for fixing this and for updating the docs !

gchhablani:
@mariosasko This is kinda cool!

PR #1960, opened 2021-02-28
Allow stateful function in dataset.map

### LINKED ISSUE #1940: Side effect when filtering data due to `does_function_return_dict` call in `Dataset.map()`
Hi there!

In my codebase I have a function to filter rows in a dataset, selecting only a certain number of examples per class. The function passes a extra argument to maintain a counter of the number of dataset rows/examples already selected per each class, which are the ones I want to keep in the end:

```python
      def fill_train_examples_per_class(example, per_class_limit: int, counter: collections.Counter):
          label = int(example['label'])
          current_counter = counter.get(label, 0)
          if current_counter < per_class_limit:
              counter[label] = current_counter + 1
              return True
          return False
```

At some point I invoke it through the `Dataset.filter()` method in the `arrow_dataset.py` module like this:

```python
...
kwargs = {"per_class_limit": train_examples_per_class_limit, "counter": Counter()}
datasets['train'] = datasets['train'].filter(fill_train_examples_per_class,  num_proc=1, fn_kwargs=kwargs)
...
```

The problem is that, passing a stateful container (the counter,) provokes a side effect in the new filtered dataset obtained. This is due to the fact that at some point in `filter()`, the `map()`'s function `does_function_return_dict` is invoked in  line [1290](https://github.com/huggingface/datasets/blob/96578adface7e4bc1f3e8bafbac920d72ca1ca60/src/datasets/arrow_dataset.py#L1290). 

When this occurs, the state of the counter is initially modified by the effects of the function call on the 1 or 2 rows selected in lines 1288 and 1289 of the same file (which are marked as `test_inputs` & `test_indices` respectively in lines 1288 and 1289. This happens out of the control of the user (which for example can't reset the state of the counter before continuing the execution,) provoking in the end an undesired side effect in the results obtained. 

In my case, the resulting dataset -despite of the counter results are ok- lacks an instance of the classes 0 and 1 (which happen to be the classes of the first two examples of my dataset.) The rest of the classes I have in my dataset, contain the right number of examples as they were not affected by the effects of `does_function_return_dict` call.

I've debugged my code extensively and made a workaround myself hardcoding the necessary stuff (basically putting `update_data=True` in line 1290,) and then I obtain the results I expected without the side effect.

Is there a way to avoid that call to `does_function_return_dict` in map()'s line 1290 ? (e.g. extracting the required information that `does_function_return_dict` returns without making the testing calls to the user function on dataset rows 0 & 1) 

Thanks in advance,

Francisco Perez-Sorrosal

mariosasko:
Removes the "test type" section in Dataset.map which would modify the state of the stateful function. Now, the return type of the map function is inferred after processing the first example.

Fixes #1940 

@lhoestq Not very happy with the usage of `nonlocal`. Would like to hear your opinion on this.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -1474,18 +1474,12 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
         # We set this variable in `apply_function_on_filtered_inputs` to True if
         # function returns a dict. If set to False, no new arrow table will be created
         update_data = None
-        # This variable is set in `apply_function_on_filtered_inputs` once we know
-        # the value of `update_data`
-        is_cached = None
 
         class NumExamplesMismatch(Exception):
             pass
 
-        # Check if the function returns updated examples
-        def does_function_return_dict(processed_inputs, indices):
-            """ Does the function returns a dict. """
-            does_return_dict = isinstance(processed_inputs, Mapping)
-
+        def validate_function_output(does_return_dict, processed_inputs, indices):
+            """ Validate output of the map function. """
             if does_return_dict is False and processed_inputs is not None:
                 raise TypeError(
                     "Provided `function` which is applied to all elements of table returns a variable of type {}. Make sure provided `function` returns a variable of type `dict` to update the dataset or `None` if you are only interested in side effects.".format(
@@ -1504,13 +1498,9 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
                         )
                     )
 
-            return does_return_dict
-
         def apply_function_on_filtered_inputs(inputs, indices, check_same_num_examples=False, offset=0):
             """ Utility to apply the function on a selection of columns. """
             nonlocal update_data
-            nonlocal cache_file_name
-            nonlocal is_cached
             fn_args = [inputs] if input_columns is None else [inputs[col] for col in input_columns]
             if offset == 0:
                 effective_indices = indices
@@ -1520,16 +1510,9 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
                 function(*fn_args, effective_indices, **fn_kwargs) if with_indices else function(*fn_args, **fn_kwargs)
             )
             if update_data is None:
-                update_data = does_function_return_dict(processed_inputs, indices)
-                # Check if we've already cached this computation (indexed by a hash)
-                if update_data and self._data_files:
-                    if cache_file_name is None:
-                        # we create a unique hash from the function,
-                        # current dataset file and the mapping args
-                        cache_file_name = self._get_cache_file_path(new_fingerprint)
-                    is_cached = os.path.exists(cache_file_name) and load_from_cache_file
-                else:
-                    is_cached = False
+                # Check if the function returns updated examples
+                update_data = isinstance(processed_inputs, Mapping)
+                validate_function_output(update_data, processed_inputs, indices)
             if not update_data:
                 return None  # Nothing to update, let's move on
             if remove_columns is not None:
@@ -1550,8 +1533,23 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
             inputs.update(processed_inputs)
             return inputs
 
+        def determine_cache_file_name(cache_file_name):
+            if self._data_files:
+                if cache_file_name is None:
+                    # we create a unique hash from the function,
+                    # current dataset file and the mapping args
+                    cache_file_name = self._get_cache_file_path(new_fingerprint)
+            return cache_file_name
+
+        def check_if_cached(cache_file_name):
+            if self._data_files:
+                return os.path.exists(cache_file_name) and load_from_cache_file
+            return False
+
+        buf_writer, writer, tmp_file = None, None, None
+
         # Prepare output buffer and batched writer in memory or on file if we update the table
-        def init_buffer_and_writer():
+        def init_buffer_and_writer(cache_file_name):
             writer_features = features
             if writer_features is None:
                 writer_features = self.features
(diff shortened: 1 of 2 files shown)


lhoestq (commented):
Awesome thanks !
I agree that the use of `nonlocal` is not pretty.
I feel like currently the dictionary type checking, the processing and the caching logics are mixed together. I think we could gain in readability a lot by separating all these a bit more, and also remove the `nonlocal` stuff.

I added a few comments in this direction, let me know what you think.

src/datasets/arrow_dataset.py
```diff
@@ -1559,6 +1555,12 @@ def apply_function_on_filtered_inputs(inputs, indices, check_same_num_examples=F
             if not batched:
                 for i, example in enumerate(pbar):
                     example = apply_function_on_filtered_inputs(example, i, offset=offset)
+                    # If cached, break and return a cached dataset, otherwise prepare resources
+                    if i == 0 and update_data:
+                        if not is_cached:
+                            buf_writer, writer, tmp_file = init_buffer_and_writer()
```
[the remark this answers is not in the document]
mariosasko: I agree it would be more natural, but we have to know the value of `update_data` (to calculate `cache_file_name`) to correctly initialize these resources and we know that value after processing the first example/batch, which means we have to enter the for loop.

src/datasets/arrow_dataset.py
```diff
@@ -1577,19 +1578,33 @@ def apply_function_on_filtered_inputs(inputs, indices, check_same_num_examples=F
...
+                    tmp_file.close()
+                    if os.path.exists(tmp_file.name):
+                        os.remove(tmp_file.name)
             raise
 
+        if update_data and is_cached:
+            logger.warning("Loading cached processed dataset at %s", cache_file_name)
+            info = self.info.copy()
+            info.features = features
+            return Dataset.from_file(cache_file_name, info=info, split=self.split)
```
mariosasko, line 1606: If it's ok to have a return statement inside the for loop in terms of style, then we can put this code inside the loop and remove the `is_cached` variable (for better readability?)

mariosasko (commented):
@lhoestq Addressed the comments and did some additional refactoring.

src/datasets/arrow_dataset.py
lhoestq: Sure looks good !

lhoestq (commented):
Awesome !
Feel free to add more comments to explain what's going on, and I think you can also start adding a test :)
Good job !

mariosasko:
@lhoestq Added a test. If you can come up with a better stateful callable, I'm all ears 😄.

src/datasets/arrow_dataset.py
```diff
@@ -1601,6 +1596,16 @@ def apply_function_on_filtered_inputs(inputs, indices, check_same_num_examples=F
                     for i, example in enumerate(pbar):
                         example = apply_function_on_filtered_inputs(example, i, offset=offset)
                         if update_data:
+                            if i == 0:
+                                cache_file_name = determine_cache_file_name(cache_file_name)
+                                if check_if_cached(cache_file_name):
+                                    logger.warning("Loading cached processed dataset at %s", cache_file_name)
+                                    info = self.info.copy()
+                                    info.features = features
+                                    return Dataset.from_file(cache_file_name, info=info, split=self.split)
```
lhoestq, line 1605: This can be outside the for loop, no ?
I don't think we need the first example to know if we can reload it from the cache

src/datasets/arrow_dataset.py
```diff
@@ -1618,17 +1623,28 @@ def apply_function_on_filtered_inputs(inputs, indices, check_same_num_examples=F
                                 "Using `.map` in batched mode on a dataset with attached indexes is allowed only if it doesn't create or remove existing examples. You can first run `.drop_index() to remove your index and then re-add it."
                             )
                         if update_data:
+                            if i == 0:
+                                cache_file_name = determine_cache_file_name(cache_file_name)
+                                if check_if_cached(cache_file_name):
+                                    logger.warning("Loading cached processed dataset at %s", cache_file_name)
+                                    info = self.info.copy()
+                                    info.features = features
+                                    return Dataset.from_file(cache_file_name, info=info, split=self.split)
```
lhoestq, line 1632: same here

lhoestq (commented):
Amazing ! 
The stateful mapping function you used for tests is good, thanks :)

lhoestq:
Sorry I said earlier that it was good to have it inside the loop, my mistake !

src/datasets/arrow_dataset.py
mariosasko: We need the value of `update_data`, which we know after processing the first example, to determine the name of the cache file (and to reload the dataset if this file already exists), so we have to enter the loop. 

The https://github.com/huggingface/datasets/pull/1960/commits/c38110c9b5a22e645eadd1a766b4fc270bd23758 commit has the approach where the "return" part is outside the loop, but I think the current solution is cleaner and easier to follow.

lhoestq: I don't think we need the value of `update_data` to get the cache file name ?
Before the loop you can directly compute the cache file name and if it exists and load_from_cache_file is True, then reload the dataset before entering the loop.
Does that make sense or I missed something ?

mariosasko:
@lhoestq Okay, did some refactoring and now the "cache" part comes before the for loop. Thanks for the guidance.

Think this is ready for the final review.

lhoestq (approved):
Excellent ! Looks all good to me, thanks :)

PR #1962, opened 2021-02-28
Fix unused arguments

mariosasko:
Noticed some args in the codebase are not used, so managed to find all such occurrences with Pylance and fix them.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -1536,6 +1536,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
                     writer_batch_size=writer_batch_size,
                     update_features=update_features,
                     fingerprint=new_fingerprint,
+                    disable_nullable=disable_nullable,
                 )
             else:
                 buf_writer = None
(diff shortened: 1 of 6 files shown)


src/datasets/features.py
```diff
@@ -457,7 +457,7 @@ def __array__(self, dtype=None):
         else:
             return self._data.astype(dtype)
 
-    def copy(self, deep: bool = False) -> "PandasArrayExtensionArray":
```
lhoestq, line 460: This one we must keep it since it's required in the pandas interface that is being implemented.

src/datasets/arrow_dataset.py
```diff
@@ -1537,6 +1537,7 @@ def apply_function_on_filtered_inputs(inputs, indices, check_same_num_examples=F
                     writer_batch_size=writer_batch_size,
                     update_features=update_features,
                     fingerprint=new_fingerprint,
+                    disable_nullable=disable_nullable,
```
lhoestq, line 1540: Good catch ! This one is an artifact from older versions

lhoestq (commented):
Thanks :)

There's just one argument that we actually need to keep, can you re-add it please ?

mariosasko:
@lhoestq Re-added the arg. The ConnectionError in CI seems unrelated to this PR (the same test fails on master as well).

lhoestq:
Thanks !
I'm re-running the CI, maybe this was an issue with circleCI

lhoestq:
Looks all good now, merged :)

PR #1970, opened 2021-03-02
Fixing the URL filtering for bad MLSUM examples in GEM

yjernite:
This updates the code and metadata to use the updated `gem_mlsum_bad_ids_fixed.json` file provided by @juand-r

cc @sebastianGehrmann


juand-r:
@yjernite Shouldn't ``"mlsum_es"`` in datasets/gem/gem.py also contain:

"bad_ids": "https://storage.googleapis.com/huggingface-nlp/datasets/gem/gem_mlsum_bad_ids_fixed.json"

(rather than the original json).  Thanks!

PR #1971, opened 2021-03-02
Fix ArrowWriter closes stream at exit

albertvillanova:
Current implementation of ArrowWriter does not properly release the `stream` resource (by closing it) if its `finalize()` method is not called and/or an Exception is raised before/during the call to its `finalize()` method.

Therefore, ArrowWriter should be used as a context manager that properly closes its `stream` resource at exit.


src/datasets/metric.py
```diff
@@ -336,7 +336,8 @@ def _finalize(self):
         from all the nodes if main node or all_process is True.
         """
         if self.writer is not None:
-            self.writer.finalize()
+            with self.writer as writer:
+                writer.finalize()
         self.writer = None
```
albertvillanova, line 341: Hi @lhoestq , do you think this is the cause of the test error? I am trying to understand why... It happens with `DistributedDatasetTest.test_load_dataset_distributed`. Maybe you know why...

albertvillanova:
> Hopefully this gives a better experience on windows on which it's super important to close stuff.

Exactly! On Windows, you got:
> PermissionError: [WinError 32] The process cannot access the file because it is being used by another process

when trying to access the unclosed `stream` file, e.g. by `with incomplete_dir(self._cache_dir) as tmp_data_dir`.

The reason is: https://docs.python.org/3/library/os.html#os.remove

> On Windows, attempting to remove a file that is in use causes an exception to be raised; on Unix, the directory entry is removed but the storage allocated to the file is not made available until the original file is no longer in use.

lhoestq:
The test passes on my windows. This was probably a circleCI issue. I re-ran the circleCI tests

albertvillanova:
NICE! It passed!

src/datasets/metric.py
```diff
@@ -336,7 +336,8 @@ def _finalize(self):
         from all the nodes if main node or all_process is True.
         """
         if self.writer is not None:
-            self.writer.finalize()
+            with self.writer as writer:
+                writer.finalize()
```
lhoestq, line 340: The usage of the `with` statement here can make the user think that the writer opens and creates a file here.
This is not the case since finalize just writes down the latest examples and close the file.

It would be prettier to use .close() instead of a with statement here, to remove this ambiguity.

The with statement should be used to define the start and end point of the usage of the writer (file opened until closed).
However in this case the writer is an attribute of the Metric object and so there's no code block with a defined start/end.
In this case I think it's better to use .close() instead. You can even do try/except to make sure that close() is called even if finalize fails.

It would makes things clearer IMO.

lhoestq (commented):
Really cool thanks :)

I just have one comment:

src/datasets/metric.py
albertvillanova: @lhoestq, I think the main point is in your sentence: 
> You can even do try/except to make sure that close() is called even if finalize fails.
Exactly! This is why I decided to open this PR: an exception was raised, and the resource was not closed. If an exception is raised within .finalize(), the file will no be closed. So that the solution would be: try... except... finally: close().

And that is exactly the use case for `with`: use a context manager instead of try... except... finally: close().
See: https://docs.python.org/3/reference/compound_stmts.html#the-with-statement
> This allows common try…except…finally usage patterns to be encapsulated for convenient reuse.

lhoestq: In this case maybe you can you add in the writer's docstring that the `with` can also be used only when calling `finalize` ?
As a user I'd expect the `with` statement of a `Writer` class to be necessarily used from start of writing till closing, but here you can write and then use the `with` statement to close. Let me know if that makes sense. 

That's definitely not a big deal but I'd rather have things said explicitly :)

lhoestq: Sounds good !

lhoestq (approved):

lhoestq:
Maybe you can merge master into this branch and check the CI before merging ?

albertvillanova:
@lhoestq done! ;)

PR #1975, opened 2021-03-02
Fix flake8

albertvillanova:
Fix flake8 style.

### DIFF
--- a/datasets/math_dataset/math_dataset.py
+++ b/datasets/math_dataset/math_dataset.py
@@ -232,9 +232,9 @@ class MathDataset(datasets.GeneratorBasedBuilder):
                 with open(data_file, encoding="utf-8") as f:
                     ls = f.read().split("\n")
 
-                    for l in ls[::-1]:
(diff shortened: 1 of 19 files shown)


datasets/math_dataset/math_dataset.py
```diff
@@ -232,9 +232,9 @@ def _read_data_from_all_categories(self, directory, config, categories):
                 with open(data_file, encoding="utf-8") as f:
                     ls = f.read().split("\n")
 
-                    for l in ls[::-1]:
-                        if not l:
-                            ls.remove(l)
+                    for k in ls[::-1]:
```
lhoestq, line 235: I guess you can replace "l" by "line" instead of "k"

lhoestq (approved):
Thanks :)

datasets/math_dataset/math_dataset.py
albertvillanova: Sure, done @lhoestq

PR #1976, opened 2021-03-02
Add datasets full offline mode with HF_DATASETS_OFFLINE

lhoestq:
Add the HF_DATASETS_OFFLINE environment variable for users who want to use `datasets` offline without having to wait for the network timeouts/retries to happen. This was requested in https://github.com/huggingface/datasets/issues/1939

cc @stas00


SBrandeis (approved):
Awesome!

PR #1978, opened 2021-03-03
Adding ro sts dataset

lorinczb:
Adding [RO-STS](https://github.com/dumitrescustefan/RO-STS) dataset


datasets/ro_sts_parallel/ro_sts_parallel.py
```diff
@@ -0,0 +1,121 @@
+# coding=utf-8
```
lhoestq, line 1: I feel like the parallel version of the Ro-STS dataset could be an additional configuration to the `ro_sts` dataset.

The `ro_sts` dataset already has a configuration named `ro_sts`. Maybe we can have an additional config named `parallel` ? What do you think ?

datasets/ro_sts_parallel/ro_sts_parallel.py
```diff
@@ -0,0 +1,121 @@
...
+        self.language_pair = language_pair
+
+
+class ROSTSParallel(datasets.GeneratorBasedBuilder):
+    """RO-STS-Parallel dataset"""
+
+    BUILDER_CONFIGS = [
+        ROSTSParallelConfig(
+            name="rosts-parallel-ro-en",
+            language_pair=("ro", "en"),
```
lhoestq, line 64: Since there's only one language pair, no need to have two configurations (the order doesn't matter)

lhoestq (commented):
Awesome thanks for adding this one !

I left a few comments:

datasets/ro_sts/ro_sts.py
```diff
@@ -0,0 +1,123 @@
...
+"""RO-STS: The Romanian Semantic Textual Similarity Dataset"""
+
+from __future__ import absolute_import, division, print_function
+
+import datasets
+
+
+# Find for instance the citation on arxiv or on the dataset repo/website
+_CITATION = """\
+Article under review
```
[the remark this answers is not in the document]
lorinczb: Sure, I will add the citation, once we have it published :)

datasets/ro_sts/ro_sts.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+class ROSTSConfig(datasets.BuilderConfig):
+    """BuilderConfig for RO-STS dataset"""
+
+    def __init__(self, **kwargs):
+        super(ROSTSConfig, self).__init__(**kwargs)
+
+
+class ROSTS(datasets.GeneratorBasedBuilder):
```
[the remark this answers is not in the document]
lorinczb: I will adjust this ro_sts_parallel version as well, to this camel case version

datasets/ro_sts_parallel/ro_sts_parallel.py
lorinczb: We have thought about this option as well, but the data files have different structures, and different tasks attached. On the [GitHub repo](https://github.com/dumitrescustefan/RO-STS/tree/master/dataset) we have two separate folders for them, and we kind of handle them as separate datasets. (Both datasets build upon the STS dataset translation, but it is offered in two flavors, one for semantic textual similarity and the other for machine translation.) What do you think, maybe we can leave them as two different datasets?

datasets/ro_sts_parallel/ro_sts_parallel.py
lorinczb: Yes, right, the order doesn't matter, I have kept the version that you suggested and removed the other one. Thanks! I am adding the changes now, with the dataset_infos.json regenerated as well.

datasets/ro_sts/README.md
```diff
@@ -0,0 +1,147 @@
...
+  - [Licensing Information](#licensing-information)
+  - [Citation Information](#citation-information)
+  - [Contributions](#contributions)
+
+## Dataset Description
+
+- **Homepage:** https://github.com/dumitrescustefan/RO-STS
+- **Repository:** https://github.com/dumitrescustefan/RO-STS
+- **Paper:** [Needs More Information]
+- **Leaderboard:** [LiRo]
```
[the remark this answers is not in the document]
lorinczb: Yes, sorry, I have left the leaderboard name in here, once the article is published we will launch the leaderboard site, right now it has a temporary link. I will add the link to leaderboard and citation at the same time, hopefully soon :). For now I have just changed to [Needs More Information], hope that is okay.

lorinczb:
@lhoestq thank you very much for the quick review and useful comments! 

I have tried to address them all, and a few comments that you left for ro_sts I have applied to the ro_sts_parallel as well (in read-me: fixed source_datasets, links to homepage, repository, leaderboard, thanks to me message, in ro_sts_parallel.py changed to camel case as well). In the ro_sts_parallel I have changed the order on the languages, also in the example, as you said order doesn't matter, but just to have them listed in the readme in the same order.

I have commented above on why we would like to keep them as separate datasets, hope it makes sense.

If there is anything else I should change please let me know.

Thanks again!

lorinczb:
@lhoestq I tried to adjust the ro_sts_parallel, locally when I run the tests they are passing, but somewhere it has the old name of rosts-parallel-ro-en which I am trying to change to ro_sts_parallel. I don't think I have left anything related to rosts-parallel-ro-en, but when the dataset_infos.json is regenerated it adds it. Could you please help me out, how can I fix this? Thanks in advance!

datasets/ro_sts_parallel/ro_sts_parallel.py
lhoestq: Ok that makes sense !

lhoestq (approved):
Awesome thank you for the changes ! 

I fixed the CI issue, there was just an unused file in a wrong directory that I removed.

Looks all good now :)

lorinczb:
Great, thanks for all your help!

PR #1979, opened 2021-03-03
Add article_id and process test set template for semeval 2020 task 11…

hemildesai:
… dataset

- `article_id` is needed to create the submission file for the task at https://propaganda.qcri.org/semeval2020-task11/
- The `technique classification` task provides the span indices in a template for the test set that is necessary to complete the task. This PR implements processing of that template for the dataset.


datasets/sem_eval_2020_task_11/sem_eval_2020_task_11.py
```diff
@@ -151,7 +165,7 @@ def _split_generators(self, dl_manager):
             ),
         ]
 
-    def _generate_examples(self, data_dir, keys, split, labels):
+    def _generate_examples(self, data_dir, keys, split, labels, **kwargs):
```
lhoestq, line 168:
```suggestion
    def _generate_examples(self, data_dir, keys, split, labels, tc_test_template):
```
We can directly have the right argument here.

For the train and val set, you can just set `"tc_test_template": None` in their `gen_kwargs`

lhoestq (commented):
Cool thank you :)

datasets/sem_eval_2020_task_11/sem_eval_2020_task_11.py
hemildesai: Done!

datasets/sem_eval_2020_task_11/sem_eval_2020_task_11.py
```diff
@@ -127,6 +129,12 @@ def _split_generators(self, dl_manager):
         """Returns SplitGenerators."""
         path_to_manual_file = os.path.abspath(os.path.expanduser(dl_manager.manual_dir))
 
+        tc_labels_template = os.path.join(path_to_manual_file, "test-task-tc-template.out")
+        if os.path.isfile(tc_labels_template):
+            tc_test_template = self._process_tc_labels_template(tc_labels_template)
+        else:
+            tc_test_template = None
```
lhoestq, line 136: Out of curiosity, why do you need to test if the file exists ?

hemildesai: This is a classification task on a span in a given article. So this template just provides the offsets for the spans to be classified in an article.

lhoestq: I see.
And in these lines you're checking if the file with the template exists.
Is this verification necessary ? Isn't the file always present in the `ptc-corpus.tgz` file downloaded for this dataset ?

lhoestq: For reproducibility concerns we try to have deterministic dataset generations.
Do you think we can simply assume that this file always exists ?
Otherwise we can keep it this way..

hemildesai: Ok, changed it to make it deterministic.

lhoestq:
Thanks !
Now to fix the CI the only thing left is to add a dummy `test-task-tc-template.out` file inside the `dummy_data.zip` at `./datasets/sem_eval_2020_task_11/dummy/1.1.0`
It must contain the labels template for each dummy article of the test set included in `dummy_data.zip`

After that we should be good to merge this one :)

hemildesai:
@lhoestq Made the changes! The failure now seems to be unrelated to the changes. Any idea what's going on?

lhoestq:
This is a bug on master that we're investigating. You can ignore it

lhoestq (approved):
Looks all good now ! Thanks again :)

merging

gchhablani:
Hi @lhoestq @hemildesai,

I get a flake8 issue on the changes:
```bash
datasets/sem_eval_2020_task_11/sem_eval_2020_task_11.py:240:58: E741 ambiguous variable name 'l'
```

PR #1980, opened 2021-03-03
Loading all answers from drop

KaijuML:
Hello all,

I propose this change to the DROP loading script so that all answers are loaded no matter their type. Currently, only "span" answers are loaded, which excludes a significant amount of answers from drop (i.e. "number" and "date").

I updated the script with the version I use for my work. However, I couldn't find a way to verify that all is working when integrated with the datasets repo, since the `load_dataset` method seems to always download the script from github and not local files.

Note that 9 items from the train set have no answers, as well as 1 from the validation set. The script I propose simply do not load them.

Let me know if there is anything else I can do,
Clément


datasets/drop/drop.py
```diff
@@ -69,28 +89,75 @@ def _split_generators(self, dl_manager):
...
+
+                    if split == "train":
+                        answers = [qa["answer"]]
+                    else:
+                        answers = qa["validated_answers"]
+
+                    try:
+                        example["answers_spans"] = self.build_answers(answers)
+                        yield example["query_id"], example
+                    except AssertionError:
```
lhoestq, line 124: Maybe you can have a more explicit error type here ?
Feel free to create a `AnswerParsingError` exception instead of using `assert`.

lhoestq (commented):
Hi Clément :)
Thanks for proposing this change ! This looks really nice

When you're working locally on a dataset script and you want to test it using `load_dataset`, please provide the local path to the script. For example:
```python
load_dataset("./datasets/drop")
```

lhoestq:
Nice thanks for the change !
This looks all good to me

Before we merge can you just update the dataset_infos.json file of drop ? You can do it by running
```
datasets-cli test ./datasets/drop --all_configs --save_infos --ignore_verifications
```

KaijuML:
Done!

lhoestq (approved):
Thanks ! :)

PR #1995, opened 2021-03-05
[Timit_asr] Make sure not only the first sample is used 

patrickvonplaten:
When playing around with timit I noticed that only the first sample is used for all indices. I corrected this typo so that the dataset is correctly loaded.


patrickvonplaten:
cc @lhoestq  @vrindaprabhu

patrickvonplaten:
Failing `run (push)` is unrelated -> merging

elgeish:
Thanks for fixing this, it was affecting my runs for https://github.com/huggingface/transformers/pull/10581/

vrindaprabhu:
I am seeing this very late! Sorry for the blunder everyone! :(

PR #1998, opened 2021-03-06
Add -DOCSTART- note to dataset card of conll-like datasets

### LINKED ISSUE #1983: The size of CoNLL-2003 is not consistant with the official release.
Thanks for the dataset sharing! But when I use conll-2003, I meet some questions.
The statistics of conll-2003 in this repo is  : 
\#train 14041  \#dev 3250 \#test 3453
While the official statistics is:
\#train 14987 \#dev 3466 \#test 3684
Wish for your reply~

mariosasko:
Closes #1983


lhoestq (commented):
Thanks !

I just looked for mentions of "DOCSTART" in the other datasets and none of them contain mentions of "DOCSTART". The only one with these lines is `conll2003`.

I guess the `if line.startswith("-DOCSTART-")` code comes from many bad copy-pasting here...

So I would only mention that "DOCSTART" lines are filtered out for `conll2003` and not the others.

In a next PR we can remove the bad `if line.startswith("-DOCSTART-")` checks...

mariosasko:
Nice catch! Yes I didn't check the actual data, instead I was just looking for the `if line.startswith("-DOCSTART-")` pattern.

Issue #2009, opened 2021-03-09
Ambiguous documentation


theo-m:
https://github.com/huggingface/datasets/blob/2ac9a0d24a091989f869af55f9f6411b37ff5188/templates/new_dataset_script.py#L156-L158

Looking at the template, I find this documentation line to be confusing, the method parameters don't include the `gen_kwargs` so I'm unclear where they're coming from.

Happy to push a PR with a clearer statement when I understand the meaning.

SBrandeis:
Hi @theo-m !

A few lines above this line, you'll find that the `_split_generators` method returns a list of `SplitGenerator`s objects:

```python
datasets.SplitGenerator(
    name=datasets.Split.VALIDATION,
    # These kwargs will be passed to _generate_examples
    gen_kwargs={
        "filepath": os.path.join(data_dir, "dev.jsonl"),
        "split": "dev",
    },
),
```

Notice the `gen_kwargs` argument passed to the constructor of `SplitGenerator`: this dict will be unpacked as keyword arguments to pass to the `_generat_examples` method (explaining in this case the `filepath` and `split` arguments).

theo-m:
Oh ok I hadn't made the connection between those two, will offer a tweak to the comment and the template then.

PR #2014, opened 2021-03-09
more explicit method parameters

theo-m:
re: #2009

not super convinced this is better, and while I usually fight against kwargs here it seems to me that it better conveys the relationship to the `_split_generator` method.

### DIFF
--- a/templates/new_dataset_script.py
+++ b/templates/new_dataset_script.py
@@ -153,24 +153,25 @@ class NewDataset(datasets.GeneratorBasedBuilder):
             ),
         ]
 
-    def _generate_examples(self, **gen_kwargs):
-        """ Yields examples. """
-        # TODO: This method will receive as arguments the `gen_kwargs` defined in the previous `_split_generators` method.
-        # It is in charge of opening the given file and yielding (key, example) tuples from the dataset
-        # The key is not important, it's more here for legacy reason (legacy from tfds)
+    def _generate_examples(
+        self, filepath, split  # method parameters are unpacked from `gen_kwargs` as given in `_split_generators`
+    ):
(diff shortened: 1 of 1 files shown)


templates/new_dataset_script.py
```diff
@@ -153,24 +153,24 @@ def _split_generators(self, dl_manager):
             ),
         ]
 
-    def _generate_examples(self, filepath, split):
+    def _generate_examples(self, **gen_kwargs):
```
lhoestq, line 156: I agree it makes the connection between what's defined in the SplitGenerator gen_kwargs clearer when reading the template. But when implementing an actual dataset it's better to not have kwargs.

Plus it's already mentioned in each SplitGenerator definition that
```
# These kwargs will be passed to _generate_examples
``` 
and also in `_generate_examples`:
```
# TODO: This method will receive as arguments the `gen_kwargs` defined in the previous `_split_generators` method.
```

Maybe we can add another comment:
```suggestion
    def _generate_examples(self, filepath, split):  # you can add arguments to this method by passing them in the `gen_kwargs`
```

Let me know what you think

theo-m: Yeah I'll definitely concede that it's already written in the clear - maybe it's more that I've not read python for a while rather than the template being ambiguous.

> But when implementing an actual dataset it's better to not have kwargs.

Yup I agree for sure. Weak opinion: all the examples in the repo have named parameters and set the tone, while the template is for onboarding and should favour clearer "magic" where possible. 
But all in all my aversion for kwargs prevents me from pushing too much in that direction - will push an additional comment

lhoestq (approved):

PR #2019, opened 2021-03-09
Replace print with logging in dataset scripts

mariosasko:
Replaces `print(...)` in the dataset scripts with the library logger.


mariosasko:
Maybe a script or even a test in `test_dataset_common.py` that verifies that a dataset script meets some set of quality standards (print calls and todos from the dataset script template are not present, etc.) could be added?

lhoestq:
Yes definitely !

PR #2023, opened 2021-03-10
Add Romanian to XQuAD

M-Salti:
On Jan 18, XQuAD was updated with a new Romanian validation file ([xquad commit link](https://github.com/deepmind/xquad/commit/60cac411649156efb6aab9dd4c9cde787a2c0345))


lhoestq:
Hi ! Thanks for updating XQUAD :)

The slow test is failing though since there's no dummy data nor metadata in dataset_infos.json for the romanian configuration.

Could you please generate the dummy data with
```
datasets-cli dummy_data ./datasets/xquad --auto_generate --json_field data
```
This will update all the dummy data files, and also add the new one for the romanian configuration.


You can also update the metadata with
```
datasets-cli test ./datasets/xquad --name xquad.ro --save_infos
```
This will update the dataset_infos.json file with the metadata of the romanian config :)

Thanks in advance !

M-Salti:
Hello Quentin, and thanks for your help.

I found that running

```python
datasets-cli test ./datasets/xquad --name xquad.ro --save_infos
```

was not enough to pass the slow tests, because it was not adding the new `xquad.ro.json` checksum to the other configs infos and becuase of that an `UnexpectedDownloadedFile` error was being thrown, so instead I used:

```python
datasets-cli test ./datasets/xquad --save_infos --all_configs --ignore_verifications
```

`--ignore_verifications` was necessary to bypass the same `UnexpectedDownloadedFile` error.

Additionally, I deleted `dummy_data_copy.zip` and the `copy.sh` script because they both seem now unnecessary.

The slow tests for both the real and dummy data now pass successfully, so I hope that I didn't mess anything up :)

lhoestq:
You're right, you needed the `--ignore_verifications` flag !
Thanks for updating them :)

Although I just noticed that the new dummy_data.zip files are quite big (170KB each) because they contain the json files of all the languages, while only one json file per language is necessary. Could you remove the unnecessary json files to reduce the size of the dummy_data.zip files if you don't mind ?

M-Salti:
Done. I created a script (`remove_unnecessary_langs.sh`) to automate the process.

lhoestq (approved):
Thanks !

PR #2025, opened 2021-03-10
[Refactor] Use in-memory/memory-mapped/concatenation tables in Dataset

### LINKED ISSUE #1877: Allow concatenation of both in-memory and on-disk datasets
This is a prerequisite for the addition of the `add_item` feature (see #1870).
Currently there is one assumption that we would need to change: a dataset is either fully in memory (dataset._data_files is empty), or the dataset can be reloaded from disk (using the dataset._data_files).
This assumption is used for pickling for example:
- in-memory dataset can just be pickled/unpickled in-memory
- on-disk dataset can be unloaded to only keep the filepaths when pickling, and then reloaded from the disk when unpickling

Maybe let's have a design that allows a Dataset to have a Table that can be rebuilt from heterogenous sources like in-memory tables or on-disk tables ? This could also be further extended in the future

One idea would be to define a list of sources and each source implements a way to reload its corresponding pyarrow Table.
Then the dataset would be the concatenation of all these tables.

Depending on the source type, the serialization using pickle would be different. In-memory data would be copied while on-disk data would simply be replaced by the path to these data.

If you have some ideas you would like to share about the design/API feel free to do so :)

cc @albertvillanova

lhoestq:
## Intro

Currently there is one assumption that we need to change: a dataset is either fully in memory (dataset._data_files is empty), or the dataset can be reloaded from disk with memory mapping (using the dataset._data_files).
This assumption is used for pickling for example:
- in-memory dataset can just be pickled/unpickled in-memory
- on-disk dataset can be unloaded to only keep the filepaths when pickling, and then reloaded from the disk when unpickling

## Issues

Because of this assumption, we can't easily implement methods like `Dataset.add_item` to append more rows to a dataset, or `dataset.add_column` to add a column, since we can't mix data from memory and data from the disk.
Moreover, `concatenate_datasets` doesn't work if the datasets to concatenate are not all from memory, or all form the disk.

## Solution provided in this PR

I changed this by allowing several types of Table to be used in the Dataset object.
More specifically I added three pyarrow Table wrappers: InMemoryTable, MemoryMappedTable and ConcatenationTable.
The in-memory and memory-mapped tables implement the pickling behavior described above.
The ConcatenationTable can be made from several tables (either in-memory or memory mapped) called "blocks". Pickling a ConcatenationTable simply pickles the underlying blocks.

## Implementation details

The three tables classes mentioned above all inherit from a `Table` class defined in `table.py`, which is a wrapper of a pyarrow table. The `Table` wrapper implements all the attributes and methods of the underlying pyarrow table.

Regarding the MemoryMappedTable:
Reloading a pyarrow table from the disk makes you lose all the changes you may have applied (slice, rename_columns, drop, cast etc.). Therefore the MemoryMappedTable implements a "replay" mechanism to re-apply the changes when reloading the pyarrow table from the disk.

## Checklist

- [x] add InMemoryTable
- [x] add MemoryMappedTable
- [x] add ConcatenationTable
- [x] Update the ArrowReader to use these new tables depending on the `in_memory` parameter
- [x] Update Dataset.from_xxx methods
- [x] Update load_from_disk and save_to_disk
- [x] Backward compatibility of load_from_disk
- [x] Add tests for the new tables
- [x] Update current tests
- [ ] Documentation

----------

I would be happy to discuss the design of this PR :)

Close #1877


shamanez:
There is one more thing I would love to see. Let's say we iteratively keep updating a data source that loaded from **load_dataset** or **load_from_disk**. Now we need to save it to the same location by overriding the previous file inorder to save the disk space. At the moment **save_to_disk** can not assign a name. So I do not see an easy way to override the previous files.  @lhoestq  is this possible?

lhoestq:
> There is one more thing I would love to see. Let's say we iteratively keep updating a data source that loaded from **load_dataset** or **load_from_disk**. Now we need to save it to the same location by overriding the previous file inorder to save the disk space. At the moment **save_to_disk** can not assign a name. So I do not see an easy way to override the previous files. @lhoestq is this possible?

In the new save_to_disk, the filename of the arrow file is fixed: `dataset.arrow`.
This way is will be overwritten if you save your dataset again

> Is there a way to flush out any connection to a data source loaded from **load_from_disk** or **load_dataset** methods? At the moment I suspect when we use any of those functions, it will always keep a pointer although we override it again with a new version of the dataset source. This is really useful in an iterative process.

If you update an arrow file, then you must reload it with `load_from_disk` for example in order to have the updated data.
Does that answer the question ? How does this "pointer" behavior manifest exactly on your side ?

lhoestq:
Apparently the usage of the compute layer of pyarrow requires pyarrow>=1.0.0 (otherwise there are some issues on windows with file permissions when doing dataset concatenation).

I'll bump the pyarrow requirement from, 0.17.1 to 1.0.0

shamanez:
> If you update an arrow file, then you must reload it with `load_from_disk` for example in order to have the updated data.
> Does that answer the question? How does this "pointer" behavior manifest exactly on your side?

Yes,  I checked this behavior.. if we update the .arrow file it kind of flushes out the previous one. So your solution is perfect <3.

lhoestq:
Sorry for spamming, there's a a bug that only happens on the CI so I have to re-run it several times

src/datasets/arrow_dataset.py
```diff
@@ -439,6 +439,16 @@ def save_to_disk(self, dataset_path: str, fs=None):
             dataset_path = extract_path_from_uri(dataset_path)
         else:
             fs = fsspec.filesystem("file")
+            cache_files_paths = [Path(cache_filename) for cache_filename in self.cache_files]
+            # Check that the dataset doesn't overwrite iself. It can cause a permission error on Windows and a segfault on linux.
+            if Path(dataset_path).joinpath(DATASET_ARROW_FILENAME) in cache_files_paths:
+                raise PermissionError(
+                    f"Tried to overwrite {Path(dataset_path).joinpath(DATASET_ARROW_FILENAME)} but a dataset can't overwrite itself."
```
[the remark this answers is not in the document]
lhoestq: No it just means that you can't write a Table in the file where it is loaded from (or the Table breaks itself).
i.e. this is forbidden:
```python
path = "path/to/my/dataset"
d = load_from_disk(path)
d = d.rename_column("a", "b")  # this DOESN'T create a new arrow file (the dataset still has the same data)
d.save_to_disk(path)
```

But this is ok:
```python
path = "path/to/my/dataset"
d = load_from_disk(path)
d = d.map(...)  # this DOES create a new arrow file for the new data
d.save_to_disk(path)
```

I'm not 100% convinced by this though. I can probably make this work in all caes by using a temp file rather than writing directly to the arrow file. I'll let you know how this goes.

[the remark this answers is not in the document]
shamanez: oh got it.  I like map functions creating a new arrow file since we can keep the original dataset unchanged. Thanks a lot.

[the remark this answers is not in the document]
lhoestq: On windows it's not possible to overwrite a file that is already opened for a Table unfortunately so I think we'll have to keep this for now

lhoestq:
Alright I finally added all the tests I wanted !
I also fixed all the bugs and now all the tests are passing :)

Let me know if you have comments.

I also noticed that two methods in pyarrow seem to bring some data in memory even for a memory mapped table: filter and cast:
- for filter I took a look at the C++ code on the arrow's side and found [this part](https://github.com/apache/arrow/blob/55c8d74d5556b25238fb2028e9fb97290ea24684/cpp/src/arrow/compute/kernels/vector_selection.cc#L93-L160) that "builds" the array during filter. It seems to indicate that it allocates new memory for the filtered array but not 100% sure.
- regarding cast I noticed that it happens when changing the precision of an array of integers. Not sure if there are other cases.


Maybe we'll need to investigate this a bit for your PR on improving `filter` @theo-m , since we don't want to fill the users memory.

theo-m:
> Maybe we'll need to investigate this a bit for your PR on improving `filter` @theo-m , since we don't want to fill the users memory.

I'm a bit unclear on this, I thought the point of the refactor was to use `Table.filter` to speed up our own `.filter` and stop using `.map` that offloaded too much stuff on disk. 
At some point I recall we decided to use `keep_in_memory=True`  as the expectations were that it would be hard to fill the memory?

SBrandeis (approved):
Just a few minor suggestions/questions.
Awesome work, @lhoestq !!

src/datasets/table.py
```diff
@@ -0,0 +1,585 @@
...
+        raise NotImplementedError()
+
+    @wraps_arrow_table_method(pa.Table.drop)
+    def drop(self, *args, **kwargs):
+        raise NotImplementedError()
+
+
+class TableBlock(Table):
+    pass
+
```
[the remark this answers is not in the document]
lhoestq: TableBlock is the allowed class inside a ConcanetationTable. Only MemoryMappedTable and InMemoryTable are TableBlock.
This way you can't have a ConcanetationTable made out of other ConcanetationTables.
I'll add a docstring to make this clearer

lhoestq:
> I'm a bit unclear on this, I thought the point of the refactor was to use Table.filter to speed up our own .filter and stop using .map that offloaded too much stuff on disk.
> At some point I recall we decided to use keep_in_memory=True as the expectations were that it would be hard to fill the memory?

Yes it's ok to have the mask in memory, but not the full table. I was not aware that the table returned by filter could actually be in memory (it's not part of the pyarrow documentation afaik).
To be more specific I noticed that every time you call `filter`, the pyarrow total allocated memory increases.
I haven't checked on a big dataset though, but it would be nice to see how much memory it uses with respect to the size of the dataset.

src/datasets/formatting/formatting.py
```diff
@@ -303,7 +319,9 @@ def key_to_query_type(key: Union[int, slice, range, str, Iterable]) -> str:
 
 
 def query_table(
-    pa_table: pa.Table, key: Union[int, slice, range, str, Iterable], indices: Optional[pa.lib.UInt64Array] = None
+    pa_table: Union[pa.Table, "Table"],
```
theo-m, line 322: since you're not inside the `Table` class body you could refer to it directly rather than with a string?

theo-m: also unclear why you'd accept both and not just your new abstraction?

src/datasets/arrow_dataset.py
```diff
@@ -223,31 +196,23 @@ def wrapper(*args, **kwargs):
...
 
 class Dataset(DatasetInfoMixin, IndexableMixin):
-    """A Dataset backed by an Arrow table or Record Batch."""
+    """A Dataset backed by an Arrow table."""
 
     def __init__(
         self,
-        arrow_table: pa.Table,
-        data_files: Optional[List[dict]] = None,
+        arrow_table: Union[pa.Table, Table],
```
theo-m, line 203: same question, why not rely only on your new abstraction?

src/datasets/arrow_dataset.py
```diff
@@ -589,68 +503,69 @@ def save_to_disk(self, dataset_path: str, fs=None):
...
+                json.dumps(state["_format_kwargs"][k])
+            except TypeError as e:
+                raise TypeError(str(e) + f"\nThe format kwargs must be JSON serializable, but key '{k}' isn't.")
+
+        # Get json serializable dataset info
+        dataset_info = asdict(self._info)
+
+        # Save dataset + indices + state + info
+        fs.makedirs(dataset_path, exist_ok=True)
+        with fs.open(Path(dataset_path).joinpath(DATASET_ARROW_FILENAME).as_posix(), "wb") as dataset_file:
```
theo-m, line 548: optional nit, pathlib overloads `/`, so that you can do `(mydirectory / myfile).as_posix()`, it just expects the left element to be a `pathlib.Path`.

src/datasets/table.py
```diff
@@ -0,0 +1,591 @@
+import copy
+from functools import wraps
+from typing import List, TypeVar, Union
+
+import pyarrow as pa
+
+
+def wraps_arrow_table_method(arrow_table_method):
```
theo-m, line 8:
```suggestion
def inject_arrow_table_documentation(arrow_table_method):
```

src/datasets/table.py
```diff
@@ -0,0 +1,591 @@
...
+
+
+def in_memory_arrow_table_from_buffer(buffer: pa.Buffer) -> pa.Table:
+    stream = pa.BufferReader(buffer)
+    opened_stream = pa.ipc.open_stream(stream)
+    table = opened_stream.read_all()
+    return table
+
+
+def memory_mapped_arrow_table_from_file(filename: str) -> pa.Table:
```
theo-m, line 31: these could be `@classmethod`?

src/datasets/table.py
```diff
@@ -0,0 +1,591 @@
...
+    This is because we don't want a ConcanetationTable made out of other ConcanetationTables.
+    """
+
+    pass
+
+
+class InMemoryTable(TableBlock):
+    """
+    Wraps a pyarrow Table using composition.
+    The table is said in-memory so pickling it does copy all the data using memory.
```
theo-m, line 188: worth explaining why it matters

src/datasets/table.py
```diff
@@ -0,0 +1,591 @@
...
+    Only MemoryMappedTable and InMemoryTable are TableBlock.
+    This is because we don't want a ConcanetationTable made out of other ConcanetationTables.
+    """
+
+    pass
+
+
+class InMemoryTable(TableBlock):
+    """
+    Wraps a pyarrow Table using composition.
```
theo-m, line 187: I'm not sure this line is very helpful, for `MemoryMappedTable` as well.

src/datasets/table.py
```diff
@@ -0,0 +1,591 @@
...
+            else:
+                blocks.append([t.slice(offset, length) for t in tables])
+                length, offset = 0, 0
+        return ConcatenationTable(table, blocks)
+
+    @wraps_arrow_table_method(pa.Table.filter)
+    def filter(self, mask, *args, **kwargs):
+        table = self.table.filter(mask, *args, **kwargs)
+        blocks = []
+        for (start, end), tables in zip(self._slices, self.blocks):
```
theo-m, line 492: the naming makes this quite hard to read, why is it called `end`?

src/datasets/table.py
```diff
@@ -0,0 +1,591 @@
...
+        return MemoryMappedTable(self.table.rename_columns(*args, **kwargs), self.path, replay)
+
+    @wraps_arrow_table_method(pa.Table.drop)
+    def drop(self, *args, **kwargs):
+        replay = [("drop", copy.deepcopy(args), copy.deepcopy(kwargs))]
+        replay = self._merge_replay(replay)
+        return MemoryMappedTable(self.table.drop(*args, **kwargs), self.path, replay)
+
+
+BlockPart = TypeVar("BlockPart", TableBlock, List[TableBlock], List[List[TableBlock]])
```
theo-m, line 381: why would you need to handle lists of lists?

theo-m (commented):
looks cool! left some comments, but mostly a shallow review because not comfortable enough with `pyarrow` to spot shady stuff

src/datasets/table.py
lhoestq: First axis concatenates the rows, and second axis concatenates the columns.
This way we can implement Dataset.add_column for example.

src/datasets/table.py
lhoestq: `end` is a very bad name here. Let me change it

src/datasets/table.py
lhoestq: will add more context in the docstring, especially regarding memory handling with pyarrow

src/datasets/table.py
lhoestq: Since they return arrow tables and not InMemoryTable or MemoryMapped table, I hesitated to have them as classmethods. No strong opinion though

src/datasets/table.py
lhoestq: much clearer, thanks
doing this change now

src/datasets/arrow_dataset.py
lhoestq: I liked the idea of allowing users to use regular arrow tables as well, but maybe it's simpler if we only allow the new abstraction

src/datasets/table.py
theo-m: Oh ok, I had missed the fact that concatenation could be "horizontal" too. worth a comment?

src/datasets/table.py
theo-m: oh yes my bad, misread the `pa.Table` into `Table`. maybe prefix them with an underscore as they seem to be a class util that wouldn't be used from outside?

src/datasets/arrow_dataset.py
theo-m: Yes I think it would make everything clearer, documentation wise - since the typing here is not enforced.
Contractually (e.g. not statically but for runtime) a `pa.Table` is a `Table` too, so this use is allowed too? not sure this is a very good point to make, but whenever we can simplify stuff... :)

src/datasets/table.py
lhoestq: yup I'm adding one

src/datasets/arrow_dataset.py
lhoestq: Yes this is allowed since they have the exact same API

src/datasets/arrow_dataset.py
theo-m: tbh you can also just `Path(mydir, myfile)` which is shorter and as explicit. not as fun as using the `/` to visually signify fs hierarchy, but has its merits 😛

src/datasets/table.py
```diff
@@ -0,0 +1,591 @@
...
+        return state
+
+    def __setstate__(self, state):
+        state = state.copy()
+        state["table"] = ConcatenationTable.from_blocks(state["blocks"]).table
+        self.__dict__ = state
+
+    @staticmethod
+    def from_blocks(blocks_part: BlockPart):
+        if isinstance(blocks_part, Table):
```
albertvillanova, line 424: Why not checking more specifically that is instance of `TableBlock`?

The same below.

src/datasets/table.py
```diff
@@ -0,0 +1,591 @@
...
+            table = blocks_part
+            return ConcatenationTable(table.table, [[table]])
+        elif isinstance(blocks_part[0], Table):
```
albertvillanova, line 427: Maybe define a method called `concatenate_rows()`?

src/datasets/table.py
```diff
@@ -0,0 +1,591 @@
...
+            table = pa.concat_tables([t.table for t in blocks_part])
+            blocks = [[t] for t in blocks_part]
+            return ConcatenationTable(table, blocks)
+        else:
```
albertvillanova, line 431: Maybe define a method `concatenate_columns_and_rows()`?

src/datasets/arrow_dataset.py
```diff
@@ -72,6 +73,11 @@
     PYARROW_V0 = False
 
 
+DATASET_ARROW_FILENAME = "dataset.arrow"
+DATASET_INDICES_FILENAME = "indices.arrow"
+DATASET_STATE_JSON_FILENAME = "state.json"
```
albertvillanova, line 78: Maybe these could be defined in `config.py`.

albertvillanova (commented):
You did an awesome job, thanks!!!

src/datasets/arrow_dataset.py
theo-m: yes, my point was that it would mean that our documentation-typing is slightly misleading, but since we don't have interfaces, what can we do, right? 🤷 
so still in favour of the simplification.

lhoestq:
I have addressed your comments @theo-m @albertvillanova ! Thanks for the suggestions

albertvillanova (commented):
Hi @lhoestq, again thanks for your great work!! :heart:

There is one thing I have been thinking about... I have not a clear opinion yet, but maybe it is worth to discuss it with you (and the others).

It is about using Composition instead of Inheritance, for `Table` (and subclasses). I am a big fan of composition over inheritance: much more flexibility, loose coupling, etc. Indeed, it is the default approach I usually use in 99% of the cases. However, there are some rare cases where it might be better using inheritance. And I am wondering if this is not one of such rare cases...

Why?
- `Table` (and subclasses) must expose ALL the methods of `pa.Table`; this is done automatically by inheritance.
- Most of the methods just use the `pa.Table` implementation; using inheritance would avoid having to repeat all this methods in `Table`: they are just inherited.
- Some methods need a specific implementation; this can be achieved by overriding these methods from the base class implementation.

Therefore, by using inheritance, in `Table` code you would just need to implement the overridden methods (no need to repeat ALL the other inherited `pa.Table` methods).

What do you think? As I told you above, I have not a clear opinion yet.

lhoestq:
I totally agree with you. I would have loved to use inheritance instead.
However because `pa.Table` is a cython class without proper initialization methods (you can't call `__init__` for example): you can't instantiate a subclass of `pa.Table` in python.
To be more specific, you actually can instantiate a subclass of `pa.Table` with no data BUT this is not a valid table so you get an error.
And since `pa.Table` objects are immutable you can't even set the data in `__new__` or `__init__`.

albertvillanova:
@lhoestq, but in which cases you would like to instantiate directly either `InMemoryTable` or `MemoryMappedTable`? You normally use some of their `from_xxx` methods...

lhoestq:
Yes I was thinking of these cases. The issue is that they return `pa.Table` objects even from a subclass of `pa.Table`

albertvillanova:
That is indeed a weird behavior...

albertvillanova:
I guess that in this case, the best approach is as you did, using composition over inheritance...

https://github.com/apache/arrow/pull/5322

albertvillanova (approved):

theo-m (approved):
LGTM!

albertvillanova:
@lhoestq I think you forgot to add the new classes to the docs?

lhoestq:
Yes you're right, let me add them

PR #2030, opened 2021-03-11
Implement Dataset from text

albertvillanova:
Implement `Dataset.from_text`.

Analogue to #1943, #1946.


albertvillanova:
I am wondering why only one test of "keep_in_memory=True" fails, when there are many other tests that test the same and it happens only in pyarrow_1...

lhoestq (approved):
LGTM thanks :)

PR #2044, opened 2021-03-12
Add CBT dataset

gchhablani:
This PR adds the [CBT Dataset](https://arxiv.org/abs/1511.02301).

Note that I have also added the `raw` dataset as a separate configuration. I couldn't find a suitable "task" for it in YAML tags.

The dummy files have one example each, as the examples are slightly big. For `raw` dataset, I just used top few lines, because they are entire books and would take up a lot of space.

Let me know in case of any issues.


datasets/cbt/README.md
```diff
@@ -0,0 +1,334 @@
...
+
+- **Homepage:**[The bAbI project](https://research.fb.com/downloads/babi/)
+- **Repository:**
+- **Paper:** [arXiv Paper](https://arxiv.org/pdf/1511.02301.pdf)
+- **Leaderboard:**
+- **Point of Contact:** [Felix Hill](mailto:felix.hill@cl.cam.ac.uk) or [Antoine Bordes](mailto:abordes@fb.com).
+
+### Dataset Summary
+
+[More Information Needed]
```
lhoestq, line 91: Can you add a high level description of the dataset here ?

Can you also explain the differences between the different configurations ?

datasets/cbt/cbt.py
```diff
@@ -0,0 +1,335 @@
...
+            name="NE",
+            version=VERSION,
+            description="This part of my dataset covers the named entity answer CBT dataset",
+        ),
+        datasets.BuilderConfig(
+            name="CN", version=VERSION, description="This part of my dataset covers the common noun answer CBT dataset"
+        ),
+    ]
+
+    DEFAULT_CONFIG_NAME = "V"  # It's not mandatory to have a default configuration. Just use one if it make sense.
```
lhoestq, line 225: Why have the Verb config by default ?

datasets/cbt/cbt.py
```diff
@@ -0,0 +1,335 @@
...
+    def _generate_examples(
+        self, filepath  # method parameters are unpacked from `gen_kwargs` as given in `_split_generators`
+    ):
+        """ Yields examples as (key, example) tuples. """
+        # This method handles input defined in _split_generators to yield (key, example) tuples from the dataset.
+        # The `key` is here for legacy reason (tfds) and is not important in itself.
+
+        if self.config.name != "raw":
+            with open(filepath, encoding="utf-8") as f:
+                data = f.readlines()
```
lhoestq, line 293: Can you iter over the lines instead of reading everything at once ?
This should save memory

datasets/cbt/cbt.py
```diff
@@ -0,0 +1,335 @@
...
+                        example_idx += 1
+                    else:
+                        if len(line.split()[0]) == 1:
+                            sentences.append(line[1:].strip())
+                        else:
+                            sentences.append(line[2:].strip())
+                            # Text might contain double spaces.
+        else:
+            with open(filepath, encoding="utf=8") as f:
+                data = f.read()
```
lhoestq, line 325: same here

lhoestq (commented):
Awesome, thanks for adding this one :)

a few comments:

datasets/cbt/cbt.py
gchhablani: Here I need to search over the entire data for `_BOOK_TITLE_`, it's not possible to do this without reading entire data.

The alternative is to look for lines and check the first few characters for `_BOOK_TITLE_`. So, I'll do that.

datasets/cbt/README.md
```diff
@@ -0,0 +1,334 @@
...
+
+### Licensing Information
+
+The license comes with the compressed file containing the dataset:
+```
+GNU Free Documentation License
+
+Version 1.3, 3 November 2008
```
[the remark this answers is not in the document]
gchhablani: Okay, I'm also making the same changes to `.py` file.

gchhablani:
Hi @lhoestq,

I have added changes from the review.

lhoestq (approved):
Thank you @gchhablani ! :)

gchhablani:
Thanks for approving @lhoestq

PR #2049, opened 2021-03-13
Fix text-classification tags

gchhablani:
There are different tags for text classification right now: `text-classification` and `text_classification`:
image.

This PR fixes it.


julien-c:
LGTM, thanks for fixing.

julien-c (approved):

lhoestq (approved):
Thanks !

PR #2057, opened 2021-03-15
update link to ZEST dataset

matt-peters:
Updating the link as the original one is no longer working.


lhoestq (approved):
Thanks for fixing the URL !

I also updated the metadata in the dataset_infos.json file

will merge once the CI is all green :)

PR #2069, opened 2021-03-17
Add and fix docstring for NamedSplit

albertvillanova:
Add and fix docstring for `NamedSplit`, which was missing.


albertvillanova:
Maybe we should add some other split classes?

lhoestq (approved):
thanks !

PR #2072, opened 2021-03-17
Fix docstring issues

albertvillanova:
Fix docstring issues.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -104,9 +104,9 @@ html_static_path = ['_static']
 #
 # html_sidebars = {}
 
-# This must be the name of an image file (path relative to the configuration 
-# directory) that is the favicon of the docs. Modern browsers use this as 
-# the icon for tabs, windows and bookmarks. It should be a Windows-style 
+# This must be the name of an image file (path relative to the configuration
+# directory) that is the favicon of the docs. Modern browsers use this as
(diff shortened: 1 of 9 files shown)


docs/source/conf.py
```diff
@@ -190,3 +190,5 @@ def setup(app):
     app.add_js_file('js/custom.js')
 
 # -- Extension configuration -------------------------------------------------
+# Napoleon settings
+napoleon_use_ivar = True
```
lhoestq, line 194: What does this do exactly ?

lhoestq (approved):
Nice thanks !

docs/source/conf.py
albertvillanova: This tunes the rendering of "Attributes" list (see for example class: DatasetInfo).

Default value is `False`, and in that case each attribute is rendered with a format analog to class methods.

Otherwise, if set to `True`, Attributes are rendered with the same format as Args (Parameters).

lhoestq: Cool !

albertvillanova:
I think I will stop pushing to this PR, so that it can me merged for today release. 

I will open another PR for further fixing docs.

Do you agree, @lhoestq ?

lhoestq:
Sounds good thanks !

PR #2079, opened 2021-03-18
Refactorize Metric.compute signature to force keyword arguments only

albertvillanova:
Minor refactoring of Metric.compute signature to force the use of keyword arguments, by using the single star syntax.


lhoestq (approved):
Thanks ! this is consistent with `add` and `add_batch` now :)

PR #2081, opened 2021-03-18
Fix docstrings issues

albertvillanova:
Fix docstring issues.


src/datasets/arrow_dataset.py
```diff
@@ -1308,41 +1324,45 @@ def map(
         and update the table (if function does updated examples).
 
         Args:
-            function (`callable`): with one of the following signature:
+            function (`callable`): Function with one of the following signatures:
```
lhoestq, line 1327: shouldn't we use double backticks here ? same for the other arguments: \`\`callable\`\`, \`\`bool\`\`, etc.

lhoestq (commented):
cool thank you :)

src/datasets/arrow_dataset.py
albertvillanova: We should make a decision, because now different conventions are used.

In other open source projects, the convention is:
- When listing the Parameters, the types are specified just bare (not set as in-line)
  ```python
  Args:
      a (int): First param.
      b (str): Second param.
      c (DatasetInfo): Third param.
  ```
- Only in the descriptions and examples, this are displayed as in-line: name of a parameter, Python code, the name of a module, function, built-in, type, literal, class, method.

albertvillanova: Another option might be not including by hand the types in the docstring, and let Sphinx insert automatically in the docstring the type hints from the signature ;)

albertvillanova: @lhoestq, however in the latter, the links to the corresponding classes (if they exist) are not created.

Once we make some decision, we could document it for contributors ;)

lhoestq: We should do the same as in `transformers` using `:obj:<Type>`

Here that would be
```
function (:obj:`Callable`): Function with one of the following signatures:
```

lhoestq: If there's a default value, the template is
```
arg_name (:obj:`<Type>`, `optional`, defaults to :obj:`<default_value>`):
```

albertvillanova: @lhoestq therefore I think I will have to open this discussion to `transformers` members, because honestly I do not understand why they use the `:obj:` role in those cases. 

The purpose of roles is simply to create a cross-reference link to the corresponding description directive. In our docs, we have not defined any description directive for `Callable`, `str`, `int`,... Therefore, why using a role to try to create a cross-reference link to a non-existing description directive?

And second, the purpose of role `obj` is simply to create a cross-reference link to _an object of unspecified type_. Why using it for default values (literals)? What are we trying to link?

albertvillanova: Also note that this is not the convention used in other Open Source projects, like TensorFlow or PyArrow.

Also note that we are injecting all PyArrow Table docstrings (with their standard convention) in our Table(s).

thomwolf: @LysandreJik and @sgugger made a docupmentation and docstring writing guide some time ago: https://github.com/huggingface/transformers/tree/master/docs#writing-documentation---specification

Maybe you can find some answers there?

albertvillanova: Thanks @thomwolf for pointing out this guide.

I would really like to discuss with the authors @LysandreJik and @sgugger what is the reason for what I consider abusing of the usage of the role `:obj:`:
- When you write:
  ```
  :obj:`str`
  ```
  you are telling Sphinx: try to create a cross-reference link to the definition (the corresponding description directive) of `str` in our docs. But we not define `str` in our docs; that is defined in the Python Standard Library docs. Therefore it is impossible to create a link to it.
- When you write:
  ```
  :obj:`(batch_size, sequence_length)`
  ```
  you are telling to Sphinx: try to create a cross-reference link to the definition of `(batch_size, sequence_length)`. Well, I think this clearly makes no sense.

Moreover, I think it is important to note that I have not seen this usage of `:obj:` before in other Open Source projects, like TensorFlow, TensorFlow Datasets, Keras, PyArrow,...

sgugger: Yes using `:obj:` tells Sphinx to try to cross-reference the object inside but it will fail graciously if it does not find the reference **and** apply proper code formatting to what is inside, which is mostly why we use it. Leaving a type between backsticks will have it be rendered in italics (since Sphinx delegates it to restructured text which interprets it that way) which is something we did not want for the documentation of Transformers, as it is incorrect.

Even using double-backsticks does not apply the same formatting.

Note that this decision has been taken a long time ago (even before I joined) and it is very unlikely we will change the way it's done on the Tranformers side, as it would require changing 204k lines of codes (not all of them are docstrings but you get the idea).

albertvillanova: Thanks for your clear explanation, @sgugger.

I am just wondering why double-backticks do not apply the same formatting...

sgugger: Not sure either. It sometimes ends up as code-formatting + bold (the same as with an `:obj:`), sometimes just code-formatting without the bold.

lhoestq (approved):
LGTM thanks !

PR #2088, opened 2021-03-20
change bibtex template to author instead of authors

PhilipMay:
Hi,
IMO when using BibTex Author should be used instead of Authors.
See here: http://www.bibtex.org/Using/de/

Thanks
Philip


PhilipMay:
Trailing whitespace was removed. So more changes in diff than just this fix.

lhoestq (approved):
thank you :)

PR #2091, opened 2021-03-20
Fix copy snippet in docs

mariosasko:
With this change the lines starting with `...` in the code blocks can be properly copied to clipboard.


lhoestq (approved):
good catch ! thanks

PR #2100, opened 2021-03-23
Fix deprecated warning message and docstring

albertvillanova:
Fix deprecated warnings:
- Use deprecated Sphinx directive in docstring
- Fix format of deprecated message
- Raise FutureWarning

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -749,14 +749,16 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
 
         return self._data.column(column).unique().to_pylist()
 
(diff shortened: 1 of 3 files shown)


lhoestq:
`dictionary_encode_column_ ` should be deprecated since it never worked correctly. It will be removed in a major release.
This has to be deprecated in `DatasetDict` as well.
And `Dataset.dictionary_encode_column` doesn't exist indeed.

albertvillanova:
Thanks @lhoestq. I have fixed deprecated for `dictionary_encode_column_`.

lhoestq (approved):
Thanks :)

PR #2101, opened 2021-03-23
MIAM dataset - new citation details

eusip:
Hi @lhoestq, I have updated the citations to reference an OpenReview preprint.


lhoestq:
Hi !
Looks like there's a unicode error in the new citation in the miam.py file.
Could you try to fix it ? Not sure from which character it comes from though

You can test if it works on your side with
```
RUN_SLOW=1 pytest tests/test_dataset_common.py::LocalDatasetTest::test_load_dataset_all_configs_miam
```

eusip:
Unicode error resolved!

lhoestq (approved):

PR #2102, opened 2021-03-23
Move Dataset.to_csv to csv module

albertvillanova:
Move the implementation of `Dataset.to_csv` to module `datasets.io.csv`.


lhoestq (approved):
Good idea thanks !

PR #2107, opened 2021-03-24
Metadata validation

theo-m:
- `pydantic` metadata schema with dedicated validators against our taxonomy
- ci script to validate new changes against this schema and start a vertuous loop
- soft validation on tasks ids since we expect the taxonomy to undergo some changes in the near future

for reference with the current validation we have ~365~ 378 datasets with invalid metadata! full error report [_here_.](https://gist.github.com/theo-m/61b3c0c47fc6121d08d3174bd4c2a26b)


src/datasets/utils/resources/creators.json
```diff
@@ -0,0 +1,17 @@
+{
```
theo-m, line 1: these files were brought from https://github.com/huggingface/datasets-tagging

src/datasets/utils/metadata.py
```diff
@@ -0,0 +1,188 @@
...
+from pathlib import Path
+from typing import Any, Callable, Dict, List, Optional, Tuple
+
+
+# loading package files: https://stackoverflow.com/a/20885799
+try:
+    import importlib.resources as pkg_resources
+except ImportError:
+    # Try backported to PY<37 `importlib_resources`.
+    import importlib_resources as pkg_resources
```
lhoestq, line 12: We should add `importlib_resources` to the test dependencies for python<3.7 then ?

This makes me also think we should have a CI for python 3.6

src/datasets/utils/resources/creators.json
lhoestq: Is it possible to have these json files in a `resources` folder at the root of the repo or something like that ?
So that it's easier to find them

Is that compatible with how importlib.resources works ? Or do they have to be inside the `src/datasets` directory ?

src/datasets/utils/metadata.py
```diff
@@ -0,0 +1,188 @@
...
+    "10B<n<100B",
+    "100B<n<1T",
+    "n>1T",
+]
+known_multilingualities = {
+    "monolingual": "contains a single language",
+    "multilingual": "contains multiple languages",
+    "translation": "contains translated or aligned text",
+    "other": "other type of language distribution",
+}
```
lhoestq, line 53: would be nice to have json files for those as well

src/datasets/utils/metadata.py
```diff
@@ -0,0 +1,188 @@
...
+    def licenses_must_be_in_known_set(cls, licenses: List[str]):
+        others, to_validate = escape_validation_for_predicate(licenses, lambda e: "-other-" in e)
+        return [*tagset_validator(to_validate, list(known_licenses.keys()), "licenses", known_licenses_url), *others]
+
+    @validator("task_categories")
+    def task_category_must_be_in_known_set(cls, task_categories: List[str]):
+        # TODO: we're currently ignoring all values starting with 'other' as our task taxonomy is bound to change
+        #   in the near future and we don't want to waste energy in tagging against a moving taxonomy.
+        known_set = list(known_task_ids.keys())
+        others, to_validate = escape_validation_for_predicate(task_categories, lambda e: e.startswith("other"))
```
lhoestq, line 143: you can ignore any tag value containing "other" (not only at the beginning of the tag value)

For example: `text-classification-other-pokemon-classification`

(no, there's no dataset with this tag (yet))

lhoestq (commented):
This is really cool thanks :)

I left a few comments.

Also I was wondering this is really needed to have `utils.metadata` as a submodule of `datasets` ? This is only used by the CI so I'm not sure we should have this in the actual `datasets` package.

src/datasets/utils/metadata.py
theo-m: tbh I added this copy pasting from SO, I didn't look at the repo's policy for python versions support.

src/datasets/utils/metadata.py
theo-m: that's true for `tasks_ids` but not categories

src/datasets/utils/resources/creators.json
theo-m: I find that it being in the same directory as the schema definition is better, easier to find for lib devs and for the others the url to the directory would be enough. 
wdyt?

theo-m:
> Also I was wondering this is really needed to have `utils.metadata` as a submodule of `datasets` ? This is only used by the CI so I'm not sure we should have this in the actual `datasets` package.

I'm unclear on the suggestion, would you rather have a root-level `./metadata.py` file? I think it's well where it is, if anything we could move it out of utils and into `datasets` as it could be used by e.g. `DatasetDict` so that users can pull the metadata easily rather than have to reparse the readme.

lhoestq:
Ok that makes sense if we want to have functions that parse the metadata for users

src/datasets/utils/metadata.py
lhoestq: You just need to add
```python
"importlib_resources;python_version<'3.7'"
```
to the metadata validation requirements:
```python
    # metadata validation
    "langcodes[data]>=3.1.0",
    "pydantic>=1.8.1",
    "importlib_resources;python_version<'3.7'",
```

theo-m: ooh i understand now, i thought it was already in the stdlib but under a different name!

gchhablani:
Hi @theo-m @lhoestq 

This seems very interesting. Should I add the descriptions to the PR on `datasets-tagging`? Alternatively, I can also create a google-sheet/markdown table :)

Sorry for the delay in responding.

Thanks,
Gunjan

theo-m:
> Hi @theo-m @lhoestq
> 
> This seems very interesting. Should I add the descriptions to the PR on `datasets-tagging`? Alternatively, I can also create a google-sheet/markdown table :)
> 
> Sorry for the delay in responding.
> 
> Thanks,
> Gunjan

Hi @gchhablani, yes I think at the moment the best solution is for you to write in `datasets-tagging`, as the PR will allow us to discuss and review, even though the work will be ported to this repo in the end. 
Or we wait for this to be merged and you reopen the PR here, your call :)

src/datasets/utils/metadata.py
```diff
@@ -0,0 +1,172 @@
...
+
+if __name__ == "__main__":
+    from argparse import ArgumentParser
+
+    ap = ArgumentParser(usage="Validate the yaml metadata block of a README.md file.")
+    ap.add_argument("readme_filepath")
+    args = ap.parse_args()
+
+    readme_filepath = Path(args.readme_filepath)
+    DatasetMetadata.from_readme(readme_filepath)
```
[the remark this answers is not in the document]
theo-m: up to you, the idea was to keep a simple script that one could run to validate a readme in using `python -m datasets.utils.metadata myreadme.md`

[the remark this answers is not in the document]
lhoestq: We usually add scripts to the `datasets-cli`. This would be a nice addition though.
We can keep this here if it's considered experimental.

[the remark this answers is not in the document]
lhoestq: let's keep it here for now then

SBrandeis:
cc @abhi1thakur

src/datasets/utils/resources/languages.json
```diff
@@ -0,0 +1,792 @@
+{
```
lhoestq, line 1: Where did you get this data from ? (for future references and eventual updates)

lhoestq (approved):
Thank you @theo-m @SBrandeis !

src/datasets/utils/resources/languages.json
SBrandeis: I got the language codes from `datahub.io` (https://datahub.io/core/language-codes) and mapped the Language names manually with the `langcodes` package.
I have added a comment here: https://github.com/huggingface/datasets/pull/2107/files#diff-00db9a5082ea37ce6236856b9c01bdcca69facd06a0d9484141dd2c81c62f5f0R30-R32
Let me know if the statement is not clear 😄

lhoestq: Ok thanks :)

PR #2109, opened 2021-03-25
Add more issue templates and customize issue template chooser

albertvillanova:
When opening an issue, it is not evident for the users how to choose a blank issue template. There is a link at the bottom of all the other issue templates (`Don’t see your issue here? Open a blank issue.`), but this is not very visible for users. This is the reason why many users finally chose the `add-dataset` template instead (this is more visible) for issues that indeed are not requesting the addition of a new dataset.

~~With this PR, the default blank issue template would be as visible as the other templates (as the `add-dataset` template), thus making easier for the users to choose it.~~

With this PR:
- more issue templates, besides `add-dataset`, are added: `bug-report` and `feature-request`
- the issue template chooser is customized, so that it now includes a link to `Discussions` for questions


albertvillanova:
If you agree, I could also add a link to [Discussions](https://github.com/huggingface/datasets/discussions) in order to reinforce the use of Discussion to make Questions (instead of Issues).

I could also add some other templates: Bug, Feature Request,...

theo-m (commented):
Cool initiative, it was on my todo too :) 
What do you say we instead provide templates for common usecases, e.g. a `bug` template, a `feature idea` one, etc.?

albertvillanova:
@theo-m we wrote our same comments at the same time... 😉

.github/ISSUE_TEMPLATE/open-issue.md
```diff
@@ -0,0 +1,8 @@
+---
+name: Open Issue
+about: Open an Issue.
+title: ''
+labels: ''
+assignees: ''
+
+---
```
theo-m, line 8: I feel like we could remove this altogether. Arguably if they're not filing a bug, a request or a feature request, maybe it's best they go through discussions.

theo-m (approved):
lgtm

.github/ISSUE_TEMPLATE/open-issue.md
albertvillanova: @theo-m done

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,4 @@
+contact_links:
+  - name: Discussions
+    url: https://github.com/huggingface/datasets/discussions
+    about: Please ask and answer questions here, and engage with other community members
```
albertvillanova, line 4: Maybe adding additionally/instead a link to the Forum? https://discuss.huggingface.co/

albertvillanova: @theo-m @lhoestq do you know which is our preferred channel for Discussions? 
- https://github.com/huggingface/datasets/discussions
- https://discuss.huggingface.co/

Once this is clear, I will merge this PR :)

albertvillanova: @lhoestq maybe we should keep accepting questions as GH Issues? 

Otherwise, we should regularly keep track of the questions posted in the Discussion forums, while we are already doing this for GH Issues.

In that case, I would create a template for Questions...

lhoestq: We should definitely put the forum https://discuss.huggingface.co/ forward !

lhoestq (approved):
thanks :)

PR #2110, opened 2021-03-25
Fix incorrect assertion in builder.py

dreamgonfly:
Fix incorrect num_examples comparison assertion in builder.py

### DIFF
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -981,7 +981,6 @@ class GeneratorBasedBuilder(DatasetBuilder):
             finally:
                 num_examples, num_bytes = writer.finalize()
 
-        assert split_info.num_examples == num_examples, f"Expected to write {split_info.num_examples} but wrote {num_examples}"
         split_generator.split_info.num_examples = num_examples
         split_generator.split_info.num_bytes = num_bytes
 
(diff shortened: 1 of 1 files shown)


lhoestq:
Hi ! The SplitInfo is not always available. By default you would get `split_info.num_examples == 0`
So unfortunately we can't use this assertion you suggested

dreamgonfly:
> Hi ! The SplitInfo is not always available. By default you would get `split_info.num_examples == 0`
> So unfortunately we can't use this assertion you suggested

Then it would be better to just remove the assertion, because the existing assertion does nothing.

src/datasets/builder.py
```diff
@@ -981,7 +981,7 @@ def _prepare_split(self, split_generator):
             finally:
                 num_examples, num_bytes = writer.finalize()
 
-        assert num_examples == num_examples, f"Expected to write {split_info.num_examples} but wrote {num_examples}"
+        assert split_info.num_examples == num_examples, f"Expected to write {split_info.num_examples} but wrote {num_examples}"
```
dreamgonfly, line 984:
```suggestion

```

lhoestq (approved):
Thanks.
You can ignore the CI fail on the documentation, this has been fixed on master

PR #2111, opened 2021-03-25
Compute WER metric iteratively

### LINKED ISSUE #2078: MemoryError when computing WER metric
Hi, I'm trying to follow the ASR example to try Wav2Vec. This is the code that I use for WER calculation:

```
wer = load_metric("wer")
print(wer.compute(predictions=result["predicted"], references=result["target"]))
```

However, I receive the following exception:

`Traceback (most recent call last):
  File "/home/diego/IpGlobal/wav2vec/test_wav2vec.py", line 51, in <module>
    print(wer.compute(predictions=result["predicted"], references=result["target"]))
  File "/home/diego/miniconda3/envs/wav2vec3.6/lib/python3.6/site-packages/datasets/metric.py", line 403, in compute
    output = self._compute(predictions=predictions, references=references, **kwargs)
  File "/home/diego/.cache/huggingface/modules/datasets_modules/metrics/wer/73b2d32b723b7fb8f204d785c00980ae4d937f12a65466f8fdf78706e2951281/wer.py", line 94, in _compute
    return wer(references, predictions)
  File "/home/diego/miniconda3/envs/wav2vec3.6/lib/python3.6/site-packages/jiwer/measures.py", line 81, in wer
    truth, hypothesis, truth_transform, hypothesis_transform, **kwargs
  File "/home/diego/miniconda3/envs/wav2vec3.6/lib/python3.6/site-packages/jiwer/measures.py", line 192, in compute_measures
    H, S, D, I = _get_operation_counts(truth, hypothesis)
  File "/home/diego/miniconda3/envs/wav2vec3.6/lib/python3.6/site-packages/jiwer/measures.py", line 273, in _get_operation_counts
    editops = Levenshtein.editops(source_string, destination_string)
MemoryError`

My system has more than 10GB of available RAM. Looking at the code, I think that it could be related to the way jiwer does the calculation, as it is pasting all the sentences in a single string before calling Levenshtein editops function.

albertvillanova:
Compute WER metric iteratively to avoid MemoryError.

Fix #2078.

### DIFF
--- a/metrics/wer/wer.py
+++ b/metrics/wer/wer.py
@@ -53,10 +53,13 @@ performance of the ASR system with a WER of 0 being a perfect score.
 """
 
 _KWARGS_DESCRIPTION = """
-Computes WER score of transcribed segments against references.
+Compute WER score of transcribed segments against references.
+
 Args:
-    references: list of references for each speech input.
-    predictions: list of transcribtions to score.
+    references: List of references for each speech input.
+    predictions: List of transcriptions to score.
+    concatenate_texts (bool, default=False): Whether to concatenate all input texts or compute WER iteratively.
+
 Returns:
     (float): the word error rate
 
@@ -90,11 +93,14 @@ class WER(datasets.Metric):
             ],
         )
 
-    def _compute(self, predictions, references):
-        incorrect = 0
-        total = 0
-        for prediction, reference in zip(predictions, references):
-            measures = compute_measures(reference, prediction)
-            incorrect += measures["substitutions"] + measures["deletions"] + measures["insertions"]
-            total += measures["substitutions"] + measures["deletions"] + measures["hits"]
-        return incorrect / total
+    def _compute(self, predictions=None, references=None, concatenate_texts=False):
+        if concatenate_texts:
+            return compute_measures(references, predictions)["wer"]
+        else:
+            incorrect = 0
+            total = 0
+            for prediction, reference in zip(predictions, references):
+                measures = compute_measures(reference, prediction)
+                incorrect += measures["substitutions"] + measures["deletions"] + measures["insertions"]
+                total += measures["substitutions"] + measures["deletions"] + measures["hits"]
+            return incorrect / total
--- a/metrics/wer/wer.py
+++ b/metrics/wer/wer.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 """ Word Error Ratio (WER) metric. """
 
-from jiwer import wer
+from jiwer import compute_measures
 
 import datasets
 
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
LGTM thank you !

cc @patrickvonplaten any opinion on this ?

patrickvonplaten (approved):

metrics/wer/wer.py
```diff
@@ -91,4 +91,10 @@ def _info(self):
...
 
     def _compute(self, predictions, references):
-        return wer(references, predictions)
+        incorrect = 0
+        total = 0
+        for prediction, reference in zip(predictions, references):
+            measures = compute_measures(reference, prediction)
+            incorrect += measures["substitutions"] + measures["deletions"] + measures["insertions"]
+            total += measures["substitutions"] + measures["deletions"] + measures["hits"]
+        return incorrect / total
```
elgeish, line 100: Just for safety, you may want to handle the case of `total = 0` (e.g., empty set). Unit tests to verify the change would be great as well. Thanks for making this change!

albertvillanova: Thanks @elgeish for your comments.

The reasons why I have not explicitly handled the edge case `total = 0` are:
- `compute_measure` does already raise a ValueError exception if any of the `references` is the empty string: see https://github.com/jitsi/jiwer/blob/f8e5404e4ddb7259081191443def1b6670480244/jiwer/measures.py#L242
- if `references` and `predictions` have length = 0, the previous behavior was also raising the same ValueError; the current implementation will raise a ZeroDivisionError in this case

In relation to unit tests, currently we do not test scripts. Maybe @lhoestq can give more insight on this.

elgeish: Sounds good; makes sense!

albertvillanova:
@lhoestq @patrickvonplaten are you sure of the parameter name `concatenate_texts`? I was thinking about something like `iter`...

lhoestq:
Not sure about the name, if you can improve it feel free to do so ^^'

albertvillanova:
@lhoestq yes, but the end user does not necessarily know the details of the implementation of the WER computation.

From the end user perspective I think it might make more sense: how do you want to compute the metric?
- all in once, more RAM memory needed?
- iteratively, less RAM requirements?

Because of that I was thinking of something like `iter` or `iterative`...

lhoestq:
Personally like `concatenate_texts` better since I feel like `iter` or `iterate` are quite vague

albertvillanova:
Therefore, you can merge... ;)

lhoestq:
Ok ! merging :)

PR #2113, opened 2021-03-25
Implement Dataset as context manager

albertvillanova:
When used as context manager, it would be safely deleted if some exception is raised.

This will avoid 
> During handling of the above exception, another exception occurred:


lhoestq (approved):
thanks :)

finally ! readable errors on windows

src/datasets/arrow_dataset.py
```diff
@@ -485,6 +485,12 @@ def __del__(self):
         if hasattr(self, "_indices"):
             del self._indices
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.__del__()
```
lhoestq, line 492:
```suggestion
        # Here `del` is used to del the pyarrow tables. This properly closes the files used for memory mapped tables
        self.__del__()
```

PR #2126, opened 2021-03-28
Replace legacy torch.Tensor constructor with torch.tensor

mariosasko:
The title says it all (motivated by [this issue](https://github.com/pytorch/pytorch/issues/53146) in the pytorch repo).


lhoestq (approved):
Thanks :)

PR #2136, opened 2021-03-29
fix dialogue action slot name and value

### LINKED ISSUE #2128: Dialogue action slot name and value are reversed in MultiWoZ 2.2
Hi @yjernite, thank you for adding MultiWoZ 2.2 in the huggingface datasets platform. It is beneficial!

I spot an error that the order of Dialogue action slot names and values are reversed.

https://github.com/huggingface/datasets/blob/649b2c469779bc4221e1b6969aa2496d63eb5953/datasets/multi_woz_v22/multi_woz_v22.py#L251-L262

adamlin120:
fix #2128


lhoestq (approved):
Thank you !

PR #2143, opened 2021-03-30
task casting via load_dataset

theo-m:
wip
not satisfied with the API, it means as a dataset implementer I need to write a function with boilerplate and write classes for each `<dataset><task>` "facet".

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,8 +56,8 @@ import itertools
 import os
 import sys
 
-from setuptools import find_packages
-from setuptools import setup
+from setuptools import find_packages, setup
+
 
 DOCLINES = __doc__.split("\n")
 
@@ -89,6 +89,9 @@ REQUIRED_PKGS = [
     "fsspec",
     # To get datasets from the Datasets Hub on huggingface.co
     "huggingface_hub<0.1.0",
+    # pydantic allows us to express simply validation patterns for dataset metadata and serialize object schemas
+    # for tasks descriptions.
+    "pydantic",
 ]
 
 BENCHMARKS_REQUIRE = [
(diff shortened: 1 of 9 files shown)


setup.py
```diff
@@ -89,6 +89,9 @@
     "fsspec",
     # To get datasets from the Datasets Hub on huggingface.co
     "huggingface_hub<0.1.0",
+    # pydantic allows us to express simply validation patterns for dataset metadata and serialize object schemas
+    # for tasks descriptions.
+    "pydantic",
```
lhoestq, line 94: Serialization is also possible with dataclasses.
Validation is a good reason to have an extra dependency, but we usually don't like adding extra dependencies to avoid issues in the future (breaking changes, bugs, etc.).
Maybe we can have a simple function for validation, without requiring an extra dependency

src/datasets/dataset_dict.py
```diff
@@ -17,7 +17,10 @@
 from .utils.typing import PathLike
 
 
-class DatasetDict(dict):
+T = TypeVar("T")
+
+
+class DatasetDict(dict, MutableMapping[str, T]):
```
lhoestq, line 23: This is actually a MutableMapping[str, Dataset]

tests/test_dataset_common.py
```diff
@@ -231,6 +232,13 @@ def test_builder_configs(self, dataset_name):
         if builder_configs[0] is not None:
             all(self.assertTrue(isinstance(config, BuilderConfig)) for config in builder_configs)
 
+    def test_task_implementations(self, dataset_name):
+        if dataset_name != "rotten_tomatoes":
+            return
+
+        # this is failing at the moment since the new code for "rotten_tomatoes" is not uploaded.
+        load_dataset(dataset_name, as_task=tasks.ClassificationSingleLabelDataset.id)
```
lhoestq, line 240: it would be good to have a more general test than check for the dataset names.
Maybe by checking the available tasks for the builder instance ?

lhoestq (commented):
That's a good start thanks :)

I left a few comments
(most of them we've already discussed them, but I added hem anyway if people want to share their thoughts)

PR #2145, opened 2021-03-30
Implement Dataset add_column

### LINKED ISSUE #1954: add a new column
Hi
I'd need to add a new column to the dataset, I was wondering how this can be done? thanks 
@lhoestq

albertvillanova:
Implement `Dataset.add_column`.

Close #1954.


src/datasets/arrow_dataset.py
```diff
@@ -2460,6 +2460,18 @@ def to_pandas(
...
 
+    def add_column(self, column: dict):
+        """Add column to Dataset.
+
+        Args:
+            column (dict): Column data to be added.
+        """
+        column_table = InMemoryTable.from_pydict(column)
+        # Concatenate tables horizontally
+        self._data = ConcatenationTable.from_blocks([[self._data, column_table]])
```
lhoestq, line 2471: `self._data` may not be a valid table block (i.e. either a InMemoryTable or a MemoryMappedTable object).
For example if `self._data` is a `ConcatenationTable`, this won't work.

Maybe we can use another `ConcatenationTable` constructor for this ?
Like for example a version of `ConcatenationTable.from_tables` but for axis=1 ?
Under the hood this uses `from_blocks` anyway, but it allows any kind of tables as input.

src/datasets/arrow_dataset.py
```diff
@@ -2460,6 +2460,18 @@ def to_pandas(
                 for offset in range(0, len(self), batch_size)
             )
 
+    def add_column(self, column: dict):
```
lhoestq, line 2463: A dict may have several columns, this may be confusing.
I would expect `column` to be a `List` or a `np.ndarray`.
What do you think ?

lhoestq (commented):
This feature has been asked many times :)
This is awesome to finally have it ! Thanks for adding it !

I added some suggestions:

src/datasets/arrow_dataset.py
albertvillanova: But we also need a "name" for the column. I was thinking of an input of the form: `{"column_name": column_data}`, where `column name` is a `str`, and  `column_data` is either a `List` or `np.ndarray`.

src/datasets/arrow_dataset.py
albertvillanova: I am going to finish this PR first: #2151 😉

src/datasets/arrow_dataset.py
albertvillanova: @lhoestq, other option could be to pass 2 parameters: one for the column name and other for the column data.

src/datasets/arrow_dataset.py
```diff
@@ -2654,6 +2654,26 @@ def to_pandas(
                 for offset in range(0, len(self), batch_size)
             )
 
+    def add_column(self, name: str, column: Union[list, np.array]):
```
lhoestq, line 2657:
```suggestion
    @transmit_format
    @fingerprint_transform(inplace=False)
    def add_column(self, name: str, column: Union[list, np.array], new_fingerprint: str):
```

The format of the dataset should be transmitted to the new dataset.
and the fingerprint updated accordingly

src/datasets/arrow_dataset.py
```diff
@@ -2654,6 +2654,26 @@ def to_pandas(
...
+        """Add column to Dataset.
+
+        .. versionadded:: 1.6
```
lhoestq, line 2660:
```suggestion
        .. versionadded:: 1.7
```

tests/test_arrow_dataset.py
```diff
@@ -1948,6 +1948,27 @@ def test_concatenate_datasets_duplicate_columns(dataset):
     assert "duplicated" in str(excinfo.value)
 
 
+@pytest.mark.parametrize("in_memory", [False, True])
+def test_dataset_add_column(in_memory, dataset_dict, arrow_path):
```
lhoestq, line 1952: If you want to test the cases where a dataset is formatted / has an indices mapping / has classlabels, feel free to check out how I did it in #2254 for the `add_item` test

lhoestq (commented):
Cool ! Added a few comments :)

src/datasets/arrow_dataset.py
```diff
@@ -2654,6 +2654,26 @@ def to_pandas(
...
+        Returns:
+            :class:`Dataset`
+        """
+        column_table = InMemoryTable.from_pydict({name: column})
+        # Concatenate tables horizontally
+        table = ConcatenationTable.from_tables([self._data, column_table], axis=1)
+        # Update features
+        info = self.info
+        info.features = Features.from_arrow_schema(table.schema)
+        return Dataset(table, info=info)
```
mariosasko, line 2675: @albertvillanova Metadata needs to be updated since this function modifies the features of the dataset.

```suggestion
        table = update_metadata_with_features(table, info.features)
        return Dataset(table, info=info)
```

albertvillanova: I have finally reused
```python
info = copy.deepcopy(self.info)
```
as suggested by @lhoestq.

mariosasko: But this line doesn't actually update the metadata. For instance, this can be verified with:
```python
>>> sst = load_dataset("sst", split="train")
>>> sst1 = sst.add_column("my_col", list(range(len(sst))))
>>> print(sst1._data.schema.metadata) # should contain "my_col", but it doesn't
OrderedDict([(b'huggingface', b'{"info": {"features": {"sentence": {"dtype": "string", "id": null, "_type": "Value"}, "label": {"dtype": "float32", "id": null, "_type": "Value"}, "tokens": {"dtype": "string", "id": null, "_type": "Value"}, "tree": {"dtype": "string", "id": null, "_type": "Value"}}}}')])
```

lhoestq: Metadata stored in the schema is just a redundant information regarding the feature types.
It is used when calling `Dataset.from_file` to know which feature types to use.
These metadata are stored in the schema of the pyarrow table by using `update_metadata_with_features`.
However this something that's almost never tested properly so we should find a way to always make sure that the metadata are synced with the actual feature types in self.info

In your case you just need to add this line @albertvillanova , thanks @mariosasko for pointing that out
```python
table = update_metadata_with_features(table, info.features)
```

albertvillanova: I have created issue #2271 in relation with this.

lhoestq (approved):
Looks all good to me :)

Once #2274 is merged we can update the test function to also check the metadata

lhoestq:
#2274 has been merged. You can now merge master into this branch and use `assert_arrow_metadata_are_synced_with_dataset_features(dset)` to make sure that the metadata are good :)

PR #2151, opened 2021-03-30
Add support for axis in concatenate datasets

### LINKED ISSUE #853: concatenate_datasets support axis=0 or 1 ？
I want to achieve the following result
image

albertvillanova:
Add support for `axis` (0 or 1) in `concatenate_datasets`.

Close #853.


albertvillanova:
@lhoestq I am going to implement the consolidation step you mentioned in #1870.

albertvillanova:
@lhoestq I was thinking that the order of the TableBlocks is not relevant, isn't it?

I mean, in order to consolidate _consecutive_ in-memory table blocks, in this case:
```
blocks = [in_memory_1, memory_mapped_2, in_memory_3]
```
I could reorder the list:
```
blocks = [in_memory_1, in_memory_3, memory_mapped_2]
```
so that the first 2 can be consolidated into a single one:
```
blocks = [in_memory_4, memory_mapped_2]
```

lhoestq:
I think the order is important, users won't expect the dataset to be "shuffled" when they add a new item

src/datasets/table.py
```diff
@@ -521,30 +528,48 @@ def _concat_blocks_horizontally_and_vertically(blocks: List[List[TableBlock]]) -
             tables_to_concat_vertically.append(combined_table)
         return pa.concat_tables(tables_to_concat_vertically)
 
+    @classmethod
+    def _consolidate_blocks(cls, blocks, axis: Optional[int] = None):
+        if axis is not None:
+            consolidated_blocks = []
+            for is_in_memory, block_group in groupby(blocks, key=lambda x: isinstance(x, InMemoryTable)):
```
lhoestq, line 535: `groupby` won't take the order of the blocks into account.
While we don't care of the order for axis=1, we actually do want to keep the right order for axis 0.

That means that when adding a new blocks in axis 0, we just need to check whether both the new block and the last block are InMemoryTable objects and merge them.

lhoestq (commented):
Thanks ! I just added a comment about the `groupby` step

src/datasets/table.py
albertvillanova: But `groupby` does keep the order of the blocks...

albertvillanova:
```ipython
In [4]: for k, v in groupby([1, 2, 2, 1, 1, 1]):
   ...:     print(f"{k}: {list(v)}")
   ...:
1: [1]
2: [2, 2]
1: [1, 1, 1]

```

albertvillanova:
> I think the order is important, users won't expect the dataset to be "shuffled" when they add a new item

OK, therefore I leave `_consolidate_blocks` as it is, which currently keeps the order of the blocks (no shuffling).

src/datasets/table.py
lhoestq: Oh right ! That's perfect then :)

src/datasets/table.py
```diff
@@ -558,53 +562,89 @@ def __setstate__(self, state):
...
+    def from_tables(cls, tables: List[Union[pa.Table, Table]], axis: int = 0) -> "ConcatenationTable":
         blocks = []
         for table in tables:
             if isinstance(table, pa.Table):
-                blocks.append([InMemoryTable(table)])
+                blocks.append([InMemoryTable(table)] if axis == 0 else InMemoryTable(table))
             elif isinstance(table, ConcatenationTable):
-                blocks.extend(copy.deepcopy(table.blocks))
+                blocks.extend(copy.deepcopy(table.blocks) if axis == 0 else copy.deepcopy(table.blocks)[0])  # TODO
             else:
```
albertvillanova, line 645: @lhoestq here we have a potential issue if we want to concatenate horizontally a block `A` with a ConcatenationTable with multiple block rows `[[B], [C]]`. The resulting blocks should be something like `[[A_1, B], [A_2, C]]`.

For the moment I only concatenate horizontally to the first block row.

lhoestq: Blocks of a single row should have the same length or the column concatenation fails.
So ideally we should end up with something like `[[A_1, B], [A_2, C]]` as you mentioned.
Feel free to construct `A_1` and `A_2` by calling `A.slice(offset, length)`.

albertvillanova: Done.

lhoestq (approved):
This is ready now

I fixed some tests and I also improved the `from_tables` test

lucaguarro:
Thank you guys for implementing this. Minor thing I noticed in the [documentation](https://huggingface.co/docs/datasets/package_reference/main_classes.html#datasets.concatenate_datasets): it says "Converts a list of Dataset with **the same schema** into a single Dataset". With the addition of the axis parameter, perhaps this should be reworded, no?

PR #2154, opened 2021-03-31
Adding the NorNE dataset for Norwegian POS and NER

versae:
NorNE is a manually annotated corpus of named entities which extends the annotation of the existing Norwegian Dependency Treebank. Comprising both of the official standards of written Norwegian (Bokmål and Nynorsk), the corpus contains around 600,000 tokens and annotates a rich set of entity types including persons, organizations, locations, geo-political entities, products, and events, in addition to a class corresponding to nominals derived from names.

See #1720.


lhoestq (approved):
Perfect thank you !
Good job :)

versae:
Awesome!

PR #2155, opened 2021-03-31
Add table classes to the documentation

lhoestq:
Following #2025 , I added the table classes to the documentation

cc @albertvillanova


albertvillanova:
Just note that docstrings injected from PyArrow do not follow the same convention for formatting types in `Args` or `Returns` as we do... Not a big problem, anyway! 😄

PR #2157, opened 2021-03-31
updated user permissions based on umask

bhavitvyamalik:
Updated user permissions based on running user's umask (#2065). Let me know if `0o666` is looking good or should I change it to `~umask` only (to give execute permissions as well)

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -1645,7 +1645,9 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
         if update_data and tmp_file is not None:
             tmp_file.close()
             shutil.move(tmp_file.name, cache_file_name)
-            os.chmod(cache_file_name, 0o644)
+            umask = os.umask(0)
+            os.chmod(cache_file_name, 0o666 & ~umask)
+
 
         if update_data:
             # Create new Dataset from buffer or file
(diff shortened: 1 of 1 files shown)


src/datasets/arrow_dataset.py
```diff
@@ -1645,7 +1645,8 @@ def init_buffer_and_writer():
         if update_data and tmp_file is not None:
             tmp_file.close()
             shutil.move(tmp_file.name, cache_file_name)
-            os.chmod(cache_file_name, 0o644)
+            umask = os.umask(0)
```
lhoestq, line 1648: While getting the umask value, you are also setting the os.umask to zero here.

First, it appears that setting the value to zero might a [security issue](https://stackoverflow.com/a/44130549) so please use `0o666` instead.

Then, you can set it back to the original value.

```suggestion
            umask = os.umask(0o666)
            os.umask(umask)
```

src/datasets/arrow_dataset.py
```diff
@@ -1884,7 +1885,8 @@ def select(
         if tmp_file is not None:
             tmp_file.close()
             shutil.move(tmp_file.name, indices_cache_file_name)
-            os.chmod(indices_cache_file_name, 0o644)
+            umask = os.umask(0)
```
lhoestq, line 1888:
```suggestion
            umask = os.umask(0o666)
            os.umask(umask)
```
same here

lhoestq (commented):
Thank you !
I noticed that you didn't set the umask back to its original value and added a comment about that. I also mentioned a potential security issue that we can avoid

src/datasets/arrow_dataset.py
bhavitvyamalik: TIL ^^

lhoestq (approved):
Thanks ! LGTM :)

PR #2164, opened 2021-04-03
Replace assertTrue(isinstance with assertIsInstance in tests

mariosasko:
Replaces all the occurrences of the `assertTrue(isinstance(` pattern with `assertIsInstance`.


lhoestq (approved):
thanks :)

PR #2168, opened 2021-04-04
Preserve split type when realoding dataset

### LINKED ISSUE #2167: Split type not preserved when reloading the dataset
A minimal reproducible example:
```python
>>> from datasets import load_dataset, Dataset
>>> dset = load_dataset("sst", split="train")
>>> dset.save_to_disk("sst")
>>> type(dset.split)
<class 'datasets.splits.NamedSplit'>
>>> dset = Dataset.load_from_disk("sst")
>>> type(dset.split)  # NamedSplit expected
<class 'str'>
```

It seems like this bug was introduced in #2025.

mariosasko:
Fixes #2167 

Using `eval` is not ideal for security reasons (in web apps I assume), but without it the code would be much more complex IMO.

In terms of style, instead of explicitly importing a private member (`_RelativeInstruction`), we can add these imports at the top of the module:
```python
from . import arrow_reader  # gives us access to ReadInstruction and _RelativeInstruction
from . import splits # gives us access to NamedSplit
```

and then define the `eval` globals as follows:
```python
{**arrow_reader.__dict__, **splits.__dict__}
```


lhoestq:
Thanks for diving into this !

Before going further, I just want to make sure if using `eval` is the right solution
Personally I'm not a big fan of `eval` since it has many security concerns. Also storing string representations of python objects in the json files is not ideal either IMO, so maybe it's possible to change this aspect instead.

Maybe it would be better to convert the `_RelativeInstruction` to a string (or "specs") ?
It looks like `ReadInstruction.from_spec` already exists, but not the other way around.
The specs are the string representation of instructions. For example: `train+validation[:50%]`.

Let me know what you think ! And thanks again, this issue has been here for a while now ^^

mariosasko:
@lhoestq Yes, before going with `eval`, I thought about this approach with the "spec". The only issue with this approach is that we have to come up with a represenation for the `rounding` arg.

What do you think about this (maybe too verbose)?
```python
>>> print(ReadInstruction("train", rounding="pct1_dropremainder", from_=10, to=30).to_spec())
train[10:30](pct1_dropremainder)

**Comment by lhoestq** [comment]
Good idea !

First we must note that the rounding is only used for percentage instructions.
For absolute instructions there's no rounding ambiguity.

By default the rounding is set to `closest`. For example if you have a train set of 999 examples and if you provide an instruction spec `"train[:1%]"`, you're going to get the first ten examples (while the `pct1_dropremainder ` rounding would return 9 examples).

Currently there's no way to get an instruction with a `pct1_dropremainder` rounding strategy from an instruction spec.
So we can either drop the support of `pct1_dropremainder` or define a way to use this strategy from a spec.
I don't think dropping `pct1_dropremainder` would be a good idea since it allows to load each percent to all have the same number of examples (even the last one). Therefore I think your suggestion makes total sense and we should add a representation of this rounding strategy.

I like what you suggested `train[10%:30%](pct1_dropremainder)` is fine, and it seems compatible with the regex that parses the instructions specs.

**Comment by mariosasko** [comment]
@lhoestq I've made the changes as you suggested. Ready for the review.

File: src/datasets/arrow_dataset.py [hunk 1]
```diff
@@ -577,17 +577,22 @@ def save_to_disk(self, dataset_path: str, fs=None):
...
                 "_format_type",
                 "_indexes",
                 "_output_all_columns",
-                "_split",
             ]
         }
+
+        split = self.__dict__["_split"]
+        state["_split"] = {"spec": str(split) if split is not None else split}
+        if isinstance(split, (NamedSplit, ReadInstruction)):
```
lhoestq, line 593: According to the typing, `split` should be either None or a NamedSplit, not a ReadInstruction.
Should we fix the typing as well ?

Another option I have in mind is to always use NamedSplit for the `_split` attribute. I think we would just need to use `to_spec` at this line:

https://github.com/huggingface/datasets/blob/20ffe98d1da381f28464684a62c8ed3a499e10aa/src/datasets/arrow_reader.py#L238-L238

`original_instruction` could be a string, a NamedSplit or a ReadInstruction (damn this function lacks typing -_-)

I have a small preference for this last option because that simplify things and make the serialization a simple string instead of `{"spec": ..., "type": ...}`. Currently this is a string that is used, so that we could avoid breaking backward compatibility this way.

Let me know what you think !

lhoestq (commented):
This is awesome thanks !

Could you also add a test in `test_arrow_reader.py` to make sure the conversion to spec works as expected ?

Also I added a comment about the possibility about always using `NamedSplit`. Let me know what you think, and if I can help on this

src/datasets/arrow_dataset.py
mariosasko: Yes, I agree this "spec + type" thing is not ideal and that the `_split` attribute should either be None or a NamedSplit to get a simpler solution.

mariosasko:
@lhoestq I've added a test and addressed the comments.

Additionally, `ReadInstruction` is converted to its spec form in `builder.py` to avoid a circular import that would happen if this logic was in `arrow_reader.py`.  If you think it's better to have this logic in `arrow_reader.py`, the import can be delayed by putting it inside a function.

lhoestq (approved):
Looks all good now ! I just moved the instruction to split conversion from the builder to the reader

Thank you !

PR #2174, opened 2021-04-06
Pin docutils for better doc

sgugger:
The latest release of docutils make the navbar in the documentation weird and the Markdown wrongly interpreted:

image

We had the same problem in Transformers and solved it by pinning docutils (a dep of sphinx).

You can see the version after the change [here](https://32769-250213286-gh.circle-artifacts.com/0/docs/_build/html/index.html).


lhoestq (approved):
Thanks !

PR #2183, opened 2021-04-07
Fix s3fs tests for py36 and py37+

lhoestq:
Recently several changes happened:
1. latest versions of `fsspec` require python>3.7 for async features
2. `s3fs` added a dependency on `aiobotocore`, which is not compatible with the `moto` s3 mock context manager

This PR fixes both issues, by pinning `fsspec` and `s3fs` for python 3.6, and by using `moto` in server mode to support running the tests on python>=3.7 with the latest version of `fsspec` and `s3fs`.

cc @philschmid


setup.py
```diff
@@ -172,7 +181,8 @@
     "s3": [
         "fsspec[s3]",
```
philschmid, line 182:
```suggestion
        _deps["fsspec"],
```
to have compatible versions, since `fsspec[s3]` is installing no pinned `s3fs` version. https://github.com/intake/filesystem_spec/blob/bb7c23305ce76cb22cd10305f87a49009f363a11/setup.py#L50

setup.py
```diff
@@ -185,6 +195,7 @@
         "sphinx-rtd-theme==0.4.3",
         "sphinx-copybutton",
         "fsspec[s3]",
```
philschmid, line 197:
```suggestion
        _deps["fsspec"],
```
to have compatible versions, since `fsspec[s3]` is installing no pinned `s3fs` version. https://github.com/intake/filesystem_spec/blob/bb7c23305ce76cb22cd10305f87a49009f363a11/setup.py#L50

philschmid (approved):
LGTM!! Thank you for taking this. Might be interesting if we can profit from this async features they added.

philschmid (approved):

PR #2191, opened 2021-04-08
Refactorize tests to use Dataset as context manager

albertvillanova:
Refactorize Dataset tests to use Dataset as context manager.


lhoestq (approved):
Thanks !

This makes things a bit harder to read sometimes but this is worth it since it makes the errors on windows way more readable.

Maybe in the future we could use pytest instead, and use a fixture that properly closes the datasets created during each test ?

albertvillanova:
I find very interesting that idea of using a fixture instead!

Let me rework a little bit this PR, @lhoestq.

albertvillanova:
@lhoestq, as this is a big refactoring, I had many problems to solve the conflicts with the master branch...

Therefore, I thinks it is better to merge this as it is, and then to make other PRs with additional refactorings, before I get conflicts again with the master branch...

lhoestq:
There are still some conflicts that prevent merging.
Moreover I noticed that you added one fixture per method of the Dataset object to be mocked. The code of all these fixtures is pretty much the same, feel free to factorize them into one fixture.

Also feel free to create another branch from `master` if you don't want to fix the conflicts of this branch.
Let me know if I can help you on this

albertvillanova:
@lhoestq, yes, the new conflicts appeared after today merge commits on master...

I am definitely going to split this PR into smaller ones in order to avoid having to resolve many conflicts after each commit on master.

PR #2192, opened 2021-04-08
Fix typo in huggingface hub

LysandreJik:
pip knows how to resolve to `huggingface_hub`, but conda doesn't!

The `packaging` dependency is also required for the build to complete.


lhoestq (approved):
Thanks :)

PR #2198, opened 2021-04-09
added file_permission in load_dataset

bhavitvyamalik:
As discussed in #2065 I've added `file_permission` argument in `load_dataset`. 

Added mainly 2 things here:
1) Permission of downloaded datasets when converted to .arrow files  can be changed with argument `file_permission` argument in `load_dataset` (default is 0o644 only)
2) Incase the user uses `map` later on to generate another cache file of dataset, it ensures the permissions of newly generated file are similar to that of` *-train.arrow` file inside cache_dir for that dataset.


lhoestq:
From offline discussions: we want to make the permissions handling consistent with `transformers`. However from discussion  in https://github.com/huggingface/transformers/pull/11119 it looks like it might not be a good solution to provide this argument. Users should use umask for now, and we'll see how things evolve.

@bhavitvyamalik I'm closing the PR for now if you don't mind

Issue #2195, opened 2021-04-09
KeyError: '_indices_files' in `arrow_dataset.py`


samsontmr:
After pulling the latest master, I'm getting a crash when `load_from_disk` tries to load my local dataset.

Trace:
```
Traceback (most recent call last):
  File "load_data.py", line 11, in <module>
    dataset = load_from_disk(SRC)
  File "/opt/conda/envs/py38/lib/python3.8/site-packages/datasets/load.py", line 784, in load_from_disk
    return DatasetDict.load_from_disk(dataset_path, fs, keep_in_memory=keep_in_memory)
  File "/opt/conda/envs/py38/lib/python3.8/site-packages/datasets/dataset_dict.py", line 692, in load_from_disk
    dataset_dict[k] = Dataset.load_from_disk(dataset_dict_split_path, fs, keep_in_memory=keep_in_memory)
  File "/opt/conda/envs/py38/lib/python3.8/site-packages/datasets/arrow_dataset.py", line 634, in load_from_disk
    if state["_indices_files"]:
KeyError: '_indices_files'
```

I believe this is the line causing the error since there may not be a `_indices_files` key in the older versions:
https://github.com/huggingface/datasets/blob/b70141e3c5149430951773aaa0155555c5fb3e76/src/datasets/arrow_dataset.py#L634

May I suggest using `state.get()` instead of directly indexing the dictionary?

@lhoestq

albertvillanova:
It seems a backward compatibility issue...

lhoestq:
Thanks @samsontmr this should be fixed on master now

Feel free to reopen if you're still having issues

PR #2199, opened 2021-04-09
Fix backward compatibility in Dataset.load_from_disk

albertvillanova:
Fix backward compatibility when loading from disk an old dataset saved to disk with indices using key "_indices_data_files".

Related to #2195.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -589,7 +589,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
             ]
         }
         state["_data_files"] = [{"filename": config.DATASET_ARROW_FILENAME}]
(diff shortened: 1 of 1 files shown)


albertvillanova:
Hi @lhoestq, could you please check if this makes sense? Thanks.

lhoestq (approved):
Good catch !!

lhoestq:
What about using `_indices_data_files` field in save_to_diskinstead of `_indices_files` ?
This way future datasets can also be reloaded from older versions of the lib

albertvillanova:
Yes, I have seen it is not released yet...

You are right! It was your awesome PR on Tables which renamed this. If there is no particular reason for this renaming, yes, we could switch it back to the previous `_indices_data_files`. ;)

PR #2204, opened 2021-04-10
Add configurable options to `seqeval` metric

### LINKED ISSUE #2148: Add configurable options to `seqeval` metric
Right now `load_metric("seqeval")` only works in the default mode of evaluation (equivalent to conll evaluation).

However, seqeval library [supports](https://github.com/chakki-works/seqeval#support-features) different evaluation schemes (IOB1, IOB2, etc.), which can be plugged in just by supporting additional kwargs in `Seqeval._compute`
https://github.com/huggingface/datasets/blob/85cf7ff920c90ca2e12bedca12b36d2a043c3da2/metrics/seqeval/seqeval.py#L109

Things that would be relevant are, for example, supporting `mode="strict", scheme=IOB2` to count only full entity match as a true positive and omit partial matches.

The only problem I see is that the spirit of `metrics` seems to not require additional imports from user. `seqeval` only supports schemes as objects, without any string aliases. 

It can be solved naively with mapping like `{"IOB2": seqeval.scheme.IOB2}`. Or just left as is and require user to explicitly import scheme from `seqeval` if he wants to configure it past the default implementation.

If that makes sense, I am happy to implement the change.

marrodion:
Fixes #2148

Adds options to use strict mode, different schemes of evaluation, sample weight and adjust zero_division behavior, if encountered.

`seqeval` provides schemes as objects, hence dynamic import from string, to avoid making the user do the import (thanks to @albertvillanova for the `importlib` idea).


tests/test_metric.py
```diff
@@ -469,3 +473,11 @@ def test_input_tf(self):
             metric.add(prediction=pred, reference=ref)
         self.assertDictEqual(expected_results, metric.compute())
         del metric
+
+
+def test_seqeval_raises_when_incorrect_scheme():
```
lhoestq, line 478: Could you move this to test_metric_common.py ?

metrics/seqeval/seqeval.py
```diff
@@ -106,8 +117,32 @@ def _info(self):
...
+        suffix: bool = False,
+        scheme: Optional[str] = None,
+        mode: Optional[str] = None,
+        sample_weight: Optional[List[int]] = None,
+        zero_division: Union[str, int] = "warn",
+    ):
+        if scheme is not None:
+            try:
+                scheme_module = importlib.import_module("seqeval.scheme")
+                scheme = getattr(scheme_module, scheme)
```
lhoestq, line 133: Why not import seqeval at the top of the file and then do
```python
scheme = getattr(seqeval.schema, scheme)
```
?

instead of catching AttributeError you can do
```python
scheme = getattr(seqeval.schema, scheme, None)
if scheme is None:
    raise ...
```

lhoestq (commented):
Nice thank you !

tests/test_metric.py
marrodion: Done, thank you

metrics/seqeval/seqeval.py
marrodion: `None` is a valid value for scheme parameter

So in that case it will look like this
```python
if sheme is not None:
    scheme = getattr(seqeval.scheme, scheme, None)
    if scheme is None:
        raise ...
```
Which I find confusing, personally, because of two `if` clauses, while catching an exception explicitly seems to be clearer to me. However, I don't feel strongly about that, so if you think that `getattr(seqeval.scheme, scheme, None)` is better, I'll be happy to make the change.

lhoestq: Ok I see.
Let's keep it this way then or we may end up with something confusing. Thanks for the explanation !

lhoestq (approved):
LGTM thanks for adding these parameters :)

PR #2205, opened 2021-04-11
Updating citation information on LinCE readme

gaguilar:
Hi!

I just updated the citation information in this PR. It had an additional bibtex from one of the datasets used in LinCE and then the LinCE bibtex. I removed the former and added a link that shows the full list of citations for each dataset. 

Thanks!


lhoestq (approved):
Thank you !

PR #2216, opened 2021-04-13
added real label for glue/mrpc to test set

philschmid:
Added real label to `glue.py` `mrpc` task for test split.


lhoestq (approved):
Thank you !

PR #2220, opened 2021-04-14
Fix infinite loop in WindowsFileLock

albertvillanova:
Raise exception to avoid infinite loop.

### DIFF
--- a/src/datasets/utils/filelock.py
+++ b/src/datasets/utils/filelock.py
@@ -338,8 +338,8 @@ class WindowsFileLock(BaseFileLock):
 
         try:
             fd = os.open(self._lock_file, open_mode)
-        except OSError:
-            pass
+        except OSError as e:
+            raise e
         else:
             try:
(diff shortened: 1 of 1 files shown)


lhoestq:
How is it possible to get an infinite loop ? Can you add more details ?

albertvillanova:
Yes, in Windows, if the filename is too long, a `FileNotFoundError` is raised. The exception should be raised in this case.

If other process has the file locked, then `PermissionError` is raised. In this case, `pass` is OK.

lhoestq:
Note that the filelock module comes from this project that hasn't changed in years - while still being used by ten of thousands of projects:
https://github.com/benediktschmitt/py-filelock

Unless we have proper tests for this, I wouldn't recommend to change it

lhoestq:
I'm pretty sure many things from the library could break for windows users that haven't disabled the max path length limit.
Maybe it would be simpler to simply raise an error on startup. For exampe, for windows users the error could ask them to disable the limit if it's not been disabled yet ?

PR #2222, opened 2021-04-14
Fix too long WindowsFileLock name

albertvillanova:
Fix WindowsFileLock name longer than allowed MAX_PATH by shortening the basename.

### DIFF
--- a/src/datasets/utils/filelock.py
+++ b/src/datasets/utils/filelock.py
@@ -333,6 +333,24 @@ class WindowsFileLock(BaseFileLock):
     windows systems.
     """
 
+    def __init__(self, lock_file, timeout=-1):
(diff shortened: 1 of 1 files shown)


lhoestq:
Windows users should disable the max path length limit. It's a nightmare to handle it.
Also the lock path must not be changed in a random way. Otherwise from another process the lock path might not be the same and the locking mechanism won't work.

albertvillanova:
Do you agree with handling the case where MAX_PATH is not disabled? If not, we can close this PR.

If so, would it work a deterministic lock path instead of random?

lhoestq:
I'd rather not handle this at all, since there will be other places in the code where the limit will break things

PR #2223, opened 2021-04-14
Set test cache config

albertvillanova:
Currently, running the tests populates the default cache directory `"~/.cache"`.

This PR monkey-patches the config to set the cache directory within the temporary test directory, avoiding side effects.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -13,7 +13,9 @@ from .s3_fixtures import *  # noqa: load s3 fixtures
 
 @pytest.fixture(autouse=True)
 def set_test_cache_config(tmp_path_factory, monkeypatch):
-    test_hf_cache_home = tmp_path_factory.mktemp("cache", numbered=False)
+    # test_hf_cache_home = tmp_path_factory.mktemp("cache")  # TODO: why a cache dir per test function does not work?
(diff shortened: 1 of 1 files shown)


lhoestq:
>  why a cache dir per test function does not work?

Probably because we end up with multiple `datasets_module` in the python path. This breaks the import of all the datasets/metrics modules.
If you want to use one modules cache per test, you may need remove the `datasets_module` that was added to the python path during the test.
Indeed if the module cache hasn't been initialized, then it's added to the python path by calling `init_dynamic_modules`:

https://github.com/huggingface/datasets/blob/ba76012a19193a35053b9e20243ff40c2b4204ab/src/datasets/load.py#L291-L291

albertvillanova:
@lhoestq, for the moment, this PR avoids populating the `~/.cache` dir during training, which is already an improvement, isn't it?

lhoestq:
Yes we can merge it this way if you're fine with it !

lhoestq (approved):

albertvillanova:
I will eventually try to implement a `cache_dir` per test function in another PR, but I think I should first fix some side effects in tests: each test function should be atomic and able to have its own `cache_dir` without being affected by the `cache_dir` set in other test functions.

lhoestq:
Yes this would be ideal !

PR #2225, opened 2021-04-15
fixed one instance of 'train' to 'test'

alexwdong:
I believe this should be 'test' instead of 'train'


albertvillanova (approved):
Hi @alexwdong.

Thanks for reporting. Yes, you are right: it should be `test`.

lhoestq:
Thanks ! good catch

Could you also update the metadata of this dataset ?
You can do so by running
```
datasets-cli test ./datasets/newsgroup --all_configs --save_infos --ignore_verifications
```
This should update the dataset_infos.json file that contains the size of all the splits for example.

alexwdong:
Hi,
`dataset_infos.json` should be updated now.

lhoestq (approved):
thank you :)

PR #2227, opened 2021-04-16
Use update_metadata_with_features decorator in class_encode_column method

SBrandeis:
Following @mariosasko 's comment


lhoestq (approved):
Thank you and @mariosasko !

Issue #2146, opened 2021-03-30
Dataset file size on disk is very large


jblemoine:
Hi, 

I have created my own dataset using the provided dataset loading script. It is an image dataset where images are stored as 3D Array with dtype=uint8. 

The actual size on disk is surprisingly large. It takes 520 MB. Here is some info from `dataset_info.json`. 

`{
    "description": "",
    "citation": "",
    "homepage": "",
    "license": "",
    "features": {
        "image": {
            "shape": [224, 224, 3],
            "dtype": "uint8",
            "id": null,
            "_type": "Array3D",
        }
    },
    "post_processed": null,
    "supervised_keys": null,
    "builder_name": "shot_type_image_dataset",
    "config_name": "default",
    "version": {
        "version_str": "0.0.0",
        "description": null,
        "major": 0,
        "minor": 0,
        "patch": 0,
    },
    "splits": {
        "train": {
            "name": "train",
            "num_bytes": 520803408,
            "num_examples": 1479,
            "dataset_name": "shot_type_image_dataset",
        }
    },
    "download_checksums": {
        "": {
            "num_bytes": 16940447118,
            "checksum": "5854035705efe08b0ed8f3cf3da7b4d29cba9055c2d2d702c79785350d72ee03",
        }
    },
    "download_size": 16940447118,
    "post_processing_size": null,
    "dataset_size": 520803408,
    "size_in_bytes": 17461250526,
}`

I have created the same dataset with tensorflow_dataset and it takes only 125MB on disk.

I am wondering, is it normal behavior ? I understand `Datasets` uses Arrow for serialization wheres tf uses TF Records.

This might be a problem for large dataset. 

Thanks for your help.

lhoestq:
Hi ! In the arrow file we store all the integers as uint8.
So your arrow file should weigh around `height x width x n_channels x n_images` bytes.

What feature type do your TFDS dataset have ?

If it uses a `tfds.features.Image` type, then what is stored is the encoded data (as png or jpg for example). Since these encodings are made for compression, the resulting tfrecord is smaller that the arrow file.

We are working on adding a similar feature in `datasets`: the ability to store the encoded data instead of the raw integers for images, but also for audio data. This way, arrow files will have similar sizes as tfrecords for images.

jblemoine:
Thanks for the prompt response. You're right about the encoding since I used `tfds.features.Image` feature type as you mentioned.
[...]

PR #2228, opened 2021-04-16
Add ArrayXD support for fixed size list.

jblemoine:
Add support for fixed size list for ArrayXD when shape is known . See https://github.com/huggingface/datasets/issues/2146
Since offset are not stored anymore, the file size is now roughly equal to the actual data size.

### DIFF
--- a/src/datasets/features.py
+++ b/src/datasets/features.py
@@ -346,9 +346,8 @@ class _ArrayXDExtensionType(pa.PyExtensionType):
     def _generate_dtype(self, dtype):
         dtype = string_to_arrow(dtype)
         for d in reversed(self.shape):
-            dtype = pa.list_(dtype)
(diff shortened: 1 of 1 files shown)


lhoestq:
Awesome thanks ! To fix the CI you just need to merge master into your branch.
The error is unrelated to your PR

src/datasets/features.py
```diff
@@ -380,15 +379,17 @@ def __getitem__(self, i):
 
     def to_numpy(self):
         storage: pa.ListArray = self.storage
-        size = 1
+        size = len(self)
         for i in range(self.type.ndims):
             size *= self.type.shape[i]
             storage = storage.flatten()
+        storage = storage[:size]
```
lhoestq, line 386: Why do you need this line ?

PR #2233, opened 2021-04-16
Fix `xnli` dataset tuple key

### LINKED ISSUE #2229: `xnli` dataset creating a tuple key while yielding instead of `str` or `int`
When using  `ds = datasets.load_dataset('xnli', 'ar')`, the dataset generation script uses the following section of code in the egging, which yields a tuple key instead of the specified `str` or `int` key:
https://github.com/huggingface/datasets/blob/56346791aed417306d054d89bd693d6b7eab17f7/datasets/xnli/xnli.py#L196

Since, community datasets in Tensorflow Datasets also use HF datasets, this causes a Tuple key error while loading HF's `xnli` dataset. 
I'm up for sending a fix for this, I think we can simply use `file_idx + "_" + row_idx` as a unique key instead of a tuple.

NikhilBartwal:
Closes #2229 
The `xnli` dataset yields a tuple key in case of `ar` which is inconsistant with the acceptable key types (str/int).
The key was thus ported to `str` keeping the original information intact.

### DIFF
--- a/datasets/xnli/xnli.py
+++ b/datasets/xnli/xnli.py
@@ -193,7 +193,8 @@ class Xnli(datasets.GeneratorBasedBuilder):
                     file = open(filepath, encoding="utf-8")
                     reader = csv.DictReader(file, delimiter="\t", quoting=csv.QUOTE_NONE)
                     for row_idx, row in enumerate(reader):
-                        yield (file_idx, row_idx), {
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
Thanks !

The CI failure is not related to your PR and has been fixed on master.
Merging :)

PR #2234, opened 2021-04-17
Fix bash snippet formatting in ADD_NEW_DATASET.md

mariosasko:
This PR indents the paragraphs around the bash snippets in ADD_NEW_DATASET.md to fix formatting.


albertvillanova (approved):
Thanks @mariosasko.

PR #2238, opened 2021-04-19
NLU evaluation data

dkajtoch:
New intent classification dataset from https://github.com/xliuhw/NLU-Evaluation-Data


datasets/nlu_evaluation_data/README.md
```diff
@@ -0,0 +1,271 @@
...
+email, game, general, IoT, lists, music, news, podcasts, general Q&A, radio, recommendations, social, food takeaway, transport, and weather.
+The questions put to Turkers were designed to capture the different requests
+within each given scenario. 
+In the ‘calendar’ scenario, for example, these pre-designed intents were included: ‘set event’, ‘delete event’ and ‘query event’. 
+An example question for intent ‘set event’ is: “How would you ask your PDA to schedule a meeting with someone?” for which a user’s answer example was “Schedule
+a chat with Adam on Thursday afternoon”. 
+The Turkers would then type in their answers to these questions and select possible entities from the pre-designed suggested entities list for each of their answers.The Turkers didn’t always follow the
+instructions fully, e.g. for the specified ‘delete event’ Intent, an answer was: “PDA
+what is my next event?”; which clearly belongs to ‘query event’ Intent. 
+We have manually corrected all such errors either during post-processing or the subsequent annotations.
```
lhoestq, line 221:
```suggestion
> To build the NLU component we collected real user data via Amazon Mechanical Turk (AMT). We designed tasks where the Turker’s goal was to answer questions about how people would interact with the home robot, in a wide range of scenarios designed in advance, namely: alarm, audio, audiobook, calendar, cooking, datetime, email, game, general, IoT, lists, music, news, podcasts, general Q&A, radio, recommendations, social, food takeaway, transport, and weather.
The questions put to Turkers were designed to capture the different requests within each given scenario. 
In the ‘calendar’ scenario, for example, these pre-designed intents were included: ‘set event’, ‘delete event’ and ‘query event’. 
An example question for intent ‘set event’ is: “How would you ask your PDA to schedule a meeting with someone?” for which a user’s answer example was “Schedule a chat with Adam on Thursday afternoon”. 
The Turkers would then type in their answers to these questions and select possible entities from the pre-designed suggested entities list for each of their answers.The Turkers didn’t always follow the instructions fully, e.g. for the specified ‘delete event’ Intent, an answer was: “PDA what is my next event?”; which clearly belongs to ‘query event’ Intent. 
We have manually corrected all such errors either during post-processing or the subsequent annotations.
```
removed bad end of lines

datasets/nlu_evaluation_data/nlu_evaluation_data.py
```diff
@@ -0,0 +1,234 @@
...
+                    define_intent_name(ex["scenario"], ex["intent"])
+                ),
+            ),
+            batched=False,
+            with_indices=False,
+            features=datasets.Features(self.info.features, **self.info.post_processed.features),
+        )
+        # drop unused columns
+        dataset = dataset.remove_columns(["answer_annotation", "intent"])
+        return dataset
```
lhoestq, line 211: In general we do simple processing of examples directly in `_generate_examples`.
Could you do this parsing/processing in `_generate_examples` instead of using `_post_process` please ?

datasets/nlu_evaluation_data/nlu_evaluation_data.py
```diff
@@ -0,0 +1,234 @@
...
+    """
+    Intent name is defined as concatenation of `scenario` and `intent` values.
+
+    See Also:
+        https://github.com/xliuhw/NLU-Evaluation-Data/issues/5
+    """
+    return f"{scenario}_{intent}"
+
+
+class NLUEvaluationDataset(datasets.GeneratorBasedBuilder):
```
lhoestq, line 82:
```suggestion
class NLUEvaluationData(datasets.GeneratorBasedBuilder):
```
This must be the camel case version of the dataset script name `nlu_evaluation_data`

lhoestq (commented):
Awesome thank you ! Good job with the script and the dataset card :)

My only concern is that we should do examples processing in `_generate_examples` instead of using `_post_process`. Do you think you could do that change please ?

After this change you'll need to regenerate the datasets_infos.json:
```
datasets-cli test ./datasets/nlu_evaluation_data --save_infos --ignore_verifications
```

datasets/nlu_evaluation_data/nlu_evaluation_data.py
dkajtoch: @lhoestq it shouldn't be a problem except filtering (remove empty examples). If I do this in the for loop in `_generate_examples` (as I did previously) summary of the dataset will have an incorrect number of rows

lhoestq: Thanks for the change !
It's ok to have less rows as soon as you mention in the dataset card that you removed empty rows.
Overall the idea is to have everything in _generate_examples

dkajtoch: Ok so I will push one more change for that

dkajtoch: @lhoestq some problems with filtering in for loop. I tried to put something like this in the for loop in `_generate_examples`, but this hack does not pass tests since it produces a different dataset size. Any recommendation on how to do this properly?
```
                if answer_annotation == "null":
                    continue

                yield id_, {
                    "text": remove_annotations(answer_annotation),
                    "scenario": scenario,
                    "label": define_intent_name(scenario, intent),
                }
```

lhoestq: I think you can fix this by updating the number of expected examples in the dataset metadata in `dataset_infos.json`:
```
datasets-cli test ./datasets/nlu_evaluation_data --save_infos --all_configs --ignore_verifications
```

dkajtoch: thanks for a quick reply!

lhoestq (approved):
Thanks this looks all good now !

Issue #2239, opened 2021-04-19
Error loading wikihow dataset


odellus:
## Describe the bug

When attempting to load wikihow into a dataset with
```python
from datasets import load_dataset
dataset = load_dataset('wikihow', data_dir='./wikihow')
```
I get the message:
```
AttributeError: 'BuilderConfig' object has no attribute 'filename'
```
at the end of a [full stack trace](https://gist.github.com/odellus/602c3b2de52f541d353b1022f320ffc2).

## Steps to reproduce the bug

I have followed the instructions for creating a wikihow dataset. The [wikihow dataset site](https://huggingface.co/datasets/wikihow) says to use 
```python
from datasets import load_dataset
dataset = load_dataset('wikihow')
```
to load the dataset. I do so and I get the message
```
AssertionError: The dataset wikihow with config all requires manual data.
 Please follow the manual download instructions:   You need to manually download two wikihow files. An overview of which files to download can be seen at https://github.com/mahnazkoupaee/WikiHow-Dataset.
  You need to download the following two files manually:
    1) https://ucsb.app.box.com/s/ap23l8gafpezf4tq3wapr6u8241zz358 and save the file under <path/to/folder>/wikihowAll.csv
    2) https://ucsb.app.box.com/s/7yq601ijl1lzvlfu4rjdbbxforzd2oag and save the file under <path/to/folder>/wikihowSep.csv

  The <path/to/folder> can e.g. be "~/manual_wikihow_data".

  Wikihow can then be loaded using the following command `datasets.load_dataset("wikihow", data_dir="<path/to/folder>")`.
  .
 Manual data can be loaded with `datasets.load_dataset(wikihow, data_dir='<path/to/manual/data>')
```

So I create a directory `./wikihow` and download `wikihowAll.csv` and `wikihowSep.csv` into the new directory.

Then I run 
```python
from datasets import load_dataset
dataset = load_dataset('wikihow', data_dir='./wikihow')
```

that's when I get the [stack trace](https://gist.github.com/odellus/602c3b2de52f541d353b1022f320ffc2)

## Expected results
I expected it to load the downloaded files into a dataset.

## Actual results
```python
Using custom data configuration default-data_dir=.%2Fwikihow
[...]

PR #2240, opened 2021-04-20
Clarify how to load wikihow

albertvillanova:
Explain clearer how to load the dataset in the manual download instructions.

En relation with #2239.


lhoestq (approved):
thanks !

PR #2241, opened 2021-04-20
Add SLR32 to OpenSLR

cahya-wirawan:
I would like to add SLR32 to OpenSLR. It contains four South African languages: Afrikaans, Sesotho, Setswana and isiXhosa


lhoestq (approved):
And yet another one ! Thanks a lot :)

cahya-wirawan:
> And yet another one ! Thanks a lot :)

I just hope you don’t get fed up with openslr PR 😊 there are still few other datasets created by google in openslr that is not in hf dataset yet

PR #2247, opened 2021-04-22
Implement Dataset from Parquet

albertvillanova:
Implement instantiation of Dataset from Parquet file.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -324,6 +324,22 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
         )
 
     @classmethod
+    def from_parquet(cls, filename: str, in_memory: bool = True):
+        """Create Dataset from Parquet file.
+
+        Args:
+            filename (:obj:`str`): Path of the Parquet file.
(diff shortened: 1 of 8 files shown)


src/datasets/arrow_dataset.py
```diff
@@ -323,6 +318,22 @@ def from_file(
...
+        """Create Dataset from Parquet file.
+
+        Args:
+            filename (:obj:`str`): Path of the Parquet file.
+            in_memory (:obj:`bool`, default ``True``): Whether to copy the data in-memory.
+
+        Returns:
+            :class:`Dataset`
+        """
+        from .arrow_reader import ParquetReader
```
lhoestq, line 332: The ParquetReader unfortunately always bring data in memory. The memory-map option is here only for reading performance. Under the hood the data from parquet are compressed so you can't just memory map the file since there's no on-the-fly decoding afaik.

I think the best option would be to implement a ArrowBasedBuilder as we did for csv/json/etc.

lhoestq:
Hi @albertvillanova , I'll implement the parquet builder as an ArrowBasedBuilder if you don't mind

lhoestq:
closing in favor of #2537 that is already merged

PR #2249, opened 2021-04-22
Allow loading only specific splits

albertvillanova:
Allow loading only specific splits without loading the other splits.

### DIFF
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -500,6 +500,15 @@ class DatasetBuilder:
                     use_etag=False,
                     use_auth_token=use_auth_token,
                 )  # We don't use etag for data files to speed up the process
+            else:
+                if not download_config.cache_dir:
+                    download_config.cache_dir = os.path.join(self._cache_dir_root, "downloads")
+                if download_config._is_force_download_default:
+                    download_config.force_download = bool(download_mode == FORCE_REDOWNLOAD)
+                if download_config._is_use_etag_default:
+                    download_config.use_etag = False
+                if download_config.use_auth_token is None:
+                    download_config.use_auth_token = use_auth_token
 
             dl_manager = DownloadManager(
                 dataset_name=self.name,
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -641,6 +641,12 @@ class DatasetBuilder:
         split_dict = SplitDict(dataset_name=self.name)
         split_generators_kwargs = self._make_split_generators_kwargs(prepare_split_kwargs)
         split_generators = self._split_generators(dl_manager, **split_generators_kwargs)
+        if dl_manager._download_config.only_splits:
+            split_generators = [
+                split_generator
+                for split_generator in split_generators
+                if split_generator.name in dl_manager._download_config.only_splits
+            ]
 
         # Checksums verification
         if verify_infos:
--- a/src/datasets/utils/file_utils.py
+++ b/src/datasets/utils/file_utils.py
@@ -229,19 +229,28 @@ class DownloadConfig:
     """
 
     cache_dir: Optional[Union[str, Path]] = None
-    force_download: bool = False
+    force_download: Optional[bool] = None  # default False
+    _is_force_download_default = False
     resume_download: bool = False
     local_files_only: bool = False
     proxies: Optional[Dict] = None
     user_agent: Optional[str] = None
     extract_compressed_file: bool = False
     force_extract: bool = False
-    use_etag: bool = True
+    use_etag: Optional[bool] = None  # default True
+    _is_use_etag_default = False
     num_proc: Optional[int] = None
     max_retries: int = 1
     use_auth_token: Optional[Union[str, bool]] = None
     only_splits: Optional[list] = None
 
+    def __post_init__(self):
+        if self.use_etag is None:
+            self.use_etag = True
+            self._is_use_etag_default = True
+        if self.force_download is None:
+            self.force_download = False
+            self.is_force_download_default = True
 
     def copy(self) -> "DownloadConfig":
         return self.__class__(**{k: copy.deepcopy(v) for k, v in self.__dict__.items()})
--- a/src/datasets/utils/file_utils.py
+++ b/src/datasets/utils/file_utils.py
@@ -240,6 +240,8 @@ class DownloadConfig:
     num_proc: Optional[int] = None
     max_retries: int = 1
     use_auth_token: Optional[Union[str, bool]] = None
+    only_splits: Optional[list] = None
+
 
     def copy(self) -> "DownloadConfig":
         return self.__class__(**{k: copy.deepcopy(v) for k, v in self.__dict__.items()})
--- a/tests/test_load.py
+++ b/tests/test_load.py
@@ -247,17 +247,22 @@ def test_load_dataset_local_with_default_in_memory(
 
 
 # @pytest.mark.parametrize("keep_in_memory", [False, True])
-def test_load_dataset_local_only_splits(dataset_loading_script_dir, data_dir):  # , keep_in_memory
+def test_load_dataset_local_only_splits(dataset_loading_script_dir, data_dir, tmp_path):  # , keep_in_memory
+    cache_dir = str(tmp_path / "cache")
     download_config = DownloadConfig(only_splits="test")
     # with assert_arrow_memory_increases() if keep_in_memory else assert_arrow_memory_doesnt_increase():
     dataset = datasets.load_dataset(
-        dataset_loading_script_dir, data_dir=data_dir, split="test", download_config=download_config
+        dataset_loading_script_dir,
+        data_dir=data_dir,
+        cache_dir=cache_dir,
+        split="test",
+        download_config=download_config,
     )  # , keep_in_memory=keep_in_memory)
     assert isinstance(dataset, Dataset)
     assert dataset.split == "test"
     assert dataset.shape == (10, 1)
     # pattern = "*/0.0.0/74c0095031cf868e2486de6e08bb3ca4a9f9de3a81b10af67a42aed21393e640/*.arrow"
-    generated_arrow_files = Path(datasets.config.HF_DATASETS_CACHE, dataset.builder_name).glob("**/*.arrow")
+    generated_arrow_files = Path(cache_dir, dataset.builder_name).glob("**/*.arrow")
     assert len(list(generated_arrow_files)) == 1
 
 
--- a/tests/test_load.py
+++ b/tests/test_load.py
@@ -4,6 +4,7 @@ import shutil
 import tempfile
 import time
 from hashlib import sha256
+from pathlib import Path
 from unittest import TestCase
 from unittest.mock import patch
 
@@ -11,7 +12,7 @@ import pytest
 import requests
 
 import datasets
-from datasets import load_dataset, load_from_disk
+from datasets import Dataset, DownloadConfig, load_dataset, load_from_disk
 
 from .utils import OfflineSimulationMode, assert_arrow_memory_doesnt_increase, assert_arrow_memory_increases, offline
 
@@ -243,6 +244,21 @@ def test_load_dataset_local_with_default_in_memory(
     with assert_arrow_memory_increases() if expected_in_memory else assert_arrow_memory_doesnt_increase():
         dataset = load_dataset(dataset_loading_script_dir, data_dir=data_dir)
     assert (dataset["train"].dataset_size < max_in_memory_dataset_size) is expected_in_memory
+
+
+# @pytest.mark.parametrize("keep_in_memory", [False, True])
+def test_load_dataset_local_only_splits(dataset_loading_script_dir, data_dir):  # , keep_in_memory
+    download_config = DownloadConfig(only_splits="test")
+    # with assert_arrow_memory_increases() if keep_in_memory else assert_arrow_memory_doesnt_increase():
+    dataset = datasets.load_dataset(
+        dataset_loading_script_dir, data_dir=data_dir, split="test", download_config=download_config
+    )  # , keep_in_memory=keep_in_memory)
+    assert isinstance(dataset, Dataset)
+    assert dataset.split == "test"
+    assert dataset.shape == (10, 1)
+    # pattern = "*/0.0.0/74c0095031cf868e2486de6e08bb3ca4a9f9de3a81b10af67a42aed21393e640/*.arrow"
+    generated_arrow_files = Path(datasets.config.HF_DATASETS_CACHE, dataset.builder_name).glob("**/*.arrow")
+    assert len(list(generated_arrow_files)) == 1
 
 
 @pytest.mark.parametrize("max_in_memory_dataset_size", ["default", None, 0, 100, 1000])


src/datasets/utils/download_manager.py
```diff
@@ -182,6 +183,12 @@ def download(self, url_or_urls):
             downloaded_path(s): `str`, The downloaded paths matching the given input
                 url_or_urls.
         """
+        if self._download_config.only_splits:
+            if isinstance(url_or_urls, dict) and all(
+                split in url_or_urls for split in self._download_config.only_splits
+            ):
+                url_or_urls = {split: url_or_urls[split] for split in self._download_config.only_splits}
```
lhoestq, line 190: That's a great trick ! However as you know this won't work on all cases unfortunately.
For example some datasets use a dictionary with keys that are not related to the splits since we allow any nested structure to be passed for downloading.

What do you think about adding an argument to `_split_generators` to tell it to return only a subset of the split generators ? In this case it would download only the files related to the requested splits

Then we can just edit the datasets that can benefit from this feature the most.

albertvillanova: The problem is that `split_generators` is already implemented in ALL canonical datasets without that parameter...

lhoestq: Maybe we can pass the parameter only if it's part of the signature of `_split_generators` ?
We're already doing that for other optional parameters like `pipeline` in `BeamBasedBuilder._split_generators`:

https://github.com/huggingface/datasets/blob/909c58f6cf9568d73e87c89516a5e423983ad966/src/datasets/builder.py#L1065-L1067

I'm fine with this option but let me know if you have other ideas

albertvillanova: Now I pass the parameter to `_split_generators`. However, it must be handled properly by the method implementations.

Note that with my previous trick, it already works for `natural_questions` without even touching its current implementation of `split_generators.` 😉

src/datasets/builder.py
```diff
@@ -627,7 +636,19 @@ def _download_and_prepare(self, dl_manager, verify_infos, **prepare_split_kwargs
         # Generating data for all splits
         split_dict = SplitDict(dataset_name=self.name)
         split_generators_kwargs = self._make_split_generators_kwargs(prepare_split_kwargs)
-        split_generators = self._split_generators(dl_manager, **split_generators_kwargs)
+        try:
+            split_generators = self._split_generators(
+                dl_manager, splits=dl_manager._download_config.splits, **split_generators_kwargs
+            )
+        except TypeError:
+            split_generators = self._split_generators(dl_manager, **split_generators_kwargs)
```
lhoestq, line 644: You can use `"splits" in inspect.signature(self._split_generators).parameters.keys()` instead of try/catch

src/datasets/utils/download_manager.py
```diff
@@ -182,6 +183,10 @@ def download(self, url_or_urls):
             downloaded_path(s): `str`, The downloaded paths matching the given input
                 url_or_urls.
         """
+        if self._download_config.splits:
+            if isinstance(url_or_urls, dict) and all(split in url_or_urls for split in self._download_config.splits):
```
lhoestq, line 187: I would add an additional check just to avoid unwanted behaviors.
For example right now if a dataset script passes to the dl_manager a dict like this:
```
{"main_data": url_to_main_data, "secondary_data": url_to_sec_data}
```
then this trick here would use `url_or_urls = {}` since the keys of the dict are not split names.

Maybe you could check that `sorted(url_or_urls.keys()) == sorted(self._download_config.splits)` before filtering ?

src/datasets/utils/download_manager.py
```diff
@@ -266,7 +271,11 @@ def extract(self, path_or_paths, num_proc=None):
         path_or_paths = NestedDataStructure(path_or_paths)
         extracted_paths = NestedDataStructure(extracted_paths)
         self.extracted_paths.update(dict(zip(path_or_paths.flatten(), extracted_paths.flatten())))
-        return extracted_paths.data
+        return (
+            extracted_paths.data
+            if not isinstance(extracted_paths.data, dict)
+            else defaultdict(str, extracted_paths.data)
+        )
```
lhoestq, line 278: Why do you need this ?

src/datasets/utils/file_utils.py
```diff
@@ -229,17 +229,28 @@ class DownloadConfig:
     """
 
     cache_dir: Optional[Union[str, Path]] = None
-    force_download: bool = False
+    force_download: Optional[bool] = None  # default False
+    _is_force_download_default: bool = False
```
lhoestq, line 233: Why do you need this ?

src/datasets/utils/file_utils.py
```diff
@@ -229,17 +229,28 @@ class DownloadConfig:
...
     user_agent: Optional[str] = None
     extract_compressed_file: bool = False
     force_extract: bool = False
-    use_etag: bool = True
+    use_etag: Optional[bool] = None  # default True
+    _is_use_etag_default: bool = False
     num_proc: Optional[int] = None
     max_retries: int = 1
     use_auth_token: Optional[Union[str, bool]] = None
+    splits: Optional[list] = None
```
lhoestq, line 245: The DownloadConfig class was used until now as a the class that defines the parameters that we pass to the `cached_path` function, and it had no logic related to the DownloadManager or to datasets specifically.
Therefore I'm not sure this is best place to put this argument. Let me know what you think.

Maybe this could be an argument of the DownloadManager itself.

tests/test_load.py
```diff
@@ -245,6 +246,42 @@ def test_load_dataset_local_with_default_in_memory(
...
+class TestLoadDatasetOnlySplits:
+    def test_load_dataset_local_only_splits_processed_files(self, dataset_loading_script_dir, data_dir, tmp_path):
+        download_config = DownloadConfig(splits=["test"])
+        cache_dir = str(tmp_path / "cache")
+        datasetdict = datasets.load_dataset(
+            dataset_loading_script_dir,
+            data_dir=data_dir,
+            cache_dir=cache_dir,
+            download_config=download_config,
+        )
```
lhoestq, line 258: In my opinion the users shouldn't have another place to choose the split. The `split` parameter of `load_dataset` is made for this.

tests/test_load.py
```diff
@@ -245,6 +246,42 @@ def test_load_dataset_local_with_default_in_memory(
     assert (dataset["train"].dataset_size < max_in_memory_dataset_size) is expected_in_memory
 
 
+class TestLoadDatasetOnlySplits:
```
lhoestq, line 249: We should also check this scenario:
1. users load a dataset with the "train" split only
2. users reloads this dataset but this time asking for the "test" set only

The caching mechanism should notice that the "test" set is missing and download and prepare it. If I'm not wrong currently this would fail because of this line:

https://github.com/huggingface/datasets/blob/097129df8e3f357c24b3ad439b252e0d8b849dfa/src/datasets/builder.py#L504-L511

Instead of checking for the cache_dir, we should check for the actual arrow split file to exist inside cache_dir

lhoestq (commented):
Love it ! thank you :)

I added some comments:

src/datasets/builder.py
albertvillanova: I just applied the Pythonic EAFP principle: "It is Easier to Ask for Forgiveness than Permission".

If you really think it is better to make 4 additional calls (`inspect.signature`, `keys`, `if`, `in`, besides the import of the `inspect` package), I can refactor the code. 😅

src/datasets/utils/download_manager.py
albertvillanova: Because the user implementation of the method _split_generators will access all possible keys in the extracted path dict and may certainly access them with __getitem__. If our downloaed hasn't downloaded all possible splits (but only a subset), accessing the missing ones with __getitem__ will raise a `KeyError`.

If a `dafaultdict` is returned instead, trying to access a missing (not downloaded) splits will be silently ignored.

src/datasets/utils/file_utils.py
albertvillanova: I have not a clear opinion on this yet (I'm going to think about it), but I can tell you that for readability, a data class called `DownloadConfig` should be clearly used by a `DownloadManager` class. Analogously, `cached_path` should use something like `CachedPathConfig`.

In my opinion I find quite sensible to include a download configuration setting (whether to download all or only some of the files) in a `DownloadConfig` object. And in order to avoid passing lots of parameters to `load_dataset`, it makes sense to pass the parameters related to `download` stuff in a `DownloadConfig` class.

As a side note, I am planning to refactor also `cached_path` because it contains too many different functionalities. For the moment, I have extracted all Extract functionalities. But I am planning further refactoring...

lhoestq: The separation between the two configs makes sense to me, and I totally agree with you with the naming.
Maybe we could have `DownloadConfig` as a subclass of `CachedPathConfig` or something like that (or another relation between the two).

Regarding cached_path, note that this is a function that we have in common with the other libraries (transformers and huggingface_hub) so it could be worth discussing with the other maintainers about the changes we want to do.

src/datasets/utils/download_manager.py
lhoestq: Edit:

I meant data like this 
```
{"main_metadata": url_to_main_data, "secondary_metadata": url_to_sec_data, "train": url_train_data, "test": url_test_data}
```

src/datasets/builder.py
lhoestq: I have a small preference for `inspect` since the implementation of _split_generators may raise a TypeError, especially when someone is implementing this method and may have added bugs (but no strong opinion). It would be nice to not catch and hide errors from the users' code or they may have trouble debugging.

src/datasets/utils/download_manager.py
lhoestq: Makes sense ! Maybe use a more explicit placeholder that an empty string, in case users experience errors using this ?

tests/test_load.py
albertvillanova: As far as I know, the `split` parameter of `load_dataset` is for _loading_: loading from cache.

On the contrary, the `splits` parameter passed to `DownloadConfig` is for _downloading_.

src/datasets/utils/file_utils.py
albertvillanova: This has to do with the fact of current over-use of `DownloadConfig` for many different tasks/concerns...

tests/test_load.py
albertvillanova: Current implementation of `load_dataset` assumes that if there is something in the cache, then it is _all_ you can get, unless you pass `force_download`.

I assumed that if the user passes a parameter to download only one split, then they are aware that only that split is in the cache and only that split can be loaded. In order to force a subsequent download of other split, they should pass `force_download`.

lhoestq: Not a big fan of that. In my opinion it should detect that the requested split is missing and generate it.
This will avoid some confusions for users

src/datasets/builder.py
albertvillanova: In the except clause, the same method is called again (the only difference: without passing the parameter `splits`). Therefore, any user-created exception will be raised anyway. 😉

lhoestq: Right :)

src/datasets/utils/file_utils.py
lhoestq: I see !
What if the value `force_download` is updated during `download_and_prepare` ? In this case `_is_force_download_default` would still be True even though it's not the default value "None" anymore.
I find the name of the attribute confusing in this case

tests/test_load.py
lhoestq: This is also a breaking change we can't afford IMO

src/datasets/utils/download_manager.py
albertvillanova: Hi @lhoestq. Sorry, I'm not sure of understanding what you mean...

What I am checking here is that the all the keys in `self._download_config.splits` are a subset of the keys in `url_or_urls`.

src/datasets/utils/download_manager.py
albertvillanova: I've set `"<NOT_DOWNLOADED>"` instead.

src/datasets/utils/download_manager.py
lhoestq: If you pass a dictionary like this:
```
{"main_metadata": url_to_main_data,
"secondary_metadata": url_to_sec_data,
"train": url_train_data,
"test": url_test_data}
```
then only the train and test keys will be kept, which I feel not intuitive.

For example if the users asks to load the "train" split, then the main and secondary metadata won't be downloaded.
You can fix that by keeping all the keys except the splits to ignore

src/datasets/utils/file_utils.py
albertvillanova: I have renamed the attributes, e.g. `_is_force_download_set_by_user`.

src/datasets/utils/download_manager.py
albertvillanova: Thanks @lhoestq, I understand it now.

albertvillanova:
> If you pass a dictionary like this:
> 
> ```
> {"main_metadata": url_to_main_data,
> "secondary_metadata": url_to_sec_data,
> "train": url_train_data,
> "test": url_test_data}
> ```
> 
> then only the train or test keys will be kept, which I feel not intuitive.
> 
> For example if the users asks to load the "train" split, then the main and secondary metadata won't be downloaded.
> You can fix that by keeping all the keys except the splits to ignore

Hi @lhoestq, I have been thinking about this and I think it is worth that we discuss about it.

When I created this PR, my first idea was to create a "hack" inside the download manager that will be able to filter some split(s) without touching any dataset script. Of course, the download manager does not know about splits logic, and thus this trick would only work for some very specific datasets: only the ones containing that pass a dict to the download manager containing only the keys "train", "validation", "test" (or the one passed by the user for advanced users knowing they can do it), e.g. the `natural_questions` dataset (which was one of the targets).

The big inconvenient of this approach is that it is not applicable to many datasets (or worse, it should be constantly tweaked to cope with exceptional cases). One exceptional case is the one you pointed out. But I see others:
- the split keys can be different: train, test, dev, val, validation, eval,...
- in `hope_edi` dataset, the split keys are: TRAIN_DOWNLOAD_URL, VALIDATION_DOWNLOAD_URL
- in `few_rel` dataset, the split keys are: train_wiki, val_nyt, val_pubmed,..., pid2name
- in `curiosity_dialogs`, the split keys are: train, val, test, test_zero; this means that for every split we pass, we will also get test_zero
- in `deal_or_no_dialog`, each of the splits URL is passed separately to the download manager, so all splits would be always downloaded
- etc.

Then after discussing, another idea emerged: pass a `split` parameter to `_split_generators`, which know about the splits logic, so that it can handle which splits are passed to the download manager. This approach is more accurate and can be tweaked so that it works with all the datasets we want. The only inconvenient is that then for every target dataset, we must modify its corresponding `_split_generators` script method.

My point is that I don't think it is a good idea to implement both approaches. They could even interfere with each other! 

If you agree, I would implement ONLY the second one, which is simpler, more consistent and stable and will avoid future problems.

src/datasets/utils/download_manager.py
albertvillanova: See my comment below.

lhoestq:
Hi @albertvillanova !
Yup I agree with you, implementing the 2nd approach seems to be the right solution

PR #2254, opened 2021-04-23
Update format, fingerprint and indices after add_item

lhoestq:
Added fingerprint and format update wrappers + update the indices by adding the index of the newly added item in the table.


albertvillanova (approved):
Thanks. I will take this PR as an example to complete my PR on `add_column`.

lhoestq:
I renamed the variable, added a test for dataset._indices and fixed an issue with class_encode_column

PR #2255, opened 2021-04-23
Task casting for text classification & question answering

SBrandeis:
This PR implements task preparation for a given task, in the continuation of #2143 

Task taxonomy follows 🤗 Transformers's pipelines taxonomy: https://github.com/huggingface/transformers/tree/master/src/transformers/pipelines

Edit by @lewtun:

This PR implements support for the following tasks:

* `text-classification`
* `question-answering`

The intended usage is as follows:

```python
# Load a dataset with default column names / features
ds = load_dataset("dataset_name")
# Cast column names / features to schema. Casting is defined in the dataset's `DatasetInfo`
ds = ds.prepare_for_task(task="text-classification")
# Casting can also be realised during load
ds = load_dataset("dataset_name", task="text-classification")
# We can also combine shared tasks across dataset concatenation
ds1 = load_dataset("dataset_name_1", task="text-classification")
ds2 = load_dataset("dataset_name_2", task="text-classification")
# If the tasks have the same schema, so will `ds_concat`
ds_concat = concatenate_datasets([ds1, ds2])
```

Note that the current implementation assumes that `DatasetInfo.task_templates` has been pre-defined by the user / contributor when overriding the `MyDataset(GeneratorBasedBuilder)._info` function.

As pointed out by @SBrandeis, for evaluation we'll need a way to detect which datasets are already have a compatible schema so we don't have to edit hundreds of dataset scripts. One possibility is to check if the schema features are a subset of the dataset ones, e.g.

```python
squad = load_dataset("./datasets/squad", split="train")
qa = QuestionAnswering()
schema = Features({**qa.input_schema, **qa.label_schema})
assert all(item in squad.features.items() for item in schema.items())
```


SBrandeis:
cc @abhi1thakur

lhoestq:
Looks really nice so far, thanks !
Maybe if a dataset doesn't have a template for a specific task we could try the default template of this task ?

lewtun:
hey @SBrandeis @lhoestq,

i now have a better idea about what you guys are trying to achieve with the task templates and have a few follow-up questions:

1. how did you envision using `DatasetInfo` for running evaluation? my understanding is that all `dataset_infos.json` files are stored in the `datasets` repo (unlike `transformers` where each model's weights etc are stored in a dedicated repo). 
this suggests the following workflow:

```
- git clone datasets
- load target dataset to evaluate
- load `dataset_infos.json` for target dataset
- run eval for each task template in `task_templates`
- store metrics as evaluation cards (similar to what is done in `autonlp`)
```
2. assuming the above workflow, i see that the current `TaskTemplate` attributes of `task`, `input_schema`, and `label_schema` still require some wrangling from `dataset_infos.json` to reproduce additional mappings like `label2id` that we'd need for e.g. text classification. an alternative would be to instantiate the task template class directly from the JSON with something like
```python
from datasets.tasks import TextClassification
from transformers import AutoModelForSequenceClassification, AutoConfig

tc = TextClassification.from_json("path/to/dataset_infos.json")
# load a model with the desired config
model_ckpt = ...
config = AutoConfig.from_pretrained(model_ckpt, label2id=tc.label2id, id2label=tc.id2label)
model = AutoModelForSequenceClassification.from_pretrained(model_ckpt, config=config)
# run eval ...
```
perhaps this is what @SBrandeis had in mind with the `TaskTemplate.from_dict` method?
3. i personally prefer using `task_templates` over `supervised_keys` because it encourages the contributor to think in terms of 1 or more tasks. my question here is do we currently use `supervised_keys` for anything important in the `datasets` library?

SBrandeis:
1. How do you envision using DatasetInfo for running evaluation?

The initial idea was to be able to do something like this:
```python
dset = load_dataset("name", task="binary_classification")
# OR
dset = load_dataset("name")
dset = dset.prepare_for-task("binary_classification")
```

2. I don't think that's needed if we proceed as mentioned above

3. `supervised_keys` are mostly a legacy compatibility thing with TF datasets, not sure it's used for anything right now. I'll let @lhoestq give more details on that

lewtun:
> The initial idea was to be able to do something like this:
> 
> ```python
> from datasets import load_dataset
> dset = load_dataset("name", task="binary_classification")
> # OR
> dset = load_dataset("name")
> dset = dset.prepare_for_task("binary_classification")
> ```

ah that's very elegant! just so i've completely understood, the result would be that the relevant column names of `dset` would be mapped to e.g. `text` and `label` and thus we'd have a uniform schema for the evaluation of all `binary_classification` tasks?

SBrandeis:
That's correct! Also, the features need to be appropriately casted
For a classification task for example, we would need to cast the datasets features to something like this:
```python
datasets.Features({
    "label": datasets.ClassLabel(names=[...]),
})
```

lhoestq:
3. We can ignore `supervised_keys` (it came from TFDS and we're not using it) and use `task_templates`

lewtun:
great, thanks a lot for your answers! now it's much clearer what i need to do next 😃

src/datasets/arrow_dataset.py
```diff
@@ -1372,6 +1372,33 @@ def with_transform(
...
+        if len(compatible_templates) > 1:
+            raise ValueError(
+                f"""Expected 1 task template but found {len(compatible_templates)}! Please ensure that \
+                :attr:`datasets.DatasetInfo.task_templates` contains a unique set of task types."""
+            )
+        template = compatible_templates[0]
+        column_mapping = template.column_mapping
+        columns_to_drop = [column for column in self.column_names if column not in column_mapping]
+        dataset = self.remove_columns(columns_to_drop)
+        # TODO(sbrandeis): Add support for unnesting columns too
```
lewtun, line 1397: hey @SBrandeis could you clarify what you mean by "unnesting columns" here? 

do you mean that `Datasets.rename_columns` should supported columns with nested features like `answers.text` in SQuAD:

```python
features=datasets.Features(
    {
        "id": datasets.Value("string"),
        "title": datasets.Value("string"),
        "context": datasets.Value("string"),
        "question": datasets.Value("string"),
        "answers": datasets.features.Sequence(
            {
                "answer_start": datasets.Value("int32"),
            }
        ),
    }
)
```

lewtun:
hey @lhoestq @SBrandeis, 

i've made some small tweaks to @SBrandeis's code so that `Dataset.prepare_for_task` is called in `DatasetBuilder`. using the `emotion` dataset as a test case, the following now works:

 ```python
# DatasetDict with default columns
ds = load_dataset("./datasets/emotion/")
# DatasetDict({
#     train: Dataset({
#         features: ['tweet', 'emotion'],
#         num_rows: 16000
#     })
#     validation: Dataset({
#         features: ['tweet', 'emotion'],
#         num_rows: 2000
#     })
#     test: Dataset({
#         features: ['tweet', 'emotion'],
#         num_rows: 2000
#     })
# })

# DatasetDict with remapped columns
ds = load_dataset("./datasets/emotion/", task="text_classification")
DatasetDict({
#     train: Dataset({
#         features: ['text', 'label'],
#         num_rows: 16000
#     })
#     validation: Dataset({
#         features: ['text', 'label'],
#         num_rows: 2000
#     })
#     test: Dataset({
#         features: ['text', 'label'],
#         num_rows: 2000
#     })
# })

# Dataset with default columns
ds = load_dataset("./datasets/emotion/", split="train")
# Map/cast features
ds = ds.prepare_for_task("text_classification")
# Dataset({
#     features: ['text', 'label'],
#     num_rows: 16000
# })
```

i have a few follow-up questions / remarks:

1. i'm working under the assumption that contributors / users only provide a unique set of task types. in particular, the current implementation does not support something like:
```python
task_templates=[TextClassification(labels=class_names, text_column="tweet", label_column="emotion"), TextClassification(labels=class_names, text_column="some_other_column", label_column="some_other_column")]
```
since we use `TaskTemplate.task` and the filter for compatible templates in `Dataset.prepare_for_task`. should we support these scenarios? my hunch is that this is rare in practice, but please correct me if i'm wrong.

2. when we eventually run evaluation for `transformers` models, i expect we'll be using the `Trainer` for which we can pass the standard label names to `TrainingArguments.label_names`. if that's the case, it might be prudent to heed the warning from the [docs](https://huggingface.co/transformers/main_classes/trainer.html?highlight=trainer#trainer) and use `labels` instead of `label` in the schema:
> your model can accept multiple label arguments (use the label_names in your TrainingArguments to indicate their name to the Trainer) but none of them should be named "label".

3. i plan to forge ahead on the rest of the pipeline taxonomy. please let me know if you'd prefer smaller, self-contained pull requests (e.g. one per task)

src/datasets/arrow_dataset.py
SBrandeis: Hi @lewtun! Yes that's correct

src/datasets/builder.py
```diff
@@ -813,6 +815,9 @@ def _build_single_dataset(
                         )
                     else:
                         ds.info.features = self.info.post_processed.features
+            # Rename and cast features to match task schema
+            if self.task is not None:
+                ds = ds.prepare_for_task(self.task)
```
lhoestq, line 820: Maybe let's move this to the `load_dataset` function directly ? This way you don't have to add the `task` attribute to the builder itself.

src/datasets/tasks/question_answering.py
```diff
@@ -0,0 +1,40 @@
+from typing import Dict
+
+from ..features import Features, Value
+from .base import TaskTemplate
+
+
+class QuestionAnswering(TaskTemplate):
+    task = "question_answering"
+    input_schema = Features({"question": Value("string"), "context": Value("string")})
+    label_schema = Features({"answer_start": Value("int32"), "answer_end": Value("int32")})
```
lhoestq, line 10: In general we only have `answer_text` and `answer_start`, not `answer_end` and `answer_start`

datasets/emotion/emotion.py
```diff
@@ -33,17 +34,33 @@
...
+        )
+    ]
+
     def _info(self):
         class_names = ["sadness", "joy", "love", "anger", "fear", "surprise"]
         return datasets.DatasetInfo(
             description=_DESCRIPTION,
             features=datasets.Features(
-                {"text": datasets.Value("string"), "label": datasets.ClassLabel(names=class_names)}
+                {"tweet": datasets.Value("string"), "emotion": datasets.ClassLabel(names=class_names)}
```
lhoestq, line 58: I guess you did this change to test the column renaming step of the task preparation.
Though ultimately we'll probably not rename the features of this dataset to avoid breaking changes

lhoestq (commented):
Good job @lewtun ! :) thank you

1. this looks reasonable. If it happens we could still workaround this by having separate dataset configurations
2. right indeed !
3. We can do one PR per task if you want

I also added a few comments:

datasets/emotion/emotion.py
lewtun: yes, this was just to show how it might look in practice - i'll revert the changes before the final review!

src/datasets/builder.py
lewtun: good idea!

src/datasets/tasks/question_answering.py
lewtun: good catch!

lewtun:
hey @lhoestq @SBrandeis, i think this is ready for another review 😃 

in addition to a few comments / questions i've left in the pr, here's a few remarks:

* after some experimentation, i decided against allowing the user to specify nested column names for question-answering. i couldn't find a simple solution with the current api and suspect that i'd have to touch many areas of `datasets` to "unflatten" columns in a generic fashion.
* in the current implementation, the user can specify the outer column name for question-answering, but is expected to follow the inner schema for e.g. `answers.text` and `answers.answer_start`. we can decide later how much flexibility we want to give users
* i added a few unit tests
* as discussed, let's keep this pr focused on text classification / question answering and i'll add the other tasks in separate prs

datasets/squad/squad.py
```diff
@@ -98,6 +99,9 @@ def _info(self):
             supervised_keys=None,
             homepage="https://rajpurkar.github.io/SQuAD-explorer/",
             citation=_CITATION,
+            task_templates=[
+                QuestionAnswering(),
```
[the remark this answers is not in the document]
SBrandeis: > more generally, am i correct in understanding that we'll need to add the relevant tasks to each of the ~600 canonical datasets at some point?

Yes... To make our lives easier, we need to find a way to infer supported tasks automatically for simple datasets, without changing the datasets scripts. For example, we could check if the dataset is already formatted as expected (no need for column renaming / casting).

src/datasets/info.py
```diff
@@ -189,13 +206,17 @@ def unique(values):
         features = None
         supervised_keys = None
 
+        # Can be prepared for a pipeline if all members can also be prepared for this specific pipeline
+        task_templates = None  # TODO(sbrandeis)
```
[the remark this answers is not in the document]
SBrandeis: Of course 😄 

For context, this method is called when concatenating datasets with the [`concatenate_datasets`](https://github.com/huggingface/datasets/blob/master/src/datasets/arrow_dataset.py#L3027) method. The `DatasetInfo` of the concatenated `Dataset` object is constructed using `DatasetInfo.from_merge`: https://github.com/huggingface/datasets/blob/master/src/datasets/arrow_dataset.py#L3098

What I meant here is that it would be nice to propagate the `task_templates` from the components `Dataset` objects to the resulting concatenated `Dataset` object (I'm happy to reformulate if it's not clear 😅 ). "members" refer to the components `Dataset` objects (the ones being concatenated).

If all the component `Dataset`s can be prepared for the `question_answering` task / pipeline, then the resulting concatenated `Dataset` can also be prepared for the `question_answering` task.

Let me know if this helps 😄

src/datasets/arrow_dataset.py
```diff
@@ -1384,6 +1384,32 @@ def with_transform(
...
+        """
+        tasks = [template.task for template in (self.info.task_templates or [])]
+        compatible_templates = [template for template in (self.info.task_templates or []) if template.task == task]
+        if not compatible_templates:
+            raise ValueError(f"Task {task} is not compatible with this dataset! Available tasks: {tasks}")
+
+        if len(compatible_templates) > 1:
+            raise ValueError(
+                f"""Expected 1 task template but found {len(compatible_templates)}! Please ensure that \
+                :attr:`datasets.DatasetInfo.task_templates` contains a unique set of task types."""
```
SBrandeis, line 1403: I tend to prefer the single quote syntax but that's a matter of personal taste
```suggestion
                f"Expected 1 task template but found {len(compatible_templates)}! Please ensure that "
                ":attr:`datasets.DatasetInfo.task_templates` contains a unique set of task types."
```

src/datasets/info.py
```diff
@@ -108,6 +109,7 @@ class DatasetInfo:
         post_processing_size (int, optional):
         dataset_size (int, optional):
         size_in_bytes (int, optional):
+        task_templates (List[TaskTemplate], optional): TODO(sbrandeis) - document this
```
SBrandeis, line 112: @lhoestq should document this attribute of the DatasetInfo?

datasets/squad/squad.py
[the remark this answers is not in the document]
lewtun: good point re task detection! i wonder if we could do this with a simple check on whether the schema features are a subset of the dataset, e.g.

```python
squad = load_dataset("./datasets/squad", split="train")
qa = QuestionAnswering()
schema = Features({**qa.input_schema, **qa.label_schema})
assert all(item in squad.features.items() for item in schema.items())
```

src/datasets/tasks/base.py
```diff
@@ -0,0 +1,26 @@
+import abc
+from dataclasses import dataclass
+from typing import Dict
+
+from ..features import Features
+
+
+@dataclass
```
SBrandeis, line 8: I'm not sure about this being a `dataclass`

I'd rather have `task`, `input_schema` and `label_schema` be class attributes, so they can be shared by all instances of the `TaskTemplate` concrete class.

On the other hand, for the classification task, we need the label_featrues to be instance specific 🤔

src/datasets/tasks/text_classification.py
```diff
@@ -0,0 +1,45 @@
...
+    def __init__(
+        self,
+        labels,
+        text_column="text",
+        label_column="labels",
+    ):
+        assert sorted(set(labels)) == sorted(labels), "Labels must be unique"
+        self.ordered_labels = sorted(labels)
+        self.label_schema["labels"] = ClassLabel(names=self.ordered_labels)
+        self.labels = labels
```
SBrandeis, line 26: Is this needed if we have `ordered_labels` ?

SBrandeis (commented):
Awesome work @lewtun 🎉 Thanks a lot!
I have a few comments and suggestions

src/datasets/arrow_dataset.py
lewtun: good idea! reverted to single quotes :)

src/datasets/info.py
lewtun: looking at the [docs](https://huggingface.co/docs/datasets/master/package_reference/main_classes.html?highlight=datasetinfo#datasetinfo) it seems that none of the `DatasetInfo` attributes are documented. perhaps this is best addressed in a separate PR?

src/datasets/tasks/text_classification.py
lewtun: good catch!

src/datasets/tasks/text_classification.py
```diff
@@ -0,0 +1,45 @@
...
+        self.text_column = text_column
+        self.label_column = label_column
+        # TODO(lewtun): consider adding id2label mapping if we need these attributes for instatiating models
+        self.label2id = {label: idx for idx, label in enumerate(self.ordered_labels)}
```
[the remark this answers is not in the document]
lewtun: i think we want a `dict` for `id2label` so i've added 

```
 self.id2label = {idx: label for label, idx in self.label2id.items()}
```

tests/test_arrow_dataset.py
```diff
@@ -1850,6 +1851,89 @@ def test_with_transform(self, in_memory):
...
+                "input_text": Value("string"),
+                "input_labels": Value("int32"),
+            }
+        )
+        self.features_after_cast = Features(
+            {
+                "text": Value("string"),
+                "labels": ClassLabel(names=self.labels),
+            }
+        )
```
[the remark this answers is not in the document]
lewtun: ugh, a remnant from my initial testing - good catch!

src/datasets/tasks/text_classification.py
lewtun: actually i think we do need `labels` as an attribute of `TextClassification` because we rely on having access to it when we map between the class and dict (see the `from_dict` method).

tests/test_arrow_dataset.py
```diff
@@ -1850,6 +1851,89 @@ def test_with_transform(self, in_memory):
...
+                "context": Value("string"),
+                "question": Value("string"),
+                "answers": Sequence(
+                    {
+                        "text": Value("string"),
+                        "answer_start": Value("int32"),
+                    }
+                ),
+            }
+        )
```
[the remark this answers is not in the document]
lewtun: good catch!

tests/test_tasks.py
```diff
@@ -0,0 +1,54 @@
...
+                "answers": Sequence(
+                    {
+                        "text": Value("string"),
+                        "answer_start": Value("int32"),
+                    }
+                )
+            }
+        )
+        template_dict = {
+            "context_column": "input_input_contexttext",
```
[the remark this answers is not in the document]
lewtun: oops, my IDE's autocomplete got a bit carried away here :)

tests/test_arrow_dataset.py
```diff
@@ -1850,6 +1851,89 @@ def test_with_transform(self, in_memory):
...
+        data = {"input_text": ["i love transformers!"], "input_labels": [1]}
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            dset = Dataset.from_dict(data, info=info)
+            dset = self._to(in_memory, tmp_dir, dset)
+            self.assertSetEqual(set(["input_text", "input_labels"]), set(dset.column_names))
+            self.assertDictEqual(self.features_before_cast, dset.features)
+            dset = dset.prepare_for_task(task="text-classification")
+            self.assertSetEqual(set(["labels", "text"]), set(dset.column_names))
+            self.assertDictEqual(self.features_after_cast, dset.features)
+            del dset
```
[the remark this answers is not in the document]
lewtun: thanks for the info! i just worked from one of the existing tests, but agree it's better to use a context manager - i've now added this :)

tests/test_arrow_dataset.py
```diff
@@ -1850,6 +1851,83 @@ def test_with_transform(self, in_memory):
...
+            task_templates=QuestionAnswering(
+                context_column="input_context", question_column="input_question", answers_column="input_answers"
+            ),
+        )
+        data = {
+            "input_context": ["huggingface is going to the moon!"],
+            "input_question": ["where is huggingface going?"],
+            "input_answers": [{"text": ["to the moon!"], "answer_start": [2]}],
+        }
+        with tempfile.TemporaryDirectory() as tmp_dir, Dataset.from_dict(data, info=info) as dset:
```
lewtun, line 1917: for some reason, using a context manager for `dset` produces an error on the windows tests:

```
tests\test_arrow_dataset.py:1879: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
C:\tools\miniconda3\lib\tempfile.py:809: in __exit__
    self.cleanup()
C:\tools\miniconda3\lib\tempfile.py:813: in cleanup
    _shutil.rmtree(self.name)
C:\tools\miniconda3\lib\shutil.py:500: in rmtree
    return _rmtree_unsafe(path, onerror)
C:\tools\miniconda3\lib\shutil.py:395: in _rmtree_unsafe
    onerror(os.unlink, fullname, sys.exc_info())
```

src/datasets/info.py
```diff
@@ -189,13 +206,18 @@ def unique(values):
         features = None
         supervised_keys = None
 
+        # Find common task templates across all dataset infos
+        all_task_templates = [info.task_templates for info in dataset_infos]
+        task_templates = list(set(all_task_templates[0]).intersection(*all_task_templates[1:]))
```
lewtun, line 211: @SBrandeis this is a first pass at trying to find common members across task templates. 

since the task template classes are unhashable by default, i defined my own `__hash__` functions for `TextClassification` and `QuestionAnswering`. i'm not really happy with this approach (it feels clunky and prone to failure), so if you have a better idea on computing the intersection across task templates then i'm all ears!

src/datasets/info.py
SBrandeis: Fine by me :)

lewtun: i created an issue for this in #2354 and will tackle it on friday :)

tests/test_arrow_dataset.py
lewtun: update: i've now added context management for all `dset` objects in the tests. hopefully the tests will now pass 🤞

src/datasets/tasks/base.py
SBrandeis: Actually we can use dataclasses AND class attributes together, if we type hint the class attributes with `ClassVar`:

```python
from typing import ClassVar

# [...]

@dataclass
class TaskTemplate(abc.ABC):
    task: ClassVar[str]
    input_schema: ClassVar[Features]
# [...]
```

The attributes marked as `ClassVar` will be ignored by the "dataclass mechanism" (left out of `__init__`, `__eq__`, etc)

Source: https://docs.python.org/3/library/dataclasses.html#class-variables

src/datasets/tasks/base.py
```diff
@@ -0,0 +1,30 @@
...
+        return NotImplemented
+
+    @classmethod
+    @abc.abstractmethod
+    def from_dict(cls, template_dict: dict) -> "TaskTemplate":
+        return NotImplemented
+
+    @abc.abstractmethod
+    def __hash__(self):
+        raise NotImplementedError
```
SBrandeis, line 30: I'd rather mark the dataclass as "frozen": it would generate the `__hash__` method for us (see https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass, the paragraph about `hash`)

I think it makes sense to mark a TaskTemplate as immutable, it should only be specified once when instantiating the `DatasetInfo` object.

```python
@dataclass(frozen=True)
class TaskTemplate(abc.ABC):
    # [...]
```

src/datasets/tasks/text_classification.py
SBrandeis: Maybe we should sort the labels in-place then?

I think two `TextClassification` tasks with the same set of labels should be perfectly equivalent. WDYT ?

SBrandeis: Maybe we should rename this method `__post_init__` in order not to interfere with the `dataclass` generated `__init__` method. See : https://docs.python.org/3/library/dataclasses.html#post-init-processing

src/datasets/info.py
```diff
@@ -189,13 +206,23 @@ def unique(values):
...
         supervised_keys = None
 
+        # Find common task templates across all dataset infos
+        all_task_templates = [info.task_templates for info in dataset_infos if info.task_templates is not None]
+        if len(all_task_templates) > 1:
+            task_templates = list(set(all_task_templates[0]).intersection(*all_task_templates[1:]))
+        elif len(all_task_templates):
+            task_templates = list(set(all_task_templates[0]))
+        else:
+            task_templates = None
```
SBrandeis, line 216: I'd love to see this unit tested, to make sure it actually does what we want, ie select the TaskTemplate(s) that are present in every DatasetInfo.

src/datasets/info.py
SBrandeis: Using the hash is a good idea IMO if we are certain two TaskTemplates with the same attributes have the same hash!

I believe we can achieve that by decorating the classes with `@dataclass(frozen=True)`

src/datasets/info.py
lewtun: great idea! i've added some tests to cover the various scenarios that can arise

src/datasets/tasks/base.py
lewtun: i did not know about `ClassVar` so thank you for the suggestion and pointed to the docs! i've now added this

src/datasets/arrow_dataset.py
```diff
@@ -1384,6 +1384,31 @@ def with_transform(
         dataset.set_transform(transform=transform, columns=columns, output_all_columns=output_all_columns)
         return dataset
 
+    def prepare_for_task(self, task: str) -> "Dataset":
+        """Prepare a dataset for the given task.
+
+        Casts :attr:`datasets.DatasetInfo.features` according to a task-specific schema.
```
lhoestq, line 1390: Let's have a list of the supported tasks here ?

src/datasets/tasks/question_answering.py
```diff
@@ -0,0 +1,44 @@
...
+                    "answer_start": Value("int32"),
+                }
+            )
+        }
+    )
+    question_column: str
+    context_column: str
+    answers_column: str
+
+    def __init__(self, question_column="question", context_column="context", answers_column="answers"):
```
lhoestq, line 26: any reason to do have the default values in `__init__` instead this ?
```python
    question_column: str = "question"
    context_column: str = "context"
    answers_column: str = "answers"
```

src/datasets/tasks/base.py
lewtun: i've had a stab at this and am running into an issue that you might know the answer to :)

first, if we mark `TaskTemplate` as immutable with `frozen=True` than any subclass must also be marked as immutable (we get a `TypeError` otherwise)

with that in mind, let's modify our `TextClassification` template as follows:

```python
@dataclass(frozen=True)
    task = "text-classification"
    input_schema = Features({"text": Value("string")})
    label_schema = Features({"labels": ClassLabel})
    labels: List[str]
    text_column: str
    label_column: str

    def __init__(
        self,
        labels,
        text_column="text",
        label_column="labels",
    ):
        assert sorted(set(labels)) == sorted(labels), "Labels must be unique"
        object.__setattr__(self, "labels", sorted(labels))
        object.__setattr__(self, "text_column", text_column)
        object.__setattr__(self, "label_column", label_column)
        self.label_schema["labels"] = ClassLabel(names=self.labels)
        object.__setattr__(self, "label2id", {label: idx for idx, label in enumerate(self.labels)})
        object.__setattr__(self, "id2label", {idx: label for label, idx in self.label2id.items()})
```

note that the use of `object.__setattr__` is needed to initialize the fields (see [docs](https://docs.python.org/3/library/dataclasses.html#frozen-instances)).

the problem i'm running into is that `labels` is of type `list` and thus not hashable by default. i can exclude it from the comparison methods like `__eq__` by defining:

```python
from dataclasses import field 

@dataclass(frozen=True)
    task = "text-classification"
    input_schema = Features({"text": Value("string")})
    label_schema = Features({"labels": ClassLabel})
    labels: List[str] = field(compare=False)
    text_column: str
    label_column: str
```
but this isn't ideal because now the hash for two `TextClassifcation` instances with different labels, but identical column names is the same:

```python
tc1 = TextClassification(labels=["pos", "neg"])
tc2 = TextClassification(labels=["pos", "neg", "neutral"])
# returns True
hash(tc1) == hash(tc2)
```

do you happen to know of an elegant way to incorporate the information in `labels` into the dataclass' `__hash__` function? in my original implementation i used `frozenset(self.labels)` but could not find a nice way to generate this behaviour automagically from the dataclass.

src/datasets/tasks/text_classification.py
```diff
@@ -0,0 +1,48 @@
...
+    task = "text-classification"
+    input_schema = Features({"text": Value("string")})
+    label_schema = Features({"labels": ClassLabel})
+    labels: List[str]
+    text_column: str
+    label_column: str
+
+    def __init__(
+        self,
+        labels,
```
lhoestq, line 19: Models have a config with label2id. And we have the same for datasets with the ClassLabel feature type. At one point either the model or the dataset must sync with the other. It would be great to do that on the dataset side.

For example for sentiment classification on amazon reviews with you could have these labels:
- "1 star", "2 stars", "3 stars", "4 stars", "5 stars"
- "1", "2", "3", "4", "5"

Some models may use the first set, while other models use the second set.

Here in the `TextClassification` class, the user can only specify one set of labels, while many models could actually be compatible but have different sets of labels. Should we allow users to pass a list of compatible labels sets ?

Then in terms of API, users could use `dataset.prepare_for_task("text-classification", labels=model.labels)` or something like that.

The label set could also be the same but not in the same order. For NLI for example, some models use `["neutral", "entailment", "contradiction"]` and some others use `["neutral", "contradiction", "entailment"]`, so we should take care of updating the order of the labels in the dataset to match the labels order of the model.

Let me know what you think ! This can be done in a future PR

src/datasets/tasks/base.py
lhoestq: Maybe you can use a tuple instead of a list ?

lhoestq (commented):
I love the design and how it's integrated in the DatasetInfo for json serialization :)

I added a few comments. One of them about having several label sets could be taken care of in a future PR

src/datasets/tasks/base.py
lhoestq: Maybe you can use a tuple instead of a list ?

lewtun: resolved by @SBrandeis' great idea to cast `labels` to a tuple 🚀

src/datasets/tasks/text_classification.py
lewtun: good idea! done :)

src/datasets/arrow_dataset.py
lewtun: good idea! i've added the list of tasks under the `task` arg since it felt a bit more natural to be there

src/datasets/tasks/question_answering.py
lewtun: good catch! this is now fixed with the new `__post_init__` usage

src/datasets/tasks/text_classification.py
lewtun: this is a great point - thanks for raising it! i think it's best tackled in a separate PR so i've opened a separate issue for it here: #2359

lewtun:
i'm not sure why the benchmarks are getting cancelled - is this expected?

lhoestq:
> i'm not sure why the benchmarks are getting cancelled - is this expected?

Hmm I don't know. It's certainly unrelated to this PR though. Maybe github has some issues

SBrandeis:
Something is happening with actions: https://www.githubstatus.com/

lewtun:
hey @lhoestq and @SBrandeis, i've extended the `prepare_for_task` API along the lines that @lhoestq suggested and added a few new tests to check that we trigger the value errors on incorrect input.

i think this is ready for another review :)

tests/test_arrow_dataset.py
```diff
@@ -1850,6 +1924,133 @@ def test_with_transform(self, in_memory):
...
+        )
+        task = TextClassification(text_column="input_text", label_column="input_labels", labels=labels)
+        info = DatasetInfo(
+            features=features,
+            task_templates=task,
+        )
+        data = {"input_text": ["i love transformers!"], "input_labels": [1]}
+        with tempfile.TemporaryDirectory() as tmp_dir, Dataset.from_dict(data, info=info) as dset:
+            with self._to(in_memory, tmp_dir, dset) as dset:
+                with self._to(in_memory, tmp_dir, dset) as dset:
```
lhoestq, line 2044: calling `self._to` once is enough ;)

lhoestq (approved):
Looks all good thank you :)

Can you also add `prepare_for_task` in the `main_classes.rst` file of the documentation ?

tests/test_arrow_dataset.py
lewtun: good catch!

lewtun:
> Looks all good thank you :)
> 
> Can you also add `prepare_for_task` in the `main_classes.rst` file of the documentation ?

Done! I also remembered that I needed to do the same for `DatasetDict`, so included this as well :)

src/datasets/load.py
```diff
@@ -635,6 +635,7 @@ def load_dataset(
     save_infos: bool = False,
     script_version: Optional[Union[str, Version]] = None,
     use_auth_token: Optional[Union[bool, str]] = None,
+    task: Optional[str] = None,
```
SBrandeis, line 638: Should we add `TaskTempplate` here too? 👀

src/datasets/tasks/text_classification.py
```diff
@@ -0,0 +1,40 @@
...
+
+from ..features import ClassLabel, Features, Value
+from .base import TaskTemplate
+
+
+@dataclass(frozen=True)
+class TextClassification(TaskTemplate):
+    task = "text-classification"
+    input_schema = Features({"text": Value("string")})
+    label_schema = Features({"labels": ClassLabel})
```
SBrandeis, line 12: Nit: Since we update this in `__post_init__` do we need to declare a default?

SBrandeis (commented):
Awesome job @lewtun, thanks a lot!

src/datasets/load.py
lewtun: good catch!

src/datasets/tasks/text_classification.py
lewtun: good catch! i'll fix it :)

lewtun: ah actually we need to declare a default because `label_schema` is a required argument for `__init__`. i've reverted the change and added a TODO so we can think about a more elegant approach as we iterate on the other tasks

src/datasets/arrow_dataset.py
```diff
@@ -1384,6 +1385,44 @@ def with_transform(
         dataset.set_transform(transform=transform, columns=columns, output_all_columns=output_all_columns)
         return dataset
 
+    def prepare_for_task(self, task: Union[str, TaskTemplate]) -> "Dataset":
+        """Prepare a dataset for the given task.
```
thomwolf, line 1389: I think it's good to explain a bit what are the operations contained in the expression "prepare a dataset"

src/datasets/arrow_dataset.py
```diff
@@ -1384,6 +1385,44 @@ def with_transform(
...
+
+        Casts :attr:`datasets.DatasetInfo.features` according to a task-specific schema.
+
+        Args:
+            task (:obj:`Union[str, TaskTemplate]`): The task to prepare the dataset for during training and evaluation. If :obj:`str`, supported tasks include:
+
+                - :obj:`"text-classification"`
+                - :obj:`"question-answering"`
```
thomwolf, line 1397: `question-answering` exists in two forms: abstractive and extractive question answering.

we can keep a generic `question-answering` but then it will probably mean diferrent schema of input/output for both (abstractive will have text for both while extractive can use spans indication as well as text).

Or we can also propose to use `abstractive-question-answering` and `extractive-question-answering` for instance.

thomwolf: Maybe we could have `question-answering-abstractive` and `question-answering-extractive` if somehow we can use a for a completion or search in the future (detail).

src/datasets/load.py
```diff
@@ -694,6 +696,7 @@ def load_dataset(
               You can specify a different version that the default "main" by using a commit sha or a git tag of the dataset repository.
         use_auth_token (``str`` or ``bool``, optional): Optional string or boolean to use as Bearer token for remote files on the Datasets Hub.
             If True, will get token from `"~/.huggingface"`.
+        task (``str``): The task to prepare the dataset for during training and evaluation. Casts the dataset's :class:`Features` according to one of the schemas in `~tasks`.
```
thomwolf, line 699:
```suggestion
        task (``str``): The task to prepare the dataset for during training and evaluation. Casts the dataset's :class:`Features` according to standardized column names and types as detailed in `~tasks`.
```

src/datasets/tasks/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+from .text_classification import TextClassification
+
+
+__all__ = ["TaskTemplate", "QuestionAnswering", "TextClassification"]
+
+
+NAME2TEMPLATE = {QuestionAnswering.task: QuestionAnswering, TextClassification.task: TextClassification}
+
+
+def task_template_from_dict(task_template_dict: dict) -> Optional[TaskTemplate]:
```
thomwolf, line 14: Is this function supposed to be used by the user? If so it should have a doc string and doc

src/datasets/tasks/question_answering.py
```diff
@@ -0,0 +1,41 @@
+from dataclasses import dataclass
+from typing import Dict
+
+from ..features import Features, Sequence, Value
+from .base import TaskTemplate
+
+
+@dataclass(frozen=True)
+class QuestionAnswering(TaskTemplate):
```
thomwolf, line 9: `QuestionAnsweringExtractive` ?

src/datasets/tasks/question_answering.py
```diff
@@ -0,0 +1,41 @@
...
+    task = "question-answering"
+    input_schema = Features({"question": Value("string"), "context": Value("string")})
```
thomwolf, line 11: Maybe you want to check with a few QA datasets that this schema make sense. Typically NaturalQuestions, TriviaQA and can be good second datasets to compare to and be sure of the generality of the schema.

A good recent list of QA datasets to compare the schemas among, is for instance in the UnitedQA paper: https://arxiv.org/abs/2101.00178

src/datasets/arrow_dataset.py
thomwolf: Actually I see that people are more organizing in terms of general and sub-tasks, for instance on paperwithcode: https://paperswithcode.com/area/natural-language-processing and on nlpprogress: https://github.com/sebastianruder/NLP-progress/blob/master/english/question_answering.md#squad

Probably the best is to align with one of these in terms of denomination, PaperWithCode is probably the most active and maintained and we work with them as well.

thomwolf (commented):
Added a few high-level comment.

I like the general idea!

src/datasets/arrow_dataset.py
lewtun: good idea! i opted for the following:

> Prepare a dataset for the given task by casting the dataset's :class:`Features` to standardized column names and types as detailed in `~tasks`.

src/datasets/arrow_dataset.py
lewtun: our idea was to start by following the pipeline taxonomy in `transformers`, where `question-answering` is indeed restricted to the extractive case. 

but you're 100% correct that we should already start considering the abstractive case or grouping in terms of the sub-domains that PwC adopts. 

since our focus right now is on getting `text-classification` integrated and tested with `autonlp`, i've opened an issue  here to tackle this next: https://github.com/huggingface/datasets/issues/2371

lhoestq: PWC uses "Question Answering" (meaning extractive) vs "Generative Question Answering" (includes abstractive) which encapsulates most of the QA tasks except Open/Closed Domain QA and Knowledge base QA (they require no context since the knowledge is not part of the query).

I'm fine with these names, or simply extractive and abstractive

src/datasets/tasks/question_answering.py
lewtun: thanks for the tip! added to #2371

src/datasets/tasks/question_answering.py
lewtun: thanks for the suggestion! i'll keep this unchanged for now since it will soon be overhauled by #2371

src/datasets/tasks/__init__.py
lewtun: no this is not supposed to be used by the user (at least for now). still no harm in having a docstring, so i've added:

>     """Create one of the supported task templates in :obj:`datasets.tasks` from a dictionary."""

PR #2257, opened 2021-04-24
added metrics for CUAD

bhavitvyamalik:
For now I've added F1, AUPR, Precision at 80% recall, and Precision at 90%. Last 3 metrics were reported in the [paper](https://arxiv.org/pdf/2103.06268.pdf). Please let me know if we require `exact_match` metric too here


MohammedRakib:
> For now I've added F1, AUPR, Precision at 80% recall, and Precision at 90%. Last 3 metrics were reported in the [paper](https://arxiv.org/pdf/2103.06268.pdf). Please let me know if we require `exact_match` metric too here

@bhavitvyamalik  I guess the mentioned metrics are enough but it would be better if exact match is also added since the standard SQUAD dataset also has it.

bhavitvyamalik:
I'm a little confused when it comes to 2 ground truths which can be a possible answer. Like here for eg.

`predictions = [{'prediction_text': ['The seller:', 'The buyer/End-User: Shenzhen LOHAS Supply Chain Management Co., Ltd.'], 'id': 'LohaCompanyltd_20191209_F-1_EX-10.16_11917878_EX-10.16_Supply Agreement__Parties'}]`

` references = [{'answers': {'answer_start': [143, 49], 'text': ['The seller:', 'The buyer/End-User: Shenzhen LOHAS Supply Chain Management Co., Ltd.']}, 'id': 'LohaCompanyltd_20191209_F-1_EX-10.16_11917878_EX-10.16_Supply Agreement__Parties'}]`

Should I ensure at least 1 predicted answer matches or both predicted answers should match (like in this case) for EM to be 1?

MohammedRakib:
I would like to quote it from the website that I am following to learn
these things.
Exact Match:
This metric is as simple as it sounds. For each question+answer pair, if
the characters of the model's prediction exactly match the characters of *(one
of) the True Answer(s)*, EM = 1, otherwise EM = 0. This is a strict
all-or-nothing metric; being off by a single character results in a score
of 0. When assessing against a negative example, if the model predicts any
text at all, it automatically receives a 0 for that example.

So, I guess you need to ensure at least 1 predicted answer matches for EM
to be 1.
Source:
https://qa.fastforwardlabs.com/no%20answer/null%20threshold/bert/distilbert/exact%20match/f1/robust%20predictions/2020/06/09/Evaluating_BERT_on_SQuAD.html

You can go to their homepage and read the other links. They have detailed
explanations on evaluation metrics. You can also have a look at the
squad_v2 metric file for further clarification.

Regards,
Mohammed Rakib

MohammedRakib:
I would like to quote it from the website that I am following to learn
these things.
Exact Match:
This metric is as simple as it sounds. For each question+answer pair, if
the characters of the model's prediction exactly match the characters of
*(one
of) the True Answer(s)*, EM = 1, otherwise EM = 0. This is a strict
all-or-nothing metric; being off by a single character results in a score
of 0. When assessing against a negative example, if the model predicts any
text at all, it automatically receives a 0 for that example.

So, I guess you need to ensure at least 1 predicted answer matches for EM
to be 1.
Source:
https://qa.fastforwardlabs.com/no%20answer/null%20threshold/bert/distilbert/exact%20match/f1/robust%20predictions/2020/06/09/Evaluating_BERT_on_SQuAD.html

You can go to their homepage and read the other links. They have detailed
explanations on evaluation metrics. You can also have a look at the
squad_v2 metric file for further clarification.

Regards,
Mohammed Rakib

bhavitvyamalik:
Updated the same @MohammedRakib! Even if a single answer matches I'm returning 1 in that case for EM (not traversing all predictions once we have one `exact_match` from prediction)

metrics/cuad/cuad.py
```diff
@@ -0,0 +1,117 @@
...
+    def _info(self):
+        return datasets.MetricInfo(
+            description=_DESCRIPTION,
+            citation=_CITATION,
+            inputs_description=_KWARGS_DESCRIPTION,
+            features=datasets.Features(
+                {
+                    "predictions": {
+                        "id": datasets.Value("string"),
+                        "prediction_text": datasets.features.Sequence(datasets.Value("string")),
```
lhoestq, line 82:
```suggestion
                        "prediction_text": datasets.Value("string"),
```
There should only be one prediction per question right ? Or is this something specific to CUAD ?

metrics/cuad/evaluate.py
```diff
@@ -0,0 +1,207 @@
+""" Official evaluation script for CUAD dataset. """
```
lhoestq, line 1: Did you have to do some modifications to this script ?
If so please add comments to show the lines you modified

lhoestq (commented):
Really cool thanks !

metrics/cuad/cuad.py
bhavitvyamalik: I looked at the test samples in the dataset and found that few examples have multiple answers. In order to accomodate those answers too I changed it to `Sequence`

metrics/cuad/evaluate.py
bhavitvyamalik: Most the things were kept same (as reported in the paper). Only `exact_match` metric was added

metrics/cuad/cuad.py
bhavitvyamalik: Should I add a comment in the code too? In the docs though I've mentioned that 

> predictions: List of question-answers dictionaries with the following key-values:
>         - 'id': id of the question-answer pair as given in the references (see below)
>         - 'prediction_text': list of possible texts for the answer, as a list of strings

lhoestq: > I looked at the test samples in the dataset and found that few examples have multiple answers. In order to accomodate those answers too I changed it to Sequence

If ome examples have multiple answers then it means that the `references` feature should allow for multiple reference answers (which is currently the case since you used Sequence in `references`). Not sure why you'd need a Sequence for `prediction_text` ?

bhavitvyamalik: We'll need all possible predictions to calculate precision and recall. Say if I use only one prediction_text when my references are multiple then it'll increase my `false_negatives` and I'll get low value of `recall` according to the evaluation script

lhoestq: Usually when someone evaluates a model, there's only one model prediction per question right ? Is this different for CUAD ?

bhavitvyamalik: One can get multiple or single prediction depending on `conf` score [here (official evaluation script)](https://github.com/TheAtticusProject/cuad/blob/4e6b4f3a938dbc0498f800cb50046fe5f207bfcc/evaluate.py#L27). This is how they are building the `prediction_text` to evaluate precision and recall.

lhoestq: Ok that makes sense !! Maybe mention in the docstring that users can use multiple answers depending on a threshold on the confidence probability of each prediction ?

PR #2258, opened 2021-04-25
Fix incorrect update_metadata_with_features calls in ArrowDataset

mariosasko:
Fixes bugs in the `unpdate_metadata_with_features` calls (caused by changes in #2151)


lhoestq (approved):
Good catch ! thank you

mariosasko:
@lhoestq Maybe a test that runs the functions that call `update_metadata_with_features` and checks if metadata was updated would be nice to prevent this from happening in the future.

PR #2259, opened 2021-04-25
Add support for Split.ALL

mariosasko:
The title says it all.


src/datasets/arrow_reader.py
```diff
@@ -105,6 +105,8 @@ def make_file_instructions(name, split_infos, instruction, filetype_suffix=None)
     """
     name2len = {info.name: info.num_examples for info in split_infos}
     if not isinstance(instruction, ReadInstruction):
+        if str(instruction) == "all":
+            instruction = "+".join(name2len.keys())
```
lhoestq, line 109: we may need to sort the keys here to make the order deterministic

lhoestq (commented):
Awesome thank you !
I left one minor comment

src/datasets/arrow_reader.py
mariosasko: `dict` preserves insertion order in Python 3.6+, so the ordering of the keys is deterministic (this is an implementation detail in CPython in 3.6 and was documented in 3.7).

mariosasko:
Honestly, I think we should fix some other issues in Split API before this change. E. g. currently the following will not work, even though it should:
```python
import datasets
datasets.load_dataset("sst", split=datasets.Split.TRAIN+datasets.Split.TEST)  # AssertionError
```

lhoestq (approved):
Right, merging this one then :)
thank you !

PR #2261, opened 2021-04-25
Improve ReadInstruction logic and update docs

mariosasko:
Improve ReadInstruction logic and docs.


docs/source/splits.rst
```diff
@@ -77,11 +80,15 @@ Examples using the ``ReadInstruction`` API (equivalent as above):
     train_10_20_ds = datasets.load_dataset('bookcorpus', split=datasets.ReadInstruction(
         'train', from_=10, to=20, unit='abs'))
 
-    # The first 10% of train split.
+    # The first 10% of `train` split.
     train_10_20_ds = datasets.load_dataset('bookcorpus', split=datasets.ReadInstruction(
         'train', to=10, unit='%'))
+    
+    # The first 10% of `train` split with the `pct1_dropremainder` rounding.
```
lhoestq, line 87: Maybe explain the difference between the "closest" rounding and the "" rounding ?
From the docstrings:
```
The rounding behaviour to use when percent slicing is used. Ignored when slicing with absolute indices.
Possible values:
- 'closest' (default): The specified percentages are rounded to the closest value. Use this if you want specified 
    percents to be as much exact as possible.
- 'pct1_dropremainder': the specified percentages are treated as multiple of 1%. Use this option if you want 
    consistency. Eg: len(5%) == 5 * len(1%).
    Using this option, one might not be able to use the full set of examples, if the number of those is not a multiple of 100.
```

lhoestq (commented):
Nice thanks !
It's indeed better this way.

Maybe let's describe the difference between the two rounding strategies in the doc ? (see my comment)

docs/source/splits.rst
mariosasko: Yes, will do.

mariosasko:
Ready for the final review

lhoestq (approved):
Thanks ! LGTM :)

Issue #2256, opened 2021-04-24
Running `datase.map` with `num_proc > 1` uses a lot of memory


roskoN:
## Describe the bug
Running `datase.map` with `num_proc > 1`  leads to a tremendous memory usage that requires swapping on disk and it becomes very slow.

## Steps to reproduce the bug
```python
from datasets import load_dataset

dstc8_datset = load_dataset("roskoN/dstc8-reddit-corpus", keep_in_memory=False)


def _prepare_sample(batch):
    return {"input_ids": list(), "attention_mask": list()}


for split_name, dataset_split in list(dstc8_datset.items()):
    print(f"Processing {split_name}")
    encoded_dataset_split = dataset_split.map(
        function=_prepare_sample,
        batched=True,
        num_proc=4,
        remove_columns=dataset_split.column_names,
        batch_size=10,
        writer_batch_size=10,
        keep_in_memory=False,
    )
    print(encoded_dataset_split)

    path = f"./data/encoded_{split_name}"

    encoded_dataset_split.save_to_disk(path)
```

## Expected results
Memory usage should stay within reasonable boundaries.

## Actual results
This is htop-output from running the provided script.

![image](https://user-images.githubusercontent.com/8143425/115954836-66954980-a4f3-11eb-8340-0153bdc3a475.png)

## Versions
```
- Datasets: 1.6.0
- Python: 3.8.8 (default, Apr 13 2021, 19:58:26)
[GCC 7.3.0]
- Platform: Linux-4.19.128-microsoft-standard-x86_64-with-glibc2.10
```
Running on WSL2

lhoestq:
Thanks for reporting ! We are working on this and we'll do a patch release very soon.

lhoestq:
We did a patch release to fix this issue.
It should be fixed in the new version 1.6.1

Thanks again for reporting and for the details :)

PR #2264, opened 2021-04-26
Fix memory issue in multiprocessing: Don't pickle table index

lhoestq:
The table index is currently being pickled when doing multiprocessing, which brings all the record batches of the dataset in memory.

I fixed that by not pickling the index attributes. Therefore each process has to rebuild the index when unpickling the table.

Fix issue #2256

We'll do a patch release asap !


lhoestq:
The code quality check is going to be fixed by #2265

src/datasets/table.py
```diff
@@ -145,31 +145,30 @@ def __deepcopy__(self, memo: dict):
...
+            table = self.table
             with tempfile.NamedTemporaryFile("wb", delete=False, suffix=".arrow") as tmp_file:
                 filename = tmp_file.name
                 logger.debug(
                     f"Attempting to pickle a table bigger than 4GiB. Writing it on the disk instead at {filename}"
                 )
                 _write_table_to_file(table=table, filename=filename)
-                state["path"] = filename
-                return state
+                return {"path": filename}
```
albertvillanova, line 158: Are there other items (previously in state) that we are missing now? Or just with `path` is enough?

albertvillanova (approved):
I guess the source of the problem was `state.copy()` and `self.__dict__.copy()`...

I am just wondering if we are missing some items in the state dict... Or is ti enough just with `path`, `table`, `replays` and `blocks`...

We should also implement regression tests (either in this PR or in another).

src/datasets/table.py
lhoestq: We only need "path" in this case:

- There were other items yes: _batches, _schema and _offsets. However we don't want to pickle _batches or it causes memory issues. Therefore all these attributes are reloaded when Table.__init__ is called.

- Then the only thing left to pickle is `table`. But if it's bigger than 4GB we are writing it on disk since pickle doesn't support pickling objects bigger than 4GB by default. In this case we just pickle the path to the table on disk

lhoestq:
The memory issue didn't come from `self.__dict__.copy()` but from the fact that this dic contains `_batches` which has all the batches of the table in it.
Therefore for a MemoryMappedTable all the data in `_batches` were copied in memory when pickling and this is the issue.

lhoestq:
I'm still investigating why we didn't catch this issue in the tests.
This test should have caught it but didn't:

https://github.com/huggingface/datasets/blob/3db67f5ff6cbf807b129d2b4d1107af27623b608/tests/test_table.py#L350-L353

lhoestq:
I'll focus on the patch release and fix the test in another PR after the release

PR #2268, opened 2021-04-27
Don't use pyarrow 4.0.0 since it segfaults when casting a sliced ListArray of integers

lhoestq:
This test `tests/test_table.py::test_concatenation_table_cast` segfaults with the latest update of pyarrow 4.0.0.
Setting `pyarrow<4.0.0` for now. I'll open an issue on JIRA once I know more about the origin of the issue

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ jobs:
             - run: source venv/bin/activate
             - run: pip install .[tests]
             - run: pip install -r additional-tests-requirements.txt --no-deps
-            - run: pip install pyarrow --upgrade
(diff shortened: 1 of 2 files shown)


albertvillanova:
@lhoestq note that the segfault also occurs on Linux.

lhoestq:
Created the ticket at
https://issues.apache.org/jira/browse/ARROW-12568

symeneses:
@lhoestq the ticket you mentioned is now in state resolved. Pyarrow supports AArch64 after version 4.0.0. Because of this restriction `datasets` is not installing in AArch64 systems.

PR #2270, opened 2021-04-27
Fix iterable interface expected by numpy

albertvillanova:
Numpy expects the old iterable interface with `__getitem__` instead of `__iter__`.


lhoestq:
It's been fixed in this commit: https://github.com/huggingface/datasets/commit/549110e08238b3716a5904667095fb003acda54e

Basically #2246 broke querying an index with a simple iterable.
With the fix, it's again possible to use iterables and we can keep RandIter as it is.

Closing since the fix is already on master

PR #2273, opened 2021-04-27
Added CUAD metrics

bhavitvyamalik:
`EM`, `F1`, `AUPR`, `Precision@80%Recall`, and `Precision@90%Recall` metrics supported for CUAD

### DIFF
--- a/metrics/cuad/cuad.py
+++ b/metrics/cuad/cuad.py
@@ -1,0 +1,115 @@
+# coding=utf-8
+# Copyright 2020 The HuggingFace Datasets Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


metrics/cuad/cuad.py
```diff
@@ -0,0 +1,115 @@
...
+_DESCRIPTION = """
+This metric wrap the official scoring script for version 1 of the Contract
+Understanding Atticus Dataset (CUAD).
+Contract Understanding Atticus Dataset (CUAD) v1 is a corpus of more than 13,000 labels in 510
+commercial legal contracts that have been manually labeled to identify 41 categories of important
+clauses that lawyers look for when reviewing contracts in connection with corporate transactions.
+"""
+
+_KWARGS_DESCRIPTION = """
+Computes CUAD scores (EM, F1, AUPR, Precision@80%Recall, and Precision@90%Recall).
```
lhoestq, line 40: Maybe mention in the docstring that users can use multiple answers depending on a threshold on the confidence probability of each prediction ?

lhoestq (commented):
Thanks !!
I added one last comment

metrics/cuad/cuad.py
bhavitvyamalik: Added changes in docstring for `prediction_text`. Let me know if something is also required

lhoestq (approved):
perfect thanks :)

PR #2287, opened 2021-04-29
Avoid copying table's record batches

### LINKED ISSUE #2276: concatenate_datasets loads all the data into memory
## Describe the bug
When I try to concatenate 2 datasets (10GB each) , the entire data is loaded into memory instead of being written directly to disk.

Interestingly, this happens when trying to save the new dataset to disk or concatenating it again.

image

## Steps to reproduce the bug
```python
from datasets import concatenate_datasets, load_from_disk

test_sampled_pro = load_from_disk("test_sampled_pro")
val_sampled_pro = load_from_disk("val_sampled_pro")

big_set = concatenate_datasets([test_sampled_pro, val_sampled_pro])

# Loaded to memory
big_set.save_to_disk("big_set")

# Loaded to memory
big_set = concatenate_datasets([big_set, val_sampled_pro])
```

## Expected results
The data should be loaded into memory in batches and then saved directly to disk.

## Actual results
The entire data set is loaded into the memory and then saved to the hard disk.

## Versions
Paste the output of the following code:
```python
- Datasets: 1.6.1
- Python: 3.8.8 (default, Apr 13 2021, 19:58:26) 
[GCC 7.3.0]
- Platform: Linux-5.4.72-microsoft-standard-WSL2-x86_64-with-glibc2.10
```

mariosasko:
Fixes #2276

### DIFF
--- a/src/datasets/table.py
+++ b/src/datasets/table.py
@@ -158,6 +158,7 @@ class Table(IndexedTableMixin):
         # moreover calling deepcopy on a pyarrow table seems to make pa.total_allocated_bytes() decrease for some reason
         # by adding it to the memo, self.table won't be copied
         memo[id(self.table)] = self.table
+        memo[id(self._batches)] = self._batches
         return _deepcopy(self, memo)
 
     def __getstate__(self):


lhoestq:
Thanks for fixing it. I actually included a similar fix in #2291 along with some updates in tests
I'm closing this one in favor of #2291 if you don't mind.

Thanks again !

PR #2295, opened 2021-04-30
Create ExtractManager

albertvillanova:
Perform refactoring to decouple extract functionality.


freehob (approved):

albertvillanova:
Hi @lhoestq,

Once that #2578 has been merged, I would like to ask you to have a look at this PR: it implement the same logic as the one in #2578 but for all the other file compression formats.

Thanks.

src/datasets/utils/extract.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def extract(self, input_path, force_extract=False):
+        is_extractable, extractor = self.extractor.is_extractable(input_path, return_extractor=True)
+        if not is_extractable:
+            return input_path
+        output_path = self._get_output_path(input_path)
+        if self._do_extract(output_path, force_extract):
+            try:
+                self.extractor.extract(input_path, output_path, extractor=extractor)
+            except Exception:
```
lhoestq, line 42: Maybe only catch only the exception that happens when the archive formatted is not identified ? Which errors would you like to catch here ?

Otherwise this could catch unexpected errors.

You can also add the message of the original error to the `EnvironmentError` as additional information

src/datasets/utils/extract.py
```diff
@@ -0,0 +1,186 @@
...
+        lock_path = input_path + ".lock"
+        with FileLock(lock_path):
+            shutil.rmtree(output_path, ignore_errors=True)
+            os.makedirs(os.path.dirname(output_path), exist_ok=True)
+            if extractor:
+                return extractor.extract(input_path, output_path)
+            for extractor in cls.extractors:
+                if extractor.is_extractable(input_path):
+                    extractor.extract(input_path, output_path)
+                    break
```
lhoestq, line 186: nit for consistency with the return 2 lines above this one
```suggestion
                    return extractor.extract(input_path, output_path)
```

tests/test_file_utils.py
```diff
@@ -88,28 +87,22 @@ def gen_random_output():
...
-
-@pytest.mark.parametrize("compression_format", ["xz", "zstd"])
-def test_cached_path_extract(compression_format, xz_file, zstd_path, tmp_path, text_file):
-    path = {"xz": xz_file, "zstd": zstd_path}
-    input_path = path[compression_format]
+@pytest.mark.parametrize(
+    "compression_format, expected_text_path_name", [("gzip", "text_path"), ("xz", "text_file"), ("zstd", "text_file")]
+)
+def test_cached_path_extract(
+    compression_format, expected_text_path_name, text_gz_path, xz_file, zstd_path, tmp_path, text_file, text_path
```
lhoestq, line 94: You can also add a gz_file fixture if you want. This way you don't have to deal with different text path names

lhoestq (approved):
This is so clean ! Thanks :)

src/datasets/utils/extract.py
albertvillanova: You are right: in the original code there was no exception catch. 

Indeed, in the original code the raise of `EnvironmentError` was dead code: 
- previously if none of the if conditions was met there was a return (`return output_path`), 
  - See https://github.com/huggingface/datasets/blob/d7a72234436673d6394bc67cccb5489aacf5f981/src/datasets/utils/file_utils.py#L305-L313
- and this part of the code was in the else clause after all of the if conditions repeated again: the else clause was not possible to be executed because of the previous condition I mentioned above
  - See: https://github.com/huggingface/datasets/blob/d7a72234436673d6394bc67cccb5489aacf5f981/src/datasets/utils/file_utils.py#L338-L369

I would suggest just remove this exception raise and the corresponding try and the exception catch.

tests/test_file_utils.py
albertvillanova: Done

albertvillanova:
I think all is done @lhoestq ;)

lhoestq (approved):
Nice thanks !

PR #2310, opened 2021-05-04
Update README.md

cryoff:
Provides description of data instances and dataset features


albertvillanova:
Hi @cryoff, thanks for completing the dataset card.

Now there is an automatic validation tool to assure that all dataset cards contain all the relevant information. This is the cause of the non-passing test on your Pull Request:
```
Found fields that are not non-empty list of strings: {'annotations_creators': [], 'language_creators': []}
```

PR #2315, opened 2021-05-04
Datasets cli improvements

mariosasko:
This PR:
* replaces the code from the `bug_report.md` that was used to get relevant system info with a dedicated command (a more elegant approach than copy-pasting the code IMO)
* removes the `download` command (copied from the transformers repo?)
* adds missing help messages to the cli commands


mariosasko:
Additionally, I've deleted the points that are not very relevant for this repo (I guess the deleted points originate from the transformers repo). With this change, running `datasets-cli` is identical to copy-pasting the code from `bug_report.md`, but is more elegant because it doesn't require launching the REPL and copy-pasting the code.

src/datasets/commands/env.py
```diff
@@ -1,54 +1,29 @@
...
-                # deprecated in v2.1
-                tf_cuda_available = tf.test.is_gpu_available()
-            except AttributeError:
-                # returns list of devices, convert to bool
-                tf_cuda_available = bool(tf.config.list_physical_devices("GPU"))
-
         info = {
             "`datasets` version": version,
-            "Platform": platform.platform(),
+            "Platform": sys.version,
```
lhoestq, line 22: We actually need the `platform.platform()` here. This is useful to know if users are on windows or not for example.

src/datasets/commands/env.py
```diff
@@ -1,54 +1,29 @@
...
             "Python version": platform.python_version(),
-            "PyTorch version (GPU?)": "{} ({})".format(pt_version, pt_cuda_available),
-            "Tensorflow version (GPU?)": "{} ({})".format(tf_version, tf_cuda_available),
-            "Using GPU in script?": "<fill in>",
-            "Using distributed or parallel set-up in script?": "<fill in>",
```
lhoestq, line 48: Maybe let's add the PyArrow version here, this is often useful

lhoestq (commented):
Nice thanks !

You did right removing the download command, it indeed came form `transformers`.

lhoestq (approved):
Thanks :)

PR #2326, opened 2021-05-05
Enable auto-download for PAN-X / Wikiann domain in XTREME

lewtun:
This PR replaces the manual download of the `PAN-X.lang` domains with an auto-download from a Dropbox link provided by the Wikiann author. We also add the relevant dummy data for these domains.

While re-generating `dataset_infos.json` I ran into a `KeyError` in the `udpos.Arabic` domain so have included a fix for this as well.


albertvillanova (approved):
Thanks @lewtun. It is great to get rid of the manual download...

lhoestq (approved):
Thank you !!

PR #2328, opened 2021-05-06
Add Matthews/Pearson/Spearman correlation metrics

lhoestq:
Added three metrics:
- The Matthews correlation coefficient (from sklearn)
- The Pearson correlation coefficient (from scipy)
- The Spearman correlation coefficient (from scipy)

cc @sgugger


sgugger (approved):
Great, thanks a lot for adding those!

PR #2346, opened 2021-05-11
Add Qasper Dataset

cceyda:
[Question Answering on Scientific Research Papers](https://allenai.org/project/qasper/home)

Doing NLP on NLP papers to do NLP ♻️ I had to add it~

- [x] Add README (just gotta fill out some more )
- [x] Dataloader code
- [x] Make dummy dataset
- [x] generate dataset infos
- [x] Tests


lhoestq (commented):
Nice thank you !

Feel free to ping me when you're done with the dataset card :)

cceyda:
I saw that the README [template](https://github.com/huggingface/datasets/blob/master/templates/README.md) changed while I was working on this 😅 Some TOC titles may be different but I filled it to the best of my knowledge & readme quality check passes now.
ready for review @lhoestq

datasets/qasper/README.md
```diff
@@ -0,0 +1,204 @@
...
+{
+  'id': "Paper ID (string)",
+  'title': "Paper Title",
+  'abstract': "paper abstract ...",
+  'full_text': {
+      'paragraphs':[["section1_paragraph1_text","section1_paragraph2_text",...],["section2_paragraph1_text","section2_paragraph2_text",...]],
+      'section_name':["section1_title","section2_title"],...},
+  'qas': [explained below],
+}
+```
```
lhoestq, line 94: Could we have an actual example here instead ?

Since it can be long, feel free to crop some parts

datasets/qasper/README.md
```diff
@@ -0,0 +1,204 @@
...
+multilinguality:
+- monolingual
+size_categories: 
+- 10K<n<100K
+source_datasets:
+- extended|s2orc
+task_categories:
+- question-answering
+task_ids:
+- open-domain-qa
```
lhoestq, line 21: Is this really open domain ? There's a context associated to each question

lhoestq (commented):
Thanks ! Good job with the dataset card, the TOC/sections are all good

I have two final comments and then we can merge :)

datasets/qasper/README.md
cceyda: aha, true! changed to `closed-domain-qa`

datasets/qasper/README.md
cceyda: an actual example was too verbose and hard to understand no matter how I cropped/simplified it. imo an abstract version is easier to understand at first look.
I added some more explanations to the `qas` key and how it lists questions and its answers. Since The pyarrow format can feel confusing at first.
Let me know if this is okay

lhoestq (approved):
Thanks ! It's more readable this way indeed :)

PR #2349, opened 2021-05-11
Update task_ids for Ascent KB

phongnt570:
This "other-other-knowledge-base" task is better suited for the dataset.


lhoestq (approved):
Makes sense, thanks !

PR #2352, opened 2021-05-12
Set to_json default to JSON lines

albertvillanova:
With this PR, the method `Dataset.to_json`:
- is added to the docs
- defaults to JSON lines


stas00:
This is perfect, @albertvillanova - thank you! Tested it to work.

Might it be a good idea to document the args to `to_json`?

lhoestq (commented):
Cool ! I agree it's more convenient this way

Same comment as Stas: let's add `orient` and `lines` parameters to the docstring.
And maybe update the test to make sure these parameters work as expected.

Regarding the progress bar, we can definitely use tqdm in the `for` loop in the `_write` method of the json writer.

src/datasets/arrow_dataset.py
```diff
@@ -2765,17 +2765,28 @@ def to_json(
         batch_size: Optional[int] = None,
         **to_json_kwargs,
     ) -> int:
-        """Exports the dataset to JSON lines.
+        """Export the dataset to JSON or JSON Lines.
```
stas00, line 2768:
```suggestion
        """Export the dataset to JSON lines or JSON.
```

If the default is jsonl, may be listing it first?

stas00 (approved):
Fantastic - thank you, @albertvillanova

albertvillanova:
@lhoestq I added tests for both `lines` and `orient`.

lhoestq (approved):
Thanks !

PR #2355, opened 2021-05-12
normalized TOCs and titles in data cards

yjernite:
I started fixing some of the READMEs that were failing the tests introduced by @gchhablani but then realized that there were some consistent differences between earlier and newer versions of some of the titles (e.g. Data Splits vs Data Splits Sample Size, Supported Tasks vs Supported Tasks and Leaderboards). We also had different versions of the Table of Content

This PR normalizes all of them to the newer version


lhoestq (approved):
Awesome !

Also I noticed that you added `Contributions` to many tables of content, but most of time the dataset cards still don't have this section at the end of the readme

yjernite:
Oh right! I'd be in favor of still having the same TOC across the board, we can either leave it as is or add a `[More Info Needed]` `Contributions Section` wherever it's currently missing, wdyt?

yjernite:
(I thought those were programmatically updated based on git history :D )

yjernite:
Merging for now to avoid conflict since there are so many changes but let's figure out the contributions section next ;)

PR #2361, opened 2021-05-14
Preserve dtype for numpy/torch/tf/jax arrays

### LINKED ISSUE #625: dtype of tensors should be preserved
After switching to `datasets` my model just broke. After a weekend of debugging, the issue was that my model could not handle the double that the Dataset provided, as it expected a float (but didn't give a warning, which seems a [PyTorch issue](https://discuss.pytorch.org/t/is-it-required-that-input-and-hidden-for-gru-have-the-same-dtype-float32/96221)). 

As a user I did not expect this bug. I have a `map` function that I call on the Dataset that looks like this:

```python
def preprocess(sentences: List[str]):
    token_ids = [[vocab.to_index(t) for t in s.split()] for s in sentences]

    sembeddings = stransformer.encode(sentences)
    print(sembeddings.dtype)
    return {"input_ids": token_ids, "sembedding": sembeddings}
```

Given a list of `sentences` (`List[str]`), it converts those into token_ids on the one hand (list of lists of ints; `List[List[int]]`) and into sentence embeddings on the other (Tensor of dtype `torch.float32`). That means that I actually set the column "sembedding" to a tensor that I as a user expect to be a float32.

It appears though that behind the scenes, this tensor is converted into a **list**. I did not find this documented anywhere but I might have missed it. From a user's perspective this is incredibly important though, because it means you cannot do any data_type or tensor casting yourself in a mapping function! Furthermore, this can lead to issues, as was my case. 

My model expected float32 precision, which I thought `sembedding` was because that is what `stransformer.encode` outputs. But behind the scenes this tensor is first cast to a list, and when we then set its format, as below, this column is cast not to float32 but to double precision float64.

```python
dataset.set_format(type="torch", columns=["input_ids", "sembedding"])
```

This happens because apparently there is an intermediate step of casting to a **numpy** array (?) **whose dtype creation/deduction is different from torch dtypes** (see the snippet below).  As you can see, this means that the dtype is not preserved: if I got it right, the dataset goes from torch.float32 -> list -> float64 (numpy) -> torch.float64. 

```python
import torch
import numpy as np

l = [-0.03010837361216545, -0.035979013890028, -0.016949838027358055]
torch_tensor = torch.tensor(l)
np_array = np.array(l)
np_to_torch = torch.from_numpy(np_array)

print(torch_tensor.dtype)
# torch.float32
print(np_array.dtype)
# float64
print(np_to_torch.dtype)
# torch.float64
```

This might lead to unwanted behaviour. I understand that the whole library is probably built around casting from numpy to other frameworks, so this might be difficult to solve. Perhaps `set_format` should include a `dtypes` option where for each input column the user can specify the wanted precision.

The alternative is that the user needs to cast manually after loading data from the dataset but that does not seem user-friendly, makes the dataset less portable, and might use more space in memory as well as on disk than is actually needed.

bhavitvyamalik:
Fixes #625. This lets the user preserve the dtype of numpy array to pyarrow array which was getting lost due to conversion of numpy array -> list -> pyarrow array.


bhavitvyamalik:
Hi @lhoestq, 
It turns out that pyarrow `ListArray` are not recognized as list-like when we get output from `numpy_to_pyarrow_listarray`. This might cause tests to fail. If possible can we convert that `ListArray` output to list inorder for tests to pass? Under the hood it'll maintain the dtype as that of numpy array passed during input only

src/datasets/features.py
```diff
@@ -165,7 +167,7 @@ def _cast_to_python_objects(obj: Any) -> Tuple[Any, bool]:
         import torch
 
     if isinstance(obj, np.ndarray):
-        return obj.tolist(), True
+        return numpy_to_pyarrow_listarray(obj).tolist(), True
```
lhoestq, line 170: Thanks to your changes in the arrow writer, it is now able to take numpy arrays as inputs.
Therefore we don't need to cast the numpy arrays to list here, and we can leave the numpy array as is:

```suggestion
        return obj, False
```

Also feel free to rename this function `cast_to_python_or_numpy`.

Moreover for the pytorch and tensorflow tensors we can do the same: we can just get the numpy array instead of the python lists. It means that we can remove the `.tolist()` at line 172 and 174

src/datasets/arrow_writer.py
```diff
@@ -87,14 +88,18 @@ def __arrow_array__(self, type=None):
...
+        if type is not None:
+            pass
+        elif type is None and self.try_type:
             type = self.try_type
             trying_type = True
         else:
             type = self.type
         try:
             if isinstance(type, _ArrayXDExtensionType):
                 out = pa.ExtensionArray.from_storage(type, pa.array(self.data, type.storage_dtype))
```
lhoestq, line 100: Here we can also support numpy arrays for extension types as well (Array2D, Array3D etc.):

```suggestion
                storage = pa.array(self.data, type.storage_dtype) if not isinstance(self.data, np.ndarray) else numpy_to_pyarrow_listarray(self.data, type=type.value_type)
                out = pa.ExtensionArray.from_storage(type, storage)
```

We just need to edit `numpy_to_pyarrow_listarray` to have a `type` parameter that is passed at
```python
values = pa.array(arr.flatten(), type=type)
```

tests/test_features.py
```diff
@@ -129,7 +129,7 @@ def test_cast_to_python_objects_tuple(self):
         casted_obj = cast_to_python_objects(obj)
         self.assertDictEqual(casted_obj, expected_obj)
 
-    def test_cast_to_python_objects_numpy(self):
+    def test_cast_to_python_or_numpy(self):
```
lhoestq, line 132: We'll need to update this test to make sure that numpy arrays are unchanged instead of being casted to python lists.

lhoestq (commented):
Thanks ! To fix this I added a comment to actually keep the numpy array unchanged until it is passed to a TypedSequence. This way we don't have to deal with the ListArray issue

src/datasets/features.py
bhavitvyamalik: We can't compare 2 dictionaries if they contain numpy array using `assertDictEqual` (essentially, `dict1==dict2` doesn't work). In place of `self.assertDictEqual(casted_obj, expected_obj)` can we use `np.testing.assert_equal(casted_obj, expected_obj)`? We'll have to do this for all cases where we are casting to numpy array. Since `assertDictEqual` is part of unittest case module I'm a little unsure how to proceed ahead

lhoestq: Sure you can use `np.testing.assert_equal`. Also please note that `np.testing.assert_equal` won't check for the dtypes of the numpy arrays, which can be an issue.

bhavitvyamalik: Then we can use `np.testing.assert_equal` to check if they're same and individually check dtype of `casted_obj["col_1"][0]["vec"]` and `casted_obj["col_2"]` which should match that of `expected_obj["col_1"][0]["vec"]` and `expected_obj["col_2"]`?

bhavitvyamalik:
Brought down the failing tests from 7 to 4. Let me know if that part looks good. Failing tests are looking quite similar. In `test_map_torch` https://github.com/huggingface/datasets/blob/3d46bc384f811435e59e3916faa3aa20a1cf87bc/tests/test_arrow_dataset.py#L1039 and `test_map_tf`https://github.com/huggingface/datasets/blob/3d46bc384f811435e59e3916faa3aa20a1cf87bc/tests/test_arrow_dataset.py#L1056 
they're expecting `float64`. Shouldn't that be `float32`?

lhoestq:
It's normal: pytorch and tensorflow use `float64` by default, unlike numpy which uses `float32`.

I think that we should always keep the precision of the original tensor (torch/tf/numpy).
It means that as it is in this PR it's fine (the precision is conserved when doing the torch/tf -> numpy conversion).

This is a breaking change but in my opinion the fact that we had Value("float64") for torch.float32 tensors was an issue already.

Let me know what you think. Cc @albertvillanova if you have an opinion on this

If we agree on doing this breaking change, we can just change the test.

bhavitvyamalik:
Hi @lhoestq, 
Merged master into this branch. Only changing the test is left for now (mentioned below) after which all tests should pass.

> Brought down the failing tests from 7 to 4. Let me know if that part looks good. Failing tests are looking quite similar. In `test_map_torch`
> 
> https://github.com/huggingface/datasets/blob/3d46bc384f811435e59e3916faa3aa20a1cf87bc/tests/test_arrow_dataset.py#L1039
> 
> and `test_map_tf`
> https://github.com/huggingface/datasets/blob/3d46bc384f811435e59e3916faa3aa20a1cf87bc/tests/test_arrow_dataset.py#L1056
> 
> 
> they're expecting `float64`. Shouldn't that be `float32` now?

src/datasets/features.py
```diff
@@ -169,13 +171,13 @@ def _cast_to_python_objects(obj: Any) -> Tuple[Any, bool]:
...
+        return obj.tolist(), False
     elif config.TORCH_AVAILABLE and "torch" in sys.modules and isinstance(obj, torch.Tensor):
-        return obj.detach().cpu().numpy().tolist(), True
+        return obj.detach().cpu().numpy(), True
     elif config.TF_AVAILABLE and "tensorflow" in sys.modules and isinstance(obj, tf.Tensor):
-        return obj.numpy().tolist(), True
+        return obj.numpy(), True
     elif config.JAX_AVAILABLE and "jax" in sys.modules and isinstance(obj, jnp.ndarray):
-        return obj.tolist(), True
+        return jnp.asarray(obj), True
```
lhoestq, line 180: `jnp.asarray` is to convert an array to a jnp.ndarray, you can use `np.array` instead:
```suggestion
        return np.array(obj), True
```

lhoestq:
> they're expecting float64. Shouldn't that be float32 now?

Yes feel free to update those tests :)

It would be nice to have the same test for JAX as well

src/datasets/features.py
bhavitvyamalik: Oh, I saw only `ndarray` return type and not `np`. My bad :/

bhavitvyamalik:
Added same test for for JAX too. Also, I saw that I missed changing `test_cast_to_python_objects_jax` like I did for TF and PyTorch. Finished that as well

lhoestq (approved):
It looks all good !
Thanks a lot :)

PR #2362, opened 2021-05-14
Fix web_nlg metadata

julien-c:
Our metadata storage system does not support `.` inside keys. cc @Pierrci

### DIFF
--- a/datasets/web_nlg/README.md
+++ b/datasets/web_nlg/README.md
@@ -8,15 +8,15 @@ languages:
   - en
   release_v2:
   - en
-  release_v2.1:
+  release_v2_1:
   - en
-  release_v2.1_constrained:
+  release_v2_1_constrained:
   - en
   release_v2_constrained:
   - en
-  release_v3.0_en:
+  release_v3_0_en:
   - en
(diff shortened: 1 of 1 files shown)


lhoestq:
Hi ! `release_v2.1` and the others are dataset configuration names.

The configuration names are used to show the right code snippet in the UI to load the dataset.
For example if the parsing of the web_nlg tags worked correctly we would have:
image

Therefore I don't think it's a good idea to rename the configurations from `release_v2.1` to `release_v2_1` as the code snippet would be wrong in this case.

Moreover we can't really disallow dots in configuration names since it would be a big breaking change. It's commonly used, especially with multilingual datasets. For example `load_dataset("indic_glue", "sna.bn")`.

Is this something that can be fixed on the moonlanding side instead ?

julien-c:
> Is this something that can be fixed on the moonlanding side instead ?

Not really unless we change database:)

We'll maybe try to find another workaround, but super low-prio given that it's the only dataset that has those dotted keys in the YAML metadata

lhoestq:
Ok, should we close this PR then ?

PR #2364, opened 2021-05-15
README updated for SNLI, MNLI

### LINKED ISSUE #2275: SNLI dataset has labels of -1
There are a number of rows with a label of -1 in the SNLI dataset. The dataset descriptions [here](https://nlp.stanford.edu/projects/snli/) and [here](https://github.com/huggingface/datasets/tree/master/datasets/snli) don't list  -1 as a label possibility, and neither does the dataset viewer. As examples, see index 107 or 124 of the test set.

It isn't clear what these labels mean. I found a [line of code](https://github.com/huggingface/datasets/blob/80e59ef178d3bb2090d091bc32315c655eb0633d/datasets/snli/snli.py#L94) that seems to put them in but it seems still unclear why they are there. The current workaround is to just drop the rows from any model being trained. 

Perhaps the documentation should be updated.

bhavitvyamalik:
Closes #2275. Mentioned about -1 labels in MNLI, SNLI and how they should be removed before training. @lhoestq `check_code_quality` test might fail for MNLI as the license name `other-Open Portion of the American National Corpus`  is not a registered tag for 'licenses'

### DIFF
--- a/datasets/multi_nli/README.md
+++ b/datasets/multi_nli/README.md
@@ -111,7 +111,7 @@ The data fields are the same among all splits.
 - `{premise,hypothesis} parse`: Each sentence as parsed by the Stanford PCFG Parser 3.5.2
 - `{premise,hypothesis} binary parse`: parses in unlabeled binary-branching format
 - `genre`: a `string` feature.
-- `label`: a classification label, with possible values including `entailment` (0), `neutral` (1), `contradiction` (2)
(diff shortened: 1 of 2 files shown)


lhoestq (approved):
thanks !

lhoestq:
Regarding the license issue, I think we should allow it since it starts with `other-`. Cc @gchhablani what do you think ?

gchhablani:
@lhoestq I agree, I'll look into it.

PR #2367, opened 2021-05-17
Remove getchildren from hyperpartisan news detection

ghomasHudson:
`Element.getchildren()` is now deprecated in the ElementTree library (I think in python 3.9, so it still passes the automated tests which are using 3.6. But for those of us on bleeding-edge distros it now fails).

https://bugs.python.org/issue29209


lhoestq (approved):
Thanks for the fix !

PR #2369, opened 2021-05-17
correct labels of conll2003

philschmid:
# What does this PR

It fixes/extends the `ner_tags` for conll2003 to include all. 
Paper reference https://arxiv.org/pdf/cs/0306050v1.pdf
Model reference https://huggingface.co/elastic/distilbert-base-cased-finetuned-conll03-english/blob/main/config.json


lhoestq (approved):
Thanks !

PR #2370, opened 2021-05-17
Adding HendrycksTest dataset

andyzoujm:
Adding Hendrycks test from https://arxiv.org/abs/2009.03300.
I'm having a bit of trouble with dummy data creation because some lines in the csv files aren't being loaded properly (only the first entry loaded in a row of length 6). The dataset is loading just fine. Hope you can kindly help!
Thank you!


datasets/hendrycks_test/hendrycks_test.py
```diff
@@ -0,0 +1,188 @@
...
+        ]
+
+    def _generate_examples(self, datadir, split):
+        """Yields examples as (key, example) tuples."""
+        # This method handles input defined in _split_generators to yield (key, example) tuples from the dataset.
+        # The `key` is here for legacy reason (tfds) and is not important in itself.
+
+        id_ = 0
+        if split == "auxiliary_train":
+            for f in os.listdir(datadir):
```
lhoestq, line 177: we need to make the order deterministic:
```suggestion
            for f in sorted(os.listdir(datadir)):
```

andyzoujm:
@lhoestq Thank you for the review. I've made the suggested changes. There still might be some problems with dummy data though due to some csv loading issues (which I haven't found the cause to).

lhoestq:
I took a look at the dummy data and some csv lines were cropped. I fixed them :)

lhoestq (approved):
Looks all good now ! Thank you so much for adding it :)

albertvillanova:
@andyzoujm Any reason why this dataset scrip was called "hendrycks_test" instead of "mmlu"?

We are thinking of renaming it...

andyzoujm:
That's because we didn't call it MMLU in the paper (the shorthand didn't
emerge until over a year later), and people at OpenAI were calling it that.

Andy

albertvillanova:
Thanks for your reply. Just for the records: we have renamed it to "cais/mmlu": https://huggingface.co/datasets/cais/mmlu

PR #2375, opened 2021-05-18
Dataset Streaming

lhoestq:
# Dataset Streaming

## API

Current API is

```python

# Load an IterableDataset without downloading data
snli = load_dataset("snli", streaming=True)

# Access examples by streaming data
print(next(iter(snli["train"]))) 
# {'premise': 'A person on a horse jumps over a broken down airplane.',
#  'hypothesis': 'A person is training his horse for a competition.',
#  'label': 1}
```

I already implemented a few methods:
- IterableDataset.map: apply transforms on-the-fly to the examples
- IterableDataset.shuffle: shuffle the data _a la_ TFDS, i.e. with a shuffling buffer
- IterableDataset.with_format: set the format to `"torch"` to get a `torch.utils.data.IterableDataset`
- merge_datasets: merge two iterable datasets by alternating one or the other (you can specify the probabilities)

I would love to have your opinion on the API design :)

### Streaming

Data streaming is done using `fsspec` which has nice caching features.

To make dataset streaming work I extend the `open` function of dataset scripts to support opening remote files without downloading them entirely. It also works with remote compressed archives (currently only zip is supported):

```python
# Get a file-like object by streaming data from a remote file
open("https://github.com/davidsbatista/NER-datasets/raw/master/CONLL2003/train.txt")

# Get a file-like object by streaming data from a remote compressed archive by using the hop separator "::"
open("zip://snli_1.0_train.txt::https://nlp.stanford.edu/projects/snli/snli_1.0.zip")
```

I also extend the `os.path.join` function to support navigation in remote compressed archives, since it has to deal with the `"::"` separator. This separator is used by `fsspec`.

Finally I also added a retry mechanism in case the connection fails during data streaming.

### Transforms

An IterableDataset wraps an ExamplesIterable instance. There are different subclasses depending on the transforms we want to apply:
- ExamplesIterable: the basic one
- MappedExamplesIterable: an iterable with a `map` function applied on the fly
- BufferShuffledExamplesIterable: an iterable with a shuffling buffer
- CyclingMultiSourcesExamplesIterable: alternates between several ExamplesIterable
- RandomlyCyclingMultiSourcesExamplesIterable: randomly alternates between several ExamplesIterable

### DatasetBuilder

I use the same builders as usual. I just added a new method `_get_examples_iterable_for_split` to get an ExamplesIterable for a given split. Currently only the GeneratorBasedBuilder and the ArrowBasedBuilder implement it.

The BeamBasedBuilder doesn't implement it yet.
It means that datasets like wikipedia and natural_questions can't be loaded as IterableDataset for now.

## Other details

<S>I may have to do some changes in many dataset script to use `download` instead of `download_and_extract` when extraction is not needed. This will avoid errors for streaming.</s>

EDIT: Actually I just check for the extension of the file to do extraction only if needed.

EDIT2: It's not possible to stream from .tar.gz files without downloading the file completely. For now I raise an error if one want to get a streaming dataset based on .tar.gz files.

## TODO

usual stuff:

- [x] make streaming dependency "aiohttp" optional: `pip install datasets[streaming]`
- [x] tests
- [x] docs


src/datasets/utils/resources/languages.json
```diff
@@ -6,8 +6,10 @@
     "agq-CM": "Aghem (Cameroon)",
     "ak": "Akan",
     "ak-GH": "Akan (Ghana)",
+    "als": "Tosk Albanian",
```
albertvillanova, line 9: Why are you adding languages in this PR? 😕

lhoestq:
I can make a separate PR if you want.
Basically these are the missing languages from OSCAR, which is a dataset that required minor changes to be compatible with streaming. Because of the change it triggered the metadata validation in the CI so I fixed the metadata, i.e. the languages.
The nez codes are ISO 639-3 codes and don't have ISO 639-1 nor ISO 639-2 codes.

src/datasets/utils/resources/languages.json
lhoestq: I can make a separate PR if you want.
Basically these are the missing languages from OSCAR, which is a dataset that required minor changes to be compatible with streaming. Because of the change it triggered the metadata validation in the CI so I fixed the metadata, i.e. the languages.
The nez codes are ISO 639-3 codes and don't have ISO 639-1 nor ISO 639-2 codes.

docs/source/dataset_streaming.rst
```diff
@@ -0,0 +1,138 @@
+Load a Dataset in Streaming mode
+==============================================================
+
+When a dataset is in streaming mode, you can iterate over it directly, without having to download the entire dataset.
```
lewtun, line 4: nit:

```suggestion
When a dataset is in streaming mode, you can iterate over it directly without having to download the entire dataset.
```

docs/source/dataset_streaming.rst
```diff
@@ -0,0 +1,138 @@
...
+Here is a demonstration:
+
+.. code-block::
+
+    >>> from datasets import load_dataset
+    >>> dataset = load_dataset('oscar', "unshuffled_deduplicated_en", split='train', streaming=True)
+    >>> print(next(iter(dataset)))
+    {'text': 'Mtendere Village was inspired by the vision of Chief Napoleon Dzombe, which he shared with John Blanchard during his first visit to Malawi. Chief Napoleon conveyed the desperate need for a program to intervene and care for the orphans and vulnerable children (OVC) in Malawi, and John committed to help...
+
+Even though the dataset is 1.2 terabytes of data, you can start using it right away. Under the hood, it downloaded only the first examples of the dataset
```
lewtun, line 19: nit:

```suggestion
Even though the dataset is 1.2 terabytes of data, you can start using it right away! Under the hood, it only downloaded the first example of the dataset.
```

also, does it download 1 or more than one example when we use `next(iter(dataset))`?

docs/source/dataset_streaming.rst
```diff
@@ -0,0 +1,138 @@
...
+Shuffling the dataset: ``shuffle``
+--------------------------------------------------
+
+Shuffle the dataset
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To shuffle your dataset, the :func:`datasets.IterableDataset.shuffle` method fills a buffer of size ``buffer_size`` and randomly samples examples from this buffer.
+The selected examples in the buffer are replaced by new examples.
+
+For instance, if your dataset contains 1,000,000 examples but buffer_size is set to 1,000, then shuffle will initially select a random examples from only the first 1,000 examples in the buffer.
```
lewtun, line 36: nit:

```suggestion
For instance, if your dataset contains 1,000,000 examples but ``buffer_size`` is set to 1,000, then shuffle will initially select a random examples from only the first 1,000 examples in the buffer.
```

docs/source/dataset_streaming.rst
```diff
@@ -0,0 +1,138 @@
...
+Once an example is selected, its space in the buffer is replaced by the next (i.e. 1,001-st) example, maintaining the 1,000 example buffer.
+
+.. note::
+    For perfect shuffling, you need to set the ``buffer_size`` to be greater than the size of your dataset. But in this case it will download the full dataset in the buffer.
```
lewtun, line 40: nit:

```suggestion
    For perfect shuffling, you need to set ``buffer_size`` to be greater than the size of your dataset. But in this case it will download the full dataset in the buffer.
```

docs/source/dataset_streaming.rst
```diff
@@ -0,0 +1,138 @@
...
+    {'input_ids': [101, 11047, 10497, 7869, 2352...], 'token_type_ids': [0, 0, 0, 0, 0...], 'attention_mask': [1, 1, 1, 1, 1...]}
+
+
+Mix several iterable datasets together with ``interleave_datasets``
+----------------------------------------------------------------------------------------------------
+
+It is common to use several datasets to use a model. For example BERT was trained on a mix of Wikipedia and BookCorpus.
+You can mix several iterable datasets together using :func:`datasets.interleave_datasets`.
+
+By default, the resulting dataset alternates between the original datasets, but can also define sampling probabilities to sample randomly from the different datasets.
```
lewtun, line 102: wow, this is very cool!

docs/source/dataset_streaming.rst
```diff
@@ -0,0 +1,138 @@
...
+.. code-block::
+
+    >>> import torch
+    >>> tokenized_dataset = dataset.map(lambda x: tokenizer(x["text"], return_tensors="pt"))
+    >>> torch_tokenized_dataset = tokenized_dataset.with_format("torch")
+    >>> assert isinstance(torch_tokenized_dataset, torch.utils.data.IterableDataset)
+    >>> print(next(iter(torch_tokenized_dataset)))
+    {'input_ids': tensor([[101, 11047, 10497, 7869, 2352...]]), 'token_type_ids': tensor([[0, 0, 0, 0, 0...]]), 'attention_mask': tensor([[1, 1, 1, 1, 1...]])}
+
+For now, only the pytorch format is supported but support for TensorFlow and others will be added soon.
```
lewtun, line 138: nit:

```suggestion
For now, only the PyTorch format is supported but support for TensorFlow and others will be added soon.
```

src/datasets/filesystems/compression/gzip.py
```diff
@@ -0,0 +1,72 @@
...
+
+class GZipFileSystem(AbstractArchiveFileSystem):
+    """Read contents of GZIP archive as a file-system with one file inside."""
+
+    root_marker = ""
+    protocol = "gzip"
+
+    def __init__(
+        self,
+        fo="",
```
lewtun, line 16: question: in other places in `datasets` we use typing for each function argument. is there a reason we don't do that here?

docs/source/dataset_streaming.rst
```diff
@@ -0,0 +1,138 @@
...
+    >>> from datasets import load_dataset
+    >>> dataset = load_dataset('oscar', "unshuffled_deduplicated_en", split='train', streaming=True)
+    >>> print(next(iter(dataset)))
+    {'text': 'Mtendere Village was inspired by the vision of Chief Napoleon Dzombe, which he shared with John Blanchard during his first visit to Malawi. Chief Napoleon conveyed the desperate need for a program to intervene and care for the orphans and vulnerable children (OVC) in Malawi, and John committed to help...
+
+Even though the dataset is 1.2 terabytes of data, you can start using it right away. Under the hood, it downloaded only the first examples of the dataset
+
+.. note::
+
+    The dataset that is returned is a :class:`datasets.IterableDataset`, not the classic map-style :class:`datasets.Dataset`. To get examples from an iterable dataset, you have to iterate over it using a for loop for example. To get the very last example of the dataset, you first have to iterate on all the previous examples.
```
lewtun, line 23: is it possible to create a classic `Dataset` from an `IterableDataset`? 

one application that i have in mind is picking the first N examples of a huge dataset, collecting them in a standard `Dataset` and then doing all my exploration / preprocessing / task preparation etc on that dataset.

e.g. something like

```python

dataset = load_dataset('oscar', "unshuffled_deduplicated_en", split='train', streaming=True)
# create a `Dataset` i can play with?
sample = dataset.select(range(100))
```

lewtun (commented):
this feature look mega cool and will solve a major pain point i've experienced with task templates for large datasets!

i left a few nits in the docs and a couple of questions

src/datasets/iterable_dataset.py
```diff
@@ -0,0 +1,439 @@
...
+    return cls(
+        ex_iterable=ex_iterable,
+        info=info,
+        split=split,
+        format_type=format_type,
+        shuffling=shuffling,
+    )
+
+
+def interleave_datasets(
```
albertvillanova, line 410: I was thinking that the idea is to eventually have this function supporting regular Datasets. So maybe it should be moved to its own module?

docs/source/dataset_streaming.rst
lhoestq: It downloads the first examples (buffering + caching) and yield the first one

lhoestq: What about this ?
```suggestion
Even though the dataset is 1.2 terabytes of data, you can start using it right away. Under the hood, it downloaded only the first examples of the dataset for buffering, and returned the first example.
```

src/datasets/filesystems/compression/gzip.py
lhoestq: The code style comes from fsspec. I added the types

src/datasets/iterable_dataset.py
lhoestq: I moved it to a `combine` module. Maybe `concatenate_datasets` can be moved there as well

lhoestq (commented):
Thanks for the feedback :) I took your comments into account

docs/source/dataset_streaming.rst
lhoestq: Sure definitely :)

I was thinking of adding something like this in a next PR.
Maybe `IterableDataset.to_map_style_dataset()` ?
To get only the first examples we can also add `IterableDataset.take(n)`

docs/source/dataset_streaming.rst
lewtun: perfect!

docs/source/dataset_streaming.rst
lewtun: yeah, both those features would be great!

thomwolf: I like the `IterableDataset.take(n)` as well. Could we also have a `IterableDataset.sample(n)` taking a random sample?

lewtun: a random sample would be very neat as well. here we might want to use something like reservoir sampling to deal with unbounded streams: https://en.wikipedia.org/wiki/Reservoir_sampling

src/datasets/filesystems/compression/gzip.py
```diff
@@ -0,0 +1,73 @@
...
+        mode: str = "rb",
+        target_protocol: Optional[str] = None,
+        target_options: Optional[dict] = None,
+        block_size: int = DEFAULT_BLOCK_SIZE,
+        **kwargs,
+    ):
+        """
+        Parameters
+        ----------
+        fo: str
```
lewtun, line 27: just a question (not important to address): why is the docstring for this class different to the format we use elsewhere in `datasets` where the type is written as ``param (`type`)``?

lewtun (approved):
LGTM 🚀

src/datasets/filesystems/compression/gzip.py
albertvillanova: Yes, I wonder why using the NumPy style just in this docstring, instead of the usual Google style we use in all the other docstrings. 😅

src/datasets/combine.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def interleave_datasets(
+    datasets: List["IterableDataset"], probabilities: Optional[List[float]] = None, seed: Optional[int] = None
+) -> "IterableDataset":
+    """
+    Itervleave several iterable datasets (sources) into one iterable dataset.
+    The new iterable dataset alternates between the sources to yield the examples.
+    By default it cycles through the sources in order, but you can also make the new
+    iterable dataset sample examples from one random source at a time.
```
thomwolf, line 17:
```suggestion
    Iterleave several iterable datasets (sources) into a single iterable dataset.
    The new iterable dataset alternates between the sources to yield examples.
    If `probabilities = None` (default) the iterable dataset will cycles through the sources
    in order for each next example in the iteration.
    If `probabilities` is not `None, the iterable dataset will sample a random source according
    to the provided probabilities for each next examples in the iteration.
```

thomwolf (approved):
This is really super cool!

src/datasets/filesystems/compression/gzip.py
lhoestq: It comes from the fsspec docstrings, let me update that ;)

docs/source/dataset_streaming.rst
lhoestq: As soon as we have `.take()`, you can do `iterable_dataset.shuffle(buffer_size=buffer_size, seed=seed).take(n)` to take random samples.
This could be simplified by adding a `.sample()` method indeed

vblagoje: How about slicing support i.e `iterable_dataset[100:200]` to get an iterator or `Dataset` at a particular slice?

lhoestq: I would like to avoid allowing users to get items using `__getitem__` since it's not a map-style dataset.
So I agree it would be nice to get a slice of the data, but with a different API. Maybe something like
```python
sliced_dataset = iterable_dataset.skip(100).take(100)
```
What do you think ?

This is pretty close to the tf.data.Dataset API, which is also an iterable dataset.

vblagoje: Yes, very cool @lhoestq

PR #2376, opened 2021-05-18
Improve task api code quality

mariosasko:
Improves the code quality of the `TaskTemplate` dataclasses.

Changes:
* replaces `return NotImplemented` with raise `NotImplementedError` 
* replaces `sorted` with `len` in the uniqueness check 
* defines `label2id` and `id2label` in the `TextClassification` template as properties
* replaces the `object.__setattr__(self, attr, value)` syntax with (IMO nicer) `self.__dict__[attr] = value`

### DIFF
--- a/src/datasets/tasks/base.py
+++ b/src/datasets/tasks/base.py
@@ -1,8 +1,12 @@
 import abc
+import dataclasses
 from dataclasses import dataclass
-from typing import ClassVar, Dict
+from typing import ClassVar, Dict, Type, TypeVar
 
 from ..features import Features
+
+
+T = TypeVar("T", bound="TaskTemplate")
 
 
 @dataclass(frozen=True)
@@ -18,9 +22,9 @@ class TaskTemplate(abc.ABC):
     @property
     @abc.abstractmethod
     def column_mapping(self) -> Dict[str, str]:
-        return NotImplemented
+        raise NotImplementedError
 
     @classmethod
-    @abc.abstractmethod
-    def from_dict(cls, template_dict: dict) -> "TaskTemplate":
-        return NotImplemented
+    def from_dict(cls: Type[T], template_dict: dict) -> T:
+        field_names = set(f.name for f in dataclasses.fields(cls))
+        return cls(**{k: v for k, v in template_dict.items() if k in field_names})
--- a/src/datasets/tasks/text_classification.py
+++ b/src/datasets/tasks/text_classification.py
@@ -9,6 +9,8 @@ from .base import TaskTemplate
 class TextClassification(TaskTemplate):
     task: ClassVar[str] = "text-classification"
     input_schema: ClassVar[Features] = Features({"text": Value("string")})
+    # TODO(lewtun): Since we update this in __post_init__ do we need to set a default? We'll need it for __init__ so
+    # investigate if there's a more elegant approach.
     label_schema: ClassVar[Features] = Features({"labels": ClassLabel})
     labels: List[str]
     text_column: str = "text"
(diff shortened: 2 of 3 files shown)


src/datasets/tasks/text_classification.py
```diff
@@ -1,30 +1,26 @@
...
 
     def __post_init__(self):
-        assert sorted(set(self.labels)) == sorted(self.labels), "Labels must be unique"
+        assert len(self.labels) == len(set(self.labels)), "Labels must be unique"
         # Cast labels to tuple to allow hashing
-        object.__setattr__(self, "labels", tuple(sorted(self.labels)))
-        object.__setattr__(self, "text_column", self.text_column)
-        object.__setattr__(self, "label_column", self.label_column)
+        self.__dict__["labels"] = tuple(sorted(self.labels))
         self.label_schema["labels"] = ClassLabel(names=self.labels)
```
mariosasko, line 23: This line is an issue. Modifying a class attribute based on the value of an instance attribute doesn't make sense if multiple instances of the same class are allowed (if that's the case, the class attribute will have a valid value only for the instance that was initialized last). One way to fix this is with the help of descriptors:
```python
class FeaturesWithLazyClassLabel:
    def __init__(self, features, label_column="labels"):
        assert label_column in features, f"Key '{label_column}' missing in features {features}"
        self._features = features
        self._label_column = label_column

    def __get__(self, obj, objtype=None):
        if obj is None:
            return self._features
     
        assert hasattr(obj, self._label_column), f"Object has no attribute '{self._label_column}'"
        # note: this part is not cached, but we can add this easily
        features = self._features.copy()
        features["labels"] = ClassLabel(names=getattr(obj, self._label_column))
        return features
```

and then in `TextClassification`:

```python
label_schema: ClassVar[Features] = FeaturesWithLazyClassLabel(Features({"labels": ClassLabel}))
```

lhoestq: +1 on this !

The fix using `FeaturesWithLazyClassLabel` is fine IMO

lhoestq:
Looks good thanks, what do you think @lewtun ?

src/datasets/tasks/text_classification.py
lewtun: nice catch @mariosasko! i'm also happy with the `FeatureWithLazyClassLabel` fix for now

src/datasets/tasks/text_classification.py
```diff
@@ -1,30 +1,26 @@
...
 
 from ..features import ClassLabel, Features, Value
 from .base import TaskTemplate
 
 
 @dataclass(frozen=True)
 class TextClassification(TaskTemplate):
-    task = "text-classification"
-    input_schema = Features({"text": Value("string")})
+    task: ClassVar[str] = "text-classification"
```
lewtun, line 10: thanks for spotting this!

src/datasets/tasks/base.py
```diff
@@ -18,9 +22,9 @@ def features(self) -> Features:
...
     @abc.abstractmethod
     def column_mapping(self) -> Dict[str, str]:
-        return NotImplemented
+        raise NotImplementedError
 
     @classmethod
-    @abc.abstractmethod
-    def from_dict(cls, template_dict: dict) -> "TaskTemplate":
-        return NotImplemented
+    def from_dict(cls: Type[T], template_dict: dict) -> T:
```
lewtun, line 28: very elegant approach - thanks!

lewtun (approved):
wow, the code is not even one day old and already getting improved 😍  - thanks a lot @mariosasko, it LGTM!

src/datasets/tasks/text_classification.py
mariosasko: Yes, this is a temporary solution. I agree we should try to find a simpler way to fix this. The current API design does not seem to fit the task very well (having `label_schema` defined as a class attribute but depends on the value of instances?), so we should rethink that part IMO.

PR #2383, opened 2021-05-19
Improve example in rounding docs

mariosasko:
Improves the example in the rounding subsection of the Split API docs. With this change, it should more clear what's the difference between the `closest` and the `pct1_dropremainder` rounding.


lhoestq (approved):
Great ! It shows that you don't always end up with the same number of examples per percent by default :)
thanks !

PR #2389, opened 2021-05-21
Insert task templates for text classification

lewtun:
This PR inserts text-classification templates for datasets with the following properties:

* Only one config
* At most two features of `(Value, ClassLabel)` type

Note that this misses datasets like `sentiment140` which only has `Value` type features - these will be handled in a separate PR


lewtun:
Update: found a few datasets that slipped through the net. Adding them shortly!

yjernite:
You might have thought about this already, but would it make sense to use the `datasets.features.ClassLabel` values when possible instead of declaring the list once for the `feature` and once for the `template`?

lewtun:
> You might have thought about this already, but would it make sense to use the `datasets.features.ClassLabel` values when possible instead of declaring the list once for the `feature` and once for the `template`?

hi @yjernite, these code insertions are auto-generated so could certainly be improved :) 

just so i understand, your idea is that instead of doing something like

```python
class AGNews(datasets.GeneratorBasedBuilder):
    """AG News topic classification dataset."""

    def _info(self):
        return datasets.DatasetInfo(
            description=_DESCRIPTION,
            features=datasets.Features(
                {
                    "label": datasets.features.ClassLabel(
                        names=["World", "Sports", "Business", "Sci/Tech"]
                    ),
                }
            ),
            homepage="http://groups.di.unipi.it/~gulli/AG_corpus_of_news_articles.html",
            citation=_CITATION,
            task_templates=[
                TextClassification(
                    labels=("Business", "Sci/Tech", "Sports", "World"),
                    text_column="text",
                    label_column="label",
                )
            ],
        )
```

we could do the following:

```python
class AGNews(datasets.GeneratorBasedBuilder):
    """AG News topic classification dataset."""

    def _info(self):
        info = datasets.DatasetInfo(
            description=_DESCRIPTION,
            features=datasets.Features(
                {
                    "label": datasets.features.ClassLabel(
                        names=["World", "Sports", "Business", "Sci/Tech"]
                    ),
                }
            ),
            homepage="http://groups.di.unipi.it/~gulli/AG_corpus_of_news_articles.html",
            citation=_CITATION,
        )

        info.task_templates = [
            TextClassification(
                labels=info.features.names,
                text_column="text",
                label_column="label",
            )
        ]
        return info
```

lhoestq:
Or we could simply not specify the labels and update the template in the DatasetInfo postinit to give it the labels ?

yjernite:
> Or we could simply not specify the labels and update the template in the DatasetInfo postinit to give it the labels ?

Oh yes, that would be great! It does mean enforcing that people use the right feature type (sometimes people still use a `string` feature still because they don't want to enumerate the classes, but I guess you've been catching most of those in reviews @lhoestq )

There might be reasons where there should be a legitimate difference, but I can't really think of nay right now, and we can always duplicate the feature

datasets/caner/caner.py
```diff
@@ -90,7 +91,7 @@ def _info(self):
             homepage=_HOMEPAGE,
             license=_LICENSE,
             citation=_CITATION,
-        )
+        task_templates=[TextClassification(text_column="token", label_column="ner_tag")])
```
lewtun, line 94: Hmm, this dataset is described as a NER dataset, but the features suggest otherwise (i would expect a `Sequence` feature for `ner_tag` ...)

lewtun: Will fix this separately

lhoestq (approved):
Thank you !!

lhoestq:
Let's ignore the CI fails since they are unrelated to your changes. They're about dataset cards issues

PR #2390, opened 2021-05-21
Add check for task templates on dataset load

lewtun:
This PR adds a check that the features of a dataset match the schema of each compatible task template.


lhoestq (approved):
Nice thanks !

tests/test_dataset_common.py
```diff
@@ -170,6 +170,14 @@ def check_if_url_is_valid(url):
...
                     # check that loaded datset is not empty
                     self.parent.assertTrue(len(dataset[split]) > 0)
+
+                # check that we can cast features for each task template
+                task_templates = dataset_builder.info.task_templates
+                if task_templates:
+                    for task in task_templates:
+                        task_features = {**task.input_schema, **task.label_schema}
+                        dataset = dataset.prepare_for_task(task)
+                        self.parent.assertDictEqual(task_features, dataset.features)
```
lhoestq, line 180: Wait a sec before merging, I noticed something, sorry I approved too fast x)

Here `dataset` is supposed to be a dataset dict so `dataset.features` shouldn't work. Can you make sure everything is correct ?

lhoestq: Oh maybe it didn't fail because no datasets have templates yet ?

lewtun: good catch! i misread the meaning of `dataset` 😄 . yes the reason the tests pass is that we don't have any templates yet - i'll update this to work with `DatasetDict` objects instead

lewtun: oh i just remembered that `prepare_for_task` works for both `Dataset` and `DatasetDict` objects:

* https://huggingface.co/docs/datasets/package_reference/main_classes.html?highlight=prepare_for_task#datasets.Dataset.prepare_for_task
* https://huggingface.co/docs/datasets/package_reference/main_classes.html?highlight=prepare_for_task#datasets.DatasetDict.prepare_for_task

so this test is actually ok IMO

lhoestq: Well I don't think so ;)
`dataset.features` will raise an AttributeError

You can do this instead:
```suggestion
                        for dataset_split in dataset.values():
                            self.parent.assertDictEqual(task_features, dataset_split.features)
```

lewtun: oh i missed that bit - nice catch!

lhoestq:
LGTM now, thank you =)

PR #2392, opened 2021-05-21
[WIP] Update text classification template labels in DatasetInfo __post_init__

lewtun:
Hi @yjernite and @lhoestq, here's a first stab at the suggestion discussed in #2389 to update the `labels` of the `TextClassification` template in the `DatasetInfo.__post_init__`.

One problem I've spotted is that my current implementation introduces state into the `__post_init__`: 

* When we call `load_dataset`, `DatasetInfo.features` are the "raw" features without any casting so we can access the column names by the `label_column` specified in `TextClassification`
* When we call `Dataset.prepare_for_task` we run into a problem because the `DatasetInfo.features` are first cast into the new schema which triggers a `KeyError` when we update the infos [here](https://github.com/huggingface/datasets/blob/8b2a78520828e0cc13c14a31f413a5395ef25110/src/datasets/arrow_dataset.py#L1959).

What do you think? I did this a bit quickly, so maybe I'm overlooking something obvious :)

### DIFF
--- a/src/datasets/info.py
+++ b/src/datasets/info.py
@@ -36,8 +36,10 @@ import os
 from dataclasses import asdict, dataclass, field
 from typing import List, Optional, Union
 
+from datasets.tasks.text_classification import TextClassification
+
 from . import config
-from .features import Features, Value
+from .features import Features, Value, ClassLabel
 from .splits import SplitDict
 from .tasks import TaskTemplate, task_template_from_dict
 from .utils import Version
@@ -165,6 +167,15 @@ class DatasetInfo:
             else:
                 template = task_template_from_dict(self.task_templates)
                 self.task_templates = [template] if template is not None else []
+            # insert labels and mappings for text classification
+            for idx, template in enumerate(self.task_templates):
+                if isinstance(template, TextClassification) and self.features is not None:
+                    # This introduces state and raises a KeyError when we call Dataset.prepare_for_task :(
+                    # The reason is that Dataset.prepare_for_task calls Dataset.cast which converts the
+                    # DatasetInfo.features to the new schema and thus template.label_column is no longer a valid key
+                    object.__setattr__(template, "labels", tuple(self.features[template.label_column].names))
+                    template.label_schema["labels"] = ClassLabel(names=template.labels)
+                    self.task_templates[idx] = template
 
     def _license_path(self, dataset_info_dir):
         return os.path.join(dataset_info_dir, config.LICENSE_FILENAME)
--- a/src/datasets/tasks/text_classification.py
+++ b/src/datasets/tasks/text_classification.py
@@ -12,19 +12,14 @@ class TextClassification(TaskTemplate):
     # TODO(lewtun): Since we update this in __post_init__ do we need to set a default? We'll need it for __init__ so
     # investigate if there's a more elegant approach.
     label_schema = Features({"labels": ClassLabel})
-    labels: List[str]
+    labels: List[str] = None
     text_column: str = "text"
     label_column: str = "labels"
 
     def __post_init__(self):
-        assert sorted(set(self.labels)) == sorted(self.labels), "Labels must be unique"
-        # Cast labels to tuple to allow hashing
-        object.__setattr__(self, "labels", tuple(sorted(self.labels)))
+        object.__setattr__(self, "labels", self.labels)
         object.__setattr__(self, "text_column", self.text_column)
         object.__setattr__(self, "label_column", self.label_column)
-        self.label_schema["labels"] = ClassLabel(names=self.labels)
-        object.__setattr__(self, "label2id", {label: idx for idx, label in enumerate(self.labels)})
-        object.__setattr__(self, "id2label", {idx: label for label, idx in self.label2id.items()})
 
     @property
     def column_mapping(self) -> Dict[str, str]:
@@ -40,3 +35,11 @@ class TextClassification(TaskTemplate):
             label_column=template_dict["label_column"],
             labels=template_dict["labels"],
         )
+
+    @property
+    def label2id(self):
+        return {label: idx for idx, label in enumerate(self.labels)}
+
+    @property
+    def id2label(self):
+        return {idx: label for idx, label in enumerate(self.labels)}


mariosasko:
If I'm not mistaken, one way to fix this would be to drop the task templates when copying the info by inserting `dataset.info.task_templates = None` before the `Dataset.cast` call in `Dataset.prepare_for_task`. Moreover, we should do this change independently of the KeyError being raised because currently the following is possible:
```python
dset = load_dataset("some_dataset") # let's say 'some_dataset' supports text classification and question answering
dset_tc = dset.prepare_for_task("text-classification")
dset_tc.preprare_for_task("question-answering") # this should raise an error because the schema is no longer valid for this task; currently this fails on 'rename_columns'
```
I see 2 options:
1. to drop the task templates after the first `Dataset.prepare_for_task` call
2. to save only the tasks compatible with the new schema after Dataset.prepare_for_task` (but then we have to update the column names of the compatible tasks to make sure the column mapping is still valid)

lewtun:
> If I'm not mistaken, one way to fix this would be to drop the task templates when copying the info by inserting `dataset.info.task_templates = None` before the `Dataset.cast` call in `Dataset.prepare_for_task`. Moreover, we should do this change independently of the KeyError being raised because currently the following is possible:
> 
> ```python
> dset = load_dataset("some_dataset") # let's say 'some_dataset' supports text classification and question answering
> dset_tc = dset.prepare_for_task("text-classification")
> dset_tc.preprare_for_task("question-answering") # this should raise an error because the schema is no longer valid for this task; currently this fails on 'rename_columns'
> ```
> 
> I see 2 options:
> 
> 1. to drop the task templates after the first `Dataset.prepare_for_task` call
> 2. to save only the tasks compatible with the new schema after Dataset.prepare_for_task` (but then we have to update the column names of the compatible tasks to make sure the column mapping is still valid)

thanks for the great idea @mariosasko and for spotting the problem with sequential task preparation! i am in favour of your option (1) since it is simple and saves us from having to keep track of the column mappings across multiple steps. 

i've implemented the change and refactored the tests to account for the new approach (including a new test that the templates are flushed after we call `prepare_for_task`). perhaps the slightly inelegant aspect here is that if we want to allow the user to set `labels` in the `TextClassification` template, then we have two places (`DatasetInfo.__post_init__` and `TextClassification.__post_init__`) where we need to update `label_schema`. 

on the other hand, dropping `labels` from the `TextClassification` signature would have the nice effect that users only have to think about column names when defining their tasks.

in any case, i think it would be a good idea to merge #2376 soon as the current PR is touching a lot of the same places in the codebase 😄

datasets/emotion/emotion.py
```diff
@@ -44,6 +45,12 @@ def _info(self):
             supervised_keys=("text", "label"),
             homepage=_URL,
             citation=_CITATION,
+            task_templates=[
```
lewtun, line 48: this is here for demo purposes only and appears to be the reason why the tests are failing. i'll remove it if we decide to merge this PR

lewtun:
cc @SBrandeis who might also be interested in this feature :)

lhoestq:
Tests are failing only because the `emotion` dataset card doesn't pass our dataset card validator (tags are missing), you can ignore this since it's unrelated to this PR.

src/datasets/info.py
```diff
@@ -166,6 +168,13 @@ def __post_init__(self):
             else:
                 template = task_template_from_dict(self.task_templates)
                 self.task_templates = [template] if template is not None else []
+            # Insert labels and mappings for text classification
+            for idx, template in enumerate(self.task_templates):
+                if isinstance(template, TextClassification) and self.features is not None:
+                    # Cast labels to tuple to enable hashing of task template
+                    object.__setattr__(template, "labels", tuple(sorted(self.features[template.label_column].names)))
```
lhoestq: Maybe raise an error if `template.labels` is not None and and doesn't match the labels in info.features ?

If users want to create a custom TaskTemplate with custom labels, then they must match the dataset labels.
e.g. this should fail:

```python
ds = load_dataset("emotion")
labels = ['label_that_doesnt_exist', 'another_one_that_doesnt_exist']
tc = TextClassification(text_column="text", label_column="label", labels=labels)
ds.prepare_for_task(tc)
```

lhoestq (commented):
Looks great ! I added 2 comments:

src/datasets/info.py
lewtun: > Maybe raise an error if `template.labels` is not None and and doesn't match the labels in info.features ?

Great idea! I decided to raise the error in `prepare_for_task` because `template.labels` is `None` by default in `DatasetInfo.__post_init__` and is the reason we set the attribute there with:

```python
# Cast labels to tuple to enable hashing of task template                   
 template.__dict__["labels"] = tuple(sorted(self.features[template.label_column].names))
```

I've also added a test to verify that we raise the error as expected (I also discovered some of my previous assertion tests were passing for the wrong reason 😅 ).

Perhaps one drawback from this approach is that we force the user to explicitly define the labels if they want to create their own `TextClassification` instance. On the other hand, being explicit is the Python way 😄 

> Also maybe we should create another instance of `TextClassification`, so that the user could reuse a TaskTemplate with unset labels later, instead of updating the current object.

I think that by raising the error in `prepare_for_task`, we no longer need a new instance because an instance with unset labels will raise a `ValueError` and not be modified. Please correct me if I misunderstood your remark.

src/datasets/arrow_dataset.py
```diff
@@ -1416,10 +1418,18 @@ def prepare_for_task(self, task: Union[str, TaskTemplate]) -> "Dataset":
...
+            if template.labels is None or template.labels != dataset_labels:
+                raise ValueError(
+                    f"Incompatible labels between the dataset and task template! Expected labels {dataset_labels} but got {template.labels}. Please ensure that `datasets.tasks.TextClassification.labels` matches the features of the dataset."
+                )
         column_mapping = template.column_mapping
         columns_to_drop = [column for column in self.column_names if column not in column_mapping]
         dataset = self.remove_columns(columns_to_drop)
         dataset = dataset.rename_columns(column_mapping)
+        # We found a template so now flush `DatasetInfo` to skip the template update in `DatasetInfo.__post_init__`
+        dataset.info.task_templates = None
```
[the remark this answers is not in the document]
lewtun: It's actually worse than I anticipated: subsequent calls to `load_dataset` produce no templates

```
ds1 = load_dataset("./datasets/emotion/", split="train")
# task templates not empty (expected behaviour)
assert ds1.info.task_templates
ds2 = load_dataset("./datasets/emotion/", split="train")
# fails - task templates are empty
assert ds2.info.task_templates
```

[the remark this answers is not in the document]
lewtun: It turns out that losing the templates on subsequent calls to `load_dataset` is not connected to `prepare_for_task`, but is actually due to the logic in `DatasetInfo.__post_init__`: I can reproduce the above example on `master` as well

Need to investigate the `__post_init__` first

src/datasets/tasks/text_classification.py
```diff
@@ -23,30 +24,26 @@ def __get__(self, obj, objtype=None):
 
 @dataclass(frozen=True)
 class TextClassification(TaskTemplate):
-    task: ClassVar[str] = "text-classification"
+    # `task` is not a ClassVar since we want it to be part of the `asdict` output for JSON serialization
+    task: str = "text-classification"
```
lhoestq, line 28: This fixed your issue when reloading the dataset @lewtun 
Indeed `task_template_from_dict` wasn't able to reload the template from json data since the "task" field was missing

src/datasets/info.py
```diff
@@ -167,6 +170,16 @@ def __post_init__(self):
                 template = task_template_from_dict(self.task_templates)
                 self.task_templates = [template] if template is not None else []
 
+        # Insert labels and mappings for text classification
+        if self.task_templates is not None:
+            self.task_templates = list(self.task_templates)
```
lhoestq, line 175: I'm making a new list here to avoid side effects

src/datasets/tasks/text_classification.py
lewtun: amazing, thanks for spotting and fixing this!

src/datasets/info.py
lhoestq: Thanks :) Let's allow labels to be None and sync with the features when appropriate (currently in DatasetInfo.__post_init__)

Regarding the new instance: I created a new instance in the changes I pushed.

Looks all good now

lewtun:
@lhoestq @SBrandeis i've fixed the tests and think this is now in a good state for another review :)

tests/test_arrow_dataset.py
```diff
@@ -702,8 +702,12 @@ def test_concatenate_with_no_task_templates(self, in_memory):
...
 
     def test_concatenate_with_equal_task_templates(self, in_memory):
-        task_template = TextClassification(text_column="text", label_column="labels", labels=["pos", "neg"])
-        info = DatasetInfo(task_templates=task_template)
+        labels = ["neg", "pos"]
+        task_template = TextClassification(text_column="text", label_column="labels", labels=labels)
+        info = DatasetInfo(
+            features=Features({"text": Value("string"), "labels": ClassLabel(names=labels)}),
+            task_templates=task_template,
+        )
```
lhoestq, line 710: Maybe add a test to make sure that the DatasetInfo post init does pass the labels to the task templates, so that we can instantiate the task_template without labels as in the demo ?

lhoestq (approved):
Looks all good ! Thanks for the fix :)

lhoestq:
Maybe @SBrandeis you can also take a look to make sure you're fine with it ?

tests/test_arrow_dataset.py
lewtun: good idea! i've now set up the unit test to cover the following cases:

1. `DatasetInfo` defined with a `TextClassification` template without `labels`
2. `DatasetInfo` defined with a `TextClassification` template with `labels` (this one might be overkill, but best to be safe)
3. Passing a `TextClassification` template with labels directly to `prepare_for_task` (the no labels case is covered in `test_task_with_incompatible_templates`)

src/datasets/tasks/text_classification.py
```diff
@@ -23,30 +24,26 @@ def __get__(self, obj, objtype=None):
...
 
     def __post_init__(self):
-        assert len(self.labels) == len(set(self.labels)), "Labels must be unique"
-        # Cast labels to tuple to allow hashing
-        self.__dict__["labels"] = tuple(sorted(self.labels))
+        if self.labels:
+            assert len(self.labels) == len(set(self.labels)), "Labels must be unique"
+            # Cast labels to tuple to allow hashing
+            self.__dict__["labels"] = tuple(sorted(self.labels))
+            self.__dict__["label_schema"] = copy.deepcopy(self.label_schema)
```
mariosasko, line 41: Should be safe to remove the `FeaturesWithLazyClassLabel` descriptor after this to reduce the code complexity (and then the deepcopy call can be replaced with `self.label_schema.copy()`).

mariosasko (commented):
Few suggestions.

src/datasets/info.py
```diff
@@ -167,6 +170,16 @@ def __post_init__(self):
...
+            for idx, template in enumerate(self.task_templates):
+                if isinstance(template, TextClassification) and self.features is not None:
+                    labels = self.features[template.label_column].names
+                    self.task_templates[idx] = TextClassification(
+                        text_column=template.text_column, label_column=template.label_column, labels=labels
+                    )
```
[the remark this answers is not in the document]
lewtun: good idea!

src/datasets/tasks/text_classification.py
lewtun: yes i agree, thanks for the suggestion! since we now create a new instance of `TextClassification` in `DatasetInfo.__post_init__` I think we should be safe from the side-effects you spotted earlier

PR #2395, opened 2021-05-22
`pretty_name` for dataset in YAML tags

bhavitvyamalik:
I'm updating `pretty_name` for datasets in YAML tags as discussed with @lhoestq. Here are the first 10, please let me know if they're looking good.

If dataset has 1 config, I've added `pretty_name` as `config_name: full_name_of_dataset` as config names were `plain_text`, `default`, `squad` etc (not so important in this case) whereas when dataset has >1 configs, I've added `config_name: full_name_of_dataset+config_name` so as to let user know about the `config` here.


bhavitvyamalik:
Initially I removed the ` - `since there was only one `pretty_name` per config but turns out it was breaking [here](https://github.com/huggingface/datasets/blob/74751e3f98c74d22c48c6beb1fab0c13b5dfd075/src/datasets/utils/metadata.py#L197) in `/utils/metadata.py`

gchhablani:
@lhoestq I guess this will also need some validation?

lhoestq:
Looks like the parser doesn't allow things like
```
pretty_name:
  config_name1: My awesome config number 1
  config_name2: My amazing config number 2
```
therefore you had to use `-` and consider them as a list.

I would be nice to add support for this case in the validator.

There's one thing though: the DatasetMetadata object currently corresponds to the yaml tags that are flattened: the config names are just ignored, and the lists are concatenated.

Therefore I think we would potentially need to instantiate several `DatasetMetadata` objects: one per config. Otherwise we would end up with a list of several pretty_name while we actually need at most 1 per config.

What do you think @gchhablani ?

bhavitvyamalik:
I was thinking of returning `metada_dict` (on line 193) whenever `load_dataset_card` is called (we can pass an extra parameter to `from_readme` or `from_yaml_string` for that to happen).

One just needs config_name as key for the dictionary inside `pretty_name` dict and for single config, there would be only one value to print. We can do this for other fields as well like `size_categories`, `languages` etc. This will obviate the need to flatten the YAML tags so that don't have to instantiate several DatasetMetadata objects. What do you guys think @lhoestq @gchhablani?

gchhablani:
Hi @lhoestq @bhavitvyamalik 

Few things that come to my mind:

1. Instead of instantiating a new `DatasetMetadata` for each config with flattened tags, why can't we make it more flexible and validate only non-dict items? However, in that case, the types wouldn't be as strict for the class attributes. It would also not work for cases that are like `Dict[str,List[Dict[str,str]]`, but I guess that won't be needed anyway.

    Ideally, it would be something like - Check the metadata tag type, do a DFS, and find the non-dict objects, and validate them. Is this an overkill to handle the problem?
2. For single-config datasets, there can be slightly different validation for `pretty_names`, than for multi-config. The user shouldn't need to provide a config name for single-config datasets, wdyt @bhavitvyamalik @lhoestq?
3. @bhavitvyamalik, I'm not sure I understand your approach, can you please elaborate? The `metadata_dict` is flattened before instantiating the object, do you want to remove that? Still confused.

gchhablani:
Btw, `pretty_names` can also be used to handle this during validation :P 

```
-# Dataset Card for [Dataset Name]
+# Dataset Card for Allegro Reviews
```

gchhablani:
@bhavitvyamalik

Now, I think I better understand what you're saying. So you want to skip validation for the unflattened metadata and just return it? And let the validation run for the flattened version?

bhavitvyamalik:
Exactly! Validation is important but once the YAML tags are validated I feel we shouldn't do that again while calling `load_dataset_card`. +1 for default config name for single-config datasets.

gchhablani:
@bhavitvyamalik
Actually, I made the `ReadMe` validation similar to `DatasetMetadata` validation and the class was validating the metadata during the creation. 

Maybe we need to have a separate validation method instead of having it in `__post_init__`? Wdyt @lhoestq? 

I'm sensing too many directions to look into. It'd be great to discuss these sometime. 

But if this PR is urgent then @bhavitvyamalik's logic seems good to me. It doesn't need major modifications in validation.

lhoestq:
> Maybe we need to have a separate validation method instead of having it in __post_init__? Wdyt @lhoestq?

We can definitely have a `is_valid()` method instead of doing it in the post init.

> What about adding a pretty name across all configs, and then config-specific names?

Let's keep things simple to starts with. If we can allow both single-config and multi-config cases it would already be great :)

for single-config:
```yaml
pretty_name: Allegro Reviews
```

for multi-config:
```yaml
pretty_name:
  mrpc: Microsoft Research Paraphrase Corpus (MRPC)
  sst2: Stanford Sentiment Treebank
  ...
```

To support the multi-config case I see two options:
1. Don't allow DatasetMetadata to have dictionaries but instead have separate DatasetMetadata objects per config
2. allow DatasetMetadata to have dictionaries. It implies to remove the flattening step. Then we could get metadata for a specific config this way for example:
```python
from datasets import load_dataset_card

glue_dataset_card = load_dataset_card("glue")
print(glue_dataset_card.metadata)
# DatasetMetatada object with dictionaries since there are many configs
print(glue_dataset_card.metadata.get_metadata_for_config("mrpc"))
# DatasetMetatada object with no dictionaries since there are only the mrpc tags
```

Let me know what you think or if you have other ideas.

gchhablani:
I think Option 2 is better.

Just to clarify, will `get_metadata_for_config`  also return common details, like language, say?

lhoestq:
> Just to clarify, will get_metadata_for_config also return common details, like language, say?

Yes that would be more convenient IMO. For example a dataset card like this
```yaml
languages:
- en
pretty_name:
  config1: Pretty Name for Config 1
  config3: Pretty Name for Config 2
```

then `metadat.get_metadata_for_config("config1")`  would return something like
```python
DatasetMetadata(languages=["en"], pretty_name="Pretty Name for Config 1")
```

bhavitvyamalik:
@lhoestq, should we do this post-processing in `load_dataset_card` by returning unflattened dictionary from `DatasetMetadata` or send this from `DatasetMetadata`?

lhoestq:
Not sure I understand the difference @bhavitvyamalik

bhavitvyamalik:
I was talking about this unflattened dictionary:

> I was thinking of returning the whole dictionary before flattening so that user can access whatever they want with specific configs. Let's say [this](https://pastebin.com/eJ84314f) is my metadata_dict before flattening (the loaded YAML string), so instead of validating it and then returning the items individually we can return it just after loading the YAML string.

lhoestq:
I still don't understand what you mean by "returning unflattened dictionary from DatasetMetadata or send this from DatasetMetadata", sorry. Can you give an example or rephrase this ?

IMO load_dataset_card can return a dataset card object with a metadata field. If the metadata isn't flat (i.e. it has several configs), you can get the flat metadata of 1 specific config with `get_metadata_for_config`. But of course if you have better ideas or suggestions, we can discuss this

gchhablani:
@lhoestq,  I think he is saying whatever `get_metadata_for_config` is doing can be done in `load_dataset_card` by taking the unflattened `metadata_dict` as input.

@bhavitvyamalik, I think it'd be better to have this "post-processing" in `DatasetMetadata` instead of `load_dataset_card`, as @lhoestq has shown. I'll quickly get on that.

---
Three things that are to be changed in `DatasetMetadata`:
1. Allow Non-flat elements and their validation.
2.Create a method to get metadata by config name.
3. Create a `validate()` method.

Once that is done, this PR can be updated and merged, wdys?

lhoestq:
Thanks @gchhablani for the help ! Now that https://github.com/huggingface/datasets/pull/2436 is merged you can remove the `-` in the pretty_name :)

datasets/ade_corpus_v2/README.md
```diff
@@ -33,6 +33,10 @@ task_ids:
   Ade_corpus_v2_drug_dosage_relation:
   - coreference-resolution
 paperswithcode_id: null
+pretty_name:
+  Ade_corpus_v2_classification: Adverse Drug Reaction Data v2 (Ade_corpus_v2_classification)
+  Ade_corpus_v2_drug_ade_relation: Adverse Drug Reaction Data v2 (Ade_corpus_v2_drug_ade_relation)
+  Ade_corpus_v2_drug_dosage_relation: Adverse Drug Reaction Data v2 (Ade_corpus_v2_drug_dosage_relation)
```
lhoestq, line 39:
```suggestion
pretty_name:
  Ade_corpus_v2_classification: Adverse Drug Reaction Data v2 (Classification)
  Ade_corpus_v2_drug_ade_relation: Adverse Drug Reaction Data v2 (AE relations)
  Ade_corpus_v2_drug_dosage_relation: Adverse Drug Reaction Data v2 (Dosages relations)
```

datasets/adversarial_qa/README.md
```diff
@@ -19,6 +19,11 @@ task_ids:
 - extractive-qa
 - open-domain-qa
 paperswithcode_id: adversarialqa
+pretty_name:
+  adversarialQA: adversarialQA (adversarialQA)
+  dbert: adversarialQA (dbert)
+  dbidaf: adversarialQA (dbidaf)
+  droberta: adversarialQA (droberta)
```
lhoestq, line 26:
```suggestion
  adversarialQA: Adversarial QA (combined)
  dbert: Adversarial QA (dbert)
  dbidaf: Adversarial QA (dbidaf)
  droberta: Adversarial QA (droberta)
```

datasets/aeslc/README.md
```diff
@@ -2,6 +2,7 @@
 languages:
 - en
 paperswithcode_id: aeslc
+pretty_name: '"aeslc"'
```
lhoestq, line 5:
```suggestion
pretty_name: Annotated Enron Subject Line Corpus (AESLC)
```

datasets/afrikaans_ner_corpus/README.md
```diff
@@ -18,6 +18,7 @@ task_categories:
 task_ids:
 - named-entity-recognition
 paperswithcode_id: null
+pretty_name: Afrikaans Ner Corpus
```
lhoestq, line 21:
```suggestion
pretty_name: Afrikaans NER Corpus
```

datasets/ag_news/README.md
```diff
@@ -18,6 +18,7 @@ task_categories:
 task_ids:
 - topic-classification
 paperswithcode_id: ag-news
+pretty_name: '"ag_news"'
```
lhoestq, line 21:
```suggestion
pretty_name: AG news
```

datasets/ai2_arc/README.md
```diff
@@ -19,6 +19,9 @@ task_ids:
 - open-domain-qa
 - multiple-choice-qa
 paperswithcode_id: null
+pretty_name:
+  ARC-Challenge: '"ai2_arc" (ARC-Challenge)'
+  ARC-Easy: '"ai2_arc" (ARC-Easy)'
```
lhoestq, line 24:
```suggestion
  ARC-Challenge: AI2 Reasoning Challenge (ARC) 2018 (ARC-Challenge)
  ARC-Easy: AI2 Reasoning Challenge (ARC) 2018 (ARC-Easy)
```

datasets/air_dialogue/README.md
```diff
@@ -21,6 +21,9 @@ task_ids:
 - dialogue-modeling
 - language-modeling
 paperswithcode_id: null
+pretty_name:
+  air_dialogue_data: air_dialogue (air_dialogue_data)
+  air_dialogue_kb: air_dialogue (air_dialogue_kb)
```
lhoestq, line 26:
```suggestion
  air_dialogue_data: AirDialogue (air_dialogue_data)
  air_dialogue_kb: AirDialogue (air_dialogue_kb)
```

lhoestq (commented):
Thanks :)

I added a few suggestions for the pretty names to try to make them prettier :P

albertvillanova:
Thanks @bhavitvyamalik.

I think this PR was superseded by these others also made by you:
-  #3498
- #3536

I'm closing this.

PR #2397, opened 2021-05-24
Fix number of classes in indic_glue sna.bn dataset

albertvillanova:
As read in the [paper](https://www.aclweb.org/anthology/2020.findings-emnlp.445.pdf), Table 11.


albertvillanova:
@lhoestq there are many things missing in the README.md file, but this correction is right despite not passing the validation tests...

lhoestq:
Yes indeed. We run the validation in all modified readme because we think that it is the time when contributors are the most likely to fix a dataset card - or it will never be done

PR #2404, opened 2021-05-25
Paperswithcode dataset mapping

julien-c:
This is a continuation of https://github.com/huggingface/huggingface_hub/pull/43, encoded directly inside dataset cards.

As discussed:
- `paperswithcode_id: null` when the dataset doesn't exist on paperswithcode's side.
- I've added this new key at the end of the yaml instead of ordering all keys alphabetically as pyyaml's default. No strong opinion on that one though


lhoestq (approved):
Thanks !

I added `paperswithcode_id` to our tags validator.

The CI still fails because many model cards are still missing some sections or have invalid tags. The errors are unrelated to this PR so let's merge this one.

julien-c:
live mapping can be found at https://huggingface.co/api/pwc/datasets-mapping and will be kept up to date going forward

PR #2417, opened 2021-05-27
Make datasets PEP-561 compliant

SBrandeis:
Allows to type-check datasets with `mypy` when imported as a third-party library

PEP-561: https://www.python.org/dev/peps/pep-0561
MyPy doc on the subject: https://mypy.readthedocs.io/en/stable/installed_packages.html


julien-c:
This is super cool, I love that ❤️

PR #2416, opened 2021-05-27
Add KLUE dataset

jungwhank:
Add `KLUE (Korean Language Understanding Evaluation)` dataset released recently from [paper](https://arxiv.org/abs/2105.09680), [github](https://github.com/KLUE-benchmark/KLUE) and [webpage](https://klue-benchmark.com/tasks).
Please let me know if there's anything missing in the code or README.
Thanks!

PR #2418, opened 2021-05-27
add utf-8 while reading README

bhavitvyamalik:
It was causing tests to fail in Windows (see #2416). In Windows, the default encoding is CP1252 which is unable to decode the character byte 0x9d

### DIFF
--- a/src/datasets/utils/metadata.py
+++ b/src/datasets/utils/metadata.py
@@ -55,7 +55,7 @@ class NoDuplicateSafeLoader(yaml.SafeLoader):
 
 
 def yaml_block_from_readme(path: Path) -> Optional[str]:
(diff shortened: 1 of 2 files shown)


mariosasko:
Can you please add encoding to this line as well to fix the issue (and maybe replace `path.open(...)` with `open(path, ...)`)?
https://github.com/huggingface/datasets/blob/7bee4be44706a59b084b9b69c4cd00f73ee72f76/src/datasets/utils/metadata.py#L58

bhavitvyamalik:
Sure, in fact even I was thinking of adding this in order to maintain the consistency!

lhoestq (approved):
thanks for the fix !

PR #2423, opened 2021-05-28
add `desc` in `map` for `DatasetDict` object

bhavitvyamalik:
`desc` in `map` currently only works with `Dataset` objects. This PR adds support for `DatasetDict` objects as well

### DIFF
--- a/src/datasets/dataset_dict.py
+++ b/src/datasets/dataset_dict.py
@@ -410,6 +410,7 @@ class DatasetDict(dict):
         disable_nullable: bool = False,
         fn_kwargs: Optional[dict] = None,
         num_proc: Optional[int] = None,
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
Great thank you :)

lhoestq:
The CI error is unrelated to the PR, merging

bhavitvyamalik:
@lhoestq, can we release this feature if you guys are planning for any patch release for Datasets. It'll slow down [#11927](https://github.com/huggingface/transformers/pull/11927) otherwise :/

lhoestq:
Sure definitely, having a discrepancy between Dataset.map and DatasetDict.map is an issue that we should fix and include in a patch release. Will it in the coming days

PR #2427, opened 2021-05-31
Add copyright info to MLSUM dataset

PhilipMay:


PhilipMay:
Build fails but this change should not be the reason...

PhilipMay:
rebased on master

lhoestq (approved):
Thanks for adding the licensing infos !

PR #2429, opened 2021-05-31
Rename QuestionAnswering template to QuestionAnsweringExtractive

lewtun:
Following the discussion with @thomwolf  in #2255, this PR renames the QA template to distinguish extractive vs abstractive QA. The abstractive template will be added in a future PR.


lhoestq (commented):
I like having "extractive" in the name to make things explicit. However this creates an inconsistency with transformers.

See
https://huggingface.co/transformers/task_summary.html#extractive-question-answering

But this is minor IMO and I'm ok with this renaming

lhoestq (approved):

SBrandeis (approved):

lewtun:
> I like having "extractive" in the name to make things explicit. However this creates an inconsistency with transformers.
> 
> See
> https://huggingface.co/transformers/task_summary.html#extractive-question-answering
> 
> But this is minor IMO and I'm ok with this renaming

yes i chose this convention because it allows us to match the `QuestionAnsweringXxx` naming and i think it's better to have `task_name-subtask_name` should auto-complete ever become part of the Hub :)

PR #2411, opened 2021-05-27
Add DOI badge to README

albertvillanova:
Once published the latest release, the DOI badge has been automatically generated by Zenodo.

PR #2430, opened 2021-05-31
Add version-specific BibTeX

albertvillanova:
As pointed out by @lhoestq in #2411, after the creation of the Zenodo DOI for Datasets, a new BibTeX entry is created with each release.

This PR adds a version-specific BibTeX entry, besides the existing one which is generic for the project.

See version-specific BibTeX entry here: https://zenodo.org/record/4817769/export/hx#.YLSyd6j7RPY

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,4 +153,43 @@ If you want to cite this framework you can use this:
   volume={1},
   year={2020}
 }
+@software{quentin_lhoest_2021_4817769,
+  author       = {Quentin Lhoest and
+                  Patrick von Platen and
+                  Thomas Wolf and
+                  Albert Villanova del Moral and
+                  Yacine Jernite and
+                  Mario Šaško and
+                  Suraj Patil and
+                  mariamabarham and
+                  Julien Chaumond and
+                  Julien Plu and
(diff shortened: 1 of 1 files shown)


lhoestq:
Maybe we should only keep one citation ?
cc @thomwolf @yjernite

albertvillanova:
For info:
- The one automatically generated by Zenodo is version-specific, and a new one will be generated after each release.
- Zenodo has also generated a project-specific DOI, but currently this only redirects to the DOI page of the latest version.
- All the information automatically generated by Zenodo can be corrected/customized if necessary.

lhoestq:
Thanks for the details ! As zenodo says we should probably just show the versioned DOI. And we can remove the old citation.

albertvillanova:
I have removed the old citation.

What about the new one? Should we customize it? I have fixed some author names (replaced nickname with first and family names). Note that the list of authors is created automatically by Zenodo from this list: https://github.com/huggingface/datasets/graphs/contributors
I do not know if this default automatic list of authors is what we want to show in the citation...

lhoestq (approved):
Thank you :)

PR #2436, opened 2021-06-01
Update DatasetMetadata and ReadMe

gchhablani:
This PR contains the changes discussed in #2395.

**Edit**:
In addition to those changes, I'll be updating the `ReadMe` as follows:

Currently, `Section` has separate parsing and validation error lists. In `.validate()`, we add these lists to the final lists and throw errors.

One way to make `ReadMe` consistent with `DatasetMetadata` and add a separate `.validate()` method is to throw separate parsing and validation errors. 

This way, we don't have to throw validation errors, but only parsing errors in `__init__ ()`. We can have an option in `__init__()` to suppress parsing errors so that an object is created for validation. Doing this will allow the user to get all the errors in one go.

In `test_dataset_cards` , we are already catching error messages and appending to a list. This can be done for `ReadMe()`  for parsing errors, and `ReadMe(...,suppress_errors=True); readme.validate()`  for validation, separately.

**Edit 2**:
The only parsing issue we have as of now is multiple headings at the same level with the same name. I assume this will happen very rarely, but it is still better to throw an error than silently pick one of them. It should be okay to separate it this way. 

Wdyt @lhoestq ?


src/datasets/utils/readme.py
```diff
@@ -89,6 +86,12 @@ def parse(self):
                     if self.text != "" and self.text not in FILLER_TEXT:
                         self.is_empty_text = False
 
+        if self.level == "" and not suppress_parsing_errors:
+            if self.parsing_error_list != [] or self.parsing_warning_list != []:
+                errors = "\n".join(list(map(lambda x: "-\t" + x, self.parsing_error_list + self.parsing_warning_list)))
```
lhoestq, line 91: you can use a comprehension list
```suggestion
                errors = "\n".join("-\t" + x for x in self.parsing_error_list + self.parsing_warning_list)
```

src/datasets/utils/metadata.py
```diff
@@ -77,61 +77,126 @@ def metadata_dict_from_readme(path: Path) -> Optional[Dict[str, List[str]]]:
...
-    size_categories: List[str]
-    source_datasets: List[str]
-    task_categories: List[str]
-    task_ids: List[str]
+    annotations_creators: Union[List[str], Dict[str, List[str]]]
+    language_creators: Union[List[str], Dict[str, List[str]]]
+    languages: Union[List[str], Dict[str, List[str]]]
+    licenses: Union[List[str], Dict[str, List[str]]]
+    multilinguality: Union[List[str], Dict[str, List[str]]]
+    pretty_names: Union[str, Dict[str, str]]
```
lhoestq, line 192: I would rename it `pretty_name` without an `s` since there's only one (or one per config), contrary to the other tags like `task_ids` that may have several values

lhoestq (commented):
Thank you ! 
Could you rename pretty_names to pretty_name ?
Other than that it looks all good to me :)

src/datasets/utils/metadata.py
gchhablani: Fixed.

src/datasets/utils/readme.py
gchhablani: Fixed.

lhoestq (approved):
Thanks a lot ! Merging ^^

PR #2442, opened 2021-06-02
add english language tags for ~100 datasets

VictorSanh:
As discussed on Slack, I have manually checked for ~100 datasets that they have at least one subset in English. This information was missing so adding into the READMEs.

Note that I didn't check all the subsets so it's possible that some of the datasets have subsets in other languages than English...


lhoestq (approved):
Thank you !
There are a few multilingual datasets that are still missing other languages than english but we can add them in another PR

lhoestq:
Fixing the tags of all the datasets is out of scope for this PR so I'm merging even though the CI fails because of the missing tags

PR #2448, opened 2021-06-05
Fix flores download link

mariosasko:


lhoestq (approved):
Thanks for the update :)

Maybe we should use an URL that doesn't point to the master branch, but instead use a specific commit hash in order to avoid issues like this in the future. This is also valid for all the other datasets that have data files hosted in github repos.

Issue #2366, opened 2021-05-17
Json loader fails if user-specified features don't match the json data fields order


lhoestq:
If you do
```python
dataset = load_dataset("json", data_files=data_files, features=features)
```
Then depending on the order of the features in the json data field it fails:
```python
[...]
~/Desktop/hf/datasets/src/datasets/packaged_modules/json/json.py in _generate_tables(self, files)
     94             if self.config.schema:
     95                 # Cast allows str <-> int/float, while parse_option explicit_schema does NOT
---> 96                 pa_table = pa_table.cast(self.config.schema)
     97             yield i, pa_table
[...]
ValueError: Target schema's field names are not matching the table's field names: ['tokens', 'ner_tags'], ['ner_tags', 'tokens']
```

This is because one must first re-order the columns of the table to match the `self.config.schema` before calling cast.

One way to fix the `cast` would be to replace it with:
```python
# reorder the arrays if necessary + cast to schema
# we can't simply use .cast here because we may need to change the order of the columns
pa_table = pa.Table.from_arrays([pa_table[name] for name in schema.names], schema=schema)
```

PR #2453, opened 2021-06-07
Keep original features order

albertvillanova:
When loading a Dataset from a JSON file whose column names are not sorted alphabetically, we should get the same column name order, whether we pass features (in the same order as in the file) or not.

I found this issue while working on #2366.

### DIFF
--- a/src/datasets/features.py
+++ b/src/datasets/features.py
@@ -829,7 +829,7 @@ def get_nested_type(schema: FeatureType) -> pa.DataType:
         return pa.list_(value_type)
     elif isinstance(schema, Sequence):
         value_type = get_nested_type(schema.feature)
-        # We allow to reverse list of dict => dict of list for compatiblity with tfds
+        # We allow to reverse list of dict => dict of list for compatibility with tfds
         if isinstance(value_type, pa.StructType):
             return pa.struct({f.name: pa.list_(f.type, schema.length) for f in value_type})
         return pa.list_(value_type, schema.length)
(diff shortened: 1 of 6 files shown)


lhoestq:
The arrow writer was supposing that the columns were always in the sorted order. I just pushed a fix to reorder the arrays accordingly to the schema. It was failing for many datasets like squad

lhoestq:
and obviously it broke everything

lhoestq:
Feel free to revert my commit. I can investigate this in the coming days

albertvillanova:
@lhoestq I do not understand when you say:
> It was failing for many datasets like squad

All the test were green after my last commit.

lhoestq:
> All the tests were green after my last commit.

Yes but loading the actual squad dataset was failing :/

src/datasets/features.py
```diff
@@ -817,8 +817,8 @@ def get_nested_type(schema: FeatureType) -> pa.DataType:
     # Nested structures: we allow dict, list/tuples, sequences
     if isinstance(schema, Features):
         return pa.struct(
-            {key: get_nested_type(schema[key]) for key in sorted(schema)}
-        )  # sort to make the order of columns deterministic
+            {key: get_nested_type(schema[key]) for key in schema}
+        )  # Features is subclass of dict, and dict order is deterministic since Python 3.7
```
lhoestq, line 821: Actually it's deterministic since 3.6 but at that time it was not acknowledged as a feature of the language.
```suggestion
        )  # Features is subclass of dict, and dict order is deterministic since Python 3.6
```

lhoestq (approved):
Thanks for figuring out why some columns were missing !

PR #2454, opened 2021-06-07
Rename config and environment variable for in memory max size

albertvillanova:
As discussed in #2409, both config and environment variable have been renamed.

cc: @stas00, huggingface/transformers#12056


stas00:
Thank you for the rename, @albertvillanova!

lhoestq (approved):
thanks for renaming

PR #2455, opened 2021-06-08
Update version in xor_tydi_qa.py

cccntu:
Fix #2449

@lhoestq Should I revert to the old `dummy/1.0.0` or  delete it and keep only `dummy/1.1.0`?


lhoestq:
Hi ! Thanks for updating the version

> Should I revert to the old dummy/1.0.0 or delete it and keep only dummy/1.1.0?

Feel free to delete the old dummy data files

lhoestq (approved):
Thanks !

PR #2457, opened 2021-06-08
Add align_labels_with_mapping function

lewtun:
This PR adds a helper function to align the `label2id` mapping between a `datasets.Dataset` and a classifier (e.g. a transformer with a `PretrainedConfig.label2id` dict), with the alignment performed on the dataset itself.

This will help us with the Hub evaluation, where we won't know in advance whether a model that is fine-tuned on say MNLI has the same mappings as the MNLI dataset we load from `datasets`.

An example where this is needed is if we naively try to evaluate `microsoft/deberta-base-mnli` on `mnli` because the model config has the following mappings:

```python
  "id2label": {
    "0": "CONTRADICTION",
    "1": "NEUTRAL",
    "2": "ENTAILMENT"
  },
  "label2id": {
    "CONTRADICTION": 0,
    "ENTAILMENT": 2,
    "NEUTRAL": 1
  }
```

while the `mnli` dataset has the `contradiction` and `neutral` labels swapped:

```python
id2label = {0: 'entailment', 1: 'neutral', 2: 'contradiction'}
label2id = {'contradiction': 2, 'entailment': 0, 'neutral': 1}
```

As a result, we get a much lower accuracy during evaluation:

```python
from transformers.trainer_utils import EvalPrediction
from transformers import AutoModelForSequenceClassification, Trainer

# load dataset for evaluation
mnli = load_dataset("glue", "mnli", split="test")
# load model
model_ckpt = "microsoft/deberta-base-mnli"
model = AutoModelForSequenceClassification.from_pretrained(checkpoint)
# preprocess, create trainer ...
mnli_enc = ...
trainer = Trainer(model, args=args, tokenizer=tokenizer)
# generate preds
preds = trainer.predict(mnli_enc)
# preds.label_ids misalinged with model.config => returns wrong accuracy (too low)!
compute_metrics(EvalPrediction(preds.predictions, preds.label_ids))
```

The fix is to use the helper function before running the evaluation to make sure the label IDs are aligned:

```python
mnli_enc_aligned = mnli_enc.align_labels_with_mapping(label2id=config.label2id, label_column="label")
# preds now aligned and everyone is happy :)
preds = trainer.predict(mnli_enc_aligned)
```

cc @thomwolf @lhoestq

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -3097,20 +3097,20 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
             fingerprint=new_fingerprint,
         )
 
-    def align_labels_with_mapping(self, id2label: Dict, label2id: Dict, label_column: str = "labels") -> "Dataset":
+    def align_labels_with_mapping(self, label2id: Dict, label_column: str = "labels") -> "Dataset":
         features = self.features
         int2str_function = features[label_column].int2str
-        label_name_to_id = {k.lower(): v for k, v in label2id.items()}
-        label_names = [k.lower() for k in id2label.values()]
+        # Some label mappings use uppercase label names so we lowercase them
+        label2id = {k.lower(): v for k, v in label2id.items()}
+        label_names = list(label2id.keys())
         features[label_column] = ClassLabel(num_classes=len(label_names), names=label_names)
 
-        def _process_labels(batch):
-            label_names = [int2str_function(label_id) for label_id in batch[label_column]]
-            # TODO(lewtun): do we need to also lowercase dataset label names?
-            batch[label_column] = [label_name_to_id[label_name] for label_name in label_names]
+        def process_label_ids(batch):
+            dset_label_names = [int2str_function(label_id).lower() for label_id in batch[label_column]]
+            batch[label_column] = [label2id[label_name] for label_name in dset_label_names]
             return batch
 
-        return self.map(_process_labels, features=features, batched=True)
+        return self.map(process_label_ids, features=features, batched=True)
 
 
 def concatenate_datasets(
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -3098,7 +3098,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
         )
 
     def align_labels_with_mapping(self, label2id: Dict, label_column: str = "labels") -> "Dataset":
-        features = self.features
+        features = self.features.copy()
         int2str_function = features[label_column].int2str
         # Some label mappings use uppercase label names so we lowercase them
         label2id = {k.lower(): v for k, v in label2id.items()}
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -3097,13 +3097,33 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
             fingerprint=new_fingerprint,
         )
 
-    def align_labels_with_mapping(self, label2id: Dict, label_column: str = "labels") -> "Dataset":
+    def align_labels_with_mapping(self, label2id: Dict, label_column: str) -> "Dataset":
+        """Align the dataset's label ID and label name mappings to match an input :obj:`label2id` mapping.
+        This is useful when you want to ensure that a model's predicted labels are aligned with the evaluation dataset.
+
+        Args:
+            label2id (:obj:`dict`):
+                The label name to ID mapping to align the dataset with.
+            label_column (:obj:`str`):
+                The column name of labels to align on.
+
+        Example:
+            .. code-block:: python
+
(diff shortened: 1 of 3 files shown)


lewtun:
@lhoestq i think this is ready for another review 🙂

src/datasets/arrow_dataset.py
```diff
@@ -3097,6 +3097,21 @@ def add_item(self, item: dict, new_fingerprint: str):
             fingerprint=new_fingerprint,
         )
 
+    def align_labels_with_mapping(self, label2id: Dict, label_column: str = "labels") -> "Dataset":
```
lhoestq, line 3100: I wouldn't set "labels" as the default and let the user specify explicitly which column they want to change.

Also could you add a docstring please ?
You can also add this method in the documentation in `main_classes.rst`

src/datasets/arrow_dataset.py
```diff
@@ -3097,6 +3097,21 @@ def add_item(self, item: dict, new_fingerprint: str):
...
+        features = self.features.copy()
+        int2str_function = features[label_column].int2str
+        # Some label mappings use uppercase label names so we lowercase them
+        label2id = {k.lower(): v for k, v in label2id.items()}
+        label_names = list(label2id.keys())
+        features[label_column] = ClassLabel(num_classes=len(label_names), names=label_names)
```
lhoestq, line 3106: Here you use the lowercase label names. IMO we should keep the original ones (before lowering)

lhoestq (commented):
Thanks for adding this method and writing the test !
Just have 2 comments:

src/datasets/arrow_dataset.py
lewtun: done!

src/datasets/arrow_dataset.py
lewtun: good idea! done :)

src/datasets/arrow_dataset.py
```diff
@@ -3097,6 +3097,41 @@ def add_item(self, item: dict, new_fingerprint: str):
...
+                # dataset with mapping {'entailment': 0, 'neutral': 1, 'contradiction': 2}
+                ds = load_dataset("glue", "mnli", split="train")
+                # mapping to align with
+                label2id = {'CONTRADICTION': 0, 'NEUTRAL': 1, 'ENTAILMENT': 2}
+                ds_aligned = ds.align_labels_with_mapping(label2id, "label")
+        """
+        features = self.features.copy()
+        int2str_function = features[label_column].int2str
+        # Sort input mapping by ID value to ensure the label names are aligned
+        label2id = dict(sorted(label2id.items(), key=lambda item: item[1]))
```
lewtun, line 3122: while adding a test that the label names match their expected values, i realised that we need to sort the input label mapping by ID because the user may provide something like 

```
label2id = {"CONTRADICTION": 0, "ENTAILMENT": 2, "NEUTRAL": 1}
```

which results in an `int2str` function in the aligned dataset that doesn't match what you'd expect from

```
id2label = {v: k for k, v in label2id.items()}
```

lewtun:
@lhoestq thanks for the feedback - it's now integrated :) 

i also added a comment about sorting the input label IDs

src/datasets/arrow_dataset.py
lhoestq: Indeed ! good catch

src/datasets/arrow_dataset.py
```diff
@@ -3097,6 +3097,41 @@ def add_item(self, item: dict, new_fingerprint: str):
             fingerprint=new_fingerprint,
         )
 
+    def align_labels_with_mapping(self, label2id: Dict, label_column: str) -> "Dataset":
+        """Align the dataset's label ID and label name mappings to match an input :obj:`label2id` mapping.
+        This is useful when you want to ensure that a model's predicted labels are aligned with the evaluation dataset.
```
lhoestq, line 3102:
```suggestion
        """Align the dataset's label ID and label name mapping to match an input :obj:`label2id` mapping.
        This is useful when you want to ensure that a model's predicted labels are aligned with the dataset.
        The alignment in done using the lowercase label names.
```

lhoestq (approved):
Thanks ! Looks all good now :)

We will also need to have the `DatasetDict.align_labels_with_mapping` method. Let me quickly add it

lhoestq:
Created the PR here: https://github.com/huggingface/datasets/pull/2510

lewtun:
> Thanks ! Looks all good now :)
> 
> We will also need to have the `DatasetDict.align_labels_with_mapping` method. Let me quickly add it

thanks a lot! i always forget about `DatasetDict` - will be happy when it's just one "dataset" object :)

retoj:
So, there seems to be a problem with the function align_labels_with_mapping for models like this: https://huggingface.co/huggingface/distilbert-base-uncased-finetuned-mnli]. At least with this model, but perhaps also with others, the model.config.label2id values are of type str not int, which crashes said function. After manually converting the model.config.label2id values to int, the script runs smoothly.

PR #2460, opened 2021-06-08
Revert default in-memory for small datasets

### LINKED ISSUE #2458: Revert default in-memory for small datasets
Users are reporting issues and confusion about setting default in-memory to True for small datasets.

We see 2 clear use cases of Datasets:
- the "canonical" way, where you can work with very large datasets, as they are memory-mapped and cached (after every transformation)
- some edge cases (speed benchmarks, interactive/exploratory analysis,...), where default in-memory can explicitly be enabled, and no caching will be done

After discussing with @lhoestq we have agreed to:
- revert this feature (implemented in #2182)
- explain in the docs how to optimize speed/performance by setting default in-memory

cc: @stas00 https://github.com/huggingface/datasets/pull/2409#issuecomment-856210552

albertvillanova:
Close #2458


lhoestq (approved):
Thanks for setting the default IN_MEMORY_MAX_SIZE to zero and updating the docs :)

stas00:
Thank you for this welcome change guys!

PR #2464, opened 2021-06-08
fix: adjusting indexing for the labels.

drugilsberg:
The labels index were mismatching the actual ones used in the dataset. Specifically `0` is used  for `SUPPORTS` and `1` is used for `REFUTES`
After this change, the `README.md` now reflects the content of `dataset_infos.json`.

Signed-off-by: Matteo Manica <drugilsberg@gmail.com>


lhoestq (approved):
Good catch ! Thanks for fixing it

drugilsberg:
> Good catch ! Thanks for fixing it

My pleasure🙏

PR #2468, opened 2021-06-09
Implement ClassLabel encoding in JSON loader

### LINKED ISSUE #2365: Missing ClassLabel encoding in Json loader
Currently if you want to load a json dataset this way
```python
dataset = load_dataset("json", data_files=data_files, features=features)
```
Then if your features has ClassLabel types and if your json data needs class label encoding (i.e. if the labels in the json files are strings and not integers), then it would fail:
```python
[...]
~/Desktop/hf/datasets/src/datasets/packaged_modules/json/json.py in _generate_tables(self, files)
     94             if self.config.schema:
     95                 # Cast allows str <-> int/float, while parse_option explicit_schema does NOT
---> 96                 pa_table = pa_table.cast(self.config.schema)
     97             yield i, pa_table
[...]
ArrowInvalid: Failed to parse string: 'O' as a scalar of type int64
```

This is because it just tries to cast the string data to integers, without applying the mapping str->int first

The current workaround is to do instead
```python
dataset = load_dataset("json", data_files=data_files)
dataset = dataset.map(features.encode_example, features=features)
```

albertvillanova:
Close #2365.


lhoestq (approved):
Sorry for the delay to review this one
This looks all good, feel free to merge if you have no other changes to make @albertvillanova

albertvillanova:
No, nevermind @lhoestq. Thanks to you for your reviews!

PR #2469, opened 2021-06-09
Bump tqdm version

lewtun:


lewtun:
i tried both the latest version of `tqdm` and the version required by `autonlp` - no luck with windows 😞 

it's very weird that a progress bar would trigger these kind of errors, so i'll have a look to see if it's something unique to `datasets`

lewtun:
Closing since this is now fixed in #2482

PR #2473, opened 2021-06-10
Add Disfl-QA

bhavitvyamalik:
Dataset: https://github.com/google-research-datasets/disfl-qa

To-Do: Update README.md and add YAML tags


lhoestq (commented):
Thanks for adding this datasets. This looks all good so far :)

I'm wondering if we should load the actual squad_v2 data or not. It would be more convenient to have the data from squad_v2 and in the squadformat. WDYT ?

bhavitvyamalik:
Sounds great! It'll make things easier for the user while accessing the dataset. I'll make some changes to the current file then.

datasets/disfl_qa/disfl_qa.py
```diff
@@ -0,0 +1,192 @@
...
+            # If there's a common (input, target) tuple from the features,
+            # specify them here. They'll be used if as_supervised=True in
+            # builder.as_dataset.
+            supervised_keys=None,
+            # Homepage of the dataset for documentation
+            homepage=_HOMEPAGE,
+            # License for the dataset if available
+            license=_LICENSE,
+            # Citation for the dataset
+            citation=_CITATION,
```
lhoestq, line 95: To automate the task formatting you can add this, as in squad or squad_v2 :)

```suggestion
            citation=_CITATION,
            task_templates=[
                QuestionAnsweringExtractive(
                    question_column="disfluent question", context_column="context", answers_column="answers"
                )
            ],
```

datasets/disfl_qa/README.md
```diff
@@ -0,0 +1,179 @@
...
+languages:
+- en-US
+licenses:
+- cc-by-4.0
+multilinguality:
+- monolingual
+pretty_name: 'DISFL-QA: A Benchmark Dataset for Understanding Disfluencies in Question
+  Answering'
+size_categories:
+- unknown
```
lhoestq, line 15:
```suggestion
- 10K<n<100K
```

datasets/disfl_qa/disfl_qa.py
```diff
@@ -0,0 +1,198 @@
...
+            task_templates=[
+                QuestionAnsweringExtractive(
+                    question_column="disfluent question", context_column="context", answers_column="answers"
+                )
+            ],
```
lhoestq, line 101: Awesome !

datasets/disfl_qa/disfl_qa.py
```diff
@@ -0,0 +1,198 @@
...
+        """Returns SplitGenerators."""
+
+        squad_v2_downloaded_files = dl_manager.download_and_extract(_URLS_squad_v2)
+        merge_squad_v2_json = {}
+
+        for file in squad_v2_downloaded_files:
+            with open(squad_v2_downloaded_files[file], encoding="utf-8") as f:
+                merge_squad_v2_json.update(json.load(f))
+
+        squad_v2_dict = _helper_dict(merge_squad_v2_json)  # contains all squad_v2 data in a dict with id as key
```
lhoestq, line 114: Would be nice to move this part into the `_generate_examples` method. It is in `_generate_examples` that we open files and process data, while `_split_generators` must be used only to define the splits and download their required resources

lhoestq (commented):
Nice thank you ! :D 

I added my final comments. My only concern is just about some part of the code that should be moved in `_generate_examples` rather than `_split_generators`.

But apart from that, the code and the dataset card are all good !

lhoestq (approved):
Perfect thanks :)

PR #2477, opened 2021-06-11
Fix docs custom stable version

albertvillanova:
Currently docs default version is 1.5.0. This PR fixes this and sets the latest version instead.

### DIFF
--- a/benchmarks/benchmark_array_xd.py
+++ b/benchmarks/benchmark_array_xd.py
@@ -2,11 +2,10 @@ import json
 import os
 import tempfile
 
-from utils import generate_examples, get_duration
-
 import datasets
 from datasets.arrow_writer import ArrowWriter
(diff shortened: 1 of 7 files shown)


albertvillanova:
I see that @lhoestq overlooked this PR with his commit 07e2b05. 😢 

I'm adding a script so that this issue does not happen again.

lhoestq:
I think we just need to clarify the release process in setup.py instead of adding a script that does the replacement

albertvillanova:
@lhoestq I really think we should implement a script that performs the release (instead of doing it manually as it is done now), as it is already the case in `transformers`. I will do it in a next PR.

For the moment, this PR includes one of the steps of the release script.

lhoestq (approved):
Let's automate this then :)

PR #2483, opened 2021-06-11
Use gc.collect only when needed to avoid slow downs

lhoestq:
In https://github.com/huggingface/datasets/commit/42320a110d9d072703814e1f630a0d90d626a1e6 we added a call to gc.collect to resolve some issues on windows (see https://github.com/huggingface/datasets/pull/2482)

However calling gc.collect too often causes significant slow downs (the CI run time doubled).
So I just moved the gc.collect call to the exact place where it's actually needed: when post-processing a dataset

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -553,10 +553,6 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
             del self._data
         if hasattr(self, "_indices"):
             del self._indices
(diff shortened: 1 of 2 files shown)


albertvillanova:
I continue thinking that the origin of the issue has to do with tqdm (and not with Arrow): this issue only arises for version 4.50.0 (and later) of tqdm, not for previous versions of tqdm.

My guess is that tqdm made a change from version 4.50.0 that does not properly release the iterable.

aDrIaNo34500:
FR

PR #2490, opened 2021-06-12
Allow latest pyarrow version

albertvillanova:
Allow latest pyarrow version, once that version 4.0.1 fixes the segfault bug introduced in version 4.0.0.

Close #2489.

PR #2497, opened 2021-06-14
Use default cast for sliced list arrays if pyarrow >= 4

albertvillanova:
From pyarrow version 4, it is supported to cast sliced lists.

This PR uses default pyarrow cast in Datasets to cast sliced list arrays if pyarrow version is >= 4.

In relation with PR #2461 and #2490.

cc: @lhoestq, @abhi1thakur, @SBrandeis

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -935,7 +935,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
         schema = pa.schema({col_name: type[col_name].type for col_name in self._data.column_names})
         dataset = self.with_format("arrow")
         dataset = dataset.map(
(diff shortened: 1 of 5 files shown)


SBrandeis:
I believe we don't use PyArrow >= 4.0.0 because of some segfault issues:
https://github.com/huggingface/datasets/blob/1206ffbcd42dda415f6bfb3d5040708f50413c93/setup.py#L78
Can you confirm @lhoestq ?

albertvillanova:
@SBrandeis pyarrow version 4.0.1 has fixed that issue: #2490 😉

lhoestq (approved):
Nice thanks ! I didn't know it was supported now :)

PR #2500, opened 2021-06-14
Add load_dataset_builder

### LINKED ISSUE #2484: Implement loading a dataset builder
As discussed with @stas00 and @lhoestq, this would allow things like:

```python
dataset_name = "openwebtext"
builder = load_dataset_builder(dataset_name)
print(builder.cache_dir)
```

mariosasko:
Adds the `load_dataset_builder` function. The good thing is that we can reuse this function to load the dataset info without downloading the dataset itself.

TODOs:
- [x] Add docstring and entry in the docs
- [x] Add tests

Closes #2484


src/datasets/load.py
```diff
@@ -621,6 +621,50 @@ def load_metric(
...
+        data_dir=data_dir,
+        data_files=data_files,
+        hash=hash,
+        features=features,
+        **config_kwargs,
+    )
+
+    if _return_resolved_file_path:
+        return builder_instance, resolved_file_path
+    return builder_instance
```
mariosasko, line 665: This part is not very nice. Maybe it's better to define `base_path` as an optional attribute of `DatasetBuilder` and then in `DatasetBuilder.download_and_prepare` we can do the following:
```python
base_path = base_path if base_path is not None else self._base_path
```

albertvillanova:
Hi @mariosasko, thanks for taking on this issue.

Just a few logistic suggestions, as you are one of our most active contributors ❤️ :
- When you start working on an issue, you can self-assign it to you by commenting on the issue page with the keywords: `#self-assign`; we have implemented a GitHub Action to take care of that... 😉 
- When you are still working on your Pull Request, instead of using the `[WIP]` in the PR name, you can instead create a *draft* pull request: use the drop-down (on the right of the *Create Pull Request* button) and select **Create Draft Pull Request**, then click **Draft Pull Request**.

I hope you find this hints useful. 🤗

mariosasko:
@albertvillanova Thanks for the tips. When creating this PR, it slipped my mind that this should be a draft. GH has an option to convert already created PRs to draft PRs, but this requires write access for the repo, so maybe you can help.

src/datasets/load.py
lhoestq: Good idea !

mariosasko:
Ready for the review!

One additional change. I've modified the `camelcase_to_snakecase`/`snakecase_to_camelcase` conversion functions to fix conversion of the names with 2 or more underscores (e.g. `camelcase_to_snakecase("__DummyDataset__")` would return `___dummy_dataset__`; notice one extra underscore at the beginning). The implementation is based on the [inflection](https://pypi.org/project/inflection/) library.

src/datasets/packaged_modules/__init__.py
```diff
@@ -12,9 +12,9 @@
 def hash_python_lines(lines: List[str]) -> str:
     filtered_lines = []
     for line in lines:
-        line.replace("\n", "")  # remove line breaks, white space and comments
-        line.replace(" ", "")
-        line.replace("\t", "")
+        line = line.replace("\n", "")  # remove line breaks, white space and comments
+        line = line.replace(" ", "")
+        line = line.replace("\t", "")
```
lhoestq, line 17: I would just remove this replacements.
They could create some errors (return the same hash while the code is different). And since it would change the hashes, it means that all the cache directories that have been created until now would now be ignored. 

```suggestion
```

lhoestq (commented):
Thank you for adding `load_dataset_builder` :)
This is really helpful.

I just have one comment about the part that hashes the builder's code:

lhoestq (approved):
Looks all good now !

Thank you for adding this :)

docs/source/loading_datasets.rst
```diff
@@ -431,7 +431,7 @@ For example, run the following to skip integrity verifications when loading the
 Loading datasets offline
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Each dataset builder (e.g. "squad") is a python script that is downloaded and cached from either from the huggingface/datasets GitHub repository or from the `HuggingFace Hub <https://huggingface.co/datasets>`__.
+Each dataset builder (e.g. "squad") is a python script that is downloaded and cached from either from the 🤗Datasets GitHub repository or from the `HuggingFace Hub <https://huggingface.co/datasets>`__.
```
stas00, line 434: I'd suggested a white space here: 🤗Datasets => 🤗 Datasets

docs/source/loading_datasets.rst
```diff
@@ -441,6 +441,29 @@ You can even set the environment variable `HF_DATASETS_OFFLINE` to ``1`` to tell
 This mode disables all the network calls of the library.
 This way, instead of waiting for a dataset builder download to time out, the library looks directly at the cache.
 
+.. _load_dataset_load_builder:
```
stas00, line 444: .. load_dataset_builder?

stas00:
Thank you for adding this feature, @mariosasko - this is really awesome!

Tried with:
```
python -c "from datasets import load_dataset_builder; b = load_dataset_builder('openwebtext-10k'); print(b.cache_dir)"
Using the latest cached version of the module from /home/stas/.cache/huggingface/modules/datasets_modules/datasets/openwebtext-10k/3a8df094c671b4cb63ed0b41f40fb3bd855e9ce2e3765e5df50abcdfb5ec144b (last modified on Wed May 12 20:22:53 2021) since it couldn't be found locally at openwebtext-10k/openwebtext-10k.py or remotely (FileNotFoundError).
/home/stas/.cache/huggingface/datasets/openwebtext10k/plain_text/1.0.0/3a8df094c671b4cb63ed0b41f40fb3bd855e9ce2e3765e5df50abcdfb5ec144b
```

The logger message is a bit confusing to the user - that is what does `FileNotFoundError` refer to? 

1. May be replace  `FileNotFoundError` with where it was looking for a file online. But then the remote file is there - it's found 
2. I'm not sure why it says "since it couldn't be found locally" - as it is locally found at the cache folder and again what does " locally at openwebtext-10k/openwebtext-10k.py" mean - i.e. where does it look for it? Is it `./openwebtext-10k/openwebtext-10k.py` it's looking for? or in some specific dir?

If the cached version always supersedes any other versions perhaps this is what it should say?

found cached version at xxx, not looking for a local at yyy, not downloading remote at zzz

lhoestq:
Hi ! Thanks for the comments

Regarding your last message:
You must pass `stas/openwebtext-10k` as in `load_dataset`, otherwise it doesn't know how to retrieve the builder from the HF Hub.

When you specify a dataset name without a slash, it tries to load a canonical dataset or it looks locally at ./openwebtext-10k/openwebtext-10k.py
Here since `openwebtext-10k` is not a canonical dataset and doesn't exist locally at ./openwebtext-10k/openwebtext-10k.py: it raised a FileNotFoundError.
As a fallback it managed to find the dataset script in your cache and it used this one.

docs/source/loading_datasets.rst
lhoestq: `load_dataset` is the prefix of all the anchors of this page of the documentation x)

stas00:
Oh, I see, so I actually used an incorrect input. so it was a user error. Correcting it:

```
python -c "from datasets import load_dataset_builder; b = load_dataset_builder('stas/openwebtext-10k'); print(b.cache_dir)"
/home/stas/.cache/huggingface/datasets/openwebtext10k/plain_text/1.0.0/3a8df094c671b4cb63ed0b41f40fb3bd855e9ce2e3765e5df50abcdfb5ec144b
```

Now there is no logger message. Got it!

OK, I'm not sure the magical recovery it did in first place is most beneficial in the long run. I'd have rather it failed and said: "incorrect input there is no such dataset as 'openwebtext-10k' at <this path> or <this url>" - because if it doesn't fail I may leave it in the code and it'll fail later when another user tries to use my code and won't have the cache. Does it make sense? Giving me `this url` allows me to go to the datasets hub and realize that the dataset is missing the username qualifier.

> Here since openwebtext-10k is not a canonical dataset and doesn't exist locally at ./openwebtext-10k/openwebtext-10k.py: it raised a FileNotFoundError.

Except it slapped the exception name to ` remotely (FileNotFoundError).` which makes no sense.

Plus for the local it's not clear where is it looking relatively too when it gets `FileNotFoundError` - perhaps it'd help to use absolute path and use it in the message?

---------------

Finally, the logger format is not set up so the user gets a warning w/o knowing it's a warning. As you can see it's missing the WARNING pre-amble in  https://github.com/huggingface/datasets/pull/2500#issuecomment-874250500

Newread23 (commented):
#7624 #5481 #5481

PR #2506, opened 2021-06-15
Add course banner

sgugger:
This PR adds a course banner similar to the one you can now see in the [Transformers repo](https://github.com/huggingface/transformers) that links to the course. Let me know if placement seems right to you or not, I can move it just below the badges too.


albertvillanova (approved):
Thanks. It is really a good idea to have this banner to the Course!!! 🤗 

For the location I think it is OK...

PR #2507, opened 2021-06-15
Rearrange JSON field names to match passed features schema field names

### LINKED ISSUE #2366: Json loader fails if user-specified features don't match the json data fields order

albertvillanova:
This PR depends on PR #2453 (which must be merged first).

Close #2366.


lhoestq (approved):
Thanks for fixing this !

Issue #2443, opened 2021-06-03
Some tests hang on Windows


mariosasko:
Currently, several tests hang on Windows if the max path limit of 260 characters is not disabled. This happens due to the changes introduced by #2223 that cause an infinite loop in `WindowsFileLock` described in #2220.  This can be very tricky to debug, so I think now is a good time to address these issues/PRs. IMO throwing an error is too harsh, but maybe we can emit a warning in the top-level `__init__.py ` on startup if long paths are not enabled.

lhoestq:
Hi ! That would be nice indeed to at least have a warning, since we don't handle the max path length limit.
Also if we could have an error instead of an infinite loop I'm sure windows users will appreciate that

albertvillanova:
Unfortunately, I know this problem very well... 😅 

I remember having proposed to throw an error instead of hanging in an infinite loop #2220: 60c7d1b6b71469599a27147a08100f594e7a3f84, 8c8ab60018b00463edf1eca500e434ff061546fc 
but @lhoestq told me:
> Note that the filelock module comes from this project that hasn't changed in years - while still being used by ten of thousands of projects:
https://github.com/benediktschmitt/py-filelock
> 
> Unless we have proper tests for this, I wouldn't recommend to change it

I opened an Issue requesting a warning/error at startup for that case: #2224

mariosasko:
@albertvillanova Thanks for additional info on this issue.

Yes, I think the best option is to throw an error instead of suppressing it in a loop. I've considered 2 more options, but I don't really like them:
1. create a temporary file with a filename longer than 255 characters on import; if this fails, long paths are not enabled and raise a warning. I'm not sure about this approach because I don't like the idea of creating a temporary file on import for this purpose.
2. check if long paths are enabled with [this code](https://stackoverflow.com/a/46546731/14095927). As mentioned in the comment, this code relies on an undocumented function and Win10-specific.

PR #2509, opened 2021-06-16
Fix fingerprint when moving cache dir

### LINKED ISSUE #2496: Dataset fingerprint changes after moving the cache directory, which prevent cache reload when using `map`
`Dataset.map` uses the dataset fingerprint (a hash) for caching.
However the fingerprint seems to change when someone moves the cache directory of the dataset.

This is because it uses the default fingerprint generation:
1. the dataset path is used to get the fingerprint
2. the modification times of the arrow file is also used to get the fingerprint

To fix that we could set the fingerprint of the dataset to be a hash of (<dataset_name>, <config_name>, <version>, <script_hash>), i.e. a hash of the the cache path relative to the cache directory.

lhoestq:
The fingerprint of a dataset changes if the cache directory is moved.
I fixed that by setting the fingerprint to be the hash of:
- the relative cache dir (dataset_name/version/config_id)
- the requested split

Close #2496 

I had to fix an issue with the filelock filename that was too long (>255). It prevented the tests to run on my machine. I just added `hash_filename_if_too_long` in case this happens, to not get filenames longer than 255.
We usually have long filenames for filelocks because they are named after the path that is being locked. In case the path is a cache directory that has long directory names, then the filelock filename could en up being very long.

### DIFF
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -847,7 +847,7 @@ class DatasetBuilder:
         )
         hasher = Hasher()
         hasher.update(self._relative_data_dir().replace(os.sep, "/"))
-        hasher.update(split.to_spec() if isinstance(split, ReadInstruction) else str(split))
+        hasher.update(str(split))
         fingerprint = hasher.hexdigest()
         return Dataset(fingerprint=fingerprint, **dataset_kwargs)
 
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -25,7 +25,6 @@ import shutil
 import urllib
 from dataclasses import dataclass
 from functools import partial
-from pathlib import Path
 from typing import Dict, List, Optional, Tuple, Union
 
 from datasets.features import Features
@@ -393,11 +392,11 @@ class DatasetBuilder:
         hash = self.hash
         if builder_config:
             # use the enriched name instead of the name to make it unique
-            builder_data_dir = Path(builder_data_dir, self.config_id).as_posix()
+            builder_data_dir = os.path.join(builder_data_dir, self.config_id)
         if with_version:
-            builder_data_dir = Path(builder_data_dir, str(self.config.version)).as_posix()
+            builder_data_dir = os.path.join(builder_data_dir, str(self.config.version))
         if with_hash and hash and isinstance(hash, str):
-            builder_data_dir = Path(builder_data_dir, hash).as_posix()
+            builder_data_dir = os.path.join(builder_data_dir, hash)
         return builder_data_dir
 
     def _build_cache_dir(self):
@@ -847,7 +846,7 @@ class DatasetBuilder:
             in_memory=in_memory,
         )
         hasher = Hasher()
-        hasher.update(self._relative_data_dir())
+        hasher.update(self._relative_data_dir().replace(os.sep, "/"))
         hasher.update(split.to_spec() if isinstance(split, ReadInstruction) else str(split))
         fingerprint = hasher.hexdigest()
         return Dataset(fingerprint=fingerprint, **dataset_kwargs)
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -25,6 +25,7 @@ import shutil
 import urllib
 from dataclasses import dataclass
 from functools import partial
+from pathlib import Path
 from typing import Dict, List, Optional, Tuple, Union
 
 from datasets.features import Features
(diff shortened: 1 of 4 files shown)


lhoestq:
Windows, why are you doing this ?

albertvillanova:
Thanks @lhoestq, I'm starting reviewing this PR.

src/datasets/builder.py
```diff
@@ -845,7 +845,11 @@ def _as_dataset(self, split: Union[ReadInstruction, Split] = Split.TRAIN, in_mem
             split_infos=self.info.splits.values(),
             in_memory=in_memory,
         )
-        return Dataset(**dataset_kwargs)
+        hasher = Hasher()
+        hasher.update(self._relative_data_dir().replace(os.sep, "/"))
+        hasher.update(split.to_spec() if isinstance(split, ReadInstruction) else str(split))
+        fingerprint = hasher.hexdigest()
```
albertvillanova, line 851: This might be extracted into a funcion/method.

src/datasets/utils/filelock.py
```diff
@@ -96,6 +104,24 @@ def __str__(self):
         return temp
 
 
+# Helpers
+# ------------------------------------------------
+
+
+MAX_LOCK_FILENAME_LENGTH = 255
```
albertvillanova, line 111: Is this a parameter that the user should be allowed to customize? If yes, I would suggest moving it to `config`; if not (I think this is the case), then I would suggest to remove the variable and pass the value directly in the function signature of `hash_filename_if_too_long`.

src/datasets/utils/filelock.py
```diff
@@ -96,6 +104,24 @@ def __str__(self):
...
+
+
+def hash_filename_if_too_long(path: str, max_length=MAX_LOCK_FILENAME_LENGTH) -> str:
```
albertvillanova, line 114: As this function is only used by FileLock, maybe it could become a method of that class: it is the class responsibility to check whether the filename is too long and hash it.

albertvillanova (changes requested):
Some comments below.

src/datasets/utils/filelock.py
albertvillanova: And I would recommend adding some unit tests for this function (even if they seem quite trivial).

mariosasko (commented):
I was hoping `hash_filename_if_too_long` would fix issues with FileLock on Windows if long paths are not enabled (#2443), but sadly it doesn't.

lhoestq:
Yea issues on windows are about long paths, not long filenames.
We can make sure the lock filenames are not too long, but not for the paths

src/datasets/builder.py
```diff
@@ -845,7 +845,11 @@ def _as_dataset(self, split: Union[ReadInstruction, Split] = Split.TRAIN, in_mem
             split_infos=self.info.splits.values(),
             in_memory=in_memory,
         )
-        return Dataset(**dataset_kwargs)
+        hasher = Hasher()
+        hasher.update(self._relative_data_dir().replace(os.sep, "/"))
+        hasher.update(split.to_spec() if isinstance(split, ReadInstruction) else str(split))
```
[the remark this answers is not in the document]
lhoestq: Nice thanks !

lhoestq:
Took your suggestions into account @albertvillanova :)

tests/test_filelock.py
```diff
@@ -0,0 +1,29 @@
...
+
+import pytest
+
+from datasets.utils.filelock import FileLock, Timeout
+
+
+def test_filelock(tmpdir):
+    lock1 = FileLock(tmpdir / "foo.lock")
+    lock2 = FileLock(tmpdir / "foo.lock")
+    timeout = 0.01
```
albertvillanova, line 12: Are you really willing to include a test that increases time latency by 0.01?

tests/test_filelock.py
```diff
@@ -0,0 +1,29 @@
...
+
+def test_long_filename(tmpdir):
+    filename = "a" * 1000 + ".lock"
+    lock1 = FileLock(tmpdir / filename)
+    assert lock1._lock_file.endswith(".lock")
+    assert not lock1._lock_file.endswith(filename)
+    assert len(os.path.basename(lock1._lock_file)) <= 255
+    lock2 = FileLock(tmpdir / filename)
+    with lock1.acquire():
+        with pytest.raises(Timeout):
```
albertvillanova, line 28: Increased latency by 0.01 two times?

We are already facing too-long test suite issue...

src/datasets/utils/filelock.py
```diff
@@ -125,8 +132,10 @@ class BaseFileLock:
     Implements the base class of a file lock.
     """
 
-    def __init__(self, lock_file, timeout=-1):
+    def __init__(self, lock_file, timeout=-1, filename_max_length=255):
```
albertvillanova, line 135: Sorry for nitpicking, but maybe `max_filename_length` instead? I leave to your judgment.

albertvillanova (approved):
Once the code quality check is fixed, it is OK for me.
Thanks!

tests/test_filelock.py
lhoestq: I think that's ok

tests/test_filelock.py
lhoestq: This is nothing compared to the tests that take several seconds

src/datasets/utils/filelock.py
lhoestq: doing the change right now :)

Issue #2434, opened 2021-05-31
Extend QuestionAnsweringExtractive template to handle nested columns


lewtun:
Currently the `QuestionAnsweringExtractive` task template and `preprare_for_task` only support "flat" features. We should extend the functionality to cover QA datasets like:

* `iapp_wiki_qa_squad`
* `parsinlu_reading_comprehension`

where the nested features differ with those from `squad` and trigger an `ArrowNotImplementedError`:

```
---------------------------------------------------------------------------
ArrowNotImplementedError                  Traceback (most recent call last)
<ipython-input-12-50e5b8f69c20> in <module>
----> 1 ds.prepare_for_task("question-answering-extractive")[0]

~/git/datasets/src/datasets/arrow_dataset.py in prepare_for_task(self, task)
   1436         # We found a template so now flush `DatasetInfo` to skip the template update in `DatasetInfo.__post_init__`
   1437         dataset.info.task_templates = None
-> 1438         dataset = dataset.cast(features=template.features)
   1439         return dataset
   1440 

~/git/datasets/src/datasets/arrow_dataset.py in cast(self, features, batch_size, keep_in_memory, load_from_cache_file, cache_file_name, writer_batch_size, num_proc)
    977         format = self.format
    978         dataset = self.with_format("arrow")
--> 979         dataset = dataset.map(
    980             lambda t: t.cast(schema),
    981             batched=True,

~/git/datasets/src/datasets/arrow_dataset.py in map(self, function, with_indices, input_columns, batched, batch_size, drop_last_batch, remove_columns, keep_in_memory, load_from_cache_file, cache_file_name, writer_batch_size, features, disable_nullable, fn_kwargs, num_proc, suffix_template, new_fingerprint, desc)
   1600 
   1601         if num_proc is None or num_proc == 1:
-> 1602             return self._map_single(
   1603                 function=function,
   1604                 with_indices=with_indices,

~/git/datasets/src/datasets/arrow_dataset.py in wrapper(*args, **kwargs)
    176         }
    177         # apply actual function
--> 178         out: Union["Dataset", "DatasetDict"] = func(self, *args, **kwargs)
    179         datasets: List["Dataset"] = list(out.values()) if isinstance(out, dict) else [out]
    180         # re-apply format to the output

~/git/datasets/src/datasets/fingerprint.py in wrapper(*args, **kwargs)
    395             # Call actual function
    396 
--> 397             out = func(self, *args, **kwargs)
    398 
[...]

PR #2518, opened 2021-06-18
Add task templates for tydiqa and xquad

lewtun:
This PR adds question-answering templates to the remaining datasets that are linked to a model on the Hub.

Notes: 

* I could not test the tydiqa implementation since I don't have enough disk space 😢 . But I am confident the template works :)
* there exist other datasets like `fquad` and `mlqa` which are candidates for question-answering templates, but some work is needed to handle the ordering of nested column described in #2434


lhoestq:
Just tested TydiQA and it works fine :)

lhoestq (approved):
Thanks for adding the task templates and the tags :)

I also updated the dataset_infos.json files to include the task templates infos

PR #2524, opened 2021-06-20
Raise FileNotFoundError in WindowsFileLock

### LINKED ISSUE #2443: Some tests hang on Windows

mariosasko:
Closes #2443

### DIFF
--- a/src/datasets/utils/filelock.py
+++ b/src/datasets/utils/filelock.py
@@ -338,6 +338,8 @@ class WindowsFileLock(BaseFileLock):
 
         try:
             fd = os.open(self._lock_file, open_mode)
+        except FileNotFoundError:
+            raise
         except OSError:
             pass
         else:


lhoestq:
Hi ! Could you clarify what it fixes exactly and give more details please ?

lhoestq (approved):

mariosasko:
This has already been merged, but I'll clarify the idea of this PR. Before this merge, FileLock was the only component affected by the max path limit on Windows (that came to my notice) because of its infinite loop that would suppress errors. So instead of suppressing the `FileNotFoundError` that is thrown by `os.open` if the file name is longer than the max allowed path length, this PR reraises it to notify the user.

PR #2525, opened 2021-06-21
Use scikit-learn package rather than sklearn in setup.py

lesteve:
The sklearn package is an historical thing and should probably not be used by anyone, see https://github.com/scikit-learn/scikit-learn/issues/8215#issuecomment-344679114 for some caveats.

Note: this affects only TESTS_REQUIRE so I guess only developers not end users.


lhoestq (approved):
Good catch thank you !

PR #2529, opened 2021-06-21
Add summarization template

lewtun:
This PR adds a task template for text summarization. As far as I can tell, we do not need to distinguish between "extractive" or "abstractive" summarization - both can be handled with this template.

Usage:

```python
from datasets.tasks import Summarization

ds = load_dataset("xsum", split="train")
# Dataset({
#     features: ['document', 'summary', 'id'],
#     num_rows: 204045
# })

summarization = Summarization(text_column="document", summary_column="summary")
ds.prepare_for_task(summarization)
# Dataset({
#     features: ['text', 'summary'],
#     num_rows: 204045
# })
```

### DIFF
--- a/tests/test_arrow_dataset.py
+++ b/tests/test_arrow_dataset.py
@@ -705,172 +705,6 @@ class BaseDatasetTest(TestCase):
                     self.assertEqual(dset_concat.info.description, "Dataset1\n\nDataset2\n\n")
             del dset1, dset2, dset3
 
-    def test_concatenate_with_no_task_templates(self, in_memory):
-        info = DatasetInfo(task_templates=None)
-        data = {"text": ["i love transformers!"], "labels": [1]}
-        with tempfile.TemporaryDirectory() as tmp_dir:
-            with Dataset.from_dict(data, info=info) as dset1, Dataset.from_dict(
-                data, info=info
-            ) as dset2, Dataset.from_dict(data, info=info) as dset3:
-                with self._to(in_memory, tmp_dir, dset1) as dset1, self._to(
-                    in_memory, tmp_dir, dset2
-                ) as dset2, self._to(in_memory, tmp_dir, dset3) as dset3:
-                    with concatenate_datasets([dset1, dset2, dset3]) as dset_concat:
-                        self.assertEqual(dset_concat.info.task_templates, None)
-
-    def test_concatenate_with_equal_task_templates(self, in_memory):
-        labels = ["neg", "pos"]
-        task_template = TextClassification(text_column="text", label_column="labels", labels=labels)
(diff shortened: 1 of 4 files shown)


tests/test_arrow_dataset.py
```diff
@@ -2144,6 +2144,37 @@ def test_task_question_answering(self, in_memory):
...
+        features_after_cast = Features({"text": Value("string"), "summary": Value("string")})
+        task = Summarization(text_column="input_text", summary_column="input_summary")
+        info = DatasetInfo(features=features_before_cast, task_templates=task)
+        data = {
+            "input_text": ["jack and jill took a taxi to attend a super duper party in the city."],
+            "input_summary": ["jack and jill attend party"],
+            "dummy": ["123456"],
+        }
+        # Test we can load from task name
+        with Dataset.from_dict(data, info=info) as dset:
```
lewtun, line 2161: it seems that whatever magic @lhoestq did with windows now allows me to avoid using the annoying context manager that we have in many unit tests:

```python
with tempfile.TemporaryDirectory() as tmp_dir, Dataset.from_dict(data) as dset:
    with self._to(in_memory, tmp_dir, dset) as dset:
        # do operations on `dset`
```

i never really understood _why_ we had to do this so please correct me if it's actually needed 😄

SBrandeis (approved):
LGTM 🚀

tests/test_arrow_dataset.py
lhoestq: The `with` statement are used to properly close open arrow files on windows (it really doesn't like keeping things open).
What you mention though is the call to `self._to` call that moves a dataset on disk or in memory, which can be useful in some tests. Here you don't need to do this indeed.

Since the `in_memory` parameter is not used in this test, you can move this test outside of the `BaseDatasetTest` class and have it as a regular pytest test case.
This way it won't be run twice (once for `in_memory=True` and once for `in_memory=False`)

lhoestq (commented):
Nice thanks !
Could you just move the test outside of the BaseDatasetTest class please ? Otherwise it will unnecessarily be run twice.

lewtun:
> Nice thanks !
> Could you just move the test outside of the BaseDatasetTest class please ? Otherwise it will unnecessarily be run twice.

sure, on it! thanks for the explanations about the `self._to` method :)

lewtun:
@lhoestq i've moved all the task template tests outside of `BaseDatasetTest` and collected them in their dedicated test case. (at some point i'll revisit this so we can just use `pytest` natively, but the PR is already getting out-of-scope :))

lhoestq (approved):
Looks all good, thanks :)

PR #2530, opened 2021-06-22
Fixed label parsing in the ProductReviews dataset

yavuzKomecoglu:
Fixed issue with parsing dataset labels.


yavuzKomecoglu:
@lhoestq, can you please review this PR?
What exactly is the problem in the test case? Should it matter?

lhoestq:
Hi ! Thanks for fixing this :)

The CI fails for two reasons:
- the `pretty_name` tag is missing in yaml tags in ./datasets/turkish_product_reviews/README.md. You can fix that by adding this in the yaml tags:
```yaml
pretty_name: Turkish Product Reviews
```
- The test that runs the turkish_product_reviews.py file on the dummy_data.zip data returned 0 examples. Indeed it looks like you changed dummy_data.zip file and now it is an empty zip file. I think you can fix that by reverting your change to the dummy_data.zip file

lhoestq (approved):
Thanks !
The CI was failing because of the dataset card that was missing some sections. I fixed that.

It's all good now

yavuzKomecoglu:
> Thanks !
> The CI was failing because of the dataset card that was missing some sections. I fixed that.
> 
> It's all good now

Super.  Thanks for the support.

PR #2533, opened 2021-06-22
Add task template for automatic speech recognition

lewtun:
This PR adds a task template for automatic speech recognition. In this task, the input is a path to an audio file which the model consumes to produce a transcription.

Usage:

```python
from datasets.tasks import AutomaticSpeechRecognition

ds = load_dataset("timit_asr", split="train[:10]")
# Dataset({
#     features: ['file', 'text', 'phonetic_detail', 'word_detail', 'dialect_region', 'sentence_type', 'speaker_id', 'id'],
#     num_rows: 10
# })

task = AutomaticSpeechRecognition(audio_file_column="file", transcription_column="text")
ds.prepare_for_task(task)
# Dataset({
#     features: ['audio_file', 'transcription'],
#     num_rows: 10
# })
```

### DIFF
--- a/src/datasets/tasks/automatic_speech_recognition.py
+++ b/src/datasets/tasks/automatic_speech_recognition.py
@@ -8,6 +8,8 @@ from .base import TaskTemplate
 @dataclass(frozen=True)
 class AutomaticSpeechRecognition(TaskTemplate):
     task: str = "automatic-speech-recognition"
+    # TODO(lewtun): Replace input path feature with dedicated `Audio` features
+    # when https://github.com/huggingface/datasets/pull/2324 is implemented
     input_schema: ClassVar[Features] = Features({"audio_file_path": Value("string")})
     label_schema: ClassVar[Features] = Features({"transcription": Value("string")})
     audio_file_path_column: str = "audio_file_path"
--- a/src/datasets/tasks/automatic_speech_recognition.py
+++ b/src/datasets/tasks/automatic_speech_recognition.py
@@ -8,11 +8,11 @@ from .base import TaskTemplate
 @dataclass(frozen=True)
 class AutomaticSpeechRecognition(TaskTemplate):
     task: str = "automatic-speech-recognition"
-    input_schema: ClassVar[Features] = Features({"audio_file": Value("string")})
+    input_schema: ClassVar[Features] = Features({"audio_file_path": Value("string")})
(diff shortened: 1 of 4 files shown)


src/datasets/tasks/automatic_speech_recognition.py
```diff
@@ -0,0 +1,18 @@
...
+from typing import ClassVar, Dict
+
+from ..features import Features, Value
+from .base import TaskTemplate
+
+
+@dataclass(frozen=True)
+class AutomaticSpeechRecognition(TaskTemplate):
+    task: str = "automatic-speech-recognition"
+    input_schema: ClassVar[Features] = Features({"audio_file": Value("string")})
```
lhoestq, line 11: Note that in the feature we'll have an Audio feature type that will probably have additional parameters (like ClassLabel) such as the sampling rate or the audio format.

SBrandeis, line 11: Maybe `audio_file_path` would be more explicit on what the column represent ?

tests/test_arrow_dataset.py
```diff
@@ -2144,6 +2144,39 @@ def test_task_question_answering(self, in_memory):
                 )
                 self.assertDictEqual(features_after_cast, dset.features)
 
+    def test_task_automatic_speech_recognition(self, in_memory):
```
lhoestq, line 2147: don't hesitate to move it outside of the BaseDatasetTest class, as mentioned in #2529

src/datasets/tasks/automatic_speech_recognition.py
SBrandeis: Also, paths are not portable between machines a priori. This is probably good enough for now, but at some point, we'll need to replace the `Value("string")` with an `Audio` or `Signal` feature!

tests/test_arrow_dataset.py
lewtun: yep, will do this here as well!

src/datasets/tasks/automatic_speech_recognition.py
lewtun: good to know!

SBrandeis (commented):
Thank you @lewtun !

lewtun:
@SBrandeis @lhoestq i've integrated your suggestions, so this is ready for another review :)

lhoestq:
Merging !

PR #2539, opened 2021-06-23
remove wi_locness dataset due to licensing issues

aseifert:
It was brought to my attention that this dataset's license is not only missing, but also prohibits redistribution. I contacted the original author to apologize for this oversight and asked if we could still use it, but unfortunately we can't and the author kindly asked to take down this dataset.


lhoestq:
Hi ! I'm sorry to hear that.
Though we are not redistributing the dataset, we just provide a python script that downloads and process the dataset from its original source hosted at https://www.cl.cam.ac.uk

Therefore I'm not sure what's the issue with licensing. What do you mean exactly ?

SimonHFL:
I think that the main issue is that the licesenses of the data are not made clear in the huggingface hub – other people wrongly assumed that the data was license-free, which resulted in commercial use, which is against the licenses.
Is it possible to add the licenses from the original download to huggingface? that would help clear any confusion (licenses can be found here: https://www.cl.cam.ac.uk/research/nl/bea2019st/data/wi+locness_v2.1.bea19.tar.gz)

lhoestq:
Thanks for the clarification @SimonHFL 
You're completely right, we need to show the licenses.
I just added the them here: https://huggingface.co/datasets/wi_locness#licensing-information

chrisjbryant:
Hi guys, I'm one of the authors of this dataset. 

To clarify, we're happy for you to keep the data in the repo on 2 conditions:
1. You don't host the data yourself.
2. You make it clear that anyone who downloads the data via HuggingFace should read and abide by the license. 

I think you've now met these conditions, so we're all good, but I just wanted to make it clear in case there are any issues in the future. Thanks again to @aseifert for bringing this to our attention! :)

lhoestq:
Thanks for your message @chrisjbryant :)
I'm closing this PR then.

And thanks for reporting @aseifert

PR #2540, opened 2021-06-23
Remove task templates if required features are removed during `Dataset.map`

lewtun:
This PR fixes a bug reported by @craffel where removing a dataset's columns during `Dataset.map` triggered a `KeyError` because the `TextClassification` template tried to access the removed columns during `DatasetInfo.__post_init__`:

```python

# `yelp_polarity` comes with a `TextClassification` template
ds = load_dataset("yelp_polarity", split="test")
ds
# Dataset({
#     features: ['text', 'label'],
#     num_rows: 38000
# })

# Triggers KeyError: 'label' - oh noes!
ds.map(lambda x: {"inputs": 0}, remove_columns=ds.column_names)
```

I wrote a unit test to make sure I could reproduce the error and then patched a fix.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -1998,9 +1998,10 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
             # Create new Dataset from buffer or file
             info = self.info.copy()
             # Remove task templates if the required features have been removed
-            for idx, task in enumerate(info.task_templates):
-                if not all(k in writer._features.keys() for k in task.features.keys()):
-                    _ = info.task_templates.pop(idx)
+            if info.task_templates:
+                for idx, task in enumerate(info.task_templates):
+                    if not all(k in writer._features.keys() for k in task.features.keys()):
+                        _ = info.task_templates.pop(idx)
             info.features = writer._features
             print(info.features)
             if buf_writer is None:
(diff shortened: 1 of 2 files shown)


src/datasets/arrow_dataset.py
```diff
@@ -1997,6 +1997,11 @@ def init_buffer_and_writer():
         if update_data:
             # Create new Dataset from buffer or file
             info = self.info.copy()
+            # Remove task templates if the required features have been removed
+            if info.task_templates:
+                for idx, task in enumerate(info.task_templates):
+                    if not all(k in writer._features.keys() for k in task.features.keys()):
+                        _ = info.task_templates.pop(idx)
```
SBrandeis, line 2004: Using  `pop` here seems dangerous - what if we remove two templates? The index of the second one is not `idx`

SBrandeis: Personnally, I would have used a list comprehension here:

```python
info.task_templates = [template if all(k in writer._features.keys() for k in template.features.keys()) for template in info.task_templates]
```

SBrandeis (changes requested):

src/datasets/arrow_dataset.py
lewtun: oh yes you are indeed correct - well spotted! i've now switch to using a list comprehension as you suggested

SBrandeis (approved):
LGTM :)

lhoestq (approved):
Thanks for reporting @craffel and for fixing @lewtun :)

PR #2544, opened 2021-06-24
Fix logging levels

### LINKED ISSUE #2543: switching some low-level log.info's to log.debug?
In https://github.com/huggingface/transformers/pull/12276 we are now changing the examples to have `datasets` on the same log level as `transformers`, so that one setting can do a consistent logging across all involved components.

The trouble is that now we get a ton of these:

```
06/23/2021 12:15:31 - INFO - datasets.utils.filelock -   Lock 139627640431136 acquired on /home/stas/.cache/huggingface/metrics/sacrebleu/default/default_experiment-1-0.arrow.lock
06/23/2021 12:15:31 - INFO - datasets.arrow_writer -   Done writing 50 examples in 12280 bytes /home/stas/.cache/huggingface/metrics/sacrebleu/default/default_experiment-1-0.arrow.
06/23/2021 12:15:31 - INFO - datasets.arrow_dataset -   Set __getitem__(key) output type to python objects for no columns  (when key is int or slice) and don't output other (un-formatted) columns.
06/23/2021 12:15:31 - INFO - datasets.utils.filelock -   Lock 139627640431136 released on /home/stas/.cache/huggingface/metrics/sacrebleu/default/default_experiment-1-0.arrow.lock
```

May I suggest that these can be `log.debug` as it's no informative to the user.

More examples: these are not informative - too much information:
```
06/23/2021 12:14:26 - INFO - datasets.load -   Checking /home/stas/.cache/huggingface/datasets/downloads/459933f1fe47711fad2f6ff8110014ff189120b45ad159ef5b8e90ea43a174fa.e23e7d1259a8c6274a82a42a8936dd1b87225302c6dc9b7261beb3bc2daac640.py for additional imports.
06/23/2021 12:14:27 - INFO - datasets.builder -   Constructing Dataset for split train, validation, test, from /home/stas/.cache/huggingface/datasets/wmt16/ro-en/1.0.0/0d9fb3e814712c785176ad8cdb9f465fbe6479000ee6546725db30ad8a8b5f8a

```

While these are:
```
06/23/2021 12:14:27 - INFO - datasets.info -   Loading Dataset Infos from /home/stas/.cache/huggingface/modules/datasets_modules/datasets/wmt16/0d9fb3e814712c785176ad8cdb9f465fbe6479000ee6546725db30ad8a8b5f8a
06/23/2021 12:14:27 - WARNING - datasets.builder -   Reusing dataset wmt16 (/home/stas/.cache/huggingface/datasets/wmt16/ro-en/1.0.0/0d9fb3e814712c785176ad8cdb9f465fbe6479000ee6546725db30ad8a8b5f8a)
```

I also realize that `transformers` examples don't have do use `info` for `datasets` to let the default `warning` keep logging to less noisy.

But I think currently the log levels are slightly misused and skewed by 1 level. Many `warnings` will better be `info`s and most `info`s be `debug`.

e.g.:

```
06/23/2021 12:14:27 - WARNING - datasets.builder -   Reusing dataset wmt16 (/home/stas/.cache/huggingface/datasets/wmt16/ro-en/1.0.0/0d9fb3e814712c785176ad8cdb9f465fbe6479000ee6546725db30ad8a8b5f8a)
```

why is this a warning? it is informing me that the cache is used, there is nothing to be worried about. I'd have it as `info`.

Warnings are typically something that's bordering error or the first thing to check when things don't work as expected.

infrequent info is there to inform of the different stages or important events.

Everything else is debug.

At least the way I understand things.

albertvillanova:
Sometimes default `datasets` logging can be too verbose. One approach could be reducing some logging levels, from info to debug, or from warning to info.

Close #2543.

cc: @stas00

### DIFF
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -729,7 +729,7 @@ class DatasetBuilder:
                 % (self.name, self._cache_dir_root)
             )
 
-        logger.info(
+        logger.debug(
             "Constructing Dataset for split %s, from %s", split or ", ".join(self.info.splits), self._cache_dir
         )
 
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -507,7 +507,7 @@ class DatasetBuilder:
         with FileLock(lock_path):
             data_exists = os.path.exists(self._cache_dir)
             if data_exists and download_mode == GenerateMode.REUSE_DATASET_IF_EXISTS:
-                logger.warning("Reusing dataset %s (%s)", self.name, self._cache_dir)
+                logger.info("Reusing dataset %s (%s)", self.name, self._cache_dir)
                 # We need to update the info in case some splits were added in the meantime
                 # for example when calling load_dataset from multiple workers.
                 self.info = self._load_info()
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -507,7 +507,7 @@ class DatasetBuilder:
         with FileLock(lock_path):
             data_exists = os.path.exists(self._cache_dir)
             if data_exists and download_mode == GenerateMode.REUSE_DATASET_IF_EXISTS:
-                logger.info("Reusing dataset %s (%s)", self.name, self._cache_dir)
+                logger.warning("Reusing dataset %s (%s)", self.name, self._cache_dir)
                 # We need to update the info in case some splits were added in the meantime
                 # for example when calling load_dataset from multiple workers.
                 self.info = self._load_info()
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -1320,7 +1320,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin):
         self._format_kwargs = format_kwargs
         self._format_columns = columns
         self._output_all_columns = output_all_columns
-        logger.info(
+        logger.debug(
             "Set __getitem__(key) output type to %s for %s columns "
             " (when key is int or slice) and %s output other (un-formatted) columns.",
             "python objects" if type is None else type,
--- a/src/datasets/arrow_writer.py
+++ b/src/datasets/arrow_writer.py
@@ -423,7 +423,7 @@ class ArrowWriter:
         self.pa_writer.close()
         if close_stream:
             self.stream.close()
-        logger.info(
+        logger.debug(
             "Done writing %s %s in %s bytes %s.",
             self._num_examples,
             self.unit,
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -171,7 +171,7 @@ def get_imports(file_path: str):
     with open(file_path, mode="r", encoding="utf-8") as f:
         lines.extend(f.readlines())
 
-    logger.info("Checking %s for additional imports.", file_path)
+    logger.debug("Checking %s for additional imports.", file_path)
     imports: List[Tuple[str, str, str, Optional[str]]] = []
     is_in_docstring = False
     for line in lines:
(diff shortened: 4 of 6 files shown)


src/datasets/builder.py
```diff
@@ -507,7 +507,7 @@ def download_and_prepare(
         with FileLock(lock_path):
             data_exists = os.path.exists(self._cache_dir)
             if data_exists and download_mode == GenerateMode.REUSE_DATASET_IF_EXISTS:
-                logger.warning("Reusing dataset %s (%s)", self.name, self._cache_dir)
+                logger.info("Reusing dataset %s (%s)", self.name, self._cache_dir)
```
lhoestq, line 510: I think this should stay a warning.
When something is reloaded from the cache we should show it to the users or it could lead to very bad experiences. For example when users are doing changes and expect things to get recomputed instead of reloaded.

Especially since caching is not something that intuitive for some users: what causes recomputing vs what allows things to be reloaded is not something you can guess easily without reading the documentation.

lhoestq (commented):
Thanks for making things less verbose :)

I just have one comment about the "Reusing dataset" log, that should stay a warning IMO:

src/datasets/builder.py
stas00: tl;dr: given that the default level is WARNING and this is important then yes, I totally agree.

-------------

the longer view.

I totally hear that this is very important.

I think the key here is to discuss what is expected in the normal usage, when the person fires off the same code again and again vs develop/tune up/troubleshooting.

So to me a warning is something that is bordering-line to errors, and I only want to see such notifications when something could be going wrong with the results.

If we agree than an average user wants to enable INFO (as we are doing now for when transformers examples are used) then it's already included. For example, to stress the significance the wording could be changed to:

```
logger.info("Important: Reusing dataset %s (%s)", self.name, self._cache_dir)
```

If most of the time users run `datasets` with verbosity `WARNING` then indeed it will be missed and then you're absolutely correct, it should remain under `logger.warning`.

In other words they way I think an efficient development cycle is:

- normally run with the highest log-level and keep things silent
- when something doesn't work as expected or I'm filing a bug report - run with INFO or DEBUG enabled and inspect/share the logs.

So in this case I say to myself: `datasets` warnings are really `info`s so I'll switch to ERROR to remove the noise.

Frankly, `transformers` has many warnings which are in the same boat - not really important most of the time. So I've switched to `--log_level error` to compensate for that noise, and it's awesome! (and of course we had the env vars all along, but these are more difficult to use)

stas00 (approved):
Awesome! Thank you for this adjustment, @albertvillanova!

src/datasets/builder.py
stas00: Hmm, but it logs this on every instance, e.g. currently with 2 processes:
```
06/24/2021 19:01:54 - WARNING - datasets.arrow_dataset -   Loading cached processed dataset at /home/stas/.cache/huggingface/datasets/wmt16/ro-en/1.0.0/0d9fb3e814712c785176ad8cdb9f465fbe6479000ee6546725db30ad8a8b5f8a/cache-1bb178ab5058b59a.arrow
06/24/2021 19:01:54 - WARNING - datasets.arrow_dataset -   Loading cached processed dataset at /home/stas/.cache/huggingface/datasets/wmt16/ro-en/1.0.0/0d9fb3e814712c785176ad8cdb9f465fbe6479000ee6546725db30ad8a8b5f8a/cache-1bb178ab5058b59a.arrow
```
So if I have 64 processes this will be a lot of identical reports.

But I guess `datasets` can't be aware of ranks and distributed environment :(

OK, the user fixes this issue with `--log_on_each_node 0 --log_level_replica error` which is already part of the HF Trainer and soon in all examples. And voila now there is only one warning no matter how many nodes and gpus there are.

All is good!

albertvillanova: I've reverted this change.

PR #2565, opened 2021-06-29
Inject templates for ASR datasets

lewtun:
This PR adds ASR templates for 5 of the most common speech datasets on the Hub, where "common" is defined by the number of models trained on them.

I also fixed a bunch of the tags in the READMEs 😎


lewtun:
Wait until #2567 is merged so we can benefit from the tagger :)

datasets/common_voice/README.md
```diff
@@ -61,9 +61,9 @@ languages:
 - uk
 - vi
 - vot
-- zh-CN
-- zh-HK
-- zh-TW
+- zh-Hans-CN
+- zh-Hans-HK
+- zh-Hant-TW
```
lhoestq, line 66: We should keep  zh-CN, zh-HK and zh-TW since they are valid codes according to [BCP-47](https://tools.ietf.org/search/bcp47)

The `Hans` subtag means that the text uses the Simplified Chinese script

Therefore I'd suggest to add these three to our list of supported languages

datasets/common_voice/README.md
```diff
@@ -43,11 +44,10 @@ languages:
 - mt
 - nl
 - or
-- pa-IN
```
lhoestq, line 46: we could add this language to our list of supported languages

datasets/common_voice/README.md
```diff
@@ -43,11 +44,10 @@ languages:
...
+- pa
 - pl
 - pt
-- rm-sursilv
-- rm-vallader
```
lhoestq, line 50: these are valid BCP-47 codes according to wikipedia, we can add them to our list of languages

lhoestq (commented):
Awesome thank you !
Great work for the tags as well

A few comments about some language codes that we are missing in our languages list:

datasets/common_voice/README.md
lewtun: sounds good, will add them to `datasets`!

lewtun:
thanks for the feedback @lhoestq! i've added the new language codes and this PR should be ready for a merge :)

lhoestq (approved):
Thanks ^^

PR #2567, opened 2021-06-29
Add ASR task and new languages to resources

lewtun:
This PR adds a new `automatic-speech-recognition` task to the list of supported tasks in `tasks.json` and also includes a few new languages missing from `common_voice`.

Note: I used the [Papers with Code list](https://www.paperswithcode.com/area/speech/speech-recognition) as inspiration for the ASR subtasks

### DIFF
--- a/src/datasets/utils/resources/tasks.json
+++ b/src/datasets/utils/resources/tasks.json
@@ -77,6 +77,19 @@
             "other"
         ]
     },
+    "automatic-speech-recognition": {
+        "description": "recognising speech within audio and converting it to text",
+        "options": [
+            "speech-recognition",
+            "visual-speech-recognition",
+            "distant-speech-recognition",
+            "robust-speech-recognition",
+            "seq2seq-speech-recognition",
+            "accented-speech-recognition",
+            "noisy-speech-recognition",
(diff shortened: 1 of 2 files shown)


src/datasets/utils/resources/tasks.json
```diff
@@ -77,10 +77,23 @@
...
         ]
     },
+    "automatic-speech-recognition": {
+        "description": "recognising speech within audio and converting it to text",
+        "options": [
+            "speech-recognition",
+            "visual-speech-recognition",
+            "distant-speech-recognition",
+            "robust-speech-recognition",
+            "seq2seq-speech-recognition",
```
lhoestq, line 87: `seq2seq-speech-recognition` is just regular `speech-recognition` with seq2seq models ?
Not sure this should be a separate task, since the difference is about an aspect of the model. I think we can remove it

lhoestq (approved):
Thanks :)
I didn't know about all those subtasks

src/datasets/utils/resources/tasks.json
lewtun: yeah good idea - i'll remove it!

PR #2576, opened 2021-07-01
Add mC4

lhoestq:
AllenAI is now hosting the processed C4 and mC4 dataset in this repo: https://huggingface.co/datasets/allenai/c4
Thanks a lot to them !

In this PR I added the mC4 dataset builder. It supports 108 languages

You can load it with
```python

en_mc4 = load_dataset("mc4", "en")
fr_mc4 = load_dataset("mc4", "fr")
en_and_fr_mc4 = load_dataset("mc4", languages=["en", "fr"])
```

It also supports streaming, if you don't want to download hundreds of GB of data:
```python
en_mc4 = load_dataset("mc4", "en", streaming=True)
```

Regarding the dataset_infos.json, I will add them once I have them.

Also we can work on the dataset card at that will be at https://huggingface.co/datasets/mc4
For now I just added a link to https://huggingface.co/datasets/allenai/c4 as well as a few sections


datasets/mc4/README.md
```diff
@@ -0,0 +1,521 @@
...
+
+| config   | train   | validation   |
+|:---------|:--------|:-------------|
+| af       | ?       | ?            |
+| am       | ?       | ?            |
+| ar       | ?       | ?            |
+| az       | ?       | ?            |
+| be       | ?       | ?            |
+| bg       | ?       | ?            |
+| bg-Latn  | ?       | ?            |
```
dirkgr, line 349: I don't know how much you want to keep this generic, but it might be good to explain somewhere what the *-Latn variants are (They are the same language, but using roman characters.).

datasets/mc4/README.md
```diff
@@ -0,0 +1,521 @@
...
+
+- `url`: url of the source as a string
+- `text`: text content as a string
+- `timestamp`: timestamp as a string
+
+### Data Splits
+
+| config   | train   | validation   |
+|:---------|:--------|:-------------|
+| af       | ?       | ?            |
```
dirkgr, line 343: These languages were extracted by CLD3. As @adarob explained elsewhere, CLD3 is not the best language detector there is, but it is the best compromise allowing for scale and reproducibility outside of Google. I'm mentioning this here because we immediately got questions about that, so you could head off some confusion by stating the same thing here.

PR #2581, opened 2021-07-02
Faster search_batch for ElasticsearchIndex due to threading

mwrzalik:
Hey, 
I think it makes sense to perform search_batch threaded, so ES can perform search in parallel.
Cheers!


lhoestq (approved):
This is really cool thanks !
Looks good to me :)

PR #2587, opened 2021-07-05
Add aiohttp to tests extras require

albertvillanova:
Currently, none of the streaming tests are runned within our CI test suite, because the streaming tests require aiohttp and this is missing from our tests extras require dependencies.

Our CI test suite should be exhaustive and test all the library functionalities.


lhoestq (approved):
Good catch ! thanks

PR #2597, opened 2021-07-06
Remove redundant prepare_module

albertvillanova:
I have noticed that after implementing `load_dataset_builder` (#2500), there is a redundant call to `prepare_module`.


lhoestq (approved):
LGTM :)

PR #2602, opened 2021-07-07
Remove import of transformers

albertvillanova:
When pickling a tokenizer within multiprocessing, check that is instance of transformers PreTrainedTokenizerBase without importing transformers.

Related to huggingface/transformers#12549 and #502.


lhoestq (approved):
Thanks !

PR #2609, opened 2021-07-08
Fix potential DuplicatedKeysError

albertvillanova:
Fix potential DiplicatedKeysError by ensuring keys are unique.

We should promote as a good practice, that the keys should be programmatically generated as unique, instead of read from data (which might be not unique).


albertvillanova:
Finally, I'm splitting this PR.

PR #2612, opened 2021-07-09
Return Python float instead of numpy.float64 in sklearn metrics

lewtun:
This PR converts the return type of all `sklearn` metrics to be Python `float` instead of `numpy.float64`.

The reason behind this is that our Hub evaluation framework relies on converting benchmark-specific metrics to YAML ([example](https://huggingface.co/datasets/autonlp/autonlp-benchmark-raft-neelalex__raft-test-neelalex__raft-predictions-3/blob/main/README.md#L11)) and the `numpy.float64` format produces garbage like:

```python
import yaml

metric = load_metric("accuracy")
score = metric.compute(predictions=[0,1], references=[0,1])
print(yaml.dump(score["accuracy"])) # output below
# !!python/object/apply:numpy.core.multiarray.scalar
# - !!python/object/apply:numpy.dtype
#   args:
#   - f8
#   - false
#   - true
#   state: !!python/tuple
#   - 3
#   - <
#   - null
#   - null
#   - null
#   - -1
#   - -1
#   - 0
# - !!binary |
#   AAAAAAAA8D8=
```

### DIFF
--- a/metrics/accuracy/accuracy.py
+++ b/metrics/accuracy/accuracy.py
@@ -83,5 +83,7 @@ class Accuracy(datasets.Metric):
 
     def _compute(self, predictions, references, normalize=True, sample_weight=None):
         return {
-            "accuracy": accuracy_score(references, predictions, normalize=normalize, sample_weight=sample_weight),
+            "accuracy": accuracy_score(
+                references, predictions, normalize=normalize, sample_weight=sample_weight
+            ).tolist(),
         }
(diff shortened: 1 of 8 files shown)


lewtun:
I opened an issue on the `sklearn` repo to understand why `numpy.float64` is the default: https://github.com/scikit-learn/scikit-learn/discussions/20490

lhoestq (approved):
Thanks ! :)

lhoestq:
It could be surprising at first to use `tolist()` on numpy scalars but it works ^^

lhoestq:
did the same for Pearsonr here: https://github.com/huggingface/datasets/pull/2614

PR #2614, opened 2021-07-09
Convert numpy scalar to python float in Pearsonr output

lhoestq:
Following of https://github.com/huggingface/datasets/pull/2612

### DIFF
--- a/metrics/pearsonr/pearsonr.py
+++ b/metrics/pearsonr/pearsonr.py
@@ -94,5 +94,5 @@ class Pearsonr(datasets.Metric):
 
     def _compute(self, predictions, references):
         return {
-            "pearsonr": pearsonr(references, predictions)[0].tolist(),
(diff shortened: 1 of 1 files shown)


metrics/pearsonr/pearsonr.py
```diff
@@ -94,5 +94,5 @@ def _info(self):
 
     def _compute(self, predictions, references):
         return {
-            "pearsonr": pearsonr(references, predictions)[0],
+            "pearsonr": pearsonr(references, predictions)[0].tolist(),
```
lewtun, line 97: i just discovered in #2613 that `numpy.ndarray.item` might be a closer fit to what we're trying to achieve and also less confusing to future developers of the library :)

lhoestq: I switched to using .item() here as well, thanks !

PR #2616, opened 2021-07-09
Support remote data files

albertvillanova:
Add support for (streaming) remote data files:

```python
data_files = f"https://huggingface.co/datasets/{repo_id}/resolve/main/{relative_file_path}"
ds = load_dataset("json", split="train", data_files=data_files, streaming=True)
```

cc: @thomwolf

### DIFF
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -99,8 +99,8 @@ class BuilderConfig:
         """
         The config id is used to build the cache directory.
         By default it is equal to the config name.
-        However the name of a config is not sufficent to have a unique identifier for the dataset being generated since
-        it doesn't take into account:
+        However the name of a config is not sufficient to have a unique identifier for the dataset being generated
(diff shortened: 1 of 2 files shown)


albertvillanova:
@lhoestq maybe we could also use (if available) the ETag of the remote file in `create_config_id`?

lhoestq:
> @lhoestq maybe we could also use (if available) the ETag of the remote file in `create_config_id`?

Sure ! We can get the ETag with
```python
headers = get_authentication_headers_for_url(url, use_auth_token=use_auth_token)  # auth for private repos
etag = http_head(url, headers=headers).headers.get("ETag")
```

Since the computation of the `config_id` is done in the `DatasetBuilder.__init__`, then this means that we need to add a new parameter `use_auth_token` in `DatasetBuilder.__init__`

lhoestq (approved):

PR #2620, opened 2021-07-09
Add speech processing tasks

lewtun:
This PR replaces the `automatic-speech-recognition` task category with a broader `speech-processing` category. 

The tasks associated with this category are derived from the [SUPERB benchmark](https://arxiv.org/abs/2105.01051), and ASR is included in this set.

### DIFF
--- a/src/datasets/utils/resources/tasks.json
+++ b/src/datasets/utils/resources/tasks.json
@@ -77,16 +77,19 @@
             "other"
         ]
     },
-    "automatic-speech-recognition": {
-        "description": "recognising speech within audio and converting it to text",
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
Thanks !
This task category looks even more interesting now :)

lhoestq:
Are there any `task_categories:automatic-speech-recognition` dataset for which we should update the tags ?

lewtun:
> Are there any `task_categories:automatic-speech-recognition` dataset for which we should update the tags ?

Yes there's a few - I'll fix them tomorrow :)

PR #2621, opened 2021-07-09
Use prefix to allow exceed Windows MAX_PATH

albertvillanova:
By using this prefix, you can exceed the Windows MAX_PATH limit.

See: https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN#win32-file-namespaces

Related to #2524, #2220.

### DIFF
--- a/src/datasets/utils/filelock.py
+++ b/src/datasets/utils/filelock.py
@@ -356,7 +356,7 @@ class WindowsFileLock(BaseFileLock):
         open_mode = os.O_RDWR | os.O_CREAT | os.O_TRUNC
 
         try:
-            fd = os.open(self._lock_file, open_mode)
+            fd = os.open("\\\\?\\" + self._lock_file, open_mode)
         except FileNotFoundError:
             raise
         except OSError:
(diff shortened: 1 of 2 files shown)


mariosasko:
Does this mean the `FileNotFoundError` that avoids infinite loop can be removed?

albertvillanova:
Yes, I think so...

albertvillanova:
Or maybe we could leave it in case a relative path exceeds the MAX_PATH limit...

mariosasko:
> Or maybe we could leave it in case a relative path exceeds the MAX_PATH limit?

What about converting relative paths to absolute?

src/datasets/utils/filelock.py
```diff
@@ -377,7 +375,7 @@ def _release(self):
         os.close(fd)
 
         try:
-            os.remove(self._lock_file)
+            os.remove("\\\\?\\" + os.path.abspath(self._lock_file))
```
mariosasko, line 378: A tiny nit. `os.path.abspath(os.path.expanduser(os.path.expandvars(self._lock_file)))` is a more robust conversion.

lhoestq:
Nice ! Have you had a chance to test it on a windows machine with the max path limit enabled ? Afaik the CI doesn't have the path limit

albertvillanova:
Sure @lhoestq: I've tested on my machine... And this fixes most of the tests... 😅

PR #2628, opened 2021-07-12
Use ETag of remote data files

albertvillanova:
Use ETag of remote data files to create config ID.

Related to #2616.


lhoestq (approved):
Awesome thanks !

PR #2631, opened 2021-07-12
Delete extracted files when loading dataset

### LINKED ISSUE #2481: Delete extracted files to save disk space
As discussed with @stas00 and @lhoestq, allowing the deletion of extracted files would save a great amount of disk space to typical user.

### LINKED ISSUE #2591: Cached dataset overflowing disk space
I'm training a Swedish Wav2vec2 model on a Linux GPU and having issues that the huggingface cached dataset folder is completely filling up my disk space (I'm training on a dataset of around 500 gb).

The cache folder is 500gb (and now my disk space is full).

Is there a way to toggle caching or set the caching to be stored on a different device (I have another drive with 4 tb that could hold the caching files).

This might not technically be a bug, but I was unsure and I felt that the bug was the closest one.

  File "/home/birger/miniconda3/envs/wav2vec2/lib/python3.7/site-packages/multiprocess/pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "/home/birger/miniconda3/envs/wav2vec2/lib/python3.7/site-packages/datasets/arrow_dataset.py", line 186, in wrapper
  File "/home/birger/miniconda3/envs/wav2vec2/lib/python3.7/site-packages/datasets/fingerprint.py", line 397, in wrapper
  File "/home/birger/miniconda3/envs/wav2vec2/lib/python3.7/site-packages/datasets/arrow_dataset.py", line 1983, in _map_single
    writer.finalize()
  File "/home/birger/miniconda3/envs/wav2vec2/lib/python3.7/site-packages/datasets/arrow_writer.py", line 418, in finalize
    self.pa_writer.close()
  File "pyarrow/ipc.pxi", line 402, in pyarrow.lib._CRecordBatchWriter.close
  File "pyarrow/error.pxi", line 97, in pyarrow.lib.check_status
OSError: [Errno 28] Error writing bytes to file. Detail: [errno 28] No space left on device
"""

The above exception was the direct cause of the following exception:

### LINKED ISSUE #2604: Add option to delete temporary files (e.g. extracted files) when loading dataset
I'm loading a dataset constituted of 44 GB of compressed JSON files.

When loading the dataset with the JSON script, extracting the files create about 200 GB of uncompressed files before creating the 180GB of arrow cache tables

Having a simple way to delete the extracted files after usage (or even better, to stream extraction/delete) would be nice to avoid disk cluter.

I can maybe tackle this one in the JSON script unless you want a more general solution.

albertvillanova:
Close #2481, close #2604, close #2591.

cc: @stas00, @thomwolf, @BirgerMoell


src/datasets/utils/patching.py
```diff
@@ -0,0 +1,91 @@
...
+
+        >>> import importlib
+        >>> from datasets.load import prepare_module
+        >>> from datasets.streaming import patch_submodule, xjoin
+        >>>
+        >>> snli_module_path, _ = prepare_module("snli")
+        >>> snli_module = importlib.import_module(snli_module_path)
+        >>> patcher = patch_submodule(snli_module, "os.path.join", xjoin)
+        >>> patcher.start()
+        >>> assert snli_module.os.path.join is xjoin
```
stas00, line 35: Totally unrelated to the PR, but why the `>>>` prefix for example? It makes it impossible to copy-n-paste the code. I know pytorch uses this style and the code examples in their docs are very hard to use because of that.

The `::` already ensures that sphinx will render this section as a code segment.

stas00 (approved):
Thank you for implementing this time and space-saving strategy, @albertvillanova 

My only request is that there will be a way for a user to disable the auto-delete if they need to. For example, if something goes wrong and they need to analyze the extracted files they probably should be able to pass some flag that says not to delete the extracted files?

albertvillanova:
Sure @stas00, it is still a draft pull request. :)

stas00:
Yes, I noticed it after reviewing - my apologies.

albertvillanova:
The problem with this approach is that it also deletes the downloaded files (if they need not be extracted). 😟

src/datasets/utils/patching.py
albertvillanova: No longer applicable to this PR once that I have followed another approach different from patching. Anyway, I take on this in another PR.

albertvillanova: I know the `>>>` prefix is normally used, so that the examples are tested with [`doctest`](https://docs.python.org/3/library/doctest.html). However I do not think testing is intended for this examples in this docstring... So I agree it would be better to remove the prefix.

stas00: Thank you for validating that, @albertvillanova!

So it only impacts those of us who read the source code.

stas00:
> The problem with this approach is that it also deletes the downloaded files (if they need not be extracted). worried

Right! These probably should not be deleted by default, but having an option for those users who are tight on disc space?

albertvillanova:
> Right! These probably should not be deleted by default, but having an option for those users who are tight on disc space?

I propose leaving that for another PR, and leave this one handling only with "extracted" files. Is it OK for you? :)

lhoestq:
Awesome thanks !
I just have one question: what about image/audio datasets for which we store the path to the extracted file on the arrow data ?
In this case the default should be to keep the extracted files.

So for now I would just make `keep_extracted=True` by default until we have a way to separate extracted files that can be deleted and extracted files that are actual resources of the dataset.

albertvillanova:
@lhoestq, current implementation only deletes extracted "files", not extracted "directories", as it uses: `os.remove(path)`. I'm going to add a filter on files, so that this line does not throw an exception when passed a directory.

For audio datasets, the audio files are inside the extracted "directory", so they are not deleted.

lhoestq:
I'm still more in favor of having `keep_extracted=True` by default:
- When working with a dataset, you call `load_dataset` many times. By default we want to keep objects extracted to not extract them over and over again (it can take a long time). Then once you know what you're doing and you want to optimize disk space, you can do `keep_extracted=False`. Deleting the extracted files by default is a regression that can lead to slow downs for people calling `load_dataset` many times, which is common when experimenting
- This behavior doesn't sound natural as a default behavior. In the rest of the library, things are cached and not removed unless you explicitly say do (`map` caching for example). Moreover the function in the download manager is called `download_and_extract`, not `download_and_extract_and_remove_extracted_files`

Let me know what you think !

stas00:
I think the main issue is that after doing some work users typically move on to other datasets and the amount of disc space used keeps on growing. So your logic is very sounds and perhaps what's really needed is a cleansweep function that can go through **all** datasets and clean them up to the desired degree:

- delete extracted files
- delete source
- delete old unaccessed datasets in 6 months?
- more?

So a user can launch a little application, choose what they want to clean up and voila they have just freed up a huge amount of disc space. Makes me think of Ubuntu Tweak's Janitor app - very useful.

At the moment, this process of linting is very daunting and error-prone, especially due to all those dirs/files with hash names.

mariosasko:
@stas00 I've had the same idea. Instead of the full-fledged app, a simpler approach would be to add a new command to the CLI.

stas00:
oh, CLI would be perfect. I didn't mean to request a GUI-one specifically, was just using it as an example.

One could even do a crontab to delete old datasets that haven't been accesses in X months.

albertvillanova:
@lhoestq I totally agree with you. I'm addressing that change.

@stas00, @mariosasko, that could eventually be addressed in another pull request. The objective of this PR is:
- add an option to pass to `load_dataset`, so that extracted files are deleted
- do this deletion file per file, once the file has been already used to generate the cache Arrow file

lhoestq:
I also like the idea of having a CLI tool to help users clean their cache and save disk space, good idea !

lhoestq (approved):
Thanks ! LGTM :)

PR #2633, opened 2021-07-12
Update ASR tags

lewtun:
This PR updates the ASR tags of the 5 datasets added in #2565 following the change of task categories in #2620


albertvillanova (approved):
Thanks!

Issue #2573, opened 2021-07-01
Error while loading JSON Lines


albertvillanova:
As reported by @thomwolf, while loading a JSON Lines file with "json" loading script, he gets
> json.decoder.JSONDecodeError: Extra data: line 2 column 1 (char 383)

thomwolf:
This was actually a second error arising from a too small block-size in the json reader.

Finding the right block size is difficult for the layman user

PR #2638, opened 2021-07-13
Streaming for the Json loader

lhoestq:
It was not using `open` in the builder. Therefore `pyarrow.json.read_json` was downloading the full file to start yielding rows.

Moreover, it appeared that `pyarrow.json.read_json` was not really suited for streaming as it was downloading too much data and failing if `block_size` was not properly configured (related to #2573).

So I switched to using `open` which is extended to support reading from remote file progressively, and I removed the pyarrow json reader which was not practical.
Instead, I'm using the classical `json.loads` from the standard library.

### DIFF
--- a/src/datasets/packaged_modules/json/json.py
+++ b/src/datasets/packaged_modules/json/json.py
@@ -2,13 +2,14 @@
 
 import json
 from dataclasses import dataclass
-from io import BytesIO
 from typing import Optional
 
 import pyarrow as pa
-import pyarrow.json as paj
 
 import datasets
+
+
+logger = datasets.utils.logging.get_logger(__name__)
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


thomwolf (approved):
I think it's probably better in the end indeed

thomwolf:
A note is that I think we should add a few indicator of status (as mentioned by @stas00 in #2649), probably at the (1) downloading, (2) extracting and (3) reading steps. In particular when loading many very large files it's interesting to know a bit where we are in the process.

albertvillanova (commented):
Maybe, we should benchmark potential impacts on performance.

Using the `json` package from the Python Standard Library instead of `pyarrow.json` might have an impact on performance.

There are benchmarks that compare different JSON packages, with the Standard Library one among the worst performant: 
- https://github.com/ultrajson/ultrajson#benchmarks
- https://github.com/ijl/orjson#performance

I don't know the compared performance of `pyarrow.json`, but maybe we should check it.

lhoestq:
I tested locally, and the builtin `json` loader is 4x slower than `pyarrow.json`. Thanks for the comment @albertvillanova !

Therefore I switched back to using `pyarrow.json`, but only on the batch that is read. This way we don't have to deal with its `block_size`, and it only loads in memory one batch at a time.

albertvillanova (approved):
Thanks!! 🤗

PR #2639, opened 2021-07-13
Refactor patching to specific submodule

albertvillanova:
Minor reorganization of the code, so that additional patching functions (not related to streaming) might be created.

In relation with the initial approach followed in #2631.


lhoestq (approved):
Good idea thanks, this could be useful for many other things

PR #2534, opened 2021-06-22
Sync with transformers disabling NOTSET

albertvillanova:
Close #2528.

PR #2652, opened 2021-07-14
Fix logging docstring

mariosasko:
Remove "no tqdm bars" from the docstring in the logging module to align it with the changes introduced in #2534.


lhoestq (approved):
Thanks for updating the docstrings :)

PR #2656, opened 2021-07-15
Change `from_csv` default arguments

SBrandeis:
Passing `sep=None` to pandas's `read_csv` lets pandas guess the CSV file's separator

This PR allows users to use this pandas's feature by passing `sep=None` to `Dataset.from_csv`:

```python
Dataset.from_csv(
    ...,
    sep=None
)
```

### DIFF
--- a/docs/source/loading_datasets.rst
+++ b/docs/source/loading_datasets.rst
@@ -189,7 +189,7 @@ CSV files
 
 🤗 Datasets can read a dataset made of on or several CSV files.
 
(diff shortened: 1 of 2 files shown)


lhoestq:
This is not the default in pandas right ?
We try to align our CSV loader with the pandas API.

Moreover according to their documentation, the python parser is used when sep is None, which might not be the fastest one.

Maybe users could just specify `sep=None` themselves ?
In this case we should add some documentation about this

Issue #2629, opened 2021-07-12
Load datasets from the Hub without requiring a dataset script


lhoestq:
As a user I would like to be able to upload my csv/json/text/parquet/etc. files in a dataset repository on the Hugging Face Hub and be able to load this dataset with `load_dataset` without having to implement a dataset script.

Moreover I would like to be able to specify which file goes into which split using the `data_files` argument.

This feature should be compatible with private repositories and dataset streaming.

This can be implemented by checking the extension of the files in the dataset repository and then by using the right dataset builder that is already packaged in the library (csv/json/text/parquet/etc.)

julien-c:
This is so cool, let us know if we can help with anything on the hub side (@Pierrci @elishowk) 🎉

PR #2662, opened 2021-07-16
Load Dataset from the Hub (NO DATASET SCRIPT)

lhoestq:
## Load the data from any Dataset repository on the Hub

This PR adds support for loading datasets from any dataset repository on the hub, without requiring any dataset script.

As a user it's now possible to create a repo and upload some csv/json/text/parquet files, and then be able to load the data in one line. Here is an example with the `allenai/c4` repository that contains a lot of compressed json lines files:
```python

data_files = {"train": "en/c4-train.*.json.gz"}
c4 = load_dataset("allenai/c4", data_files=data_files, split="train", streaming=True)

print(c4.n_shards)
# 1024
print(next(iter(c4)))
# {'text': 'Beginners BBQ Class Takin...'}
```

By default it loads all the files, but as shown in the example you can choose the ones you want with unix style patterns.

Of course it's still possible to use dataset scripts since they offer the most flexibility.

## Implementation details

It uses `huggingface_hub` to list the files in a dataset repository.

If you provide a path to a local directory instead of a repository name, it works the same way but it uses `glob`.

Depending on the data files available, or passed in the `data_files` parameter, one of the available builders will be used among the csv, json, text and parquet builders.

Because of this, it's not possible to load both csv and json files at once. In this case you have to load them separately and then concatenate the two datasets for example.

## TODO

- [x] tests
- [x] docs
- [x] when huggingface_hub gets a new release, update the CI and the setup.py

Close https://github.com/huggingface/datasets/issues/2629


lhoestq:
This is ready for review now :)

I would love to have some feedback on the changes in load.py @albertvillanova. There are many changes so if you have questions let me know, especially on the `resolve_data_files` functions and on the changes in `prepare_module`.

And @thomwolf if you want to take a look at the documentation, feel free to share your suggestions :)

docs/source/loading_datasets.rst
```diff
@@ -158,22 +158,52 @@ Apart from :obj:`name` and :obj:`split`, the :func:`datasets.load_dataset` metho
...
 Generic loading scripts are provided for:
 
 - CSV files (with the :obj:`csv` script),
 - JSON files (with the :obj:`json` script),
 - text files (read as a line-by-line dataset with the :obj:`text` script),
+- parquet files (with the :obj:`parquet` script).
 - pandas pickled dataframe (with the :obj:`pandas` script).
 
-If you want to control better how your files are loaded, or if you have a file format exactly reproducing the file format for one of the datasets provided on the `HuggingFace Hub <https://huggingface.co/datasets>`__, it can be more flexible and simpler to create **your own loading script**, from scratch or by adapting one of the provided loading scripts. In this case, please go check the :doc:`add_dataset` chapter.
+If you want to control better how your files are loaded, or if you have a file format exactly reproducing the file format for one of the datasets provided on the `Hugging Face Hub <https://huggingface.co/datasets>`__, it can be more flexible and simpler to create **your own loading script**, from scratch or by adapting one of the provided loading scripts. In this case, please go check the :doc:`add_dataset` chapter.
```
lewtun, line 206:
```suggestion
If you want more fine-grained control on how your files are loaded or if you have a file format that matches the format for one of the datasets provided on the `Hugging Face Hub <https://huggingface.co/datasets>`__, it can be more  simpler to create **your own loading script**, from scratch or by adapting one of the provided loading scripts. In this case, please go check the :doc:`add_dataset` section.
```

docs/source/loading_datasets.rst
```diff
@@ -190,6 +220,8 @@ Let's see an example of all the various ways you can provide files to :func:`dat
     >>> dataset = load_dataset('csv', data_files=['my_file_1.csv', 'my_file_2.csv', 'my_file_3.csv'])
     >>> dataset = load_dataset('csv', data_files={'train': ['my_train_file_1.csv', 'my_train_file_2.csv'],
                                                   'test': 'my_test_file.csv'})
+    >>> base_url = 'https://huggingface.co/datasets/lhoestq/demo1/resolve/main/data/'
```
lewtun, line 223: really nice to see an explicit example with the expected url!

docs/source/loading_datasets.rst
```diff
@@ -465,8 +512,8 @@ For example, run the following to skip integrity verifications when loading the
 Loading datasets offline
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Each dataset builder (e.g. "squad") is a python script that is downloaded and cached either from the 🤗 Datasets GitHub repository or from the `HuggingFace Hub <https://huggingface.co/datasets>`__.
-Only the ``text``, ``csv``, ``json`` and ``pandas`` builders are included in ``datasets`` without requiring external downloads.
+Each dataset builder (e.g. "squad") is a python script that is downloaded and cached either from the 🤗 Datasets GitHub repository or from the `Hugging Face Hub <https://huggingface.co/datasets>`__.
```
lewtun, line 515: turbo nit:

```suggestion
Each dataset builder (e.g. "squad") is a Python script that is downloaded and cached either from the 🤗 Datasets GitHub repository or from the `Hugging Face Hub <https://huggingface.co/datasets>`__.
```

docs/source/loading_datasets.rst
```diff
@@ -218,6 +250,13 @@ Here is an example loading two CSV file to create a ``train`` split (default spl
     >>> from datasets import load_dataset
     >>> dataset = load_dataset('csv', data_files=['my_file_1.csv', 'my_file_2.csv'])
 
+You can also provide the URLs of remote csv files:
+
+.. code-block::
+
+    >>> from datasets import load_dataset
+    >>> dataset = load_dataset('csv', data_files="https://huggingface.co/datasets/lhoestq/demo1/resolve/main/data/train.csv")
```
lewtun, line 258: i'm not sure where we should mention this, but showing that you can download from private repos with `use_auth_token=True` would be useful

lewtun (commented):
this features looks really nice - thank you for improving our quality of life 🥳 !

i left a few nits (feel free to ignore them) and a question about whether we should show you can download private files

src/datasets/builder.py
```diff
@@ -902,8 +919,8 @@ def as_streaming_dataset(
             raise ValueError(f"Builder {self.name} is not streamable.")
         if not config.AIOHTTP_AVAILABLE:
             raise ImportError(
-                f"To be able to use dataset streaming, you need to install dependencies like aiohttp "
-                f'using "pip install \'datasets[streaming]\'" or "pip install aiohttp" for instance'
+                "To be able to use dataset streaming, you need to install dependencies like aiohttp "
+                'using \'datasets[streaming]\'" or "pip install aiohttp" for instance'
```
thomwolf, line 923: I think you can move `aiohttp` to required dep since we are now pushing for the streaming feature a lot

src/datasets/load.py
```diff
@@ -305,22 +457,52 @@ def prepare_module(
...
+    # - if os.path.join(path, name) is a local python file
+    #   -> use the module from the python file
+    # - if path is a local directory (but no python file)
+    #   -> use a packaged module (csv, text etc.) based on content of the directory
+    # - if path has no "/" and is a module on github (in /datasets or in /metrics)
+    #   -> use the module from the python file on github
+    # - if path has one "/" and is dataset repository on the HF hub with a python file
+    #   -> the module from the python file in the dataset repository
+    # - if path has one "/" and is dataset repository on the HF hub without a python file
+    #   -> use a packaged module (csv, text etc.) based on content of the repository
```
thomwolf, line 470: Maybe this should be in the docstring or doc?

src/datasets/utils/streaming_download_manager.py
```diff
@@ -15,6 +15,7 @@
 
 logger = get_logger(__name__)
 BASE_KNOWN_EXTENSIONS = ["txt", "csv", "json", "jsonl", "tsv", "conll"]
+UNSUPPORTED_ARCHIVE_EXTENSIONS_FOR_STREAMING = ["tar", "xz", "rar", "zst"]
```
thomwolf, line 18: Here also we should have the list of supported (and unsupported) extension/file-format clearly stated in the doc

thomwolf: I'm a bit surprised you do a black-list instead of a white-list here

thomwolf (approved):
This is really cool and will unlock a lot of easy workflows!

src/datasets/utils/streaming_download_manager.py
lhoestq: This is because most scripts call `download_and_extract` even for files that don't need extraction. We shouldn't raise an error for an uncompressed file. If we had a white list and you passed un uncompressed file, it wouldn't find any supported compression extension in the filename, so it would fail.

src/datasets/load.py
lhoestq: It is, but the formulation is a bit different to be clearer for users.

docs/source/loading_datasets.rst
lhoestq: I added a note at https://github.com/huggingface/datasets/blob/01f53281b267a1020bbb3b2ecaa0848f376897e8/docs/source/loading_datasets.rst#from-local-or-remote-files

lhoestq:
I took your comments into account thanks !
And I made `aiohttp` a required dependency :)

lhoestq:
Just updated the documentation :)

Let me know if you have some comments

src/datasets/utils/streaming_download_manager.py
lhoestq: in the future we'll fallback on reading the header of the file to know if it's a compressed file or not

lhoestq:
Merging this one :) 

We can try to integrate the changes in the docs to #2718 @stevhliu !

stevhliu:
Baked this into the [docs](https://44335-250213286-gh.circle-artifacts.com/0/docs/_build/html/loading.html#hugging-face-hub) already, let me know if there is anything else I should add! :)

Issue #2657, opened 2021-07-15
`to_json` reporting enhancements


stas00:
While using `to_json` 2 things came to mind that would have made the experience easier on the user:

1. could we have a `desc` arg for the tqdm use and a fallback to just `to_json` so that it'd be clear to the user what's happening? Surely, one can just print the description before calling json, but I thought perhaps it'd help to have it self-identify like we did for other bars recently.

2. it took me a while to make sense of the reported numbers:
```
 22%|██▏       | 1536/7076 [12:30:57<44:09:42, 28.70s/it]
```
So iteration here happens to be 10K samples, and the total is 70M records. But the user does't know that, so the progress bar is perfect, but the numbers it reports are meaningless until one discovers that 1it=10K samples. And one still has to convert these in the head - so it's not quick. Not exactly sure what's the best way to approach this, perhaps it can be part of `desc`? or report M or K, so it'd be built-in if it were to print, e.g.:
```
 22%|██▏       | 15360/70760K [12:30:57<44:09:42, 28.70s/it]
```
or 
```
 22%|██▏       | 15.36M/70.76M [12:30:57<44:09:42, 28.70s/it]
```
(while of course remaining friendly to small datasets)

I forget if tqdm lets you add a magnitude identifier to the running count.

Thank you!

PR #2667, opened 2021-07-17
Use tqdm from tqdm_utils

mariosasko:
This PR replaces `tqdm` from the `tqdm` lib with `tqdm` from `datasets.utils.tqdm_utils`. With this change, it's possible to disable progress bars just by calling `disable_progress_bar`. Note this doesn't work on Windows when using multiprocessing due to how global variables are shared between processes. Currently, there is no easy way to disable progress bars in a multiprocess setting on Windows (patching logging with `datasets.utils.logging.get_verbosity = lambda: datasets.utils.logging.NOTSET` doesn't seem to work as well), so adding support for this is a future goal. Additionally, this PR adds a unit ("ba" for batches) to the bar printed by `Dataset.to_json` (this change is motivated by https://github.com/huggingface/datasets/issues/2657).

### DIFF
--- a/src/datasets/__init__.py
+++ b/src/datasets/__init__.py
@@ -74,7 +74,6 @@ from .splits import (
     percent,
 )
 from .utils import *
-from .utils.tqdm_utils import disable_progress_bar
 
 
 SCRIPTS_VERSION = "master" if __version__.split(".")[-1].startswith("dev") else __version__
--- a/src/datasets/utils/py_utils.py
+++ b/src/datasets/utils/py_utils.py
@@ -195,7 +195,7 @@ def map_nested(
     if not isinstance(data_struct, dict) and not isinstance(data_struct, types):
         return function(data_struct)
 
-    disable_tqdm = bool(logging.get_verbosity() == logging.NOTSET)
+    disable_tqdm = bool(logger.getEffectiveLevel() > logging.INFO)
     iterable = list(data_struct.values()) if isinstance(data_struct, dict) else data_struct
 
     if num_proc is None:
--- a/src/datasets/utils/py_utils.py
+++ b/src/datasets/utils/py_utils.py
@@ -33,9 +33,10 @@ from typing import Callable, ClassVar, Generic, Optional, Tuple, Union
 
 import dill
 import numpy as np
-from tqdm import tqdm
+from tqdm.auto import tqdm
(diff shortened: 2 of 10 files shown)


src/datasets/utils/py_utils.py
```diff
@@ -194,14 +195,15 @@ def map_nested(
     if not isinstance(data_struct, dict) and not isinstance(data_struct, types):
         return function(data_struct)
 
-    disable_tqdm = bool(logger.getEffectiveLevel() > INFO)
+    disable_tqdm = bool(logger.getEffectiveLevel() > logging.INFO)
```
mariosasko, line 198: Replacing this line with `disable_tqdm = bool(logging.get_verbosity() == logging.NOTSET)` (this check was introduced in https://github.com/huggingface/datasets/pull/2534) causes 2 tests to fail on Windows, which is very strange.

mariosasko:
The current CI failure is due to modifications in the dataset script.

src/datasets/utils/py_utils.py
lhoestq: Good job discovering what caused this issue !

src/datasets/__init__.py
```diff
@@ -74,7 +74,6 @@
     percent,
 )
 from .utils import *
-from .utils.tqdm_utils import disable_progress_bar
```
lhoestq, line 77: I think we have to keep this no ?

lhoestq (commented):
Nice thanks :)

I just have one comment about the change in `src/datasets/__init__.py`

src/datasets/__init__.py
mariosasko: It's not needed anymore because I've added the following import to `datasets.utils.__init__.py`: 
https://github.com/huggingface/datasets/blob/df35e94c11a10ff7ac86bff3e8273fd6665a4c49/src/datasets/utils/__init__.py#L38
which is picked up by `from .utils import *` in `datasets.__init__.py`.

lhoestq: Indeed, great !

lhoestq (approved):
Thanks a lot :)

lhoestq:
Merging since the CI is only failing because of dataset card issues, which is unrelated to this PR

PR #2668, opened 2021-07-17
Add Russian SuperGLUE

slowwavesleep:
Hi,

This adds the [Russian SuperGLUE](https://russiansuperglue.com/) dataset. For the most part I reused the code for the original SuperGLUE, although there are some relatively minor differences in the structure that I accounted for.


datasets/russian_super_glue/README.md
```diff
@@ -0,0 +1,370 @@
...
+|Russian Reading Comprehension with Commonsense Reasoning (RuCoS)|Reading Comprehension with Commonsense Reasoning (ReCoRD)|
+
+### Languages
+
+All tasks are in Russian.
+
+## Dataset Structure
+
+### Data Instances
+
```
albertvillanova, line 103: Could you please add as Data Instances, a dataset example for each of the tasks?

For example, for task LiDiRus, add the example in https://russiansuperglue.com/tasks/task_info/LiDiRus#Example
```python
{
     'sentence1': "Кошка сидела на коврике.",
     'sentence2': "Кошка не сидела на коврике.",
     'label': 'not_entailment',
     'knowledge': '',
     'lexical-semantics': '',
     'logic': 'Negation',
     'predicate-argument-structure': ''
}
```

albertvillanova (changes requested):
Hi @slowwavesleep, thanks a lot for adding this dataset! You did a great job.

The code is excellent. I just leave some comments about the dataset card (we are trying to improve them in order to facilitate dataset discoverability and usage).

datasets/russian_super_glue/README.md
slowwavesleep: I've added separate examples for train/dev and test, because the differences aren't always obvious. Also, I decided to sacrifice authenticity for sake of readability and wrapped the examples with exceedingly long text fragments with line breaks, although I'm still on the fence about this. On another note, the examples are specifically after the transformations, so the demonstrated format isn't completely identical to what's actually downloaded (as is the case with the original SuperGLUE). This is the least confusing way, in my opinion, since that's the format the end user is (presumably) going to use, after all.

albertvillanova (approved):
Thank you! It is awesome!

datasets/russian_super_glue/README.md
```diff
@@ -0,0 +1,673 @@
...
+  - [Other Known Limitations](#other-known-limitations)
+- [Additional Information](#additional-information)
+  - [Dataset Curators](#dataset-curators)
+  - [Licensing Information](#licensing-information)
+  - [Citation Information](#citation-information)
+  - [Contributions](#contributions)
+
+## Dataset Description
+
+- **Homepage:** https://russiansuperglue.com/
```
lhoestq, line 58: This website is unreachable for me

datasets/russian_super_glue/README.md
```diff
@@ -0,0 +1,673 @@
...
+
+#### TERRa
+- `idx`: an `int32` feature
+- `label`: a classification label, with possible values `entailment` (0), `not_entailment` (1)
+- `premise`: a `string` feature
+- `hypothesis`: a `string` feature
+
+#### RUSSE
+- `idx`: an `int32` feature
+- `label` : a classification label, with possible values `0`, `1`
```
lhoestq, line 509: Could you explain what 0 and 1 mean ?

datasets/russian_super_glue/README.md
```diff
@@ -0,0 +1,673 @@
...
+- `gold_sense2`: an `int32` feature
+- `start1`: an `int32` feature
+- `start2`: an `int32` feature
+- `end1`: an `int32` feature
+- `end2`: an `int32` feature
+
+#### RWSD
+
+- `idx`: an `int32` feature
+- `label` : a classification label, with possible values `0`, `1`
```
lhoestq, line 523: same here

datasets/russian_super_glue/README.md
```diff
@@ -0,0 +1,673 @@
...
+- `text`: a `string` feature
+- `span1_index`: an `int32` feature
+- `span2_index`: an `int32` feature
+- `span1_text`: a `string` feature
+- `span2_text`: a `string` feature
+
+
+#### DaNetQA
+- `idx`: an `int32` feature
+- `label` : a classification label, with possible values `false` (0), `true` (1)
```
lhoestq, line 533: same here

datasets/russian_super_glue/README.md
lhoestq: Thanks ! Could you also write explicitly at the beginning of the `Data Instance` section that the test sets are missing labels ?

datasets/russian_super_glue/russian_super_glue.py
```diff
@@ -0,0 +1,527 @@
...
+        RussianSuperGlueConfig(
+            name="terra",
+            description=_TERRA_DESCRIPTION,
+            features=["premise", "hypothesis"],
+            label_classes=["entailment", "not_entailment"],
+            data_url="https://russiansuperglue.com/tasks/download/TERRa",
+            citation="",
+            url="https://russiansuperglue.com/tasks/task_info/TERRa",
+        ),
+        RussianSuperGlueConfig(
```
lhoestq, line 297: This configuration should have `label_classes ` no ? I can see the `label` field in the examples in the readme

datasets/russian_super_glue/russian_super_glue.py
```diff
@@ -0,0 +1,527 @@
...
+                "end1",
+                "end2",
+                "gold_sense1",
+                "gold_sense2",
+            ],
+            data_url="https://russiansuperglue.com/tasks/download/RUSSE",
+            citation=_RUSSE_CITATION,
+            url="https://russiansuperglue.com/tasks/task_info/RUSSE",
+        ),
+        RussianSuperGlueConfig(
```
lhoestq, line 315: same here

datasets/russian_super_glue/russian_super_glue.py
```diff
@@ -0,0 +1,527 @@
...
+            name="rwsd",
+            description=_RWSD_DESCRIPTION,
+            features=["text", "span1_index", "span2_index", "span1_text", "span2_text"],
+            data_url="https://russiansuperglue.com/tasks/download/RWSD",
+            citation="",
+            url="https://russiansuperglue.com/tasks/task_info/RWSD",
+        ),
+        RussianSuperGlueConfig(
```
lhoestq, line 323: same here

lhoestq (commented):
Awesome thank you ! This is really nice :) 

I just have a few comments for the README and the python code - though both are already in really good shape

datasets/russian_super_glue/README.md
slowwavesleep: Could you check again? Might've been something temporary. Currently, I can access it and I don't believe there are any regional restrictions. I'll need more details if it still is unreachable for you.

datasets/russian_super_glue/russian_super_glue.py
slowwavesleep: Do you mean the TERRa dataset? The label classes are there.

slowwavesleep:
Added the missing label classes and their explanations (to the best of my understanding)

datasets/russian_super_glue/README.md
lhoestq: It's working now, thanks :)

datasets/russian_super_glue/russian_super_glue.py
lhoestq: What I mean is that we could name the label classes of RUSSE, RWSD and DANETQA: "same sense"/"not same sense", "cofererence"/"not coreference", "yes"/"no" via the `label_classes` parameter that you can pass to the builder config.

It's useful to add this info to know what the label represents programmatically, since the label names are used to define the ClassLabel feature type of the `label` column of your data.

lhoestq:
Thanks a lot ! Once the last comment about the label names is addressed we can merge :)

datasets/russian_super_glue/russian_super_glue.py
slowwavesleep: Are the label names supposed to be used for inference? Conceptually, I mean. Submissions on the leaderboard are expected to be in particular format (such as `idx: 1, label: "true"`), so I was thinking that it would be useful to be able reuse label names for that. Here, and in the original SuperGLUE, the datasets with binary answers are generally labeled with boolean values. However, now I noticed that these values are used inconsistently in the sample submission. It appears as "true"/"false", "True"/"False", and 0/1.

Anyway, my question is are label names meant to be just supplementary info?

Meanwhile, I'll have to check that the code written so far works as intended with inconsistent label names.

slowwavesleep: So the choice is to use whatever label names are expected in the submission or make the names meaningful. I'm not so sure that it's an obvious one in this case.

lhoestq: I didn't know about the expected names in the submission. We should probably keep the names from what's expected in the submission then.

lhoestq: In doubt maybe we can just stay consistent with the `super_glue` dataset script, i.e. don't have `label_classes` for those tasks. Does that sound good to you ?

If yes, I guess we're done since it doesn't have `label_classes` for those tasks already.

lhoestq (approved):
Let's keep the label as is then for consistency with super_glue

Thanks a lot for adding this one !

PR #2680, opened 2021-07-20
feat: 🎸 add paperswithcode id for qasper dataset

severo:
The reverse reference exists on paperswithcode:
https://paperswithcode.com/dataset/qasper


lhoestq (approved):
Thanks :)

PR #2694, opened 2021-07-21
fix: 🐛 change string format to allow copy/paste to work in bash

severo:
Before: copy/paste resulted in an error because the square bracket
characters `[]` are special characters in bash


lhoestq (approved):
Thanks :)

PR #2696, opened 2021-07-21
Add support for disable_progress_bar on Windows

mariosasko:
This PR is a continuation of #2667 and adds support for `utils.disable_progress_bar()` on Windows when using multiprocessing. This [answer](https://stackoverflow.com/a/6596695/14095927) on SO explains it nicely why the current approach (with calling `utils.is_progress_bar_enabled()` inside `Dataset._map_single`) would not work on Windows.


mariosasko:
The CI failure seems unrelated to this PR (probably has something to do with Transformers).

lhoestq (approved):
Thanks for making it work correctly in multiprocessing !

Issue #2716, opened 2021-07-26
Calling shuffle on IterableDataset will disable batching in case any functions were mapped


amankhandelia:
When using dataset in streaming mode, if one applies `shuffle` method on the dataset and `map` method for which `batched=True` than the batching operation will not happen, instead `batched` will be set to `False`

I did RCA on the dataset codebase, the problem is emerging from [this line of code](https://github.com/huggingface/datasets/blob/d25a0bf94d9f9a9aa6cabdf5b450b9c327d19729/src/datasets/iterable_dataset.py#L197) here as it is
`self.ex_iterable.shuffle_data_sources(seed), function=self.function, batch_size=self.batch_size`, as one can see it is missing batched argument, which means that the iterator fallsback to default constructor value, which in this case is `False`.
To remedy the problem we can change this line to
`self.ex_iterable.shuffle_data_sources(seed), function=self.function, batched=self.batched, batch_size=self.batch_size`

lhoestq:
Hi :) Good catch ! Feel free to open a PR if you want to contribute, this would be very welcome ;)

amankhandelia:
Have raised the PR [here](https://github.com/huggingface/datasets/pull/2717)

albertvillanova:
Fixed by #2717.

PR #2717, opened 2021-07-26
Fix shuffle on IterableDataset that disables batching in case any functions were mapped

amankhandelia:
Made a very minor change to fix the issue#2716. Added the missing argument in the constructor call.

As discussed in the bug report, the change is made to prevent the `shuffle` method call from resetting the value of `batched` attribute in `MappedExamplesIterable`


lhoestq (approved):
Thanks for the fix !

albertvillanova:
Fix #2716.

PR #2721, opened 2021-07-27
Deal with the bad check in test_load.py

mariosasko:
This PR removes a check that's been added in #2684. My intention with this check was to capture an URL in the error message, but instead, it captures a substring of the previous regex match in the test function. Another option would be to replace this check with:
```python
m_paths = re.findall(r"\S*_dummy/_dummy.py\b", str(exc_info.value))  # on Linux this will match an URL as well as a local_path due to different os.sep, so take the last element (an URL always comes last in the list)
assert len(m_paths) > 0 and is_remote_url(m_paths[-1])  # is_remote_url comes from datasets.utils.file_utils
```

@lhoestq Let me know which one of these two approaches (delete or replace) do you prefer?

### DIFF
--- a/tests/test_load.py
+++ b/tests/test_load.py
@@ -20,7 +20,7 @@ from datasets.dataset_dict import DatasetDict, IterableDatasetDict
 from datasets.features import Features, Value
 from datasets.iterable_dataset import IterableDataset
 from datasets.load import prepare_module
-from datasets.utils.file_utils import DownloadConfig, is_remote_url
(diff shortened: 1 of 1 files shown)


lhoestq:
Hi ! I did a change for this test already in #2662 :

https://github.com/huggingface/datasets/blob/00686c46b7aaf6bfcd4102cec300a3c031284a5a/tests/test_load.py#L312-L316

I guess it's ok to remove this check for now :)

lhoestq (approved):

PR #2731, opened 2021-07-29
First draft of a method to auto-convert our datasets to TF datasets!

Rocketknight1:
Oh my **god** do not merge this yet, it's just a draft.

I've added a method (via a mixin) to the `arrow_dataset.Dataset` class that automatically converts our Dataset classes to TF Dataset classes ready for training. It hopefully has most of the features we want, including streaming from disk (no need to load the whole dataset in memory!), correct shuffling, variable-length batches to reduce compute, and correct support for unusual padding. It achieves that by calling the tokenizer `pad` method in the middle of a TF compute graph via a very hacky call to `tf.py_function`, which is heretical but seems to work.

A number of issues need to be resolved before it's ready to merge, though:

1) Is a MixIn the right way to do this? Do other classes besides `arrow_dataset.Dataset` need this method too?
2) Needs an argument to support constant-length batches for TPU training - this is easy to add and I'll do it soon.
3) Needs the user to supply the list of columns to drop from the arrow `Dataset`. Is there some automatic way to get the columns we want, or see which columns were added by the tokenizer?
4) Assumes the label column is always present and always called "label" - this is probably not great, but I'm not sure what the 'correct' thing to do here is.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -159,6 +159,51 @@ class DatasetInfoMixin:
         return self._info.version
 
 
+class TensorflowDatasetMixIn:
+    def __init__(self):
+        pass
+
+    def to_tf_dataset(self, tokenizer, cols_to_remove, batch_size, shuffle):
+        import tensorflow as tf
+        dataset_in = self.remove_columns(cols_to_remove)
+        tf_cols = [col for col in dataset_in.features]
+        label_index = tf_cols.index("label")
+        dtypes_out = []
+        for col in tf_cols:
+            try:
+                col_feature = dataset_in.features[col]
+                if hasattr(col_feature, 'feature'):
+                    col_feature = col_feature.feature
+                dtype_str = col_feature.dtype
+                dtypes_out.append(tf.as_dtype(dtype_str))
+            except TypeError:
+                raise TypeError(f"Couldn't convert column {col}, dtype {dtype_str} to TF Tensor!")
+
+        def indices_to_samples(indices):
+            batch = dataset_in.select(list(indices), keep_in_memory=True).to_dict()
+            batch = tokenizer.pad(batch)
+            output = []
+            for col in tf_cols:
+                output.append(batch[col])
+            return output
+
+        def graph_indices_to_samples(indices):
+            return tf.py_function(indices_to_samples, [indices], Tout=dtypes_out)
+
+        def reform_dict(*batch_list):
+            return ({col: batch_list[i] for i, col in enumerate(tf_cols)}, batch_list[label_index])
+
+        indices = tf.range(len(dataset_in))
+        tf_dataset = tf.data.Dataset.from_tensor_slices(indices)
+        if shuffle:
+            tf_dataset = tf_dataset.shuffle(buffer_size=len(tf_dataset))
+        tf_dataset = tf_dataset.batch(batch_size)
+        tf_dataset = tf_dataset.map(graph_indices_to_samples).map(reform_dict)
+        return tf_dataset
+
+
+
+
 class DatasetTransformationNotAllowedError(Exception):
     pass
 
@@ -232,7 +277,7 @@ def _check_table(table) -> Table:
         raise TypeError(f"Expected a pyarrow.Table or a datasets.table.Table object, but got {table}.")
 
 
-class Dataset(DatasetInfoMixin, IndexableMixin):
+class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixIn):
     """A Dataset backed by an Arrow table."""
 
     def __init__(


src/datasets/arrow_dataset.py
```diff
@@ -159,6 +159,51 @@ def version(self):
...
+
+        def reform_dict(*batch_list):
+            return ({col: batch_list[i] for i, col in enumerate(tf_cols)}, batch_list[label_index])
+
+        indices = tf.range(len(dataset_in))
+        tf_dataset = tf.data.Dataset.from_tensor_slices(indices)
+        if shuffle:
+            tf_dataset = tf_dataset.shuffle(buffer_size=len(tf_dataset))
+        tf_dataset = tf_dataset.batch(batch_size)
+        tf_dataset = tf_dataset.map(graph_indices_to_samples).map(reform_dict)
```
sgugger, line 201: This looks super slick! Did you have a chance to try it on multi-GPU and TPU to check it all works okay?

src/datasets/arrow_dataset.py
```diff
@@ -159,6 +159,51 @@ def version(self):
...
+                if hasattr(col_feature, 'feature'):
+                    col_feature = col_feature.feature
+                dtype_str = col_feature.dtype
+                dtypes_out.append(tf.as_dtype(dtype_str))
+            except TypeError:
+                raise TypeError(f"Couldn't convert column {col}, dtype {dtype_str} to TF Tensor!")
+
+        def indices_to_samples(indices):
+            batch = dataset_in.select(list(indices), keep_in_memory=True).to_dict()
+            batch = tokenizer.pad(batch)
```
sgugger, line 184: Here I think a `data_collator` would be better in terms of generalization.

src/datasets/arrow_dataset.py
```diff
@@ -159,6 +159,51 @@ def version(self):
         return self._info.version
 
 
+class TensorflowDatasetMixIn:
+    def __init__(self):
+        pass
+
+    def to_tf_dataset(self, tokenizer, cols_to_remove, batch_size, shuffle):
```
sgugger, line 166: In terms of general design, I would like the tokenizer to be optional. In most cases this won't be necessary.
And in terms of optional argument, better than a tokenizer, we should just take an `data_collator` function that could just take the samples and form a batch from it. This way we can just use the same data collators as in PyTorch with a few tweaks to make tensorflow tensors at the end.

sgugger (commented):
This is very... PyTorchic XD 
I like the design, I just think it can be made more general by using a data_collator instead of a tokenizer. My only concern is how it will go in term of performance (since TF might not like the PyTorch-iness of it all) but since we're just grabbing tokenized texts and maybe padding, this shouldn't be too much of a problem.

For computer vision though, we should see if there is a way to make sure to use several processes to prepare the batches, or does the final `map` do that automatically? Knowing TF I doubt it but one can hope.

src/datasets/arrow_dataset.py
```diff
@@ -159,6 +159,51 @@ def version(self):
...
+        import tensorflow as tf
+        dataset_in = self.remove_columns(cols_to_remove)
+        tf_cols = [col for col in dataset_in.features]
+        label_index = tf_cols.index("label")
```
sgugger, line 170: We should probably have a `label_cols`  argument since it's not always `"label"` and there could be more than one. QA models use `"start_positions"` and `"end_positions"` for instance

src/datasets/arrow_dataset.py
Rocketknight1: Trying that out now, but any working `tf.data.Dataset` should work on all of them!

src/datasets/arrow_dataset.py
Rocketknight1: Done!

src/datasets/arrow_dataset.py
Rocketknight1: Resolved! We now take an optional collate_fn argument, and pad with zeroes if one isn't provided

src/datasets/arrow_dataset.py
Rocketknight1: Done! We support (potentially overlapping) lists of columns and label_cols now

src/datasets/arrow_dataset.py
Rocketknight1: Performance is way better than before - we can iterate through all of fashion-mnist in about 2-3 seconds on a single core, and it can prefetch data in parallel with model execution. I can go back and try to add in multiprocessing later for people with really big training setups, but performance should be good enough for almost all use cases now.

Rocketknight1:
This seems to be working reasonably well in testing, and performance is way better. `tf.py_function` has been dropped for an input generator, but I moved as much of the code as possible outside the generator to allow TF to compile it correctly. I also avoid `tf.RaggedTensor` at all costs, and do the shuffle in the dataset followed by accessing sequential chunks, instead of shuffling an index tensor. The combination of all of these gives us a more flexible data loader as well as a ~20X boost in performance compared to the first solution.

src/datasets/arrow_dataset.py
```diff
@@ -159,6 +159,131 @@ def version(self):
...
+        if label_cols is not None:
+            cols_to_retain = list(set(columns + label_cols))
+        else:
+            cols_to_retain = columns
+        for col in cols_to_retain:
+            if col not in self.features:
+                raise ValueError(f"Couldn't find column {col} in dataset!")
+        dataset = self.remove_columns([col for col in self.features if col not in cols_to_retain])
+        gen_signature = self._get_output_signature(dataset, batch_size)
+        num_batches = len(dataset) // batch_size  # Because we drop the remainder
```
sgugger, line 232: You should had a flag to enable/disable this: you never want to drop the remainder for evaluation, only training.

src/datasets/arrow_dataset.py
```diff
@@ -159,6 +159,131 @@ def version(self):
...
+            for i in range(0, len(epoch_dataset) - batch_size + 1, batch_size):
+                batch = epoch_dataset[i : i + batch_size]
+                if collate_fn is not None:
+                    batch = collate_fn(batch)
+                    batch = {key: np.array(val) for key, val in batch.items()}
+                yield batch
+
+        tf_dataset = tf.data.Dataset.from_generator(tf_generator, output_signature=gen_signature)
+
+        if pad_to > 0:
```
sgugger, line 254: If we include padding here, it would be nice to have the option to pad to max length of the batch, not a fixed number. And it also should deal with tensors that are too long and truncate them? Also why always pad on the last dimension and not another one?

This is all way too specific in my opinion, so I stand with just leaving a generic collate function, that can default to something that convert to tensors (and only work for things that are already of the same size), leaving any other use case up to the user.

sgugger (commented):
Looking good! Just a few more comments on the API.

src/datasets/arrow_dataset.py
Rocketknight1: One thing to note here is that when we fetch numpy/torch tensors from our datasets, they automatically pad with zeroes on the right up to the maximum size in the batch. So maybe the easiest thing to do is to just rely on that behaviour when the user doesn't supply a collate_fn, and leave out any pad arguments?

src/datasets/arrow_dataset.py
Rocketknight1: Will do!

src/datasets/arrow_dataset.py
sgugger: Oh? I have never seen that automatic padding in PyTorch, I get errors but maybe it's a recent thing.

Rocketknight1: Maybe it's just Numpy, but I swear I remember testing with torch outputs too!

src/datasets/arrow_dataset.py
Rocketknight1: Done!

src/datasets/arrow_dataset.py
Rocketknight1: Either way, padding is now removed outside of the user-specified `collate_fn`

src/datasets/arrow_dataset.py
```diff
@@ -159,6 +160,168 @@ def version(self):
...
+                # about the dimensions we're unsure of
+                shape = [batch_size] + [None for dim in tensor.shape.as_list()[1:]]
+            signatures[column] = tf.TensorSpec(shape=shape, dtype=tensor.dtype)
+        return signatures
+
+    def to_tf_dataset(
+        self,
+        columns,
+        batch_size,
+        shuffle,
```
lhoestq, line 221: Could they be optional parameters ?

lhoestq (commented):
I'm fine with this change to use tf tensors instead of ragged tensors, and it's nice to see that there is a fallback on ragged tensors anyway. I'm very impressed by the speed gains

This is indeed a breaking change, but I agree with you that in the end it's the only way to have a proper speed. It's also always better to get actual tensors rather than ragged tensors when possible.

The API looks fine to me :)

Maybe in the future people will be happy to have more control over the shuffling (setting the parameters to pass to `Dataset.shuffle`), but for now I think it's fine

src/datasets/arrow_dataset.py
Rocketknight1: I was thinking about that! It's unclear to me what the defaults should be for `columns` or `batch_size` though, and I really wanted `shuffle` to be a required parameter to ensure people were aware of it, and that they didn't accidentally shuffle or skip shuffling their data when they didn't mean to.

I could maybe set `batch_size` to something like 32 by default and leave the other two as required parameters?

lhoestq: Oh I see your point about shuffle.
And actually thinking more about it, it looks like we should require `batch_size` as well no ?

Maybe if columns is not specified then all of them are used ?

Rocketknight1: I was thinking about that, but usually our datasets have one or more string columns that we don't want, so the default of using all columns will probably not work most of the time. It'd be nice if we had some way to auto-detect relevant columns, but I can't think of how we'd do that, so I think the safest thing to do is to just ask them to specify.

Rocketknight1:
Hi @lhoestq, the tests have been modified and everything is passing. The Windows tests look to be failing for an unrelated reason, but other than that I'm ready to merge if you are!

lhoestq:
Hi @Rocketknight1 ! Feel free to merge `master` into this branch to fix and run the full CI :)

Rocketknight1:
@lhoestq rebased onto master and it looks good! I'm doing some testing with new notebook examples, but are you happy to merge if that looks good?

mariosasko (commented):
This feature seems super cool.

A few nits in terms of style:

src/datasets/arrow_dataset.py
```diff
@@ -159,6 +159,186 @@ def version(self):
...
+        drop_remainder=None,
+        collate_fn=None,
+        collate_fn_args=None,
+        label_cols=None,
+        prefetch=True,
+    ):
+        import tensorflow as tf
+
+        if collate_fn_args is None:
+            collate_fn_args = dict()
```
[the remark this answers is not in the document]
Rocketknight1: Done!

src/datasets/arrow_dataset.py
```diff
@@ -159,6 +159,186 @@ def version(self):
...
 
+class TensorflowDatasetMixIn:
+    def __init__(self):
+        pass
+
+    @staticmethod
+    def _get_output_signature(dataset, test_batch, batch_size):
+        import tensorflow as tf
+
+        signatures = dict()
```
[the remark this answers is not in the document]
Rocketknight1: Done!

src/datasets/arrow_dataset.py
```diff
@@ -159,6 +159,186 @@ def version(self):
...
+        columns,
+        batch_size,
+        shuffle,
+        drop_remainder=None,
+        collate_fn=None,
+        collate_fn_args=None,
+        label_cols=None,
+        prefetch=True,
+    ):
+        import tensorflow as tf
```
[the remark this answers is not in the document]
Rocketknight1: Added here and in the `_get_output_signature` method.

src/datasets/arrow_dataset.py
```diff
@@ -159,6 +159,198 @@ def version(self):
...
+                    array = np.array(batch[key])
+                    cast_dtype = np.int64 if np.issubdtype(array.dtype, np.integer) else np.float32
+                    array = array.astype(cast_dtype)
+                    out_batch.append(array)
+            else:
+                for key in cols_to_retain:
+                    array = batch[key]
+                    array = numpy_pad(array)
+                    cast_dtype = np.int64 if np.issubdtype(array.dtype, np.integer) else np.float32
+                    array = array.astype(cast_dtype)
```
lhoestq, line 304: Would this work for string types or nested types ?

lhoestq (commented):
Thanks for pushing this :)
Feel free to add docstrings + type hints + tests.
Let me know if I can help you with this

Also what do you think of adding it to the documentation as well ?

src/datasets/arrow_dataset.py
Rocketknight1: I've had some success with nested dtypes (in multiple choice datasets). This does fail on string types though - the `tf.data.Dataset` is intended to be passed straight to a model, so the assumption was that everything coming out of it would be convertable to a tf.Tensor. We could possibly make strings work in this context, though - but I'd need to think about a more generic approach to building the dataset and doing shape inference.

src/datasets/arrow_dataset.py
```diff
@@ -159,6 +159,198 @@ def version(self):
...
+        if "labels" in cols_to_retain and "labels" not in self.features and "label" in self.features:
+            cols_to_retain[cols_to_retain.index("labels")] = "label"
+        for col in cols_to_retain:
+            if col not in self.features:
+                raise ValueError(f"Couldn't find column {col} in dataset!")
+        if drop_remainder is None:
+            # We assume that if you're shuffling it's the train set, so we drop the remainder unless told not to
+            drop_remainder = shuffle
+        dataset = self.remove_columns([col for col in self.features if col not in cols_to_retain])
+        dataset.set_format("python")
```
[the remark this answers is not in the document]
Rocketknight1: Noted! I'll try to make it work in numpy format.

src/datasets/arrow_dataset.py
lhoestq: Ok ! Maybe we can mention this in the docstring ?

src/datasets/arrow_dataset.py
```diff
@@ -159,6 +159,252 @@ def version(self):
...
+        label_cols: Union[str, List[str]] = None,
+        dummy_labels: bool = True,
+        prefetch: bool = True,
+    ):
+        """Create a tf.data.Dataset from the underlying Dataset. This tf.data.Dataset will load and collate batches from
+        the Dataset, and is suitable for passing to methods like model.fit() or model.predict().
+
+        Args:
+            columns (:obj:`List[str]` or :obj:`str`): Dataset column(s) to load in the tf.data.Dataset. In general,
+            only columns that the model can use as input should be included here.
```
lhoestq, line 247:
```suggestion
            only columns that the model can use as input should be included here (numeric data only)
```

src/datasets/arrow_dataset.py
lhoestq: I just mentioned that numeric data only are expected in the docstring :)

lhoestq (commented):
Thanks a lot ! I think the PR is ready to be merged now :)

After that we may to update parts of the documentation:
- add the method to the list of documented Dataset method in `main_classes.rst`
- update the demo google colab
- update the tensorflow parts of the documentation

Are there other changes that you wanted to do before merging ?

Rocketknight1:
@lhoestq No, I'm happy to merge it as-is and add documentation afterwards!

lhoestq (approved):
Perfect then :)

Issue #2737, opened 2021-07-30
SacreBLEU update


devrimcavusoglu:
With the latest release of [sacrebleu](https://github.com/mjpost/sacrebleu), `datasets.metrics.sacrebleu` is broken, and getting error.

    AttributeError: module 'sacrebleu' has no attribute 'DEFAULT_TOKENIZER'

this happens since in new version of sacrebleu there is no `DEFAULT_TOKENIZER`, but sacrebleu.py tries to import it anyways. This can be fixed currently with fixing `sacrebleu==1.5.0`

## Steps to reproduce the bug
```python
sacrebleu= datasets.load_metric('sacrebleu')
predictions = ["It is a guide to action which ensures that the military always obeys the commands of the party"]
references = ["It is a guide to action that ensures that the military will forever heed Party commands"]
results = sacrebleu.compute(predictions=predictions, references=references)
print(results)
```

## Environment info

- `datasets` version: 1.11.0
- Platform: Windows-10-10.0.19041-SP0
- Python version: Python 3.8.0
- PyArrow version: 5.0.0

bhavitvyamalik:
Hi @devrimcavusoglu, 
I tried your code with latest version of `datasets`and `sacrebleu==1.5.1` and it's running fine after changing one small thing:
```
sacrebleu = datasets.load_metric('sacrebleu')
predictions = ["It is a guide to action which ensures that the military always obeys the commands of the party"]
references = [["It is a guide to action that ensures that the military will forever heed Party commands"]]  # double brackets here should do the work
results = sacrebleu.compute(predictions=predictions, references=references)
print(results)
output: {'score': 41.180376356915765, 'counts': [11, 8, 6, 4], 'totals': [18, 17, 16, 15], 'precisions': [61.111111111111114, 47.05882352941177, 37.5, 26.666666666666668], 'bp': 1.0, 'sys_len': 18, 'ref_len': 16}
```

devrimcavusoglu:
@bhavitvyamalik hmm. I forgot double brackets, but still didn't work. It may be an isseu with platform (using win-10 currently), or versions. What is your platform and your version info for datasets, python, and sacrebleu ?

bhavitvyamalik:
You can check here. I've reproduced your code in [Google colab](https://colab.research.google.com/drive/1X90fHRgMLKczOVgVk7NDEw_ciZFDjaCM?usp=sharing)

albertvillanova:
[...]

PR #2739, opened 2021-08-02
Pass tokenize to sacrebleu only if explicitly passed by user

albertvillanova:
Next `sacrebleu` release (v2.0.0) will remove `sacrebleu.DEFAULT_TOKENIZER`: https://github.com/mjpost/sacrebleu/pull/152/files#diff-2553a315bb1f7e68c9c1b00d56eaeb74f5205aeb3a189bc3e527b122c6078795L17-R15

This PR passes `tokenize` to `sacrebleu` only if explicitly passed by the user, otherwise it will not pass it (and `sacrebleu` will use its default, no matter where it is and how it is called).

Close: #2737.

### DIFF
--- a/metrics/sacrebleu/sacrebleu.py
+++ b/metrics/sacrebleu/sacrebleu.py
@@ -113,7 +113,7 @@ class Sacrebleu(datasets.Metric):
         smooth_value=None,
         force=False,
         lowercase=False,
-        tokenize=scb.DEFAULT_TOKENIZER,
+        tokenize=None,
         use_effective_order=False,
     ):
         references_per_prediction = len(references[0])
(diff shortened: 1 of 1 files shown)


metrics/sacrebleu/sacrebleu.py
```diff
@@ -127,8 +127,8 @@ def _compute(
             smooth_value=smooth_value,
             force=force,
             lowercase=lowercase,
-            tokenize=tokenize,
             use_effective_order=use_effective_order,
+            **(dict(tokenize=tokenize) if tokenize else {}),
```
devrimcavusoglu, line 131: we may want to improve readability ? like ` tokenize = tokenize if tokenize else None`, and then pass `tokenize=tokenize #L131`, would it be compatible ?

albertvillanova: Unfortunately, that will not work: we cannot pass `tokenize=None` to `scb.corpus_bleu()`.

On the contrary, if our `tokenize` si None, we must not pass it to `scb.corpus_bleu()` (and leave them use their default value).

Take into account that it is not the same passing `tokenize=None` as not passing `tokenize` parameter.

albertvillanova: With this PR implementation, we only pass `tokenize` parameter if it is not None; if it is None, we do not pass it.

devrimcavusoglu (approved):

PR #2766, opened 2021-08-06
fix typo (ShuffingConfig -> ShufflingConfig)

daleevans:
pretty straightforward, it should be Shuffling instead of Shuffing


albertvillanova (approved):
Hi @daleevans,

The tests were not passing but due to other issue (that now is fixed by #2770).

I'm merging into master.

Thanks for the typo fix!

PR #2769, opened 2021-08-07
Allow PyArrow from source

patrickvonplaten:
When installing pyarrow from source the version is:

```python
>>> import pyarrow; pyarrow.__version__
'2.1.0.dev612'
```

-> however this breaks the install check at init of `datasets`. This PR makes sure that everything coming after the last `'.'` is removed.

### DIFF
--- a/src/datasets/__init__.py
+++ b/src/datasets/__init__.py
@@ -24,7 +24,7 @@ import pyarrow
 from pyarrow import total_allocated_bytes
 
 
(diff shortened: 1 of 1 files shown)


albertvillanova (approved):
Thanks for the fix, @patrickvonplaten.

I'm merging it into master so that this is already hotfixed.

However, I think we should eventually use `packaging.version` for all version checks/handling. ;)

PR #2774, opened 2021-08-09
Prevent .map from using multiprocessing when loading from cache

thomasw21:
## Context

On our setup, we use different setup to train vs proprocessing datasets. Usually we are able to obtain a high number of cpus to preprocess, which allows us to use `num_proc` however we can't use as many during training phase. Currently if we use `num_proc={whatever the preprocessing value was}` we load from cache, but we get:

```
  File "lib/python3.8/site-packages/multiprocess/pool.py", line 131, in worker
    put((job, i, result))
  File "lib/python3.8/site-packages/multiprocess/queues.py", line 371, in put
    self._writer.send_bytes(obj)
  File "lib/python3.8/site-packages/multiprocess/connection.py", line 203, in send_bytes
    self._send_bytes(m[offset:offset + size])
  File "lib/python3.8/site-packages/multiprocess/connection.py", line 414, in _send_bytes
    self._send(header + buf)
  File "lib/python3.8/site-packages/multiprocess/connection.py", line 371, in _send
    n = write(self._handle, buf)
BrokenPipeError: [Errno 32] Broken pipe
```

Our current guess, is that we're spawning too many processes compared to the number of cpus available, and it's running OOM.  Also we're loading this in DDP setting which means that for each gpu, I need to spawn a high number of processes to match the preprocessing fingerprint.

Instead what we suggest:
 - Allow loading shard sequentially, sharing the same fingerprint as the multiprocessed one, in order to leverage multiprocessing when we actually generate the cache, and remove it when loading from cache.

## Current issues

~I'm having a hard time making fingerprints match. For some reason, the multiprocessing and the sequential version generate two different hash.~

**EDIT**: Turns out multiprocessing and sequential have different `transform` value for fingerprinting (check `fingerprint_transform`) when running `_map_single`:
 - sequential : `datasets.arrow_dataset.Dataset._map_single`
 - multiprocessing: `datasets.arrow_dataset._map_single`
 
 This discrepancy is caused by multiprocessing pickling the transformer function, it doesn't seem to keep the `Dataset` hierarchy. I'm still unclear on why `func.__qual_name__` isn't handled correctly in multiprocessing. But replacing `__qualname__` by `__name__` fixes the issue.

## What was done

~We try to prevent the usage of multiprocessing when loading a dataset. Instead we load all cached shards sequentially.~

I couldn't find a nice way to obtain the cached_file_name and check they all exist before deciding to use the multiprocessing flow or not. Instead I expose an optional boolean `sequential` in `map` method.

## TODO
 - [x] Check that the multiprocessed version and the sequential version output the same output
 - [x] Check that sequential can load multiprocessed
 - [x] Check that multiprocessed can load sequential
 
 ## Test

```python
from multiprocessing import Pool
import random

def process(batch, rng):
  length = len(batch["text"])
  return {**batch, "processed_text": [f"PROCESSED {rng.random()}" for _ in range(length)]}

dataset = load_dataset("stas/openwebtext-10k", split="train")
print(dataset.column_names)
print(type(dataset))

rng = random.Random(42)
dataset1 = dataset.map(process, batched=True, batch_size=50, num_proc=4, fn_kwargs={"rng": rng})

# This one should be loaded from cache
rng = random.Random(42)
dataset2 = dataset.map(process, batched=True, batch_size=50, num_proc=4, fn_kwargs={"rng": rng}, sequential=True)

# Just to check that the random generator was correct
print(dataset1[-1]["processed_text"])
print(dataset2[-1]["processed_text"])
```
 
 ## Other solutions

I chose to load everything sequentially, but we can probably find a way to load shards in parallel using another number of workers (essentially this would be an argument not used for fingerprinting, allowing to allow `m` shards using `n` processes, which would be very useful when same dataset have to be loaded on two different setup, and we still want to leverage cache).

Also we can use a env variable similarly to `TOKENIZERS_PARALLELISM` as this seems generally setup related (though this changes slightly if we use multiprocessing).
cc @lhoestq (since I had asked you previously on `num_proc` being used for fingerprinting). Don't know if this is acceptable.


src/datasets/arrow_dataset.py
```diff
@@ -1707,46 +1709,60 @@ def format_cache_file_name(cache_file_name, rank):
...
-                transformed_shards = [r.get() for r in results]
-                logger.info("Concatenating {} shards from multiprocessing".format(num_proc))
-                result = concatenate_datasets(transformed_shards)
-                if new_fingerprint is not None:
-                    result._fingerprint = new_fingerprint
-                return result
+                transformed_shards = [self.__class__._map_single(**kwds) for kwds in kwds_per_shard]
+            else:
+                with Pool(num_proc, initargs=initargs, initializer=initializer) as pool:
+                    os.environ = prev_env
```
thomasw21, line 1756: I'm a bit unclear on why this is used, especially since `os.environ["TOKENIZERS_PARALLELISM"] = "false"` is used after we create `prev_env`.

src/datasets/fingerprint.py
```diff
@@ -381,7 +381,7 @@ def wrapper(*args, **kwargs):
                     kwargs_for_fingerprint.pop(default_varname)
 
             # compute new_fingerprint and add it to the args of not in-place transforms
-            transform = func.__module__ + "." + func.__qualname__
+            transform = func.__module__ + "." + func.__name__
```
thomasw21, line 384: This is probably the weirdest thing about this PR. Upon investigating, the fingerprint discrepancy between the mutiprocess version and the sequential one, I noticed that `__qualname__` had different values:
 - multiprocessing: `_map_single`
 - sequential: `Dataset._map_single`
 
 I tried figuring out where this behaviour was defined in `multiprocess` but I couldn't seem to find that. Instead, I'm not sure we actually need `__qualname__` as this seems to apply only to internal HF functions. Instead I suggest we replace it with `__name__` in order to fix this discrepancy, the multiprocessing version remains unchanged.
 
 Better if you know where I can try to fix this issue in `multiprocess`, I'd be more than happy to revert.

src/datasets/arrow_dataset.py
```diff
@@ -1577,6 +1577,7 @@ def map(
         disable_nullable: bool = False,
         fn_kwargs: Optional[dict] = None,
         num_proc: Optional[int] = None,
+        sequential: Optional[bool] = None,
```
thomasw21, line 1580: Though the context of the PR was just to load sequentially, I couldn't find a good way to check that all shards were already cached. Instead I expose this value to map. Is there a way to check before the transformation that the shard are all already cached?

tests/test_arrow_dataset.py
```diff
@@ -832,6 +832,27 @@ def test_map_multiprocessing(self, in_memory):
...
+                        self.assertDictEqual(
+                            dset_test.features,
+                            Features({"filename": Value("string"), "id": Value("int64")}),
+                        )
+                        self.assertEqual(len(dset_test.cache_files), 0 if in_memory else 2)
+                        self.assertListEqual(dset_test["id"], list(range(30)))
+                        self.assertNotEqual(dset_test._fingerprint, fingerprint)
+                        assert_arrow_metadata_are_synced_with_dataset_features(dset_test)
+
+                        self.assertEqual(dset_test._fingerprint, dset_reference._fingerprint)
```
thomasw21, line 852: This is interesting as that would give us the ability to preprocess in multiprocessing fashion and then load in sequential fashion.

src/datasets/arrow_dataset.py
```diff
@@ -1707,46 +1710,60 @@ def format_cache_file_name(cache_file_name, rank):
...
-                ]
-                logger.info("Spawning {} processes".format(num_proc))
-                results = [pool.apply_async(self.__class__._map_single, kwds=kwds) for kwds in kwds_per_shard]
-                transformed_shards = [r.get() for r in results]
-                logger.info("Concatenating {} shards from multiprocessing".format(num_proc))
-                result = concatenate_datasets(transformed_shards)
-                if new_fingerprint is not None:
-                    result._fingerprint = new_fingerprint
-                return result
+                transformed_shards = [self.__class__._map_single(**dill.copy(kwds)) for kwds in kwds_per_shard]
```
thomasw21, line 1754: `diil.copy`: This is needed in order to mimick multiprocessing pickling. The issue is if we pass arguments using a reference, then modification get propagated through the sequence. Typically, using a random generator.

thomasw21:
I'm guessing tests are failling, because this was pushed before https://github.com/huggingface/datasets/pull/2779 was merged? cc @albertvillanova

albertvillanova:
Hi @thomasw21, yes you are right: those failing tests were fixed with #2779.

Would you mind to merge current upstream master branch and push again?
```
git checkout sequential_map_when_cached
git fetch upstream
git merge upstream/master
git push -u origin sequential_map_when_cached
```

lhoestq:
Thanks for working on this ! I'm sure we can figure something out ;)

Currently `map` starts a process to apply the map function on each shard. If the shard has already been processed, then the process that has been spawned loads the processed shard from the cache and returns it.

I think we should be able to simply not start a process if a shard is already processed and cached.
This way:
- you won't need to specify `sequential=True`
- it won't create new processes if the dataset is already processed and cached
- it will properly reload each processed shard that is cached

To know if we have to start a new process for a shard you can use the function `update_fingerprint` from fingerprint.py to know the expected fingerprint of the processed shard.
Then, if the shard has already been processed, there will be a cache file named `cached-<new_fingerprint>.arrow` and you can load it with
```
Dataset.from_file(path_to_cache_file, info=self.info, split=self.split)
```

Let me know if that makes sense !

thomasw21:
Yes that makes total sense, I tried to initially do that, except the way fingerprint is handled doesn't allow to easily manipulate such a field. Typically the fingerprinting is handled in `@fingerprint_transform` which has a bunch of params that aren't quite easy to extract. Those params are used to manipulate args, kwargs in fancy ways in order to finally obtain a dictionary used for fingerprint. I could duplicate everything, but this look like a very risky thing to do. I'll take a look if I can make something work with `inspect` if I can make a very simple wrapper.

A much more simpler solution I think is adding an optional `shard: Optional[int] = None` parameter. If None, we use the number of proc as the number of shards, otherwise we pass down the expected number of shards and use either sequential/multiprocessing (with arbitrary number of workers) to load the shards?

tests/test_arrow_dataset.py
```diff
@@ -967,6 +968,19 @@ def test_map_caching(self, in_memory):
...
+            with self._caplog.at_level(WARNING):
+                with self._create_dummy_dataset(in_memory, tmp_dir) as dset:
+                    with dset.map(lambda x: {"foo": "bar"}, num_proc=2) as dset_test1:
+                        dset_test1_data_files = list(dset_test1.cache_files)
+                    with dset.map(lambda x: {"foo": "bar"}, num_proc=2) as dset_test2:
+                        self.assertEqual(dset_test1_data_files, dset_test2.cache_files)
+                        self.assertTrue(
+                            (len(re.findall("Loading cached processed dataset", self._caplog.text)) == 2) ^ in_memory
+                        )
+                        # TODO: check that dset_test2 didn't use multiprocess
```
thomasw21, line 982: I don't really know how to test this part.

thomasw21:
@lhoestq friendly ping as I feel it's up for review.

src/datasets/fingerprint.py
lhoestq: I'm not able to reproduce this:
```python
from multiprocess import Pool

def func_name():
    import datasets
    print(datasets.Dataset._map_single.__name__)

def func_qualname():
    import datasets
    print(datasets.Dataset._map_single.__qualname__)

func_name()
with Pool(2) as p:
    p.apply_async(func_name).get()
# _map_single
# _map_single

func_qualname()
with Pool(2) as p:
    p.apply_async(func_qualname).get()
# Dataset._map_single
# Dataset._map_single
```
(tested on colab and on my macbook pro)

Therefore I would leave it as before, since `__name__` could create collisions in the names in the future.

Could you provide a reproducible example ?

tests/test_arrow_dataset.py
lhoestq: Maybe mock `multiprocess.Pool` and check that it hasn't been called ?

lhoestq (commented):
Amazing thanks a lot !

This is a very clean solution :)

Before we merge I just have one comment about the `__qualname__` vs `__name__` change, and a minor suggestion for the tests:

src/datasets/fingerprint.py
thomasw21:
```python
from functools import wraps

def print_func_name(func):
  @wraps(func)
  def wrapper(*args, **kwargs):
    print(func.__qualname__, func.__name__)
  return wrapper

class Foo:
  @print_func_name
  def bar():
    pass

func_qualname()
with Pool(2) as p:
    p.apply_async(Foo.bar).get()
```

thomasw21: I was able to find a way to bypass this issue, but it really feels like a hack ...

```python
from functools import wraps
from multiprocess import Pool
def print_func_name(func):
  temp = func.__qualname__
  def wrapper(*args, **kwargs):
    print(func.__qualname__, func.__name__, temp)
  return wrapper

class Foo:
  @print_func_name
  def bar():
    pass

Foo.bar()
with Pool(2) as p:
    p.apply_async(Foo.bar).get()
    ```

lhoestq: Nice ! I added a test and a comment to explain this, so I think it's fine :)

lhoestq (approved):
Thanks for pushing this feature :)
It looks all good to me !

lhoestq:
The CI error is unrelated to the changes of this PR - it looks like an SSL issue with conda

PR #2777, opened 2021-08-10
Use packaging to handle versions

albertvillanova:
Use packaging module to handle/validate/check versions of Python packages.

Related to #2769.

### DIFF
--- a/src/datasets/__init__.py
+++ b/src/datasets/__init__.py
@@ -21,10 +21,11 @@
 __version__ = "1.11.1.dev0"
 
 import pyarrow
+from packaging import version as _version
 from pyarrow import total_allocated_bytes
 
 
-if tuple(int(i) for i in pyarrow.__version__.split(".")[:3]) < (1, 0, 0):
+if _version.parse(pyarrow.__version__).major < 1:
     raise ImportWarning(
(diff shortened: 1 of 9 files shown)


src/datasets/__init__.py
```diff
@@ -21,10 +21,11 @@
 __version__ = "1.11.1.dev0"
 
 import pyarrow
+from packaging import version as _version
 from pyarrow import total_allocated_bytes
 
 
-if tuple(int(i) for i in pyarrow.__version__.split(".")[:3]) < (1, 0, 0):
+if _version.parse(pyarrow.__version__).major < 1:
```
albertvillanova, line 28: I wonder why we make this validation, as `pyarrow` least version is already pinned in setup.py `install_requires`.

lhoestq: Because in certain environments (google colab, kaggle kernels), it may create some issues.

For example on colab, pyarrow is already installed. When colab only had pyarrow 0.0.6, installing `datasets` would install pyarrow >= 1.0.0 BUT it would also require the kernel to be restarted for the update to take effect.
Because of that you could end up with an old version of pyarrow until your restart the kernel

lhoestq (approved):
Awesome ! This is much cleaner :)

PR #2783, opened 2021-08-10
Add KS task to SUPERB

anton-l:
Add the KS (keyword spotting) ask as described in the [SUPERB paper](https://arxiv.org/abs/2105.01051).

- [s3prl instructions](https://github.com/s3prl/s3prl/blob/master/s3prl/downstream/README.md#ks-keyword-spotting)
- [s3prl implementation](https://github.com/s3prl/s3prl/blob/master/s3prl/downstream/speech_commands/dataset.py)
- [TFDS implementation](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/audio/speech_commands.py)

Some notable quirks:
- The dataset is originally single-archive (train+val+test all in one), but the test set has a "canonical" distribution in a separate archive, which is also used here (see `_split_ks_files()`). 
- The `_background_noise_`/`_silence_` audio files are much longer than others, so they require some sort of slicing for downstream training. I decided to leave the implementation of that up to the users, since TFDS and s3prl take different approaches (either slicing wavs deterministically, or subsampling randomly at runtime)

Related to #2619.


lewtun:
thanks a lot for implementing this @anton-l !!

i won't have time to review this while i'm away, so happy for @albertvillanova and @patrickvonplaten to decide when to merge :)

patrickvonplaten (approved):
Looks very clean to me!

datasets/superb/superb.py
```diff
@@ -148,6 +148,27 @@ class Superb(datasets.GeneratorBasedBuilder):
...
+            Keyword Spotting (KS) detects preregistered keywords by classifying utterances into a predefined set of
+            words. The task is usually performed on-device for the fast response time. Thus, accuracy, model size, and
+            inference time are all crucial. SUPERB uses the widely used [Speech Commands dataset v1.0] for the task.
+            The dataset consists of ten classes of keywords, a class for silence, and an unknown class to include the
+            false positive. The evaluation metric is accuracy (ACC)"""
+            ),
+            features=datasets.Features(
+                {
+                    "file": datasets.Value("string"),
+                    "label": datasets.Value("string"),
```
albertvillanova, line 164: For the "label" field, we should better use `datasets.ClassLabel` (similarly as tfds does).

datasets/superb/superb.py
```diff
@@ -383,3 +439,26 @@ def _get_speakers(rec, data, args):
         }
         for segment in data.segments[rec]
     ]
+
+
+def _split_ks_files(archive_path, split):
+    audio_path = os.path.join(archive_path, "**/*.wav")
+    audio_paths = sorted(glob.glob(audio_path))
```
albertvillanova, line 446: NIT: this `sorted` is redundant as we already call `sorted` in the return from this function: line 291 `for audio_file in sorted(splits[split]):`

datasets/superb/superb.py
```diff
@@ -244,6 +284,22 @@ def _generate_examples(self, archive_path, split=None):
                             "text": transcript,
                         }
                         key += 1
+        elif self.config.name == "ks":
+            words = ["yes", "no", "up", "down", "left", "right", "on", "off", "stop", "go"]
+            splits = _split_ks_files(archive_path, split)
+            key = 0
+            for audio_file in sorted(splits[split]):
```
albertvillanova, line 291: As we only have a for loop, maybe it would be simpler to autogenerate the `key` with `enumerate`: see below

albertvillanova, line 291:
```suggestion
            for key, audio_file in enumerate(sorted(splits[split])):
```

datasets/superb/superb.py
```diff
@@ -244,6 +284,22 @@ def _generate_examples(self, archive_path, split=None):
...
+                _, word = os.path.split(base_dir)
+                id_ = "{}_{}".format(word, file_name)
+                if word in words:
+                    label = word
+                elif word == "_silence_" or word == "_background_noise_":
+                    label = "_silence_"
+                else:
+                    label = "_unknown_"
+                yield key, {"id": id_, "file": audio_file, "label": label}
+                key += 1
```
albertvillanova, line 302:
```suggestion
```

datasets/superb/superb.py
```diff
@@ -244,6 +284,22 @@ def _generate_examples(self, archive_path, split=None):
...
                         }
                         key += 1
+        elif self.config.name == "ks":
+            words = ["yes", "no", "up", "down", "left", "right", "on", "off", "stop", "go"]
+            splits = _split_ks_files(archive_path, split)
+            key = 0
+            for audio_file in sorted(splits[split]):
+                base_dir, file_name = os.path.split(audio_file)
+                _, word = os.path.split(base_dir)
+                id_ = "{}_{}".format(word, file_name)
```
albertvillanova, line 294: Maybe remove `id_`

datasets/superb/superb.py
```diff
@@ -244,6 +284,22 @@ def _generate_examples(self, archive_path, split=None):
...
+                if word in words:
+                    label = word
+                elif word == "_silence_" or word == "_background_noise_":
+                    label = "_silence_"
+                else:
+                    label = "_unknown_"
+                yield key, {"id": id_, "file": audio_file, "label": label}
```
albertvillanova, line 301: Maybe remove "id" key-value pair.

datasets/superb/superb.py
```diff
@@ -383,3 +439,26 @@ def _get_speakers(rec, data, args):
...
+        val_paths = [os.path.join(archive_path, p) for p in val_paths]
+    with open(test_list_file, encoding="utf-8") as f:
+        test_paths = f.read().strip().splitlines()
+        test_paths = [os.path.join(archive_path, p) for p in test_paths]
+
+    # the paths for the train set is just whichever paths that do not exist in
+    # either the test or validation splits
+    train_paths = list(set(audio_paths) - set(val_paths) - set(test_paths))
+
+    return {"train": train_paths, "val": val_paths, "test": test_paths}
```
albertvillanova, line 464: Maybe better removing the "test" key-value pair, as it is never used.

albertvillanova (changes requested):
Hi @anton-l, thanks a lot for the addition of this SUPERB task. You did an awesome job! ^^

Just some comments and suggested changes before we merge it into master.

datasets/superb/README.md
```diff
@@ -183,8 +184,15 @@ An example from each split looks like:
...
 #### ks
 
-[More Information Needed](https://github.com/huggingface/datasets/blob/master/CONTRIBUTING.md#how-to-contribute-to-the-dataset-cards)
+An example from each split looks like:
 
+```python
+{
+  'file': '/path/yes/af7a8296_nohash_1.wav', 
+  'label': 'yes', 
+  'id': 'yes_af7a8296_nohash_1.wav'
```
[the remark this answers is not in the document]
anton-l: Yes, since it's a synthetic field anyway, I removed it

anton-l:
@albertvillanova thanks! Everything should be ready now :)

datasets/superb/README.md
```diff
@@ -252,8 +259,9 @@ An example from each split looks like:
 
 #### ks
 
-[More Information Needed](https://github.com/huggingface/datasets/blob/master/CONTRIBUTING.md#how-to-contribute-to-the-dataset-cards)
-
+- `id` (`string`): ID of the record.
```
albertvillanova, line 262: I think you just missed this one! 😅

albertvillanova (changes requested):
Just one minor missing "id" and that's all! :)

albertvillanova (approved):
Thank you!!

anton-l:
@albertvillanova yeah, I'm not sure how to best implement it in pure `datasets` yet. It's something like this, where `sample_noise()` needs to be called from a pytorch batch collator or other framework-specific variant:

```python
def map_to_array(example):
    import soundfile as sf

    speech_array, sample_rate = sf.read(example["file"])
    example["speech"] = speech_array
    example["sample_rate"] = sample_rate
    return example


def sample_noise(example):
    # Use a version of this function in a stateless way to extract random 1 sec slices of background noise
    # on each epoch
    from random import randint

    # _silence_ audios are longer than 1 sec
    if example["label"] == "_silence_":
        random_offset = randint(0, len(example["speech"]) - example["sample_rate"] - 1)
        example["speech"] = example["speech"][random_offset : random_offset + example["sample_rate"]]

    return example
```

albertvillanova:
I see... Yes, not trivial indeed. Maybe for the moment you could add those functions above to the README (as it is the case for now in diarization)? What do you think?

PR #2786, opened 2021-08-11
Support streaming compressed files

albertvillanova:
Add support to stream compressed files (current options in fsspec):
- bz2
- lz4
- xz
- zstd

cc: @lewtun


src/datasets/utils/streaming_download_manager.py
```diff
@@ -74,8 +88,13 @@ def xopen(file, mode="r", *args, **kwargs):
...
         kwargs["headers"] = get_authentication_headers_for_url(file, use_auth_token=kwargs.pop("use_auth_token", None))
-    file_obj = fsspec.open(file, mode=mode, *args, **kwargs).open()
-    _add_retries_to_file_obj_read_method(file_obj)
+    compression = fsspec.core.get_compression(file, "infer")
+    if not compression or compression in ["gzip", "zip"]:
+        file_obj = fsspec.open(file, mode=mode, *args, **kwargs).open()
+        file_obj = _add_retries_to_file_obj_read_method(file_obj)
+    else:
+        file_obj = fsspec.open(file, mode=mode, compression=compression, *args, **kwargs)
+        file_obj = _add_retries_to_fsspec_open_file(file_obj)
```
lhoestq, line 97: `xopen` is an extension of `open` to make it work with remote files.

Here you change its behavior for compressed files: you automatically uncompress them. Therefore if you try to open a compressed file and then use `gzip` (or any other compressing tool) to uncompress it, it won't work since it's already uncompressed.

I think we should revert this change, and explicitly use some tool in the dataset scripts to uncompress the files as we do in standard python. Otherwise we may end up with code that works in streaming mode but not in standard mode and vice-versa.

Let me know what you think @albertvillanova

tests/test_load.py
```diff
@@ -247,6 +247,24 @@ def test_load_dataset_streaming_gz_json(jsonl_gz_path):
     assert ds_item == {"col_1": "0", "col_2": 0, "col_3": 0.0}
 
 
+@require_streaming
+@pytest.mark.parametrize(
+    "path", ["sample.jsonl", "sample.jsonl.gz", "sample.tar", "sample.jsonl.xz", "sample.zip", "sample.jsonl.zst"]
+)
+def test_load_dataset_streaming_compressed_files(path):
+    repo_id = "albertvillanova/datasets-tests-compression"
+    data_files = f"https://huggingface.co/datasets/{repo_id}/resolve/main/{path}"
```
lhoestq, line 256: This is a really nice feature !

I think the glob logic has to be moved in a data files resolution module, as done in https://github.com/huggingface/datasets/pull/2662

The current implementation may not be robust enough to work with path manipulations by users in compressed files

src/datasets/utils/streaming_download_manager.py
albertvillanova: No, `fsspec.open` (even if passed the `compression` parameter) does not uncompress the file immediately: it returns an `OpenFile` instance, which will return a file-object wrapped with a decompressor instance (when called within a context manager), which will decompress on the fly... ;)

tests/test_load.py
albertvillanova: I have not touched the glob logic in this PR though... 🤔

src/datasets/utils/streaming_download_manager.py
albertvillanova: And yes, at the end, the result (after having called `dl_manager.download_and_extract` will be an uncompressed file, either streaming or not. That is the objective! 😉

albertvillanova: The issue is: how do you make that `StreamingDownloadManager.extract()` passes the parameter `compression=compression` to `fsspec.open(urlpath, compression=compression)` if they can communicate only through the parameter `urlpath`?

Because of this, I always pass `compression="infer"`, which assumes that all datasets scripts have called `.extract` (or `.download_and_extract`) before calling `fsspec.open`. This assumption is sensible and will work for all dataset scripts, except for oscar (as you told me yesterday), because you changed oscar with a call: `gzip.open(open())`.

PR #2798, opened 2021-08-13
Fix streaming zip files

albertvillanova:
Currently, streaming zip files gives `FileNotFoundError` message.

This PR fixes it by adding a glob string.

The corresponding test is implemented in PR #2786.

### DIFF
--- a/src/datasets/utils/streaming_download_manager.py
+++ b/src/datasets/utils/streaming_download_manager.py
@@ -126,7 +126,7 @@ class StreamingDownloadManager(object):
             # there is one single file which is the uncompressed gzip file
             return f"{protocol}://{os.path.basename(urlpath.split('::')[0]).rstrip('.gz')}::{urlpath}"
         else:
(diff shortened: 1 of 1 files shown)


lhoestq:
Hi ! I don't fully understand this change @albertvillanova 
The `_extract` method used to return the compound URL that points to the root of the inside of the archive.
This way users can use the usual os.path.join or other functions to point to the relevant files. I don't see why you're using a glob pattern ?

albertvillanova:
This change is to allow this:
```python
data_files = f"https://huggingface.co/datasets/albertvillanova/datasets-tests-compression/resolve/main/sample.zip"
ds = load_dataset("json", split="train", data_files=data_files, streaming=True)
assert isinstance(ds, IterableDataset)
```

Before this PR it gave error because pointing to the root, without any subsequent join, gives error:
```python
fsspec.open("zip://::https://huggingface.co/datasets/albertvillanova/datasets-tests-compression/resolve/main/sample.zip")
```

PR #2807, opened 2021-08-16
Add cats_vs_dogs dataset

nateraw:
Adds Microsoft's [Cats vs. Dogs](https://www.microsoft.com/en-us/download/details.aspx?id=54765) dataset.


lhoestq (approved):
It looks like it's all good now :)
Merging !

Thanks again for adding this one :)

PR #2810, opened 2021-08-16
Add WIT Dataset

hassiahk:
Adds Google's [WIT](https://github.com/google-research-datasets/wit) dataset.


mariosasko:
Google's version of WIT is now available here: https://huggingface.co/datasets/google/wit

PR #2811, opened 2021-08-17
Fix stream oscar

albertvillanova:
Previously, an additional `open` was added to oscar to make it stream-compatible: 587bbb94e891b22863b312b99696e32708c379f4.

This was argued that might be problematic: https://github.com/huggingface/datasets/pull/2786#discussion_r690045921

This PR:
- removes that additional `open`
- patches `gzip.open` with `xopen` + `compression="gzip"`

### DIFF
--- a/datasets/oscar/oscar.py
+++ b/datasets/oscar/oscar.py
@@ -355,7 +355,7 @@ class Oscar(datasets.GeneratorBasedBuilder):
         current_lines = []
         for filepath in filepaths:
             logger.info("generating examples from = %s", filepath)
-            with gzip.open(open(filepath, "rb"), "rt", encoding="utf-8") as f:
+            with gzip.open(filepath, "rt", encoding="utf-8") as f:
                 for line in f:
                     if len(line.strip()) > 0:
                         current_lines.append(line)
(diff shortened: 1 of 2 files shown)


lhoestq (commented):
Ok this fixes the issue for OSCAR.

However it doesn't fix the underlying issue which is that the current behavior of `xopen` is not the one of the one of the regular `open`, since it returns a file-like object with automatic uncompression, which shouldn't be the case.

This is confusing for users, and it makes the current OSCAR code fail.

I agree this PR does fix OSCAR, but IMO this is not how we have to fix it. Instead we should make `xopen` mimic `open` in standard conditions. Otherwise it will fail for many other datasets and confuse users that deal with compressed files.

lhoestq:
One additional note: if we can try to not change the code of oscar.py too often, I'm sure users that have it in their cache directory will be happy to not have to redownload it every time they update the library ;)

(since changing the code changes the cache directory of the dataset)

albertvillanova:
I don't think this is confusing for users because users don't even know we have patched `open`. The only thing users care is that if the pass `streaming=True`, they want to be able to load the dataset in streaming mode.

I don't see any other dataset where patching `open` with `fsspec.open`+`compression` is an "underlying issue". Are there other datasets where this is an issue?

The only dataset where this was an issue is in oscar and the issue is indeed due to the additional `open` you added inside `zip.open`.

PR #2814, opened 2021-08-18
Bump tqdm version

mariosasko:
The recently released tqdm 4.62.1 includes a fix for PermissionError on Windows (submitted by me in https://github.com/tqdm/tqdm/pull/1207), which means we can remove expensive `gc.collect` calls by bumping tqdm to that version. This PR does exactly that and, additionally, fixes a `disable_tqdm` definition that would previously, if used, raise a PermissionError on Windows.


lhoestq (approved):
Good news ! Thanks for fixing it on tqdm's side :)

Issue #2813, opened 2021-08-18
Remove compression from xopen


albertvillanova:
We implemented support for streaming with 2 requirements:
- transparent use for the end user: just needs to pass the parameter `streaming=True`
- no additional work for the contributors: previous loading scripts should also work in streaming mode with no (or minor) changes; and new loading scripts should not involve additional code to support streaming

In order to fulfill these requirements, streaming implementation patched some Python functions:
- the `open(urlpath)` function was patched with `fsspec.open(urlpath)`
- the `os.path.join(urlpath, *others)` function was patched in order to add to `urlpath` hops (`::`) and extractor protocols (`zip://`), which are required by `fsspec.open`

Recently, we implemented support for streaming all archive+compression formats: zip, tar, gz, bz2, lz4, xz, zst; tar.gz, tar.bz2,...
Under the hood, the implementation:
- passes an additional parameter `compression` to `fsspec.open`, so that it performs the decompression on the fly: `fsspec.open(urlpath, compression=...)`

Some concerns have been raised about passing the parameter `compression` to `fsspec.open`:
- https://github.com/huggingface/datasets/pull/2786#discussion_r689550254
- #2811 

The main argument is that if `open` decompresses the file and afterwards we call `gzip.open` on it, that will raise an error in `oscar` dataset:
```python
gzip.open(open(urlpath
```
While this is true:
- it is not natural/usual to call `open` inside `gzip.open` (never seen this before)
- indeed, this was recently (2 months ago) coded that way in `datasets` in order to allow streaming support (with previous implementation of streaming)

In this particular case, there is a natural fix solution: #2811:
- Revert the `open` inside the `gzip.open` (change done 2 months ago): `gzip.open(open(urlpath` => `gzip.open(urlpath`
- Patch `gzip.open(urlpath` with `fsspec.open(urlpath, compression="gzip"` 

Are there other issues apart from this?

Note that there is an issue just because the open inside of the gzip.open. There is no issue in the other cases where datasets loading scripts use just
- `gzip.open` 
- `open` (after having called dl_manager.download_and_extract)

TODO:
- [ ] Is this really an issue? Please enumerate the `datasets` loading scripts where this is problematic.
  - For the moment, there are only 3 datasets where we have an `open` inside a `gzip.open`:
    - oscar (since 23 June), mc4 (since 2 July) and c4 (since 2 July)
  - In the 3 datasets, the only reason to put an open inside a gzip.open was indeed to force supporting streaming
[...]

PR #2822, opened 2021-08-20
Add url prefix convention for many compression formats

lhoestq:
## Intro

When doing dataset streaming, the uncompression of compressed files is done on the fly using `fsspec`.

In particular, the download manager method `download_and_extract` doesn't return a path to the local download and extracted file, but instead a chained URL so that the uncompression can be done when the file is opened. A few examples of chained URLS:
- `gz://file.txt::https://foo.bar/file.txt.gz`
- `bz2://file.txt::https://foo.bar/file.txt.bz2`
- `zip://::https://foo.bar/archive.zip`
- `tar://::https://foo.bar/archive.tar.gz` (the TAR uncompression includes gz, bz2 etc. uncompression in `fsspec`)

This syntax is highly inspired by the `fsspec` URL chaining syntax from https://filesystem-spec.readthedocs.io/en/latest/features.html#url-chaining

This url prefixing allows `open` to know what kind of uncompression to do in a dataset script when doing
```python
def _generate_examples(self, urlpath):
    with open(urlpath) as f:
        ....
```

## What it changes

This changes the previous behavior from https://github.com/huggingface/datasets/pull/2786 , in which `open` was trying to infer the compression automatically. Infering the compression made it impossible to know whether the user wanted `open` to return compressed data (as the default behavior of the buitin open), or the uncompressed data. By adding uncompression prefixes to the URL, `open` know directly if it has to uncompress or not, and also which protocol to use.

## Additional notes

This PR should close https://github.com/huggingface/datasets/issues/2813

It should also close this PR https://github.com/huggingface/datasets/pull/2811 since the oscar dataset script won't try to uncompress twice anymore

Note that I had to temporarily remove the support for passing tar and zip files to `data_files` for streaming to make it work, since it makes it ambiguous whether a zip file passed as `data_files` should be uncompressed or not. IMO we can make it work again by changing the syntax to make the glob explicit:
```python
load_dataset("json", data_files="zip://*.jsonl::https://foo.bar/archive.zip")
```
This is the exact same convention as fsspec and it removes all ambiguities

cc @albertvillanova @lewtun


src/datasets/filesystems/compression.py
```diff
@@ -0,0 +1,168 @@
...
+
+class Bz2FileSystem(BaseCompressedFileFileSystem):
+    """Read contents of BZ2 file as a filesystem with one file inside."""
+
+    protocol = "bz2"
+    compression = "bz2"
+    extension = ".bz2"
+
+
+class GZipFileSystem(BaseCompressedFileFileSystem):
```
lewtun, line 83: out of curiosity, why is this class name `GZipFileSystem` instead of `GzipFileSystem` like the other compression protocol classes?

src/datasets/filesystems/compression.py
```diff
@@ -0,0 +1,168 @@
...
+    """Read contents of GZIP file as a filesystem with one file inside."""
+
+    protocol = "gzip"
+    compression = "gzip"
+    extension = ".gz"
+
+
+class Lz4FileSystem(BaseCompressedFileFileSystem):
+    """Read contents of Lz4 file as a filesystem with one file inside."""
```
lewtun, line 92: nit:
```suggestion
    """Read contents of LZ4 file as a filesystem with one file inside."""
```

lewtun (commented):
this is really elegant @lhoestq !

from the user perspective, do i need to know about the url prefix convention when using `load_dataset`? if yes, perhaps we can add some details in the documentation to explain how this works - what do you think?

src/datasets/filesystems/compression.py
```diff
@@ -0,0 +1,168 @@
...
+
+    root_marker = ""
+    protocol: str = (
+        None  # protocol passed in prefix to the url. ex: "gzip", for gzip://file.txt::http://foo.bar/file.txt.gz
+    )
+    compression: str = None  # compression type in fsspec. ex: "gzip"
+    extension: str = None  # extension of the filename to strip. ex: "".gz" to get file.txt from file.txt.gz
+
+    def __init__(
+        self, fo: str = "", target_protocol: Optional[str] = None, target_options: Optional[dict] = None, **kwargs
```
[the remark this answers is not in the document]
lhoestq: `fo` is the variable name in `fsspec` to wrap any file to open. I went with the same convention.
However it's indeed worth adding some explanation in the docstring :)

lhoestq:
Thanks for the feedback :) I will also complete the documentation to explain this convention

lhoestq:
I just added come documentation about how streaming works with chained URLs.

I will also add some docs about how to use chained URLs directly in `load_dataset` in #2662, since #2662 does change the documentation already and to avoid having to resolve conflicts.

docs/source/dataset_streaming.rst
```diff
@@ -164,3 +164,107 @@ It is possible to get a ``torch.utils.data.IterableDataset`` from a :class:`data
...
+
+Now that you are aware of how dataset streaming works, you can make sure your dataset script work in streaming mode:
+
+1. make sure you use ``open`` to open the data files: it is extended to work with remote files
+2. if you have to deal with archives like ZIP files, make sure you use ``os.path.join`` to navigate in the archive
+
+Currently a few python functions or classes are not supported for dataset streaming:
+
+- ``pathlib.Path`` and all its methods are not supported, please use ``os.path.join`` and string objects
+- ``os.walk``, ``os.listdir``, ``glob.glob`` are not supportet yet
```
severo, line 270: nit: supported

docs/source/dataset_streaming.rst
```diff
@@ -164,3 +164,107 @@ It is possible to get a ``torch.utils.data.IterableDataset`` from a :class:`data
...
+
+The StreamingDownloadManager
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The standard (i.e. non-streaming) way of loading a dataset has two steps:
+
+1. download and extract the raw data files of the dataset by using the :class:`datasets.DownloadManager`
+2. process the data files to generate the Arrow file used to load the :class:`datasets.Dataset` object.
+
+For example in non-streaming mode a file is simply downloaded like this:
```
lewtun, line 180: turbo nit:

```suggestion
For example, in non-streaming mode a file is simply downloaded like this:
```

docs/source/dataset_streaming.rst
```diff
@@ -164,3 +164,107 @@ It is possible to get a ``torch.utils.data.IterableDataset`` from a :class:`data
...
+
+    >>> from datasets import DownloadManager
+    >>> url = "https://huggingface.co/datasets/lhoestq/test/resolve/main/some_text.txt"
+    >>> filepath = DownloadManager().download(url)  # the file is downloaded here
+    >>> print(filepath)
+    '/Users/user/.cache/huggingface/datasets/downloads/16b702620cad8d485bafea59b1d2ed69e796196e6f2c73f005dee935a413aa19.ab631f60c6cb31a079ecf1ad910005a7c009ef0f1e4905b69d489fb2bd162683'
+    >>> with open(filepath) as f:
+    ...     print(f.read())
+
+When you load a dataset in streaming mode, the download manager being used instead is the :class:`datasets.StreamingDownloadManager`.
```
lewtun, line 192: nit:

```suggestion
When you load a dataset in streaming mode, the download manager that is used instead is the :class:`datasets.StreamingDownloadManager`.
```

docs/source/dataset_streaming.rst
```diff
@@ -164,3 +164,107 @@ It is possible to get a ``torch.utils.data.IterableDataset`` from a :class:`data
...
+Instead of actually downloading and extracting all the data when you load the dataset, it is done lazily.
+The file starts to be downloaded and extracted only when ``open`` is called.
+This is made possible by extending ``open`` to support opening remote files via http.
```
lewtun, line 195: nit:

```suggestion
This is made possible by extending ``open`` to support opening remote files via HTTP.
```

docs/source/dataset_streaming.rst
```diff
@@ -164,3 +164,107 @@ It is possible to get a ``torch.utils.data.IterableDataset`` from a :class:`data
...
+    >>> urlpath = StreamingDownloadManager().download(url)
+    >>> print(urlpath)
+    'https://huggingface.co/datasets/lhoestq/test/resolve/main/some_text.txt'
+    >>> with xopen(urlpath) as f:
+    ...     print(f.read())  # the file is actually downloaded here
+
+As you can see, since it's possible to open remote files via an URL, the streaming download manager just returns the URL instead of the path to the local downloaded file.
+
+Then the file is downloaded in a streaming fashion: it is downloaded progessively as you iterate over the data file.
+This is made possible because it is based on ``fsspec``, a library that allows to open and iterate on remote files. You can find more infos about ``fsspec`` in `its documentation <https://filesystem-spec.readthedocs.io/>`_
```
lewtun, line 213: nit:

```suggestion
This is made possible because it is based on ``fsspec``, a library that allows to open and iterate on remote files. You can find more information about ``fsspec`` in `its documentation <https://filesystem-spec.readthedocs.io/>`_
```

docs/source/dataset_streaming.rst
```diff
@@ -164,3 +164,107 @@ It is possible to get a ``torch.utils.data.IterableDataset`` from a :class:`data
...
+
+Compressed files and archives
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You noticed that the streaming download manager returns the exact same URL that was given as input for a text file.
```
lewtun, line 218:
```suggestion
You may have noticed that the streaming download manager returns the exact same URL that was given as input for a text file.
```

docs/source/dataset_streaming.rst
```diff
@@ -164,3 +164,107 @@ It is possible to get a ``torch.utils.data.IterableDataset`` from a :class:`data
...
+However if you use ``download_and_extract`` on a compressed file instead, then the output url will be a chained URL.
+Chained URLs are used by ``fsspec`` to navigate in remote compressed archives.
+
+Some examples of chained URL are:
+
+.. code-block::
+
+    >>> from datasets.utils.streaming_download_manager import xopen
+    >>> chained_url = "zip://combined/train.json::https://adversarialqa.github.io/data/aqa_v1.0.zip"
```
lewtun, line 227: in a separate pr / location in the docs, it would be great to see this syntax explained a little bit.

src/datasets/filesystems/compression.py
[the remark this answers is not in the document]
lewtun: ok sounds good! since we're sticking close to `fsspec` let's keep is as `fo`

lewtun (commented):
the new docs are great and very instructive - thanks for adding them! i left a few minor nits

lhoestq:
Merging this one now, next step is resolve the conflicts in #2662 and update the docs for URL chaining :)

There is also the glob feature of zip files that I need to add, to be able to do this for example:
```python
load_dataset("json", data_files="zip://*::https://foo.bar/archive.zip")
```

PR #2828, opened 2021-08-23
Add code-mixed Kannada Hope speech dataset

adeepH:
## Adding a Dataset
- **Name:** *KanHope*
- **Description:** *A code-mixed English-Kannada dataset for Hope speech detection*
- **Paper:** *https://arxiv.org/abs/2108.04616* 
- **Data:** *https://github.com/adeepH/KanHope/tree/main/dataset*
- **Motivation:** *The dataset is amongst the very few resources available for code-mixed low-resourced Dravidian languages of India*


lhoestq (commented):
Hi ! Thanks a lot for adding this dataset :)

It looks all good to me !

Before merging we just need to fix the CI. The error comes from the name "kanHope" for the directory and the python script. We are using the snake_case convention for the dataset directory/script names.

Could you rename them `kan_hope` ? It should fix the CI

Also because of the rename you will also need to regenerate the dataset_infos.json file

PR #2836, opened 2021-08-25
Optimize Dataset.filter to only compute the indices to keep

lhoestq:
Optimize `Dataset.filter` to only compute the indices of the rows to keep, instead of creating a new Arrow table with the rows to keep. Creating a new table was an issue because it could take a lot of disk space.

This will be useful to process audio datasets for example cc @patrickvonplaten


patrickvonplaten:
Maybe worth updating the docs here as well?

lhoestq:
Yup, will do !

PR #2843, opened 2021-08-27
Fix extraction protocol inference from urls with params

lhoestq:
Previously it was unable to infer the compression protocol for files at URLs like
```
https://foo.bar/train.json.gz?dl=1
```
because of the query parameters.

I fixed that, this should allow 10+ datasets to work in streaming mode:
```
  "discovery",
  "emotion",
  "grail_qa",
  "guardian_authorship",
  "pragmeval",
  "simple_questions_v2",
  "versae/adobo",
  "w-nicole/childes_data",
  "w-nicole/childes_data_no_tags_",
  "w-nicole/childes_data_with_tags",
  "w-nicole/childes_data_with_tags_"
```

cc @severo

### DIFF
--- a/src/datasets/utils/streaming_download_manager.py
+++ b/src/datasets/utils/streaming_download_manager.py
@@ -150,7 +150,7 @@ class StreamingDownloadManager(object):
 
     def _get_extraction_protocol(self, urlpath: str) -> Optional[str]:
         path = urlpath.split("::")[0]
-        extension = path.split(".")[-1]
+        extension = path.split(".")[-1].split("?")[0]  # https://foo.bar/train.json.gz?dl=1 -> gz
         if extension in BASE_KNOWN_EXTENSIONS:
             return None
         elif path.endswith(".tar.gz") or path.endswith(".tgz"):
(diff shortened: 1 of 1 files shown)


src/datasets/utils/streaming_download_manager.py
```diff
@@ -149,8 +149,12 @@ def _extract(self, urlpath: str) -> str:
             return f"{protocol}://::{urlpath}"
 
     def _get_extraction_protocol(self, urlpath: str) -> Optional[str]:
+        # get inner file: zip://train-00000.json.gz::https://foo.bar/data.zip -> zip://data/train-00000.json.gz
```
severo, line 152: I think it should be

~~~
         # get inner file: zip://train-00000.json.gz::https://foo.bar/data.zip -> zip://train-00000.json.gz
~~~

lhoestq:
merging since the windows error is just a CircleCI issue

severo:
It works, eg https://observablehq.com/@huggingface/datasets-preview-backend-client#{%22datasetId%22%3A%22discovery%22} and https://datasets-preview.huggingface.tech/rows?dataset=discovery&config=discovery&split=train

lhoestq:
Nice !

Issue #2552, opened 2021-06-28
Keys should be unique error on code_search_net


thomwolf:
## Describe the bug
Loading `code_search_net` seems not possible at the moment.

## Steps to reproduce the bug
```python
>>> load_dataset('code_search_net')
Downloading: 8.50kB [00:00, 3.09MB/s]                                                                                                                                           
Downloading: 19.1kB [00:00, 10.1MB/s]                                                                                                                                           
No config specified, defaulting to: code_search_net/all
Downloading and preparing dataset code_search_net/all (download: 4.77 GiB, generated: 5.99 GiB, post-processed: Unknown size, total: 10.76 GiB) to /Users/thomwolf/.cache/huggingface/datasets/code_search_net/all/1.0.0/b3e8278faf5d67da1d06981efbeac3b76a2900693bd2239bbca7a4a3b0d6e52a...
Traceback (most recent call last):         
  File "/Users/thomwolf/Documents/GitHub/datasets/src/datasets/builder.py", line 1067, in _prepare_split
    writer.write(example, key)
  File "/Users/thomwolf/Documents/GitHub/datasets/src/datasets/arrow_writer.py", line 343, in write
    self.check_duplicate_keys()
  File "/Users/thomwolf/Documents/GitHub/datasets/src/datasets/arrow_writer.py", line 354, in check_duplicate_keys
    raise DuplicatedKeysError(key)
datasets.keyhash.DuplicatedKeysError: FAILURE TO GENERATE DATASET !
Found duplicate Key: 48
Keys should be unique and deterministic in nature
```

## Environment info
- `datasets` version: 1.8.1.dev0
- Platform: macOS-10.15.7-x86_64-i386-64bit
- Python version: 3.8.5
- PyArrow version: 2.0.0

thomwolf:
Two questions:
- with `datasets-cli env` we don't have any information on the dataset script version used. Should we give access to this somehow? Either as a note in the Error message or as an argument with the name of the dataset to `datasets-cli env`?
- I don't really understand why the id is duplicated in the code of `code_search_net`, how can I debug this actually?

lhoestq:
Thanks for reporting. There was indeed an issue with the keys. The key was the addition of the file id and row id, which resulted in collisions. I just opened a PR to fix this at https://github.com/huggingface/datasets/pull/2555

To help users debug this kind of errors we could try to show a message like this
```python
DuplicateKeysError: both 42th and 1337th examples have the same keys `48`.
[...]

PR #2844, opened 2021-08-27
Fix: wikicorpus - fix keys

lhoestq:
As mentioned in https://github.com/huggingface/datasets/issues/2552, there is a duplicate keys error in `wikicorpus`.

I fixed that by taking into account the file index in the keys

### DIFF
--- a/datasets/wikicorpus/wikicorpus.py
+++ b/datasets/wikicorpus/wikicorpus.py
@@ -159,7 +159,7 @@ class Wikicorpus(datasets.GeneratorBasedBuilder):
                             pass
                         elif row.startswith("ENDOFARTICLE") or row.startswith("\n"):
                             if len(words) > 1:  # some content besides only (. . Fp 0)
-                                yield (file_idx, row_idx), {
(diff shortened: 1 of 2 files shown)


datasets/wikicorpus/wikicorpus.py
```diff
@@ -159,7 +159,7 @@ def _generate_examples(self, dirpath):
                             pass
                         elif row.startswith("ENDOFARTICLE") or row.startswith("\n"):
                             if len(words) > 1:  # some content besides only (. . Fp 0)
-                                yield id_, {
+                                yield (file_idx, row_idx), {
```
albertvillanova, line 162: Is there a reason to yield a tuple, differently from above, where it is yielded a string instead?
```suggestion
                                yield f"{file_idx}_{row_idx}", {
```

lhoestq: Good catch ! This needs to be a string indeed - we don't support tuples anymore since we detect duplicate keys

lhoestq:
The CI error is unrelated to this PR

... merging !

PR #2851, opened 2021-08-30
Update `column_names` showed as `:func:` in exploring.st

ClementRomac:
Hi,  

One mention of `column_names` in exploring.st was showing it as `:func:` instead of `:attr:`.


lhoestq (approved):
Good catch, thanks !

Issue #2825, opened 2021-08-23
The datasets.map function does not load cached dataset


hobbitlzy:
## Describe the bug
The datasets.map function caches the processed data to a certain directory. When the map function is called another time with totally the same parameters, the cached data are supposed to be reloaded instead of re-processing. However, it doesn't reuse cached data sometimes. I use the common data processing in different tasks, the datasets are processed again, the only difference is that I run them in different files.

## Steps to reproduce the bug
Just run the following codes in different .py files.
```python
if __name__ == '__main__':
    from datasets import load_dataset
    from transformers import AutoTokenizer
    raw_datasets = load_dataset("wikitext", "wikitext-2-raw-v1")

    tokenizer = AutoTokenizer.from_pretrained("bert-base-uncased")


    def tokenize_function(examples):
        return tokenizer(examples["text"], padding="max_length", truncation=True)


    tokenized_datasets = raw_datasets.map(tokenize_function, batched=True)
```

## Expected results
The map function should reload data in the second or any later runs.

## Actual results
The processing happens in each run.

## Environment info

- `datasets` version: 1.8.0
- Platform: linux
- Python version: 3.7.6
- PyArrow version: 3.0.0

This is the first time I report a bug. If there is any problem or confusing description, please let me know 😄.

lhoestq:
Thanks for reporting ! This is indeed a bug, I'm looking into it

lhoestq:
#2854 fixed the issue :)

We'll do a new release of `datasets` soon to make the fix available.
In the meantime, feel free to try it out by installing `datasets` from source

If you have other issues or any question, feel free to re-open the issue :)

TheTahaaa:
Hello there, 

Although I don't change any parameter in the map function, I've faced the same issue.

This is the code that I use to map and tokenise my dataset:

`
train_data = train_data.map(
            lambda samples: self.tokenizer(
                text=samples['source'], text_target=samples['target'], padding='max_length', truncation=True), batched=True, batch_size=128)
`

albertvillanova:
Hi @TheTahaaa,

Could you please open a new Bug issue with all the information about your local environment?

TheTahaaa:
> Hi @TheTahaaa,
> 
> Could you please open a new Bug issue with all the information about your local environment?

Hey there,

[...]

PR #2854, opened 2021-08-31
Fix caching when moving script

lhoestq:
When caching the result of a `map` function, the hash that is computed depends on many properties of this function, such as all the python objects it uses, its code and also the location of this code.

Using the full path of the python script for the location of the code makes the hash change if a script like `run_mlm.py` is moved.

I changed this by simply using the base name of the script instead of the full path.

Note that this change also affects the hash of the code used from imported modules, but I think it's fine. Indeed it hashes the code of the imported modules anyway, so the location of the python files of the imported modules doesn't matter when computing the hash.

Close https://github.com/huggingface/datasets/issues/2825


lhoestq:
Merging since the CI failure is unrelated to this PR

PR #2856, opened 2021-08-31
fix: 🐛 remove URL's query string only if it's ?dl=1

severo:
A lot of URL use the query strings, for example
http://opus.nlpl.eu/download.php?f=Bianet/v1/moses/en-ku.txt.zip, we
must not remove it when trying to detect the protocol. We thus remove it
only in the case of the query string being ?dl=1 which occurs on dropbox
and dl.orangedox.com. Also: add unit tests.

See https://github.com/huggingface/datasets/pull/2843 for the original
discussion.


lhoestq (approved):
Thanks for the fix :)
LGTM

merging since the CI failure isn't related to this PR and has been fixed on master

PR #2858, opened 2021-08-31
Fix s3fs version in CI

lhoestq:
The latest s3fs version has new constrains on aiobotocore, and therefore on boto3 and botocore

This PR changes the constrains to avoid the new conflicts

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -97,7 +97,7 @@ REQUIRED_PKGS = [
     "importlib_metadata;python_version<'3.8'",
     # to save datasets locally or on any filesystem
     # minimum 2021.05.0 to have the AbstractArchiveFileSystem
-    "fsspec[http]>=2021.05.0",
+    "fsspec[http]==2021.08.1",
     # for data streaming via http
     "aiohttp",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -121,14 +121,14 @@
...
-    "botocore==1.19.52",
+    "aiobotocore",
+    "boto3",
+    "botocore",
     "faiss-cpu",
     "fsspec[s3]",
     "moto[s3,server]==2.0.4",
     "rarfile>=4.0",
-    "s3fs",
+    "s3fs==2021.08.1",
```
albertvillanova, line 131: Is it necessary to pin this version? I guess this will have to be eventually updated/removed to allow future versions of `s3fs`...

lhoestq: This is useful to get the right versions of aiobotocore/boto3/botocore. If we don't set a minimum version, then pip would pick a version that is too old (s3fs 0.5.3). If we set a minimum version without pinning the exact version, the pip resolver takes forever to find the right versions for s3fs/aiobotocore/boto3/botocore (>10min).

Maybe an alternative would be to have minimum versions for all of them, but I haven't tested

Issue #2859, opened 2021-08-31
Loading allenai/c4 in streaming mode does too many HEAD requests


lhoestq:
This does 60,000+ HEAD requests to get all the ETags of all the data files:
```python
from datasets import load_dataset
load_dataset("allenai/c4", streaming=True)
```
It makes loading the dataset completely impractical.

The ETags are used to compute the config id (it must depend on the data files being used).
Instead of using the ETags, we could simply use the commit hash of the dataset repository on the hub, as well and the glob pattern used to resolve the files (here it's `*` by default, to load all the files of the repository)

severo:
https://github.com/huggingface/datasets/blob/6c766f9115d686182d76b1b937cb27e099c45d68/src/datasets/builder.py#L179-L186

PR #2861, opened 2021-09-01
fix: 🐛 be more specific when catching exceptions

severo:
The same specific exception is catched in other parts of the same
function.

### DIFF
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -31,6 +31,7 @@ from urllib.parse import urlparse
 
 import fsspec
 import huggingface_hub
+import requests
 from huggingface_hub import HfApi
 
 from . import config
@@ -544,7 +545,7 @@ def prepare_module(
                         dataset_info = hf_api.dataset_info(
                             repo_id=path, revision=script_version, token=download_config.use_auth_token
                         )
-                    except FileNotFoundError:
+                    except requests.exceptions.RequestException:
                         raise FileNotFoundError(
                             f"Couldn't find a directory or a {resource_type} named '{path}'. "
                             f"It doesn't exist locally at {expected_dir_for_combined_path_abs} or remotely on {hf_api.endpoint}/datasets"
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -544,7 +544,7 @@ def prepare_module(
                         dataset_info = hf_api.dataset_info(
                             repo_id=path, revision=script_version, token=download_config.use_auth_token
                         )
-                    except Exception:
+                    except FileNotFoundError:
                         raise FileNotFoundError(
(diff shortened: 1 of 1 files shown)


albertvillanova (commented):
The test suite is not passing because with your change, we do not longer catch `requests` exceptions:
```
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://huggingface.co/api/datasets/lhoestq/_dummy?full=true
```

albertvillanova:
To give more context: after our discussion, if I understood properly, you are trying to fix a call to `datasets` that takes 15 minutes: https://github.com/huggingface/datasets-preview-backend/issues/17 Is this right?

severo:
Yes, that's it. And to do that I'm trying to use https://pypi.org/project/stopit/, which will raise a stopit.TimeoutException exception. But currently, if this exception is raised, it's caught and considered as a "FileNotFoundError" while it should not be caught.

albertvillanova:
And what about passing the `timeout` parameter instead?

severo:
It might be a good idea, but I would have to add a timeout argument to several methods, I'm not sure we want that (I want to ensure all my queries in https://github.com/huggingface/datasets-preview-backend/tree/master/src/datasets_preview_backend/queries resolve in a given time, be it with an error in case of timeout, or with the successful response). The methods are `prepare_module`, `import_main_class`, *builder_cls.*`get_all_exported_dataset_infos`, `load_dataset_builder`, and `load_dataset`

src/datasets/load.py
```diff
@@ -544,7 +545,7 @@ def prepare_packaged_module(name):
                         dataset_info = hf_api.dataset_info(
                             repo_id=path, revision=script_version, token=download_config.use_auth_token
                         )
-                    except Exception:
+                    except requests.exceptions.RequestException:
```
albertvillanova, line 548: Now the test suite is green: all `requests` exceptions (and only them) are catched. I'm wondering if there could be another kind of exception thrown by `HfApi.dataset_info`...

severo: Reading https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/hf_api.py#L524, I don't think so. On the other end, as the huggingface_hub does not officially state that the exceptions of this method are requests.exceptions.RequestException, it might break if they touch the code.

albertvillanova:
I understand, you are trying to find a fix for your use case. OK.

Just note that it is also an issue for `datasets` users. Once #2859 fixed in `datasets`, you will no longer have this issue...

severo:
Closing, since 1. my problem is more #2859, and I was asking for that change in order to make a hack work on my side, 2. if we want to change how exceptions are handled, we surely want to do it on all the codebase, not only in this particular case.

Issue #2866, opened 2021-09-02
"counter" dataset raises an error in normal mode, but not in streaming mode


severo:
## Describe the bug

`counter` dataset raises an error on `load_dataset()`, but simply returns an empty iterator in streaming mode.

## Steps to reproduce the bug

```python
>>> import datasets as ds
>>> a = ds.load_dataset('counter', split="train", streaming=False)
Using custom data configuration default
Downloading and preparing dataset counter/default (download: 1.29 MiB, generated: 2.48 MiB, post-processed: Unknown size, total: 3.77 MiB) to /home/slesage/.cache/huggingface/datasets/counter/default/1.0.0/9f84962fa0f35bec5a34fe0bdff8681838d497008c457f7856c48654476ec0e9...
Traceback (most recent call last):
  File "/home/slesage/hf/datasets-preview-backend/.venv/lib/python3.8/site-packages/datasets/builder.py", line 726, in _download_and_prepare
    self._prepare_split(split_generator, **prepare_split_kwargs)
  File "/home/slesage/hf/datasets-preview-backend/.venv/lib/python3.8/site-packages/datasets/builder.py", line 1124, in _prepare_split
    for key, record in utils.tqdm(
  File "/home/slesage/hf/datasets-preview-backend/.venv/lib/python3.8/site-packages/tqdm/std.py", line 1185, in __iter__
    for obj in iterable:
  File "/home/slesage/.cache/huggingface/modules/datasets_modules/datasets/counter/9f84962fa0f35bec5a34fe0bdff8681838d497008c457f7856c48654476ec0e9/counter.py", line 161, in _generate_examples
    with derived_file.open(encoding="utf-8") as f:
  File "/home/slesage/.pyenv/versions/3.8.11/lib/python3.8/pathlib.py", line 1222, in open
    return io.open(self, mode, buffering, encoding, errors, newline,
  File "/home/slesage/.pyenv/versions/3.8.11/lib/python3.8/pathlib.py", line 1078, in _opener
    return self._accessor.open(self, flags, mode)
FileNotFoundError: [Errno 2] No such file or directory: '/home/slesage/.cache/huggingface/datasets/downloads/extracted/b57aa6db5601a738e57b95c1fd8cced54ff28fc540efcdaf0f6c4f1bb5dfe211/COUNTER/0032p.xml'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #2874, opened 2021-09-07
Support streaming datasets that use pathlib

albertvillanova:
This PR extends the support in streaming mode for datasets that use `pathlib.Path`.

Related to: #2866.
CC: @severo

### DIFF
--- a/src/datasets/utils/streaming_download_manager.py
+++ b/src/datasets/utils/streaming_download_manager.py
@@ -118,6 +118,10 @@ def xopen(file, mode="r", *args, **kwargs):
     return file_obj
 
 
+def xpathopen(path, **kwargs):
+    return xopen(_as_posix(path), **kwargs)
+
+
 class StreamingDownloadManager(object):
     """
     Download manager that uses the "::" separator to navigate through (possibly remote) compressed archives.
--- a/src/datasets/utils/streaming_download_manager.py
+++ b/src/datasets/utils/streaming_download_manager.py
@@ -56,6 +56,14 @@ def xjoin(a, *p):
     return "::".join([a] + b)
 
 
+def _as_posix(path):
+    return path.as_posix().replace(":/", "://")
+
+
+def xpathjoin(a, *p):
+    return type(a)(xjoin(_as_posix(a), *p))
+
+
 def _add_retries_to_file_obj_read_method(file_obj):
     read = file_obj.read
     max_retries = config.STREAMING_READ_MAX_RETRIES
--- a/src/datasets/utils/streaming_download_manager.py
+++ b/src/datasets/utils/streaming_download_manager.py
@@ -2,7 +2,7 @@ import os
 import re
 import time
 from pathlib import Path
-from typing import Optional
+from typing import Optional, Tuple
 
 import fsspec
 import posixpath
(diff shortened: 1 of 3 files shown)


src/datasets/utils/streaming_download_manager.py
```diff
@@ -56,6 +56,14 @@ def xjoin(a, *p):
     return "::".join([a] + b)
 
 
+def _as_posix(path):
+    return path.as_posix().replace(":/", "://")
```
lhoestq, line 60: I think it can cause an error in a case like this:
```
PosixPath('zip:/test.txt::/Users/username/bar.zip') -> 'zip://test.txt:://Users/username/bar.zip'
```
a simple workaround would be:
```suggestion
    return path.as_posix().replace(":/", "://").replace(":://", "::/")
```

src/datasets/utils/streaming_download_manager.py
```diff
@@ -56,6 +56,14 @@ def xjoin(a, *p):
     return "::".join([a] + b)
 
 
+def _as_posix(path):
```
lhoestq, line 59: Could you add docstrings + type hints to these functions ?

tests/test_streaming_download_manager.py
```diff
@@ -29,36 +37,33 @@
     ],
```
lhoestq, line 37: maybe we can add a test with local paths (both relative and absolute)
```suggestion
        (
            ".",
            ("file.txt",),
            "file.txt",
        ),
        (
            Path().resolve().as_posix(),
            ("file.txt",),
            (Path().resolve() / "file.txt").as_posix(),
        ),
    ],
```

tests/test_streaming_download_manager.py
```diff
@@ -29,36 +37,33 @@
...
     assert output_path == expected_path
+    output_path = xpathjoin(Path(input_path), *paths_to_join)
+    assert output_path == Path(expected_path)
 
 
 def test_xopen_local(text_path):
-
     with xopen(text_path, encoding="utf-8") as f, open(text_path, encoding="utf-8") as expected_file:
         assert list(f) == list(expected_file)
+    with xpathopen(Path(text_path), encoding="utf-8") as f, open(text_path, encoding="utf-8") as expected_file:
```
lhoestq, line 49: Since `Path` converts `://` to `:/` it would be nice to have a test with `zip://` and a local path, and also with a remote path

tests/test_streaming_download_manager.py
```diff
@@ -29,36 +37,33 @@
...
+        assert list(f) == list(expected_file)
 
 
 def test_xopen_remote():
-    from datasets.utils.streaming_download_manager import xopen
-
     with xopen(TEST_URL, encoding="utf-8") as f:
         assert list(f) == TEST_URL_CONTENT.splitlines(keepends=True)
+    with xpathopen(Path(TEST_URL), encoding="utf-8") as f:
```
lhoestq, line 56: same here

lhoestq (commented):
Thanks, this is very nice ! It should fix a lot of dataset in streaming mode :)

I added a few suggestions, mostly about the tests

src/datasets/utils/streaming_download_manager.py
lhoestq: or, this
```suggestion
    return re.sub("([A-z]:/)([A-z])", "\g<1>/\g<2>", path.as_posix())
```

albertvillanova: Good catch!

albertvillanova: I finally did:
```python
return re.sub(r"(?<!:):/", "://", path.as_posix())
```

tests/test_streaming_download_manager.py
albertvillanova: This is indirectly tested by `test_as_posix`, do you agree?

tests/test_streaming_download_manager.py
albertvillanova: Same here. This will save testing time, specially relevant when performing network requests...

tests/test_streaming_download_manager.py
lhoestq: yup thanks !

lhoestq (approved):
LGTM thanks you !

severo:
I've tried https://github.com/huggingface/datasets/issues/2866 again, and I get the same error.

albertvillanova:
@severo Issue #2866 is not fully fixed yet: multiple patches need to be implemented for `pathlib`, as that dataset uses quite a lot of `pathlib` functions... 😅

severo:
No worry and no stress, I just wanted to check for that case :) I'm very happy that you're working on issues I'm interested in!

PR #2876, opened 2021-09-07
Extend support for streaming datasets that use pathlib.Path.glob

albertvillanova:
This PR extends the support in streaming mode for datasets that use `pathlib`, by patching the method `pathlib.Path.glob`.

Related to #2874, #2866.

CC: @severo


albertvillanova:
I am thinking that ideally we should call `fs.glob()` instead...

src/datasets/utils/streaming_download_manager.py
```diff
@@ -180,6 +180,28 @@ def xpathopen(path: Path, **kwargs):
...
+        :obj:`~pathlib.Path`
+    """
+    posix_path = _as_posix(path)
+    main_hop, *rest_hops = posix_path.split("::")
+    if is_local_path(main_hop):
+        yield from Path(main_hop).glob(pattern)
+    else:
+        fs, fs_token, globbed_paths = fsspec.get_fs_token_paths(xjoin(posix_path, pattern))
+        if "*" not in pattern:
+            globbed_paths = fs.glob(globbed_paths[0])
```
lhoestq, line 200: Maybe add a comment to mention that `get_fs_token_paths` does pattern matching only if there's a `*` in the pattern ? Otherwise it could be unclear why you would need to `glob` when there's no `*` in the pattern.

```suggestion
        fs, fs_token, globbed_paths = fsspec.get_fs_token_paths(xjoin(posix_path, pattern))
        # if there's no "*" in the pattern, get_fs_token_paths() doesn't do any pattern matching
        # so to be able to glob patterns like "[0-9]", we have to call `fs.glob`
        if "*" not in pattern:
            globbed_paths = fs.glob(globbed_paths[0])
```

albertvillanova: I have finally refactor this so that it always calls `fs.glob` because:
- `get_fs_token_paths` does pattern matching only if there is a `*`
- The `*` in `get_fs_token_paths` only matches file names, not directory names
- `get_fs_token_paths` doesn't do pattern matching if there is `**`

albertvillanova:
Thanks, @lhoestq: the idea of adding the mock filesystem is to avoid network calls and reduce testing time ;) 

I have added `rglob` aswell and fixed some bugs.

PR #2880, opened 2021-09-08
Extend support for streaming datasets that use pathlib.Path stem/suffix

albertvillanova:
This PR extends the support in streaming mode for datasets that use `pathlib`, by patching the properties `pathlib.Path.stem` and `pathlib.Path.suffix`.

Related to #2876, #2874, #2866.

CC: @severo


lhoestq (approved):
Nice thanks !

Issue #887, opened 2020-11-25
pyarrow.lib.ArrowNotImplementedError: MakeBuilder: cannot construct builder for type extension<arrow.py_extension_type>


AmitMY:
I set up a new dataset, with a sequence of arrays (really, I want to have an array of (None, 137, 2), and the first dimension is dynamic) 

```python
    def _info(self):
        return datasets.DatasetInfo(
            description=_DESCRIPTION,
            # This defines the different columns of the dataset and their types
            features=datasets.Features(
                {
                    "pose": datasets.features.Sequence(datasets.features.Array2D(shape=(137, 2), dtype="float32"))
                }
            ),
            homepage=_HOMEPAGE,
            citation=_CITATION,
        )
    def _generate_examples(self):
        """ Yields examples. """

        yield 1, {
            "pose": [np.zeros(shape=(137, 2), dtype=np.float32)]
        }
```

But this doesn't work -
> pyarrow.lib.ArrowNotImplementedError: MakeBuilder: cannot construct builder for type extension<arrow.py_extension_type>

lhoestq:
Yes right now `ArrayXD` can only be used as a column feature type, not a subtype.
With the current Arrow limitations I don't think we'll be able to make it work as a subtype, however it should be possible to allow dimensions of dynamic sizes (`Array3D(shape=(None, 137, 2), dtype="float32")` for example since the [underlying arrow  type](https://github.com/huggingface/datasets/blob/master/src/datasets/features.py#L236) allows dynamic sizes.

For now I'd suggest the use of nested `Sequence` types. Once we have the dynamic sizes you can update the dataset.
What do you think ?

AmitMY:
> Yes right now ArrayXD can only be used as a column feature type, not a subtype. 

Meaning it can't be nested under `Sequence`?
If so, for now I'll just make it a python list and make it with the nested `Sequence` type you suggested.

lhoestq:
Yea unfortunately..
That's a current limitation with Arrow ExtensionTypes that can't be used in the default Arrow Array objects.
We already have an ExtensionArray that allows us to use them as column types but not for subtypes.
Maybe we can extend it, I haven't experimented with that yet

AmitMY:
Cool
So please consider this issue as a feature request for:
```
Array3D(shape=(None, 137, 2), dtype="float32")
```

its a way to represent videos, poses, and other cool sequences

AmitMY:
[...]

PR #2891, opened 2021-09-10
[WIP] Allow dynamic first dimension for ArrayXD

rpowalski:
Add support for dynamic first dimension for ArrayXD features. See issue [#887](https://github.com/huggingface/datasets/issues/887).
Following changes allow for `to_pylist` method of `ArrayExtensionArray` to return a list of numpy arrays where fist dimension can vary.

@lhoestq Could you suggest how you want to extend test suit. For now I added only very limited testing.

### DIFF
--- a/src/datasets/features.py
+++ b/src/datasets/features.py
@@ -412,9 +412,32 @@ class ArrayExtensionArray(pa.ExtensionArray):
         numpy_arr = numpy_arr.reshape(len(self), *self.type.shape)
         return numpy_arr
 
+    def to_list_of_numpy(self, zero_copy_only=True):
+        storage: pa.ListArray = self.storage
+        shape = self.type.shape
+        arrays = []
+        for dim in range(1, self.type.ndims):
+            assert shape[dim] is not None, f"Support only dynamic size on first dimension. Got: {shape}"
+
+        first_dim_offsets = np.array([off.as_py() for off in storage.offsets])
+        for i in range(len(storage)):
+            storage_el = storage[i:i+1]
+            first_dim = first_dim_offsets[i+1] - first_dim_offsets[i]
+            # flatten storage
+            for dim in range(self.type.ndims):
+                storage_el = storage_el.flatten()
+
+            numpy_arr = storage_el.to_numpy(zero_copy_only=zero_copy_only)
+            arrays.append(numpy_arr.reshape(first_dim, *shape[1:]))
+
+        return arrays
+
     def to_pylist(self):
         zero_copy_only = _is_zero_copy_only(self.storage.type)
-        return self.to_numpy(zero_copy_only=zero_copy_only).tolist()
+        if self.type.shape[0] is None:
+            return self.to_list_of_numpy(zero_copy_only=zero_copy_only)
+        else:
+            return self.to_numpy(zero_copy_only=zero_copy_only).tolist()
 
 
 class PandasArrayExtensionDtype(PandasExtensionDtype):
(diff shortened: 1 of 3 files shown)


lhoestq (commented):
Thanks for opening the PR :)

I think we can add test to make sure both `to_pylist` and `to_numpy` work as expected using an Array3D for example, with first dimension being dynamic.

Also It would be nice to add a test to make sure that `Dataset.to_pandas` works with a dataset with such data.

rpowalski:
@lhoestq, thanks for your review.

I added test for `to_pylist`, I didn't do that for `to_numpy` because this method shouldn't be called for dynamic dimension ArrayXD - this method will try to make a single numpy array for the whole column which cannot be done for dynamic arrays.

I dig into `to_pandas()` functionality and I found it quite difficult to implement. `PandasArrayExtensionArray` takes single np.array as an argument. It might be a bit of changes to make it work with the list of arrays. Do you mind if we exclude this work from this PR. I added an error message for the case if somebody tries to use dynamic arrays with `to_pandas`

rpowalski:
@lhoestq, I just fixed all the tests. Let me know if there is anything else to add.

lhoestq:
Hi ! Sorry for the delay

It looks good to me ! I think the only thing missing is the support for passing a list of numpy arrays to `map` when the first dimension is dynamic.

Currently it raises an error:
```python
from datasets import *
import numpy as np

features=  Features({"a": Array3D(shape=(None, 5, 2), dtype="int32")})
d = Dataset.from_dict({"a": [np.zeros((5,5,2)), np.zeros((2,5,2))]}, features=features)
d = d.map(lambda a: {"a": np.concatenate([a]*2)}, input_columns="a")
print(d[0])
```
raises
```python
  File "playground/ttest.py", line 6, in <module>
    d = d.map(lambda x: {"a": np.concatenate([x]*2)}, input_columns="a")
  File "/home/truent/hf/datasets/src/datasets/arrow_dataset.py", line 1932, in map
    return self._map_single(
  File "/home/truent/hf/datasets/src/datasets/arrow_dataset.py", line 426, in wrapper
  File "/home/truent/hf/datasets/src/datasets/fingerprint.py", line 406, in wrapper
  File "/home/truent/hf/datasets/src/datasets/arrow_dataset.py", line 2317, in _map_single
    writer.finalize()
  File "/home/truent/hf/datasets/src/datasets/arrow_writer.py", line 443, in finalize
    self.write_examples_on_file()
  File "/home/truent/hf/datasets/src/datasets/arrow_writer.py", line 312, in write_examples_on_file
  File "pyarrow/array.pxi", line 222, in pyarrow.lib.array
  File "pyarrow/array.pxi", line 110, in pyarrow.lib._handle_arrow_array_protocol
  File "/home/truent/hf/datasets/src/datasets/arrow_writer.py", line 108, in __arrow_array__
    storage = pa.array(self.data, type.storage_dtype)
  File "pyarrow/array.pxi", line 305, in pyarrow.lib.array
  File "pyarrow/array.pxi", line 39, in pyarrow.lib._sequence_to_array
  File "pyarrow/error.pxi", line 84, in pyarrow.lib.check_status
pyarrow.lib.ArrowInvalid: Can only convert 1-dimensional array values
```

I think the issue is that here we don't cover the case where self.data is a list of numpy arrays:

https://github.com/huggingface/datasets/blob/55fd140a63b8f03a0e72985647e498f1fc799d3f/src/datasets/arrow_writer.py#L104-L109

We should remove the `isinstance(self.data[0], np.ndarray)` part and add these lines to cover this case:

https://github.com/huggingface/datasets/blob/55fd140a63b8f03a0e72985647e498f1fc799d3f/src/datasets/arrow_writer.py#L112-L113

stevhliu:
Hi @rpowalski, thanks for adding this feature! 

I wanted to check if you are still interested in documenting this, otherwise I'd be happy to help with it :)

Issue #2816, opened 2021-08-18
Add Mostly Basic Python Problems Dataset


osanseviero:
## Adding a Dataset
- **Name:** Mostly Basic Python Problems Dataset
- **Description:** The benchmark consists of around 1,000 crowd-sourced Python programming problems, designed to be solvable by entry level programmers, covering programming fundamentals, standard library functionality, and so on. Each problem consists of a task description, code solution and 3 automated test cases.
- **Paper:** *link to the dataset paper if available*
- **Data:** https://github.com/google-research/google-research/tree/master/mbpp
- **Motivation:** Simple, small dataset related to coding problems.

Instructions to add a new dataset can be found [here](https://github.com/huggingface/datasets/blob/master/ADD_NEW_DATASET.md).

lvwerra:
I started working on that.

PR #2893, opened 2021-09-10
add mbpp dataset

lvwerra:
This PR adds the mbpp dataset introduced by Google [here](https://github.com/google-research/google-research/tree/master/mbpp) as mentioned in #2816.

The dataset contain two versions: a full and a sanitized one. They have a slightly different schema and it is current state the loading preserves the original schema. An open question is whether to harmonize the two schemas when loading the dataset or to preserve the original one. Since not all fields are overlapping the schema will not be exactly the same.


datasets/mbpp/README.md
```diff
@@ -0,0 +1,188 @@
...
+- expert-generated
+languages:
+- en-US
+licenses:
+- cc-by-4.0
+multilinguality:
+- monolingual
+pretty_name: Mostly Basic Python Problems
+size_categories:
+- unknown
```
lhoestq, line 16:
```suggestion
- n<1K
```

datasets/mbpp/README.md
```diff
@@ -0,0 +1,188 @@
+---
+annotations_creators:
+- crowdsourced
+- expert-generated
+language_creators:
+- crowdsourced
+- expert-generated
+languages:
+- en-US
```
lhoestq, line 9:
```suggestion
- en
```

lhoestq (approved):
Cool ! Thanks for adding this dataset :)

lhoestq:
I think it's fine to have the original schema

PR #2906, opened 2021-09-14
feat: 🎸 add a function to get a dataset config's split names

severo:
Also: pass additional arguments (use_auth_token) to get private configs + info of private datasets on the hub

Questions:

- [x] I'm not sure how the versions work: I changed 1.12.1.dev0 to 1.12.1.dev1, was it correct?
    -> no: reverted
- [x] Should I add a section in https://github.com/huggingface/datasets/blob/master/docs/source/load_hub.rst? (there is no section for get_dataset_infos)
    -> yes: added

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -220,7 +220,7 @@ setup(
     version="1.12.2.dev0",  # expected format is one of x.y.z.dev0, or x.y.z.rc1 or x.y.z (no to dashes, yes to dots)
     description=DOCLINES[0],
     long_description="\n".join(DOCLINES[2:]),
-    long_description_content_type='text/markdown',
+    long_description_content_type="text/markdown",
     author="HuggingFace Inc.",
     author_email="thomas@huggingface.co",
     url="https://github.com/huggingface/datasets",
--- a/src/datasets/inspect.py
+++ b/src/datasets/inspect.py
@@ -19,9 +19,10 @@
 from typing import Optional
 
 from .hf_api import HfApi
-from .load import import_main_class, prepare_module
+from .load import import_main_class, load_dataset_builder, prepare_module
 from .utils import DownloadConfig
 from .utils.logging import get_logger
+from .utils.streaming_download_manager import StreamingDownloadManager
 
 
 logger = get_logger(__name__)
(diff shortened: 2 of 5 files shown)


severo:
I'm not sure how the versions work: should I change 1.12.1.dev0 to 1.12.1.dev1?

setup.py
```diff
@@ -217,7 +217,7 @@
 
 setup(
     name="datasets",
-    version="1.12.1.dev0",  # expected format is one of x.y.z.dev0, or x.y.z.rc1 or x.y.z (no to dashes, yes to dots)
+    version="1.12.1.dev1",  # expected format is one of x.y.z.dev0, or x.y.z.rc1 or x.y.z (no to dashes, yes to dots)
```
lhoestq, line 220: why this change ?

severo: I was not sure if every commit must increase the version or not.

severo: Note that in another project I depend on a commit of the git repository of datasets:

https://github.com/huggingface/datasets-preview-backend/blob/10fd611042b41f2c8bed767a080f3a57b4429653/pyproject.toml#L28

And the dependency is updated only if the version increases. Without this, I have to do:

```
poetry remove datasets
```

add the line manually to pyproject.toml with the new commit, then

```
poetry lock
poetry install
```

severo: I will revert this.

src/datasets/inspect.py
```diff
@@ -129,3 +130,27 @@ def get_dataset_config_names(path: str):
...
+    """
+    builder = load_dataset_builder(path, name=config_name)
+    if builder.info.splits is None:
+        try:
+            return [
+                split_generator.name
+                for split_generator in builder._split_generators(StreamingDownloadManager(base_path=builder.base_path))
+            ]
+        except Exception as err:
+            raise Exception("The split names could not be parsed from the dataset config.") from err
```
severo, line 155: Maybe we should give that Exception its own name? I'm not sure what's the usage in datasets for that case

lhoestq: Yes feel free to create a new error type for this

lhoestq (commented):
Looks good to me ! Can you just merge master into your branch ?
We just deprecated the `script_version` parameter in favor of `revision` in `load_dataset_builder` and `prepare_module`, so you might need to do a few adjustments

lhoestq:
> Should I add a section in https://github.com/huggingface/datasets/blob/master/docs/source/load_hub.rst? (there is no section for get_dataset_infos)

Yes totally :) This tutorial should indeed mention this, given how fundamental it is

src/datasets/inspect.py
severo: Done

src/datasets/inspect.py
```diff
@@ -99,7 +107,18 @@ def inspect_metric(path: str, local_path: str, download_config: Optional[Downloa
...
-def get_dataset_infos(path: str):
+def get_dataset_infos(
+    path: str,
+    revision: Optional[Union[str, Version]] = None,
+    download_config: Optional[DownloadConfig] = None,
+    download_mode: Optional[GenerateMode] = None,
+    force_local_path: Optional[str] = None,
+    dynamic_modules_path: Optional[str] = None,
+    return_resolved_file_path: bool = False,
+    return_associated_base_path: bool = False,
```
lhoestq, line 118: These parameters may get removed from `prepare_module` in #2986 , so maybe let's not have them here

severo: ok

severo: Done

lhoestq (approved):
Thanks for the change and for adding it to the documentation :)
LGTM

PR #2910, opened 2021-09-14
feat: 🎸 pass additional arguments to get private configs + info

severo:
`use_auth_token` can now be passed to the functions to get the configs
or infos of private datasets on the hub


severo:
Included in https://github.com/huggingface/datasets/pull/2906

PR #2911, opened 2021-09-14
Fix exception chaining

albertvillanova:
Fix exception chaining to avoid tracebacks with message: `During handling of the above exception, another exception occurred:`

### DIFF
--- a/src/datasets/utils/file_utils.py
+++ b/src/datasets/utils/file_utils.py
@@ -152,8 +152,8 @@ def head_hf_s3(
             hf_bucket_url(identifier=identifier, filename=filename, use_cdn=use_cdn, dataset=dataset),
             max_retries=max_retries,
         )
-    except Exception as e:  # TODO: return or raise?
-        return e
+    except Exception as e:
+        return e  # TODO: return or raise?
 
 
 def hf_github_url(path: str, name: str, dataset=True, version: Optional[str] = None) -> str:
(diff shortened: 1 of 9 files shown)


src/datasets/utils/file_utils.py
```diff
@@ -152,7 +152,7 @@ def head_hf_s3(
             hf_bucket_url(identifier=identifier, filename=filename, use_cdn=use_cdn, dataset=dataset),
             max_retries=max_retries,
         )
-    except Exception as e:
+    except Exception as e:  # TODO: return or raise?
```
albertvillanova, line 155: @lhoestq is this a bug? The exception is returned instead of raised.

src/datasets/utils/file_utils.py
```diff
@@ -153,7 +153,7 @@ def head_hf_s3(
             max_retries=max_retries,
         )
     except Exception as e:
-        return e
+        return e  # TODO: return or raise?
```
albertvillanova, line 156: @lhoestq is this a bug? The exception is returned instead of raised.

lhoestq: No this is on purpose, we don't want any error to be raised from it in case there's no internet

lhoestq (approved):
Nice !

src/datasets/utils/file_utils.py
albertvillanova: Then, this makes no sense: https://github.com/huggingface/datasets/blob/master/src/datasets/load.py#L431-L434

lhoestq: Oh right !

lhoestq: Let's try to change it to raise then to remove the confusion

albertvillanova: Now `head_hf_s3` may raise an exception. I fix `prepare_module` to ignore the exception when it is not pertinent.

Issue #2914, opened 2021-09-15
Having a dependency defining fsspec entrypoint raises an AttributeError when importing datasets


pierre-godard:
## Describe the bug
In one of my project, I defined a custom fsspec filesystem with an entrypoint.
My guess is that by doing so, a variable named `spec` is created in the module `fsspec` (created by entering a for loop as there are entrypoints defined, see the loop in question [here](https://github.com/intake/filesystem_spec/blob/0589358d8a029ed6b60d031018f52be2eb721291/fsspec/__init__.py#L55)).
So that `fsspec.spec`, that was previously referring to the `spec` submodule, is now referring to that `spec` variable.
This make the import of datasets failing as it is using that `fsspec.spec`.

## Steps to reproduce the bug
I could reproduce the bug with a dummy poetry project.

Here is the pyproject.toml:
```toml
[tool.poetry]
name = "debug-datasets"
version = "0.1.0"
description = ""
authors = ["Pierre Godard"]

[tool.poetry.dependencies]
python = "^3.8"
datasets = "^1.11.0"

[tool.poetry.dev-dependencies]

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"

[tool.poetry.plugins."fsspec.specs"]
"file2" = "fsspec.implementations.local.LocalFileSystem"
```

The only other file being a `debug_datasets/__init__.py` empty file.

The overall structure of the project is as follows:
```
.
├── pyproject.toml
└── debug_datasets
    └── __init__.py
```

Then, within the project folder run:

```
poetry install
poetry run python
```

And in the python interpreter, try to import `datasets`:

```
import datasets
```

## Expected results
The import should run successfully.

## Actual results

Here is the trace of the error I get:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/godarpi/.cache/pypoetry/virtualenvs/debug-datasets-JuFzTKL--py3.8/lib/python3.8/site-packages/datasets/__init__.py", line 33, in <module>
    from .arrow_dataset import Dataset, concatenate_datasets
  File "/home/godarpi/.cache/pypoetry/virtualenvs/debug-datasets-JuFzTKL--py3.8/lib/python3.8/site-packages/datasets/arrow_dataset.py", line 48, in <module>
    from .filesystems import extract_path_from_uri, is_remote_filesystem
[...]

PR #2915, opened 2021-09-15
Fix fsspec AbstractFileSystem access

pierre-godard:
This addresses the issue from #2914 by changing the way fsspec's AbstractFileSystem is accessed.


lhoestq (approved):
Thanks for the fix :) LGTM

PR #2897, opened 2021-09-11
Add OpenAI's HumanEval dataset

lvwerra:
This PR adds OpenAI's [HumanEval](https://github.com/openai/human-eval) dataset. The dataset consists of 164 handcrafted programming problems with solutions and unittests to verify solution. This dataset is useful to evaluate code generation models.

PR #2916, opened 2021-09-15
Add OpenAI's pass@k code evaluation metric

lvwerra:
This PR introduces the `code_eval` metric which implements [OpenAI's code evaluation harness](https://github.com/openai/human-eval) introduced in the [Codex paper](https://arxiv.org/abs/2107.03374). It is heavily based on the original implementation and just adapts the interface to follow the `predictions`/`references` convention.

The addition of this metric should enable the evaluation against the code evaluation datasets added in #2897 and #2893.

A few open questions:

- The implementation makes heavy use of multiprocessing which this PR does not touch. Is this conflicting with multiprocessing natively integrated in `datasets`?
- This metric executes generated Python code and as such it poses dangers of executing malicious code. OpenAI addresses this issue by 1) commenting the `exec` call in the code so the user has to actively uncomment it and read the warning and 2) suggests using a sandbox environment (gVisor container). Should we add a similar safeguard? E.g. a prompt that needs to be answered when initialising the metric? Or at least a warning message?
- Naming: the implementation sticks to the `predictions`/`references` naming, however, the references are not reference solutions but unittest to test the solution. While reference solutions are also available they are not used. Should the naming be adapted?


lhoestq (commented):
What an interesting metric ! Thanks for adding it :)

I think we can keep predictions and references, it's not the best name but it would be difficult to change.

Also regarding the execution of code, I agree the strict minimum is to show a BIG warning to the user. 

One thing that we can do is raise an error that explains the risks and that says that the user has to set an environment variable like `HF_ALLOW_CODE_EVAL` is set to `1`, in order to make sure they commit to taking those risks. This could be a parameter to pass to `load_metric` as well. Let me know what you think is best or if you have other ideas

lhoestq:
> The implementation makes heavy use of multiprocessing which this PR does not touch. Is this conflicting with multiprocessing natively integrated in datasets?

It should work normally, but feel free to test it.
There is some documentation about using metrics in a distributed setup that uses multiprocessing [here](https://huggingface.co/docs/datasets/loading.html?highlight=rank#distributed-setup)
You can test to spawn several processes where each process would load the metric. Then in each process you add some references/predictions to the metric. Finally you call compute() in each process and on process 0 it should return the result on all the references/predictions

metrics/code_eval/code_eval.py
```diff
@@ -0,0 +1,186 @@
...
+    references: a list with a test for each prediction. Each test should evaluate the
+        correctness of a code candidate.
+    k: number of code candidates to consider in the evaluation (Default: [1, 10, 100])
+    n_workers: number of workers used to evaluate the canidate programs (Default: 4).
+    timeout:
+Returns:
+    pass_at_k: dict with pass rates for each k
+Examples:
+    >>> import os
+    >>> os.environ["HF_ALLOW_CODE_EVAL"] = "1"
```
lvwerra, line 79: @lhoestq I really liked your idea of having to set an environment variable. Currently, there is one issue: I need to set this also in the example in the doc string since the example is tested. I would like to avoid this since I think a lot of users would blindly copy-paste the example and then never see the warning that is raised if it is not set. Is there a way to set the environment variable in the test environment. Or do you propose another approach?

lvwerra:
Is there a good way to debug the Windows tests? I suspect it is an issue with `multiprocessing`, but I can't see the error messages.

metrics/code_eval/code_eval.py
lhoestq: Yes we can set the env variable in the CI runner in `.circleci/config.yml`

lhoestq:
Indeed it has an issue on windows.
In your example it's supposed to output
```python
{'pass@1': 0.5, 'pass@2': 1.0}
```
but it gets
```python
{'pass@1': 0.0, 'pass@2': 0.0}
```

I'm not on my windows machine today so I can't take a look at it. I can dive into it early next week if you want

lvwerra:
> I'm not on my windows machine today so I can't take a look at it. I can dive into it early next week if you want

That would be great - unfortunately I have no access to a windows machine at the moment. I am quite sure it is an issue with in exectue.py because of multiprocessing.

lhoestq (approved):
Thanks ! It looks all good :)

PR #2931, opened 2021-09-16
Fix bug in to_tf_dataset

Rocketknight1:
Replace `set_format()` to `with_format()` so that we don't alter the original dataset in `to_tf_dataset()`


lhoestq (approved):
Thanks for the fix :)

I think leaving the test with the try/except when deleting the temp dir could leave temp files on windows. Feel free to merge, I can try to find what we can do (not a big deal though)

Rocketknight1:
I'm going to merge it, but yeah - hopefully the CI runner just cleans that up automatically and few other people run the tests on Windows anyway!

PR #2936, opened 2021-09-17
Check that array is not Float as nan != nan

Iwontbecreative:
The Exception wants to check for issues with StructArrays/ListArrays but catches FloatArrays with value nan as nan != nan.
Pass on FloatArrays as we should not raise an Exception for them.


lhoestq (approved):
Thanks for the quick fix.

Once we set the minimum supported version of PyArrow to be 2.0.0 we'll be able to remove the overflow exceptions :)

PR #2939, opened 2021-09-17
MENYO-20k repo has moved, updating URL

cdleong:
Dataset repo moved to https://github.com/uds-lsv/menyo-20k_MT, now editing URL to match.

https://github.com/uds-lsv/menyo-20k_MT/blob/master/data/train.tsv is the file we're looking for


lhoestq (approved):
Thanks !

I also updated the dataset_infos.json file that registers the URLs and file checksums for data integrity verifications

PR #2949, opened 2021-09-20
Introduce web and wiki config in triviaqa dataset

shirte:
The TriviaQA paper suggests that the two subsets (Wikipedia and Web)
should be treated differently. There are also different leaderboards
for the two sets on CodaLab. For that reason, introduce additional
builder configs in the trivia_qa dataset.


lhoestq (commented):
Nice thanks ! It looks all good to me :)

I only have one comment:

We try to keep the dummy data as small as possible. The rc, rc.web, rc.wikipedia and unfiltered ones are a bit big because of the evidence files. Feel free to only keep a few sentences in them in order to make the dummy data even smaller.

lhoestq:
I just made the dummy data smaller :)
Once github refreshes the change I think we can merge !

shirte:
Thank you so much for reviewing and accepting my pull request!! :)

I created these rather large dummy data sets to cover all different cases for the row structure. E.g. in the web configuration, it's possible that a row has evidence from wikipedia ("EntityPages") and the web ("SearchResults"). But it also might happen that either EntityPages or SearchResults is empty. Probably, I will add this thought to the dataset description in the future.

lhoestq:
Ok I see ! Yes feel free to mention it in the dataset card, this can be useful.

For the dummy data though we can keep the small ones, as the tests are mainly about testing the parsing from the dataset script rather than the actual content of the dataset.

PR #2951, opened 2021-09-20
Dummy labels no longer on by default in `to_tf_dataset`

Rocketknight1:
After more experimentation, I think I have a way to do things that doesn't depend on adding `dummy_labels` - they were quite a hacky solution anyway!


lhoestq (approved):
Great !
Should we actually keep this feature now that it seems to work without it ?

Rocketknight1:
@lhoestq Let me make sure we never need it, and if not then I'll remove it entirely in a follow-up PR.

PR #2954, opened 2021-09-22
Run tests in parallel

albertvillanova:
Run CI tests in parallel to speed up the test suite.

Speed up results:
- Linux: from `7m 30s` to `5m 32s`
- Windows: from `13m 52s` to `11m 10s`


albertvillanova:
There is a speed up in Windows machines:
- From `13m 52s` to `11m 10s`

In Linux machines, some workers crash with error message:
```
OSError: [Errno 12] Cannot allocate memory
```

albertvillanova:
There is also a speed up in Linux machines:
- From `7m 30s` to `5m 32s`

PR #2959, opened 2021-09-23
Added computer vision tasks

merveenoyan:
Added various image processing/computer vision tasks.

### DIFF
--- a/src/datasets/utils/resources/tasks.json
+++ b/src/datasets/utils/resources/tasks.json
@@ -92,6 +92,57 @@
             "emotion-recognition"
         ]
     },
+    "semantic-segmentation": {
+        "description": "tasks related to linking pixels to class labels",
+        "options": [
+            "semantic-segmentation",
+            "3d-semantic-segmentation"
+        ]
+    },
+    "object-detection": {
+        "description": "tasks related to detecting instances of objects from a particular class in an image",
+        "options": [
+            "object-detection",
+            "3d-object-detection"
+        ]
+    },
+    "image-generation": {
+        "description": "tasks related to generating images from an existing dataset",
+        "options": [
+            "image-generation",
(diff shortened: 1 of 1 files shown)


lhoestq:
Looks great, thanks ! If the 3d ones are really rare we can remove them for now.

And I can see that `object-detection` and `semantic-segmentation` are both task categories (top-level) and task ids (bottom-level). Maybe there's a way to group them and have less granularity for the task categories. For example `speech-processing` is a high level task category. What do you think ?

We can still update the list of tasks later if needed when we have more vision datasets

osanseviero (approved):
Thanks for the PR!

> If the 3d ones are really rare we can remove them for now.

+1. This is an area in which we'll want to grow in the future but right now we don't have any 3D stuff.

I was also wondering if `optical-character-recognition` can not be modeled as `object-detection` and hence be under same tag. Not sure though!

merveenoyan:
@lhoestq  @osanseviero I used the categories (there were main ones and subcategories) in the paperswithcode, I got rid of some of them that could be too granular. I can put it there if you'd like (I'll wait for your reply before committing it again)

lhoestq:
We can ignore the ones that are too granular IMO. What we did for audio tasks is to have them all under "audio-processing". Maybe we can do the same here for now until we have more comprehensive tasks/applications ?

lhoestq:
Following the discussion in (private) https://github.com/huggingface/moon-landing/issues/2020, what do you think of aligning the top level tasks list with the model tasks taxonomy ?

* Image Classification
* Object Detection
* Image Segmentation
* Text-to-Image
* Image-to-Text

merveenoyan:
I moved it to [a branch](https://github.com/huggingface/datasets/pull/3800) for ease.

PR #2974, opened 2021-09-27
Actually disable dummy labels by default

Rocketknight1:
So I might have just changed the docstring instead of the actual default argument value and not realized.

@lhoestq I'm sorry >.>


lhoestq (approved):
Haha good catch x) I didn't even notice

PR #2975, opened 2021-09-27
ignore dummy folder and dataset_infos.json

### LINKED ISSUE #2877: Don't keep the dummy data folder or dataset_infos.json when resolving data files
When there's no dataset script, all the data files of a folder or a repository on the Hub are loaded as data files.

There are already a few exceptions:
- files starting with "." are ignored
- the dataset card "README.md" is ignored
- any file named "config.json" is ignored (currently it isn't used anywhere, but it could be used in the future to define splits or configs for example, but not 100% sure)

However any data files in a folder named "dummy" should be ignored as well as they should only be used to test the dataset.
Same for "dataset_infos.json" which should only be used to get the `dataset.info`

Ishan-Kumar2:
Fixes #2877

Added the `dataset_infos.json` to the ignored files list and also added check to ignore files which have parent directory as `dummy`.
Let me know if it is correct. Thanks :)

### DIFF
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -233,7 +233,7 @@ def _resolve_data_files_locally_or_by_urls(
     Return the absolute paths to all the files that match the given patterns.
     It also supports absolute paths in patterns.
     If an URL is passed, it is returned as is."""
-    data_files_ignore = ["README.md", "config.json"]
+    data_files_ignore = ["README.md", "config.json", "dataset_infos.json"]
     if isinstance(patterns, str):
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
Hi ! Thanks for helping on this !

I just did a small change: I blacklisted "dummy_data.zip" instead of files inside a "dummy" directory, since I noticed that it would be more general. Indeed the "dummy_data.zip" are the ones to be ignored, but they can not only be inside a directory named "dummy", but also inside subdirectories of a directory named "dummy"

PR #3002, opened 2021-10-02
Remove a reference to the open Arrow file when deleting a TF dataset created with to_tf_dataset

### LINKED ISSUE #2934: to_tf_dataset keeps a reference to the open data somewhere, causing issues on windows
To reproduce:
```python
import datasets as ds
import weakref
import gc

d = ds.load_dataset("mnist", split="train")
ref = weakref.ref(d._data.table)
tfd = d.to_tf_dataset("image", batch_size=1, shuffle=False, label_cols="label")
del tfd, d
gc.collect()
assert ref() is None, "Error: there is at least one reference left"
```

This causes issues because the table holds a reference to an open arrow file that should be closed. So on windows it's not possible to delete or move the arrow file afterwards.

Moreover the CI test of the `to_tf_dataset` method isn't able to clean up the temporary arrow files because of this.

cc @Rocketknight1

mariosasko:
This [comment](https://github.com/huggingface/datasets/issues/2934#issuecomment-922970919) explains the issue. This PR fixes that with a `weakref` callback, and additionally:
* renames `TensorflowDatasetMixIn` to `TensorflowDatasetMixin` for consistency
* correctly indents `TensorflowDatasetMixin`'s docstring
* replaces `tf.data.AUTOTUNE` with `tf.data.experimental.AUTOTUNE` (we support TF>=2.2 according to the [setup.py](https://github.com/huggingface/datasets/blob/fc46bba66ba4f432cc10501c16a677112e13984c/setup.py#L188) and `AUTOTUNE` has been moved to the experimental part of `tf.data` in 1.X if I'm not mistaken)

Fixes #2934

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -247,22 +247,22 @@ class TensorflowDatasetMixIn:
             only columns that the model can use as input should be included here (numeric data only).
             batch_size (:obj:`int`): Size of batches to load from the dataset.
             shuffle(:obj:`bool`): Shuffle the dataset order when loading. Recommended True for training, False for
-             validation/evaluation.
+                validation/evaluation.
             drop_remainder(:obj:`bool`, default ``None``): Drop the last incomplete batch when loading. If not provided,
-             defaults to the same setting as shuffle.
+                defaults to the same setting as shuffle.
             collate_fn(:obj:`Callable`): A function or callable object (such as a `DataCollator`) that will collate
-             lists of samples into a batch.
+                lists of samples into a batch.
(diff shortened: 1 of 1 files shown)


lhoestq (commented):
Nice thanks ! It may also fix the PermissionError issue we have in the windows CI here:
https://github.com/huggingface/datasets/blob/8c95023c8439dc14cdb580d7dc5c476d492c8811/tests/test_arrow_dataset.py#L2055-L2059

Could you try removing the try/except and see if it fixes it ?

Also cc @Rocketknight1 can you take a look at this change please ?

mariosasko:
@lhoestq The test passes even without the try/except block!

lhoestq (approved):
Cool, thanks !

PR #3022, opened 2021-10-05
Add further details about the MeDAL dataset

xhlulu:
Added more details in the following sections:
* Dataset Structure
* Data Instances
* Data Splits
* Source Data
* Annotations
* Discussions of Biases
* LIcensing Information


xhlulu:
Thanks! Seems like it did the trick since the tests are passing. Let me know if there's anything else I can do in this PR!

PR #3028, opened 2021-10-05
Properly install ruamel-yaml for windows CI

lhoestq:


albertvillanova:
@lhoestq I would say this does not "properly" install `ruamel-yaml`, but the contrary, it overwrites it.

According to `pip` docs:
>  This can break your system if the existing package is of a different version or was installed with a different package manager!

That is why I did not fix the issue this way, made a hotfix pinning `huggingface_hub` (#3025), while searching a permanent solution for the issue:

lhoestq:
Yea I did this because we need the latest version of `huggingface_hub` for #2986 and because I didn't want to ssh to the windows worker x)
Maybe it can be fixed by installing it with conda - let me try

lhoestq:
Oh yea it may not work since it was first installed with distutils

PR #3030, opened 2021-10-05
(WIP) Add `remove_columns` to `IterableDataset`

### LINKED ISSUE #2944: Add  `remove_columns` to `IterableDataset `
A clear and concise description of what the problem is.

```python
dataset = load_dataset("c4", 'realnewslike', streaming =True, split='train')
dataset = dataset.remove_columns('url')
```
```
AttributeError: 'IterableDataset' object has no attribute 'remove_columns'
```

It would be nice to have `.remove_columns()` to match the `Datasets` api. 

This can be done with a single call to `.map()`, 

I can try to help add this. 🤗

cccntu:
Fixes #2944

WIP
* Not tested yet.
* We might want to allow batched remove for efficiency.

@lhoestq Do you think it should have `batched=` and `batch_size=`?

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -471,6 +471,29 @@ class IterableDataset(DatasetInfoMixin):
             shuffling=copy.deepcopy(self._shuffling),
         )
 
+    def remove_columns(self, column_names: Union[str, List[str]]) -> "IterableDataset":
+        """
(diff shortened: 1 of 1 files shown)


lhoestq:
I don't think that batching would help. Indeed we're dealing with python iterators that yield elements one by one, so batched `map` needs to accumulate a batch, apply the function, and then yield examples from the batch.

Though once we have parallel processing in `map`, we can reconsider it

PR #3067, opened 2021-10-12
add story_cloze

zaidalyafeai:


zaidalyafeai:
@lhoestq can't fix the last test fails.

zaidalyafeai:
Thanks Albert, I fixed the suggested comments. This dataset has no train splits, it is only used for evaluation.

PR #3068, opened 2021-10-13
feat: increase streaming retry config

borisdayma:
Increase streaming config parameters:
* retry interval set to 5 seconds
* max retries set to 20 (so 1mn 40s)


borisdayma:
@lhoestq I had 2 runs for more than 2 days each, continuously streaming (they were failing before with 3 retries at 1 sec interval).

They are running on TPU's (so great internet connection) and only had connection errors a few times each (3 & 4). Each time it worked after only 1 retry.
The reason for a higher number of retries is for local connections. It would allow for almost 2mn of a wifi/ethernet disconnection. In practice this should not happen very often.

Let me know if you think it's too much.

albertvillanova:
Thanks @borisdayma, the failing test is due to an issue in the master branch, that has already been fixed.

You can include the fix:
```
git checkout feat-streamtime
git fetch upstream master
git merge upstream/master
```

Issue #3089, opened 2021-10-15
JNLPBA Dataset


sciarrilli:
## Steps to reproduce the bug
```python
# Sample code to reproduce the bug
```

## Expected results
The dataset loading script for this dataset is incorrect. This is a biomedical dataset used for named entity recognition. The entities in the [script](https://github.com/huggingface/datasets/blob/master/datasets/jnlpba/jnlpba.py#L81-L83) are: O, B, and I.  The correct entities from the original data file are: 

['O',
 'B-DNA',
 'I-DNA',
 'B-RNA',
 'I-RNA',
 'B-cell_line',
 'I-cell_line',
 'B-cell_type',
 'I-cell_type',
 'B-protein',
 'I-protein']

## Actual results
The dataset loader script needs to include the following NER names:

['O',
 'B-DNA',
 'I-DNA',
 'B-RNA',
 'I-RNA',
 'B-cell_line',
 'I-cell_line',
 'B-cell_type',
 'I-cell_type',
 'B-protein',
 'I-protein']

And the [data](https://github.com/huggingface/datasets/blob/master/datasets/jnlpba/jnlpba.py#L46) that is being pulled has been modified from the original dataset and does not include the original NER tags.

## Environment info

- `datasets` version:
- Platform:
- Python version:
- PyArrow version:

sciarrilli:
# Steps to reproduce

`from datasets import load_dataset
dataset = load_dataset('jnlpba')

dataset['train'].features['ner_tags']
`

sciarrilli:
# Steps to reproduce

To reproduce:

```python
from datasets import load_dataset
dataset = load_dataset('jnlpba')

dataset['train'].features['ner_tags']
```
Output:
```python
Sequence(feature=ClassLabel(num_classes=3, names=['O', 'B', 'I'], names_file=None, id=None), length=-1, id=None)
```

PR #3092, opened 2021-10-15
Fix JNLBA dataset

bhavitvyamalik:
As mentioned in #3089, I've added more tags and also updated the link for dataset which was earlier using a Google Drive link. 
I'm having problem with generating dummy data as `datasets-cli dummy_data  ./datasets/jnlpba --auto_generate --match_text_files "*.iob2"` is giving `datasets.keyhash.DuplicatedKeysError: FAILURE TO GENERATE DATASET !
` error. I'll try to add dummy data manually.

### DIFF
--- a/datasets/jnlpba/README.md
+++ b/datasets/jnlpba/README.md
@@ -71,7 +71,11 @@ English
 
 ### Data Instances
 
-[Needs More Information]
+{
+  'id': '1',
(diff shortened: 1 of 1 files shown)


albertvillanova:
Fix #3089.

bhavitvyamalik:
@albertvillanova all tests are passing now. Either you or @lhoestq can review it!

PR #3110, opened 2021-10-19
Stream TAR-based dataset using iter_archive

lhoestq:
I converted all the dataset based on TAR archive to use iter_archive instead, so that they can be streamable.

It means that around 80 datasets become streamable :)


lhoestq:
The CI fails are only related to missing sections or tags in the dataset cards - which is unrelated to this PR

PR #3133, opened 2021-10-21
Support Audio feature in streaming mode

### LINKED ISSUE #3132: Support Audio feature in streaming mode
Currently, Audio feature is only supported for non-streaming datasets.

Due to the large size of many speech datasets, we should also support Audio feature in streaming mode.

albertvillanova:
Fix #3132.


lhoestq (approved):
Cool thanks !

Issue #2841, opened 2021-08-26
Adding GLUECoS Hinglish and Spanglish code-switching bemchmark


yjernite:
## Adding a Dataset
- **Name:** GLUECoS
- **Description:** a Microsoft Benchmark to evaluate code-switching for only two language pairs but a variety of tasks
- **Paper:** https://aclanthology.org/2020.acl-main.329/
- **Data:** https://github.com/microsoft/GLUECoS
- **Motivation:** We currently only have [one other](https://huggingface.co/datasets/lince) dataset for code-switching

Instructions to add a new dataset can be found [here](https://github.com/huggingface/datasets/blob/master/ADD_NEW_DATASET.md).

Ishan-Kumar2:
Hi @yjernite I am interested in adding this . 
In the repo they have also added a code mixed MT task from English to Hinglish [here](https://github.com/microsoft/GLUECoS#code-mixed-machine-translation-task). I think this could be a good dataset addition in itself and then I can add the rest of the GLUECoS tasks as one dataset. What do you think?

PR #3149, opened 2021-10-22
Add CMU Hinglish DoG Dataset for MT

Ishan-Kumar2:
Address part of #2841 

Added the CMU Hinglish DoG Dataset as in GLUECoS. Added it as a seperate dataset as unlike other tasks of GLUE CoS this can't be evaluated for a BERT like model. 
Consists of parallel dataset between Hinglish (Hindi-English) and English, can be used for Machine Translation between the two. 

The data processing part is inspired from the GLUECoS repo [here](https://github.com/microsoft/GLUECoS/blob/7fdc51653e37a32aee17505c47b7d1da364fa77e/Data/Preprocess_Scripts/preprocess_mt_en_hi.py)
The dummy data part is not working properly, it shows 
``` UnboundLocalError: local variable 'generator_splits' referenced before assignment ``` 
when I run without ``--auto_generate``.

Please let me know how I can fix that.
Thanks

### CODE REVIEW

**Comment by Ishan-Kumar2** [comment]
Hi @lhoestq, thanks a lot for the help. I have moved the part as suggested. 
Although still while running the dummy data script, I face this issue

```
  File "/home/ishan/anaconda3/bin/datasets-cli", line 8, in <module>
    sys.exit(main())
  File "/home/ishan/anaconda3/lib/python3.8/site-packages/datasets/commands/datasets_cli.py", line 33, in main
    service.run()
  File "/home/ishan/anaconda3/lib/python3.8/site-packages/datasets/commands/dummy_data.py", line 318, in run
    self._autogenerate_dummy_data(
  File "/home/ishan/anaconda3/lib/python3.8/site-packages/datasets/commands/dummy_data.py", line 363, in _autogenerate_dummy_data
    dataset_builder._prepare_split(split_generator)
  File "/home/ishan/anaconda3/lib/python3.8/site-packages/datasets/builder.py", line 1103, in _prepare_split
    example = self.info.features.encode_example(record)
  File "/home/ishan/anaconda3/lib/python3.8/site-packages/datasets/features/features.py", line 981, in encode_example
    return encode_nested_example(self, example)
  File "/home/ishan/anaconda3/lib/python3.8/site-packages/datasets/features/features.py", line 775, in encode_nested_example
    return {
  File "/home/ishan/anaconda3/lib/python3.8/site-packages/datasets/features/features.py", line 775, in <dictcomp>
    return {
  File "/home/ishan/anaconda3/lib/python3.8/site-packages/datasets/utils/py_utils.py", line 99, in zip_dict
    yield key, tuple(d[key] for d in dicts)
  File "/home/ishan/anaconda3/lib/python3.8/site-packages/datasets/utils/py_utils.py", line 99, in <genexpr>
    yield key, tuple(d[key] for d in dicts)
KeyError: 'status'
```
This KeyError is at times different from 'status' also.
when I run 
```
datasets-cli dummy_data datasets/cmu_hinglish_dog --auto_generate --json_field='history'
```
I have tried removing unnecessary feature type definition, but that didn't help. Please let me know if I am missing something, thanks!

File: datasets/cmu_hinglish_dog/cmu_hinglish_dog.py [hunk 1]
```diff
@@ -0,0 +1,190 @@
...
+        data_dir_hi_en = os.path.join(
+            hi_en_path, "CMUHinglishDoG", "Conversations_Hinglish"
+        )
+
+        # Extensions in Hinglish dataset are .json_1 and .json_2, converting them to .json
+        for split in os.listdir(data_dir_hi_en):
+            for f in os.listdir(os.path.join(data_dir_hi_en, split)):
+                f = os.path.join(os.path.join(data_dir_hi_en, split, f))
+                p = Path(f)
+                os.rename(f, p.with_suffix(".json"))
```
lhoestq, line 128: You can remove this part an open the .json_1 and .json_2 files

datasets/cmu_hinglish_dog/cmu_hinglish_dog.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def _split_generators(self, dl_manager):
+        """The linking part between Hinglish data and English data is inspired from the implementation in GLUECoS.
+        Refer here for the original script https://github.com/microsoft/GLUECoS/blob/7fdc51653e37a32aee17505c47b7d1da364fa77e/Data/Preprocess_Scripts/preprocess_mt_en_hi.py"""
+
+        eng_path = dl_manager.download_and_extract(_URL_ENGLISH)
+
+        # WikiData file is not required for this
+        if os.path.isdir(os.path.join(eng_path, "datasets-CMU_DoG-master", "WikiData")):
+            shutil.rmtree(os.path.join(eng_path, "datasets-CMU_DoG-master", "WikiData"))
```
lhoestq, line 114: I'm not sure how deleting a directory is handled for the cache - would it be possible to remove this part so that the cache stays intact ?

lhoestq (commented):
Nice ! it looks much cleaner now :)

I think you don't have to pass `--json_field='history'`, otherwise it will ignore all the other fields like "status", can you try again without passing this parameter ?

Also some more minor comments:

datasets/cmu_hinglish_dog/cmu_hinglish_dog.py
Ishan-Kumar2: Yes ``.json_1`` and ``.json_2`` can be opened, but the dummy data generation script skips it because the extension is not ``.json`` [here](https://github.com/huggingface/datasets/blob/b6bfe88d9ae11ba51eb8b4ba720826cf9274969c/src/datasets/commands/dummy_data.py#L136)

Ishan-Kumar2: Also I tested out without ``--json_field`` also, it gives the error Can't parse output
```
Using custom data configuration default
Couldn't generate dummy file 'datasets/cmu_hinglish_dog/dummy/0.0.0/dummy_data/Conversations.zip/datasets-CMU_DoG-master/README.md'. Ignore that if this file is not useful for dummy data.
  File "/home/ishan/anaconda3/bin/datasets-cli", line 8, in <module>
    sys.exit(main())
  File "/home/ishan/anaconda3/lib/python3.8/site-packages/datasets/commands/datasets_cli.py", line 33, in main
    service.run()
  File "/home/ishan/anaconda3/lib/python3.8/site-packages/datasets/commands/dummy_data.py", line 318, in run
    self._autogenerate_dummy_data(
  File "/home/ishan/anaconda3/lib/python3.8/site-packages/datasets/commands/dummy_data.py", line 346, in _autogenerate_dummy_data
    dl_manager.auto_generate_dummy_data_folder(
  File "/home/ishan/anaconda3/lib/python3.8/site-packages/datasets/commands/dummy_data.py", line 87, in auto_generate_dummy_data_folder
    total += self._create_dummy_data(
  File "/home/ishan/anaconda3/lib/python3.8/site-packages/datasets/commands/dummy_data.py", line 178, in _create_dummy_data
    total += self._create_dummy_data(
  File "/home/ishan/anaconda3/lib/python3.8/site-packages/datasets/commands/dummy_data.py", line 145, in _create_dummy_data
    raise ValueError(
ValueError: Couldn't parse columns ['date', 'history', 'rating', 'status', 'uid1LogInTime', 'uid1LogOutTime', 'uid1response', 'uid2LogInTime', 'uid2LogOutTime', 'uid2response', 'user1_id', 'user2_id', 'whoSawDoc', 'wikiDocumentIdx']. Maybe specify which json field must be used to read the data with --json_field <my_field>.
```

Ishan-Kumar2: [This line](https://github.com/huggingface/datasets/blob/b6bfe88d9ae11ba51eb8b4ba720826cf9274969c/src/datasets/commands/dummy_data.py#L142) requires every value of json to a be a ``list``. That is not true in this data since each datapoint is a seperate json, so do I need to pre-process the data?

lhoestq: Good catch, that's indeed an issue for your case. I guess this format is not supported for the dummy data generator. You can still create the dummy data manually.

To do so:
1. go into the directory `datasets/cmu_hinglish_dog/dummy/0.0.0` and create a directory `dummy_data`.
2. go into this directory and create two directories: `Conversations.zip`and `CMUHinglishDoG.zip` (don't zip them, this is just to say that they contains the dummy data for each archive)
3. copy the content of the actual archives in them, but only keep 1 or 2 files per split. You can also remove the unnecessary files like the wikidata folder
4. zip the dummy_data directory into `dummy_data.zip`
5. run the command
```bash
RUN_SLOW=1 pytest tests/test_dataset_common.py::LocalDatasetTest::test_load_dataset_all_configs_cmu_hinglish_dog
```

Let me know if that worked, and feel free to ping me if I can help

Ishan-Kumar2: Okay sure, I'll do it using these steps. Thanks!

datasets/cmu_hinglish_dog/cmu_hinglish_dog.py
Ishan-Kumar2: Was causing an issue in automatic dummy data part, fixed now.

datasets/cmu_hinglish_dog/cmu_hinglish_dog.py
```diff
@@ -108,70 +104,74 @@ def _split_generators(self, dl_manager):
...
+                en = json.load(
+                    open(os.path.join(os.path.join(data_dir_en, "valid"), en_file_path))
+                )
+            elif en_file_path in english_files_test:
+                en = json.load(
+                    open(os.path.join(os.path.join(data_dir_en, "test"), en_file_path))
+                )
+            else:
+                found = False
+            if found:
```
Ishan-Kumar2, line 170: I have modified the part of searching for the corresponding file in the english set. Earlier the English counterparts of a split where only being searched for in that specific split. I found that a lot of Hinglish files in ``val`` and ``test`` have english counterparts in ``train`` of English dir, so it didn't make sense to not use them.

datasets/cmu_hinglish_dog/cmu_hinglish_dog.py
```diff
@@ -180,10 +180,28 @@ def _generate_examples(self, hi_en_dir, en_dir):
...
-                        "uid": x["uid"],
-                        "docIdx": x["docIdx"],
-                        "hi_en": y["text"],
-                        "en": x["text"],
+                        "date": hi_en["date"],
+                        "history": {
+                            "uid": x["uid"],
+                            "docIdx": [x["docIdx"]],
+                            "utcTimestamp": x["utcTimestamp"],
+                            "text": x["text"] + "_" + y["text"],
```
Ishan-Kumar2, line 188: Since the type of input and output data types have to be the same, I had to output both Hinglish and English texts in a single text field (as input has only 1 text field).
Not to sure about this way of returning the texts seperated by a _. Please let me know if I am missing something or if there is an alternative :)

lhoestq (approved):
Thanks for the changes :)

I just fixed the feature type of the "translation" field to use the `Translation` type, and regenerated the daatset_infos.json

I also added more information in the dataset card

I think it's all good now :) thanks for the contribution !

lhoestq:
The CI fail is unrelated to this PR and fixed on master. Merging !

PR #3160, opened 2021-10-25
Better error msg if `len(predictions)` doesn't match `len(references)` in metrics

mariosasko:
Improve the error message in `Metric.add_batch` if `len(predictions)` doesn't match `len(references)`.

cc: @BramVanroy (feel free to test this code on your examples and review this PR)


src/datasets/metric.py
```diff
@@ -435,13 +436,20 @@ def add_batch(self, *, predictions=None, references=None):
...
+            if match is not None:
+                error_msg = (
+                    f"Mismatch in the number of predictions ({match.group(1)}) and references ({match.group(2)})"
+                )
+            else:
+                error_msg = (
+                    f"Predictions and/or references don't match the expected format.\n"
+                    f"Expected format: {self.features},\n"
+                    f"Input predictions: {predictions},\n"
+                    f"Input references: {references}"
```
lhoestq, line 450: Maybe let's not print the full list of predictions and references ? It could be very very long ^^'

lhoestq (approved):
Love it ! Thanks :)

Just one minor comment:

src/datasets/metric.py
mariosasko: Yes, I agree. We can implement a simple summarization similar to NumPy arrays. 

Currently, [NumPy](https://numpy.org/doc/stable/reference/generated/numpy.set_printoptions.html#numpy.set_printoptions) triggers summarization on arrays with more than 1000 elements. Do you have a threshold in mind for our use case?

lhoestq: We can just crop at 5 elements IMO

mariosasko: Let's make it 6 to have an equal number of elements on both sides of the summarized list if `len(batch) == 6`. 😃

mariosasko:
I added a function, but it only works with the `list` type. For arrays/tensors, we delegate formatting to the frameworks.

lhoestq (approved):
Sounds good, thanks !

PR #3180, opened 2021-10-29
fix label mapping

VictorSanh:
Fixing label mapping for hlgd.
0 correponds to same event and 1 corresponds to different event
Capture d’écran 2021-10-29 à 10 39 58 AM
Capture d’écran 2021-10-29 à 10 40 09 AM
nt


VictorSanh:
heck, test failings. moving to draft. will come back to this later today hopefully

lhoestq:
Thanks for fixing this :)
I just updated the dataset_infos.json and added the missing `pretty_name` tag to the dataset card

VictorSanh:
thank you @lhoestq! running around as always it felt through as a lower priority...

PR #3182, opened 2021-10-29
Don't memoize strings when hashing since two identical strings may have different python ids

lhoestq:
When hashing an object that has several times the same string, the hashing could return a different hash if the identical strings share the same python `id()` or not.

Here is an example code that shows how the issue can affect the caching:
```python
import json
import pyarrow as pa
from datasets.features import Features

schema = pa.schema([pa.field("some_string", pa.string()), pa.field("another_string", pa.string())])
features_from_schema = Features.from_arrow_schema(schema)
Hasher.hash(features_from_schema) # dffa9dca9a73fd8c

features_dict = json.loads('{"some_string": {"dtype": "string", "id": null, "_type": "Value"}, "another_string": {"dtype": "string", "id": null, "_type": "Value"}}')
features_from_json = Features.from_dict(features_dict)
Hasher.hash(features_from_json) # 3812e76b15e6420e

features_from_schema == features_from_json # True
```

This is because in `features_dict`, some strings like "dtype" are repeated but don't share the same id, contrary to the ones in `features_from_schema`.

I fixed that by disabling memoization for strings.

This could be optimized in the future by implementing a smarter memoization with a special handling for strings.

### DIFF
--- a/src/datasets/utils/py_utils.py
+++ b/src/datasets/utils/py_utils.py
@@ -323,6 +323,11 @@ class Pickler(dill.Pickler):
         else:
             dill.Pickler.save_global(self, obj, name=name)
 
+    def memoize(self, obj):
+        # don't memoize strings since two identical strings can have different python ids
+        if type(obj) != str:
+            dill.Pickler.memoize(self, obj)
+
 
 def dump(obj, file):
     """pickle an object to a file"""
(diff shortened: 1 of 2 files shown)


lhoestq:
This change slows down the hash computation a little bit but from my tests it doesn't look too impactful. So I think it's fine to merge this.

PR #3184, opened 2021-10-30
RONEC v2

dumitrescustefan:
Hi, as we've recently finished with the new RONEC (Romanian Named Entity Corpus), we'd like to update the dataset here as well. It's actually essential as links to V1 are no longer valid. 

In reality we'd like to replace completely v1, as v2 is a full re-annotation of v1 with additional data (up to 2x size vs v1).

I've run the make style and all the dummy and real data test, and they passed.

I hope it's okay to merge the new RONEC v2 in the datasets. 

Thanks!


datasets/ronec/ronec.py
```diff
@@ -59,53 +57,52 @@ def __init__(self, **kwargs):
 class RONEC(datasets.GeneratorBasedBuilder):
     """RONEC dataset"""
 
-    VERSION = datasets.Version("1.0.0")
+    VERSION = datasets.Version("2.0.0")
     BUILDER_CONFIGS = [
         RONECConfig(name="ronec", version=VERSION, description="RONEC dataset"),
     ]
```
lhoestq, line 63: It would be nice to still let users access the V1 for reproducibility of all the work before RONEC v2

Do you think you could add a configuration that allows users to load RONEC V1 ? You can use the new links to the V1 data files

lhoestq (commented):
Awesome thanks ! I just added a comment about keeping a way to load V1, let me know what you think

dumitrescustefan:
@lhoestq Thanks for the review. I totally understand what you are saying. Normally, I would definitely agree with you, but in this particular case, the quality of v1 is poor, and the dataset itself is small (at the time we created v1 it was the only RO NER dataset, and its size was limited by the available resources). 

This is why we worked to build a larger one, with much better inter-annotator agreement. Fact is, models trained on v1 will be of very low quality and I would not recommend to anybody to use/do that. That's why I'd strongly suggest we replace v1 with v2, and kindof make v1 vanish :)  

What do you think? If you insist on having v1 accessible, I'll add the required code. Thanks!

lhoestq (approved):
Thanks again for updating the dataset !
I think we can merge now :)

PR #3195, opened 2021-11-02
More robust `None` handling

### LINKED ISSUE #3181: `None` converted to `"None"` when loading a dataset
## Describe the bug
When loading a dataset `None` values of the type `NoneType` are converted to `'None'` of the type `str`.

## Steps to reproduce the bug
```python
qasper = load_dataset("qasper", split="train", download_mode="reuse_cache_if_exists")
print(qasper[60]["full_text"]["section_name"])
```

When installing version 1.1.40, the output is
`[None, 'Introduction', 'Benchmark Datasets', ...]`

When installing from the master branch, the output is
`['None', 'Introduction', 'Benchmark Datasets', ...]`

Notice how the first element was changed from `NoneType` to `str`.

## Expected results
`None` should stay as is.

## Actual results
`None` is converted to a string.

## Environment info

- `datasets` version: master
- Platform: Linux-4.4.0-19041-Microsoft-x86_64-with-glibc2.17
- Python version: 3.8.10
- PyArrow version: 4.0.1

mariosasko:
PyArrow has explicit support for `null` values, so it makes sense to support Nones on our side as well.

[Colab Notebook with examples](https://colab.research.google.com/drive/1zcK8BnZYnRe3Ao2271u1T19ag9zLEiy3?usp=sharing)

Changes:
* allow None for the features types with special encoding (`ClassLabel, TranslationVariableLanguages, Value, _ArrayXD`)
* handle None in `class_encode_column` (also there is an option to stringify Nones and treat them as a class)
* support None sorting in `sort` (use pandas for that) 
* handle None in align_labels_with_mapping
* support for None in ArrayXD (converts `None` to `np.nan` to align the behavior with PyArrow)
* support for None in the Audio/Image feature
* allow promotion when concatenating tables (`pa.concat_tables(table_list, promote=True)`) and `null` row/~~column~~ broadcasting similar to pandas 

Additional notes:
* use `null` instead of `none` for function arguments for consistency with existing `disable_nullable` 
* fixes a bug with the `update_metadata_with_features` call in `Dataset.rename_columns`
* had to update some tests, let me know if that's ok

TODO:
- [x] check how the Audio features behaves with Nones
- [x] Better None handling in `concatenate_datasets`/`add_item`
- [x] Fix formatting with Nones
- [x] Add Colab with examples
- [x] Tests

TODOs for subsequent PRs:
-  Mention None handling in the docs
- Add `drop_null`/`fill_null` to `Dataset`/`DatasetDict`

Fix #3181 #3253


lhoestq:
I also created a PR regarding `disable_nullable` that must be always `False` by default, in order to always allow None values
https://github.com/huggingface/datasets/pull/3211

src/datasets/arrow_dataset.py
```diff
@@ -3348,13 +3402,22 @@ def add_column(self, name: str, column: Union[list, np.array], new_fingerprint:
...
         column_table = InMemoryTable.from_pydict({name: column})
+        _check_column_names(self.column_names + column_table.column_names)
+        # Pad tables if needed to match the maximum length
+        max_num_rows = max(self.num_rows, column_table.num_rows)
+        dset_table = pad_table_to_length(self._data, max_num_rows)
+        column_table = pad_table_to_length(column_table, max_num_rows)
+        indices_table = self._indices
+        if indices_table is not None and max_num_rows > self.num_rows:
+            indices_to_add = range(self._data.num_rows, self._data.num_rows + (max_num_rows - self.num_rows))
+            indices_table = add_indices_to_indices_table(self._indices, indices_to_add)
```
lhoestq, line 3413: Is this really necessary ? It could lead to more frequent silent errors rather than helping users no ?

src/datasets/arrow_dataset.py
```diff
@@ -3649,28 +3721,37 @@ def concatenate_datasets(
...
+    list_of_feautures = [dset.features for dset in dsets]
+
+    if axis == 0:
+        list_of_features = align_list_of_features(list_of_feautures)
+        tables = [table.cast(pa.schema(features.type)) for table, features in zip(tables, list_of_features)]
+    elif axis == 1:
+        _check_column_names([col_name for table in tables for col_name in table.column_names])
+
+        max_num_rows = max(dset.num_rows for dset in dsets)
+        tables = [pad_table_to_length(table, max_num_rows) for table in tables]
```
lhoestq, line 3735: same here

src/datasets/features/features.py
```diff
@@ -443,17 +450,22 @@ def to_list_of_numpy(self, zero_copy_only=True):
...
         arrays = []
         first_dim_offsets = np.array([off.as_py() for off in storage.offsets])
         for i in range(len(storage)):
-            storage_el = storage[i : i + 1]
-            first_dim = first_dim_offsets[i + 1] - first_dim_offsets[i]
-            # flatten storage
-            for _ in range(ndims):
-                storage_el = storage_el.flatten()
+            if i in null_indices:
+                arrays.append(np.nan)
```
lhoestq, line 459: Nit: maybe we can optimize this. If the number of null indices is high this can take time:
```python
for i, is_null in enumerate(storage.is_null().to_numpy(zero_copy_only=False)):
    if is_null:
        arrays.append(np.nan)
    else:
        ...

**Review by lhoestq — COMMENTED** [review]
Looking good !
I just have a question about padding when concatenating datasets or adding a column, and some nitpicks:

File: src/datasets/features/features.py (hunk 3 above)
Comment (mariosasko) [reply src/datasets/features/features.py:459]: Great catch!

File: src/datasets/arrow_dataset.py (hunk 1 above)
Comment (mariosasko) [reply src/datasets/arrow_dataset.py:3413]: Actually, I agree with you that this could lead to silent errors. Also, the code is definitely not the nicest :). So let's disable manual padding in `add_column`/`concatenate_datasets` for now, and only allow padding introduced by `pa.concat_tables(..., promote=True)`!

**Comment by mariosasko** [comment]
@lhoestq I addressed your comments, added tests, did some refactoring to make the implementation cleaner and added support for `None` values in `map` transforms when the feature type is `ArrayXD` (previously, I only implemented `None` decoding).

My only concern is that during decoding `ArrayXD` arrays with `None` values will be auto-casted to `float64` to allow `np.nan` insertion and this might be unexpected if `dtype` is not `float`, so one option would be to allow `None` values only if the storage type is `float32` or `float64`. Let me know WDYT would be the most consistent behavior here.

**Comment by lhoestq** [comment]
Cool ! :D
> My only concern is that during decoding ArrayXD arrays with None values will be auto-casted to float64 to allow np.nan insertion and this might be unexpected if dtype is not float, so one option would be to allow None values only if the storage type is float32 or float64. Let me know WDYT would be the most consistent behavior here.

Yes that makes sense to only fill with nan if the type is compatible

**Comment by mariosasko** [comment]
After some more experimenting, I think we can keep auto-cast to float because PyArrow also does it:
```python
import pyarrow as pa
arr = pa.array([1, 2, 3, 4, None], type=pa.int32()).to_numpy(zero_copy_only=False) # None present - int32 -> float64
assert arr.dtype == np.float64
```
Additional changes:
* fixes a bug in the `_is_zero_copy_only` implementation for the ArraXD types. Previously, `_is_zero_copy_only` would always return False for these types. Still have to see if it's possible to optimize copying of the non-extension types (`Sequence`, ...), but I plan to work on that in a separate PR.
* https://github.com/huggingface/datasets/pull/2891 introduced a bug where the dtype of `ArrayXD` wouldn't be preserved due to `to_pylist` call in NumPy Formatter (`np.array(np.array(..).tolist())` doesn't necessarily preserve dtype of the initial array), so I'm also fixing that.

src/datasets/arrow_dataset.py
```diff
@@ -3692,21 +3746,28 @@ def add_item(self, item: dict, new_fingerprint: str):
...
-        item_table = item_table.cast(schema)
-        # Concatenate tables
-        table = concat_tables([self._data, item_table])
+        item_table = InMemoryTable.from_pydict({k: [v] for k, v in item.items()})
+        # We don't call _check_if_features_can_be_aligned here so this cast is "unsafe"
+        dset_features, item_features = _align_features([self.features, Features.from_arrow_schema(item_table.schema)])
+        # Cast and concatenate tables
+        table = concat_tables(
+            [
+                self._data.cast(pa.schema(dset_features.type)) if self.features != dset_features else self._data,
```
lhoestq, line 3755: Why is casting needed here ? I guess this is to re-order the columns if needed ?
Feel free to add a comment in the code since we may forget that this is important x)

src/datasets/arrow_dataset.py
```diff
@@ -3778,10 +3844,25 @@ def concatenate_datasets(
...
+    else:
+        # Return first dataset if all datasets are empty
+        return dsets[0]
+
+    # Perform checks (and a potentional cast if axis=0)
+    if axis == 0:
+        _check_if_features_can_be_aligned([dset.features for dset in dsets])
+        aligned_features = _align_features([dset.features for dset in dsets])
+        dsets = [
+            dset.cast(features) if dset.features != features else dset
```
lhoestq, line 3859: `dset.cast` is costly since it calls `map`. Is the issue just about reordering the features ?
If so maybe we could have an additional (private) method that does that instead of calling `map` ?

src/datasets/formatting/formatting.py
```diff
@@ -154,26 +159,44 @@ def extract_batch(self, pa_table: pa.Table) -> dict:
...
-            # cast to list of arrays or we end up with a np.array with dtype object
-            # for ArrayExtensionArray call py_list directly to support dynamic dimensions
             if isinstance(pa_array.type, _ArrayXDExtensionType):
-                array: List = pa_array.to_pylist()
+                # don't call to_pylist() to preserve dtype of the fixed-size array
+                zero_copy_only = _is_zero_copy_only(pa_array.type.storage_dtype, unnest=True)
+                if pa_array.type.shape[0] is None:
+                    array: List = pa_array.to_list_of_numpy(zero_copy_only=zero_copy_only)
+                else:
+                    array: List = pa_array.to_numpy(zero_copy_only=zero_copy_only)
```
lhoestq, line 190: If I understand correctly the idea is to always return numpy arrays, even for the default formatting of a dataset.

I think this is a (very welcomed) breaking change, feel free to mention it in this PR's description if it's not already

lhoestq (commented):
Very nice ! My final comments:

src/datasets/arrow_dataset.py
mariosasko: Nice catch! Yes, I don't think this cast is needed. Let's remove it and see what the CI says :).

src/datasets/formatting/formatting.py
mariosasko: This is explained in the comment: `# don't call to_pylist() to preserve dtype of the fixed-size array`.
This change is needed because previously we would call `pa_array.to_pylist()`, which in turn calls `pa_array.to_numpy(zero_copy_only=zero_copy_only).tolist()` for arrays where the first dimension is not dynamic, and the `tolist()` call is dangerous because after that call we can't restore the initial dtype properly. So what I did here is I copied `the ArrayExtensionArray.to_pylist` code and removed the `to_list` call for arrays without dynamic dims. (So this should not be a breaking change.)

lhoestq: Oh yes you're right ! For some reason I thought it would affect the default python formatting x)

lhoestq (approved):
LGTM ! Thank you :)

lhoestq:
The CI fail for windows is unrelated to this PR, merging

PR #3196, opened 2021-11-02
QOL improvements: auto-flatten_indices and desc in map calls

### LINKED ISSUE #3040: [save_to_disk] Using `select()` followed by `save_to_disk` saves complete dataset making it hard to create dummy dataset
## Describe the bug

When only keeping a dummy size of a dataset (say the first 100 samples), and then saving it  to disk to upload it in the following to the hub for easy demo/use - not just the small dataset is saved but the whole dataset with an indices file. The problem with this is that the dataset is still very big.

## Steps to reproduce the bug

E.g. run the following:

```python
from datasets import load_dataset, save_to_disk

nlp = load_dataset("glue", "mnli", split="train")
nlp.save_to_disk("full")

nlp = nlp.select(range(100))
nlp.save_to_disk("dummy")
```

Now one can see that both `"dummy"` and `"full"` have the same size. This shouldn't be the case IMO.

## Expected results

IMO `"dummy"` should be much smaller so that one can easily play around with the dataset on the hub.

## Actual results
Specify the actual results or traceback.

## Environment info

- `datasets` version: 1.12.2.dev0
- Platform: Linux-5.11.0-34-generic-x86_64-with-glibc2.10
- Python version: 3.8.5
- PyArrow version: 5.0.0

mariosasko:
This PR:
* automatically calls `flatten_indices` where needed: in `unique` and `save_to_disk` to avoid saving the indices file
* adds descriptions to the map calls

Fix #3040

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -1166,7 +1166,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
             lambda batch: {column: dst_feat.str2int(batch)},
             input_columns=column,
             batched=True,
-            desc="Casting to the labels",
+            desc="Casting to class labels",
         )
         dset = concatenate_datasets([self.remove_columns([column]), dset], axis=1)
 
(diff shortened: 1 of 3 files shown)


lhoestq (approved):
Amazing ! Thanks also for adding all the `desc` parameters

PR #3203, opened 2021-11-03
Updated URL for download

MalteHB:
It seems that DaNLP has updated their download URLs and it therefore also needs to be updated in here...


lhoestq (approved):
Thanks for updating the URL :)

I also updated the metadata json file of the dataset with the new URL

lhoestq:
Actually it looks like the old URL is still working, and it's also the one that is mentioned in https://github.com/alexandrainst/danlp/blob/master/docs/docs/datasets.md

What makes you think we should use the new URL ?

MalteHB:
I might have jumped to conclusions a bit too fast here... 

I was working in Google Colab and got an error that it was unable to use the URL. I then forked the project, updated the URL, ran it locally and it worked. I therefore assumed that my URL update fixed the issue, however, I see now that it might rather be a Google Colab issue... 

Still - this seems to be the official URL for downloading the dataset, and I think that it will be most beneficial to use. :-)

lhoestq:
It looks like they're using these new urls for their new datasets. Maybe let's change to the new URL in case the old one stops working at one point. Thanks

PR #3222, opened 2021-11-05
Add docs for audio processing

stevhliu:
This PR adds documentation for the `Audio` feature. It describes:

- The difference between loading `path` and `audio`, as well as use-cases/best practices for each of them.
- Resampling audio files with `cast_column`, and then calling `ds[0]["audio"]` to automatically decode and resample to the desired sampling rate.
- Resampling with `map`.

Preview [here](https://52969-250213286-gh.circle-artifacts.com/0/docs/_build/html/audio_process.html), let me know if I'm missing anything!


docs/source/audio_process.rst
```diff
@@ -0,0 +1,97 @@
...
+
+``audio`` is the actual audio file that is loaded and resampled on-the-fly upon calling it.
+
+.. code::
+
+    >>> common_voice[0]["audio"]
+    {'array': array([ 0.0000000e+00,  0.0000000e+00,  0.0000000e+00, ...,
+        -8.8930130e-05, -3.8027763e-05, -2.9146671e-05], dtype=float32),
+    'path': '/root/.cache/huggingface/datasets/downloads/extracted/05be0c29807a73c9b099873d2f5975dae6d05e9f7d577458a2466ecb9a2b0c6b/cv-corpus-6.1-2020-12-11/tr/clips/common_voice_tr_21921195.mp3',
+    'sampling_rate': 48000}
```
lhoestq, line 62: I would first mention the "audio" column, and then the "path" column. Indeed the "path" column isn't going to be used most of the time, whereas "audio" really is the column that contain the relevant data

docs/source/audio_process.rst
```diff
@@ -0,0 +1,97 @@
...
+The next time you load the audio file, the :class:`datasets.Audio` feature will load and resample it to 16kHz:
+
+    >>> common_voice_train[0]["audio"]
+    {'array': array([ 0.0000000e+00,  0.0000000e+00,  0.0000000e+00, ...,
+        -7.4556941e-05, -1.4621433e-05, -5.7861507e-05], dtype=float32),
+    'path': '/root/.cache/huggingface/datasets/downloads/extracted/05be0c29807a73c9b099873d2f5975dae6d05e9f7d577458a2466ecb9a2b0c6b/cv-corpus-6.1-2020-12-11/tr/clips/common_voice_tr_21921195.mp3',
+    'sampling_rate': 16000}
+
+.. image:: /imgs/resample.gif
+   :align: center
```
lhoestq, line 84: O_O

docs/source/audio_process.rst
```diff
@@ -0,0 +1,97 @@
...
+
+Just like text datasets, you can apply a preprocessing function over an entire dataset with :func:`datasets.Dataset.map`, which is useful for resampling all of your audio data at once. Just make sure to include the ``audio`` key when you call :func:`datasets.Dataset.map` so that you are actually resampling the audio data:
+
+.. code-block::
+
+    >>> def prepare_dataset(batch):
+    ...     audio = batch["audio"]
+    ...     batch["input_values"] = processor(audio["array"], sampling_rate=audio["sampling_rate"]).input_values[0]
```
lhoestq, line 92: Maybe add a line to explain how the processor is created ?

lhoestq (commented):
Cool thanks :)
also pinging @anton-l @patrickvonplaten @albertvillanova

docs/source/audio_process.rst
stevhliu: Changed gif slightly to make it more clear that the resampling happens after `ds[0]["audio"]`. Let me know which one you prefer, or if there is anything else I can add! :)

lhoestq: I liked the symmetry of the first one, and having everything on one level makes it look clear that it's transforming the data like a pipeline. Maybe if you have both waves at the same level as `ds[0]["audio"]` it's clearer that the transformation is done at this step ?

stevhliu: Good idea!

lhoestq:
Nice ! love it this way. I guess you can set this PR to "ready for review" ?

docs/source/audio_process.rst
```diff
@@ -0,0 +1,114 @@
...
+Users should also install `torchaudio <https://pytorch.org/audio/stable/index.html>`_ and `librosa <https://librosa.org/doc/latest/index.html>`_, two common libraries used by 🤗 Datasets for handling audio data.
+
+.. code::
+
+    >>> pip install librosa
+    >>> pip install torchaudio
+
+.. important::
+
+    torchaudio's ``sox_io`` `backend <https://pytorch.org/audio/stable/backend.html#>`_ supports decoding ``mp3`` files. Unfortunately, the ``sox_io`` backend is only available on Linux/macOS, and is unsupported by Windows.
```
patrickvonplaten, line 28:
```suggestion
    torchaudio's ``sox_io`` `backend <https://pytorch.org/audio/stable/backend.html#>`_ supports decoding ``mp3`` files. Unfortunately, the ``sox_io`` backend is only available on Linux/macOS, and is not supported by Windows.
```

docs/source/audio_process.rst
```diff
@@ -0,0 +1,114 @@
...
+
+.. code-block::
+
+    >>> from datasets import load_dataset, load_metric, Audio
+    >>> common_voice = load_dataset("common_voice", "tr", split="train")
+
+Audio datasets
+--------------
+
+Audio datasets commonly have an ``audio`` and ``path`` column.
```
patrickvonplaten, line 40: Some important datasets like librispeech I have "audio" and "file" columns
```suggestion
Audio datasets commonly have an ``audio`` and ``path`` or ``file`` column.
```

docs/source/audio_process.rst
```diff
@@ -0,0 +1,114 @@
...
+
+    >>> common_voice[0]["audio"]
+    {'array': array([ 0.0000000e+00,  0.0000000e+00,  0.0000000e+00, ...,
+        -8.8930130e-05, -3.8027763e-05, -2.9146671e-05], dtype=float32),
+    'path': '/root/.cache/huggingface/datasets/downloads/extracted/05be0c29807a73c9b099873d2f5975dae6d05e9f7d577458a2466ecb9a2b0c6b/cv-corpus-6.1-2020-12-11/tr/clips/common_voice_tr_21921195.mp3',
+    'sampling_rate': 48000}
+
+When you access an audio file, it is automatically decoded and resampled. Generally, you should query an audio file like: ``common_voice[0]["audio"]``. If you query an audio file with ``common_voice["audio"][0]`` instead, **all** the audio files in your dataset will be decoded and resampled. This process can take a long time if you have a large dataset.
+
+``path`` is an absolute path to an audio file.
```
patrickvonplaten, line 54:
```suggestion
``path`` or ``file`` is an absolute path to an audio file.
```

docs/source/audio_process.rst
```diff
@@ -0,0 +1,114 @@
...
+    'path': '/root/.cache/huggingface/datasets/downloads/extracted/05be0c29807a73c9b099873d2f5975dae6d05e9f7d577458a2466ecb9a2b0c6b/cv-corpus-6.1-2020-12-11/tr/clips/common_voice_tr_21921195.mp3',
+    'sampling_rate': 16000}
+
+.. image:: /imgs/resample.gif
+   :align: center
+
+``Map``
+-------
+
+Just like text datasets, you can apply a preprocessing function over an entire dataset with :func:`datasets.Dataset.map`, which is useful for preprocessing all of your audio data at once. Create a convenient ``processor`` object that contains:
```
patrickvonplaten, line 90:
```suggestion
Just like text datasets, you can apply a preprocessing function over an entire dataset with :func:`datasets.Dataset.map`, which is useful for preprocessing all of your audio data at once.
For a [speech recognition model of your choice](https://huggingface.co/models?pipeline_tag=automatic-speech-recognition&sort=downloads), *e.g.* `"facebook/wav2vec2-large-xlsr-53"`, you can load a ``processor`` object that contains:
```

patrickvonplaten (commented):
That's a great document - thanks a lot for putting all of this together!

I left some tips on how the preprocessing `transformers` code example could be a bit simplified. 

In short 99% of the use cases when Audio `datasets` is used for `transformers` is either:

a) A pretrained speech model is fine-tuned

or:

b) A fine-tuned speech model is evaluated / used in inference

For both a) and b) the `feature_extractor` is always defined. So we should always advocate to use `AutoFeatureExtractor.from_pretrained(...)` here IMO.

For a) the `tokenizer` is **not** defined and has to be created as described in the docs currently. For b) the `tokenizer` is also defined so that one can directly use `Wav2Vec2Processor.from_pretrained(...)` 

Hope that helps a bit :-)

lhoestq (approved):
Looks all good to me now :)
Let us know if you have more comments or if it's ready to merge

anton-l (approved):
LGTM, great reference the `transformers` examples!

lhoestq:
I guess we can merge this one now :)

PR #3225, opened 2021-11-06
Update tatoeba to v2021-07-22

KoichiYasuoka:
Tatoeba's latest version is v2021-07-22


KoichiYasuoka:
How about this? @lhoestq @abhishekkrthakur

lhoestq:
Hi ! I think it would be nice if people could still be able to load the old version.
Maybe this can be a parameter ? For example to load the old version they could do
```python
load_dataset("tatoeba", lang1="en", lang2="mr", date="v2020-11-09")
```

If it sounds good to you, we can add this parameter to the TatoebaConfig:
```python
class TatoebaConfig(datasets.BuilderConfig):
    def __init__(self, *args, lang1=None, lang2=None, date="v2021-07-22", **kwargs):
    self.date = date
```
and then pass the date to the URL
```python
_BASE_URL = "https://object.pouta.csc.fi/OPUS-Tatoeba/{}/moses/{}-{}.txt.zip"
```
```python
        def _base_url(lang1, lang2, date):
            return _BASE_URL.format(date, lang1, lang2)
```

What do you think ?

KoichiYasuoka:
`_DATE = "v" + "-".join(s.zfill(2) for s in _VERSION.split("."))` seems rather tricky but works well. How about this? @lhoestq

datasets/tatoeba/tatoeba.py
```diff
@@ -95,10 +97,10 @@ def _info(self):
         )
 
     def _split_generators(self, dl_manager):
-        def _base_url(lang1, lang2):
-            return _BASE_URL.format(lang1, lang2)
+        def _base_url(lang1, lang2, date):
+            return.format(date, lang1, lang2)
```
lhoestq, line 101:
```suggestion
            return _BASE_URL.format(date, lang1, lang2)
```

lhoestq (approved):
Looks all good thanks !

I just renamed the parent directory of the dummy data to match the new version name, and mentioned how to change the date in the dataset card.

lhoestq:
The CI is only failing because of the missing sections in the dataset card, and because of an issue with the CER metric that is unrelated to this PR

PR #3244, opened 2021-11-09
Fix filter method for batched=True

thomasw21:


tests/test_arrow_dataset.py
```diff
@@ -1184,6 +1184,15 @@ def test_filter_with_indices_mapping(self, in_memory):
...
                         self.assertListEqual(dset["col"], [1])
 
+    def test_filter_batched(self, in_memory):
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            dset = Dataset.from_dict({"col": [0, 1, 2]})
+            with self._to(in_memory, tmp_dir, dset) as dset:
+                with dset.filter(lambda x: x["col"] > 0, batched=True) as dset:
+                    self.assertListEqual(dset["col"], [1, 2])
+                    with dset.filter(lambda x: x["col"] < 2, batched=True) as dset:
+                        self.assertListEqual(dset["col"], [1])
```
lhoestq, line 1194:
```suggestion
                with dset.filter(lambda x: [i > 0 for i in x["col"]], batched=True) as dset:
                    self.assertListEqual(dset["col"], [1, 2])
                    with dset.filter(lambda x: [i < 2 for i in x["col"]], batched=True) as dset:
                        self.assertListEqual(dset["col"], [1])
```

lhoestq (approved):
Thanks for fixing this and updating the docs !

PR #3250, opened 2021-11-11
Add ETHICS dataset

ssss1029:
This PR adds the ETHICS dataset, including all 5 sub-datasets.
From https://arxiv.org/abs/2008.02275


datasets/ethics/ethics.py
```diff
@@ -0,0 +1,169 @@
+"""ETHICS: Justice Dataset."""
+
+
+import csv
+import os
+
+import datasets
```
BithomeFX, line 7:
```suggestion
import datasets```suggestion
import datasets
```

```

**Review by BithomeFX — APPROVED** [review]

File: datasets/ethics/README.md [hunk 2]
```diff
@@ -0,0 +1,210 @@
+---
+annotations_creators:
+- crowdsourced
+- expert-generated
+language_creators:
+- crowdsourced
+- expert-generated
+languages:
+- en-US
```
lhoestq, line 9:
```suggestion
- en
```

datasets/ethics/README.md
```diff
@@ -0,0 +1,210 @@
...
+    and Collin Burns
+    and Steven Basart
+    and Andrew Critch
+    and Jerry Li
+    and Dawn Song
+    and Jacob Steinhardt},
+  journal={arXiv preprint arXiv:2008.02275},
+  year={2021}
+}
+```
```
lhoestq, line 210:
````suggestion
```

### Contributions

Thanks to [@ ssss1029](https://github.com/ssss1029) for adding this dataset.
````

File: datasets/ethics/ethics.py [hunk 4]
```diff
@@ -0,0 +1,169 @@
...
+    ]
+
+    def _info(self):
+
+        features = {
+            "commonsense_morality": datasets.Features(
+                {
+                    "label": datasets.ClassLabel(2),
+                    "input": datasets.Value("string"),
+                    "is_short": datasets.ClassLabel(names=["True", "False"]),
```
lhoestq, line 78: I would invert the two, generally True is 1 and False is 0
```suggestion
                    "is_short": datasets.ClassLabel(names=["False", "True"]),
```

datasets/ethics/ethics.py
```diff
@@ -0,0 +1,169 @@
...
+        datasets.BuilderConfig(name="utilitarianism", version=VERSION),
+        datasets.BuilderConfig(name="virtue_ethics", version=VERSION),
+    ]
+
+    def _info(self):
+
+        features = {
+            "commonsense_morality": datasets.Features(
+                {
+                    "label": datasets.ClassLabel(2),
```
lhoestq, line 76: It would be nice to have names for the positive and negative class. This is useful to avoid any ambiguity whether the positive or negative class is the problematic one for example.

Do you have ideas of class names we could use here and for the other configurations ?

datasets/ethics/README.md
```diff
@@ -0,0 +1,210 @@
...
+
+# Dataset Card for ETHICS
+
+## Table of Contents
+- [Dataset Description](#dataset-description)
+  - [Dataset Summary](#dataset-summary)
+  - [Supported Tasks](#supported-tasks-and-leaderboards)
+  - [Languages](#languages)
+- [Dataset Structure](#dataset-structure)
+  - [Data Fields and Instances](#data-instances)
```
lhoestq, line 34: We try to use the same dataset card structure for all the cards, could you re-add this please ?
```suggestion
  - [Data Instances](#data-instances)
  - [Data Fields](#data-fields)
```

The `Data Fields` section can just list the name of each field (scenario, excuse, etc.) and give a short description + say the type (integer, string)

lhoestq (commented):
Awesome thank you ! Good job with the dataset script and the dataset card, both are really good

I just left some minor comments/suggestions, let me know what you think:

albertvillanova:
Thanks for your contribution, @ssss1029. Are you still interested in adding this dataset?

PR #3252, opened 2021-11-11
Fix failing CER metric test in CI after update

mariosasko:
Fixes the [failing CER metric test](https://app.circleci.com/pipelines/github/huggingface/datasets/8644/workflows/79816553-fa2f-4756-b022-d5937f00bf7b/jobs/53298) in CI by adding support for `jiwer==2.3.0`, which was released yesterday. Also, I verified that all the tests in `metrics/cer/test_cer.py` pass after the change, so the results should be the same irrespective of the `jiwer` version.

### DIFF
--- a/metrics/cer/cer.py
+++ b/metrics/cer/cer.py
@@ -18,30 +18,50 @@ from typing import List
 
 import jiwer
 import jiwer.transforms as tr
+from packaging import version
 
 import datasets
+from datasets.config import PY_VERSION
 
 
-class SentencesToListOfCharacters(tr.AbstractTransform):
-    def process_string(self, s: str):
-        return list(s)
-
-    def process_list(self, inp: List[str]):
-        chars = []
-
-        for sentence in inp:
-            chars.extend(self.process_string(sentence))
-
-        return chars
+if PY_VERSION < version.parse("3.8"):
+    import importlib_metadata
+else:
+    import importlib.metadata as importlib_metadata
 
 
-cer_transform = tr.Compose(
-    [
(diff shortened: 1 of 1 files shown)


metrics/cer/cer.py
```diff
@@ -18,30 +18,50 @@
...
+            chars = []
+            for sent_idx, sentence in enumerate(inp):
+                chars.extend(self.process_string(sentence))
+                if self.sentence_delimiter is not None and self.sentence_delimiter != "" and sent_idx < len(inp) - 1:
+                    chars.append(self.sentence_delimiter)
+            return chars
+
+    cer_transform = tr.Compose(
+        [tr.RemoveMultipleSpaces(), tr.Strip(), SentencesToListOfCharacters(SENTENCE_DELIMITER)]
+    )
```
lhoestq, line 55: why can't we use the old transform here ?
```python
cer_transform = tr.Compose(
    [
        tr.RemoveMultipleSpaces(),
        tr.Strip(),
        SentencesToListOfCharacters(),
    ]
)
```

lhoestq (approved):
and thanks for the fix :)

metrics/cer/cer.py
mariosasko: This transform in the current state does the same thing as the old one.

However, in `jiwer 2.3.0`, we can use `tr.ReduceToSingleSentence` in place of our transform, which by default uses a single space as a delimiter between sentences. And this change allows us to pass the same delimiter to both of them.

Additionally, we can easily change the sentence delimiter in the future this way (if we want to align with `jiwer` for instance).

PR #3254, opened 2021-11-11
Update xcopa dataset (fix checksum issues + add translated data)

mariosasko:
This PR updates the checksums (as reported [here](https://discuss.huggingface.co/t/how-to-load-dataset-locally/11601/2)) of the `xcopa` dataset. Additionally, it adds new configs that hold the translated data of the original set of configs. This data was not available at the time of adding this dataset to the lib.


mariosasko:
The CI failures are unrelated to the changes (missing fields in the readme and the CER metric error fixed in #3252).

lhoestq (approved):
Thanks !

Issue #3171, opened 2021-10-27
Raise exceptions instead of using assertions for control flow


mariosasko:
Motivated by https://github.com/huggingface/transformers/issues/12789 in Transformers, one welcoming change would be replacing assertions with proper exceptions. The only type of assertions we should keep are those used as sanity checks.

Currently, there is a total of 87 files with the `assert` statements (located under `datasets` and `src/datasets`), so when working on this, to manage the PR size, only modify 4-5 files at most before submitting a PR.

manisnesan:
Adding the remaining tasks for this issue to help new code contributors.  
$ cd src/datasets && ack assert -lc 
- [ ] splits.py:4
- [ ] commands/convert.py:1
- [ ] arrow_reader.py:3
- [ ] load.py:7
- [ ] utils/patching.py:1
- [ ] utils/py_utils.py:2
- [ ] features/features.py:9
- [ ] arrow_writer.py:7
- [ ] data_files.py:1
- [ ] search.py:6
- [ ] table.py:1
- [ ] metric.py:3
- [ ] tasks/image_classification.py:1
- [ ] arrow_dataset.py:17
- [ ] fingerprint.py:6
- [ ] io/json.py:1
- [ ] io/csv.py:1

Ishan-Kumar2:
Hi all,
I am interested in taking up `fingerprint.py`, `search.py`, `arrow_writer.py` and `metric.py`. Will raise a PR soon!

manisnesan:
Let me look into `arrow_dataset.py`, `table.py`, `data_files.py` & `features.py`

manisnesan:
All the tasks are completed for this issue. This can be closed.

PR #3256, opened 2021-11-12
asserts replaced by exception for text classification task with test.

manisnesan:
I have replaced only a single assert in text_classification.py along with a unit test to verify an exception is raised based on https://github.com/huggingface/datasets/issues/3171 . 

I would like to first understand the code contribution workflow. So keeping the change to a single file rather than making too many changes. Once this gets approved, I will look into the rest. 

Thanks.

### DIFF
--- a/tests/test_tasks.py
+++ b/tests/test_tasks.py
@@ -40,6 +40,13 @@ class TextClassificationTest(TestCase):
         self.assertEqual(input_schema, task.input_schema)
         self.assertEqual(label_schema, task.label_schema)
 
+    def test_value_error_unique_labels(self):
+        with self.assertRaises(ValueError):
+            # Add duplicate labels
+            self.labels.append("neg")
+            task = TextClassification(text_column="input_text", label_column="input_label", labels=self.labels)
+            self.assertEqual("text-classification", task.task)
+
 
 class QuestionAnsweringTest(TestCase):
     def test_column_mapping(self):
--- a/tests/test_tasks.py
+++ b/tests/test_tasks.py
@@ -43,8 +43,8 @@ class TextClassificationTest(TestCase):
     def test_value_error_unique_labels(self):
         with self.assertRaises(ValueError):
             # Add duplicate labels
-            self.labels.append("neg")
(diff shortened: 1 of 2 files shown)


tests/test_tasks.py
```diff
@@ -40,6 +40,13 @@ def test_from_dict(self):
         self.assertEqual(input_schema, task.input_schema)
         self.assertEqual(label_schema, task.label_schema)
 
+    def test_value_error_unique_labels(self):
+        with self.assertRaises(ValueError):
+            # Add duplicate labels
+            self.labels.append("neg")
+            task = TextClassification(text_column="input_text", label_column="input_label", labels=self.labels)
```
lhoestq, line 47: Here you add "neg" to self.labels which is used in the other tests `test_column_mapping`, `test_from_dict`.
To not interfere with the other tests you can just copy the labels list:

```suggestion
            labels = self.labels + self.labels[:1]
            task = TextClassification(text_column="input_text", label_column="input_label", labels=labels)
```

mariosasko (approved):
Thanks for working on this! Just one suggestion:

lhoestq (approved):
Thanks for the contribution and for taking care of writing a test :)

I just did a minor modification in the test:

tests/test_tasks.py
mariosasko: Ups forgot to add a comment to the review :). 

Let's not mutate `self.labels` in-place (IMO not a good style) even though it works because `TestCase.setUp` is called before each fixture and not once per class.

```suggestion
            labels = self.labels + ["neg"]
            task = TextClassification(text_column="input_text", label_column="input_label", labels=labels)
```

mariosasko: Yes, that's exactly what I meant! More readable and avoids unintended behavior.

lhoestq:
Haha it looks like you got the chance of being reviewed twice at the same time and got the same suggestion twice x)
Anyway it's all good now so we can merge !

manisnesan:
Thanks for the feedback.

PR #3262, opened 2021-11-12
asserts replaced with exception for image classification task, csv, json

manisnesan:
Fixes for csv, json in io module and image_classification task with tests referenced in https://github.com/huggingface/datasets/issues/3171


lhoestq (approved):
thanks !

PR #3267, opened 2021-11-13
Replacing .format() and % by f-strings

### LINKED ISSUE #3257: Use f-strings for string formatting
f-strings offer better readability/performance than `str.format` and `%`, so we should use them in all places in our codebase unless there is good reason to keep the older syntax.

> **NOTE FOR CONTRIBUTORS**: To avoid large PRs and possible merge conflicts, do 1-3 modules per PR. Also, feel free to ignore the files located under `datasets/*`.

Mehdi2402:
**Fix #3257**

Replaced _.format()_ and _%_ by f-strings in the following modules : 
- [x] **tests**
- [x] **metrics**
- [x] **benchmarks**
- [x] **utils**
- [x] **templates**

Will follow in the next PR the modules left : 
- [ ] **src**

Module **datasets** will not be edited as asked by @mariosasko 

PS : black and isort applied to files


lhoestq:
Hi ! It looks like most of your changes are just `black` changes. All those changes are not necessary. In particular if you want to use `black`, please use the `make style` command instead. It runs `black` with additional parameters and you shouldn't end up with that many changes

Feel free to open a new PR that doesn't include all the unnecessary `black` changes that you have on your branch :)

Mehdi2402:
> Hi ! It looks like most of your changes are just `black` changes. All those changes are not necessary. In particular if you want to use `black`, please use the `make style` command instead. It runs `black` with additional parameters and you shouldn't end up with that many changes
> 
> Feel free to open a new PR that doesn't include all the unnecessary `black` changes that you have on your branch :)

Thank you for your answer :) , I will open a new PR with the correct changes.

Mehdi2402:
Hi @lhoestq, I submitted 3 commits in a new PR (#3277) where I did not apply black.

I can apply the ```make style``` command if asked.

Issue #3188, opened 2021-11-01
conll2002 issues


BramVanroy:
**Link:** https://huggingface.co/datasets/conll2002

The dataset viewer throws a server error when trying to preview the dataset. 

```
Message:       Extraction protocol 'train' for file at 'https://raw.githubusercontent.com/teropa/nlp/master/resources/corpora/conll2002/esp.train' is not implemented yet
```

In addition, the "point of contact" has encoding issues and does not work when clicked.

Am I the one who added this dataset ? No, @lhoestq  did

lhoestq:
Hi ! Thanks for reporting :)

This is related to https://github.com/huggingface/datasets/issues/2742, I'm working on it. It should fix the viewer for around 80 datasets.

BramVanroy:
Ah, hadn't seen that sorry.

The scrambled "point of contact" is a separate issue though, I think.

BramVanroy:
@lhoestq The "point of contact" is still an issue.

lhoestq:
It will be fixed in https://github.com/huggingface/datasets/pull/3274, thanks

PR #3274, opened 2021-11-15
Fix some contact information formats

lhoestq:
As reported in https://github.com/huggingface/datasets/issues/3188 some contact information are not displayed correctly.
This PR fixes this for CoNLL-2002 and some other datasets with the same issue


lhoestq:
The CI fail are caused by some missing sections or tags, which is unrelated to this PR. Merging !

PR #3277, opened 2021-11-15
f-string formatting

### LINKED ISSUE #3257: Use f-strings for string formatting

Mehdi2402:
**Fix #3257**

Replaced _.format()_ and _%_ by f-strings in the following modules : 
- [x] **tests**
- [x] **metrics**
- [x] **benchmarks**
- [x] **utils**
- [x] **templates**
- [x] **src/Datasets/\*.py**

Modules in **_src/Datasets/_**: 
- [x] **commands**
- [x] **features**
- [x] **formatting**
- [x] **io**
- [x] **tasks**
- [x] **utils**

Module **datasets** will not be edited as asked by @mariosasko 

-A correction of the first PR (#3267)-


Mehdi2402:
Hello @lhoestq, ```make style``` is applied as asked. :)

lhoestq (approved):
Alright it's all good now, thanks :)

Merging !

Carlosbogo (commented):
At line 30, shouldn't the new f-string also reference `old_val`  instead of `new_val`?

PR #3278, opened 2021-11-15
Proposed update to the documentation for WER

wooters:
I wanted to submit a minor update to the description of WER for your consideration. 

Because of the possibility of insertions, the numerator in the WER formula can be larger than N, so the value of WER can be greater than 1.0:

```
>>> from datasets import load_metric
>>> metric = load_metric("wer")
>>> metric.compute(predictions=["hello how are you"], references=["hello"])
3.0
```

and similarly from the underlying jiwer module's `wer` function:

```
>>> from jiwer import wer
>>> wer("hello", "hello how are you")
3.0
```

### DIFF
--- a/metrics/wer/wer.py
+++ b/metrics/wer/wer.py
@@ -48,7 +48,7 @@ I is the number of insertions,
 C is the number of correct words,
 N is the number of words in the reference (N=S+D+C).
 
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
You're right, thanks for the correction :)

PR #3279, opened 2021-11-16
Minor Typo Fix - Precision to Recall

SebastinSanty:


lhoestq (approved):
Good catch ! Thanks for the fix

PR #3302, opened 2021-11-19
fix old_val typo in f-string

Mehdi2402:
This PR is to correct a typo in #3277  that @Carlosbogo revieled in a comment.

Related closed issue : #3257 

Sorry about that 😅.


lhoestq (approved):
Oh indeed good catch ! thanks :)

PR #3305, opened 2021-11-20
asserts replaced with exception for ``fingerprint.py``, ``search.py``, ``arrow_writer.py`` and ``metric.py``

Ishan-Kumar2:
Addresses #3171 
Fixes exception for ``fingerprint.py``, ``search.py``, ``arrow_writer.py`` and ``metric.py`` and modified tests


mariosasko (approved):
@Ishan-Kumar2 Thanks for working on this! I fixed two checks that were not negated correctly, but other than that it looks good.

lhoestq (approved):
It looks all good thanks !

PR #3307, opened 2021-11-20
Add IndoNLI dataset

afaji:
This PR adds IndoNLI dataset, from https://aclanthology.org/2021.emnlp-main.821/


lhoestq (commented):
That's amazing thank you !
Good job with the dataset script and the dataset card for its comprehensiveness :)

My only comment is about the names of the labels. In my opinion it would be more practical to have the full names rather than the letters c, n or e for consistency and readability.

Once they're renamed you can regenerate the datasets_infos.json file and we should be all good !

afaji:
@lhoestq thanks for the review! I've modified the labels to follow other NLI datasets.
Please review my change and let me know if I miss anything.

lhoestq (approved):
That's perfect thank you !

Merging now :)

PR #3309, opened 2021-11-21
fix: files counted twice in inferred structure

borisdayma:
Files were counted twice in a structure like:
```
my_dataset_local_path/
├── README.md
└── data/
    ├── train/
    │   ├── shard_0.csv
    │   ├── shard_1.csv
    │   ├── shard_2.csv
    │   └── shard_3.csv
    └── valid/
        ├── shard_0.csv
        └── shard_1.csv
```

The reason is that they were matching both `*train*/*` and `*train*/**/*`.

This PR fixes it. @lhoestq

### DIFF
--- a/src/datasets/data_files.py
+++ b/src/datasets/data_files.py
@@ -41,7 +41,7 @@ DEFAULT_PATTERNS_SPLIT_IN_DIR_NAME = {
 }
 
 DEFAULT_PATTERNS_ALL = {
-    str(Split.TRAIN): ["*"],
+    str(Split.TRAIN): ["**"],
 }
 
 ALL_SPLIT_PATTERNS = [SPLIT_PATTERN_SHARDED]
--- a/src/datasets/data_files.py
+++ b/src/datasets/data_files.py
@@ -154,13 +154,29 @@ def resolve_patterns_locally_or_by_urls(
     Resolve the paths and URLs of the data files from the patterns passed by the user.
     URLs are just returned as is.
 
+    You can use patterns to resolve multiple local files. Here are a few examples:
+    - *.csv to match all the CSV files at the first level
+    - **.csv to match all the CSV files at any level
+    - data/* to match all the files inside "data"
+    - data/** to match all the files inside "data" and its subdirectories
+
+    The patterns are resolved using the fsspec glob.
+    Here are some behaviors specific to fsspec glob that are different from glob.glob, Path.glob, Path.match or fnmatch:
+    - '*' matches only first level items
+    - '**' matches all items
+    - '**/*' matches all at least second level items
+
+    More generally:
+    - '*' matches any character except a forward-slash (to match just the file or directory name)
+    - '**' matches any character including a forward-slash /
+
     Examples:
 
         >>> import huggingface_hub
-        >>> from datasets.data_files import resolve_patterns_in_dataset_repository
-        >>> base_path = /Users/username/Desktop/hf/datasets
+        >>> from datasets.data_files import resolve_patterns_locally_or_by_urls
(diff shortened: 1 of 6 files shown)


borisdayma:
I see it creates some errors in the tests.

Another solution if needed is to add something like `data_files = list(set(data_files))` after [this line](https://github.com/huggingface/datasets/blob/8555197a3fe826e98bd0206c2d031c4488c53c5c/src/datasets/data_files.py#L511)

lhoestq:
Hi ! Thanks for the correction :)

Your change seems right, let me look at the errors and try to fix this

borisdayma:
Not sure if it's due to this change but I tested `load_dataset('dalle-mini/encoded-vqgan_imagenet_f16_16384', streaming=True)` and the `validation` set is empty.

lhoestq:
So indeed there was an issue with the patterns `*` and `**/*` that would return some files twice. This issue came from the fact that we were not using the right `glob`.

Indeed we were using `Path.rglob` for local files and `Path.match` for remote files. Since these two methods don't have the same behavior for such patterns, I decided to change that.

In particular, we now use `glob.glob` (same as `fsspec` glob) as a reference for data files resolution from patterns. This is the same as dask for example.

/!\ Here are some behaviors specific to `glob.glob` that are different from Path.glob, Path.match or fnmatch:
- '*' matches only first level files
- '**/*' matches only at least second level files

This way we have a consistent behavior with respect to other python data libraries and there's no overlap anymore between the two patterns.

Some implementations details:

To ensure that we have the same behavior for local files and for files in a remote dataset repository, I decided to use `fsspec` glob for both. This was made possible by implementing the `HfFileSystem` class as a `fsspec` filesystem.

I pushed those changes directly to your PR - I hope you don't mind. Please let me know if that solves your problem, and then we can merge !

lhoestq:
There's still an issue with fsspec's glob - I'll take a look this afternoon

src/datasets/data_files.py
```diff
@@ -38,12 +40,23 @@ class Url(str):
     str(Split.VALIDATION): ["*dev*/*", "*dev*/**/*", "*valid*/*", "*valid*/**/*"],
 }
 
+DEFAULT_PATTERNS_SPLIT_IN_SUBDIR_NAME = {
+    str(Split.TRAIN): ["**/*train*/*", "**/*train*/**/*"],
+    str(Split.TEST): ["**/*test*/*", "**/*test*/**/*", "**/*eval*/*", "**/*eval*/**/*"],
+    str(Split.VALIDATION): ["**/*dev*/*", "**/dev**/**/*", "**/*valid*/*", "**/*valid*/**/*"],
```
borisdayma, line 46: I think there's a typo in the "dev" one: "**/dev**/**/*"

borisdayma (commented):
Just a little typo. I'll test it out!

lhoestq:
To recap:
```
We use fsspec glob as a reference for data files resolution from patterns.
This is the same as dask for example.

/!\ Here are some behaviors specific to fsspec glob that are different from glob.glob, Path.glob, Path.match or fnmatch:
- '*' matches only first level items
- '**' matches all items
- '**/*' matches all at least second level items

More generally:
- `*`` matches any character except a forward-slash (to match just the file or directory name)
- `**`` matches any character including a forward-slash /
```

borisdayma:
lol Windows… Maybe `Pathlib` for the tests?

I tested streaming a repo and it worked perfectly now!

lhoestq (approved):
Alright it's all good now :)

The tests are all updated accordingly, globbing is now consistent locally and in dataset repos, and there's no duplicate files anymore :)

PR #3314, opened 2021-11-23
Adding arg to pass process rank to `map`

TevenLeScao:
This PR adds a `with_rank` argument to `map` that gives the user the possibility to pass the rank of each process to their function. This is mostly designed for multi-GPU map (each process can be sent to a different device thanks to the rank). I've also added tests. I'm putting the PR up so you can check the code, I'll add a multi-GPU example to the doc (+ write a bit in the doc for the new arg)

### DIFF
--- a/docs/source/process.rst
+++ b/docs/source/process.rst
@@ -312,6 +312,24 @@ You can also use :func:`datasets.Dataset.map` with indices if you set ``with_ind
     '4: PG & E Corp. shares jumped $ 1.63 or 8 percent to $ 21.03 on the New York Stock Exchange on Friday .'
    ]
 
+You can also use :func:`datasets.Dataset.map` with the rank of the process if you set ``with_rank=True``. This is analogous to `with_indices`. The `rank` argument in the mapped function goes after the `index` one if it is already present. The main use-case for it is to parallelize your computation across several GPUs. This requires setting `multiprocess.set_start_method("spawn")`, without which you will receive a CUDA error: `RuntimeError: Cannot re-initialize CUDA in forked subprocess. To use CUDA with multiprocessing, you must use the 'spawn' start method`.
+
+
+.. code-block::
+
+   >>> from multiprocess import set_start_method
+   >>> import torch
+   >>> import os
+   >>>
+   >>> set_start_method("spawn")
+   >>>
+   >>> def gpu_computation(example, rank):
+   >>>     os.environ["CUDA_VISIBLE_DEVICES"] = str(rank % torch.cuda.device_count())
+   >>>     # Your big GPU call goes here
+   >>>     return examples
+   >>>
+   >>> updated_dataset = dataset.map(gpu_computation, with_rank=True)
+
 Multiprocessing
 ^^^^^^^^^^^^^^^
 
--- a/docs/source/process.rst
+++ b/docs/source/process.rst
@@ -312,7 +312,7 @@ You can also use :func:`datasets.Dataset.map` with indices if you set ``with_ind
     '4: PG & E Corp. shares jumped $ 1.63 or 8 percent to $ 21.03 on the New York Stock Exchange on Friday .'
    ]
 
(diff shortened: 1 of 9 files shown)


docs/source/process.rst
```diff
@@ -312,6 +312,24 @@ You can also use :func:`datasets.Dataset.map` with indices if you set ``with_ind
     '4: PG & E Corp. shares jumped $ 1.63 or 8 percent to $ 21.03 on the New York Stock Exchange on Friday .'
    ]
 
+You can also use :func:`datasets.Dataset.map` with the rank of the process if you set ``with_rank=True``. This is analogous to `with_indices`. The `rank` argument in the mapped function goes after the `index` one if it is already present. The main use-case for it is to parallelize your computation across several GPUs. This requires setting `multiprocess.set_start_method("spawn")`, without which you will receive a CUDA error: `RuntimeError: Cannot re-initialize CUDA in forked subprocess. To use CUDA with multiprocessing, you must use the 'spawn' start method`.
```
lhoestq, line 315:
```suggestion
You can also use :func:`datasets.Dataset.map` with the rank of the process if you set ``with_rank=True``. This is analogous to ``with_indices``. The ``rank`` argument in the mapped function goes after the ``index`` one if it is already present. The main use-case for it is to parallelize your computation across several GPUs. This requires setting `multiprocess.set_start_method("spawn")`, without which you will receive a CUDA error: `RuntimeError: Cannot re-initialize CUDA in forked subprocess. To use CUDA with multiprocessing, you must use the 'spawn' start method`.
```

src/datasets/arrow_dataset.py
```diff
@@ -1911,14 +1912,16 @@ def map(
...
-                - `function(example: Union[Dict, Any], indices: int) -> Union[Dict, Any]` if `batched=False` and `with_indices=True`
-                - `function(batch: Union[Dict[List], List[Any]]) -> Union[Dict, Any]` if `batched=True` and `with_indices=False`
-                - `function(batch: Union[Dict[List], List[Any]], indices: List[int]) -> Union[Dict, Any]` if `batched=True` and `with_indices=True`
+                - `function(example: Union[Dict, Any]) -> Union[Dict, Any]` if `batched=False` and `with_indices=False` and `with_rank=False`
+                - `function(example: Union[Dict, Any], *extra_args: int) -> Union[Dict, Any]` if `batched=False` and `with_indices=True` and/or `with_rank=True` (one extra arg for each)
+                - `function(batch: Union[Dict[List], List[Any]]) -> Union[Dict, Any]` if `batched=True` and `with_indices=False` and `with_rank=False`
+                - `function(batch: Union[Dict[List], List[Any]], *extra_args: List[int]) -> Union[Dict, Any]` if `batched=True` and `with_indices=True` and/or `with_rank=True` (one extra arg for each)
```
lhoestq, line 1918:
```suggestion
                - `function(example: Union[Dict, Any]) -> Union[Dict, Any]` if `batched=False` and `with_indices=False` and `with_rank=False`
                - `function(example: Union[Dict, Any], *extra_args) -> Union[Dict, Any]` if `batched=False` and `with_indices=True` and/or `with_rank=True` (one extra arg for each)
                - `function(batch: Union[Dict[List], List[Any]]) -> Union[Dict, Any]` if `batched=True` and `with_indices=False` and `with_rank=False`
                - `function(batch: Union[Dict[List], List[Any]], *extra_args) -> Union[Dict, Any]` if `batched=True` and `with_indices=True` and/or `with_rank=True` (one extra arg for each)
```

src/datasets/arrow_dataset.py
```diff
@@ -2165,12 +2171,13 @@ def _map_single(
...
```
lhoestq, line 2177:
```suggestion
                - `function(example: Union[Dict, Any]) -> Union[Dict, Any]` if `batched=False` and `with_indices=False` and `with_rank=False`
                - `function(example: Union[Dict, Any], *extra_args) -> Union[Dict, Any]` if `batched=False` and `with_indices=True` and/or `with_rank=True` (one extra arg for each)
                - `function(batch: Union[Dict[List], List[Any]]) -> Union[Dict, Any]` if `batched=True` and `with_indices=False` and `with_rank=False`
                - `function(batch: Union[Dict[List], List[Any]], *extra_args) -> Union[Dict, Any]` if `batched=True` and `with_indices=True` and/or `with_rank=True` (one extra arg for each)
```

lhoestq (approved):
Thanks for adding this and for the docs :)

The duplicate commits are not an issue here because there're not so many of them and they are quite separate from your changes. Next time prefer using merge rather than rebase. Note that this is just an issue from github that shows changes that already happened...

PR #3319, opened 2021-11-24
Add push_to_hub docs

lhoestq:
Since #3098 it's now possible to upload a dataset on the Hub directly from python using the `push_to_hub` method.
I just added a section in the "Upload a dataset to the Hub" tutorial.

I kept the section quite simple but let me know if it sounds good to you @LysandreJik @stevhliu :)

### DIFF
--- a/docs/source/upload_dataset.rst
+++ b/docs/source/upload_dataset.rst
@@ -80,7 +80,15 @@ Your dataset can now be loaded by anyone in a single line of code!
 Upload from Python
 ------------------
 
-To upload a :class:`datasets.DatasetDict` on the Hugging face Hub in python, you can use the :func:`datasets.DatasetDict.push_to_hub` method:
+To upload a :class:`datasets.DatasetDict` on the Hugging Face Hub in Python, you can login and use the :func:`datasets.DatasetDict.push_to_hub` method:
+
+1. Login from the command line:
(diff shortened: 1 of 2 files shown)


docs/source/upload_dataset.rst
```diff
@@ -77,6 +77,19 @@ Your dataset can now be loaded by anyone in a single line of code!
    })
 
 
+Upload from Python
+------------------
+
+To upload a :class:`datasets.DatasetDict` on the Hugging face Hub in python, you can use the :func:`datasets.DatasetDict.push_to_hub` method:
```
mariosasko, line 83: One nit for consistency:

```suggestion
To upload a :class:`datasets.DatasetDict` on the Hugging Face Hub in Python, you can use the :func:`datasets.DatasetDict.push_to_hub` method:
```

stevhliu:
Looks good to me! :)

Maybe we can mention that users can also set the `private` argument if they want to keep their dataset private? It would lead nicely into the next section on Privacy.

lhoestq:
Thanks for your comments, I fixed the capitalization for consistency and added an passage to mention the `private` parameter and to have a nice transition to the Privacy section :)

I also added the login instruction that was missing before the user can actually upload a dataset.

PR #3321, opened 2021-11-25
Update URL of tatoeba subset of xtreme

### LINKED ISSUE #3320: Can't get tatoeba.rus dataset
## Describe the bug
It gives an error.

> FileNotFoundError: Couldn't find file at https://github.com/facebookresearch/LASER/raw/master/data/tatoeba/v1/tatoeba.rus-eng.rus

## Steps to reproduce the bug
```python
data=load_dataset("xtreme","tatoeba.rus", split="validation")
```

## Solution
The library tries to access the **master** branch. In the github repo of facebookresearch, it is in the **main** branch.

mariosasko:
Updates the URL of the tatoeba subset of xtreme. Additionally, replaces `os.path.join` with `xjoin` to correctly join the URL segments on Windows.

Fix #3320


lhoestq (approved):
Thanks for updating the URLs

I think the changes to use `xjoin` are not necessary. I just removed them and I hope you don't mind

lhoestq:
To be more precise: `os.path.join` is replaced on-the-fly by `xjoin` anyway with patching, to extend it to remote files

lhoestq:
Oh actually just ignore what I said: they were used to concatenate URLs, which is not recommended. Let me fix that again

PR #3323, opened 2021-11-26
Fix wrongly converted assert

eliasws:
Seems like this assertion was replaced by an exception but the condition got wrongly converted.

### DIFF
--- a/src/datasets/search.py
+++ b/src/datasets/search.py
@@ -305,7 +305,7 @@ class FaissIndex(BaseIndex):
             scores (`List[List[float]`): The retrieval scores of the retrieved examples.
             indices (`List[List[int]]`): The indices of the retrieved examples.
         """
-        if len(query.shape) != 1 and (len(query.shape) == 2 and query.shape[0] != 1):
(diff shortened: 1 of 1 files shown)


eliasws:
#3327

src/datasets/search.py
```diff
@@ -305,7 +305,7 @@ def search(self, query: np.array, k=10) -> SearchResults:
             scores (`List[List[float]`): The retrieval scores of the retrieved examples.
             indices (`List[List[int]]`): The indices of the retrieved examples.
         """
-        if len(query.shape) != 1 or (len(query.shape) == 2 and query.shape[0] != 1):
+        if len(query.shape) != 1 and (len(query.shape) == 2 and query.shape[0] != 1):
```
lhoestq, line 308:
```suggestion
        if len(query.shape) != 1 and (len(query.shape) != 2 or query.shape[0] != 1):
```

lhoestq (approved):
Indeed thanks !

I also fixed the additional condition

PR #3328, opened 2021-11-27
Quick fix error formatting

NouamaneTazi:
While working on a dataset, I got the error
```
TypeError: Provided `function` which is applied to all elements of table returns a `dict` of types {[type(x) for x in processed_inputs.values()]}. When using `batched=True`, make sure provided `function` returns a `dict` of types like `{allowed_batch_return_types}`.
```

This PR should fix the formatting of this error


lhoestq (approved):
Thanks for the fix !

Issue #3272, opened 2021-11-15
Make iter_archive work with ZIP files


lhoestq:
Currently users can use `dl_manager.iter_archive` in their dataset script to iterate over all the files of a TAR archive.
It would be nice if it could work with ZIP files too !

Mehdi2402:
Hello, is this issue open for any contributor ? can I work on it ?

lhoestq:
Hi ! Sure this is open for any contributor. If you're interested feel free to self-assign this issue to you by commenting `#self-assign`. Then if you have any question or if I can help, feel free to ping me.

To begin with, feel free to take a look at both implementations of `iter_archive` for local downloads and for data strseaming:

In the `DownloadManager` for local dowloads:
https://github.com/huggingface/datasets/blob/dfa334bd8dc6cbc854b170379c7d2cb7e3d3fe4f/src/datasets/utils/download_manager.py#L218-L242

In the `StreamingDownloadManager` to stream the content of the archive directly from the remote file:
https://github.com/huggingface/datasets/blob/dfa334bd8dc6cbc854b170379c7d2cb7e3d3fe4f/src/datasets/utils/streaming_download_manager.py#L502-L526

Notice the call to `xopen` that opens and streams a file given either an URL or a local path :)

Mehdi2402:
Okay thank you for the information. I will work on this :) 

#self-assign

Mehdi2402:
#self-assign

PR #3347, opened 2021-11-30
iter_archive for zip files 

Mehdi2402:
* In this PR, I added the option to iterate through zipfiles for `download_manager.py` only. 
* Next PR will be the same applied to `streaming_download_manager.py`.
* Related issue #3272.
## Comments : 
* There is no `.isreg()` equivalent in zipfile library to check if file is Regular so I used `.is_dir()` instead to skip directories.
* For now I got `streaming_download_manager.py`  working for local zip files, but not for urls. I get the following error when I test it on an archive in google drive, so still working on it.  `BlockSizeError: Got more bytes so far (>2112) than requested (22)`

## Tasks : 
- [x] download_manager.py
- [ ] streaming_download_manager.py


src/datasets/utils/download_manager.py
```diff
@@ -225,21 +228,36 @@ def iter_archive(self, path):
             Generator yielding tuple (path_within_archive, file_obj).
             File-Obj are opened in byte mode (io.BufferedReader)
         """
-        with open(path, "rb") as f:
-            stream = tarfile.open(fileobj=f, mode="r|*")
-            for tarinfo in stream:
-                file_path = tarinfo.name
-                if not tarinfo.isreg():
+        extension = Path(path).suffix
```
lhoestq, line 231: I think cached zip or tar archives don't have the extension at the end of their filenames. Also for streaming we don't always know the filename either.

You can take a look at the `_get_extraction_protocol` function in `utils/streaming_download_manager.py`. It first checks the extension and then fallbacks to using the magic number of the file to guess the compression type

lhoestq (commented):
Hi ! Thanks for diving into this :)

Your implementation for ZIP looks all good ! I think we can just improve the compression type check:

lhoestq:
And also don't always try streaming with Google Drive - it can have issues because of how Google Drive works (with quotas, restrictions, etc.) and it can indeed cause `BlockSizeError`.

Feel free to host your test data elsewhere, such as in a dataset repository on https://huggingface.co (see [here](https://huggingface.co/docs/datasets/upload_dataset.html#upload-your-files) for a tutorial on how to upload files)

PR #3355, opened 2021-12-01
Extend support for streaming datasets that use pd.read_excel

albertvillanova:
This PR fixes error:
```
ValueError: Cannot seek streaming HTTP file
```

CC: @severo

### DIFF
--- a/src/datasets/utils/streaming_download_manager.py
+++ b/src/datasets/utils/streaming_download_manager.py
@@ -5,6 +5,7 @@ import re
 import tarfile
 import time
 from asyncio import TimeoutError
+from io import BytesIO
 from itertools import chain
 from pathlib import Path, PurePosixPath
 from typing import List, Optional, Tuple, Union
@@ -440,6 +441,12 @@ def xpandas_read_csv(filepath_or_buffer, use_auth_token: Optional[Union[str, boo
         return pd.read_csv(filepath_or_buffer, **kwargs)
     else:
         return pd.read_csv(xopen(filepath_or_buffer, use_auth_token=use_auth_token), **kwargs)
+
+
+def xpandas_read_excel(filepath_or_buffer, **kwargs):
+    import pandas as pd
+
+    return pd.read_excel(BytesIO(filepath_or_buffer.read()), **kwargs)
 
 
 class StreamingDownloadManager(object):
--- a/src/datasets/streaming.py
+++ b/src/datasets/streaming.py
@@ -13,6 +13,7 @@ from .utils.streaming_download_manager import (
     xlistdir,
     xopen,
     xpandas_read_csv,
+    xpandas_read_excel,
     xpathglob,
     xpathjoin,
     xpathopen,
@@ -74,4 +75,5 @@ def extend_module_for_streaming(module_path, use_auth_token: Optional[Union[str,
         patch.object(module.Path, "stem", property(fget=xpathstem)).start()
         patch.object(module.Path, "suffix", property(fget=xpathsuffix)).start()
     patch_submodule(module, "pd.read_csv", wrap_auth(xpandas_read_csv), attrs=["__version__"]).start()
+    patch_submodule(module, "pd.read_excel", xpandas_read_excel).start()
     module._patched_for_streaming = True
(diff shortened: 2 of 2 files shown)


src/datasets/utils/streaming_download_manager.py
```diff
@@ -442,6 +443,12 @@ def xpandas_read_csv(filepath_or_buffer, use_auth_token: Optional[Union[str, boo
         return pd.read_csv(xopen(filepath_or_buffer, use_auth_token=use_auth_token), **kwargs)
 
 
+def xpandas_read_excel(filepath_or_buffer, **kwargs):
+    import pandas as pd
+
+    return pd.read_excel(BytesIO(filepath_or_buffer.read()), **kwargs)
```
lhoestq, line 449: Maybe `.read()` must be called in the dataset scripts that have the HTTP range issues ?
This way people can still stream excel files without downloading them entirely if they want to, and it's also one less patch to worry about

albertvillanova: It is `pandas.read_excel` which calls `.seek(0)` on the `fsspec` class `HTTPStreamFile`.

```
huggingface\datasets\.venv\lib\site-packages\pandas\io\excel\_base.py in __init__(self, filepath_or_buffer)
    348         elif hasattr(filepath_or_buffer, "read"):
    349             # N.B. xlrd.Book has a read attribute too
--> 350             filepath_or_buffer.seek(0)
    351             self.book = self.load_workbook(filepath_or_buffer)
    352         elif isinstance(filepath_or_buffer, str):

huggingface\datasets\.venv\lib\site-packages\fsspec\implementations\http.py in seek(self, *args, **kwargs)
    616
    617     def seek(self, *args, **kwargs):
--> 618         raise ValueError("Cannot seek streaming HTTP file")
    619
    620     async def _read(self, num=-1):

ValueError: Cannot seek streaming HTTP file
```

albertvillanova: Moreover, Excel files have a maximum size limit of 2GB.

lhoestq: Ok I see !

I think `HTTPFile` supports `seek`, it only depends on the host of the file whether we get a `HTTPFile` or `HTTPStreamFile`. So for a good host (like the Hugging Face Hub 🙃) seeking in the file is possible.

What about catching the `ValueError: Cannot seek streaming HTTP file` and explain that in this case the user has to load the file completely with `.read()` ? This way the users say explicitly in their scripts that the data are loaded all at once and therefore they won't be confused by the bandwidth requires to start "streaming" the data

lhoestq: (not a strong opinion on this though, especially given that there are currently not that many excel datasets anyway)

albertvillanova: I tried with an Excel file in the Hub and I get the same error:
```
url = "https://huggingface.co/datasets/albertvillanova/tmp-tests/resolve/main/AJGT.xlsx"
with fsspec.open(url, "rb", block_size=0).open() as f:
    df = pd.read_excel(f, engine="openpyxl")
```
gives:
```
ValueError: Cannot seek streaming HTTP file
```

albertvillanova: Do you think there are any cases where the `pd.read_excel` can work in streaming mode (using `fsspec`) without using the `.read()`?

lhoestq: I think you're having the error because you set `block_size=0` ?

albertvillanova: No, I passed the `block_size` parameter to avoid having other error type:
```
BlockSizeError: Got more bytes so far (>107395) than requested (22)
```

lhoestq: Oh ok. It looks like a bug from `fsspec`. I'm ok with merging this PR until we have more info about this bug then

albertvillanova: - fsspec/filesystem_spec#160

lhoestq: That's weird, the Hub does accept range requests...

albertvillanova:
TODO in the future:
- If we finally find a use case where the `pd.read_excel()` can work in streaming mode (using fsspec), that is, without using the `.read()` [see https://github.com/huggingface/datasets/pull/3355#discussion_r761138011], I propose to try this first, catch the ValueError and then try with `.read`, but all implemented in `xpandas_read_excel`.

PR #3357, opened 2021-12-01
Update languages in aeslc dataset card

apergo-ai:
After having worked a bit with the dataset.
As far as I know, it is solely in English (en-US). There are only a few mails in Spanish, French or German (less than a dozen I would estimate).


albertvillanova (approved):
Thanks for the addition.

PR #3359, opened 2021-12-01
Add The Pile Free Law subset

albertvillanova:
Add:
- Free Law subset of The Pile: "free_law" config

Close bigscience-workshop/data_tooling#75.

CC: @StellaAthena

### DIFF
--- a/datasets/the_pile/README.md
+++ b/datasets/the_pile/README.md
@@ -67,7 +67,7 @@ datasets combined together.
 
 ### Languages
 
-[More Information Needed]
+This dataset is in English (`EN`).
 
 ## Dataset Structure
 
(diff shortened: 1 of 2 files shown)


StellaAthena:
@albertvillanova Is there a specific reason you’re adding the Pile under “the” instead of under “pile”? That does not appear to be consistent with other datasets.

albertvillanova:
Hi @StellaAthena,

I asked myself the same question, but at the end I decided to be consistent with previously added Pile subsets:
- #2817

I guess the reason is to stress that the definite article is always used before the name of the dataset (your site says: "The Pile. An 800GB Dataset of Diverse Text for Language Modeling"). Other datasets are not usually preceded by the definite article.

CC: @lhoestq

lhoestq:
> I guess the reason is to stress that the definite article is always used before the name of the dataset (your site says: "The Pile. An 800GB Dataset of Diverse Text for Language Modeling").

Yes that's because of this that it starts with "the"

PR #3364, opened 2021-12-01
Use the Audio feature in the AutomaticSpeechRecognition template

anton-l:
This updates the ASR template and all supported datasets to use the `Audio` feature

### DIFF
--- a/src/datasets/tasks/automatic_speech_recognition.py
+++ b/src/datasets/tasks/automatic_speech_recognition.py
@@ -1,20 +1,24 @@
 from dataclasses import dataclass
 from typing import ClassVar, Dict
 
-from ..features import Features, Value
+from ..features import Features, Value, Audio
 from .base import TaskTemplate
 
 
 @dataclass(frozen=True)
 class AutomaticSpeechRecognition(TaskTemplate):
     task: str = "automatic-speech-recognition"
-    # TODO(lewtun): Replace input path feature with dedicated `Audio` features
-    # when https://github.com/huggingface/datasets/pull/2324 is implemented
-    input_schema: ClassVar[Features] = Features({"audio_file_path": Value("string")})
+    input_schema: ClassVar[Features] = Features({"audio": Audio(_storage_dtype="string")})
     label_schema: ClassVar[Features] = Features({"transcription": Value("string")})
-    audio_file_path_column: str = "audio_file_path"
+    audio_column: str = "audio"
     transcription_column: str = "transcription"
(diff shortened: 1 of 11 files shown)


src/datasets/tasks/automatic_speech_recognition.py
```diff
@@ -1,20 +1,24 @@
...
     label_schema: ClassVar[Features] = Features({"transcription": Value("string")})
-    audio_file_path_column: str = "audio_file_path"
+    audio_column: str = "audio"
     transcription_column: str = "transcription"
+    # set the `streaming` flag when casting streamable datasets that have `Audio` stored as `struct{path, bytes}`
+    streaming: bool = False
+
+    def __post_init__(self):
+        if self.streaming:
+            self.__dict__["input_schema"] = Features({"audio": Audio(_storage_dtype="struct")})
```
anton-l, line 20: This is needed to support both the extracted and streamable datasets, avoiding type casting errors between `string` and `struct{}`. 
Maybe there's a better way to infer the underlying `Audio` storage type and not try to cast `Audio(_storage_dtype="struct")` <-> `Audio(_storage_dtype="string")`? Because setting the appropriate `streaming` flag could be confusing for contributors without additional docs.

To reproduce the type casting error, just remove this `__post_init__()` method and try converting CommonVoice:
```python
load_dataset("datasets/common_voice", "tr", split="train").prepare_for_task("automatic-speech-recognition")
```

mariosasko: If I'm not mistaken, we can avoid this logic by updating the `prepare_for_task` method to not call `cast` on the complex feature types (Audio, Image) with the `decode_example` method defined and to call `cast_column` on them instead.

Ideally, `cast` should know how to handle complex types (Audio, Image, ArrayXD), but currently, it doesn't.

lhoestq:
Cool !

I noticed that you removed the `audio_file_path_column` field of the template, note that you also have to update all the dataset_infos.json file that still contain this outdated field. For example in the common_voice you can find this:
```
"task_templates": [{"task": "automatic-speech-recognition", "audio_file_path_column": "path", "transcription_column": "sentence"}]
```

mariosasko:
Yes, will do that. I'm just busy with the bigscience task.

mariosasko:
After we merge this, we should also update the following dataset scripts: https://huggingface.co/datasets?task_ids=task_ids:automatic-speech-recognition

mariosasko:
Closing in favor of https://github.com/huggingface/datasets/pull/4006

PR #3375, opened 2021-12-03
Support streaming zipped dataset repo by passing only repo name

### LINKED ISSUE #3373: Support streaming zipped CSV dataset repo by passing only repo name
Given a community 🤗 dataset repository containing only a zipped CSV file (only raw data, no loading script), I would like to load it in streaming mode without passing `data_files`:
```
ds_name = "bigscience-catalogue-data/vietnamese_poetry_from_fsoft_ai_lab"
ds = load_dataset(ds_name, split="train", streaming=True, use_auth_token=True)
item = next(iter(ds))
```

Currently, it gives a `FileNotFoundError` because there is no glob (no "\*" after "zip://": "zip://*") in the passed URL:
```
'zip://::https://huggingface.co/datasets/bigscience-catalogue-data/vietnamese_poetry_from_fsoft_ai_lab/resolve/e5d45f1bd9a8a798cc14f0a45ebc1ce91907c792/poems_dataset.zip'
```

albertvillanova:
Proposed solution:
- I have added the method `iter_files` to DownloadManager and StreamingDownloadManager
- I use this in modules: "csv", "json", "text"
- I test for CSV/JSONL/TXT zipped (and non-zipped) files, both in streaming and non-streaming modes

Fix #3373.


lhoestq:
I just tested and I think this only opens one file ? If there are several files in the ZIP, only the first one is opened. To open several files from a ZIP, one has to call `open` several times.

What about updating the CSV loader to make it `download_and_extract` zip files, and open each extracted file ?

albertvillanova:
I have implemented the  glob of ZIP files in the packaged modules:
- csv
- json
- text

albertvillanova:
Also for streaming and non-streaming.

albertvillanova:
In https://github.com/huggingface/datasets/pull/3375/commits/c10275fe36085601cb7bdb9daee9a8f1fc734f48, there were 3 failing tests:
```
FAILED tests/test_streaming_download_manager.py::test_streaming_dl_manager_get_extraction_protocol[https://drive.google.com/uc?export=download&id=1k92sUfpHxKq8PXWRr7Y5aNHXwOCNUmqh-zip]
FAILED tests/test_streaming_download_manager.py::test_streaming_gg_drive - Fi...
FAILED tests/test_streaming_download_manager.py::test_streaming_gg_drive_zipped
= 3 failed, 3553 passed, 2950 skipped, 2 xfailed, 1 xpassed, 125 warnings in 192.79s (0:03:12) =
```

After re-running the CI in https://github.com/huggingface/datasets/pull/3375/commits/57bfe1f342cd3c59d2510b992d5f06a0761eb147, there was only 1 failing test:
```
FAILED tests/test_streaming_download_manager.py::test_streaming_gg_drive_zipped
= 1 failed, 3555 passed, 2950 skipped, 2 xfailed, 1 xpassed, 125 warnings in 199.76s (0:03:19) =
```

src/datasets/packaged_modules/csv/csv.py
```diff
@@ -145,7 +145,7 @@ def _split_generators(self, dl_manager):
         """We handle string, list and dicts in datafiles"""
         if not self.config.data_files:
             raise ValueError(f"At least one data file must be specified, but got data_files={self.config.data_files}")
-        data_files = dl_manager.download_and_extract(self.config.data_files)
+        data_files = dl_manager.download_and_extract(self.config.data_files, glob_archives=True)
```
lhoestq, line 148: there's already some globbing a few lines after with
```python
            if any(os.path.isdir(file) for file in files):
                files = [file for file in _iter_files(files)]
```
Could it be possible to make `_iter_files` work with streaming, instead of having an additional logic ?

albertvillanova: Let me think of a solution...

albertvillanova: Please also note that you implemented the functionality in the function `_iter_files` (not a method) and use it only in "csv" module, whereas that functionality is needed in all packaged modules ("json", "text",...).

My approach was to put that functionality inside the DownloadManager, so that it can be automatically used in all packaged modules ("csv", "json",...).

Therefore, the ideal solution is:
- to merge the functionality of `_iter_files` and `glob_archive`: so that we do not have redundant calls to glob
- and put it inside the DownloadManager: so that this functionality can be used in all packages modules ("csv", "json"...)

lhoestq: That makes sense

If it can help there's also a PR that adds support for ZIP archives for `iter_archive`, if it can be helpful: #3379. The `iter_archive` method should also be better in terms of performance for streaming thanks to buffering, compared to globbing + opening the inner files one by one. If you think this can be useful I can help the contributor to merge this PR faster :)

albertvillanova: Proposed approach: implement `[Streaming]DownloadManager.iter_files` using `os.walk` [`xwalk`].

albertvillanova:
@lhoestq my final proposed solution:
- I have added the method `iter_files` to DownloadManager and StreamingDownloadManager
- I use this in modules: "csv", "json", "text"
- I test for CSV/JSONL/TXT zipped (and non-zipped) files, both in streaming and non-streaming modes

src/datasets/utils/mock_download_manager.py
```diff
@@ -126,7 +126,7 @@ def download_and_extract(self, data_url, *args):
             return self.create_dummy_data_single(dummy_file, data_url)
 
     # this function has to be in the manager under this name so that testing works
-    def download(self, data_url, *args):
+    def download(self, data_url, *args, **kwargs):
```
lhoestq, line 129: I think you can remove the kwargs ?

src/datasets/utils/mock_download_manager.py
```diff
@@ -109,7 +109,7 @@ def manual_dir(self):
         return "/".join(self.dummy_file.replace(os.sep, "/").split("/")[:-1])
 
     # this function has to be in the manager under this name so that testing works
-    def download_and_extract(self, data_url, *args):
+    def download_and_extract(self, data_url, *args, **kwargs):
```
lhoestq, line 112: here as well

albertvillanova:
> Note that at one point we might consider switching to using `iter_archive` for ZIP files in the json/text/csv loaders since it should be faster.

As far as the functionality is kept... ;)

src/datasets/utils/mock_download_manager.py
albertvillanova: I left it because `*args` was already present. But I'm removing it if you think it is OK to have *args but not **kwargs.... :P

PR #3382, opened 2021-12-04
#3337 Add typing overloads to Dataset.__getitem__ for mypy

### LINKED ISSUE #3337: Typing of Dataset.__getitem__ could be improved.
## Describe the bug

The newly added typing for Dataset.__getitem__ is Union[Dict, List]. This makes tools like mypy a bit awkward to use as we need to check the type manually. We could use type overloading to make this easier. [Documentation](https://docs.python.org/3/library/typing.html#typing.overload)

## Steps to reproduce the bug
Let's have a file `test.py`

```python
from typing import List, Dict, Any

from datasets import Dataset

ds = Dataset.from_dict({
    'a': [1,2,3],
    'b': ["1", "2", "3"]
})

one_colum: List[str] = ds['a']
some_index: Dict[Any, Any] = ds[1]
```

## Expected results

Running `mypy test.py` should not give any error.

## Actual results

```
test.py:10: error: Incompatible types in assignment (expression has type "Union[Dict[Any, Any], List[Any]]", variable has type "List[str]")
test.py:11: error: Incompatible types in assignment (expression has type "Union[Dict[Any, Any], List[Any]]", variable has type "Dict[Any, Any]")
Found 2 errors in 1 file (checked 1 source file)
```

## Environment info

- `datasets` version: 1.13.3
- Python version: 3.8.8
- PyArrow version: 6.0.1

Dref360:
Add typing overloads to Dataset.__getitem__ for mypy

Fixes #3337 

**Iterable**
Iterable from `collections` cannot have a type, so you can't do `Iterable[int]` for example. `typing` has a Generic version that builds upon the one from `collections`.

**Flake8**
I had to add `# noqa: F811`, this is a bug from Flake8.

datasets uses flake8==3.7.9 which released in October 2019 if I update flake8 (4.0.1), I no longer get these errors, but I did not want to make the update without your approval. (It also triggers other errors like no args in f-strings.)


src/datasets/arrow_dataset.py
```diff
@@ -1837,7 +1850,15 @@ def _getitem(self, key: Union[int, slice, str], decoded: bool = True, **kwargs)
         )
         return formatted_output
 
-    def __getitem__(self, key: Union[int, slice, str]) -> Union[Dict, List]:
+    @overload
+    def __getitem__(self, key: Union[int, slice, Iterable[int], Iterable[bool]]) -> Dict:
```
Dref360, line 1854:
```suggestion
    def __getitem__(self, key: Union[int, slice, Iterable[int], Iterable[bool]]) -> Dict:  # noqa: F811
```

Dref360:
Locally the `make quality` passes with the same dependencies. I would suggest upgrading flake8. (I can take care of it in another PR)

src/datasets/arrow_dataset.py
```diff
@@ -1837,7 +1850,15 @@ def _getitem(self, key: Union[int, slice, str], decoded: bool = True, **kwargs)
         )
         return formatted_output
 
-    def __getitem__(self, key: Union[int, slice, str]) -> Union[Dict, List]:
+    @overload
+    def __getitem__(self, key: Union[int, slice, Iterable[int], Iterable[bool]]) -> Dict:  # noqa: F811
```
lhoestq, line 1854: We don't support lists of booleans, they are converted to integers.
```suggestion
    def __getitem__(self, key: Union[int, slice, Iterable[int]]) -> Dict:  # noqa: F811
```

lhoestq (commented):
Cool thanks ! Let me see if I can fix the flake8 issue

Dref360:
Thank you for fixing flake8! I think we are ready to merge then.

lhoestq (approved):
Thank you !

PR #3383, opened 2021-12-05
add Georgian data in cc100.

AnzorGozalishvili:
update cc100 dataset to support loading Georgian (ka) data which is originally available in CC100 dataset source.
All tests are passed. 
Dummy data generated.
metadata generated.


lhoestq (approved):
Thank you ! It looks all good :)

PR #3387, opened 2021-12-06
Create Language Modeling task

albertvillanova:
Create Language Modeling task to be able to specify the input "text" column in a dataset.

This can be useful for datasets which are not exclusively used for language modeling and have more than one column:
- for text classification datasets (with columns "review" and "rating", for example), the Language Modeling task can be used to specify the "text" column ("review" in this case).

TODO:
- [ ] Add the LanguageModeling task to all dataset scripts which can be used for language modeling


lhoestq (approved):
Cool thanks ! we can add the task template to the datasets progressively in subsequent PRs

PR #3388, opened 2021-12-06
Fix flaky test of the temporary directory used by load_from_disk

lhoestq:
The test is flaky, here is an example of random CI failure:
https://github.com/huggingface/datasets/commit/73ed6615b4b3eb74d5311684f7b9e05cdb76c989

I fixed that by not checking the content of the random part of the temporary directory name


lhoestq:
CI failed because of a server error - merging

PR #3395, opened 2021-12-07
Fix formatting in IterableDataset.map docs

mariosasko:
Fix formatting in the recently added `Map` section of the streaming docs.


lhoestq (approved):
Good catch :)

PR #3402, opened 2021-12-07
More robust first elem check in encode/cast example

### LINKED ISSUE #3306: nested sequence feature won't encode example if the first item of the outside sequence is an empty list
## Describe the bug
As the title, nested sequence feature won't encode example if the first item of the outside sequence is an empty list.

## Steps to reproduce the bug
```python
from datasets import Features, Sequence, ClassLabel
features = Features({
    'x': Sequence(Sequence(ClassLabel(names=['a', 'b']))),
})
print(features.encode_batch({
    'x': [
        [['a'], ['b']],
        [[], ['b']],
    ]
}))
```

## Expected results
print `{'x': [[[0], [1]], [[], ['1']]]}`

## Actual results
print `{'x': [[[0], [1]], [[], ['b']]]}`

## Environment info
- `datasets` version: 1.15.1
- Platform: Linux-5.13.0-21-generic-x86_64-with-glibc2.34
- Python version: 3.9.7
- PyArrow version: 6.0.0

## Additional information
I think the issue stems from [here](https://github.com/huggingface/datasets/blob/8555197a3fe826e98bd0206c2d031c4488c53c5c/src/datasets/features/features.py#L847-L848).

mariosasko:
Fix #3306

### DIFF
--- a/tests/features/test_features.py
+++ b/tests/features/test_features.py
@@ -244,6 +244,23 @@ def test_encode_nested_example_sequence_with_none():
     assert result is None
 
 
+def test_encode_batch_with_example_with_empty_first_elem():
+    features = Features(
+        {
+            "x": Sequence(Sequence(ClassLabel(names=["a", "b"]))),
+        }
+    )
+    encoded_batch = features.encode_batch(
+        {
+            "x": [
+                [["a"], ["b"]],
+                [[], ["b"]],
+            ]
+        }
+    )
+    assert encoded_batch == {"x": [[[0], [1]], [[], [1]]]}
+
+
 def iternumpy(key1, value1, value2):
     if value1.dtype != value2.dtype:  # check only for dtype
         raise AssertionError(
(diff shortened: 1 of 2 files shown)


tests/features/test_features.py
```diff
@@ -244,6 +244,23 @@ def test_encode_nested_example_sequence_with_none():
...
 
+def test_encode_batch_with_example_with_empty_first_elem():
+    features = Features(
+        {
+            "x": Sequence(Sequence(ClassLabel(names=["a", "b"]))),
+        }
+    )
+    encoded_batch = features.encode_batch(
+        {
+            "x": [
```
lhoestq, line 255: To be properly tested, does the first element has to be an empty list ?

lhoestq (approved):
Cool thanks !

tests/features/test_features.py
mariosasko: Yes, because `encode_batch` calls `encode_nested_example` which goes from the beginning of the list and tries to find the first element that is "good enough" to perform additional checks, and we consider an element "good enough" if it's not `None` or if it's not an empty sequence (possible nested). Previously, we would stop on the first element that is not `None`, but this could lead to issues such as the one this PR fixes.

PR #3406, opened 2021-12-08
Fix module inference for archive with a directory

### LINKED ISSUE #3405: ZIP format inference does not work when files located in a dir inside the archive
## Describe the bug
When a zipped file contains archived files within a directory, the function `infer_module_for_data_files_in_archives` does not work.

It only works for files located in the root directory of the ZIP file.

## Steps to reproduce the bug
```python
infer_module_for_data_files_in_archives(["path/to/zip/file.zip"], False)
```

albertvillanova:
Fix module inference for an archive file that contains files within a directory.

Fix #3405.


lhoestq (approved):
Good catch !! thanks

PR #3413, opened 2021-12-09
Add WIDER FACE dataset

mariosasko:
Adds the WIDER FACE face detection benchmark.

TODOs:
* [x] dataset card
* [x] dummy data


lhoestq (approved):
Love it ! LGTM
Good job with the dataset script and the dataset card :)

Issue #3369, opened 2021-12-02
[Audio] Allow resampling for audio datasets in streaming mode


patrickvonplaten:
Many audio datasets like Common Voice always need to be resampled. This can very easily be done in non-streaming mode as follows:

```python
from datasets import load_dataset

ds = load_dataset("common_voice", "ab", split="test")

ds = ds.cast_column("audio", Audio(sampling_rate=16_000))
```

However in streaming mode it fails currently:

```python
from datasets import load_dataset

ds = load_dataset("common_voice", "ab", split="test", streaming=True)

ds = ds.cast_column("audio", Audio(sampling_rate=16_000))
```

with the following error:

```
AttributeError: 'IterableDataset' object has no attribute 'cast_column'  
```

It would be great if we could add such a feature (I'm not 100% sure though how complex this would be)

lhoestq:
This requires implementing `cast_column` for iterable datasets. It can also be useful to be able to disable the audio/image decoding for the dataset viewer cc @severo

patrickvonplaten:
Just to clarify a bit. This feature is **always** needed when using the common voice dataset in streaming mode. So I think it's quite important

PR #3430, opened 2021-12-14
Make decoding of Audio and Image feature optional

mariosasko:
Add the `decode` argument (`True` by default) to the `Audio` and the `Image` feature to make it possible to toggle on/off decoding of these features.

Even though we've discussed that on Slack, I'm not removing the `_storage_dtype` argument of the Audio feature in this PR to avoid breaking the Audio feature tests.


mariosasko:
Closing this PR for now due to https://github.com/huggingface/datasets/issues/3145#issuecomment-993664104.

mariosasko:
Okay, after some more thinking, I'm re-opening this PR for three reasons:
* This feature will allow us to remove the `image_file_path`/`audio_file_path` columns in our vision/audio datasets. Currently, it makes sense to keep those columns because it's not obvious how to access the underlying path information of the Image/Audio feature. However, if the user is not aware and does `dataset[0]["image_file_path"]` on our vision/audio datasets, this will be costly because the image/audio file data has to be decoded first (stored in `dataset[0]["image"]`/`dataset[0]["audio"]`)
* In CV, we often work with the so-called "half life" datasets (RedCaps, WIT, ...) that only provide image URLs and not actual image data, and some of these image URLs even go down. Our solution to this problem is to give a note on how to efficiently download the image data using `map` in the datasets cards of such datasets. This feature will remove the need for a separate `image_url` column of type `Value("string")` in such datasets. Instead, we will be able to use the `image` column of type `Image()`  (the image feature knows how to decode image URLs using `xopen`), disable decoding and use `requests.get` for download, which I expect to be faster than `xopen`.
* This feature should help us in implementing `push_to_hub` for the Image/Audio where we transfer actual image/audio data and not paths

lhoestq:
> This feature will allow us to remove the image_file_path/audio_file_path columns in our vision/audio datasets. Currently, it makes sense to keep those columns because it's not obvious how to access the underlying path information of the Image/Audio feature. However, if the user is not aware and does dataset[0]["image_file_path"] on our vision/audio datasets, this will be costly because the image/audio file data has to be decoded first (stored in dataset[0]["image"]/dataset[0]["audio"])

That makes sense !

> Instead, we will be able to use the image column of type Image() (the image feature knows how to decode image URLs using xopen), disable decoding and use requests.get for download, which I expect to be faster than xopen.

I feel like it's a bit convoluted compared to having the `image_url` column as string, and say to users to `map` using `requests.get` with `image_url`.

Moreover I'm not 100% sure that we should have `Image` features with both local paths and URLs, since this behavior is a bit hidden the users and they don't give the same performance at all.

> This feature should help us in implementing push_to_hub for the Image/Audio where we transfer actual image/audio data and not paths

Cool !

lhoestq:
> Do you mean we should remove support for URLs in the Image feature? Because this is what we get for free by adding support for streaming (by using xopen instead of open) and this is also what the Audio feature does.

I think it might not be ideal to have URLs in an `Image` type column for a dataset in **non-streaming** mode, since you'd expect to have everything locally. But for a streaming dataset it must use `xopen` indeed

mariosasko:
Yes, I agree. Let's have the `image_url` columns as `Value("string")` + a note with the map function to download images for local datasets and a note with `cast_column` (which is requested https://github.com/huggingface/datasets/issues/3369) for streamed datasets (`ds.cast_column("image_url", Image())`).

mariosasko:
I fixed the merge conflicts and small bugs in nested decoding introduced by #3575. Additionally, I addressed https://github.com/huggingface/datasets/issues/3473 by adding the `_iter` method to `Dataset` (inspired by the `_getitem` method). For `iter(dset)` I set `_iter(dset, decoded=True)` to enable decoding and for `map` `_iter(dset, decoded=False)` to make it lazy.

src/datasets/features/audio.py
```diff
@@ -84,6 +87,10 @@ def decode_example(self, value: dict) -> dict:
         Returns:
             dict
         """
+        if not self.decode:
+            raise RuntimeError(
+                "Decoding is disabled for this feature. Set the `decode` attribute to `True` to enable it."
```
lhoestq, line 92: (nit) "Set the `decode` attribute to `True` to enable it" can make the user think that they can just do
```python
datasets.features[name].decode = True
```
which is not allowed since it doesn't update the `fingerprint` nor `dataset.features._column_requires_decoding`

Maybe we can rephrase this to something like this ?
```suggestion
                "Decoding is disabled for this feature. Please use Audio(decode=True) instead."
```

the same applied for Image as well

Issue #3398, opened 2021-12-07
Add URL field to Wikimedia dataset instances: wikipedia,...


albertvillanova:
As reported by @geohci, in order to host pre-processed data in the Hub, we should add the full URL to data instances (new field "url"), so that we conform to proper attribution from license requirement. See, e.g.: https://fair-trec.github.io/docs/Fair_Ranking_2021_Participant_Instructions.pdf#subsection.3.2

This should be done for all pre-processed datasets under "wikimedia" org in the Hub: https://huggingface.co/wikimedia

albertvillanova:
@geohci, I think the field "url" does not appear in the Wikimedia dumps. Therefore I guess we should generate it, using the "title" field and url-encoding it?

lhoestq:
Indeed:

> To re-distribute text on Wikipedia in any form, provide credit to the authors either by including a) a [hyperlink](https://en.wikipedia.org/wiki/Hyperlink) (where possible) or [URL](https://en.wikipedia.org/wiki/URL) to the page or pages you are re-using, b) a hyperlink (where possible) or URL to an alternative, stable online copy which is freely accessible, which conforms with the license, and which provides credit to the authors in a manner equivalent to the credit given on this website, or c) a list of all authors. (Any list of authors may be filtered to exclude very small or irrelevant contributions.) This applies to text developed by the Wikipedia community. Text from external sources may attach additional attribution requirements to the work, which should be indicated on an article's face or on its talk page. For example, a page may have a banner or other notation indicating that some or all of its content was originally published somewhere else. Where such notations are visible in the page itself, they should generally be preserved by re-users.

source: https://en.wikipedia.org/wiki/Wikipedia:Copyrights

I guess it's fine to add the URL field - it can be constructed easily from the title page IIRC.

geohci:
yep, sorry forgot that that wasn't already in the dumps. specifically `f"https://{language}.wikipedia.org/wiki/{title.replace(' ', '_')}` should do it

albertvillanova:
Thanks @geohci.

I had already been looking for information about the conversion from title to URL and I found that apart from replacing blanks with underscores, some other special character must also be percent-encoded (e.g. `"` to `%22`).

Therefore, I have finally used `urllib.parse.quote` function. This additionally percent-encodes non-ASCII characters, but Wikimedia doc say these are equivalent:
> For the other characters either the code or the character can be used in internal and external links, they are equivalent. The system does a conversion when needed.
[...]

PR #3435, opened 2021-12-15
Improve Wikipedia Loading Script

### LINKED ISSUE #3400: Improve Wikipedia loading script
As reported by @geohci, the "wikipedia" processing/loading script could be improved by some additional small suggested processing functions:
- _extract_content(filepath):
  - Replace .startswith("#redirect") with more structured approach: if elem.find(f"./{namespace}redirect") is None: continue
- _parse_and_clean_wikicode(raw_content, parser):
  - Remove rm_template from cleaning -- this is redundant with .strip_code() from mwparserformhell
  - Build a language-specific list of namespace prefixes to filter out per below get_namespace_prefixes
  - Optional: strip prefixes like categories -- e.g., Category:Towns in Tianjin becomes Towns in Tianjin
  - Optional: strip magic words

geohci:
* More structured approach to detecting redirects
* Remove redundant template filter code (covered by strip_code)
* Add language-specific lists of additional media namespace aliases for filtering
* Add language-specific lists of category namespace aliases for new link text cleaning step
* Remove magic words (parser directions like __TOC__ that occasionally occur in text)

Fix #3400

With support from @albertvillanova 
CC @yjernite


geohci:
I wanted to flag a change from since we discussed this: I initially wrote a function for using the Wikimedia APIs to collect namespace aliases, but decided that adding in more http requests to the script wasn't a great idea so instead used that code to build a static list that I just added directly to the code.

Also, an FYI that python library dependencies weren't working on my local end so I wasn't able to directly test the code. I tested a copy with the problematic elements stripped (beam etc.) that worked fine, but someone with a working local copy may want to test just to make sure I didn't accidentally break anything.

geohci:
Also, while I would argue more strongly for some of the changes in this code, they are five distinct changes so not so hard to remove one or two if other folks think they aren't worth the overhead etc.

datasets/wikipedia/wikipedia.py
```diff
@@ -357,6 +357,516 @@
     "zu",
 ]
 
+MEDIA_ALIASES = {
```
albertvillanova, line 360: Could you please add a comment with the URL you get the aliases from?

datasets/wikipedia/wikipedia.py
```diff
@@ -357,6 +357,516 @@
...
+    "yo": ["Fáìlì", "Amóhùnmáwòrán", "Àwòrán"],
+    "za": ["媒体文件", "文件", "档案", "图像", "媒体"],
+    "zea": ["Afbeelding", "Plaetje"],
+    "zh": ["媒体文件", "F", "文件", "媒體", "档案", "图像", "圖像", "媒体", "檔案"],
+    "zh-classical": ["文件", "媒體", "圖像", "檔案"],
+    "zh-min-nan": ["tóng-àn", "文件", "媒體", "Mûi-thé", "圖像", "檔案"],
+    "zh-yue": ["檔", "档", "文件", "图", "媒體", "圖", "档案", "图像", "圖像", "媒体", "檔案"],
+}
+
+CAT_ALIASES = {
```
albertvillanova, line 615: Also here: comment with source URL

albertvillanova (commented):
Thanks for your suggested improvements @geohci.

I ping the rest of the Datasets team, so that we also have their feedback.

CC: @lhoestq @mariosasko

albertvillanova:
I also add a comment by @geohci in the Issue page:
> See https://public.paws.wmcloud.org/User:Isaac_(WMF)/HuggingFace%20Wikipedia%20Processing.ipynb for more implementation details / some data around the overhead induced by adding the extra preprocessing steps (stripping link prefixes and magic words)

datasets/wikipedia/wikipedia.py
geohci: added -- it's slightly more complex because it's an API call for each language but I added enough that hopefully easy to update if ever needed.

datasets/wikipedia/wikipedia.py
geohci: added with same caveat as above

lhoestq:
Hi ! Thanks a lot, this is very cool ! Note that unfortunately if we change the processing right now, users won't be able to load the "big" languages like english anymore, because it requires an Apache Beam runtime to process them. Some Wikipedia dumps have been processed by Hugging Face so that users don't need to run Apache Beam stuff.

Therefore, we can merge this change after we have processed dumps using this new processing, and host them on the Hugging Face google storage.

I think we can take care of this and let you know once this is ready ? What do you think @albertvillanova ?

This is also an opportunity to have the latest dumps ready, the current ones are from 2020

lhoestq:
Related PR on updating to the latest dates: https://github.com/huggingface/datasets/pull/3612

albertvillanova:
@lhoestq if the additional processing steps are validated, we could go on generating the processed datasets for the big languages.

The only thing before doing that is that we should also validate other change (so that we include it also in the processed datasets):
- #3398

lhoestq:
> @lhoestq if the additional processing steps are validated, we could go on generating the processed datasets for the big languages.

Cool ! Looking forward to it :)

> The only thing before doing that is that we should also validate other change (so that we include it also in the processed datasets):
> 
> https://github.com/huggingface/datasets/issues/3398

Sounds good ! We can definitely add the URL as asked by the Wikipedia to provide credits to the authors.

geohci:
> @geohci I do not have push rights to this PR. See: [Enabling repository maintainer permissions on existing pull requests](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests).
> 
> I would like to merge the master branch so that all tests pass. Once done, I will be able approve this PR.

@albertvillanova the `Allow edits by maintainers` box was already checked (what your instructions indicated) and indicates `If checked, users with write access to huggingface/datasets can add new commits to your wikipedia-updates branch. You can always change this setting later.` so you should have permissions already. If there's something else I'm missing or can do, please let me know. If it's not easy to resolve, I am plenty comfortable with you creating a new PR with these changes under your account too.

albertvillanova (approved):
Thank you, @geohci.

I'm merging this PR into a dedicated branch where we will collect all Wikipedia updates.

Then we will generate the pre-processed datasets with the updated script (using this dedicated branch).

And finally, we will merge everything to master.

PR #3436, opened 2021-12-15
Add the OneStopQa dataset

scaperex:
Adding OneStopQA,  a multiple choice reading comprehension dataset annotated according to the STARC (Structured Annotations for Reading Comprehension) scheme.


lhoestq (approved):
Thanks a lot @scaperex ! It looks all good :)

Good job with the dataset card !

PR #3437, opened 2021-12-15
Update BLEURT hyperlink

lewtun:
The description of BLEURT on the hf.co website has a strange use of URL hyperlinking. This PR attempts to fix this, although I am not 100% sure Markdown syntax is allowed on the frontend or not.

Screen Shot 2021-12-15 at 17 31 27

### DIFF
--- a/metrics/bleurt/bleurt.py
+++ b/metrics/bleurt/bleurt.py
@@ -39,7 +39,7 @@ BLEURT a learnt evaluation metric for Natural Language Generation. It is built u
 and then employing another pre-training phrase using synthetic data. Finally it is trained on WMT human annotations. You may run BLEURT out-of-the-box or fine-tune
 it for your specific application (the latter is expected to perform better).
 
-See the project's `README <https://github.com/google-research/bleurt>`__ for more information.
+See the project's README at https://github.com/google-research/bleurt for more information.
 """
 
 _KWARGS_DESCRIPTION = """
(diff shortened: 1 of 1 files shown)


albertvillanova:
Hi @lewtun. Thanks for noticing that issue.

Unfortunately, the frontend does not support Markdown (at least for /metrics); see: https://huggingface.co/metrics/bertscore

I wonder whether:
- another syntax for links is supported (reStructuredText, HTML,...)
- or support for Markdown should be implemented, as it is already the case for /datasets; see: https://huggingface.co/datasets/speech_commands#dataset-summary

CC: @julien-c @SBrandeis @Pierrci ...

albertvillanova (commented):
Hi @lewtun. Thanks for noticing that issue.

Unfortunately, the frontend does not support Markdown (at least for /metrics); see: https://huggingface.co/metrics/bertscore

I wonder whether:
- another syntax for links is supported (reStructuredText, HTML,...)
- or support for Markdown should be implemented, as it is already the case for /datasets; see: https://huggingface.co/datasets/speech_commands#dataset-summary

CC: @julien-c @SBrandeis @Pierrci ...

julien-c:
seems like a very very low-prio improvement :)

lewtun:
@albertvillanova thanks for the feedback! I removed the formatting altogether since I think this is a bit simpler tor read than non-rendered Markdown

lhoestq (approved):

PR #3438, opened 2021-12-15
Update supported versions of Python in setup.py

mariosasko:
Update the list of supported versions of Python in `setup.py` to keep the PyPI project description updated.


lhoestq (approved):
thanks !

PR #3439, opened 2021-12-15
Add `cast_column` to `IterableDataset`

### LINKED ISSUE #3369: [Audio] Allow resampling for audio datasets in streaming mode

mariosasko:
Closes #3369. 

cc: @patrickvonplaten

### DIFF
--- a/docs/source/package_reference/main_classes.rst
+++ b/docs/source/package_reference/main_classes.rst
@@ -73,7 +73,7 @@ The base class :class:`datasets.IterableDataset` implements an iterable Dataset 
 
 .. autoclass:: datasets.IterableDataset
     :members:
-        remove_columns,
+        remove_columns, cast_column,
         __iter__,
         map, shuffle, skip, take,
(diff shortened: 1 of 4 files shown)


patrickvonplaten:
Awesome thanks a lot @mariosasko

lhoestq (approved):
Thanks @mariosasko !

This is also related to https://github.com/huggingface/datasets/issues/3444

PR #3442, opened 2021-12-16
Extend text to support yielding lines, paragraphs or documents

albertvillanova:
Add `config.row` option to `text` module to allow yielding lines (default, current case), paragraphs or documents.

Feel free to comment on the name of the config parameter `row`:
- Currently, the docs state datasets are made of rows and columns
- Other names I considered: `example`, `item`

### DIFF
--- a/src/datasets/packaged_modules/text/text.py
+++ b/src/datasets/packaged_modules/text/text.py
@@ -75,7 +75,9 @@ class Text(datasets.ArrowBasedBuilder):
                             break
                         batch += f.readline()  # finish current line
                         batch = batch.split("\n\n")
-                        pa_table = pa.Table.from_arrays([pa.array(batch[:-1])], schema=schema)
+                        pa_table = pa.Table.from_arrays(
+                            [pa.array([example for example in batch[:-1] if example])], schema=schema
+                        )
                         # Uncomment for debugging (will print the Arrow table size and elements)
                         # logger.warning(f"pa_table: {pa_table} num rows: {pa_table.num_rows}")
                         # logger.warning('\n'.join(str(pa_table.slice(i, 1).to_pydict()) for i in range(pa_table.num_rows)))
--- a/src/datasets/packaged_modules/text/text.py
+++ b/src/datasets/packaged_modules/text/text.py
@@ -84,4 +84,5 @@ class Text(datasets.ArrowBasedBuilder):
                         batch = batch[-1]
                 elif self.config.row == "document":
                     text = f.read()
-                    yield file_idx, pa.Table.from_pydict({"text": [text]})
+                    pa_table = pa.Table.from_arrays([pa.array([text])], schema=schema)
+                    yield file_idx, pa_table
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
Awesome !

lhoestq:
The parameter can also be named `separate_by`

src/datasets/packaged_modules/text/text.py
```diff
@@ -51,15 +52,37 @@ def _generate_tables(self, files):
...
+                        batch_idx += 1
+                elif self.config.row == "paragraph":
+                    batch_idx = 0
+                    batch = ""
+                    while True:
+                        batch += f.read(self.config.chunksize)
+                        if not batch:
+                            break
+                        batch += f.readline()  # finish current line
+                        batch = batch.split("\n\n")
```
lhoestq, line 77: Maybe filter empty paragraphs ?

albertvillanova:
> The parameter can also be named `split_by` with values "line", "paragraph" or "document" (no 's' at the end)

@lhoestq @mariosasko I would avoid the term `split` in this context and keep it only for "train", "validation" and "test" splits.
- https://huggingface.co/docs/datasets/process.html#split
  > datasets.Dataset.train_test_split() creates train and test splits, if your dataset doesn’t already have them.
- https://huggingface.co/docs/datasets/process.html#process-multiple-splits
  > Many datasets have splits that you can process simultaneously with datasets.DatasetDict.map().

Please note that in the documentation, one of the terms more frequently used in this context is "row":
- https://huggingface.co/docs/datasets/access.html#features-and-columns
  > A dataset is a table of rows and typed columns.

  > Return the number of rows and columns with the following standard attributes:
  > dataset.num_columns
  > 4
  > dataset.num_rows
  > 3668

- https://huggingface.co/docs/datasets/access.html#rows-slices-batches-and-columns
  > Get several rows of your dataset at a time with slice notation or a list of indices:
- https://huggingface.co/docs/datasets/process.html#map
  > This function can even create new rows and columns.

Other of the terms more frequently used in the docs (in the code as well) is "example":
- https://huggingface.co/docs/datasets/process.html#map
  > It allows you to apply a processing function to each example in a dataset, independently or in batches.
- https://huggingface.co/docs/datasets/process.html#batch-processing
  > datasets.Dataset.map() also supports working with batches of examples.
- https://huggingface.co/docs/datasets/process.html#split-long-examples
  > When your examples are too long, you may want to split them
- https://huggingface.co/docs/datasets/process.html#data-augmentation
  > With batch processing, you can even augment your dataset with additional examples.

Less frequently used: "item":
- https://huggingface.co/docs/datasets/package_reference/main_classes.html#datasets.Dataset.add_item
  > Add item to Dataset.

Other term used in the docs (although less frequently) is "sample". The advantage of this word is that it is also a verb, so we can use the parameter: "sample_by" (if you insist on using a verb instead of a noun).

In summary, these proposals:
- config.row
- config.example
- config.sample
- config.sample_by

lhoestq:
I like `sample_by`. Another idea I had was `separate_by`.

It could also by `sampling_method`, `separation_method`.

Not a big fan of the proposed nouns alone since they are very generic, that's why I tried to have something more specific.

I also agree that we actually should avoid `split` to avoid any confusion

mariosasko:
Thanks for the analysis of the used terms. I also like `sample_by` (`separate_by` is good too).

src/datasets/packaged_modules/text/text.py
albertvillanova: Done

lhoestq:
Thank you !! :D

PR #3443, opened 2021-12-16
Extend iter_archive to support file object input

albertvillanova:
This PR adds support to passing a file object to `[Streaming]DownloadManager.iter_archive`.

With this feature, we can iterate over a tar file inside another tar file.


src/datasets/utils/download_manager.py
```diff
@@ -215,17 +216,18 @@ def _download(self, url_or_filename: str, download_config: DownloadConfig) -> st
             url_or_filename = url_or_path_join(self._base_path, url_or_filename)
         return cached_path(url_or_filename, download_config=download_config)
 
-    def iter_archive(self, path):
-        """Returns iterator over files within archive.
+    def iter_archive(self, path_or_file: Union[str, io.BufferedReader]):
```
lhoestq, line 219: I think we can name the parameter `path_or_buf` like in pandas

lhoestq (approved):
Cool thank you !

PR #3456, opened 2021-12-20
[WER] Better error message for wer

patrickvonplaten:
Currently we have the following problem when using the WER. When the input format to the WER metric is wrong, instead of throwing an error message a word-error-rate is computed which is incorrect. E.g. when doing the following: 

```python

wer = load_metric("wer")

target_str = ["hello this is nice", "hello the weather is bloomy"]
pred_str = [["hello it's nice"], ["hello it's the weather"]]

print("Wrong:", wer.compute(predictions=pred_str, references=target_str))
print("Correct", wer.compute(predictions=[x[0] for x in pred_str], references=target_str))
```
We get:
```
Wrong: 1.0
Correct 0.5555555555555556
```

meaning that we get a word-error rate for incorrectly passed input formats. We should raise an error here instead so that people don't spent hours fixing a model while it's their incorrect evaluation metric is the problem for a low WER.

### DIFF
--- a/metrics/wer/wer.py
+++ b/metrics/wer/wer.py
@@ -94,6 +94,21 @@ class WER(datasets.Metric):
         )
 
     def _compute(self, predictions=None, references=None, concatenate_texts=False):
+        if type(predictions) != type(references):
+            raise ValueError(
+                f"`predictions` {predictions} are of type {type(predictions)}, "
+                f" while `targets` {references} are of type {type(references)}. "
+                "Make sure `predictions` and `targets` are of the same type."
+            )
+
+        inputs_are_lists = isinstance(predictions, (list, tuple)) and isinstance(references, (list, tuple))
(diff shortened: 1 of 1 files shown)


lhoestq:
Hi ! I don't think this would solve this issue.
Currently it looks like there's a bug that converts the list `["hello it's nice"]` to a string `'["hello it's nice"]'` since this is what the metric expects as input. The conversion is done before the data are passed to `_compute()`.

This is `Value("string").encode_example` that is called to do the conversion. Since `str()` encoding is too permissive we should consider raising an error if the example is not a string (even though it can be converted to string).

patrickvonplaten:
> called

Ah yeah you're right

lhoestq:
I just opened https://github.com/huggingface/datasets/pull/3460 to fix that. It now raises an error instead of computing the wrong WER

PR #3460, opened 2021-12-20
Don't encode lists as strings when using `Value("string")`

lhoestq:
Following https://github.com/huggingface/datasets/pull/3456#event-5792250497 it looks like `datasets` can silently convert lists to strings using `str()`, instead of raising an error.

This PR fixes this and should fix the issue with WER showing low values if the input format is not right.


src/datasets/features/features.py
```diff
@@ -306,7 +306,10 @@ def encode_example(self, value):
         elif pa.types.is_floating(self.pa_type):
             return float(value)
         elif pa.types.is_string(self.pa_type):
-            return str(value)
+            if isinstance(value, str):
```
patrickvonplaten, line 309: just out of curiosity - this won't break anything where *e.g.* an int value is passed but assumed to be of type string?

lhoestq: I think we have to discuss this indeed. I'm ok with accepting integers and floats to avoid breaking some scripts

patrickvonplaten: ok maybe just not allow lists then?

patrickvonplaten: In general, I think it's a pretty weird behavior from python that `str([1,2,3])` is allowed

albertvillanova:
Should we close this PR?

lhoestq:
since the original issue has to do with metrics that have been moved to `evaludate` I think we can close this one

PR #3469, opened 2021-12-21
Fix METEOR missing NLTK's omw-1.4

lhoestq:
NLTK 3.6.6 now requires `omw-1.4` to be downloaded for METEOR to work.
This should fix the CI on master


lhoestq:
I also modified the doctest call to raise the exception that doctest may catch, instead of `doctest.UnexpectedException`.
This will make debugging easier if it happens again

lhoestq:
The change from `___` to `--` was allowed by https://github.com/huggingface/moon-landing/pull/1657

PR #3472, opened 2021-12-22
Fix `str(Path(...))` conversion in streaming on Linux

mariosasko:
Fix `str(Path(...))` conversion in streaming on Linux. This should fix the streaming of the `beans` and `cats_vs_dogs` datasets.


albertvillanova (approved):
Thanks for fixing the issue.

I agree this allows working in streaming mode "beans" and "cats_vs_dogs" by fixing a previous malformed path, but as you told me, the source cause of the bug is `str(Path())` in streaming mode.

Maybe we should eventually fix the source cause as well.

PR #3477, opened 2021-12-23
Use `iter_files` instead of `str(Path(...)` in image dataset

mariosasko:
Use `iter_files` in the `beans` and the `cats_vs_dogs` dataset scripts as suggested by @albertvillanova.

Additional changes:
* Fix `iter_files` in `MockDownloadManager` (see this [CI error](https://app.circleci.com/pipelines/github/huggingface/datasets/9247/workflows/2657ff8a-b531-4fd9-a9fc-6541a72e8d83/jobs/57028))
* Add support for `os.path.isdir` and `os.path.isfile` in streaming (`os.path.isfile` is needed in `StreamingDownloadManager`'s `iter_files` to make `cats_vs_dogs` streamable)

TODO:
- [ ] add tests for `xisdir` and `xisfile`

### DIFF
--- a/src/datasets/streaming.py
+++ b/src/datasets/streaming.py
@@ -9,6 +9,8 @@ from .utils.streaming_download_manager import (
     xbasename,
     xdirname,
     xglob,
+    xisdir,
+    xisfile,
     xjoin,
     xlistdir,
     xopen,
@@ -68,6 +70,9 @@ def extend_module_for_streaming(module_path, use_auth_token: Optional[Union[str,
     patch_submodule(module, "os.path.join", xjoin).start()
     patch_submodule(module, "os.path.dirname", xdirname).start()
     patch_submodule(module, "os.path.basename", xbasename).start()
+    # allow checks on paths
+    patch_submodule(module, "os.path.isfile", wrap_auth(xisdir)).start()
+    patch_submodule(module, "os.path.isfile", wrap_auth(xisfile)).start()
     if hasattr(module, "Path"):
         patch.object(module.Path, "joinpath", xpathjoin).start()
         patch.object(module.Path, "__truediv__", xpathjoin).start()
--- a/src/datasets/streaming.py
+++ b/src/datasets/streaming.py
@@ -74,7 +74,7 @@ def extend_module_for_streaming(module_path, use_auth_token: Optional[Union[str,
     patch_submodule(module, "os.path.dirname", xdirname).start()
     patch_submodule(module, "os.path.basename", xbasename).start()
     # allow checks on paths
-    patch_submodule(module, "os.path.isfile", wrap_auth(xisdir)).start()
+    patch_submodule(module, "os.path.isdir", wrap_auth(xisdir)).start()
     patch_submodule(module, "os.path.isfile", wrap_auth(xisfile)).start()
     if hasattr(module, "Path"):
         patch.object(module.Path, "joinpath", xpathjoin).start()
--- a/src/datasets/utils/streaming_download_manager.py
+++ b/src/datasets/utils/streaming_download_manager.py
@@ -9,7 +9,7 @@ from asyncio import TimeoutError
 from io import BytesIO
 from itertools import chain
 from pathlib import Path, PurePath, PurePosixPath
-from typing import List, Optional, Tuple, Union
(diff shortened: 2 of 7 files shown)


src/datasets/utils/streaming_download_manager.py
```diff
@@ -160,6 +160,52 @@ def xbasename(a):
...
+        return os.path.isfile(path)
+    else:
+        if rest_hops and fsspec.get_fs_token_paths(rest_hops[0])[0].protocol == "https":
+            storage_options = {
+                "https": {"headers": get_authentication_headers_for_url(rest_hops[0], use_auth_token=use_auth_token)}
+            }
+        else:
+            storage_options = None
+        fs, *_ = fsspec.get_fs_token_paths(path, storage_options=storage_options)
+        return fs.isfile(path)
```
albertvillanova, line 183:
```suggestion
        return fs.isfile(main_hop)
```

src/datasets/utils/streaming_download_manager.py
```diff
@@ -160,6 +160,52 @@ def xbasename(a):
...
+        return os.path.isdir(path)
+    else:
+        if rest_hops and fsspec.get_fs_token_paths(rest_hops[0])[0].protocol == "https":
+            storage_options = {
+                "https": {"headers": get_authentication_headers_for_url(rest_hops[0], use_auth_token=use_auth_token)}
+            }
+        else:
+            storage_options = None
+        fs, *_ = fsspec.get_fs_token_paths(path, storage_options=storage_options)
+        return fs.isdir(path)
```
albertvillanova, line 206:
```suggestion
        return fs.isdir(main_hop)
```

albertvillanova (changes requested):
Thanks for all the enhancements, @mariosasko.

As discussed, the workaround in `iter_files` was temporal (it was valid only in some limited uses cases) and eventually would need fixing. In that moment I prioritized having a limited-working implementation rather than a general one. 

To have a generally valid implementation, first some fixing is needed in the custom FileSystems we have at `datasets`. I could have a look at that in another PR, so that you could merge it into this once that is fixed.

On the meantime, I have suggested some fixing to your implementation below.

albertvillanova (changes requested):
This PR requires first being merged:
- #3481

albertvillanova (approved):
The failing test has nothing to do with this PR.

lhoestq:
`iter_archive` is about to support ZIP archives. I think we should use this no ?

lhoestq:
I'm not a big fan of having two functions that do the same thing. What do you think ?

albertvillanova:
They do not do the same thing:
- One iterates over files in a directory
- The other I guess will iterate over the members of an archive

lhoestq:
Makes sense ! Sounds good then - sorry for my misunderstanding

Note that `iter_archive` will be more performant for data streaming that `iter_files` thanks to the buffering so maybe in the future we can `iter_archive` for some of these datasets

albertvillanova:
Yes, @lhoestq I agree with you: once `iter_archive` supports zip files, it will be more suitable than `iter_files` for these 2 datasets.

Anyway, this PR also implements `isfile`/`isdir` in streaming mode, besides fixing `iter_files`. And I'm interested in having those in master.

Maybe, could we merge this PR into master and take note to refactor the datasets to use `iter_archive` once zip is supported?
Other option could be to split this PR into 2...

src/datasets/streaming.py
```diff
@@ -71,6 +73,9 @@ def wrapper(*args, **kwargs):
     patch_submodule(module, "os.path.join", xjoin).start()
     patch_submodule(module, "os.path.dirname", xdirname).start()
     patch_submodule(module, "os.path.basename", xbasename).start()
+    # allow checks on paths
+    patch_submodule(module, "os.path.isfile", wrap_auth(xisdir)).start()
```
lhoestq, line 77:
```suggestion
    patch_submodule(module, "os.path.isdir", wrap_auth(xisdir)).start()
```

lhoestq (commented):
Sure, let me add some tests for xisdir and xisfile and we can merge

lhoestq (approved):
Alright let's merge this now :)

PR #3478, opened 2021-12-23
Extend support for streaming datasets that use os.walk

albertvillanova:
This PR extends the support in streaming mode for datasets that use `os.walk`, by patching that function.

This PR adds support for streaming mode to datasets:
1. autshumato
1. code_x_glue_cd_code_to_text
1. code_x_glue_tc_nl_code_search_adv
1. nchlt

CC: @severo


severo:
Nice. I'll update the dataset viewer once merged, and test on these four datasets

PR #3486, opened 2021-12-27
Fix weird spacing in ManualDownloadError message

bryant1410:
`textwrap.dedent` works based on the spaces at the beginning. Before this change, there wasn't any space.


albertvillanova (approved):
Thanks for the fix, @bryant1410.

PR #3492, opened 2021-12-28
Add `gzip` for `to_json`

### LINKED ISSUE #3480: the compression format requested when saving a dataset in json format is not respected
## Describe the bug
In the documentation of the `to_json` method, it is stated in the parameters that 
> **to_json_kwargs – Parameters to pass to pandas’s pandas.DataFrame.to_json.

however when we pass for example `compression="gzip"`, the saved file is not compressed.

Would you also have expected compression to be applied? :relaxed: 

## Steps to reproduce the bug
```python
my_dict = {"a": [1, 2, 3], "b": [1, 2, 3]}
```
### Result with datasets
```python
from datasets import Dataset

dataset = Dataset.from_dict(my_dict)
dataset.to_json("dic_with_datasets.jsonl.gz", compression="gzip")
!cat dic_with_datasets.jsonl.gz
```
output
```
{"a":1,"b":1}
{"a":2,"b":2}
{"a":3,"b":3}
```
Note: I would expected to see binary data here

### Result with pandas
```python
import pandas as pd

df = pd.DataFrame(my_dict)
df.to_json("dic_with_pandas.jsonl.gz", lines=True, orient="records", compression="gzip")
!cat dic_with_pandas.jsonl.gz
```
output
```
4��a�dic_with_pandas.jsonl��VJT�2�QJ��\� ��g��yƵ���������)���
```
Note: It looks like binary data

## Expected results

I would have expected that the saved result with datasets would also be a binary file

## Environment info

- `datasets` version: 1.16.1
- Platform: Linux-4.18.0-193.70.1.el8_2.x86_64-x86_64-with-glibc2.17
- Python version: 3.8.11
- PyArrow version: 5.0.0

bhavitvyamalik:
(Partially) closes #3480. I have added `gzip` compression for `to_json`. I realised we can run into this compression problem with `to_csv` as well. `IOHandler` can be used for `to_csv` too. Please let me know if any changes are required.

### DIFF
--- a/src/datasets/io/handler.py
+++ b/src/datasets/io/handler.py
@@ -1,0 +1,19 @@
+import gzip
+
+
+class IOHandler:
+    def __init__(self, file_name, file_mode, compression):
+        self._file_name = file_name
+        self._file_mode = file_mode
+        self._compression = compression
+
+    def __enter__(self):
+        if self._compression is None:
+            self._file = open(self._file_name, self._file_mode)
+        elif self._compression == "gzip":
+            self._file = gzip.open(self._file_name, self._file_mode)
+
+        return self._file
+
+    def __exit__(self, exc_type, exc_value, exc_traceback):
+        self._file.close()
--- a/src/datasets/io/json.py
+++ b/src/datasets/io/json.py
@@ -8,6 +8,7 @@ from ..packaged_modules.json.json import Json
 from ..utils import logging
 from ..utils.typing import NestedDataStructureLike, PathLike
 from .abc import AbstractDatasetReader
+from .handler import IOHandler
 
 
 class JsonDatasetReader(AbstractDatasetReader):
@@ -81,9 +82,13 @@ class JsonDatasetWriter:
         _ = self.to_json_kwargs.pop("path_or_buf", None)
         orient = self.to_json_kwargs.pop("orient", "records")
         lines = self.to_json_kwargs.pop("lines", True)
+        compression = self.to_json_kwargs.pop("compression", None)
+
+        if compression not in [None, "gzip"]:
+            raise NotImplementedError(f"Datasets currently does not support {compression} compression")
 
         if isinstance(self.path_or_buf, (str, bytes, os.PathLike)):
-            with open(self.path_or_buf, "wb+") as buffer:
+            with IOHandler(self.path_or_buf, "wb", compression) as buffer:
                 written = self._write(file_obj=buffer, orient=orient, lines=lines, **self.to_json_kwargs)
         else:
             written = self._write(file_obj=self.path_or_buf, orient=orient, lines=lines, **self.to_json_kwargs)


src/datasets/io/json.py
```diff
@@ -81,9 +82,13 @@ def write(self) -> int:
         _ = self.to_json_kwargs.pop("path_or_buf", None)
         orient = self.to_json_kwargs.pop("orient", "records")
         lines = self.to_json_kwargs.pop("lines", True)
+        compression = self.to_json_kwargs.pop("compression", None)
+
+        if compression not in [None, "gzip"]:
+            raise NotImplementedError(f"Datasets currently does not support {compression} compression")
```
lhoestq, line 88: we use backticks to refer to the library
```suggestion
            raise NotImplementedError(f"`datasets` currently does not support {compression} compression")
```

src/datasets/io/handler.py
```diff
@@ -0,0 +1,19 @@
+import gzip
+
+
+class IOHandler:
```
lhoestq, line 4: Just a naming suggestion, feel free to ignore if `IOHandler` sounds better to you
```suggestion
class open_compressed:
```

lhoestq: I also think this could simply be a function that returns the output of `open` or `gzip.open`, what do you think ?

lhoestq (commented):
Nice thank you !

src/datasets/io/json.py
```diff
@@ -81,9 +82,13 @@ def write(self) -> int:
...
 
         if isinstance(self.path_or_buf, (str, bytes, os.PathLike)):
-            with open(self.path_or_buf, "wb+") as buffer:
+            with IOHandler(self.path_or_buf, "wb", compression) as buffer:
```
albertvillanova, line 91: Any reason for removing the updating mode  "+"?

albertvillanova (approved):
Yes, indeed one of the advantages of handling directly the `compression` parameter ourselves, is that we can use it for other formats, like CSV or text (and maybe other future supported formats).

src/datasets/io/handler.py
bhavitvyamalik: Not sure about this but I think we'll need to close file too after writing which we aren't doing in `json.py` hence `__exit__` thing in context manager

bhavitvyamalik: `IOHandler` is not just opening compressed files but uncompressed as well :p

src/datasets/io/json.py
bhavitvyamalik: afair `bz2` and `gzip` don't have `wb+` mode. I didn't see any mode ending with + in these compression formats in python docs

lhoestq (approved):
Looks all good ! Thanks a lot @bhavitvyamalik :)

PR #3498, opened 2021-12-28
update `pretty_name`

bhavitvyamalik:
I made a script some time back to fetch `pretty_names` from `papers_with_code` dataset and some rules. Updating them in the `README` of `datasets`. Only changed first 200 files and after some eye balling most of them look good to me!


lhoestq (approved):
Thanks for adding them ! They look all good :)

Issue #1837, opened 2021-02-08
Add VCTK


patrickvonplaten:
## Adding a Dataset
- **Name:** *VCTK*
- **Description:** *This CSTR VCTK Corpus includes speech data uttered by 110 English speakers with various accents. Each speaker reads out about 400 sentences, which were selected from a newspaper, the rainbow passage and an elicitation paragraph used for the speech accent archive.*
- **Paper:** Homepage: https://datashare.ed.ac.uk/handle/10283/3443
- **Data:** https://datashare.ed.ac.uk/handle/10283/3443
- **Motivation:** Important speech dataset
- **TFDatasets Implementation**: https://www.tensorflow.org/datasets/catalog/vctk

If interested in tackling this issue, feel free to tag @patrickvonplaten

Instructions to add a new dataset can be found [here](https://github.com/huggingface/datasets/blob/master/ADD_NEW_DATASET.md).

patrickvonplaten:
That sounds great @jaketae - let me know if you need any help i.e. feel free to ping me on a first PR :-)

PR #3500, opened 2021-12-29
Docs: Add VCTK dataset description

jaketae:
This PR is a very minor followup to #1837, with only docs changes (single comment string).


lhoestq (approved):
thanks :)

PR #3513, opened 2022-01-03
Add desc parameter to filter

### LINKED ISSUE #3317: Add desc parameter to Dataset filter method
As I was filtering very large datasets I noticed the filter method doesn't have the desc parameter which is available in the map method. Why don't we add a desc parameter to the filter method both for consistency and it's nice to give some feedback to users during long operations on Datasets?

Add desc parameter to Dataset filter method

N/A

**Additional context**
N/A

mariosasko:
Fix #3317


lhoestq (approved):
Thank you ! :)

PR #3516, opened 2022-01-03
dataset `asset` - change to raw.githubusercontent.com URLs

VictorSanh:
Changed the URLs to the ones it was automatically re-directing.
Before, the download was failing


mariosasko (approved):
Thanks for fixing this! LGTM!

PR #3520, opened 2022-01-04
Audio datacard update - first pass

meg-huggingface:
Filling out data card "Personal and Sensitive Information" for speech datasets to note PII concerns

### DIFF
--- a/datasets/ami/README.md
+++ b/datasets/ami/README.md
@@ -292,7 +292,7 @@ All information about the dataset creation can be found
 
 ### Personal and Sensitive Information
 
-[Needs More Information]
+The dataset consists of people who have donated their voice online. You agree to not attempt to determine the identity of speakers in this dataset.
 
 ## Considerations for Using the Data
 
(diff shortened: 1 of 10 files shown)


mcmillanmajora:
I'm not sure that we want to change the tags at the top of the cards by hand. Those are used to create the tags in the hub. Although looking at all the tags now, we might want to normalize the current tags again (hyphens or no, ".0" or no). Maybe we could add a binary tag for public domain or not?

meg-huggingface:
> 

That's a good point, I didn't realize these were auto-populated.
At the same time, some of them are wrong -- how/where are they auto-populated?  Seems like we should fix it at that source for the future.
In the mean time, I see that "cc0-1.0" is the desired tag for public domain, so I will change that for now.

patrickvonplaten (approved):
Thanks a lot @meg-huggingface

lhoestq (approved):
Thanks !

PR #3525, opened 2022-01-04
Adding license information for Openbookcorpus

meg-huggingface:
Not entirely sure, following the links here, but it seems the relevant license is at https://github.com/soskek/bookcorpus/blob/master/LICENSE

PR #248, opened 2020-06-07
add Toronto BooksCorpus

richardyy1188:
1. I knew there is a branch `toronto_books_corpus`
 - After I downloaded it, I found it is all non-english, and only have one row. 
- It seems that it cites the wrong paper
- according to papar using it, it is called `BooksCorpus` but not `TornotoBooksCorpus`

2. It use a text mirror in google drive
- `bookscorpus.py` include a function `download_file_from_google_drive` , maybe you will want to put it elsewhere.
- text mirror is found in this [comment on the issue](https://github.com/soskek/bookcorpus/issues/24#issuecomment-556024973), and it said to have the same statistics as the one in the paper.
- You may want to download it and put it on your gs in case of it disappears someday.

3. Copyright ?
The paper has said

> **The BookCorpus Dataset.** In order to train our sentence similarity model we collected a corpus of 11,038 books ***from the web***. These are __**free books written by yet unpublished authors**__. We only included books that had more than 20K words in order to filter out perhaps noisier shorter stories. The dataset has books in 16 different genres, e.g., Romance (2,865 books), Fantasy (1,479), Science fiction (786), Teen (430), etc. Table 2 highlights the summary statistics of our book corpus.

and we have changed the form (not books), so I don't think it should have that problems. Or we can state that use it at your own risk or only for academic use. I know @thomwolf should know these things more.

This should solved #131

PR #3526, opened 2022-01-04
Update license to bookcorpus dataset card

meg-huggingface:
Not entirely sure, following the links here, but it seems the relevant license is at https://github.com/soskek/bookcorpus/blob/master/LICENSE


albertvillanova (commented):
I'm not sure but I would say the MIT license covers the code of the script to generate the dataset, not the dataset itself.

However, I'm not sure what the license of the dataset is. Note the dataset is not publicly accessible.

Maybe @lhoestq, do you have some hint? I have seen you generated the current data file hosted at Hugging Face:
- #248

lhoestq:
The smashwords ToS apply for this dataset, we did the same for https://github.com/huggingface/datasets/pull/3525

lhoestq (approved):

PR #3535, opened 2022-01-05
Add SVHN dataset

mariosasko:
Add the SVHN dataset. 

Additional notes:
* compared to the TFDS implementation, exposes additional the "full numbers" config
* adds the streaming support for `os.path.splitext` and `scipy.io.loadmat` 
* adds `h5py` to the requirements list for the dummy data test


lhoestq (approved):
Awesome ! Looking forward to have it on the Hub :)

PR #3538, opened 2022-01-05
Readme usage update

meg-huggingface:
Noticing that the recent commit throws a lot of errors in the automatic checks.  It looks to me that those errors are simply errors that were already there (metadata issues), unrelated to what I've just changed, but worth another look to make sure.

### DIFF
--- a/datasets/lj_speech/README.md
+++ b/datasets/lj_speech/README.md
@@ -187,7 +187,7 @@ Recordings by Linda Johnson from LibriVox. Alignment and annotation by Keith Ito
 
 ### Personal and Sensitive Information
 
-[Needs More Information]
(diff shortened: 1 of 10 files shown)


yjernite (approved):
lgtm!

datasets/lj_speech/README.md
```diff
@@ -187,7 +187,7 @@ Recordings by Linda Johnson from LibriVox. Alignment and annotation by Keith Ito
 
 ### Personal and Sensitive Information
 
-[Needs More Information]
+The dataset consists of people who have donated their voice online. You agree to not attempt to determine the identity of speakers in this dataset.
```
anton-l, line 190: Can be rephrased/omitted for LJ Speech IMO, since these recordings are done by the authors themselves, and they released them into the public domain.

PR #3542, opened 2022-01-06
Update the CC-100 dataset card

aajanki:
* summary from the dataset homepage
* more details about the data structure
* this dataset does not contain annotations


lhoestq (approved):
Thanks for adding this information :)

PR #3545, opened 2022-01-07
fix: 🐛 pass token when retrieving the split names

severo:


src/datasets/inspect.py
```diff
@@ -271,9 +271,12 @@ def get_dataset_split_names(
     )
     if builder.info.splits is None:
         try:
+            download_config.use_auth_token = use_auth_token
```
severo, line 274: by the way, I'm not sure if I should do this, or if `get_dataset_split_names` should be called with both `use_auth_token` and `download_config` parameters (seems overly complicated)

severo:
Currently, it does not work with https://huggingface.co/datasets/mozilla-foundation/common_voice_7_0/blob/main/common_voice_7_0.py#L146 (which was the goal), because `dl_manager.download_config.use_auth_token` is ignored, and the token is required to be passed using `huggingface-cli login`

src/datasets/inspect.py
lhoestq: Simply passing `download_config` is ok :)

lhoestq, line 274:
```suggestion
            download_config = download_config.copy() if download_config else DownloadConfig()
            download_config.use_auth_token = use_auth_token
```

lhoestq: in case `download_config` is `None`

lhoestq (commented):
Thank you ! It looks all good :)
Just one comment in case `download_config` is `None`:

severo:
> Simply passing download_config is ok :)
hmm, I prefer only passing use_auth_token. But the question is more: is it correct, in the (convoluted) case if `download_config.use_auth_token` exists and is different from `use_auth_token`? Which one should be used?

lhoestq:
If both are passed, `use_auth_token` should have the priority (more specific parameters have the higher priority)

lhoestq (approved):
Thanks ! It looks all good :) Feel free to merge if it's good for you

PR #3546, opened 2022-01-07
Remove print statements in datasets

mariosasko:
This is a second time I'm removing print statements in our datasets, so I've added a test to avoid these issues in the future.


mariosasko:
The CI failures are unrelated to the changes.

lhoestq (approved):
Thanks 🙏

PR #3551, opened 2022-01-07
Add more compression types for `to_json`

bhavitvyamalik:
This PR adds `bz2`, `xz`, and `zip` (WIP) for `to_json`. I also plan to add `infer` like how `pandas` does it


src/datasets/io/handler.py
```diff
@@ -1,7 +1,9 @@
+import bz2
 import gzip
+import lzma
 
 
-class IOHandler:
+class FileWriteHandler:
```
bhavitvyamalik, line 6: @lhoestq I've changed IOHandler to FileWriteHandler as I realised technically we are only writing files (compressed/uncompressed) and not reading them here. What do you think about it?

lhoestq: Sounds good !

bhavitvyamalik:
I looked into how to compress with `zipfile` for which few methods exists, let me know which one looks good:
1. create the file in normal `wb` mode and then zip it separately
2. use `ZipFile.write_str` to write file into the archive. For this we'll need to change how we're writing files from `_write` method  

How `pandas` handles it is that they have created a wrapper for standard library class `ZipFile` and allow the returned file-like handle to accept byte strings via `write` method instead of `write_str` (purpose was to change the name of function by creating that wrapper)

lhoestq:
1. sounds not ideal since it creates an intermediary file.
I like pandas' approach. Is it possible to implement 2. using the pandas class ? Or maybe we can have something similar ?

bhavitvyamalik:
Definitely, @lhoestq! I've adapted that from original code and turns out it is faster than `gz` compression. Apart from that I've also added `infer` option to automatically infer compression type from `path_or_buf` given

bhavitvyamalik:
One small thing, currently I'm assuming that user will provide compression extension in `path_or_buf`. Is it this also possible?
`dataset.to_json("from_dataset.json", compression="zip")`? 
Should I put an `assert` to ensure the file name provided always has a compression extension?

lhoestq:
Thanks !

> One small thing, currently I'm assuming that user will provide compression extension in path_or_buf. Is it this also possible?
>dataset.to_json("from_dataset.json", compression="zip")?
>Should I put an assert to ensure the file name provided always has a compression extension?

I think it's fine as it is right now :) No need to check the extension of the filename passed to `path_or_buf`.

lhoestq:
> turns out it is faster than gz compression

I think the default compression level of `gzip` is 9 in python, which is very slow. Maybe we can switch to compression level 6 instead which is faster, like the `gzip` command on unix

lhoestq:
I found that `fsspec` has something that may interest you: [fsspec.open(..., compression=...)](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.open)

It also has `zip` if I understand correctly ! see https://github.com/fsspec/filesystem_spec/blob/master/fsspec/compression.py#L70

Since `fsspec` is a dependency of `datasets` we can use all this :)

Let me know if you prefer using `fsspec` instead (I haven't tested this yet to write compressed files). IMO it sounds pretty easy to use and it would make the code base simpler

bhavitvyamalik:
Just tried `fsspec` but I'm not able to write compressed `zip` files :/
`gzip`, `xz`, `bz2` are all working fine and it's really simple (no need for `FileWriteHandler` now!)

src/datasets/io/json.py
```diff
@@ -84,11 +90,17 @@ def write(self) -> int:
...
+
+        # Infer compression from the filename/URL extension
+        if compression == "infer":
+            for compression_format, extension in self._compression_to_extension.items():
+                if self.path_or_buf.lower().endswith(extension):
+                    compression = compression_format
 
         if isinstance(self.path_or_buf, (str, bytes, os.PathLike)):
-            with IOHandler(self.path_or_buf, "wb", compression) as buffer:
+            with FileWriteHandler(self.path_or_buf, "wb", compression) as buffer:
```
bhavitvyamalik, line 103: I just changed this to:
`with fsspec.open(self.path_or_buf, "wb", compression=compression,) as buffer:`

bhavitvyamalik: Yeah, for now we can go with `gzip`, `bz2`, `xz`, and `infer`. `compresslevel` also let's handle in next PR (as of now I see same speed for all values)

lhoestq: Ok ! Would it be possible to add a test before we merge this one ?

bhavitvyamalik: Create another compressed file and check file sizes?

tests/io/test_json.py
```diff
@@ -255,3 +257,15 @@ def test_dataset_to_json_orient_invalidproc(self, dataset):
         with pytest.raises(ValueError):
             with io.BytesIO() as buffer:
                 JsonDatasetWriter(dataset, buffer, num_proc=0)
+
+    @pytest.mark.parametrize("compression, extension", [("gzip", "gz"), ("bz2", "bz2"), ("xz", "xz")])
```
bhavitvyamalik, line 261: somehow `gzip` test is failing due to few mismatches

src/datasets/io/json.py
```diff
@@ -83,11 +84,11 @@ def write(self) -> int:
...
-            raise NotImplementedError(f"Datasets currently does not support {compression} compression")
+        if compression not in [None, "infer", "gzip", "bz2", "xz"]:
+            raise NotImplementedError(f"`datasets` currently does not support {compression} compression")
 
         if isinstance(self.path_or_buf, (str, bytes, os.PathLike)):
-            with IOHandler(self.path_or_buf, "wb", compression) as buffer:
+            with fsspec.open(self.path_or_buf, "wb", compression=compression) as buffer:
                 written = self._write(file_obj=buffer, orient=orient, lines=lines, **self.to_json_kwargs)
         else:
             written = self._write(file_obj=self.path_or_buf, orient=orient, lines=lines, **self.to_json_kwargs)
```
lhoestq, line 94:
```suggestion
            if compression:
                raise NotImplementedError(
                    f"The compression parameter is not supported when writing to a buffer, but compression={compression}
                    " was passed. Please provide a local path instead."
                )
            written = self._write(file_obj=self.path_or_buf, orient=orient, lines=lines, **self.to_json_kwargs)
```

lhoestq (approved):
Thanks ! Just adding an error message in case someone passes `compression` for a buffer or file-like object:

PR #3556, opened 2022-01-10
Preserve encoding/decoding with features in `Iterable.map` call

### LINKED ISSUE #3505: cast_column function not working with map function in streaming mode for Audio features
## Describe the bug
I am trying to use Audio class for loading audio features using custom dataset. I am able to cast 'audio' feature into 'Audio' format with cast_column function. On using map function, I am not getting 'Audio' casted feature but getting path of audio file only.
I am getting features of 'audio' of string type with load_dataset call. After using cast_column 'audio' feature is converted into 'Audio' type. But in map function I am not able to get Audio type for audio feature & getting string type data containing path of file only. So I am not able to use processor in encode function.

## Steps to reproduce the bug
```python
# Sample code to reproduce the bug
from datasets import load_dataset, Audio
from transformers import Wav2Vec2Processor

def encode(batch, processor):
    print("Audio: ",batch['audio'])
    batch["input_values"] = processor(batch["audio"]['array'], sampling_rate=16000).input_values
    return batch

def print_ds(ds):
    iterator = iter(ds)
    for d in iterator:
        print("Data: ",d)
        break

processor = Wav2Vec2Processor.from_pretrained(pretrained_model_path)

dataset = load_dataset("custom_dataset.py","train",data_files={'train':'train_path.txt'},
                                data_dir="data", streaming=True, split="train")

print("Features: ",dataset.features)

print_ds(dataset)

dataset = dataset.cast_column("audio", Audio(sampling_rate=16_000))

print("Features: ",dataset.features)

print_ds(dataset)

dataset = dataset.map(lambda x: encode(x,processor))

print("Features: ",dataset.features)

print_ds(dataset)



```

## Expected results

map function not printing Audio type features be used with processor function and getting error in processor call due to this.

## Actual results

# after load_dataset call
Features:  {'sentence': Value(dtype='string', id=None), 'audio': Value(dtype='string', id=None)}
Data:  {'sentence': 'और अपने पेट को माँ की स्वादिष्ट गरमगरम जलेबियाँ हड़पते\n', 'audio': 'data/0116_003.wav'}

# after cast_column call
Features:  {'sentence': Value(dtype='string', id=None), 'audio': Audio(sampling_rate=16000, mono=True, _storage_dtype='string', id=None)}
Data:  {'sentence': 'और अपने पेट को माँ की स्वादिष्ट गरमगरम जलेबियाँ हड़पते\n', 'audio': {'path': 'data/0116_003.wav', 'array': array([ 1.2662281e-06,  1.0264218e-06, -1.3615092e-06, ...,
        1.3017889e-02,  1.0085563e-02,  4.8155054e-03], dtype=float32), 'sampling_rate': 16000}}

# after map call
Features:  None
Audio:  data/0116_003.wav

  File "demo2.py", line 36, in <module>
    print_ds(dataset)
  File "demo2.py", line 11, in print_ds
    for d in iterator:
  File "/opt/conda/lib/python3.7/site-packages/datasets/iterable_dataset.py", line 341, in __iter__
    for key, example in self._iter():
  File "/opt/conda/lib/python3.7/site-packages/datasets/iterable_dataset.py", line 338, in _iter
    yield from ex_iterable
  File "/opt/conda/lib/python3.7/site-packages/datasets/iterable_dataset.py", line 192, in __iter__
    yield key, self.function(example)
  File "demo2.py", line 32, in <lambda>
    dataset = dataset.map(lambda x: batch_encode(x,processor))
  File "demo2.py", line 6, in batch_encode
    batch["input_values"] = processor(batch["audio"]['array'], sampling_rate=16000).input_values
TypeError: string indices must be integers
## Environment info

- `datasets` version: 1.17.0
- Platform: Linux-4.14.243 with-debian-bullseye-sid
- Python version: 3.7.9
- PyArrow version: 6.0.1

mariosasko:
As described in https://github.com/huggingface/datasets/issues/3505#issuecomment-1004755657, this PR uses a generator expression to encode/decode examples with `features` (which are set to None in `map`) before applying a map transform.

Fix #3505


src/datasets/iterable_dataset.py
```diff
@@ -396,8 +412,12 @@ def map(self, function: Callable, batched: bool = False, batch_size: int = 1000)
         """
         info = copy.deepcopy(self._info)
         info.features = None
+        formatter = _features_formatter(self.features)
         ex_iterable = MappedExamplesIterable(
-            self._ex_iterable, function=function, batched=batched, batch_size=batch_size
+            ((key, formatter(example)) for key, example in self._ex_iterable) if self.features else self._ex_iterable,
```
lhoestq, line 417: If I'm not mistaken here you are passing an iterator instead of an iterable.
It means that if you iterate over the dataset once, then the second time your iterate over it, it will be empty.
Maybe we can have a TypedExamplesIterable class that does the feature formatting ?

src/datasets/iterable_dataset.py
```diff
@@ -164,7 +182,11 @@ def shuffle_data_sources(self, seed: Optional[int]) -> "RandomlyCyclingMultiSour
 
 class MappedExamplesIterable(_BaseExamplesIterable):
     def __init__(
-        self, ex_iterable: _BaseExamplesIterable, function: Callable, batched: bool = False, batch_size: int = 1000
+        self,
+        ex_iterable: Union[_BaseExamplesIterable, Generator[Dict[str, Any], None, None]],
```
lhoestq, line 186: Generators are iterators, and so have the same issue as in the previous comment.

lhoestq (commented):
Cool thanks ! I left some comments about the usage of iterators that should be avoided in favor of iterables.

src/datasets/iterable_dataset.py
mariosasko: Oh, I see. Yes, a new class that does the formatting makes sense.

src/datasets/iterable_dataset.py
```diff
@@ -281,6 +281,32 @@ def n_shards(self) -> int:
...
 
 
+class TypedExamplesIterable(_BaseExamplesIterable):
+    def __init__(self, ex_iterable: _BaseExamplesIterable, features: Features):
+        self.ex_iterable = ex_iterable
+        self.features = features
+
+    def __iter__(self):
+        iterator = iter(self.ex_iterable)
+        for key, example in iterator:
```
lhoestq, line 291: (nit) It's maybe simpler this way
```suggestion
        for key, example in self.ex_iterable:
```

lhoestq (approved):
Amazing thanks !

PR #3560, opened 2022-01-10
Run pyupgrade for Python 3.6+

bryant1410:
Run the command:

```bash
pyupgrade $(find . -name "*.py" -type f) --py36-plus
```

Which mainly avoids unnecessary lists creations and also removes unnecessary code for Python 3.6+.

It was originally part of #3489.

Tip for reviewing faster: use the CLI (`git diff`) and scroll.


lhoestq:
Hi ! Thanks for the change :)
Could it be possible to only run it for the code in `src/` ? We try to not change the code in the `datasets/` directory too often since it refreshes the users cache when they upgrade `datasets`.

bryant1410:
> Hi ! Thanks for the change :)
> Could it be possible to only run it for the code in `src/` ? We try to not change the code in the `datasets/` directory too often since it refreshes the users cache when they upgrade `datasets`.

I reverted the changes in `datasets/` instead of changing only `src/`. Does it sound good?

lhoestq (approved):
Thanks ! It looks all good :)

I think using OSError instead of IOError, EnvironmentError and socket.error is fine since they're aliases of OSError anyway. The changes in the way we instantiate NamedTuple objects is fine too since this syntax is available from python 3.6

lhoestq:
I just resolved some conflicts with the master branch. If the CI is green we can merge :)

Issue #2520, opened 2021-06-18
Datasets with tricky task templates


lewtun:
I'm collecting a list of datasets here that don't follow the "standard" taxonomy and require further investigation to implement task templates for.

## Text classification

* [hatexplain](https://huggingface.co/datasets/hatexplain): ostensibly a form of text classification, but not in the standard `(text, target)` format and each sample appears to be tokenized.
* [muchocine](https://huggingface.co/datasets/muchocine): contains two candidate text columns (long-form and summary) which in principle requires two `TextClassification` templates which is not currently supported

mariosasko:
The `task_templates` API is deprecated in favor of the `train-eval-index` YAML field, so I'm closing this issue.

PR #3562, opened 2022-01-10
Allow multiple task templates of the same type

mariosasko:
Add support for multiple task templates of the same type. Fixes (partially) #2520.

CC: @lewtun

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -1838,7 +1838,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         dataset.set_transform(transform=transform, columns=columns, output_all_columns=output_all_columns)
         return dataset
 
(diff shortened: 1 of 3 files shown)


src/datasets/arrow_dataset.py
```diff
@@ -1858,11 +1859,13 @@ def prepare_for_task(self, task: Union[str, TaskTemplate]) -> "Dataset":
             if not compatible_templates:
                 raise ValueError(f"Task {task} is not compatible with this dataset! Available tasks: {tasks}")
 
-            if len(compatible_templates) > 1:
+            if len(compatible_templates) > 1 and id is None:
```
lhoestq, line 1862: We could also return the id=0 by default, but no strong opinion on this

lhoestq (approved):
Cool !

src/datasets/arrow_dataset.py
mariosasko: My reasoning for `id=None` was to make selecting `id` more explicit, but this is also OK (`id=0` selects the most common task for a dataset).

PR #3566, opened 2022-01-12
Add initial electricity time series dataset

kashif:
Here is an initial prototype time series dataset


datasets/electricity/README.md
```diff
@@ -0,0 +1,242 @@
...
+}
+```
+
+We have two configurations `uci` and `lstnet`, which are specificed as as follows. 
+
+The time series are resampled to hourly frequency. We test on 7 rolling windows of prediction lenght of 24. 
+
+The `uci` validation therefore ends 24*7 time steps before the end of each time series. The training split ends 24 time steps before the end of the validation split. 
+
+For the `lsnet` configuration we split the training window to be 0.6 of the full time series and the validation is the 0.8-th time series and the last 0.2 time windows are used as the test set of 7 rolling windows of the 24 time steps. Finally as in the LSTNet paper we only consider time series which are active in the year 2012-2014 which leaves us with 320 time series.
```
mariosasko, line 86: Some typos/grammer:

```suggestion
We have two configurations `uci` and `lstnet`, which are specificed as follows. 

The time series are resampled to hourly frequency. We test on 7 rolling windows of prediction length of 24. 

The `uci` validation therefore ends 24*7 time steps before the end of each time series. The training split ends 24 time steps before the end of the validation split. 

For the `lsnet` configuration we split the training window to be 0.6 of the full-time series and the validation is the 0.8-th time series and the last 0.2 time windows are used as the test set of 7 rolling windows of the 24 time steps. Finally, as in the LSTNet paper, we only consider time series that are active in the year 2012-2014, which leaves us with 320 time series.
```

datasets/electricity/README.md
```diff
@@ -0,0 +1,242 @@
...
+
+
+### Data Fields
+
+For this univariate regular time series we have:
+
+- `start`: a `datetime` as string of the first entry of each time serie in the dataset
+- `target`: an `array[float32]` of the actual target values
+
+Given the `freq` and the `start` date, we can assign a datetime to each entry in the target array. Each time serie is also assigned a unique `feat_static_cat` integer id as well as an identifier `item_id`.
```
mariosasko, line 96: It's preferred to list and describe all the fields as follows: 

```suggestion
- `start`: a `datetime` as string of the first entry of each time serie in the dataset
- `target`: an `array[float32]` of the actual target values
- `feat_static_cat`: [field description]
- `item_id`: [field description]

Given the `freq` and the `start` date, we can assign a datetime to each entry in the target array.
```

(Also it would be nice if you could describe the `feat_stati_cat` field, which is a list of integer ids (you mention `integer id`), in more detail)

datasets/electricity/electricity.py
```diff
@@ -0,0 +1,217 @@
...
+    def _split_generators(self, dl_manager):
+        # dl_manager is a datasets.download.DownloadManager that can be used to download and extract URLS
+        # It can accept any type or nested list/dict and will give back the same structure with the url replaced with path to local files.
+        # By default the archives will be extracted and a path to a cached folder where they are extracted is returned instead of the archive
+        data_dir = dl_manager.download_and_extract(_URL)
+
+        # define the prediction problem # TODO save these in metadata
+        freq = "1H"
+        prediction_length = 24
+        rolling_evaluations = 7
```
mariosasko, line 116: It makes sense to add these variables to the config as attributes to support custom values from the user.
```python
class ElectrictyConfig(datasets.BuilderConfig):
    """[docstring for attributes]"""  
    freq: str = "1H"
    prediction_length: int = 24
    rolling_evaluations: int = 7

...
BUILDER_CONFIGS = [
        ElectrictyConfig(
            name="uci",
            version=VERSION,
            description="Original UCI time series.",
        ),
        ElectrictyConfig(
            name="lstnet",
            version=VERSION,
            description="Electricity time series preporcessed as in LSTNet paper.",
        ),
    ]
```
Then, you can access the values as `self.config.{attr}` in the script (e.g. `self.config.freq`)

We should also add a section to the README where we explain each attribute in more detail (we can also add sample calls to `load_dataset` with these attributes, etc.).

datasets/electricity/electricity.py
```diff
@@ -0,0 +1,217 @@
...
+            # concate the time series to be from 2012 till 2014
+            df = df[(df.index.year >= 2012) & (df.index.year <= 2014)]
+
+            # drop time series which are zero at the start
+            df = df.T[df.iloc[0] > 0].T
+
+            # validation ends at 8/10-th of the time series
+            val_end_date = df.index[int(len(df) * (8 / 10)) - 1]
+            # training ends at 6/10-th of the time series
+            train_end_date = df.index[int(len(df) * (6 / 10)) - 1]
```
mariosasko, line 145: How did you come up with these proportions for the splits?  Ideally, all the data should be in the train split, unless there are predefined splits.

Another option is to add these values to the config to be customizable:
```
val_proportion = 8 / 10
train_proportion = 6 / 10
```

datasets/electricity/electricity.py
```diff
@@ -0,0 +1,217 @@
...
+            description="Electricity time series preporcessed as in LSTNet paper.",
+        ),
+    ]
+
+    DEFAULT_CONFIG_NAME = "lstnet"
+
+    def _info(self):
+        features = datasets.Features(
+            {
+                "start": datasets.Value("string"),
```
mariosasko, line 82: Let's use the timestamp type here (you can find my message on Slack where I explain how to work with timestamps  in `datasets`):
```suggestion
                "start": datasets.Value("timestamp[<unit]"),
```
Please also update the README after that change.

mariosasko (commented):
Very nice work so far!

datasets/electricity/electricity.py
kashif: ah perfect! that makes a lot of sense I believe

mariosasko: Let's add a new section `Dataset Usage` to the card. There you can explain the meaning of the config parameters `freq`, `prediction_length` and `rolling_evaluations` in a format similar to Dataset Fields and add an example call  (e.g. `load_dataset("electricity", "uci", rolling_evaluations=..., ...)`.

mariosasko:
@kashif Some commits on the PR branch are not authored by you, so could you please open a new PR and not use rebase this time :)? You can copy and paste the dataset dir to the new branch.

kashif:
making a new PR

PR #3567, opened 2022-01-12
Fix push to hub to allow individual split push

thomasw21:
# Description of the issue

If one decides to push a split on a datasets repo, he uploads the dataset and overrides the config. However previous config splits end up being lost despite still having the dataset necessary.

The new flow is the following:
 - query the old config from the repo
 - update into a new config (add/overwrite new split for example)
 - push the new config

# Side fix

 - `repo_id` in HfFileSystem was wrongly typed.
 - I've added `indent=2` as it becomes much easier to read now.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -3597,10 +3597,11 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         organization, dataset_name = repo_id.split("/")
         info_to_dump = self.info.copy()
         info_to_dump.download_checksums = None
-        info_to_dump.download_size = uploaded_size
-        info_to_dump.dataset_size = dataset_nbytes
-        info_to_dump.size_in_bytes = uploaded_size + dataset_nbytes
+        info_to_dump.download_size = uploaded_size + self.info.download_size
+        info_to_dump.dataset_size = dataset_nbytes + self.info.dataset_size
+        info_to_dump.size_in_bytes = uploaded_size + dataset_nbytes + self.info.size_in_bytes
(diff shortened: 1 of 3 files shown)


src/datasets/arrow_dataset.py
```diff
@@ -3594,14 +3634,18 @@ def push_to_hub(
...
+
+        old_dataset_infos = self.get_remote_dataset_infos(repo_id=repo_id, token=token, branch=branch)
+
         organization, dataset_name = repo_id.split("/")
         info_to_dump = self.info.copy()
         info_to_dump.download_checksums = None
-        info_to_dump.download_size = uploaded_size
-        info_to_dump.dataset_size = dataset_nbytes
-        info_to_dump.size_in_bytes = uploaded_size + dataset_nbytes
+        info_to_dump.download_size = uploaded_size + old_dataset_infos.download_size
```
thomasw21, line 3647: I think we can add a bunch checks here. Typically if you upload a split with a different set of features, there's currently nothing preventing you from doing do, yet it'll crash when you use `load_dataset()`

src/datasets/arrow_dataset.py
```diff
@@ -3553,6 +3555,46 @@ def delete_file(file):
             )
         return repo_id, split, uploaded_size, dataset_nbytes
 
+    def get_remote_dataset_infos(
```
thomasw21, line 3558: I've hesitated to add this to `DatasetInfo.from_hub()`. The only thing I didn't like about it was that the return value is nullable. I can still make the change if you think it makes more sense.

mariosasko:
This has been addressed in https://github.com/huggingface/datasets/pull/4415. Closing.

PR #3573, opened 2022-01-13
Add Mauve metric

jthickstun:
Add support for the [Mauve](https://github.com/krishnap25/mauve) metric introduced in this [paper](https://arxiv.org/pdf/2102.01454.pdf) (Neurips, 2021).


lhoestq:
Hi ! The CI was failing because `mauve-text` wasn't installed. I added it to the CI setup :)

I also did some minor changes to the script itself, especially to remove `**kwargs` and explicitly mentioned all the supported arguments (this way if someone does a typo with some parameters they get an error)

lhoestq (approved):
Thanks a lot for adding MAUVE. It looks all good :)

Issue #3497, opened 2021-12-28
Changing sampling rate in audio dataset and subsequently mapping with `num_proc > 1` leads to weird bug


patrickvonplaten:
Running: 

```python
from datasets import load_dataset, DatasetDict
import datasets
from transformers import AutoFeatureExtractor

raw_datasets = DatasetDict()

raw_datasets["train"] = load_dataset("common_voice", "ab", split="train")

feature_extractor = AutoFeatureExtractor.from_pretrained("facebook/wav2vec2-base")

raw_datasets = raw_datasets.cast_column(
    "audio", datasets.features.Audio(sampling_rate=feature_extractor.sampling_rate)
)

num_workers = 16

def prepare_dataset(batch):
    sample = batch["audio"]

    inputs = feature_extractor(sample["array"], sampling_rate=sample["sampling_rate"])
    batch["input_values"] = inputs.input_values[0]
    batch["input_length"] = len(batch["input_values"])
    return batch

raw_datasets.map(
    prepare_dataset,
    remove_columns=next(iter(raw_datasets.values())).column_names,
    num_proc=16,
    desc="preprocess datasets",
)
```

gives

```bash
  File "/home/patrick/experiments/run_bug.py", line 25, in <module>
    raw_datasets.map(
  File "/home/patrick/python_bin/datasets/dataset_dict.py", line 492, in map
    {
  File "/home/patrick/python_bin/datasets/dataset_dict.py", line 493, in <dictcomp>
    k: dataset.map(
  File "/home/patrick/python_bin/datasets/arrow_dataset.py", line 2139, in map
    shards = [
  File "/home/patrick/python_bin/datasets/arrow_dataset.py", line 2140, in <listcomp>
    self.shard(num_shards=num_proc, index=rank, contiguous=True, keep_in_memory=keep_in_memory)
  File "/home/patrick/python_bin/datasets/arrow_dataset.py", line 3164, in shard
    return self.select(
  File "/home/patrick/python_bin/datasets/arrow_dataset.py", line 485, in wrapper
    out: Union["Dataset", "DatasetDict"] = func(self, *args, **kwargs)
  File "/home/patrick/python_bin/datasets/fingerprint.py", line 411, in wrapper
    out = func(self, *args, **kwargs)
  File "/home/patrick/python_bin/datasets/arrow_dataset.py", line 2756, in select
    return self._new_dataset_with_indices(indices_buffer=buf_writer.getvalue(), fingerprint=new_fingerprint)
[...]

Issue #3247, opened 2021-11-10
Loading big json dataset raises pyarrow.lib.ArrowNotImplementedError


maxzirps:
## Describe the bug
When trying to create a dataset from a json file with around 25MB, the following error is raised `pyarrow.lib.ArrowNotImplementedError: Unsupported cast from struct<b: int64, c: int64> to struct using function cast_struct`

Splitting the big file into smaller ones and then loading it with the `load_dataset` method did also not work.

Creating a pandas dataframe from it and then loading it with `Dataset.from_pandas` works

## Steps to reproduce the bug
```python
load_dataset("json", data_files="test.json")
```

test.json ~25MB
```json
{"a": {"c": 8, "b": 5}}
{"a": {"b": 7, "c": 6}}
{"a": {"c": 8, "b": 5}}
{"a": {"b": 7, "c": 6}}
{"a": {"c": 8, "b": 5}}
...
```

working.json ~160bytes
```json
{"a": {"c": 8, "b": 5}}
{"a": {"b": 7, "c": 6}}
{"a": {"c": 8, "b": 5}}
{"a": {"b": 7, "c": 6}}
{"a": {"c": 8, "b": 5}}
```

## Expected results
It should load the dataset from the json file without error.

## Actual results
It raises Exception `pyarrow.lib.ArrowNotImplementedError: Unsupported cast from struct<b: int64, c: int64> to struct using function cast_struct`

```
Traceback (most recent call last):
  File "/Users/m/workspace/xxx/project/main.py", line 60, in <module>
    dataset = load_dataset("json", data_files="result.json")
  File "/opt/homebrew/Caskroom/miniforge/base/envs/xxx/lib/python3.9/site-packages/datasets/load.py", line 1627, in load_dataset
    builder_instance.download_and_prepare(
  File "/opt/homebrew/Caskroom/miniforge/base/envs/xxx/lib/python3.9/site-packages/datasets/builder.py", line 607, in download_and_prepare
    self._download_and_prepare(
  File "/opt/homebrew/Caskroom/miniforge/base/envs/xxx/lib/python3.9/site-packages/datasets/builder.py", line 697, in _download_and_prepare
    self._prepare_split(split_generator, **prepare_split_kwargs)
  File "/opt/homebrew/Caskroom/miniforge/base/envs/xxx/lib/python3.9/site-packages/datasets/builder.py", line 1159, in _prepare_split
    writer.write_table(table)
[...]

PR #3575, opened 2022-01-13
Add Arrow type casting to struct for Image and Audio + Support nested casting

lhoestq:
## Intro

1. Currently, it's not possible to have nested features containing Audio or Image. 

2. Moreover one can keep an Arrow array as a StringArray to store paths to images, but such arrays can't be directly concatenated to another image array if it's stored an another Arrow type (typically, a StructType).

3. Allowing several Arrow types for a single HF feature type also leads to bugs like this one #3497 

4. Issues like #3247 are quite frequent and happen when Arrow fails to reorder StructArrays. 

5. Casting Audio feature type is blocking preparation for the ASR task template: https://github.com/huggingface/datasets/pull/3364

All those issues are linked together by the fact that:
- we are limited by the Arrow type casting which is lacking features for nested types.
- and especially for Audio and Image: they are not robust enough for concatenation and feature inference.

## Proposed solution

To fix 1 and 4 I implemented nested array type casting (which is missing in PyArrow).

To fix 2, 3 and 5 while having a simple implementation for nested array type casting, I changed the storage type of Audio and Image to always be a StructType. Also casting from StringType is directly implemented via a new function `cast_storage` that is defined individually for Audio and Image. I also added nested decoding.

### I. Better Arrow data type casting for nested data structures

I implemented new functions `array_cast` and `table_cast` that do the exact same as `pyarrow.Array.cast` or `pyarrow.Table.cast` but support nested struct casting and array re-ordering.

These functions can be used on PyArrow objects, and are already integrated in our own `datasets.table.Table.cast` functions. So one can do `my_dataset.data.cast(pyarrow_schema_with_custom_hf_types)` directly.

### II. New image and audio extension types with custom casting

I used PyArrow extension types to be able to define what casting is allowed or not. For example both StringType->ImageExtensionType and StructType->ImageExtensionType are allowed, via the `cast_storage` method.

I factorized all the PyArrow + Pandas extension stuff in the `base_extension.py` file. This aims at separating the front-facing API code of `datasets` from the Arrow back-end which requires advanced knowledge.

### III. Nested feature decoding

I added a new function `decode_nested_example` to decode image and audio data in nested data structures. For optimization's sake, this function is only called if a column has at least one feature that requires decoding.

## Alternative considered

The casting to struct type could have been done directly with python objects using some Audio and Image methods, but bringing arrow data to python objects is expensive. The Audio and Image types could also have been able to convert the arrow data directly, but this is not convenient to use when casting a full Arrow Table with nested fields. Therefore I decided to keep the Arrow data casting logic in Arrow extension types.

## Future work

This work can be used to allow the ArrayND feature types to be nested too (see issue #887)

## TODO

- [x] fix current tests
- [x] add new tests
- [x] docstrings/comments


lhoestq:
Regarding the tests I'm just missing the FixedSizeListType type casting for ListArray objects, will to it tomorrow as well as adding new tests + docstrings

lhoestq:
While writing some tests I noticed that the ExtensionArray can't be directly concatenated - maybe we can get rid of the extension types/arrays and only keep their storages in native arrow types.

In this case the `cast_storage` functions should be the responsibility of the Image and Audio classes directly. And therefore we would need two never cast to a pyarrow type again but to a HF feature - since they'd end up being the one able to tell what's castable or not. This is fine in my opinion but let me know what you think. I can take care of this on monday I think

lhoestq:
Alright I got rid of all the extension type stuff, I'm writing the new tests now :)

lhoestq:
Tests are done, I'll finish the comments and docstrings tomorrow and set the PR on ready for review once it's done !

src/datasets/arrow_dataset.py
```diff
@@ -636,8 +629,6 @@ def __init__(
 
         if self._data.schema.metadata is not None and "huggingface".encode("utf-8") in self._data.schema.metadata:
             metadata = json.loads(self._data.schema.metadata["huggingface".encode("utf-8")].decode())
-            if "info" in metadata and self.info.features is None:  # try to load features from the arrow file metadata
-                self._info.features = DatasetInfo.from_dict(metadata["info"]).features
```
lhoestq, line 640: this has been moved to Features.from_arrow_schema, called a few lines later

src/datasets/arrow_dataset.py
```diff
@@ -780,7 +771,7 @@ def from_pandas(
             info = DatasetInfo()
         info.features = features
         table = InMemoryTable.from_pandas(
-            df=df, preserve_index=preserve_index, schema=pa.schema(features.type) if features is not None else None
+            df=df, preserve_index=preserve_index, schema=features.arrow_schema if features is not None else None
```
lhoestq, line 774: This is the new way of getting the arrow schema of some features.
Contrary to the old way, it also incorporates the HF metadata about the actual feature types the schema represeent

src/datasets/arrow_dataset.py
```diff
@@ -815,10 +806,12 @@ def from_dict(
         if features is not None:
             mapping = features.encode_batch(mapping)
         mapping = {
-            col: OptimizedTypedSequence(data, type=features.type[col].type if features is not None else None, col=col)
+            col: OptimizedTypedSequence(data, type=features[col] if features is not None else None, col=col)
```
lhoestq, line 809: TypedSequence and OptimizedTypedSequence now take HF feature types as input instead of pyarrow types.

This is because it should be able to cast string to struct for Image and Audio for example, and such casting is defined by the HF feature types

src/datasets/arrow_writer.py
```diff
@@ -337,38 +365,18 @@ def write_examples_on_file(self):
...
-            )
-            pa_array = pa.array(typed_sequence)
-            inferred_type = pa_array.type
-            first_example = pa.array(OptimizedTypedSequence(typed_sequence.data[:1], type=inferred_type))[0]
-            if pa_array[0] != first_example:  # Sanity check (check for overflow in StructArray or ListArray)
-                # This check fails with FloatArrays with nans, which is not what we want, so account for that:
-                if not isinstance(pa_array[0], pa.lib.FloatScalar):
-                    raise OverflowError(
-                        f"There was an overflow in the {type(pa_array)}. Try to reduce writer_batch_size to have batches smaller than 2GB"
-                    )
```
lhoestq, line 366: I temporarily removed this part but I guess this isn't fixed yet on the arrow side ?
So I guess I'll add it back

By any chance, do you know if there's a way to test if the issue still occurs @mariosasko ? I can't find any test for that

lhoestq: I was able to find the test that was used to check if an overflow was happening:
```python
import pyarrow as pa
pa.array([["x" * 1024]] * ((4 << 20) + 1))
```
From pyarrow 3 it doesn't overflow anymore, so I think we can simply remove this check now !

mariosasko:
> While writing some tests I noticed that the ExtensionArray can't be directly concatenated - maybe we can get rid of the extension types/arrays and only keep their storages in native arrow types.
>
>In this case the cast_storage functions should be the responsibility of the Image and Audio classes directly. And therefore we would need two never cast to a pyarrow type again but to a HF feature - since they'd end up being the one able to tell what's castable or not. This is fine in my opinion but let me know what you think. I can take care of this on monday I think

Does this change affect performance?

lhoestq:
> Does this change affect performance?

In general it should have a significant impact on performance since the structure of the features is rarely complex (in general we have <20 features and <2 levels of nesting)

Regarding Audio and Image specifically, casting from a StringArray is a little bit more costly since it creates the "bytes" BinaryArray with `None` values with the same length as the "path" array. From the tests I did locally this is very fast though and shouldn't affect the user experience at the current scale of the audio/image datasets we have. It also requires a little bit of RAM though

lhoestq:
Alright this is ready for review now ! Let me know if you have comments and/or improvements :)

src/datasets/features/image.py
```diff
@@ -226,12 +92,14 @@ def decode_example(self, value):
             raise ImportError("To support decoding images, please install 'Pillow'.")
 
         if isinstance(value, str):
-            path, bytes_ = value, None
+            path, bytes_ = value, None  # TODO(QL, Mario): remove this case ?
```
mariosasko, line 95: Yes, let's remove this because it's no longer needed. Previously, we would need this for the hacky `cast_column` cast from `str` to `PIL.Image`:
```python
ds = Dataset.from_dict({"image": ["path/to/image1", "path/to/image2"]})
ds = ds.cast_column("image", Image())
```

mariosasko (approved):
Great job! Can't wait to use these features. 

Just one nit:

m-aliabbas:
I am facing the issue ArrowNotImplementedError  but no solution is working. Please help me

lhoestq:
Can you open an new issue and share the error message as well as the script you used ? We'd be happy to help :)

PR #3588, opened 2022-01-18
Update HellaSwag README.md

borgr:
Adding information from the git repo and paper that were missing


lhoestq (approved):
Thanks for adding that !

There are still some information missing in the tags, that's why the CI fails.
If you'd like to contribute feel free to take a look at the missing tags

PR #3592, opened 2022-01-18
Add QuickDraw dataset

mariosasko:
Add the QuickDraw dataset.

TODOs:
* [x] add dummy data
* [x] add dataset card
* [x] generate `dataset_info.json`


lhoestq (approved):
Awesome thank you ! LGTM :)

The dummy data are kinda big but I don't see a way to reduce them, so let's keep them this way

The CI fail is unrelated to this PR btw, you can ignore it

Issue #3582, opened 2022-01-15
conll 2003 dataset source url is no longer valid


rcanand:
## Describe the bug
Loading `conll2003` dataset fails because it was removed (just yesterday 1/14/2022) from the location it is looking for.

## Steps to reproduce the bug
```python
from datasets import load_dataset


load_dataset("conll2003")
```

## Expected results
The dataset should load.

## Actual results
It is looking for the dataset at `https://github.com/davidsbatista/NER-datasets/raw/master/CONLL2003/train.txt` but it was removed from there yesterday (see [commit](https://github.com/davidsbatista/NER-datasets/commit/9d8f45cc7331569af8eb3422bbe1c97cbebd5690) that removed the file and related [issue](https://github.com/davidsbatista/NER-datasets/issues/8)). 

- We should replace this with an alternate valid location.
- this is being referenced in the huggingface course chapter 7 [colab notebook](https://colab.research.google.com/github/huggingface/notebooks/blob/master/course/chapter7/section2_pt.ipynb), which is also broken.

```python
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-4-27c956bec93c> in <module>()
      1 from datasets import load_dataset
      2 
----> 3 raw_datasets = load_dataset("conll2003")

11 frames
/usr/local/lib/python3.7/dist-packages/datasets/utils/file_utils.py in get_from_cache(url, cache_dir, force_download, proxies, etag_timeout, resume_download, user_agent, local_files_only, use_etag, max_retries, use_auth_token, ignore_url_params)
    610             )
    611         elif response is not None and response.status_code == 404:
--> 612             raise FileNotFoundError(f"Couldn't find file at {url}")
    613         _raise_if_offline_mode_is_enabled(f"Tried to reach {url}")
    614         if head_error is not None:

FileNotFoundError: Couldn't find file at https://github.com/davidsbatista/NER-datasets/raw/master/CONLL2003/train.txt
```

## Environment info

- `datasets` version:
- Platform:
- Python version:
- PyArrow version:

GuillemGSubies:
I came to open the same issue.

lhoestq:
Thanks for reporting !

I pushed a temporary fix on `master` that uses an URL from a previous commit to access the dataset for now, until we have a better solution

lhoestq:
I changed the URL again to use another host, the fix is available on `master` and we'll probably do a new release of `datasets` tomorrow.

[...]

PR #3602, opened 2022-01-19
Update url for conll2003

lhoestq:
Following https://github.com/huggingface/datasets/issues/3582 I'm changing the download URL of the conll2003 data files, since the previous host doesn't have the authorization to redistribute the data


JonathanSum:
Hi. lhoestq 

image
What is the solution for it?
you can see it is still doesn't work here.
https://colab.research.google.com/drive/1l52FGWuSaOaGYchit4CbmtUSuzNDx_Ok?usp=sharing
Thank you.

lhoestq:
For now you can specify `load_dataset(..., revision="master")` to use the fix on `master`.

We'll also do a new release of `datasets` tomorrow I think

PR #3609, opened 2022-01-20
Fixes to pubmed dataset download function

spacemanidol:
Pubmed has updated its settings for 2022 and thus existing download script does not work.

### DIFF
--- a/datasets/pubmed/pubmed.py
+++ b/datasets/pubmed/pubmed.py
@@ -40,7 +40,7 @@ _LICENSE = ""
 # TODO: Add link to the official dataset URLs here
 # The HuggingFace dataset library don't host the datasets but only point to the original files
 # This can be an arbitrary nested dict/list of URLs (see below in `_split_generators` method)
-_URLs = [f"ftp://ftp.ncbi.nlm.nih.gov/pubmed/baseline/pubmed21n{i:04d}.xml.gz" for i in range(1, 1063)]
+_URLs = [f"ftp://ftp.ncbi.nlm.nih.gov/pubmed/baseline/pubmed22n{i:04d}.xml.gz" for i in range(1, 1114)]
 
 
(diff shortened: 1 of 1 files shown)


lhoestq:
Hi ! I think we can simply add a new configuration for the 2022 data instead of replacing them.
You can add the new configuration here:
```python
    BUILDER_CONFIGS = [
        datasets.BuilderConfig(name="2021", description="The 2021 annual record", version=datasets.Version("1.0.0")),
        datasets.BuilderConfig(name="2022", description="The 2022 annual record", version=datasets.Version("1.0.0")),
    ]
```

And we can have the URLs for these two versions this way:
```python
_URLs = {
    "2021": f"ftp://ftp.ncbi.nlm.nih.gov/pubmed/baseline/pubmed21n{i:04d}.xml.gz" for i in range(1, 1063)],
    "2022": f"ftp://ftp.ncbi.nlm.nih.gov/pubmed/baseline/pubmed22n{i:04d}.xml.gz" for i in range(1, 1114)]
}
```
and depending on the configuration name (you can get it with `self.config.name`) we can pick the URLs of 2021 or the ones of 2022 and pass them to the `dl_manager` in `_split_generators`

Feel free to ping me if you have questions or if I can help !

albertvillanova:
Hi @spacemanidol, thanks for your contribution.

The update of the PubMed dataset URL (besides the update of the corresponding metadata and the dummy data) was already merged to master branch in this other PR:
- #3692 

I'm closing this PR then.

@lhoestq please take into account that 2021 data is no longer accessible: every year PubMed releases the baseline data (containing all previous data until that year) and from that on, they release daily updates.

lhoestq:
> @lhoestq please take into account that 2021 data is no longer accessible: every year PubMed releases the baseline data (containing all previous data until that year) and from that on, they release daily updates.

Oh ok I didn't know, thanks

PR #3612, opened 2022-01-21
wikifix

apergo-ai:
This should get the wikipedia dataloading script back up and running - at least I hope so (tested with language ff and ii)

### DIFF
--- a/datasets/wikipedia/wikipedia.py
+++ b/datasets/wikipedia/wikipedia.py
@@ -23,6 +23,7 @@ import json
 import re
 import xml.etree.cElementTree as etree
 from datetime import datetime
+
 import datasets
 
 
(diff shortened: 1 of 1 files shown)


apergo-ai:
test fail because of dataset_infos.json isn't updated. Unfortunately, I cannot get the datasets-cli locally to execute without error. Would need to troubleshoot, what's missing. Maybe someone else can pick up the stick.

lhoestq:
Hi ! If we change the default date to the latest one, users won't be able to load the "big" languages like english anymore, because it requires an Apache Beam runtime to process them. On the contrary, the old data 20200501 has been processed by Hugging Face so that users don't need to run Apache Beam stuff.

Therefore I'm in favor of not changing the default date until we have processed the latest versions of wikipedia.

Users that want to load other languages or that can use Apache Beam can still pass the `language` and `date` parameter to `load_dataset` if they want anyway:
```python
load_dataset("wikipedia", language="fr", date="20220120")
```

apergo-ai:
in that case you can close the PR

lhoestq:
Ok thanks !

(oh I I just noticed that the dataset card is missing the documentation regarding the language and date parameters, let me add it)

PR #3614, opened 2022-01-21
Minor fixes

mariosasko:
This PR:
* adds "desc" to the `ignore_kwargs` list in `Dataset.filter`
* fixes the default value of `id` in `DatasetDict.prepare_for_task`


lhoestq (approved):
thanks !

PR #3616, opened 2022-01-22
Make streamable the BnL Historical Newspapers dataset

### LINKED ISSUE #3615: Dataset BnL Historical Newspapers does not work in streaming mode
## Describe the bug
When trying to load in streaming mode, it "hangs"...

## Steps to reproduce the bug
```python
ds = load_dataset("bnl_newspapers", split="train", streaming=True)
```

## Expected results
The code should be optimized, so that it works fast in streaming mode.

CC: @davanstrien

albertvillanova:
I've refactored the code in order to make the dataset streamable and to avoid it takes too long:
- I've used `iter_files`

Close #3615

### DIFF
--- a/datasets/bnl_newspapers/README.md
+++ b/datasets/bnl_newspapers/README.md
@@ -16,7 +16,7 @@ licenses:
 - cc0-1.0
 multilinguality:
 - multilingual
-pretty_name: BnL newspapers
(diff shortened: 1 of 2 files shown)


lhoestq (approved):
Thanks for the fix !

PR #3623, opened 2022-01-24
Extend support for streaming datasets that use os.path.relpath

### LINKED ISSUE #3622: Extend support for streaming datasets that use os.path.relpath
Extend support for streaming datasets that use `os.path.relpath`.

This feature will also be useful to yield the relative path of audio or image files.

albertvillanova:
This PR extends the support in streaming mode for datasets that use `os.path.relpath`, by patching that function.

This feature will also be useful to yield the relative path of audio or image files, within an archive or parent dir.

Close #3622.


lhoestq (approved):
Amazing thanks !

Issue #3561, opened 2022-01-10
Cannot load ‘bookcorpusopen’


HUIYINXUE:
## Describe the bug
Cannot load 'bookcorpusopen'

## Steps to reproduce the bug
```python
dataset = load_dataset('bookcorpusopen')
```
or
```python
dataset = load_dataset('bookcorpusopen',script_version='master')
```

## Actual results
ConnectionError: Couldn't reach https://the-eye.eu/public/AI/pile_preliminary_components/books1.tar.gz

## Environment info
- `datasets` version: 1.9.0
- Platform: Linux version 3.10.0-1160.45.1.el7.x86_64
- Python version:  3.6.13
- PyArrow version: 6.0.1

samjgorman:
The host of this copy of the dataset (https://the-eye.eu) has been down for months. 

Finding this dataset is a little esoteric, as the original authors took down the official BookCorpus dataset some time ago.

There are community-created versions of BookCorpus, such as the files hosted in the link below.
https://battle.shawwn.com/sdb/bookcorpus/

And more discussion here:
https://github.com/soskek/bookcorpus

Do we want to remove this dataset entirely?  There's a fair argument for this, given that the official BookCorpus dataset was taken down by the authors.   If not, perhaps can open a PR with the link to the community-created tar above and updated dataset description.

mariosasko:
Hi! The `bookcorpusopen` dataset is not working for the same reason as explained in this comment: https://github.com/huggingface/datasets/issues/3504#issuecomment-1004564980

albertvillanova:
Hi @HUIYINXUE, it should work now that the data owners created a mirror server with all data.

PR #3627, opened 2022-01-25
Fix host URL in The Pile datasets

### LINKED ISSUE #3626: The Pile cannot connect to host
## Describe the bug
The Pile had issues with their previous host server and have mirrored its content to another server.

The new URL server should be updated.

albertvillanova:
This PR fixes the host URL in The Pile datasets, once they have mirrored their data in another server.

Fix #3626.


mariosasko:
We should also update the `bookcorpusopen` download url (see #3561) , no?

albertvillanova:
For `the_pile_openwebtext2` and `the_pile_stack_exchange` I did not regenerate the JSON files, but instead i just changed the download_checksums URL.

TrentBrick:
Also if I git clone and edit the repo or reset it before this PR it is still trying to pull using mystic? Why is this?

PR #3641, opened 2022-01-27
Fix numpy rngs when seed is None

### LINKED ISSUE #3634: Dataset.shuffle(seed=None) gives fixed row permutation
## Describe the bug
Repeated attempts to `shuffle` a dataset without specifying a seed give the same results.

## Steps to reproduce the bug
```python
import datasets

# Some toy example
data = datasets.Dataset.from_dict(
    {"feature": [1, 2, 3, 4, 5], "label": ["a", "b", "c", "d", "e"]}
)

# Doesn't work as expected
print("Shuffle dataset")
for _ in range(3):
    print(data.shuffle(seed=None)[:])

# This seems to work with pandas
print("\nShuffle via pandas")
for _ in range(3):
    df = data.to_pandas().sample(frac=1.0)
    print(datasets.Dataset.from_pandas(df, preserve_index=False)[:])

```

## Expected results
I assumed that the default setting would initialize a new/random state of a `np.random.BitGenerator` (see [docs](https://huggingface.co/docs/datasets/package_reference/main_classes.html?highlight=shuffle#datasets.Dataset.shuffle)).

Wouldn't that reshuffle the rows each time I call `data.shuffle()`?

## Actual results

```bash
Shuffle dataset
{'feature': [5, 1, 3, 2, 4], 'label': ['e', 'a', 'c', 'b', 'd']}
{'feature': [5, 1, 3, 2, 4], 'label': ['e', 'a', 'c', 'b', 'd']}
{'feature': [5, 1, 3, 2, 4], 'label': ['e', 'a', 'c', 'b', 'd']}

Shuffle via pandas
{'feature': [4, 2, 3, 1, 5], 'label': ['d', 'b', 'c', 'a', 'e']}
{'feature': [2, 5, 3, 4, 1], 'label': ['b', 'e', 'c', 'd', 'a']}
{'feature': [5, 2, 3, 1, 4], 'label': ['e', 'b', 'c', 'a', 'd']}
```

## Environment info

- `datasets` version: 1.18.0
- Platform: Linux-5.13.0-27-generic-x86_64-with-glibc2.17
- Python version: 3.8.12
- PyArrow version: 6.0.1

mariosasko:
Fixes the NumPy RNG when `seed` is `None`.

The problem becomes obvious after reading the NumPy notes on RNG (returned by `np.random.get_state()`):
> The MT19937 state vector consists of a 624-element array of 32-bit unsigned integers plus a single integer value between 0 and 624 that indexes the current position within the main array.

`The MT19937 state vector`: the seed which we currently index, but this value stays the same for multiple rounds.
`plus a single integer value`: the `pos` value in this PR (is 624 if `seed` is set to a fixed value with `np.random.seed`, so we take the first value in the `seed` array returned by `np.random.get_state()`: https://stackoverflow.com/questions/32172054/how-can-i-retrieve-the-current-seed-of-numpys-random-number-generator)

NumPy notes: https://numpy.org/doc/stable/reference/random/bit_generators/mt19937.html

Fix #3634

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -2916,8 +2916,11 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         if generator is None:
             if seed is None:
                 _, seed, pos, *_ = np.random.get_state()
-                seed = seed[0] if pos == 624 else seed[pos]
-                _ = np.random.random()  # do 1 step of rng
+                if pos < 624:
+                    seed = seed[pos]
+                    _ = np.random.random()  # do 1 step of rng
+                else:
+                    seed = seed[0]  # fixed seed
             generator = np.random.default_rng(seed)
 
         # Check if we've already cached this computation (indexed by a hash)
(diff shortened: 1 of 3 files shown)


lhoestq (approved):
Thanks ! That makes total sense :)

cc @Rocketknight1 I recall you had a similar issue a few months ago, when shuffling several times wouldn't change anything

PR #3650, opened 2022-01-30
Allow 'to_json' to run in unordered fashion in order to lower memory footprint

thomasw21:
I'm using `to_json(..., num_proc=num_proc, compressiong='gzip')` with `num_proc>1`. I'm having an issue where things seem to deadlock at some point. Eventually I see OOM. I'm guessing it's an issue where one process starts to take a long time for a specific batch, and so other process keep accumulating their results in memory.

In order to flush memory, I propose we use optional `imap_unordered`. This will prevent one process to block the other ones. The logical thinking is that index are rarily relevant, and in one wants to keep an index, one can still create another column and reconstruct from there.

### DIFF
--- a/src/datasets/io/json.py
+++ b/src/datasets/io/json.py
@@ -80,7 +80,7 @@ class JsonDatasetWriter:
         self.to_json_kwargs = to_json_kwargs
 
         if (num_proc is None or num_proc == 1) and map_unordered:
-            raise ValueError(f"Unordered map is only defined when using multiprocessing.")
+            raise ValueError("Unordered map is only defined when using multiprocessing.")
         self.map_unordered = map_unordered
 
     def write(self) -> int:
--- a/src/datasets/io/json.py
+++ b/src/datasets/io/json.py
@@ -65,6 +65,7 @@ class JsonDatasetWriter:
         dataset: Dataset,
         path_or_buf: Union[PathLike, BinaryIO],
         batch_size: Optional[int] = None,
+        map_unordered: bool = False,
         num_proc: Optional[int] = None,
         **to_json_kwargs,
     ):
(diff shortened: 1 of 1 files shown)


bhavitvyamalik:
Hi @thomasw21, I remember suggesting `imap_unordered` to @lhoestq at that time to speed up and less memory utilisation but with `pool_imap` on multiple datasets (>9GB) , memory utilisation was almost constant and we decided to go with that only. 

1. Did you try this without `gzip`? Because `gzip`  feature was introduced recently and I didn't check multi_proc thing with `gzip`. One thing I know is that `gzip` is slow in our implementation than `zip` (it's a WIP #3551) 
2. You can try reducing your batch size, this can also help in avoiding OOM errors!

thomasw21:
Thanks @bhavitvyamalik ! I see. I'm not sure this PR actually fixes things for me either (I ended up reducing the num_proc/batch_size to lower it). It does allow the process to run for longer, but I think the reason why it was waiting is that one of the process crashes .... Unfortunately I was working on a setup with a low RAM/cpu core ratio. I'm actually very surprised that it doesn't change memory utilization, otherwise I don't see the purpose of `imap_unordered` existing. I think it's main purpose are when you have high variance in samples (in terms of bytes), which causes unecessary accumulation in `imap`
  1. Did not try without `gzip`
  2. Yeah or `num_proc`

bhavitvyamalik:
Can you please try without `gzip` to see how it performs? If it works fine then we can improve `gzip` from our side (I'm already working on it)

thomasw21:
I'll be busy for next few weeks on another project, will do as soon as I have some bandwidth.

albertvillanova:
Should we close this PR?

thomasw21:
Yes we can close this PR if considered unneeded.

PR #3652, opened 2022-01-31
sp. Columbia => Colombia

serapio:
"Columbia" is various places in North America. The country is "Colombia".


lhoestq (approved):
good catch ! Thanks for the fix

cahya-wirawan:
The original openslr site mixed both names https://openslr.org/72/ :-)

serapio:
Yeah, I filed the issue to have it fixed there last year, but it looks like they missed a few.

PR #3661, opened 2022-02-01
Remove unnecessary 'r' arg in

bryant1410:
Originally from #3489


lhoestq (approved):
Thanks !

lhoestq:
The CI failure is only because of the datasets is missing some sections in their cards - we can ignore that since it's unrelated to this PR

PR #3666, opened 2022-02-02
Multilingual Spoken Words

polinaeterna:


polinaeterna:
@lhoestq  I still have problems with processing `.opus` files with `soundfile` so I actually cannot fully check that it works but it should... Maybe this should be investigated in case of someone else would also have problems with that.

Also, as the data is in a private repo on the hub (before we come to a decision about audio data privacy), the needed checks cannot be done right now.

src/datasets/features/audio.py
```diff
@@ -95,6 +95,11 @@ def decode_example(self, value: dict) -> dict:
             raise ValueError(f"An audio sample should have one of 'path' or 'bytes' but both are None in {value}.")
         elif path is not None and path.endswith("mp3"):
             array, sampling_rate = self._decode_mp3(file if file else path)
+        elif path is not None and path.endswith("opus"):
```
polinaeterna, line 98: I can check if a file is in opus format only if `path` is not None. actually I don't understand in what we can have an empty path (I though that in the loading script we always provide either full path to file or relative path)? But I saw that audio feature tests check the case with empty path.

datasets/ml_spoken_words/README.md
```diff
@@ -0,0 +1,288 @@
...
+
+[More Information Needed]
+
+## Additional Information
+
+### Dataset Curators
+
+[More Information Needed]
+
+### Licensing Information
```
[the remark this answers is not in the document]
lhoestq: Why not ! Note that the license tag is already mandatory in `test_changed_dataset_card`

src/datasets/features/audio.py
lhoestq: It can happen when the file is not available locally (e.g. in streaming mode) - in this case one can pass a path or just the file name so that we can know the extension of the file, and therefore be able to decode it.

Note that in the future we can stop using `path` to check whether a file is in OPUS or MP3 or whatever, and check the [magic number](https://en.wikipedia.org/wiki/List_of_file_signatures) of the file instead

src/datasets/features/audio.py
```diff
@@ -135,24 +140,44 @@ def cast_storage(self, storage: Union[pa.StringArray, pa.StructArray]) -> pa.Str
...
-    def _decode_non_mp3_file_like(self, file):
+    def _decode_non_mp3_file_like(self, file, format=None):
         try:
             import librosa
             import soundfile as sf
         except ImportError as err:
             raise ImportError("To support decoding audio files, please install 'librosa' and 'soundfile'.") from err
 
-        array, sampling_rate = sf.read(file)
+        try:
```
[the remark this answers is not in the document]
lhoestq: I think you can check `soundfile.__libsndfile_version__` instead

```python
from packaging import version

if version.parse(soundfile.__libsndfile_version__) >= version.parse("1.0.30"):
    ...
```

datasets/ml_spoken_words/README.md
```diff
@@ -0,0 +1,297 @@
...
+- uk
+- vi
+- zh-CN
+licenses:
+- cc-by-4.0
+multilinguality:
+- multilingual
+pretty_name: MultilingualSpokenWords
+size_categories:
+- unknown
```
lhoestq, line 63:
```suggestion
- 10M<n<100M
```

datasets/ml_spoken_words/ml_spoken_words.py
```diff
@@ -0,0 +1,245 @@
...
+    def _split_generators(self, dl_manager):
+        splits_archive_path = [dl_manager.download(_SPLITS_URL.format(lang=lang)) for lang in self.config.languages]
+        download_audio = partial(_download_audio_archives, dl_manager=dl_manager)
+
+        return [
+            datasets.SplitGenerator(
+                name=datasets.Split.TRAIN,
+                gen_kwargs={
+                    "audio_archives": [download_audio(split="train", lang=lang) for lang in self.config.languages],
+                    "splits_archive": [dl_manager.iter_archive(path) for path in splits_archive_path],
```
lhoestq, line 178: I think currently this would cause an issue when shuffling in streaming mode. Indeed shuffling does shuffle the order of the lists that are passed to _split_generators because each list would be shuffled differently. I can fix this by making them be shuffled the exact same way.

The alternative would be to use a single list of `zip(audio_archives, splits_archive)` but I feel like it would make more sense if `datasets` takes care of that naturally

lhoestq (commented):
Cool thanks ! I left my final suggestions :)

Once everything is ok, we'll need to merge **only the changes in the src/datasets/\* files**, and remove the files from datasets/ml_spoken_words from this PR since they're going to be added on the Hub directly. Therefore after applying the latest changes, feel free to remove the files from datasets/ml_spoken_words/* from this PR, so we can merge it :)

datasets/ml_spoken_words/ml_spoken_words.py
polinaeterna: but is it a problem in this specific case? here a single splits_archive contains three csv files with metadata for train / val /test splits, so I iterate over them to take only one csv that corresponds to the current split. (I made them tar.gz archives because some csv files for large languages like English can be up to 1Gb in size)

lhoestq:
> But samples are generated from scratch each time. Is it a supposed behavior?

Yea that's the way it works right now, because we generate one arrow file per configuration. Since changing the languages creates a new configuration, then it generates a new arrow file.

datasets/ml_spoken_words/ml_spoken_words.py
lhoestq: > you mean zip(audio_archives, splits_archive) should be passed directly to the _generate_examples, as a single parameter, right?

Yup that would work. Anyway I opened this PR to fix that for every dataset, because I don't like that it could cause silent issues:

https://github.com/huggingface/datasets/issues/3744

lhoestq: Feel free to zip them and pass them as a single parameter for now, so that users that have an older version of the lib don't have issues. But in the future we should be able to have them in two separate lists :)

polinaeterna: heh it looks much uglier that way :( note that I can't refactor `[dl_manager.iter_archive(path) for path in splits_archive_path]` to a single variable as it would lead to stop iteration and the next splits could not be prepared

datasets/ml_spoken_words/ml_spoken_words.py
```diff
@@ -0,0 +1,245 @@
...
+            ),
+        ]
+
+    def _generate_examples(self, audio_archives, splits_archive, split):
+        metadata = dict()
+        for lang_idx, lang in enumerate(self.config.languages):
+            for split_filename, split_file in splits_archive[lang_idx]:
+                if split_filename.split(".csv")[0] == split:
+                    # TODO: how to correctly process csv files from tar?
+                    csv_reader = csv.reader([line.decode("utf-8") for line in split_file.readlines()], delimiter=",")
```
polinaeterna, line 215: @lhoestq is it ok to parse csv files like that?

datasets/ml_spoken_words/ml_spoken_words.py
polinaeterna: maybe it wasn't a good idea to put all the splits csv metadata files into a single tar.gz. archive... my motivation was that since the dataset is already very large I should compress all that I can compress 😈

polinaeterna: @lhoestq I checked the shuffling PR https://github.com/huggingface/datasets/pull/3746  and it seems that we misunderstood each other. I guess in my case here I actually don't need zipping archives since they are not aligned (each `splits_archive` contains only three files: `train.csv`, `val.csv`, `test.csv`, each audio_archive contains all the audio files) and their order actually doesn't matter. agree? can I roll back?

lhoestq: Oh ok my bad ! Sure feel free to roll back

src/datasets/features/audio.py
```diff
@@ -162,23 +168,38 @@ def path_to_bytes(path):
...
             import librosa
             import soundfile as sf
         except ImportError as err:
             raise ImportError("To support decoding audio files, please install 'librosa' and 'soundfile'.") from err
 
+        if format == "opus":
+            if version.parse(sf.__libsndfile_version__) < version.parse("1.0.30"):
+                raise RuntimeError(
+                    "Decoding .opus files requires 'libsndfile'>=1.0.30, "
+                    + "it can be installed via conda: `conda install -c conda-forge libsndfile==1.0.30`"
```
lhoestq, line 201: let's allow newer patch releases (they fixed some important security issues iirc)
```suggestion
                    + "it can be installed via conda: `conda install -c conda-forge libsndfile>=1.0.30`"
```

lhoestq (commented):
Alright thanks :)

To make sure this keeps working as expected in the future, could you also add a test in `test/features/test_audio.py` similar to `test_audio_decode_example_mp3` for OPUS ? This idea is just make sure that `Audio().decode_example` works when an OPUS file is passed as input.

You can add a dummy OPUS file in `test/features/data` and copy/paste + adapt the code from `test_audio_decode_example_mp3`.

tests/features/test_audio.py
```diff
@@ -114,6 +114,17 @@ def test_audio_decode_example_mp3(shared_datadir):
     assert decoded_example["sampling_rate"] == 44100
 
 
+@require_libsndfile
```
lhoestq, line 117: (nit) I would rename it `require_libsndfile_with_opus` to make it explicit that it's just just checking if `libsndfile` is available, but also checking that the installed version does support OPUS
```suggestion
@require_libsndfile_with_opus
```

tests/features/test_audio.py
```diff
@@ -7,7 +7,7 @@
 from datasets import Dataset, concatenate_datasets, load_dataset
 from datasets.features import Audio, Features, Sequence, Value
 
-from ..utils import require_sndfile, require_sox, require_torchaudio
+from ..utils import require_libsndfile, require_sndfile, require_sox, require_torchaudio
```
lhoestq, line 10:
```suggestion
from ..utils import require_libsndfile_with_opus, require_sndfile, require_sox, require_torchaudio
```

tests/utils.py
```diff
@@ -156,6 +157,28 @@ def require_sndfile(test_case):
     return test_case
 
 
+def require_libsndfile(test_case):
```
lhoestq, line 160:
```suggestion
def require_libsndfile_with_opus(test_case):
```

lhoestq (approved):
Thanks :)

Feel free to merge when it's good for you !

PR #3667, opened 2022-02-02
Process .opus files with torchaudio

polinaeterna:
@anton-l suggested to proccess .opus files with `torchaudio` instead of `soundfile` as it's faster: 
opus

(moreover, I didn't manage to load .opus files with `soundfile` / `librosa` locally on any my machine anyway for some reason, even with `ffmpeg` installed).

For now my current changes work with locally stored file:
```python
# download sample opus file (from MultilingualSpokenWords dataset)
!wget https://huggingface.co/datasets/polinaeterna/test_opus/resolve/main/common_voice_tt_17737010.opus 

from datasets import Dataset, Audio

audio_path = "common_voice_tt_17737010.opus"
dataset = Dataset.from_dict({"audio": [audio_path]}).cast_column("audio", Audio(48000))
dataset[0]
# {'audio': {'path': 'common_voice_tt_17737010.opus',
#   'array': array([ 0.0000000e+00,  0.0000000e+00,  3.0517578e-05, ...,
#          -6.1035156e-05,  6.1035156e-05,  0.0000000e+00], dtype=float32),
#   'sampling_rate': 48000}}
```
But it doesn't work when loading inside s dataset from bytes (I checked on [MultilingualSpokenWords](https://github.com/huggingface/datasets/pull/3666), the PR is a draft now, maybe the bug is somewhere there  )

```python
import torchaudio
with open(audio_path, "rb") as b:
    print(torchaudio.load(b))
# RuntimeError: Error loading audio file: failed to open file <in memory buffer>
```

### DIFF
--- a/src/datasets/features/audio.py
+++ b/src/datasets/features/audio.py
@@ -91,10 +91,11 @@ class Audio:
             raise RuntimeError("Decoding is disabled for this feature. Please use Audio(decode=True) instead.")
 
         path, file = (value["path"], BytesIO(value["bytes"])) if value["bytes"] is not None else (value["path"], None)
+        extension = path.split(".")[-1]
         if path is None and file is None:
             raise ValueError(f"An audio sample should have one of 'path' or 'bytes' but both are None in {value}.")
-        elif path is not None and path.endswith("mp3"):
-            array, sampling_rate = self._decode_mp3(file if file else path)
+        elif path is not None and extension in ["mp3", "opus"]:
+            array, sampling_rate = self._decode_mp3(file if file else path, extension)
         else:
             if file:
                 array, sampling_rate = self._decode_non_mp3_file_like(file)
@@ -161,7 +162,7 @@ class Audio:
             sampling_rate = self.sampling_rate
         return array, sampling_rate
 
-    def _decode_mp3(self, path_or_file):
+    def _decode_mp3(self, path_or_file, extension):
         try:
             import torchaudio
             import torchaudio.transforms as T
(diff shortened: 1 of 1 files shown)


src/datasets/features/audio.py
```diff
@@ -91,10 +91,11 @@ def decode_example(self, value: dict) -> dict:
             raise RuntimeError("Decoding is disabled for this feature. Please use Audio(decode=True) instead.")
 
         path, file = (value["path"], BytesIO(value["bytes"])) if value["bytes"] is not None else (value["path"], None)
+        extension = path.split(".")[-1]
```
lhoestq, line 94: The path can be None here:
```suggestion
        extension = path.split(".")[-1] if path is not None else None
```

lhoestq (commented):
Sounds good to me :) thanks !

lhoestq:
Note that torchaudio is maybe less practical to use for TF or JAX users.
This is not in the scope of this PR, but in the future if we manage to find a way to let the user control the decoding it would be nice

polinaeterna:
> Note that torchaudio is maybe less practical to use for TF or JAX users. This is not in the scope of this PR, but in the future if we manage to find a way to let the user control the decoding it would be nice

@lhoestq so maybe don't do this PR? :) if it doesn't work anyway with an opened file, only with path

lhoestq:
Yes as discussed offline there seems to be issues with torchaudio on opened files. Feel free to close this PR if it's better to stick with soundfile because of that

mariosasko:
We should be able to remove torchaudio, which has torch as a hard dependency, soon and use only soundfile for decoding: https://github.com/bastibe/python-soundfile/issues/252#issuecomment-1000246773 (opus + mp3 support is on the way).

Issue #3013, opened 2021-10-04
Improve `get_dataset_infos`?


severo:
Using the dedicated function `get_dataset_infos` on a dataset that has no dataset-info.json file returns an empty info:

```
>>> from datasets import get_dataset_infos
>>> get_dataset_infos('wit')
{}
```

While it's totally possible to get it (regenerate it) with:

```
>>> from datasets import load_dataset_builder
>>> builder = load_dataset_builder('wit')
>>> builder.info
[...]

PR #3670, opened 2022-02-02
feat: 🎸 generate info if dataset_infos.json does not exist

severo:
in get_dataset_infos(). Also: add the `use_auth_token` parameter, and create get_dataset_config_info()

✅ Closes: #3013

### DIFF
--- a/src/datasets/inspect.py
+++ b/src/datasets/inspect.py
@@ -124,14 +124,12 @@ def inspect_metric(path: str, local_path: str, download_config: Optional[Downloa
 
 def get_dataset_infos(
     path: str,
-    revision: Optional[Union[str, Version]] = None,
+    data_files: Optional[Union[Dict, List, str]] = None,
     download_config: Optional[DownloadConfig] = None,
     download_mode: Optional[GenerateMode] = None,
-    force_local_path: Optional[str] = None,
-    dynamic_modules_path: Optional[str] = None,
-    data_files: Optional[Union[Dict, List, str]] = None,
+    revision: Optional[Union[str, Version]] = None,
     use_auth_token: Optional[Union[bool, str]] = None,
-    **download_kwargs,
+    **config_kwargs,
 ):
     """Get the meta information about a dataset, returned as a dict mapping config name to DatasetInfoDict.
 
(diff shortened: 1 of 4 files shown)


severo:
It's a first attempt at solving https://github.com/huggingface/datasets/issues/3013.

src/datasets/inspect.py
```diff
@@ -154,21 +156,36 @@ def get_dataset_infos(
...
+            revision=revision,
+            download_config=download_config,
+            download_mode=download_mode,
+            data_files=data_files,
+            config_name=config_name,
+            data_dir=None,
+            cache_dir=None,
+            features=None,
+            use_auth_token=use_auth_token,
+            # **download_kwargs,
```
lhoestq, line 185: You can remove this line

lhoestq (commented):
Cool thank you ! Do you mind adding some tests for these functions ?

src/datasets/inspect.py
lhoestq: I think we need at least those ones: path/revision/config_name (obviously), download_config/download_mode_use_auth_token (it downloads the dataset script from the hub), data_files/data_dir (splits may depend on these)

So maybe these two could be potentially removed:
- cache_dir: where to download data files and arrow files (not necessary when we just check the splits or the configs)
- features: overwrite the default feature type of the dataset (completely useless in this case)

what do you think ?

severo: data_dir: in this function (get_dataset_infos) it does not make sense either, right, since it's related to a config?

severo: My point the mismatch between the two function signatures:

```
def get_dataset_infos(
    path: str,
    revision: Optional[Union[str, Version]] = None,
    download_mode: Optional[GenerateMode] = None,
    force_local_path: Optional[str] = None,
    dynamic_modules_path: Optional[str] = None,
    data_files: Optional[Union[Dict, List, str]] = None,
    **download_kwargs,
):
```

and 

```
def get_dataset_config_info(
    path: str,
    config_name: Optional[str] = None,
    data_dir: Optional[str] = None,
    data_files: Optional[Union[str, Sequence[str], Mapping[str, Union[str, Sequence[str]]]]] = None,
    cache_dir: Optional[str] = None,
    download_mode: Optional[GenerateMode] = None,
    revision: Optional[Union[str, Version]] = None,
    **config_kwargs,
) -> DatasetInfo:
```

The first one calls the second one

lhoestq: I'll check what parameters we need and update the PR, will ping you once it's done :)
Thanks for raising this concern

severo: thank you!

lhoestq:
I only kept these ones:
```
    path: str,
    data_files: Optional[Union[Dict, List, str]] = None,
    download_mode: Optional[GenerateMode] = None,
    revision: Optional[Union[str, Version]] = None,
    **config_kwargs,
```

Let me know if it's better for you now !

(note that there's no breaking change since the ones that are removed can be passed as config_kwargs if you really want)

severo:
(https://github.com/huggingface/datasets/pull/3670/commits/07c3f0800dd34dfebb9674ad46c67a907b08ded8 -> I has forgotten to update black in my venv)

PR #3674, opened 2022-02-03
Add FrugalScore metric

moussaKam:
This pull request add FrugalScore metric for NLG systems evaluation.

FrugalScore is a reference-based metric for NLG models evaluation. It is based on a distillation approach that allows to learn a fixed, low cost version of any expensive NLG metric, while retaining most of its original performance.

Paper: https://arxiv.org/abs/2110.08559?context=cs
Github: https://github.com/moussaKam/FrugalScore

@lhoestq


metrics/frugalscore/frugalscore.py
```diff
@@ -0,0 +1,96 @@
...
+"""FrugalScore metric."""
+
+from transformers import AutoModelForSequenceClassification, AutoTokenizer, Trainer, TrainingArguments
+
+import datasets
+
+
+_CITATION = """\
+@article{eddine2021frugalscore,
+  title={FrugalScore: Learning Cheaper, Lighter and Faster Evaluation Metricsfor Automatic Text Generation},
```
lhoestq, line 23: a space is missing
```suggestion
  title={FrugalScore: Learning Cheaper, Lighter and Faster Evaluation Metrics for Automatic Text Generation},
```

metrics/frugalscore/frugalscore.py
```diff
@@ -0,0 +1,96 @@
...
+    references (list of str): list of reference for each prediction. Each
+        reference should be a string.
+    pretrained_model_name_or_path (str): the pretrained model used to generate the scores.
+    batch_size (int): the batch size for predictions.
+    max_length (int): maximum sequence length.
+Returns:
+    scores (list of int): list of scores.
+Examples:
+    >>> my_new_metric = datasets.load_metric("frugalscore")
+    >>> results = my_new_metric.compute(predictions=['hello there', 'huggingface'], references=['hello world', 'hugging face'])
```
lhoestq, line 49: you can name the metric variable `frugalscore`
```suggestion
    >>> frugalscore = datasets.load_metric("frugalscore")
    >>> results = frugalscore.compute(predictions=['hello there', 'huggingface'], references=['hello world', 'hugging face'])
```

metrics/frugalscore/frugalscore.py
```diff
@@ -0,0 +1,96 @@
...
+        def tokenize_function(data):
+            return tokenizer(
+                data["sentence1"], data["sentence2"], max_length=max_length, truncation=True, padding=True
+            )
+
+        tokenized_datasets = raw_datasets.map(tokenize_function, batched=True)
+        tokenized_datasets.remove_columns(["sentence1", "sentence2"])
+        trainer = Trainer(model, training_args, tokenizer=tokenizer)
+        predictions = trainer.predict(tokenized_datasets)
+        return {"scores": list(predictions.predictions.squeeze(-1))}
```
lhoestq, line 96: Here would it make sense to also return the mean of the scores ?

metrics/frugalscore/frugalscore.py
```diff
@@ -0,0 +1,96 @@
...
+            homepage="https://github.com/moussaKam/FrugalScore",
+        )
+
+    def _compute(
+        self,
+        predictions,
+        references,
+        pretrained_model_name_or_path="moussaKam/frugalscore_tiny_bert-base_bert-score",
+        batch_size=32,
+        max_length=128,
```
lhoestq, line 77: I think it would be awesome if we had a `device` parameter here, to run it on GPU ! What do you think ?

lhoestq (commented):
Hi ! Thanks for adding this metric, it looks very interesting :)

I left a few comments below.

I also have one additional question: do you have a dummy/tiny model that we can use in the `datasets` CI to test this metric?

Usually we avoid using big models in the CI because they slow it down significantly. For example for the `bertscore` metric we simply mock the model completely

metrics/frugalscore/frugalscore.py
moussaKam: I think it's OK like that, I tried to follow [`bleurt`](https://github.com/huggingface/datasets/blob/master/metrics/bleurt/bleurt.py#L124)

moussaKam:
@lhoestq 

The model used by default (`moussaKam/frugalscore_tiny_bert-base_bert-score`) is a tiny model.

I still want to make one modification before merging.
I would like to load the model checkpoint once. Do you think it's a good idea if I load it in `_download_and_prepare`? In this case should the model name be the `self.config_name` or another variable say `self.model_name` ?

moussaKam:
OK, I added a commit that load the checkpoint in `_download_and_prepare`. Please let me know if it looks good.

moussaKam:
@lhoestq is everything OK to merge?

metrics/frugalscore/frugalscore.py
```diff
@@ -0,0 +1,112 @@
...
+        self.model = AutoModelForSequenceClassification.from_pretrained(checkpoint)
+        self.tokenizer = AutoTokenizer.from_pretrained(checkpoint)
+
+    def _compute(
+        self,
+        predictions,
+        references,
+        batch_size=32,
+        max_length=128,
+        device="gpu",
```
lhoestq, line 86: Maybe the default should be `cpu` as in deep learning frameworks like pytorch ?

metrics/frugalscore/frugalscore.py
lhoestq: Ok ! Sounds good :)

lhoestq (commented):
Thanks ! It's really nice this way. I have one last comment about the default value for `device`:

metrics/frugalscore/frugalscore.py
moussaKam: I changed it so the default value is `gpu` only if gpu is available

lhoestq: Sounds good thanks ! This is also consistent with other metrics like BERTScore

lhoestq:
I triggered the CI and it's failing, can you merge the `master` branch into yours ? It should fix the issues.

Also the doctest apparently raises an error because it outputs `{'scores': [0.6307542, 0.6449357]}` instead of `{'scores': [0.631, 0.645]}` - feel free to edit the code example in the docstring to round the scores, that should fix it

moussaKam:
@lhoestq hope it's OK now

lhoestq (approved):
Thanks for the changes ! I also did some minor changes as well and it's all good now :)

PR #3680, opened 2022-02-04
Move dataset_infos to local dir with only data files

albertvillanova:
Currently this case is missed.

CC: @lvwerra


lhoestq (approved):
good catch !

PR #3681, opened 2022-02-04
Fix TestCommand to move dataset_infos instead of copying

albertvillanova:
Why do we copy instead of moving the file?

CC: @lhoestq @lvwerra

### DIFF
--- a/src/datasets/commands/test.py
+++ b/src/datasets/commands/test.py
@@ -1,7 +1,7 @@
 import os
 from argparse import ArgumentParser
 from pathlib import Path
-from shutil import copyfile, rmtree
+from shutil import move, rmtree
 from typing import Generator
 
 import datasets.config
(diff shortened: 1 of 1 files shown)


lhoestq:
All the datasets that are loaded normally with `load_dataset`, if `dataset_infos.json` exists, have this file in the importable directory. So it's fine if we copy the file instead of moving it but it's not a big deal.

Any reason to prefer moving it rather than copying it ?

albertvillanova:
@lvwerra reported than when generating the `dataset_infos.json` for multiple dataset directories containing only JSONL files, subsequent json files contained all previous directories as configs:
- First generate JSON for dataset in dir `dir1`: dataset_infos.json contains one config for `dir1`
- Then generate JSON for dataset in dir `dir2`: dataset_infos.json contains 2 configs, for `dir1` and `dir2`

The reason is that all dataset_infos.json files are first created in the same dir (the one containing the json builder) and then **copied** to the user dir.

Subsequent calls of TestCommand don't replace the dataset_infos.json already present in the dir of the json builder, but append to it.

albertvillanova:
@lhoestq aren't you mentioning the case in the else clause?
```python
else:  # in case of a remote dataset
    dataset_dir = None
```

In that case `dataset_infos.json` is not copied: `dataset_dir = None`

lhoestq:
When using the JSON loader, calling `get_imported_module_dir()` returns a path inside the pip installed packages, so we shouldn't write files in it anyway, and the dataset_infos.json file should be written directly in the user's directory instead (some users don't have write access to the pip installed packages for example).

Maybe the packaged modules like `json` should override `_save_infos` to save them in the user's directory instead of next to the builder's script. What do you think ?

lhoestq:
Anyway as a hotfix we can just add an exception for the `json` builder for now, if the issue has to be fixed soon

cApTO23 (approved):
This commit looks.good.

albertvillanova:
I'm closing this PR.

PR #3699, opened 2022-02-10
Add dev-only config to Natural Questions dataset

### LINKED ISSUE #413: Is there a way to download only NQ dev?
Maybe I missed that in the docs, but is there a way to only download the dev set of natural questions (~1 GB)? 
As we want to benchmark QA models on different datasets, I would like to avoid downloading the 41GB of training data. 

I tried
```
dataset = nlp.load_dataset('natural_questions', split="validation", beam_runner="DirectRunner")
```
But this still triggered a big download of presumably the whole dataset. Is there any way of doing this or are splits / slicing options only available after downloading?

Thanks!

albertvillanova:
As suggested by @lhoestq and @thomwolf, a new config has been added to Natural Questions dataset, so that only dev split can be downloaded. 

Fix #413.

### DIFF
--- a/datasets/natural_questions/README.md
+++ b/datasets/natural_questions/README.md
@@ -123,7 +123,7 @@ The data fields are the same among all splits.
 | name    |  train |validation|
 |---------|-------:|---------:|
 | default | 307373 |      7830|
-| dev     |      0 |      7830|
+| dev     |    N/A |      7830|
 
 ## Dataset Creation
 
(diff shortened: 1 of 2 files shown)


albertvillanova:
I've copied the 0.0.2 folder content to 0.0.3, as suggested.

I'm updating the dataset card...

lhoestq (approved):
Cool thanks :)

PR #3702, opened 2022-02-10
Update data URL of lm1b dataset

yazdanbakhsh:
The http address doesn't work anymore


lhoestq:
Hi ! I'm getting some 503 from both the http and https addresses. Do you think we could host this data somewhere else ? (please check if there is a license and if it allows redistribution)

albertvillanova:
Both HTTP and HTTPS links are working now.

We are closing this PR.

PR #3709, opened 2022-02-11
Set base path to hub url for canonical datasets

lhoestq:
This should allow canonical datasets to use relative paths to download data files from the Hub

cc @polinaeterna this will be useful if we have audio datasets that are canonical and for which you'd like to host data files


albertvillanova (approved):
This PR allows one of the missing features of canonical datasets: to host data. Canonical and community datasets are getting more and more similar...

Only one question: I don't know the exact mechanism to sync canonical datasets from GitHub to HF Hub. I'm only wondering if data files won't be deleted during synchronization (as they are not present on GitHub, but only on the HF Hub).

Apart from this, I don't see any other potential breaking...

PR #3715, opened 2022-02-13
Fix bugs in msr_sqa dataset

Timothyxxx:
The last version has many problems,

1) Errors in table load-in. Split by a single comma instead of using pandas is wrong.
2) id reduplicated in _generate_examples function.
3) Missing information of history questions which make it hard to use.

I fix it refer to https://github.com/HKUNLP/UnifiedSKG. And we test it to perform normally.


Timothyxxx:
It shows below when I run test:

FAILED tests/test_dataset_common.py::LocalDatasetTest::test_load_dataset_all_configs_msr_sqa - ValueError: Unknown split "validation". Should be one of ['train', 'test'].

It doesn't make no sense for me.

mariosasko:
@albertvillanova Does this PR has some additional fixes compared to https://github.com/huggingface/datasets/pull/3771 or we can close it?

albertvillanova:
@mariosasko besides the fix of the DuplicatedKeysError, this PR:
- changes the reading of one of the files: use pandas instead of splitting by comma
- adds some extra logic in the processing of the data: adding a new field "question_and_history"

We should decide whether validating these additional changes.
- for example, if we accept as pertinent the addition of the field "question_and_history", this should be added as feature to the info, and the matadata should be regenerated...

Timothyxxx:
Hi guys, anything we can do to fix that bug👀? @mariosasko @albertvillanova @lhoestq

albertvillanova (approved):
@Timothyxxx, thanks for your contribution fixing the bugs. And sorry for the late response.

PR #3722, opened 2022-02-15
added electricity load diagram dataset

kashif:
Initial Electricity Load Diagram time series dataset.


lhoestq (approved):
Awesome ! Good job with the dataset script and the dataset card. I especially like the example in the dataset card that says explicitly how users can change the configuration. This is going to be the very first time series dataset available !

Let us know if it's all good on your side, so we can merge this PR :)

Also are there some things you'd like to see improved in `datasets` in the future for time series data ?

PR #3723, opened 2022-02-15
Fix flatten of complex feature types

### LINKED ISSUE #3686: `Translation` features cannot be `flatten`ed
## Describe the bug

(`Dataset.flatten`)[https://github.com/huggingface/datasets/blob/master/src/datasets/arrow_dataset.py#L1265] fails for columns with feature (`Translation`)[https://github.com/huggingface/datasets/blob/3edbeb0ec6519b79f1119adc251a1a6b379a2c12/src/datasets/features/translation.py#L8]

## Steps to reproduce the bug

```python

dataset = load_dataset("europa_ecdc_tm", "en2fr", split="train[:10]")
print(dataset.features)
# {'translation': Translation(languages=['en', 'fr'], id=None)}
print(dataset[0])
# {'translation': {'en': 'Vaccination against hepatitis C is not yet available.', 'fr': 'Aucune vaccination contre l’hépatite C n’est encore disponible.'}}

dataset.flatten()
```

## Expected results

`dataset.flatten` should flatten the `Translation` column as if it were a dict of `Value("string")`

```python
dataset[0]
# {'translation.en':  'Vaccination against hepatitis C is not yet available.', 'translation.fr': 'Aucune vaccination contre l’hépatite C n’est encore disponible.' }
dataset.features
# {'translation.en': Value("string"), 'translation.fr': Value("string")}
```

## Actual results

```python
In [31]: dset.flatten()
KeyError                                  Traceback (most recent call last)
<ipython-input-31-bb88eb5276ee> in <module>
----> 1 dset.flatten()

[...]\site-packages\datasets\fingerprint.py in wrapper(*args, **kwargs)
    411             # Call actual function
    412
--> 413             out = func(self, *args, **kwargs)
    414
    415             # Update fingerprint of in-place transforms + update in-place history of transforms

[...]\site-packages\datasets\arrow_dataset.py in flatten(self, new_fingerprint, max_depth)
   1294                 break
   1295         dataset.info.features = self.features.flatten(max_depth=max_depth)
-> 1296         dataset._data = update_metadata_with_features(dataset._data, dataset.features)
   1297         logger.info(f'Flattened dataset from depth {depth} to depth {1 if depth + 1 < max_depth else "unknown"}.')
   1298         dataset._fingerprint = new_fingerprint

[...]\site-packages\datasets\arrow_dataset.py in update_metadata_with_features(table, features)
    534 def update_metadata_with_features(table: Table, features: Features):
    535     """To be used in dataset transforms that modify the features of the dataset, in order to update the features stored in the metadata of its schema."""
--> 536     features = Features({col_name: features[col_name] for col_name in table.column_names})
    537     if table.schema.metadata is None or b"huggingface" not in table.schema.metadata:
    538         pa_metadata = ArrowWriter._build_metadata(DatasetInfo(features=features))

[...]\site-packages\datasets\arrow_dataset.py in <dictcomp>(.0)
    534 def update_metadata_with_features(table: Table, features: Features):
    535     """To be used in dataset transforms that modify the features of the dataset, in order to update the features stored in the metadata of its schema."""
--> 536     features = Features({col_name: features[col_name] for col_name in table.column_names})
    537     if table.schema.metadata is None or b"huggingface" not in table.schema.metadata:
    538         pa_metadata = ArrowWriter._build_metadata(DatasetInfo(features=features))

KeyError: 'translation.en'
```

## Environment info

- `datasets` version: 1.18.3
- Platform: Windows-10-10.0.19041-SP0
- Python version: 3.7.10
- PyArrow version: 3.0.0

mariosasko:
Fix `flatten` for the following feature types: Image/Audio, Translation, and  TranslationVariableLanguages.

Inspired by `cast`/`table_cast`, I've introduced a `table_flatten` function to handle the Image/Audio types.

CC: @SBrandeis  

Fix #3686.


src/datasets/features/features.py
```diff
@@ -1208,6 +1208,20 @@ def __setitem__(self, column_name: str, feature: FeatureType):
...
+    def update(self, iterable, **kwds):
+        if hasattr(iterable, "keys"):
+            for key in iterable.keys():
+                self[key] = iterable[key]
+        else:
+            for key, value in iterable:
+                self[key] = value
+        for key in kwds:
+            self[key] = kwds[key]
+
```
mariosasko, line 1224: @albertvillanova @lhoestq This is a temporary fix for `Features.flatten`. After we merge this one, I'll open a new PR to fix the issues with `dict` subclassing.

src/datasets/table.py
```diff
@@ -1125,6 +1127,52 @@ def table_cast(table: pa.Table, schema: pa.Schema):
...
+    from .features import Audio, Features
+
+    has_decodable_feature = False
+    has_audio_feature = False
+    features = Features.from_arrow_schema(table.schema)
+    for column_name, subfeature in features.items():
+        if hasattr(subfeature, "decode_example") and subfeature.decode:
+            has_decodable_feature = True
+            if isinstance(subfeature, Audio):
+                has_audio_feature = True
```
lhoestq, line 1151: If we have a `flatten` method per type, then we could just check if `subfeature.flatten() == subfeature`. This way we don't have hardcoded lists of feature tyeps that have special handling

src/datasets/arrow_dataset.py
```diff
@@ -616,6 +616,27 @@ def _align_features(features_list: List[Features]) -> List[Features]:
     return [Features({k: name2feature[k] for k in features.keys()}) for features in features_list]
 
 
+def _check_if_trying_to_flatten_audio_feature(feature, column_name, depth, max_depth, inplace: bool = False):
+    """Check if the user is trying to flatten an audio feature."""
+    if isinstance(feature, Audio) and feature.decode:
```
lhoestq, line 621: Why audio and not image as well here ?

lhoestq (commented):
Cool thanks ! This will help a lot :)

My only concern is that more and more feature types are hardcoded in these functions and it makes them harder to maintain. Maybe each feature type could be responsible of defining their own logic, see my comments:

src/datasets/arrow_dataset.py
mariosasko: A bit hard to explain. We can't dispatch single storage to multiple columns that are introduced by flatten ("path", "array", "sampling_rate" for Audio). And for `Image` we can, because `flatten` is a no-op in that case (singe storage -> single column).

src/datasets/table.py
```diff
@@ -1125,6 +1127,44 @@ def table_cast(table: pa.Table, schema: pa.Schema):
         return table
 
 
+def table_flatten(table: pa.Table):
+    """Improved version of pa.Table.flatten.
+
+    It behaves as pa.Table.flatten, but skips decodable features
+    unless the `decode` attribute of these features is set to False.
```
lhoestq, line 1134: Maybe mention that it does a 1-step flatten to makes things clear (vs Dataset.flatten that flattens everything)

tests/test_arrow_dataset.py
```diff
@@ -776,6 +787,83 @@ def test_flatten(self, in_memory):
...
+                    self.assertNotEqual(dset._fingerprint, fingerprint)
+                    assert_arrow_metadata_are_synced_with_dataset_features(dset)
+
+    @require_pil
+    def test_flatten_complex_image(self, in_memory):
+        # decoding turned on
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            with Dataset.from_dict(
+                {"a": [np.arange(4 * 4 * 3).reshape(4, 4, 3)] * 10, "foo": [1] * 10},
+                features=Features({"a": Image(), "foo": Value("int64")}),
```
lhoestq, line 838: Could this example also have nested features ? This way it completely test the `table_flatten` function

lhoestq (commented):
Cool thanks :)
I just spotted one bug, the rest looks all good

lhoestq:
Apparently the merge brought back some tests that use `flatten_()` that we removed recently

lhoestq (approved):
Thank you !

lhoestq (approved):
Thanks for the tests :)

PR #3734, opened 2022-02-16
Fix bugs in NewsQA dataset

### LINKED ISSUE #3733: Bugs in NewsQA dataset
## Describe the bug
NewsQA dataset has the following bugs:
- the field `validated_answers` is an exact copy of the field `answers` but with the addition of `'count': [0]` to each dict
- the field `badQuestion` does not appear in `answers` nor `validated_answers`

## Steps to reproduce the bug
By inspecting the dataset script we can see that:
- the parsing of `validated_answers` is a copy-paste of the one for `answers`
- the `badQuestion` field is ignored in the parsing of both `answers` and `validated_answers`

albertvillanova:
Fix #3733.


lhoestq (approved):
Cool thanks ! You might need to also update the sample in the dataset card to reflect this change, as well as the documentation of the fields to account for the new ones and the removed ones

PR #3736, opened 2022-02-16
Local paths in common voice

lhoestq:
Continuation of https://github.com/huggingface/datasets/pull/3664:

- pass the `streaming` parameter to _split_generator
- update @anton-l's code to use this parameter for `common_voice`
- add a comment to explain why we use `download_and_extract` in non-streaming and `iter_archive` in streaming

Now the `common_voice` dataset has a local path back in `ds["path"]`, and this field is `None` in streaming mode.

cc @patrickvonplaten @anton-l @albertvillanova


datasets/common_voice/common_voice.py
```diff
@@ -655,65 +657,126 @@ def _info(self):
...
                     "path_to_clips": path_to_clips,
                 },
             ),
         ]
 
-    def _generate_examples(self, files, filepath, path_to_clips):
+    def _generate_examples(self, archive_iterator, filepath, path_to_clips):
         """Yields examples."""
+        if archive_iterator is not None:
+            yield from self._generate_examples_streaming(archive_iterator, filepath, path_to_clips)
```
patrickvonplaten, line 744: small nit - I'd even pass the `streaming` flag here to make it super clear that they are two different modes and maybe have both a `_generate_examples_streaming(...)` and a `_generate_examples_non_streaming(...)`

patrickvonplaten (commented):
That looks great to me! Think it's quite easy to understand that there are two parallel ways now on how to download and prepare audio datasets

src/datasets/builder.py
```diff
@@ -727,8 +729,12 @@ def _save_infos(self):
 
     def _make_split_generators_kwargs(self, prepare_split_kwargs):
         """Get kwargs for `self._split_generators()` from `prepare_split_kwargs`."""
-        del prepare_split_kwargs
-        return {}
+        split_generators_kwargs = {}
+        split_generators_arg_names = inspect.signature(self._split_generators).parameters.keys()
+        if "streaming" in split_generators_arg_names:
+            streaming = isinstance(prepare_split_kwargs.get("dl_manager"), StreamingDownloadManager)
```
albertvillanova, line 735: I guess you need the `DownloadManager` instance to find out whether we are in streaming mode or not... Because the builder itself knows nothing about streaming or not...

albertvillanova (approved):
I think the importance of this PR is fixing Common Voice, once we realize that the approach of non-extracting the archive content in non-streaming mode is not optimal/suitable for users who want to have direct access to audio file paths.

So OK for approving it and merge/fix Common Voice the faster the better.

But on the other hand, IMHO, I think this specific solution adds complexity to handling streaming/non-streaming, and moves this complexity to the loading script and thus to the contributors/users who want to create the loading script for their canonical/community datasets (instead of keeping it hidden form the end users). 

Maybe we could have a discussion in the near future to see if it is possible to find other solutions (or not), while keeping the requirement of having access to the audio file paths in non-streaming. It is just a suggestion! :)

src/datasets/builder.py
lhoestq: Indeed having this logic inside the `builder` can be a bit unexpected. An alternative would be to replace the `streaming` parameter by
```python
streaming = dl_manager.is_streaming()
```
inside the dataset script

lhoestq:
I just changed to `dl_manager.is_streaming` rather than an additional parameter `streaming` that has to be handled by the DatasetBuilder class - this way the streaming logic doesn't interfere with the base builder's code.

I think it's better this way, but let me know if you preferred the previous way and I can revert

albertvillanova:
@lhoestq good idea: much cleaner this way! That way each class has its own responsibilities without mixing around...

PR #3743, opened 2022-02-17
initial monash time series forecasting repository

kashif:


datasets/monash_tsf/README.md
```diff
@@ -0,0 +1,179 @@
...
+prediction_length_map = {
+    "S": 60,
+    "T": 60,
+    "H": 48,
+    "D": 30,
+    "W": 8,
+    "M": 12,
+    "Y": 4,
+}
+```
```
mariosasko, line 108: Let's mention the default number of examples in each dataset (as a table, see [this](https://github.com/huggingface/datasets/tree/master/datasets/imdb#data-splits) for instance), and, similarly to the electricity load diagrams [readme](https://github.com/huggingface/datasets/tree/master/datasets/electricity_load_diagrams#dataset-usage), we should add a `Dataset Usage` section to explain the configuration parameters (`freq`, `prediction_length`, `rolling_evaluations`) that the user can customize.

datasets/monash_tsf/monash_tsf.py
```diff
@@ -0,0 +1,606 @@
...
+    record: Optional[str] = None
+    freq: Optional[str] = None
+    prediction_length: Optional[int] = None
+    item_id_column: Optional[str] = None
+    data_column: Optional[str] = None
+    target_fields: Optional[List[str]] = None
+    feat_dynamic_real_fields: Optional[List[str]] = None
+    multivariate: bool = False
+    rolling_evaluations: int = 1
+    ROOT: str = "https://zenodo.org/record"
```
mariosasko, line 65: A nit: it's cleaner to make `ROOT` a global variable.

datasets/monash_tsf/utils.py
```diff
@@ -0,0 +1,188 @@
...
+            # Strip white space from start/end of line
+            line = line.strip()
+
+            if line:
+                if line.startswith("@"):  # Read meta-data
+                    if not line.startswith("@data"):
+                        line_content = line.split(" ")
+                        if line.startswith("@attribute"):
+                            if len(line_content) != 3:  # Attributes have both name and type
+                                raise Exception("Invalid meta-data specification.")
```
mariosasko, line 44: Let's replace `Exception`s with `ValueError`s in this file

mariosasko (commented):
Nice. Some final comments:

Btw, my comment to add a Data Usage section similar to the one in the `electricity_load_diagrams` dataset is still not addressed, so I'm unresolving it.

datasets/monash_tsf/monash_tsf.py
```diff
@@ -0,0 +1,606 @@
...
+                "W": 8,
+                "M": 12,
+                "Y": 4,
+            }
+            freq = frequency_converter(frequency)
+            freq = to_offset(freq).name
+            forecast_horizon = prediction_length_map[freq]
+
+        if self.config.prediction_length is not None:
+            forecast_horizon = self.config.prediction_length
```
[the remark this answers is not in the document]
kashif: yes, my bad we do not need or use the `freq` in the builder config as it's fixed for this repository of dataset... arb. freq might not work as some time series are too small etc. so I will remove it.

datasets/monash_tsf/README.md
```diff
@@ -0,0 +1,234 @@
...
+| temperature_rain              | Nature    | 32072         | 1D     | 30         | [Godahewa et al. 2021](https://openreview.net/pdf?id=wEc1mgAjU- )      
+
+
+### Dataset Usage
+
+To load a particular dataset just specify its name from the table above e.g.:
+
+```python
+load_dataset("monash_tsf", "nn5_daily")
+```
```
mariosasko, line 107: The user can pass a custom `prediction_length` or `rolling_evaluations`, so we should list and describe these args here, no?

kashif: no not in this dataset as those are fixed as well (from the monash side and their preprocessing)

mariosasko (approved):
Thanks! LGTM.

datasets/monash_tsf/monash_tsf.py
```diff
@@ -0,0 +1,566 @@
...
+
+    file_name: Optional[str] = None
+    record: Optional[str] = None
+    prediction_length: Optional[int] = None
+    item_id_column: Optional[str] = None
+    data_column: Optional[str] = None
+    target_fields: Optional[List[str]] = None
+    feat_dynamic_real_fields: Optional[List[str]] = None
+    multivariate: bool = False
+    rolling_evaluations: int = 1
```
lhoestq, line 64: Awesome ! This is nice to see that it has the same `prediction_length` and `rolling_evaluations` as in `electricity_load_diagrams`

datasets/monash_tsf/README.md
```diff
@@ -0,0 +1,234 @@
...
+> Notes:
+> - Data might contain missing values as in the original datasets.
+> - The prediction length is either specified in the dataset or a default value depending on the frequency is used as in the original repository benchmark.
+
```
lhoestq, line 111: Maybe add an example if the users want to chanhe the `rolling_evaluations` or `prediction_length` ?

lhoestq (commented):
Nice thank you ! I added a few comments :)

If I understand correctly this dataset only requires ~2-3GB of RAM to be processed (the pandas dataframes are loaded in memory) - which is fine.

Btw there seems to be many dummy_data.zip file. Since they're all pretty similar, feel free to only keep one (e.g. the first one "weather"), and remove all the other ones. This way it helps keeping the size of the git repository reasonable ^^ I know you must have spent quite some time making them, sorry about that

datasets/monash_tsf/README.md
lhoestq: Feel free to mention it in the dataset card :)

datasets/monash_tsf/monash_tsf.py
```diff
@@ -0,0 +1,606 @@
...
+    file_name: Optional[str] = None
+    record: Optional[str] = None
+    freq: Optional[str] = None
+    prediction_length: Optional[int] = None
+    item_id_column: Optional[str] = None
+    data_column: Optional[str] = None
+    target_fields: Optional[List[str]] = None
+    feat_dynamic_real_fields: Optional[List[str]] = None
+    multivariate: bool = False
+    rolling_evaluations: int = 1
```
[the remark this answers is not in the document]
lhoestq: For consistency with `Electricity Load Diagrams`, should we have `freq` and drop `forecast_horizon` ? (not a big deal though)

lhoestq (approved):
thanks :)

lhoestq:
The CI fails are unrelated to this PR, merging !

kashif:
thanks 🙇🏽

PR #3757, opened 2022-02-18
Add perplexity to metrics

emibaylor:
Adding perplexity metric

This code differs from the code in [this](https://huggingface.co/docs/transformers/perplexity) HF blog post because the blogpost code fails in at least the following circumstances:
- returns nans whenever the stride = 1
- hits a runtime error when the stride is significantly larger than the max model length (e.g. if max_model_length = 512 and stride = 1024)

Note that:
- As it is, it only works for causal models. Pseudoperplexity can be added later as another metric to work with masked language models.
- It takes in a list of strings so that it can be dataset independent. This does mean that it doesn't currently batch inputs, and is therefore relatively slow.
- It overwrites the metrics compute() function for a specific perplexity compute() function. This is because the current general metrics compute() function requires model-generated predictions, which doesn't make sense in the context of perplexity


lhoestq:
Awesome thank you ! The implementation of the parent `Metric` class was assuming that all metrics were supposed to have references/predictions pairs - I just changed that so you don't have to override `compute()`. I took the liberty of doing the changes directly inside this PR to make sure it works as expected with perplexity.

Other than that it looks all good :) I just did minor changes like remove a remaining `print` as well as fixing the `Features` defined in `_info()`. I also renamed `input_text` to `input_texts` since it makes it more obvious that it's a list of strings - let me know if it sounds good to you.

Let me know if you'd like to make other changes or if it's all good for you !

metrics/perplexity/perplexity.py
```diff
@@ -0,0 +1,146 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+""" Perplexity Metric """
+
+import torch
+from tqdm import tqdm
```
lhoestq, line 18: Could you use `from datasets.utils import tqdm` instead ? This way the progress bar will be disabled if users do
```python
datasets.disable_progress_bar()
```

metrics/perplexity/perplexity.py
```diff
@@ -0,0 +1,146 @@
...
+            ),
+            reference_urls=["https://huggingface.co/docs/transformers/perplexity"],
+        )
+
+    def _compute(self, input_texts, model_id, stride=512, device=None):
+
+        if device is not None:
+            assert device in ["gpu", "cpu"], "device should be either gpu or cpu."
+        else:
+            device = "gpu" if torch.cuda.is_available() else "cpu"
```
lhoestq, line 95: I added this change for consistency with other model-based metrics

metrics/perplexity/perplexity.py
```diff
@@ -0,0 +1,146 @@
...
+
+    Example 2:
+        >>> perplexity = datasets.load_metric("perplexity")
+        >>> input_texts = datasets.load_dataset("wikitext", "wikitext-2-raw-v1", "test") # doctest:+ELLIPSIS
+        [...]
+        >>> results = perplexity.compute(model_id='gpt2',
+        ...                              input_texts=input_texts,
+        ...                              stride=256)
+        >>> round(results["perplexity"], 2)
+        285.12
```
lhoestq, line 70: In your previous commit, the perplexity was much higher (around 6,000) - can you double check that 285,12 is the correct value ?

lhoestq (commented):
Some additional suggestions:

metrics/perplexity/perplexity.py
```diff
@@ -0,0 +1,146 @@
...
+                num_predictions += 1
+
+            if len(nlls) > 0:
+                ppls.append(torch.exp(torch.stack(nlls).sum() / num_predictions))
+
+        ppl = torch.stack(ppls).sum() / len(encoded_texts)
+
+        return {
+            "perplexity": float(ppl),
+        }
```
albertvillanova, line 146: I'm wondering why returning a dict instead of a float. If finally a float is returned, docstring should be changed accordingly.

CC: @lhoestq

lhoestq: For consistency with the other metrics. Certain metrics outputs several scores (e.g. rouge), so we output a dictionary. Even `accuracy` returns a dictionary.

Note that this can change if we build a new lib for metrics. Especially since Keras metrics expect one single number.
cc @lvwerra

metrics/perplexity/perplexity.py
emibaylor: Yes! Sorry about that, looked into it and noticed a few mistakes in my code, they should be fixed now! Will update the examples as soon as I can

lhoestq: Ok thanks ! The rest looks all good, please ping me when the examples are updated :)

emibaylor: Awesome! Just updated the examples so all should be goon now

metrics/perplexity/perplexity.py
```diff
@@ -0,0 +1,149 @@
...
+
+    Example 2:
+        >>> perplexity = datasets.load_metric("perplexity")
+        >>> input_texts = datasets.load_dataset("wikitext", "wikitext-2-raw-v1", "test") # doctest:+ELLIPSIS
+        [...]
+        >>> results = perplexity.compute(model_id='gpt2',
+        ...                              input_texts=input_texts,
+        ...                              stride=256)
+        >>> round(results["perplexity"], 2)
+        304.81
```
lhoestq, line 70:
```suggestion
        107.75
```

lhoestq (commented):
I'm still not getting the same values as you when running the examples for some reason

Also feel free to run `make style` to fix the code formatting :)

metrics/perplexity/perplexity.py
```diff
@@ -0,0 +1,149 @@
...
+        in the input list.
+Examples:
+    Example 1:
+        >>> perplexity = datasets.load_metric("perplexity")
+        >>> input_texts = ["lorem ipsum", "Happy Birthday!", "Bienvenue"]
+        >>> results = perplexity.compute(model_id='gpt2',
+        ...                              input_texts=input_texts,
+        ...                              stride=1)
+        >>> round(results["perplexity"], 2)
+        837.04
```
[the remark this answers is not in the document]
lhoestq: ^ 78.22 is also the number that the CI outputs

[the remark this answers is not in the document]
emibaylor: Sorry about that! I accidentally took the result for this when stride=2, not stride=1. Fixed that and now get 78.22 as the result!

metrics/perplexity/perplexity.py
emibaylor: Fixed the code so that when the data is loaded, it's loaded as:
`input_texts = datasets.load_dataset("wikitext", "wikitext-2-raw-v1", split="test")["text"]`

This should change it so that the output is 304.81 on your end, too (but if you could confirm that, that would be great!)

lhoestq: Cool thanks !

lhoestq:
The test with the full test set seems to take too much time in the CI - maybe we can just select `split="test[:few_examples]"`

lhoestq (approved):
Alright this is all good now ! Thanks a lot :)

Feel free to merge if it's good for you

PR #3767, opened 2022-02-21
Expose method and fix param

severo:
A fix + expose a new method, following https://github.com/huggingface/datasets/pull/3670


lhoestq (approved):
Good catch ! And thanks for adding some tests :)

PR #3775, opened 2022-02-22
Update gigaword card and info

mariosasko:
Reported on the forum: https://discuss.huggingface.co/t/error-loading-dataset/14999


lhoestq:
I think it actually comes from an issue here:

https://github.com/huggingface/datasets/blob/810b12f763f5cf02f2e43565b8890d278b7398cd/src/datasets/utils/file_utils.py#L575-L579

This code doesn't seem to work anymore. This can probably be fixed with

```python
if url.startswith("https://drive.google.com/"): 
    url += "&confirm=yes"
    cookies = response.cookies 
```

because Google Drive doesn't return the `download_warning` cookie anymore.

lhoestq:
Actually it seems that is has been fixed already in https://github.com/huggingface/datasets/pull/3787 :)

I think it should have fixed the gigaword dataset loading

mariosasko:
@lhoestq The linked PR indeed fixes the issue. This PR is still worth merging IMO to update `gigaword`'s card.

lhoestq (approved):
Definitely :) thanks !

PR #3690, opened 2022-02-08
Update docs to new frontend/UI

mishig25:
### TLDR: Update `datasets` `docs` to the new syntax (markdown and mdx files) & frontend (as how it looks on [hf.co/transformers](https://huggingface.co/docs/transformers/index))

| Light mode                                                                                 | Dark mode                                                             |
|-----------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------|
| Screenshot 2022-02-17 at 14 15 34    | Screenshot 2022-02-17 at 14 16 27                                     |

## Checklist

- [x] update datasets docs to new syntax (should call `doc-builder convert`) (this PR)
- [x] discuss `@property` methods frontend https://github.com/huggingface/doc-builder/pull/87
- [x] discuss `inject_arrow_table_documentation` (this PR) https://github.com/huggingface/datasets/pull/3690#discussion_r801847860
- [x] update datasets docs path on moon-landing https://github.com/huggingface/moon-landing/pull/2089
- [x] convert pyarrow docstring from Numpydoc style to groups style  https://github.com/huggingface/doc-builder/pull/89(https://stackoverflow.com/a/24385103/6558628)
- [x] handle `Raises` section on frontend and doc-builder https://github.com/huggingface/doc-builder/pull/86
- [x] check imgs path (this PR) (nothing to update here)
- [x] doc exaples block has to follow format `Examples::` https://github.com/huggingface/datasets/pull/3693
- [x] fix [this docstring](https://github.com/huggingface/datasets/blob/6ed6ac9448311930557810383d2cfd4fe6aae269/src/datasets/arrow_dataset.py#L3339) (causing svelte compilation error)
- [x] Delete sphinx related files 
- [x] Delete sphinx CI
- [x]  Update docs config in setup.py
- [x] add `versions.yml` in doc-build  https://github.com/huggingface/doc-build/pull/1
- [x] add `versions.yml` in doc-build-dev https://github.com/huggingface/doc-build-dev/pull/1
- [x] https://github.com/huggingface/moon-landing/pull/2089
- [x] format docstrings for example `datasets.DatasetBuilder.download_and_prepare` args format look wrong
- [x] create new github actions. (can probably be in a separate PR) (see the transformers equivalents below)
1. [build_dev_documentation.yml](https://github.com/huggingface/transformers/blob/master/.github/workflows/build_dev_documentation.yml)
2. [build_documentation.yml](https://github.com/huggingface/transformers/blob/master/.github/workflows/build_documentation.yml)
[...]

PR #3777, opened 2022-02-22
Start removing canonical datasets logic

lhoestq:
I updated the source code and the documentation to start removing the "canonical datasets" logic.

Indeed this makes the documentation confusing and we don't want this distinction anymore in the future. Ideally users should share their datasets on the Hub directly.

### Changes

- the documentation about dataset loading mentions the datasets on the Hub (no difference between canonical and community, since they all have their own repository now)
- the documentation about adding a dataset doesn't explain the technical differences between canonical and community anymore, and only presents how to add a community dataset. There is still a small section at the bottom that mentions the datasets that are still on GitHub and redirects to the `ADD_NEW_DATASET.md` guide on GitHub about how to contribute a dataset to the `datasets` library
- the code source doesn't mention "canonical" anymore anywhere. There is still a `GitHubDatasetModuleFactory` class that is left, but I updated the docstring to say that it will be eventually removed in favor of the `HubDatasetModuleFactory` classes that already exist

Would love to have your feedbacks on this !

cc @julien-c @thomwolf @SBrandeis

### DIFF
--- a/docs/source/share.rst
+++ b/docs/source/share.rst
@@ -130,14 +130,16 @@ Congratulations, your dataset has now been uploaded to the Hugging Face Hub wher
 Datasets on GitHub (legacy)
 ---------------------------
 
-Most datasets have been added to the GitHub repository of the huggingface/datasets repository before being moved to the Hugging Face Hub.
-Editing them still has to be done using Pull Requests on GitHub for now.
+Datasets used to be hosted on our GitHub repository, but all datasets have now been migrated to the Hugging Face Hub.
+The legacy GitHub datasets were added originally on our GitHub repository and therefore don't have a namespace: "squad", "glue", etc. unlike the other datasets that are named "username/dataset_name" or "org/dataset_name".
+Those datasets are still maintained, and if you'd like to edit them, please open a Pull Request on the huggingface/datasets repository.
+Sharing your dataset to the Hub is the recommended way of adding a dataset.
 
 .. important::
 
     The distinction between a Hub dataset and a dataset from GitHub only comes from the legacy sharing workflow. It does not involve any ranking, decisioning, or opinion regarding the contents of the dataset itself.
 
 
-The code of these datasets are reviewed by the Hugging Face team and is considered **safe**, and they require test data in order to be regularly tested.
+The code of these datasets are reviewed by the Hugging Face team, and they require test data in order to be regularly tested.
 
 For more info, please take a look at the documentation on [How to add a new dataset in the huggingface/datasets repository](https://github.com/huggingface/datasets/blob/master/ADD_NEW_DATASET.md).
--- a/docs/source/share.rst
+++ b/docs/source/share.rst
@@ -1,57 +1,26 @@
 Share
 ======
 
-At Hugging Face, we are on a mission to democratize NLP and we believe in the value of open source. That's why we designed 🤗 Datasets so that anyone can share a dataset with the greater NLP community. There are currently over 900 datasets in over 100 languages in the Hugging Face Hub, and the Hugging Face team always welcomes new contributions!
+At Hugging Face, we are on a mission to democratize good Machine Learning and we believe in the value of open source. That's why we designed 🤗 Datasets so that anyone can share a dataset with the greater ML community. There are currently thousands datasets in over 100 languages in the Hugging Face Hub, and the Hugging Face team always welcomes new contributions!
 
-This guide will show you how to share a dataset that can be easily accessed by anyone.
+Dataset repositories offer you nice features such as:
 
-There are two options to share a new dataset:
-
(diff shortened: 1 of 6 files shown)


stevhliu (approved):
Awesome, this looks great!

severo:
I'm not sure if the documentation explains why the dataset identifiers might have a namespace or not (the user/org): 'glue' vs 'severo/glue'. Do you think we should explain it, and relate it to the GitHub/Hub distinction?

docs/source/share.rst
```diff
@@ -157,67 +126,18 @@ Congratulations, your dataset has now been uploaded to the Hugging Face Hub wher
...
-
-Clone the repository
-^^^^^^^^^^^^^^^^^^^^
 
-To share a canonical dataset:
+Datasets on GitHub (legacy)
+---------------------------
 
-1. Fork the 🤗 `Datasets repository <https://github.com/huggingface/datasets>`_ by clicking on the **Fork** button.
+Most datasets have been added to the GitHub repository of the huggingface/datasets repository before being moved to the Hugging Face Hub.
```
albertvillanova, line 133: Nit:
```suggestion
Most datasets have been added to the GitHub huggingface/datasets repository before being moved to the Hugging Face Hub.
```

src/datasets/load.py
```diff
@@ -484,8 +484,12 @@ def get_module(self) -> MetricModule:
         raise NotImplementedError
 
 
-class CanonicalDatasetModuleFactory(_DatasetModuleFactory):
-    """Get the module of a canonical dataset. The dataset script is downloaded from GitHub."""
+class GithubDatasetModuleFactory(_DatasetModuleFactory):
+    """
+    Get the module of a dataset from GitHub (legacy).
+    The dataset script is downloaded from GitHub.
+    This class will eventually be removed and a HubDatasetModuleFactory will be used instead.
```
albertvillanova, line 491: I would add a `.. deprecated::` rst directive here: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated 

With the new mdx docs, I don't know if this is still possible...

albertvillanova (approved):
Thanks for this PR!!!

src/datasets/load.py
lhoestq: I don't think we already have something dedicated to deprecation in the mdx docs. Anyway this class is not in the exported documentation so we can probably keep this this way

docs/source/share.rst
```diff
@@ -157,67 +126,18 @@ Congratulations, your dataset has now been uploaded to the Hugging Face Hub wher
...
------------------------
-
-Canonical datasets are dataset scripts hosted in the GitHub repository of the 🤗 Dataset library.
-The code of these datasets are reviewed by the Hugging Face team, and they require test data in order to be regularly tested.
-
-Clone the repository
-^^^^^^^^^^^^^^^^^^^^
 
-To share a canonical dataset:
+Datasets on GitHub (legacy)
```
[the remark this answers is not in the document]
lhoestq: I improved this part, let me know if it sounds better for you

docs/source/share.rst
```diff
@@ -157,67 +126,18 @@ Congratulations, your dataset has now been uploaded to the Hugging Face Hub wher
...
 
-8. Add your changes, and make a commit to record your changes locally. Then you can push the changes to your account:
+    The distinction between a Hub dataset and a dataset from GitHub only comes from the legacy sharing workflow. It does not involve any ranking, decisioning, or opinion regarding the contents of the dataset itself.
 
-.. code-block::
 
-   git add datasets/<my-new-dataset>
-   git commit
-   git push -u origin my-new-dataset
+The code of these datasets are reviewed by the Hugging Face team and is considered **safe**, and they require test data in order to be regularly tested.
```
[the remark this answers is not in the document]
lhoestq: added a security section https://github.com/huggingface/datasets/pull/3777/commits/d6555d35f531eea9f753ddb8065a55234ec8f612

lhoestq:
> I'm not sure if the documentation explains why the dataset identifiers might have a namespace or not (the user/org): 'glue' vs 'severo/glue'. Do you think we should explain it, and relate it to the GitHub/Hub distinction?

I added an explanation, let me know if it sounds good to you:

```
Datasets used to be hosted on our GitHub repository, but all datasets have now been migrated to the Hugging Face Hub.
The legacy GitHub datasets were added originally on our GitHub repository and therefore don't have a namespace: "squad", "glue", etc. unlike the other datasets that are named "username/dataset_name" or "org/dataset_name".
```

lhoestq:
Thanks for the feedbacks ! Merging this now - if you have some comments I can take care of them in a subsequent PR

I'll also take care of resolving the conflicts with https://github.com/huggingface/datasets/pull/3690

PR #3782, opened 2022-02-24
Error of writing with different schema, due to nonpreservation of nullability

richarddwang:
## 1. Case
```
dataset.map(
  batched=True,
  disable_nullable=True,
)
```
will get the following error at here https://github.com/huggingface/datasets/blob/c9967f55626931f8059dc416526c791444cdfdf7/src/datasets/arrow_writer.py#L516

`pyarrow.lib.ArrowInvalid: Tried to write record batch with different schema`

### 2.1 tracing
During `_map_single`, the following are called
https://github.com/huggingface/datasets/blob/c9967f55626931f8059dc416526c791444cdfdf7/src/datasets/arrow_dataset.py#L2523
https://github.com/huggingface/datasets/blob/c9967f55626931f8059dc416526c791444cdfdf7/src/datasets/arrow_writer.py#L511
### 2.2. Observation
The problem is, even after `table_cast`, `pa_table.schema != self._schema`
`pa_table.schema` (before/after `table_cast`)
```
input_ids: list<item: int32>
  child 0, item: int32
```
`self._schema`
```
input_ids: list<item: int32> not null
  child 0, item: int32
```
### 2.3. Reason
https://github.com/huggingface/datasets/blob/c9967f55626931f8059dc416526c791444cdfdf7/src/datasets/table.py#L1121
Here we lose nullability stored in `schema` because it seems that `Features` is always nullable and don't store nullability.
https://github.com/huggingface/datasets/blob/c9967f55626931f8059dc416526c791444cdfdf7/src/datasets/table.py#L1103
So, casting to schema from such `Features` loses nullability, and eventually causes error of writing with different schema

## 3. Solution
1. Let `Features` stores nullability.
2. Directly cast table with original schema but not schema from converted `Features`. (this PR)
3. Don't `cast_table` when `write_table`

### DIFF
--- a/src/datasets/table.py
+++ b/src/datasets/table.py
@@ -1102,6 +1102,21 @@ def cast_table_to_features(table: pa.Table, features: "Features"):
     arrays = [cast_array_to_feature(table[name], feature) for name, feature in features.items()]
     return pa.Table.from_arrays(arrays, schema=features.arrow_schema)
 
+def cast_table_to_schema(table: pa.Table, schema: pa.Schema):
+    """Cast an table to the arrow schema. Different from `cast_table_to_features`, this method can preserve nullability.
+
+    Args:
+        table (pa.Table): PyArrow table to cast
+        features (Features): target features.
+
+    Returns:
+        pa.Table: the casted table
+    """
+    from .features import Features
+    
+    features = Features.from_arrow_schema(schema)
+    arrays = [cast_array_to_feature(table[name], feature) for name, feature in features.items()]
+    return pa.Table.from_arrays(arrays, schema=schema)
 
 def table_cast(table: pa.Table, schema: pa.Schema):
     """Improved version of pa.Table.cast
@@ -1116,9 +1131,7 @@ def table_cast(table: pa.Table, schema: pa.Schema):
         pa.Table: the casted table
     """
     if table.schema != schema:
-        from .features import Features
-
-        return cast_table_to_features(table, Features.from_arrow_schema(schema))
+        return cast_table_to_schema(table, schema)
     elif table.schema.metadata != schema.metadata:
         return table.replace_schema_metadata(schema.metadata)
     else:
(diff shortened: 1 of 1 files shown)


src/datasets/table.py
```diff
@@ -1103,6 +1101,23 @@ def cast_table_to_features(table: pa.Table, features: "Features"):
     return pa.Table.from_arrays(arrays, schema=features.arrow_schema)
 
 
+def cast_table_to_schema(table: pa.Table, schema: pa.Schema):
+    """Cast an table to the arrow schema. Different from `cast_table_to_features`, this method can preserve nullability.
```
lhoestq, line 1105:
```suggestion
    """Cast a table to the arrow schema. Different from `cast_table_to_features`, this method can preserve nullability.
```

lhoestq:
Hi ! Thanks for reporting, indeed `disable_nullable` doesn't seem to be supported in this case. Maybe at one point we can have `disable_nullable` as a parameter of certain feature types

lhoestq (commented):
Looks good to me !

Can you run `make style` to fix the code formatting please ?

src/datasets/table.py
```diff
@@ -1103,6 +1103,23 @@ def cast_table_to_features(table: pa.Table, features: "Features"):
...
+    Args:
+        table (pa.Table): PyArrow table to cast
+        features (Features): target features.
+
+    Returns:
+        pa.Table: the casted table
+    """
+    from .features import Features
+
+    features = Features.from_arrow_schema(schema)
```
lhoestq, line 1118:
```suggestion
    features = Features.from_arrow_schema(schema)
    if sorted(table.column_names) != sorted(features):
        raise ValueError(f"Couldn't cast\n{table.schema}\nto\n{features}\nbecause column names don't match")
```

lhoestq (approved):
Thanks :)

I also added a check on the column names

PR #3783, opened 2022-02-24
Support passing str to iter_files

albertvillanova:


albertvillanova:
@mariosasko it was indeed while reading that PR, that I remembered this change I wanted to do long ago... 😉

PR #3791, opened 2022-02-25
Add `data_dir` to `data_files` resolution and misc improvements to HfFileSystem

mariosasko:
As discussed in https://github.com/huggingface/datasets/pull/2830#issuecomment-1048989764, this PR adds a QOL improvement to easily reference the files inside a directory in `load_dataset` using the `data_dir` param (very handy for ImageFolder because it avoids globbing, but also useful for the other loaders). Additionally, it fixes the issue with `HfFileSystem.isdir`, which would previously always return `False`, and aligns the path-handling logic in `HfFileSystem` with `fsspec.GitHubFileSystem`.

### DIFF
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -732,7 +732,7 @@ class LocalDatasetModuleFactoryWithoutScript(_DatasetModuleFactory):
         download_mode: Optional[DownloadMode] = None,
     ):
         if data_files is None and data_dir is not None:
-            data_files = [os.path.join(data_dir, "*"), os.path.join(data_dir, "**/*")]
+            data_files = os.path.join(data_dir, "**")
 
         self.path = path
         self.name = Path(path).stem
@@ -781,7 +781,7 @@ class PackagedDatasetModuleFactory(_DatasetModuleFactory):
         download_mode: Optional[DownloadMode] = None,
     ):
         if data_files is None and data_dir is not None:
-            data_files = [os.path.join(data_dir, "*"), os.path.join(data_dir, "**/*")]
+            data_files = os.path.join(data_dir, "**")
 
         self.name = name
         self.data_files = data_files
@@ -817,7 +817,7 @@ class HubDatasetModuleFactoryWithoutScript(_DatasetModuleFactory):
         download_mode: Optional[DownloadMode] = None,
     ):
         if data_files is None and data_dir is not None:
-            data_files = [os.path.join(data_dir, "*"), os.path.join(data_dir, "**/*")]
+            data_files = os.path.join(data_dir, "**")
 
         self.name = name
         self.revision = revision
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -1524,8 +1524,8 @@ def load_dataset_builder(
 
 
         name (:obj:`str`, optional): Defining the name of the dataset configuration.
-        data_dir (:obj:`str`, optional): Defining the data_dir of the dataset configuration. If `data_files` is not specified for the generic builders (csv, text etc.) or the Hub datasets,
-            it's equal to passing `os.path.join(data_dir, **)` as `data_files` to reference all the files inside a directory.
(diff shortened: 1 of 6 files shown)


src/datasets/load.py
```diff
@@ -803,10 +811,14 @@ def __init__(
         self,
         name: str,
         revision: Optional[Union[str, Version]] = None,
+        data_dir: Optional[str] = None,
         data_files: Optional[Union[str, List, Dict]] = None,
         download_config: Optional[DownloadConfig] = None,
         download_mode: Optional[DownloadMode] = None,
     ):
+        if data_files is None and data_dir is not None:
+            data_files = [os.path.join(data_dir, "*"), os.path.join(data_dir, "**/*")]
```
lhoestq, line 820: This would put everything in the train split no ? This goes against https://huggingface.co/docs/datasets/repository_structure.html#multiple-configuration-wip

We can work on this later in another PR

src/datasets/load.py
```diff
@@ -771,10 +775,14 @@ class PackagedDatasetModuleFactory(_DatasetModuleFactory):
     def __init__(
         self,
         name: str,
+        data_dir: Optional[str] = None,
         data_files: Optional[Union[str, List, Dict]] = None,
         download_config: Optional[DownloadConfig] = None,
         download_mode: Optional[DownloadMode] = None,
     ):
+        if data_files is None and data_dir is not None:
+            data_files = [os.path.join(data_dir, "*"), os.path.join(data_dir, "**/*")]
```
lhoestq, line 784: Does this mean that this code would only return a split "train" ?
```python
# data/ contains two files train.json and test.json
load_dataset("json", data_dir="data")  # returns only train ?
```

It can be confusing given that
```python
load_dataset("data")  # returns train and test
```

On the other hand it is normal that
```python
load_dataset("json", data_files="data/**")  # returns only train
# because it is equivalent to
load_dataset("json", data_files={"train": "data/**"})  # returns only train
```

Do you think this is fine or we need a change ?

I can imagine that some dataset builders based on standard structures would like to return several splits depending on what's inside `data_dir`

lhoestq (commented):
Cool thanks ! a few suggestions and questions about the behavior regarding the splits:

src/datasets/load.py
```diff
@@ -727,9 +727,13 @@ class LocalDatasetModuleFactoryWithoutScript(_DatasetModuleFactory):
     def __init__(
         self,
         path: str,
+        data_dir: Optional[str] = None,
         data_files: Optional[Union[str, List, Dict]] = None,
         download_mode: Optional[DownloadMode] = None,
     ):
+        if data_files is None and data_dir is not None:
+            data_files = [os.path.join(data_dir, "*"), os.path.join(data_dir, "**/*")]
```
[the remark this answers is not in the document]
lhoestq: Yes right, I forgot to remove `os.path.join(data_dir, "*")` in the suggestion

src/datasets/load.py
mariosasko: Hmm, this feature's goal is to be a shorthand syntax for `f"{data_dir}/**"`, and I'm not sure it is safe to assume that the user wants additional `data_dir` processing (automatic split globbing as we do with `data_files`, for instance) on top of that, so I'd leave this as-is. This can be changed later, depending on the feedback we get.

src/datasets/load.py
mariosasko: I don't think it "goes against" that as long as the user doesn't specify `data_dir`, and if he does, then it's safe to assume that we can turn off the repository structure inference 🙂 .

lhoestq: Ok sounds good, this will require some documentation then

src/datasets/load.py
lhoestq: Ok let's keep it this way then :)

lhoestq (approved):
Looks all good thanks ! Can you please update the docstrings to add the `data_dir` parameter and briefly explain its behavior ?

PR #3793, opened 2022-02-25
Docs new UI actions no self hosted

LysandreJik:
Removes the need to have a self-hosted runner for the dev documentation

### DIFF
--- a/.github/workflows/build_dev_documentation.yml
+++ b/.github/workflows/build_dev_documentation.yml
@@ -1,15 +1,17 @@
 name: Build dev documentation
 
 on:
+  push:
+    branches:
+      - docs_new_ui-actions-no-self-hosted
   pull_request:
 
 
 jobs:
   build_and_package:
-    runs-on: [self-hosted, doc-builder]
+    runs-on: ubuntu-latest
     container:
-      image: huggingface/doc-builder-transformers
-      options: "-v /home/github_actions:/mnt"
+      image: huggingface/transformers-doc-builder
     env:
       PR_NUMBER: ${{ github.event.number }}
       EVENT_CONTEXT: ${{ toJSON(github.event) }}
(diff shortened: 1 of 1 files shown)


LysandreJik:
It seems like the doc can't be compiled right now because of the following:

```
  File "/usr/local/bin/doc-builder", line 33, in <module>
    sys.exit(load_entry_point('doc-builder', 'console_scripts', 'doc-builder')())
  File "/__w/datasets/datasets/doc-builder/src/doc_builder/commands/doc_builder_cli.py", line 39, in main
    args.func(args)
  File "/__w/datasets/datasets/doc-builder/src/doc_builder/commands/build.py", line 95, in build_command
    build_doc(
  File "/__w/datasets/datasets/doc-builder/src/doc_builder/build_doc.py", line 361, in build_doc
    anchors_mapping = build_mdx_files(package, doc_folder, output_dir, page_info)
  File "/__w/datasets/datasets/doc-builder/src/doc_builder/build_doc.py", line 200, in build_mdx_files
    raise type(e)(f"There was an error when converting {file} to the MDX format.\n" + e.args[0]) from e
TypeError: There was an error when converting datasets/docs/source/package_reference/table_classes.mdx to the MDX format.
expected string or bytes-like object
```

lhoestq:
This is due to the injection of docstrings from PyArrow. I think I can fix that by moving all the docstrings and fix them manually.

.github/workflows/build_dev_documentation.yml
```diff
@@ -1,15 +1,17 @@
 name: Build dev documentation
 
 on:
+  push:
+    branches:
+      - docs_new_ui-actions-no-self-hosted
```
mishig25, line 6: do we need this part? since I don't see equavalent on https://github.com/huggingface/transformers/blob/master/.github/workflows/build_dev_documentation.yml

mishig25:
> It seems like the doc can't be compiled right now because of the following:

it is expected since there is something I need to change on doc-builder side.

> This is due to the injection of docstrings from PyArrow. I think I can fix that by moving all the docstrings and fix them manually.

@lhoestq I will let you know if we need to change it manually.

@LysandreJik thanks a lot for this PR! I only had one question [here](https://github.com/huggingface/datasets/pull/3793#discussion_r816100194)

mishig25 (approved):

lhoestq:
> @lhoestq I will let you know if we need to change it manually.

It would be simpler to change it manually anyway - I don't want our documentation to break if PyArrow has documentation issues

lhoestq:
For some reason it fails when `Installing node dependencies` when running `npm ci` from the `kit` directory, any idea why @mishig25 ?

.github/workflows/build_dev_documentation.yml
LysandreJik: No you're correct, this can be removed.

mishig25:
Checking it rn

.github/workflows/build_dev_documentation.yml
mishig25: I see, will keep it for now for testing reasons :)

LysandreJik:
It's very likely linked to an OOM error: https://github.com/huggingface/transformers/pull/15710#issuecomment-1051737337

PR #3794, opened 2022-02-27
Add Mahalanobis distance metric

JoaoLages:
Mahalanobis distance is a very useful metric to measure the distance from one datapoint X to a distribution P. 
In this PR I implement the metric in a simple way with the help of numpy only.

Similar to the [MAUVE implementation](https://github.com/huggingface/datasets/blob/master/metrics/mauve/mauve.py), we can make this metric accept texts as input and encode them with a featurize model, if that is desirable.

### DIFF
--- a/metrics/mahalanobis/mahalanobis.py
+++ b/metrics/mahalanobis/mahalanobis.py
@@ -21,10 +21,10 @@ import datasets
 _DESCRIPTION = """
 Compute the Mahalanobis Distance
 
-Mahalonobis distance is the distance between a point and a distribution. 
-And not between two distinct points. It is effectively a multivariate equivalent of the Euclidean distance. 
-It was introduced by Prof. P. C. Mahalanobis in 1936 
-and has been used in various statistical applications ever since 
+Mahalonobis distance is the distance between a point and a distribution.
+And not between two distinct points. It is effectively a multivariate equivalent of the Euclidean distance.
+It was introduced by Prof. P. C. Mahalanobis in 1936
+and has been used in various statistical applications ever since
 [source: https://www.machinelearningplus.com/statistics/mahalanobis-distance/]
 """
 
--- a/metrics/mahalanobis/mahalanobis.py
+++ b/metrics/mahalanobis/mahalanobis.py
@@ -66,9 +66,6 @@ class Mahalanobis(datasets.Metric):
             features=datasets.Features(
                 {
                     "X": datasets.Sequence(datasets.Value("float", id="sequence"), id="X"),
-                    "reference_distribution": datasets.Sequence(
-                        datasets.Value("float", id="sequence"), id="reference_distribution"
-                    ),
                 }
             ),
         )
--- a/metrics/mahalanobis/mahalanobis.py
+++ b/metrics/mahalanobis/mahalanobis.py
@@ -14,7 +14,9 @@
 """Mahalanobis metric."""
 
 import numpy as np
+
 import datasets
+
 
 _DESCRIPTION = """
 Compute the Mahalanobis Distance
--- a/metrics/mahalanobis/mahalanobis.py
+++ b/metrics/mahalanobis/mahalanobis.py
@@ -48,9 +48,9 @@ Returns:
 Examples:
 
     >>> mahalanobis_metric = datasets.load_metric("mahalanobis")
-    >>> results = mahalanobis_metric.compute(references=[[0, 1], [1, 0]], predictions=[0, 1])
+    >>> results = mahalanobis_metric.compute(references=[[0, 1], [1, 0]], predictions=[[0, 1]])
     >>> print(results)
-    {'mahalanobis': np.array([0.5])}
+    {'mahalanobis': array([0.5])}
 """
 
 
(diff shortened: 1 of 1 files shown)


metrics/mahalanobis/mahalanobis.py
```diff
@@ -0,0 +1,97 @@
...
+_KWARGS_DESCRIPTION = """
+Args:
+    predictions: Predicted labels, as returned by a model.
+    references: Ground truth labels.
+Returns:
+    matthews_correlation: Matthews correlation.
+Examples:
+
+    >>> mahalanobis_metric = datasets.load_metric("mahalanobis")
+    >>> results = mahalanobis_metric.compute(references=[[0, 1], [1, 0]], predictions=[[0, 1]])
```
lhoestq, line 53: `predictions` must be a single list right ?
```suggestion
    >>> results = mahalanobis_metric.compute(references=[[0, 1], [1, 0]], predictions=[0, 1])
```

metrics/mahalanobis/mahalanobis.py
```diff
@@ -0,0 +1,97 @@
...
+        # convert to numpy arrays
+        predictions = np.array(predictions)
+        references = np.array(references)
+
+        # Assert that arrays are 2D
+        assert len(predictions.shape) == 2, "Expected `predictions` to be a 2D vector"
+        assert len(references.shape) == 2, "Expected `references` to be a 2D vector"
+        assert (
+            references.shape[0] > 1
+        ), "Expected `references` to be a 2D vector with more than one element in the first dimension"
```
lhoestq, line 85: Can you `raise ValuError` instead of doing `assert` here ?

lhoestq (commented):
Thanks ! It looks very good :)

Just 2 comments:

metrics/mahalanobis/mahalanobis.py
JoaoLages: No, it can be a list of vectors. It will yield the mahalanobis distance for each of those input vectors.

metrics/mahalanobis/mahalanobis.py
JoaoLages: Added in https://github.com/huggingface/datasets/pull/3794/commits/865f680e32961872ee4a24aa9b211e59dab09346

metrics/mahalanobis/mahalanobis.py
JoaoLages: It seems that this kind of metric is not supported due to [this exception](https://github.com/huggingface/datasets/blob/207be676bffe9d164740a41a883af6125edef135/src/datasets/metric.py#L469). Do you suggest for me to change the function to only accept a single list? It will be slower to compute prediction by prediction

metrics/mahalanobis/mahalanobis.py
```diff
@@ -0,0 +1,100 @@
...
+  publisher={Elsevier}
+}
+"""
+
+_KWARGS_DESCRIPTION = """
+Args:
+    predictions: Predicted labels, as returned by a model.
+    references: Ground truth labels.
+Returns:
+    matthews_correlation: Matthews correlation.
```
lhoestq, line 49: (bad copy paste, I guess)
```suggestion
    mahalanobis: The Mahalonobis distance.
```

metrics/mahalanobis/mahalanobis.py
lhoestq: Indeed your `references` parameter is actually the definition of a distribution, not references values for each prediction. In this case, please remove `references` from the features and `_compute`, and add the `reference_distribution` parameter to `_compute`.

The example will have to be changed to
```python
results = mahalanobis_metric.compute(reference_distribution=[[0, 1], [1, 0]], predictions=[[0, 1]])
```

Under the hood it means that `predictions` is the input of the metric, and `reference_distribution` simply becomes a parameter of the metric.

(feel free to rename `predictions` and `reference_distribution` if you want btw, you can use whatever variable names make more sense)

lhoestq (commented):
Thanks, I think we can make this work with a few changes:

metrics/mahalanobis/mahalanobis.py
JoaoLages: Makes sense 👍 Thought it was a hard argument naming constraint that I had to follow.

Changed `predictions` -> `X` and `references` -> `reference_distribution` in https://github.com/huggingface/datasets/pull/3794/commits/2037075ce881901a79883880938a792362735fca

metrics/mahalanobis/mahalanobis.py
```diff
@@ -0,0 +1,102 @@
...
+        return datasets.MetricInfo(
+            description=_DESCRIPTION,
+            citation=_CITATION,
+            inputs_description=_KWARGS_DESCRIPTION,
+            features=datasets.Features(
+                {
+                    "X": datasets.Sequence(datasets.Value("float", id="sequence"), id="X"),
+                    "reference_distribution": datasets.Sequence(
+                        datasets.Value("float", id="sequence"), id="reference_distribution"
+                    ),
```
lhoestq, line 71: Thanks for the change !

Since `reference_distribution` is not an actual input (it doesn't have one reference per prediction), you can remove it from the `features`
```suggestion
```

lhoestq (approved):
It looks all good now, thanks :)

PR #3796, opened 2022-02-28
Skip checksum computation if `ignore_verifications` is `True`

mariosasko:
This will speed up the loading of the datasets where the number of data files is large (can easily happen with `imagefoler`, for instance)


lhoestq (approved):
Thank you !

PR #3798, opened 2022-02-28
Fix error message in CSV loader for newer Pandas versions

mariosasko:
Fix the error message in the CSV loader for `Pandas >= 1.4`. To fix this, I directly print the current file name in the for-loop. An alternative would be to use a check similar to this:
```python
csv_file_reader.handle.handle if datasets.config.PANDAS_VERSION >= version.parse("1.4") else csv_file_reader.f
```
CC: @SBrandeis


lhoestq (approved):
thanks :)

PR #3800, opened 2022-03-01
Added computer vision tasks

merveenoyan:
Previous PR was in my fork so thought it'd be easier if I do it from a branch. Added computer vision task datasets according to HF tasks.

### DIFF
--- a/src/datasets/utils/resources/tasks.json
+++ b/src/datasets/utils/resources/tasks.json
@@ -105,12 +105,16 @@
             "other"
         ]
     },
-    "text-to-image": {
+    "image-to-text": {
         "description": "tasks related to generating images from an existing dataset",
         "options": [
             "image-captioning",
             "other"
         ]
+    },
+    "text-to-image": {
+        "description": "tasks related to generating images from an existing dataset",
+        "options": []
     },
     "image-segmentation": {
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
Thanks !

For future references, note that the top level names come from here:
https://github.com/huggingface/huggingface_hub/blob/main/js/src/lib/interfaces/Types.ts

let us know what you think @mariosasko

src/datasets/utils/resources/tasks.json
```diff
@@ -98,6 +98,43 @@
...
     },
+    "object-detection": {
+        "description": "tasks related to detecting instances of objects from a particular class in an image",
+        "options": [
+            "face-detection",
+            "other"
+        ]
+    },
+    "image-to-text": {
+        "description": "tasks related to generating images from an existing dataset",
```
osanseviero, line 109: maybe this is clearer?
```suggestion
        "description": "tasks related to generating images from a given text",
```

osanseviero: Actually, this is the other way around I think. This should be generating text from a given image

mariosasko (approved):
I agree with @osanseviero's comment, but other than that LGTM! Well done!

osanseviero (approved):
Very cool!

PR #3805, opened 2022-03-02
Remove decode: true for image feature in head_qa

craffel:
This was erroneously added in https://github.com/huggingface/datasets/commit/701f128de2594e8dc06c0b0427c0ba1e08be3054. This PR removes it.


lhoestq (approved):
thanks for the fix !

PR #3814, opened 2022-03-03
Handle Nones in PyArrow struct

mariosasko:
This PR fixes an issue introduced by #3575 where `None` values stored in PyArrow arrays/structs would get ignored by `cast_storage` or by the `pa.array(cast_to_python_objects(..))` pattern. To fix the former, it also bumps the minimal PyArrow version to v5.0.0 to use the `mask` param in `pa.SturctArray`.

### DIFF
--- a/src/datasets/arrow_writer.py
+++ b/src/datasets/arrow_writer.py
@@ -184,9 +184,7 @@ class TypedSequence:
             elif isinstance(data, list) and data and isinstance(first_non_null_value(data)[1], np.ndarray):
                 out = list_of_np_array_to_pyarrow_listarray(data)
             else:
-                casted_data = cast_to_python_objects(data, only_1d_for_numpy=True)
-                mask = np.array([value is None for value in casted_data])
-                out = pa.array(casted_data, mask=mask)
+                out = pa.array(cast_to_python_objects(data, only_1d_for_numpy=True))
             # use smaller integer precisions if possible
             if self.trying_int_optimization:
                 if pa.types.is_int64(out.type):
@@ -211,9 +209,7 @@ class TypedSequence:
                     elif isinstance(data, list) and data and any(isinstance(value, np.ndarray) for value in data):
                         return list_of_np_array_to_pyarrow_listarray(data)
                     else:
-                        casted_data = cast_to_python_objects(data, only_1d_for_numpy=True)
-                        mask = np.array([value is None for value in casted_data])
-                        return pa.array(casted_data, mask=mask)
+                        return pa.array(cast_to_python_objects(data, only_1d_for_numpy=True))
                 except pa.lib.ArrowInvalid as e:
                     if "overflow" in str(e):
                         raise OverflowError(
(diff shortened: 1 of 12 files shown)


src/datasets/arrow_writer.py
```diff
@@ -184,8 +184,9 @@ def __arrow_array__(self, type: Optional[pa.DataType] = None):
             elif isinstance(data, list) and data and isinstance(first_non_null_value(data)[1], np.ndarray):
                 out = list_of_np_array_to_pyarrow_listarray(data)
             else:
-                out = pa.array(cast_to_python_objects(data, only_1d_for_numpy=True))
-
+                casted_data = cast_to_python_objects(data, only_1d_for_numpy=True)
+                mask = np.array([value is None for value in casted_data])
+                out = pa.array(casted_data, mask=mask)
```
lhoestq, line 189: we should keep the null values here no ?

lhoestq: Oh maybe I misunderstood: is mask removing the exampels (as in, ignoring them), or simply setting them to None ?

src/datasets/table.py
```diff
@@ -980,32 +951,26 @@ def array_cast(array: pa.Array, pa_type: pa.DataType, allow_number_to_str=True):
         if pa.types.is_struct(pa_type) and (
             set(field.name for field in pa_type) == set(field.name for field in array.type)
         ):
-            arrays = [
-                _c(array.field(field.name), field.type, allow_number_to_str=allow_number_to_str) for field in pa_type
-            ]
-            return pa.StructArray.from_arrays(arrays, fields=list(pa_type))
+            arrays = [_c(array.field(field.name), field.type) for field in pa_type]
```
lhoestq, line 954: Why did you remove `allow_number_to_str` here ?

src/datasets/arrow_writer.py
mariosasko: I think you are right. Let me remove this part and see if the newly-added tests are still green.

lhoestq: Could you explain why you need to specify `mask` here ? Isn't it inferred by pyarrow automatically ?

lhoestq (commented):
Thanks for the fix :)

lhoestq:
Looks like I added my comments while you were editing - sorry about that

src/datasets/arrow_writer.py
mariosasko: > it indeed makes them null ! nice :)
Yes, but nulls should be preserved here anyway :).

src/datasets/table.py
mariosasko: Because we already capture its value here: https://github.com/huggingface/datasets/blob/f013aff823b831fb48b458a7bbb3bb11c2981326/src/datasets/table.py#L974

I would expect an error for passing the same arg twice, but `partial`  supports this due to this line: https://github.com/python/cpython/blob/9d9dc59d07d51d73e5af7dd506d0da63aa336995/Lib/functools.py#L300

lhoestq: Ok cool !

lhoestq (approved):
Thank you !

PR #3815, opened 2022-03-03
Fix iter_archive getting reset

lhoestq:
The `DownloadManager.iter_archive` method currently returns an iterator - which is  **empty** once you iter over it once. This means you can't pass the same archive iterator to several splits.

To fix that, I changed the ouput of `DownloadManager.iter_archive` to be an iterable that you can iterate over several times, instead of a one-time-use iterator.

The `StreamingDownloadManager.iter_archive` already returns an appropriate iterable, and the code added in this PR are inspired from the one in `streaming_download_manager.py`

### DIFF
--- a/src/datasets/utils/download_manager.py
+++ b/src/datasets/utils/download_manager.py
@@ -21,7 +21,7 @@ import os
 import tarfile
 from datetime import datetime
 from functools import partial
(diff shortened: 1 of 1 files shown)


mariosasko (approved):
Thanks!

PR #3817, opened 2022-03-03
Simplify Common Voice code

lhoestq:
In #3736 we introduced one method to generate examples when streaming, that is different from the one when not streaming.

In this PR I propose a new implementation which is simpler: it only has one function, based on `iter_archive`. And you still have access to local audio files when loading the dataset in non-streaming mode.

cc @patrickvonplaten @polinaeterna @anton-l @albertvillanova since this will become the template for many audio datasets to come.

This change can also trivially be applied to the other audio datasets that already exist. Using this line, you can get access to local files in non-streaming mode:
```python
local_extracted_archive = dl_manager.extract(archive_path) if not dl_manager.is_streaming else None
```

### DIFF
--- a/datasets/common_voice/common_voice.py
+++ b/datasets/common_voice/common_voice.py
@@ -659,31 +659,37 @@ class CommonVoice(datasets.GeneratorBasedBuilder):
 
     def _split_generators(self, dl_manager):
         """Returns SplitGenerators."""
+        # Download the TAR archive that contains the audio files:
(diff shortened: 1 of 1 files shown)


polinaeterna:
I think the script looks pretty clean and readable now! cool!

patrickvonplaten (approved):
Very nice!

albertvillanova (commented):
Much cleaner!!! :+1:

PR #3824, opened 2022-03-04
Allow not specifying feature cols other than `predictions`/`references` in `Metric.compute`

### LINKED ISSUE #3818: Support for "sources" parameter in the add() and add_batch() methods in datasets.metric - SARI
The methods `add_batch` and `add` from the `Metric` [class](https://github.com/huggingface/datasets/blob/1675ad6a958435b675a849eafa8a7f10fe0f43bc/src/datasets/metric.py) does not work with [SARI](https://github.com/huggingface/datasets/blob/master/metrics/sari/sari.py) metric. This metric not only relies on the predictions and references, but also in the input. 

For example, when the `add_batch` method is used, then the `compute()` method fails:
```
metric = load_metric("sari")
metric.add_batch(
    predictions=["About 95 you now get in ."], 
    references=[["About 95 species are currently known .","About 95 species are now accepted .","95 species are now accepted ."]])
metric.compute()
> TypeError: _compute() missing 1 required positional argument: 'sources'
```

Therefore, the `compute() `method can only be used standalone:
```
metric = load_metric("sari")
result = metric.compute(
    sources=["About 95 species are currently accepted ."],
    predictions=["About 95 you now get in ."],
    references=[["About 95 species are currently known .","About 95 species are now accepted .","95 species are now accepted ."]])
> {'sari': 26.953601953601954}
```
  
Support for an additional parameter `sources` in the `add_batch` and `add` of the `Metric` class.
```
add_batch(*, sources=None, predictions=None, references=None, **kwargs)
add(*, sources=None, predictions=None, references=None, **kwargs)
compute()
```

I've tried to override the `add_batch` and `add`, however, these are highly dependent to the `Metric` class. We could also write a simple function that compute the scores of a sentences list, but then we lose the functionality from the original [add](https://huggingface.co/docs/datasets/_modules/datasets/metric.html#Metric.add) and [add_batch method](https://huggingface.co/docs/datasets/_modules/datasets/metric.html#Metric.add_batch). 

**Additional context**
These methods are used in the transformers [pytorch examples](https://github.com/huggingface/transformers/blob/master/examples/pytorch/summarization/run_summarization_no_trainer.py).

mariosasko:
Fix #3818

### DIFF
--- a/metrics/sari/sari.py
+++ b/metrics/sari/sari.py
@@ -265,6 +265,7 @@ class Sari(datasets.Metric):
             inputs_description=_KWARGS_DESCRIPTION,
             features=datasets.Features(
                 {
+                    "sources": datasets.Value("string", id="sequence"),
                     "predictions": datasets.Value("string", id="sequence"),
                     "references": datasets.Sequence(datasets.Value("string", id="sequence"), id="references"),
                 }
--- a/src/datasets/metric.py
+++ b/src/datasets/metric.py
@@ -391,11 +391,15 @@ class Metric(MetricInfoMixin):
             - None if the metric is not run on the main process (``process_id != 0``).
         """
         all_kwargs = {"predictions": predictions, "references": references, **kwargs}
-        missing_inputs = [k for k in self.features if k not in all_kwargs]
-        if missing_inputs:
-            raise ValueError(
-                f"Metric inputs are missing: {missing_inputs}. All required inputs are {list(self.features)}"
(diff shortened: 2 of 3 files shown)


lhoestq (commented):
Cool thanks ! We might have to update the test to support this. Sorry I forgot to do it when updating the metrics logic

lhoestq (approved):
Thanks for adding the test :)

PR #3834, opened 2022-03-06
Fix dead dataset scripts creation link.

dnaveenr:
Previous link gives 404 error. Updated with a new dataset scripts creation link.


lhoestq (approved):
Thanks !

PR #3842, opened 2022-03-07
Align IterableDataset.shuffle with Dataset.shuffle

lhoestq:
From #3444 , Dataset.shuffle can have the same API than IterableDataset.shuffle (i.e. in streaming mode).

Currently you can pass an optional seed to both if you want, BUT currently IterableDataset.shuffle always requires a buffer_size, used for approximate shuffling. I propose using a reasonable default value (maybe 1000) instead.

In this PR, I set the default `buffer_size` value to 1,000, and I reorder the `IterableDataset.shuffle` arguments to match `Dataset.shuffle`, i.e. making `seed` the first argument.

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -356,9 +356,9 @@ class IterableDataset(DatasetInfoMixin):
         if self._shuffling and self._epoch == 0:
             return self._shuffling.generator
         elif self._shuffling:
-            # Create effective seed using self._epoch (we substract in order to avoir overflow in long_scalars)
+            # Create effective seed using self._epoch (we subtract in order to avoir overflow in long_scalars)
             effective_seed = deepcopy(self._shuffling.generator).integers(0, 1 << 63) - self._epoch
-            effective_seed = 1 << 63 + effective seed if effective_seed < 0 else effective_seed
+            effective_seed = (1 << 63) + effective seed if effective_seed < 0 else effective_seed
             return np.random.default_rng(effective_seed)
         else:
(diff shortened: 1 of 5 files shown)


PhaniKanagala (approved):

PhaniKanagala (approved):

src/datasets/iterable_dataset.py
```diff
@@ -349,20 +352,23 @@ def __init__(
...
-    @property
-    def _effective_seed(self):
-        if self._shuffling:
-            return self._shuffling.seed + self._epoch if self._shuffling.seed is not None else None
+    def _effective_generator(self):
+        if self._shuffling and self._epoch == 0:
+            return self._shuffling.generator
+        elif self._shuffling:
+            # Create effective seed using self._epoch (we substract in order to avoir overflow in long_scalars)
+            effective_seed = deepcopy(self._shuffling.generator).integers(0, 1 << 63) - self._epoch
```
mariosasko, line 360: If I'm not mistaken, `effective_seed` can be a negative integer after this line, which will result in an error in `np.random.default_rng(effective_seed)`.
```suggestion
            effective_seed = deepcopy(self._shuffling.generator).integers(0, 1 << 63) - self._epoch
            effective_seed = 1 << 63 + effective seed if effective_seed < 0 else effective_seed
```

mariosasko (approved):
Thanks! One comment:

src/datasets/iterable_dataset.py
lhoestq: Good catch ! thanks

Issue #3608, opened 2022-01-20
Add support for continuous metrics (RMSE, MAE)


ck37:
**Is your feature request related to a problem? Please describe.**

I am uploading our dataset and models for the "Constructing interval measures" method we've developed, which uses item response theory to convert multiple discrete labels into a continuous spectrum for hate speech. Once we have this outcome our NLP models conduct regression rather than classification, so binary metrics are not relevant. The only continuous metrics available at https://huggingface.co/metrics are pearson & spearman correlation, which don't ensure that the prediction is on the same scale as the outcome.

**Describe the solution you'd like**
I would like to be able to tag our models on the Hub with the following metrics:
- RMSE
- MAE

**Describe alternatives you've considered**

I don't know if there are any alternatives.

**Additional context**
Our preprint is available here: https://arxiv.org/abs/2009.10277 . We are making it available for use in Jigsaw's Toxic Severity Rating Kaggle competition: https://www.kaggle.com/c/jigsaw-toxic-severity-rating/overview . I have our first model uploaded to the Hub at https://huggingface.co/ucberkeley-dlab/hate-measure-roberta-large

Thanks,
Chris

ariG23498:
Hey @ck37 

You can always use a custom metric as explained [in this guide from HF](https://huggingface.co/docs/datasets/master/loading_metrics.html#using-a-custom-metric-script).

If this issue needs to be contributed to (for enhancing the metric API) I think [this link](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.mean_absolute_error.html) would be helpful for the `MAE` metric.

callmekofi:
You can use a local metric script just by providing its path instead of the usual shortcut name

dnaveenr:
#self-assign I have starting working on this issue to enhance the metric API.

PR #3845, opened 2022-03-07
add RMSE and MAE metrics.

dnaveenr:
This PR adds RMSE - Root Mean Squared Error and MAE - Mean Absolute Error to the metrics API.
Both implementations are based on usage of sciket-learn.

Feature request here : Add support for continuous metrics (RMSE, MAE) [#3608](https://github.com/huggingface/datasets/issues/3608)

Please suggest any changes if required. Thank you.

### DIFF
--- a/metrics/rmse/rmse.py
+++ b/metrics/rmse/rmse.py
@@ -73,8 +73,8 @@ class Mse(datasets.Metric):
             ],
         )
 
-    def _compute(self, predictions, references):
+    def _compute(self, predictions, references, sample_weight=None, multioutput="uniform_average", squared=True):
 
-        rmse = mean_squared_error(references, predictions, squared=False)
-
-        return {"rmse": rmse}
+        mse = mean_squared_error(references, predictions, sample_weight=sample_weight, multioutput=multioutput, squared=squared)
+        
+        return {"mse": mse}
--- a/metrics/rmse/rmse.py
+++ b/metrics/rmse/rmse.py
@@ -56,7 +56,7 @@ Examples:
 
 
 @datasets.utils.file_utils.add_start_docstrings(_DESCRIPTION, _KWARGS_DESCRIPTION)
-class Rmse(datasets.Metric):
+class Mse(datasets.Metric):
     def _info(self):
         return datasets.MetricInfo(
             description=_DESCRIPTION,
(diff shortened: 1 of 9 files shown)


dnaveenr:
@mariosasko I've reopened it here. Please suggest any changes if required. Thank you.

metrics/rmse/rmse.py
```diff
@@ -0,0 +1,80 @@
...
+            reference_urls=[
+                "https://scikit-learn.org/stable/modules/generated/sklearn.metrics.mean_squared_error.html"
+            ],
+        )
+
+    def _compute(self, predictions, references):
+
+        rmse = mean_squared_error(references, predictions, squared=False)
+
+        return {"rmse": rmse}
```
mariosasko, line 80: Let's also add the missing params here (+ docstring in `KWARGS_DESCRIPTION`):
```suggestion
    def _compute(self, predictions, references, sample_weight=None, multioutput="uniform_average", squared=True):

        mse = mean_squared_error(references, predictions, sample_weight=sample_weight, multioutput=multioutput, squared=squared)
        
        return {"mse": mse}
```

mariosasko (changes requested):
Great job so far!

dnaveenr:
Thanks for suggestions. :) I have added update the KWARGS_DESCRIPTION for the missing params and also changed RMSE to MSE.
While testing, I noticed that when the input is a list of lists, we get an error :
`TypeError: float() argument must be a string or a number, not 'list'`
Could you suggest the datasets.Value() attribute to support both list of floats and list of lists containing floats ?

mariosasko:
Just add a new config to cover that case. You can do this by replacing the current `features` dict with:
```python
features=datasets.Features(
    {
        "predictions": datasets.Sequence(datasets.Value("float")),
        "references": datasets.Sequence(datasets.Value("float")),
    }
    if self.config_name == "multioutput"
    else {
        "predictions": datasets.Value("float"),
        "references": datasets.Value("float"),
     }
),
```
Feel free to suggest a better name for the config than `multioutput`

mariosasko:
Also, could you please move the changes to a new branch and open a PR from there (for the 3rd time 😄) because the diff shows changes from unrelated PRs (maybe due to rebasing?).

dnaveenr:
Thanks for the input, I have added new config to support for multi-dimensional lists and updated the examples as well.

Sure. Will do that and open a new PR for these changes.

PR #3850, opened 2022-03-08
[feat] Add tqdm arguments

penguinwang96825:
In this PR, tqdm arguments can be passed to the map() function and such, in order to be more flexible.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -59,17 +59,14 @@ from . import config, utils
 from .arrow_reader import ArrowReader
 from .arrow_writer import ArrowWriter, OptimizedTypedSequence
 from .features import (
-    Audio,
     ClassLabel,
     Features,
     FeatureType,
-    Image,
     Sequence,
     Value,
     _ArrayXD,
     decode_nested_example,
     pandas_types_mapper,
-    require_decoding,
 )
 from .filesystems import extract_path_from_uri, is_remote_filesystem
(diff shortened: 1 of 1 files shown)


src/datasets/arrow_dataset.py
```diff
@@ -2226,7 +2238,7 @@ def apply_function_on_filtered_inputs(inputs, indices, check_same_num_examples=F
                 additional_args += (effective_indices,)
             if with_rank:
                 additional_args += (rank,)
-            processed_inputs = function(*fn_args, *additional_args, **fn_kwargs)
+            processed_inputs = function(*fn_args, *additional_args, **fn_kwargs, **tqdm_kwargs)
```
lhoestq, line 2241: I don't think this is the right place to pass the `tqdm_kwargs `.

You probably want to pass it to `pbar = utils.tqdm(...)` a few lines later instead

src/datasets/arrow_dataset.py
```diff
@@ -3997,14 +4012,15 @@ def get_indices_from_mask_function(
     indices_mapping: Optional[Table] = None,
     *args,
     **fn_kwargs,
+    **tqdm_kwargs,
```
lhoestq, line 4015: Having two `**kwargs` arguments is not allowed in python.

I think you can simply remove all mentions of `**kwargs` in this function, it doesn't need it

lhoestq (commented):
Thanks for opening a new PR :)

I added some comments

PR #3862, opened 2022-03-08
Manipulate columns on IterableDataset (rename columns, cast, etc.)

lhoestq:
I added:
- add_column
- cast
- rename_column
- rename_columns

related to https://github.com/huggingface/datasets/issues/3444

TODO:
- [x] docs
- [x] tests

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -638,7 +638,7 @@ class IterableDataset(DatasetInfoMixin):
             :class:`IterableDataset`: A copy of the dataset with a renamed column.
         """
 
-        def add_new_column_fn(example):
+        def rename_column_fn(example):
             if original_column_name not in example:
                 raise ValueError(
                     f"Error when renaming {original_column_name} to {new_column_name}: column {original_column_name} is not in the dataset."
@@ -649,7 +649,7 @@ class IterableDataset(DatasetInfoMixin):
                 )
             return {new_column_name: example[original_column_name]}
 
-        return self.map(add_new_column_fn, remove_columns=[original_column_name])
+        return self.map(rename_column_fn, remove_columns=[original_column_name])
 
     def rename_columns(self, column_mapping: Dict[str, str]) -> "IterableDataset":
         """
@@ -663,7 +663,7 @@ class IterableDataset(DatasetInfoMixin):
             :class:`IterableDataset`: A copy of the dataset with renamed columns
         """
 
-        def add_new_columns_fn(example):
+        def rename_columns_fn(example):
             if any(col not in example for col in column_mapping):
                 raise ValueError(
                     f"Error when renaming {list(column_mapping)} to {list(column_mapping.values())}: columns {set(column_mapping) - set(example)} are not in the dataset."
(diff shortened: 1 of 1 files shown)


src/datasets/iterable_dataset.py
```diff
@@ -592,12 +601,84 @@ def take(self, n) -> "IterableDataset":
...
+                raise ValueError(
+                    f"Error when renaming {original_column_name} to {new_column_name}: column {original_column_name} is not in the dataset."
+                )
+            if new_column_name in example:
+                raise ValueError(
+                    f"Error when renaming {original_column_name} to {new_column_name}: column {new_column_name} is already in the dataset."
+                )
+            return {new_column_name: example[original_column_name]}
+
+        return self.map(add_new_column_fn, remove_columns=[original_column_name])
```
mariosasko, line 652:
```suggestion
        def rename_column_fn(example):
            if original_column_name not in example:
                raise ValueError(
                    f"Error when renaming {original_column_name} to {new_column_name}: column {original_column_name} is not in the dataset."
                )
            if new_column_name in example:
                raise ValueError(
                    f"Error when renaming {original_column_name} to {new_column_name}: column {new_column_name} is already in the dataset."
                )
            return {new_column_name: example[original_column_name]}

        return self.map(rename_column_fn, remove_columns=[original_column_name])
```

src/datasets/iterable_dataset.py
```diff
@@ -592,12 +601,84 @@ def take(self, n) -> "IterableDataset":
...
+            if any(col in example for col in column_mapping.values()):
+                raise ValueError(
+                    f"Error when renaming {list(column_mapping)} to {list(column_mapping.values())}: columns {set(example) - set(column_mapping.values())} are already in the dataset."
+                )
+            return {
+                new_column_name: example[original_column_name]
+                for original_column_name, new_column_name in column_mapping.items()
+            }
+
+        return self.map(add_new_columns_fn, remove_columns=list(column_mapping))
```
mariosasko, line 680:
```suggestion
        def rename_columns_fn(example):
            if any(col not in example for col in column_mapping):
                raise ValueError(
                    f"Error when renaming {list(column_mapping)} to {list(column_mapping.values())}: columns {set(column_mapping) - set(example)} are not in the dataset."
                )
            if any(col in example for col in column_mapping.values()):
                raise ValueError(
                    f"Error when renaming {list(column_mapping)} to {list(column_mapping.values())}: columns {set(example) - set(column_mapping.values())} are already in the dataset."
                )
            return {
                new_column_name: example[original_column_name]
                for original_column_name, new_column_name in column_mapping.items()
            }

        return self.map(rename_columns_fn, remove_columns=list(column_mapping))
```

mariosasko (approved):
LGTM!

IIUC we check if columns are present/not present directly in the yielded examples and not in `info.features` because `info.features` can be `None` (after `map`, for instance)? We should develop a solution that ensures `info.features` is never `None`. For example, one approach would be to infer them from examples in `map` and make them promotable from `Value("null")` to a specific type, in case of `None` values.

lhoestq:
> IIUC we check if columns are present/not present directly in the yielded examples and not in info.features because info.features can be None (after map, for instance)?

Yes exactly

> We should develop a solution that ensures info.features is never None. For example, one approach would be to infer them from examples in map and make them promotable from Value("null") to a specific type, in case of None values.

I agree this would be useful. Though inferring the type requires to start streaming some data, which takes a few seconds (compared to being instantaneous right now).

Let's discuss this in a new issue maybe ?

PR #3863, opened 2022-03-08
Update code blocks

lhoestq:
Following https://github.com/huggingface/datasets/pull/3860#issuecomment-1061756712 and https://github.com/huggingface/datasets/pull/3690 we need to update the code blocks to use markdown instead of sphinx


albertvillanova (approved):
Thanks for the update. :)

PR #3864, opened 2022-03-08
Update image dataset tags

mariosasko:
Align the existing image datasets' tags with new tags introduced in #3800.


lhoestq (approved):
Thanks ! :) LGTM

PR #3866, opened 2022-03-08
Bring back imgs so that forsk dont get broken

mishig25:


lhoestq:
I think we just need to keep `datasets_logo_name.jpg` and `course_banner.png` because they appear in the README.md of the forks of `datasets`. The other images can be removed

mishig25:
Force pushed those two imgs only

lhoestq (approved):
thanks !

PR #3867, opened 2022-03-08
Update for the rename doc-builder -> hf-doc-utils

sgugger:
This PR adapts the job to the upcoming change of name of `doc-builder`.

### DIFF
--- a/.github/workflows/build_dev_documentation.yml
+++ b/.github/workflows/build_dev_documentation.yml
@@ -11,7 +11,7 @@ jobs:
   build_and_package:
     runs-on: ubuntu-latest
     container:
-      image: huggingface/transformers-hf-doc-utils
+      image: huggingface/transformers-doc-builder
     env:
       COMMIT_SHA: ${{ github.event.pull_request.head.sha }}
       PR_NUMBER: ${{ github.event.number }}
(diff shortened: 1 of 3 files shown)


.github/workflows/build_dev_documentation.yml
```diff
@@ -11,7 +11,7 @@ jobs:
   build_and_package:
     runs-on: ubuntu-latest
     container:
-      image: huggingface/transformers-doc-builder
+      image: huggingface/transformers-hf-doc-utils
```
mishig25, line 14: was this `image` also renamed?

julien-c:
why utils? it's a builder no?

mishig25:
@julien-c there was a vote 🙂 https://huggingface.slack.com/archives/C021H1P1HKR/p1646405136644739

.github/workflows/build_dev_documentation.yml
sgugger: No, it's a mistake on my part :-)

lhoestq:
Thanks ! It looks all good to me (provided `hf-doc-utils` is the name we keep in the end). I'm fine with this name, and `hf-doc-builder` is also fine IMHO

julien-c:
ok, this is definitely not a hill I'll die on =) @mishig25 @sgugger

PR #3868, opened 2022-03-08
Ignore duplicate keys if `ignore_verifications=True`

mariosasko:
Currently, it's impossible to generate a dataset if some keys from `_generate_examples` are duplicated. This PR allows skipping the check for duplicate keys if `ignore_verifications` is set to `True`.


lhoestq:
Cool thanks ! Could you add a test please ?

lhoestq (approved):
Thank you ! LGTM

PR #3871, opened 2022-03-09
add pandas to env command

patrickvonplaten:
Pandas is a required packages and used quite a bit. I don't see any downside with adding its version to the `datasets-cli env` command.


lhoestq (approved):
Good idea, thanks :)

patrickvonplaten:
Think failures are unrelated - feel free to merge whenever you want :-)

PR #3876, opened 2022-03-09
Fix download_mode in dataset_module_factory

albertvillanova:
Fix `download_mode` value set in `dataset_module_factory`.

Before the fix, it was set to `bool` (default to `False`).

Also set properly its default value in all public functions.


lhoestq (approved):
thanks :)

PR #3882, opened 2022-03-10
Image process doc

stevhliu:
This PR is a first draft of how to process image data. It adds:

- Load an image dataset with `image` and `path` (adds tip about `decode=False` param to access the path and bytes, thanks to @mariosasko).
- Load an image using the `ImageFolder` builder. I know there is an [example](https://huggingface.co/docs/datasets/master/en/loading#image-folders) of this already, but I also wanted to add it here so users don't miss it. This doc seems important for centralizing all of the image-related things so far. Datasets has grown so quickly 🚀  now that I think maybe splitting up the How-to guides by modality may be better since working with vision/audio data is slightly different from what users have seen up until now. This way we can continue to scale the docs to better accommodate vision/audio things.
- Add a data augmentation with `set_transform`. There is only 1 example here so far, but we can certainly add more. 

Todo:
- [x] Couldn't figure out why my augmentation function works with `set_transform` but not `map` 🥲. Working with @mariosasko on this!

### DIFF
--- a/docs/source/image_process.mdx
+++ b/docs/source/image_process.mdx
@@ -124,4 +124,4 @@ Now you can visualize the results of the `ColorJitter` transform:
 >>> plt.imshow(img.permute(1, 2, 0))
 ```
 
-![image_process_jitter](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/image_process_dog.png)
+![image_process_jitter](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/image_process_jitter.png)
--- a/docs/source/image_process.mdx
+++ b/docs/source/image_process.mdx
@@ -1,0 +1,127 @@
+# Process image data
+
+🤗 Datasets support loading and processing images with the [`~Image`] feature. This guide will show you how to:
+
+- Load an image dataset.
+- Load a generic image dataset with `ImageFolder`.
+- Add data augmentations to your images.
+- Use [`~Dataset.map`] to quickly apply your transforms to the entire dataset.
+
+## Image datasets
+
+The images in an image dataset are typically either a:
+
+- PIL `image`.
+- Path to an image file that you can load. 
+
+For example, load the [Food-101](https://huggingface.co/datasets/food101) dataset and take a look:
+
+```py
+>>> from datasets import load_dataset, Image
+
+>>> dataset = load_dataset("food101", split="train[100:]")
+>>> dataset[0]
+{'image': <PIL.JpegImagePlugin.JpegImageFile image mode=RGB size=384x512 at 0x7FC45AB5C590>,
+ 'label': 6}
+```
+
+The [`~Image`] feature automatically decodes the data from the `image` column to return an image object. Now try and call the `image` column to see what the image is:
+
+```py
+>>> from datasets import load_dataset, Image
+
+>>> dataset = load_dataset("food101", split="train[100:200]")
+>>> dataset[0]["image"]
+```
+
+![image_process_beignet](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/image_process_beignet.png)
+
+
+To load an image from its path, use the [`~Dataset.cast_column`] method. The [`~Image`] feature will decode the data at the path to return an image object:
+
+```py
+>>> from datasets import load_dataset, Image
+
+>>> dataset = load_dataset("cats_vs_dogs").cast_column("image_file_path", Image())
+>>> dataset[0]["image_file_path"]
+```
+
+![image_process_dog](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/image_process_dog.png)
+
+<Tip>
+
+You can also access the path and bytes of an image file by setting `decode=False` when you load a dataset. In this case, you will need to cast the `image` column:
+
+`dataset = load_dataset("food101", split="train[100:]").cast_column('image', Image(decode=False))`
+
+</Tip>
+
+## ImageFolder
+
+You can also load your image dataset with our `ImageFolder` dataset builder without writing a custom dataloader. Your image dataset structure should look like this:
+
+```
+folder/dog/golden_retriever.png
+folder/dog/german_shepherd.png
+folder/dog/chihuahua.png
+
+folder/cat/maine_coon.png
+folder/cat/bengal.png
+folder/cat/birman.png
+```
+
+Then you can load your dataset by specifying `imagefolder` and the directory of your dataset in `data_dir`:
+
+```py
+>>> from datasets import load_dataset
+>>> dataset = load_dataset("imagefolder", data_dir="/path/to/folder")
+```
+
+Load remote datasets from their URLs with the `data_files` parameter:
+
+```py
+dataset = load_dataset("imagefolder", data_files="https://download.microsoft.com/download/3/E/1/3E1C3F21-ECDB-4869-8368-6DEBA77B919F/kagglecatsanddogs_3367a.zip", split="train")
+```
+
+## Data augmentation
+
+Adding data augmentations to a dataset is common to prevent overfitting and achieve better performance. You can use any library or package you want to apply the augmentations. This guide will use the transforms from [torchvision](https://pytorch.org/vision/stable/transforms.html).
+
+Add the [`ColorJitter`](https://pytorch.org/vision/stable/transforms.html#torchvision.transforms.ColorJitter) transform to change the color properties of the image randomly:
+
+```py
+>>> from torchvision.transforms import Compose, ColorJitter, ToTensor
+
+>>> jitter = Compose(
+    [
+     ColorJitter(brightness=0.25, contrast=0.25, saturation=0.25, hue=0.7),
+     ToTensor(),
+    ]
+)
+```
+
+Create a function to apply the `ColorJitter` transform to an image:
+
+```py
+>>> def transforms(examples):
+...     examples["pixel_values"] = [jitter(image.convert("RGB")) for image in examples["image"]]
+...     return examples
+```
+
+Use the [`~Dataset.set_transform`] function to apply the transformation on-the-fly. This is generally faster than [`~Dataset.map`] because it only prepares the data when it is used, instead of preparing it across the entire dataset ahead of time:
+
+```py
+>>> dataset.set_transform(transforms)
+```
+
+Now you can visualize the results of the `ColorJitter` transform:
+
+```py
+>>> import numpy as np
+>>> import matplotlib.pyplot as plt
+
+>>> img = dataset[0]["pixel_values"]
+>>> plt.imshow(img.permute(1, 2, 0))
+```
+
+![image_process_jitter](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/image_process_dog.png)
--- a/docs/source/image_process.mdx
+++ b/docs/source/image_process.mdx
@@ -19,7 +19,7 @@ For example, load the [Food-101](https://huggingface.co/datasets/food101) datase
 ```py
 >>> from datasets import load_dataset, Image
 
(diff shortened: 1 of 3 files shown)


docs/source/image_process.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+![image_process_beignet](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/image_process_beignet.png)
+
+
+To load an image from its path, use the [`~Dataset.cast_column`] method. The [`~Image`] feature will decode the data at the path to return an image object:
+
+```py
+>>> from datasets import load_dataset, Image
+
+>>> dataset = load_dataset("cats_vs_dogs").cast_column("image_file_path", Image())
```
mariosasko, line 45: I find this example a bit weird because here we add a `Image` column to the dataset that already has one. IMO, this would make more sense:
```suggestion
>>> dataset = Dataset.from_dict({"image_file_path": ["path/to/image_1", "path/to/image_2, ..., path/to/image_n"]}).cast_column("image_file_path", Image())
```

docs/source/image_process.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+<Tip>
+
+You can also access the path and bytes of an image file by setting `decode=False` when you load a dataset. In this case, you will need to cast the `image` column:
+
+`dataset = load_dataset("food101", split="train[100:]").cast_column('image', Image(decode=False))`
+
+</Tip>
+
+## ImageFolder
```
mariosasko, line 59: WDYT about deleting [this ImageFolder section](https://huggingface.co/docs/datasets/master/en/loading#image-folders) to have only this one, because ImageFolder is not as generic loader as CSV, JSON. It's a bit more specific and associated with CV. 

cc @lhoestq for his opinion on this

docs/source/image_process.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+Create a function to apply the `ColorJitter` transform to an image:
+
+```py
+>>> def transforms(examples):
+...     examples["pixel_values"] = [jitter(image.convert("RGB")) for image in examples["image"]]
+...     return examples
+```
+
+Use the [`~Dataset.set_transform`] function to apply the transformation on-the-fly. This is generally faster than [`~Dataset.map`] because it only prepares the data when it is used, instead of preparing it across the entire dataset ahead of time:
```
mariosasko, line 111: This is **slower** than `map` unless examples are accessed only once. However, `set_transform` consumes less memory because `map` generates a cache file the first time.

mariosasko (commented):
Thanks, looks really good!

docs/source/image_process.mdx
lhoestq: "`set_transform ` consumes less memory" -> "`set_transform ` consumes less disk space" no ?

docs/source/image_process.mdx
lhoestq: Yes I agree, let's not duplicate the documentation. Feel free to redirect users that are looking for vision-specific formats from `loading.mdx` to `image_process.mdx` though ^^

docs/source/image_process.mdx
mariosasko: Yes, less disk space. 🙂

docs/source/image_process.mdx
stevhliu: Good point, I wanted to use a real dataset example but this makes more sense!

docs/source/image_process.mdx
```diff
@@ -0,0 +1,133 @@
...
+
+## Map
+
+[`~Dataset.map`] can also apply transforms over an entire dataset. When you use [`~Dataset.map`] to apply transforms, it generates a cache file. This saves time because you don't have to execute the same transform twice. You should use [`~Dataset.map`] for operations you only run once per training - like resizing an image - instead of using it for operations that are executed for each epoch, like data augmentations.
+
+To apply the same `ColorJitter` transform from above with [`~Dataset.map`], you need to set `batched=True`:
+
+```py
+>>> processed_food = dataset.map(transforms, batched=True)
+```
```
mariosasko, line 133: One last nit:

Let's also add a transform that returns `PIL.Image` objects to show it's possible to cache them as well (e.g. resize).

Additionally, it makes sense to add a tip explaining how to reduce memory requirements of `map` with `batch_size` (controls the number of examples that are processed in a single call to the transform function) and `writer_batch_size` (controls the number of the transformed examples that are kept in memory before they are "flushed" into an arrow cache file) because `batch_size`/`writer_batch_size` is 1000 by default so it can be expensive to store 1000 images in RAM.

docs/source/image_process.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+```py
+>>> from torchvision.transforms import Compose, ColorJitter, ToTensor
+
+>>> jitter = Compose(
+... [
+...  ColorJitter(brightness=0.25, contrast=0.25, saturation=0.25, hue=0.7),
+...  ToTensor(),
+...  ]
+... )
```
mariosasko, line 96: `black` formatting 😄: 
```suggestion
>>> jitter = Compose(
...     [
...          ColorJitter(brightness=0.25, contrast=0.25, saturation=0.25, hue=0.7),
...          ToTensor(),
...     ]
... )
```

docs/source/image_process.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+Create a simple [`Resize`](https://pytorch.org/vision/stable/generated/torchvision.transforms.Resize.html) function:
+
+```py
+>>> from torchvision.transforms import Resize
+
+>>> resize = Resize([100,100])
+>>> def transforms(examples):
+...     examples["pixel_values"] = [resize(image.convert("RGB")) for image in examples["image"]]
+...     return examples
```
mariosasko, line 137: Let's avoid importing `torchvision` here to not to get angry the TF/Jax folks that want to run this example:
```suggestion
>>> def transforms(examples):
...     examples["pixel_values"] = [image.convert("RGB").resize((100, 100)) for image in examples["image"]]
...     return examples
```

mariosasko (approved):
Great job @stevhliu! Two minor nits.

docs/source/image_process.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+- PIL `image`.
+- Path to an image file you can load. 
+
+For example, load the [Food-101](https://huggingface.co/datasets/food101) dataset and take a look:
+
+```py
+>>> from datasets import load_dataset, Image
+
+>>> dataset = load_dataset("food101", split="train[100:]")
```
lhoestq, line 22: Why `train[100:]` and not just `train` ? (sorry if you've already discussed this)

docs/source/image_process.mdx
```diff
@@ -0,0 +1,154 @@
...
+![image_process_beignet](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/image_process_beignet.png)
+
+
+To load an image from its path, use the [`~Dataset.cast_column`] method. The [`Image`] feature will decode the data at the path to return an image object:
+
+```py
+>>> from datasets import load_dataset, Image
+
+>>> dataset = Dataset.from_dict({"image_file_path": ["path/to/image_1", "path/to/image_2", ..., "path/to/image_n"]}).cast_column("image_file_path", Image())
+>>> dataset[0]["image_file_path"]
```
lhoestq, line 46: I would simply call it `image` since it's an Image feature type
```suggestion
>>> dataset = Dataset.from_dict({"image": ["path/to/image_1", "path/to/image_2", ..., "path/to/image_n"]}).cast_column("image", Image())
>>> dataset[0]["image"]
```

docs/source/image_process.mdx
```diff
@@ -0,0 +1,154 @@
...
+```py
+>>> from datasets import load_dataset
+>>> dataset = load_dataset("imagefolder", data_dir="/path/to/folder")
+```
+
+Load remote datasets from their URLs with the `data_files` parameter:
+
+```py
+>>> dataset = load_dataset("imagefolder", data_files="https://download.microsoft.com/download/3/E/1/3E1C3F21-ECDB-4869-8368-6DEBA77B919F/kagglecatsanddogs_3367a.zip", split="train")
+```
```
lhoestq, line 80: Maybe it would be nice to explain that there is a `label` column and say that the label is based on the directory name ?

docs/source/image_process.mdx
```diff
@@ -0,0 +1,154 @@
...
+>>> def transforms(examples):
+...     examples["pixel_values"] = [jitter(image.convert("RGB")) for image in examples["image"]]
+...     return examples
+```
+
+Then you can use the [`~Dataset.set_transform`] function to apply the transform on-the-fly to consume less disk space. Use this function if you only need to access the examples once:
+
+```py
+>>> dataset.set_transform(transforms)
+```
```
lhoestq, line 111: Maybe add s note that one can also use any data augmentation tool they want, like https://github.com/albumentations-team/albumentations for example

docs/source/image_process.mdx
```diff
@@ -0,0 +1,154 @@
...
+...          ToTensor(),
+...     ]
+... )
+```
+
+Create a function to apply the `ColorJitter` transform to an image:
+
+```py
+>>> def transforms(examples):
+...     examples["pixel_values"] = [jitter(image.convert("RGB")) for image in examples["image"]]
```
lhoestq, line 103: Here you can just replace the `image` field no ?

docs/source/image_process.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+## Map
+
+[`~Dataset.map`] can also apply transforms over an entire dataset and it also generates a cache file.
+
+Create a simple [`Resize`](https://pytorch.org/vision/stable/generated/torchvision.transforms.Resize.html) function:
+
+```py
+>>> def transforms(examples):
+...     examples["pixel_values"] = [image.convert("RGB").resize((100,100)) for image in examples["image"]]
```
lhoestq, line 133: same here

docs/source/image_process.mdx
```diff
@@ -0,0 +1,154 @@
...
+>>> import numpy as np
+>>> import matplotlib.pyplot as plt
+
+>>> img = dataset[0]["pixel_values"]
+>>> plt.imshow(img.permute(1, 2, 0))
+```
+
+![image_process_jitter](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/image_process_jitter.png)
+
+## Map
```
lhoestq, line 125: Usually the dataset preparation workflow does loading > preprocessing > data augmentation (during training)

Should this page follow the same order ?

lhoestq (commented):
Awesome thank you ! I've added a few comments, but feel free to ignore if you've already discussed them

docs/source/image_process.mdx
stevhliu: I wanted to load and test with a smaller number of examples first, but we can use the full `train` if you'd like!

lhoestq: Then it should be `train[:100]` no ? :P

lhoestq (approved):
Thanks ! It looks all good :D

docs/source/image_process.mdx
```diff
@@ -0,0 +1,161 @@
+# Process image data
+
+🤗 Datasets support loading and processing images with the [`Image`] feature. This guide will show you how to:
+
+- Load an image dataset.
+- Load a generic image dataset with `ImageFolder`.
+- Add data augmentations to your images with [`Dataset.set_transform`].
+- Use [`~Dataset.map`] to quickly apply transforms to an entire dataset.
```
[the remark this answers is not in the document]
lhoestq: I took the liberty of swapping the order here :)

Feel free to merge if it's all good to you @stevhliu , we'll do the release after this one is merged

PR #3886, opened 2022-03-10
Retry HfApi call when 504 error inside push_to_hub

### LINKED ISSUE #3872: HTTP error 504 Server Error: Gateway Time-out
I am trying to push a large dataset(450000+) records with the help of `push_to_hub()`
While pushing, it gives some error like this.

```
  File "data_split_speech.py", line 159, in <module>
    data_new_2.push_to_hub("user-name/dataset-name",private=True)
  File "/opt/conda/lib/python3.8/site-packages/datasets/dataset_dict.py", line 951, in push_to_hub
    repo_id, split, uploaded_size, dataset_nbytes = self[split]._push_parquet_shards_to_hub(
  File "/opt/conda/lib/python3.8/site-packages/datasets/arrow_dataset.py", line 3556, in _push_parquet_shards_to_hub
    api.upload_file(
  File "/opt/conda/lib/python3.8/site-packages/huggingface_hub/hf_api.py", line 1017, in upload_file
    raise err
  File "/opt/conda/lib/python3.8/site-packages/huggingface_hub/hf_api.py", line 1008, in upload_file
    r.raise_for_status()
  File "/opt/conda/lib/python3.8/site-packages/requests/models.py", line 953, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 504 Server Error: Gateway Time-out for url: https://huggingface.co/api/datasets/user-name/dataset-name/upload/main/data/train2-00041-of-00064.parquet

```
Can anyone help me to resolve this issue.

albertvillanova:
Ass suggested by @lhoestq in #3872, this PR:
- Implements a retry function
- Retries HfApi call when 504 error. To be agreed:
  - max_retries = 2 (at 0.5 and 1 seconds)

Fix #3872.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -97,7 +97,7 @@ from .table import (
 )
 from .tasks import TaskTemplate
 from .utils import logging
-from .utils.file_utils import estimate_dataset_size
+from .utils.file_utils import _retry, estimate_dataset_size
 from .utils.info_utils import is_small_dataset
 from .utils.py_utils import temporary_assignment, unique_values
 from .utils.streaming_download_manager import xgetsize
@@ -3484,14 +3484,20 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
             buffer = BytesIO()
             shard.to_parquet(buffer)
             uploaded_size += buffer.tell()
-            api.upload_file(
-                path_or_fileobj=buffer.getvalue(),
-                path_in_repo=path_in_repo(index),
-                repo_id=repo_id,
-                token=token,
-                repo_type="dataset",
-                revision=branch,
-                identical_ok=True,
+            _retry(
+                api.upload_file,
+                func_kwargs=dict(
+                    path_or_fileobj=buffer.getvalue(),
+                    path_in_repo=path_in_repo(index),
+                    repo_id=repo_id,
+                    token=token,
+                    repo_type="dataset",
+                    revision=branch,
+                    identical_ok=True,
+                ),
+                exceptions=HTTPError,
+                status_codes=[504],
+                max_retries=2,
             )
         return repo_id, split, uploaded_size, dataset_nbytes
 
--- a/src/datasets/utils/file_utils.py
+++ b/src/datasets/utils/file_utils.py
@@ -20,7 +20,7 @@ from dataclasses import dataclass
 from functools import partial
 from hashlib import sha256
 from pathlib import Path
-from typing import Dict, Optional, TypeVar, Union
+from typing import Dict, Optional, Type, TypeVar, Union
 from urllib.parse import urljoin, urlparse
 
 import numpy as np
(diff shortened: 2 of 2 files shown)


src/datasets/arrow_dataset.py
```diff
@@ -3484,14 +3484,20 @@ def delete_file(file):
...
+                    path_in_repo=path_in_repo(index),
+                    repo_id=repo_id,
+                    token=token,
+                    repo_type="dataset",
+                    revision=branch,
+                    identical_ok=True,
+                ),
+                exceptions=HTTPError,
+                status_codes=[504],
+                max_retries=2,
```
lhoestq, line 3500: You can set the max_wait_time to a much bigger value than the default 2sec, since an upload operation can take some time before failing, I guess ? . Maybe 30sec ?

lhoestq (approved):
Thank you so much !

src/datasets/arrow_dataset.py
albertvillanova: @lhoestq thanks for your comment, because I don't know which values could be optimal/reasonable.

Just to be sure we understand the workflow:
- we use `requests`, which makes synchronous requests
- for each shard, we perform a synchronous POST request
- this can take long time to be finished until we receive the server response (I understand you are referring to this time), but, because of synch, the execution is blocked until we receive the server response
- it is just when we receive the server response with error status code 504, that we should retry the same POST request, but only after some sleep time
- this is where `max_wait_time` is relevant: this is the max time we will sleep since we receive the 504 error until we resend the same request

In order to decide the value of this, we should know: when a server sends a 504 error, how long does it typically take to be operative again, so that it can take requests normally?

This time is different from the time the server typically take to process one request, which is what I think you are referring in your comment.

lhoestq: Oh I see your point, right !

> In order to decide this value, we should know: when a server sends a 504 error, how long does it typically take to be operative again, so that it can take requests normally?

From my experience it can take up to a minute, but I'm not sure what causes them in the first place

julien-c: > From my experience it can take up to a minute

is this in prod or staging?

lhoestq: For prod. I haven't played with staging that much to experience this on staging

lhoestq:
I made it more robust by increasing the wait time, and I also added some logs when a request is retried. Let me know if it's ok for you

albertvillanova:
At the end you did not set the agreed max value of 60s. 

Moreover, with the new numbers, there is a slight contradiction: although you set max_retries=5, we will only make 3 retries at most because of the combined values of `base_wait_time` and `max_wait_time`.

lhoestq:
Yea I thought that in total we could wait 1min, but if we have a max_wait_time of 20sec between each request it's fine IMO

> Moreover, with the new numbers, there is a slight contradiction: although you set max_retries=5, we will only make 4 retries at most because of the combined values of base_wait_time and max_wait_time.

What makes you think this ? If the exponential wait time becomes bigger than `max_wait_time` then it still does the retry, but after a wait time of `max_wait_time`

albertvillanova:
Sorry, I meant 4 retries **with exponential backoff**; the fifth one is with constant backoff.

albertvillanova:
OK, and one question: do you think that the retries do not affect the time the server needs for being able to process the request (I guess in this case the cause are other users' requests, or other causes; not our specific request).

Just to be sure: 
- Then 20s at most between consecutive requests do not impact the server.
- And we expect after a total of 5 retries (within a total 50s), the server should be able to come back to normality.

lhoestq:
> do you think that the retries do not affect the time the server needs for being able to process the request (I guess in this case the cause are other users' requests, or other causes; not our specific request).

Yes I don't think the retries would affect the server, I think the cause of the 504 errors is elsewhere

> Just to be sure:
>
> Then 20s at most between consecutive requests do not impact the server.
> And we expect after a total of 5 retries (within a total 50s of wait time + request processing/uploading time), the server should be able to come back to normality.

Yes I think it's fine for now, we can still adapt this later if needed

PR #3890, opened 2022-03-10
Update beans download urls

### LINKED ISSUE #3889: Cannot load beans dataset (Couldn't reach the dataset)
## Describe the bug
The beans dataset is unavailable to download.

## Steps to reproduce the bug
```python
ds = load_dataset('beans')
```

## Expected results
The dataset would be downloaded with no issue.

## Actual results

```
ConnectionError: Couldn't reach https://storage.googleapis.com/ibeans/train.zip (error 403)
```

It looks like the billing of this project has been disabled because it is associated with a delinquent account.

## Environment info
Google Colab

mariosasko:
Replace the old URLs with the Hub [URLs](https://huggingface.co/datasets/beans/tree/main/data).

Also reported by @stevhliu. 

Fix #3889


albertvillanova (changes requested):
Thanks for the fix, @mariosasko.

Currently, when loading it gives `NonMatchingChecksumError` because of the `None` checksum values you introduced in the `dataset_infos.json` file.

Maybe you should regenerate the `dataset_infos.json`, once merged:
- #3892

mariosasko:
@albertvillanova Thanks for investigating and fixing that issue. I regenerated the `dataset_infos.json` file.

albertvillanova (commented):
Thanks!

PR #3893, opened 2022-03-11
Add default branch for doc building

sgugger:
Since other libraries use `main` as their default branch and it's now the standard default, you have to specify a different name in the doc config if you're using `master` like datasets (`doc-builder` tries to guess it, but in the job, we have weird checkout of merge commits so it doesn't always manage to get it right).

This PR makes sure it will always use master for the dev doc (until you decide to switchto main)

### DIFF
--- a/docs/source/_config.py
+++ b/docs/source/_config.py
@@ -1,0 +1,1 @@
+default_branch_name = "master"


lhoestq (approved):
Thanks ! For reference and if I understand correctly, this change follows the default branch check change at https://github.com/huggingface/doc-builder/pull/135

sgugger:
Yes! And when we discovered on the Transformers side that this check fails on the GitHub actions, we added a config attribute to have a default. Setting in Transformers fixed the issue of the doc being deployed to main, so porting the fix here too :-)

PR #3897, opened 2022-03-11
Align tqdm control/cache control with Transformers

### LINKED ISSUE #3586: Revisit `enable/disable_` toggle function prefix
As discussed in https://github.com/huggingface/transformers/pull/15167, we should revisit the `enable/disable_` toggle function prefix, potentially in favor of `set_enabled_`. Concretely, this translates to

- De-deprecating `disable_progress_bar()`
- Adding `enable_progress_bar()`
- On the caching side, adding `enable_caching` and `disable_caching`

Additional decisions have to be made with regards to the existing `set_enabled_X` functions; that is, whether to keep them as is or deprecate them in favor of the aforementioned functions. 

cc @mariosasko @lhoestq

mariosasko:
This PR:
* aligns the `tqdm` logic with Transformers (follows https://github.com/huggingface/transformers/pull/15167) by moving the code to `utils/logging.py`, adding `enable_progres_bar`/`disable_progres_bar` and removing `set_progress_bar_enabled` (a note for @lhoestq: I'm not adding `logging.tqdm` to the public namespace in this PR to avoid the situation where `from datasets import *; tqdm` would overshadow the standard `tqdm` 
* aligns the cache control with the new `tqdm` logic by adding `enable_caching`/`disable_caching` to the  public namespace and deprecating `set_caching_enabled` (not fully removing it because it's used more often than `set_progress_bar_enabled` and has a dedicated example in the old docs)

Fix #3586


src/datasets/utils/logging.py
```diff
@@ -167,3 +169,68 @@ def enable_propagation() -> None:
...
+        return empty_fn
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, type_, value, traceback):
+        return
+
+
+_active = True
```
lhoestq, line 198: Maybe rename this `_tqdm_active` ?

src/datasets/fingerprint.py
```diff
@@ -92,6 +93,49 @@ def get_datasets_with_cache_file_in_temp_dir():
...
+    - cache files are named using a random hash instead of the dataset fingerprint
+    - use :func:`datasets.Dataset.save_to_disk` to save a transformed dataset or it will be deleted when session closes
+    - caching doesn't affect :func:`datasets.load_dataset`. If you want to regenerate a dataset from scratch you should use
+    the ``download_mode`` parameter in :func:`datasets.load_dataset`.
+    """
+    global _CACHING_ENABLED
+    _CACHING_ENABLED = False
+
+
+@deprecated("Use `enable_caching` / `disable_caching` instead.")
```
lhoestq, line 138:
```suggestion
@deprecated("Use datasets.enable_caching() or datasets.disable_caching() instead. This function will be removed in a future version of datasets.")
```

PR #3898, opened 2022-03-11
Create README.md for WER metric

sashavor:
Proposing a draft WER metric card, @lhoestq I'm not very certain about "Values from popular papers" -- I don't know ASR very well, what do you think of the examples I found?

### DIFF
--- a/metrics/wer/README.md
+++ b/metrics/wer/README.md
@@ -3,7 +3,7 @@
 ## Metric description
 Word error rate (WER) is a common metric of the performance of an automatic speech recognition (ASR) system. 
 
-The general difficulty of measuring performance lies in the fact that the recognized word sequence can have a different length from the reference word sequence (supposedly the correct one). The WER is derived from the [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance), working at the word level instead of the phoneme level. 
+The general difficulty of measuring the performance of ASR systems lies in the fact that the recognized word sequence can have a different length from the reference word sequence (supposedly the correct one). The WER is derived from the [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance), working at the word level.
 
 This problem is solved by first aligning the recognized word sequence with the reference (spoken) word sequence using dynamic string alignment. Examination of this issue is seen through a theory called the power law that states the correlation between [perplexity](https://huggingface.co/metrics/perplexity) and word error rate.
 
--- a/metrics/wer/README.md
+++ b/metrics/wer/README.md
@@ -47,7 +47,7 @@ This value indicates the average number of errors per reference word.
 
 The **lower** the value, the **better** the performance of the ASR system, with a WER of 0 being a perfect score.
 
-## Popular papers that use this metric
+### Values from popular papers
 
 This metric is highly dependent on the content and quality of the dataset, and therefore users can expect very different values for the same model but on different datasets.
 
--- a/metrics/wer/README.md
+++ b/metrics/wer/README.md
@@ -5,7 +5,7 @@ Word error rate (WER) is a common metric of the performance of an automatic spee
 
 The general difficulty of measuring the performance of ASR systems lies in the fact that the recognized word sequence can have a different length from the reference word sequence (supposedly the correct one). The WER is derived from the [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance), working at the word level.
 
-This problem is solved by first aligning the recognized word sequence with the reference (spoken) word sequence using dynamic string alignment. Examination of this issue is seen through a theory called the power law that states the correlation between [perplexity](https://huggingface.co/metrics/perplexity) and word error rate.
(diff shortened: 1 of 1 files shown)


julien-c:
For ASR you can probably ping @patrickvonplaten

metrics/wer/README.md
```diff
@@ -0,0 +1,118 @@
+# Metric Card for WER
+
+## Metric description
+Word error rate (WER) is a common metric of the performance of an automatic speech recognition (ASR) system. 
+
+The general difficulty of measuring performance lies in the fact that the recognized word sequence can have a different length from the reference word sequence (supposedly the correct one). The WER is derived from the [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance), working at the word level instead of the phoneme level.
```
patrickvonplaten, line 6:
```suggestion
The general difficulty of measuring performance of ASR systems lies in the fact that the recognized word sequence can have a different length from the reference word sequence (supposedly the correct one). The WER is derived from the [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance), working at the word level.
```

metrics/wer/README.md
```diff
@@ -0,0 +1,118 @@
+# Metric Card for WER
+
+## Metric description
+Word error rate (WER) is a common metric of the performance of an automatic speech recognition (ASR) system. 
+
+The general difficulty of measuring performance lies in the fact that the recognized word sequence can have a different length from the reference word sequence (supposedly the correct one). The WER is derived from the [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance), working at the word level instead of the phoneme level. 
+
+This problem is solved by first aligning the recognized word sequence with the reference (spoken) word sequence using dynamic string alignment. Examination of this issue is seen through a theory called the power law that states the correlation between [perplexity](https://huggingface.co/metrics/perplexity) and word error rate.
```
patrickvonplaten, line 8:
```suggestion
The WER is computed by first aligning the recognized word sequence with the reference word sequence using dynamic word alignment. 

Theory on the meaningfulness of WER is closely linked to the so-called power law that states the correlation between [perplexity](https://huggingface.co/metrics/perplexity) and word error rate.
```

Do you think we could maybe add a link (wikipedia, article, paper) to the last sentence so that the reader - if interested - could take a closer look at the link between WER and perplexity?

metrics/wer/README.md
```diff
@@ -0,0 +1,118 @@
...
+```
+print(wer_score)
+0.5
+```
+
+This value indicates the average number of errors per reference word. 
+
+The **lower** the value, the **better** the performance of the ASR system, with a WER of 0 being a perfect score.
+
+## Values from popular papers
```
patrickvonplaten, line 50: I think I would actually delete this section as it's not very relevant to the definition of WER IMO. WER's are always relative, therefore I'm not sure how meaningful absolute comparisons of WER values are. Also common datasets and "recent" systems are always changing so I think it would be very difficult to keep this section up to date

patrickvonplaten, line 50:
```suggestion
```

metrics/wer/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+While initial papers such as [Graves (2014)](http://proceedings.mlr.press/v32/graves14.pdf) reported a WER of 0.27, recent systems such as [Word2vec 2.0](https://arxiv.org/pdf/2006.11477.pdf) have a much lower WER, ranging from 0.018 to 0.033). 
+
+WER values will depend on the dataset being used for evaluation, with datasets such as [LibriSpeech](https://huggingface.co/datasets/librispeech_asr), [Common Voice](https://huggingface.co/datasets/common_voice) and [Timit](https://huggingface.co/datasets/timit_asr) commonly used for evaluating ASR systems.
```
patrickvonplaten, line 54:
```suggestion
```

metrics/wer/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+```python
+from datasets import load_metric
+wer = load_metric("wer")
+predictions = ["this is the prediction", "there is an other sample"]
+references = ["this is the reference", "there is another one"]
+wer_score = wer.compute(predictions=predictions, references=references)
+print(wer_score)
+0.5
+```
```
patrickvonplaten, line 80:
```suggestion
```

The `wer_score` of 0.5 can be computed as follows:

`WER = (S + D + I) / N = (3 + 0 + 1) / 8 = 4/8 = 0.5`, with

- 3 substitions: *prediction* -> *reference*, *an* -> *another*, and *sample* -> *one*
- 1 insertion: *other*
```

patrickvonplaten (approved):
Thanks a lot for the metric card addition here. That's very useful :-) I've left some comments for improvement - feel free to disregard/discuss!

metrics/wer/README.md
sashavor: This is part of the template for metric cards (and we are trying to keep that standard across metrics). 
I tried to find some historical/popular values, and I don't think we need to keep this updated, it's just to give people an idea of the evolution of WER

sashavor: @patrickvonplaten do you have any datasets in mind that we can use?

sashavor:
I put "Values from popular papers" as a subsection of "Output values" -- I hope that's a compromise that works for everyone :hugs:

lhoestq (approved):
Sounds good, thanks !

PR #3907, opened 2022-03-14
Update README.md for SQuAD metric

sashavor:
Putting "Values from popular papers" as a subsection of "Output values"


lhoestq (approved):
thanks :)

Issue #3804, opened 2022-03-02
Text builder with custom separator line boundaries


cronoik:
**Is your feature request related to a problem? Please describe.**
The current [Text](https://github.com/huggingface/datasets/blob/207be676bffe9d164740a41a883af6125edef135/src/datasets/packaged_modules/text/text.py#L23) builder implementation splits texts with `splitlines()` which splits the text on several line boundaries. Not all of them are always wanted. 

**Describe the solution you'd like**
```python
                if self.config.sample_by == "line":
                    batch_idx = 0
                    while True:
                        batch = f.read(self.config.chunksize)
                        if not batch:
                            break
                        batch += f.readline()  # finish current line
                        if self.config.custom_newline is None: 
                            batch = batch.splitlines(keepends=self.config.keep_linebreaks)
                        else:
                            batch = batch.split(self.config.custom_newline)[:-1]
                        pa_table = pa.Table.from_arrays([pa.array(batch)], schema=schema)
                        # Uncomment for debugging (will print the Arrow table size and elements)
                        # logger.warning(f"pa_table: {pa_table} num rows: {pa_table.num_rows}")
                        # logger.warning('\n'.join(str(pa_table.slice(i, 1).to_pydict()) for i in range(pa_table.num_rows)))
                        yield (file_idx, batch_idx), pa_table
                        batch_idx += 1
```

**A clear and concise description of what you want to happen.**
Creating the dataset rows with a subset of the `splitlines()` line boundaries.

cronoik:
Gently pinging @lhoestq

lhoestq:
Hi ! Interresting :)

Could you give more details on what kind of separators you would like to use instead ?

cronoik:
In my case, I just want to use `\n` but not `U+2028`.

lhoestq:
Ok I see, maybe there can be a `sep` parameter to allow users to specify what line/paragraph separator they'd like to use

albertvillanova:
Related to:
- #3729

albertvillanova:
Thanks for requesting this enhancement. We have recently found a somehow related issue with another dataset:
- #3704

Let me make a PR proposal.

PR #3910, opened 2022-03-14
Fix text loader to split only on universal newlines

### LINKED ISSUE #3729: Wrong number of examples when loading a text dataset
## Describe the bug
when I use load_dataset to read a txt file  I find  that the number of the samples is incorrect

## Steps to reproduce the bug
```
fr = open('train.txt','r',encoding='utf-8').readlines()
print(len(fr))  # 1199637

datasets = load_dataset('text', data_files={'train': ['train.txt']}, streaming=False)
print(len(datasets['train']))  # 1199649
```
I also use command line operation to verify it
```
$ wc -l train.txt 
1199637 train.txt
```

## Expected results
please fix that issue 

## Environment info

- `datasets` version: 1.8.3
- Platform:windows&linux
- Python version:3.7
- PyArrow version:6.0.1

albertvillanova:
Currently, `text` loader breaks on a superset of universal newlines, which also contains Unicode line boundaries. See: https://docs.python.org/3/library/stdtypes.html#str.splitlines

However, the expected behavior is to get the lines splitted only on universal newlines: "\n", "\r\n" and "\r".

See: oscar-corpus/corpus#18

Fix #3729.

### DIFF
--- a/src/datasets/packaged_modules/text/text.py
+++ b/src/datasets/packaged_modules/text/text.py
@@ -1,4 +1,5 @@
 from dataclasses import dataclass
+from io import StringIO
 from typing import Optional
 
 import pyarrow as pa
@@ -52,7 +53,8 @@ class Text(datasets.ArrowBasedBuilder):
     def _generate_tables(self, files):
         schema = pa.schema(self.config.features.type if self.config.features is not None else {"text": pa.string()})
         for file_idx, file in enumerate(files):
-            with open(file, "rb") as f:
+            # open in text mode, by default translates universal newlines ("\n", "\r\n" and "\r") into "\n"
+            with open(file, encoding=self.config.encoding) as f:
                 if self.config.sample_by == "line":
                     batch_idx = 0
                     while True:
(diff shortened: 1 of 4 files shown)


lhoestq (approved):
Thanks for the fix ! This seems related to https://github.com/huggingface/datasets/issues/3804 as well

lhoestq:
Looks like the test needs to be updated for windows ^^'

adrianeboyd:
I don't think this is the same issue as in https://github.com/oscar-corpus/corpus/issues/18, where the OSCAR metadata has line offsets that use only `\n` as the newline marker to count lines, not `\r\n` or `\r`.

It looks like the OSCAR data loader is opening the data files with `gzip.open` directly and I don't think this text loader is used, but I'm not familiar with a lot of `datasets` internals so I could be mistaken?

albertvillanova:
@lhoestq I'm working on fixing the Windows tests on my Windows machine...

albertvillanova:
I finally changed the approach in order to avoid having "\r\n" and "\r" line breaks in Python `str` read from files on Windows/old Macintosh machines.

PR #3912, opened 2022-03-14
add draft of registering function for pandas

lvwerra:
This PR adds a register function for `pandas`. It allows to directly push `DataFrame` objects to the hub and in return loading datasets on the hub from `DataFrame`. The motivation for this integration is to enable the vast number of `pandas` users to be able to easily push `DataFrames` to the hub. 

Here is an example:

```python
import pandas as pd
from datasets import register_pandas

register_pandas()

# push to hub
df = pd.DataFrame.from_dict({"test": [1,2,3]})
df.push_to_hub("my_test")

# load from hub
df_retrieved = pd.DataFrame.load_from_hub("lvwerra/my_test")
```

It follows a similar philosophy as the `tqdm` [integration](https://github.com/tqdm/tqdm#pandas-integration). Also see [this issue](https://github.com/pandas-dev/pandas/issues/46000) on the `pandas` repository.

This is just a rough draft of what such integration could look like but I would like appreciate some feedback on this: is this something you would like to add the library and is this the way to go? cc @lhoestq @albertvillanova @julien-c

### DIFF
--- a/src/datasets/utils/pandas_to_hub.py
+++ b/src/datasets/utils/pandas_to_hub.py
@@ -1,0 +1,43 @@
+from datasets import Dataset, load_dataset
+
+def register_pandas():
+    """
+    Enables `pandas.DataFrame` objects to directly push to the hub or loaded from the hub.
+    
+    Example::
+
+        >>> df = pd.DataFrame.from_dict({"test": [1,2,3]})
+        >>> df.push_to_hub("my_dataset")
+
+        >>> loaded_df = df.load_from_hub("my_user/my_dataset")
+    """
+    from pandas.core.frame import DataFrame
+    
+    def push_to_hub(dataframe: DataFrame, repo_id: str, **push_to_hub_kwargs):
+        """Pushes the dataset to the hub.
+        The dataset is pushed using HTTP requests and does not need to have neither git or git-lfs installed.
+
+        Args:
+            repo_id (:obj:`str`):
+                The ID of the repository to push to in the following format: `<user>/<dataset_name>` or
+                `<org>/<dataset_name>`. Also accepts `<dataset_name>`, which will default to the namespace
+                of the logged-in user.
+        """
+        ds = Dataset.from_pandas(dataframe)
+        ds.push_to_hub(repo_id, **push_to_hub_kwargs)
+        
+    def load_from_hub(repo_id, split="train", **loading_kwargs):
+        """Loads a dataset from the hub into a DataFrame.
+        Args:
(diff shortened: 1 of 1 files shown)


lhoestq:
That's cool ! Though I would expect such an integration to only require `huggingface_hub`, not the full `datasets` library. 
 Indeed if users want to use the `datasets` lib they could just to `Dataset.from_pandas(df).push_to_hub()` already. Therefore I would explore something that doesn't not necessarily requires `datasets`.

For other could storage solutions (S3, GCS, etc.), pandas allows users to pass URIs like `s3://bucket-name/path/data.csv` to the `read_xxx` and `to_xxx` (for csv, parquet, json, etc). It also support passing the **root directory** like `s3://bucket-name/dataset-dir` instead of a single file name.

In the Hugging Face Hub case, we have one dataset = one repository. We can enter pandas' paradigm by saying one dataset = one repository = one root directory. Here is what we could have:

### push to Hub:
```python
"""
Demo script for writing a pandas data frame to a CSV file on HF using fsspec-supported pandas APIs
"""
import pandas as pd

HF_USER = os.getenv("HF_USER")
HF_TOKEN = os.getenv("HF_TOKEN")

books_df = pd.DataFrame(
    data={"Title": ["Book I", "Book II", "Book III"], "Price": [56.6, 59.87, 74.54]},
    columns=["Title", "Price"],
)

dataset_name = "books1"

books_df.to_csv(
    f"hf://{HF_USER}/{dataset_name}",
    index=False,
    storage_options={
        "repo_type": "dataset",
        "token": HF_TOKEN,
    },
)

```

### load from Hub:
```python
"""
Demo script for reading a CSV file from HF into a pandas data frame using fsspec-supported pandas
APIs
"""
import pandas as pd

HF_USER = os.getenv("HF_USER")
HF_TOKEN = os.getenv("HF_TOKEN")

dataset_name = "books1"

books_df = pd.read_csv(
    f"hf://{HF_USER}/{dataset_name}",
    storage_options={
        "repo_type": "dataset",
        "token": HF_TOKEN,
    },
)

print(books_df)
```

And you could do the same with Parquet data using `read/to_parquet` or other formats. Formats like CSV, Parquet or JSON Lines would work out of the box with `datasets`. This API would also allow anyone to use Dask with the Hugging Face Hub for example.

What do you think ?

mariosasko:
I'm closing this issue [`hffs`](https://github.com/huggingface/hffs) can now be used for reading/writing data frames from/to the Hub.

PR #3914, opened 2022-03-14
Use templates for doc-builidng jobs

sgugger:
This PR updates the jobs for all doc-building related things by using the templates introduced on `doc-builder`. By putting those once there, we make sure every repo gets the latest fixes on the doc-building github actions :-)

Note: all libraries must share the same docker image for those doc-building jobs. For now, all the one used (`huggingface/transformers-doc-builder`) contains all extra steps of the datasets install for docbuling (mainly libsndfile) but if in the future some additional steps are necessary on top of `pip install -e .[dev]`, this docker image will need to be updated with the extra deps.


lhoestq (approved):
Cool ! Would love to see some docs on how to add new dependencies when it's ready

thanks :)

lhoestq:
You can ignore the CI failures btw, they're unrelated to this PR

PR #3916, opened 2022-03-14
Create README.md for GLUE

sashavor:
I still have a hesitation regarding the format of inputs -- whether it's a list or a list of lists? -- hopefully @lhoestq will be able to clarify.

Also tagging @yjernite for the Limitations section. Happy to hear your thoughts!

### DIFF
--- a/metrics/glue/README.md
+++ b/metrics/glue/README.md
@@ -3,7 +3,7 @@
 ## Metric description
 This metric is used to compute the GLUE evaluation metric associated to each [GLUE dataset](https://huggingface.co/datasets/glue). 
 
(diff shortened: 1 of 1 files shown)


metrics/glue/README.md
```diff
@@ -0,0 +1,104 @@
...
+```python
+from datasets import load_metric
+glue_metric = load_metric('glue', 'sst2')
+references = [0, 1]
+predictions = [0, 1]
+results = glue_metric.compute(predictions=predictions, references=references)
+```
+## Output values
+
+The output of the metric depends on the GLUE subset chosen, consisting of a dictionary that contains one or several of the following metrics:
```
lhoestq, line 27: Maybe we can mention which subset uses which metric here ?

lhoestq (approved):
Thanks ! And the inputs are supposed to be lists, not lists of lists (there's a mistake in the docstring of the metric)

PR #3925, opened 2022-03-15
Fix main_classes docs index

lhoestq:
Currently the `main_classes` documentation has a wrong index. I believe this comes from issues in the examples of the Translation feature types

image


lhoestq:
Hmm it's still not good 
image

any idea what could cause this ?

lhoestq:
Ok fixed :)

PR #3926, opened 2022-03-15
Doc maintenance

stevhliu:
This PR adds some minor maintenance to the docs. The main fix is properly linking to pages in the callouts because some of the links would just redirect to a non-existent section on the same page.


lhoestq (approved):
Thanks you ! Good job figuring out the broken links

PR #3931, opened 2022-03-15
Add align_labels_with_mapping docs

stevhliu:
This PR documents the `align_labels_with_mapping` function to ensure predicted labels are aligned with the dataset, or to assign a different mapping of labels to ids (requested by @mariosasko 🎉 ).

For this specific code sample, the current dataset has a `mixed` label that the original [dataset](https://huggingface.co/datasets/poem_sentiment#data-fields) didn't. Is there a way to remove this label so it is completely aligned with the original dataset mappings? Otherwise, I'll just leave it as it is.

### DIFF
--- a/docs/source/process.mdx
+++ b/docs/source/process.mdx
@@ -246,39 +246,28 @@ Notice how the subfields are now their own independent columns: `answers.text` a
 
 ## Align
 
-The [`~datasets.Dataset.align_labels_with_mapping`] function is used to align a dataset label id with the label name, or you can also assign a different mapping of label to ids. 
+The [`~datasets.Dataset.align_labels_with_mapping`] function aligns a dataset label id with the label name. Not all 🤗 Transformers models follow the prescribed label mapping of the original dataset, especially for NLI dataset labels. For example, the [MNLI](https://huggingface.co/datasets/glue) dataset uses the following label mapping:
 
-For example, the [Poem Sentiment](https://huggingface.co/datasets/poem_sentiment) dataset uses the following label mapping:
+```py
+>>> label2id = {'entailment': 0, 'neutral': 1, 'contradiction': 2}
+```
+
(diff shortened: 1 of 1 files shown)


docs/source/process.mdx
```diff
@@ -244,6 +244,44 @@ Dataset({
...
+
+```py
+>>> from datasets import load_dataset
+
+>>> poem = load_dataset("poem_sentiment")
+>>> labels = poem["train"].features["label"].names
+>>> label2id, id2label = dict(), dict()
+>>> for i, label in enumerate(labels):
+...     label2id[label] = str(i)
+...     id2label[str(i)] = label
```
mariosasko, line 261: IMO this is cleaner:
```python
id2label = {i, label for i, label in enumerate(labels)}
label2id = {label, i for i, label in id2label}
```

stevhliu: Looked at the original paper, and it seems like they just dropped all samples with the `mixed` label from the final version of the dataset. I made a note of that, and added that we won't do any additional preprocessing to also drop the `mixed` label!

mariosasko: Thanks for the investigation!

mariosasko (approved):
LGTM! 

Gently pinging @lhoestq for the final review

lhoestq (commented):
Thanks a lot ! I think one very common case is people aligning their NLI datasets labels, because in `transformers` not all models follow the id2label `["entailment", "neutral", "contradiction"]` in this order

For example the most downloaded model for NLI is `facebook/bart-large-mnli` (1M4 downloads/month) and has
```
  "label2id": {
    "contradiction": 0,
    "entailment": 2,
    "neutral": 1
  },
```

If you think it can help, we can explain this specific use case as example here

lhoestq (approved):
Thanks ! LGTM :)

PR #3932, opened 2022-03-15
Create SARI metric card

sashavor:
SARI metric card! (do we have an expert in text simplification to validate?.. :sweat_smile: )

### DIFF
--- a/metrics/sari/README.md
+++ b/metrics/sari/README.md
@@ -17,7 +17,8 @@ where
 `F1_keep` is the n-gram F1 score for keep operations 
 
 `P_del` is the n-gram precision score for delete operations
-n = 4, as in the original paper. 
+
+The number of n grams, `n`, is equal to 4, as in the original paper. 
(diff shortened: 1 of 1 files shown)


metrics/sari/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+This metric outputs a dictionary with the SARI score:
+
+```
+print(sari_score)
+{'sari': 26.953601953601954}
+```
+
+The range of values for the SARI score is between 0 and 100 -- the higher the value, the better the performance of the model being evaluated, with a SARI of 
+100 being a perfect score.
```
albertvillanova, line 53: Nit:
```suggestion
The range of values for the SARI score is between 0 and 100 -- the higher the value, the better the performance of the model being evaluated, with a SARI of 100 being a perfect score.
```

albertvillanova (approved):
Thanks.

PR #3943, opened 2022-03-16
[Doc] Don't use v for version tags on GitHub

sgugger:
This removes the `v` automatically used by `doc-builder` for versions.


mishig25 (approved):
thanks for this!

PR #3963, opened 2022-03-18
Add Audio Folder

### LINKED ISSUE #3964: Add default Audio Loader
Writing a custom loading dataset script might be a bit challenging for users.

Add default Audio loader (analogous to ImageFolder) for small datasets with standard directory structure.

Create a custom loading script? that's what users doing now.

polinaeterna:
Would resolve  #3964

AudioFolder loads a .txt file with transcriptions and creates a dataset with all audiofiles in provided directory that has a transcription (independently of the directory structure) as a single split (train).

Can be loaded via:
```python
# for local dirs
dataset = load_dataset("audiofolder", data_dir="/path/to/folder", transcripts_filename="transcripts.txt")
```
```python
# for local and remote zip archives
dataset = load_dataset("audiofolder", data_files="path/to/archive/archive.zip", transcripts_filename="transcripts.txt")
```
default transcriptions filename is `transcripts.txt`. it should have the following structure:
```
audio_id_1   transcription text 1
audio_id_1   transcription text 1
```
separator is `\t`!

---
sorry for first old commits from other branch, don't know how that happened...


lhoestq:
Feel free to merge `master` into this branch to fix the CI errors related to Google Drive :)

I think we can just remove the test that is based on dummy data, or make it have the `sampling_rate` parameter hardcoded in the test

mariosasko:
IMO it's important to keep this loader aligned with `imagefolder`. I'm aware that the current `imagefolder` API is limiting because only labels can be inferred from the directory structure, which means it can only be used for classification and self-supervised pretraining. However, to make the loader more generic, we plan to support [metadata files](https://huggingface.slack.com/archives/C02JB9L6JKF/p1645450017434029?thread_ts=1645157416.389499&cid=C02JB9L6JKF) (will work on that this week),  and in the audio case, these files can store transcripts.

Streaming TAR archives (`iter_archive`) is not supported by any of the loaders currently, so we can add that in a separate PR for all of them (to keep this PR simple).

WDYT?

lhoestq:
> Streaming TAR archives (iter_archive) is not supported by any of the loaders currently, so we can add that in a separate PR for all of them (to keep this PR simple).

Yes definitely, we can see that later

> to make the loader more generic, we plan to support [metadata files](https://huggingface.slack.com/archives/C02JB9L6JKF/p1645450017434029?thread_ts=1645157416.389499&cid=C02JB9L6JKF) (will work on that this week), and in the audio case, these files can store transcripts.

Could you share an example of what the structure would look like in this case ?

Note that for audio we ultimately should be able to load several splits at once (common voice, librispeech, etc. all have splits), unlike the current imagefolder implementation that puts everything in `train`. If we want consistency then we would need the same for imagefolder

tests/test_packaged_modules.py
```diff
@@ -92,3 +106,16 @@ def test_imagefolder_drop_labels(image_file, drop_labels):
         assert all(example.keys() == {"image", "label"} for _, example in generator)
     else:
         assert all(example.keys() == {"image"} for _, example in generator)
+
+
+@pytest.mark.parametrize("sampling_rate", [None, 16_000])
```
polinaeterna, line 111: this test is logically a bit strange as it checks two things: that you receive an error if `sampling_rate` is not provided and that a dataset instantiates correctly when `sampling_rate` is provided + **also** that it correctly generates examples. Is it reasonable to move correct generation checking to a separate test or is it ok as it is?

polinaeterna: sorry for publishing the comment now, i accidentally pressed the wrong button...

polinaeterna:
> I think we can just remove the test that is based on dummy data, or make it have the sampling_rate parameter hardcoded in the test

Not sure what to do with `test_builder_class` and `test_load_dataset_offline`, I don't really want to drop these tests completely but do you think it's a good idea to hardcode builder loading like this: 🤔
```
if dataset_name == "audiofolder":
    builder = builder_cls(name=name, cache_dir=tmp_cache_dir, sampling_rate=16_000)
else:
    builder = builder_cls(name=name, cache_dir=tmp_cache_dir)
```
@mariosasko totally agree on that APIs should be aligned, do you think we should implement metadata support first? Or maybe we can merge this PR with explicit single transcript file and add full metadata support further.

Splits support is definitely a required feature too, I think we can implement it in the future PR too.

polinaeterna:
btw i've found a workaround for splits generation :D

```
from datasets.data_files import DataFilesDict

ds = load_dataset(
    "audiofolder",
    data_files=DataFilesDict(
        {
            "train":"../audiofolder/AudioTestSplits/train.zip",
            "test": "../audiofolder/AudioTestSplits/test.zip"
        }
    ),
    sampling_rate=16_000
)
```

lhoestq:
> Not sure what to do with test_builder_class and test_load_dataset_offline, I don't really want to drop these tests completely but do you think it's a good idea to hardcode builder loading like this: 🤔

Yes it's fine. If you you're not a fan of having such parameters directly at the core of the code you can declare a global variable `PACKAGED_MODULES_TEST_KWARGS = {"audiofolder": {"sampling_rate": 16_000}}` and do
```python
builder_kwargs = PACKAGED_MODULES_TEST_KWARGS.get(name, {})
builder = builder_cls(name=name, cache_dir=tmp_cache_dir, **builder_kwargs)
```

polinaeterna:
@lhoestq @mariosasko please take a look at the code and feel free to add your comments and discuss the potential issues
 
after we are satisfied with the code, I'll write the documentation

lhoestq (commented):
The code sounds good to me :)

I created https://github.com/huggingface/datasets/pull/4006 to update the ASR task template

polinaeterna:
@lhoestq it appeared that this PR already exists https://github.com/huggingface/datasets/pull/3364

src/datasets/packaged_modules/audiofolder/audiofolder.py
```diff
@@ -0,0 +1,161 @@
...
+    def _info(self):
+        return datasets.DatasetInfo(
+            features=datasets.Features(
+                {
+                    "path": datasets.Value("string"),
+                    "audio": datasets.Audio(sampling_rate=self.config.sampling_rate),
+                    "text": datasets.Value("string"),
+                }
+            ),
+            task_templates=[AutomaticSpeechRecognition(audio_file_path_column="path", transcription_column="text")],
```
lhoestq, line 39: you can update things here:
```suggestion
            task_templates=[AutomaticSpeechRecognition(audio_column="audio", transcription_column="text")],
```

src/datasets/packaged_modules/audiofolder/audiofolder.py
```diff
@@ -0,0 +1,161 @@
...
+class AudioFolder(datasets.GeneratorBasedBuilder):
+    BUILDER_CONFIG_CLASS = AudioFolderConfig
+
+    AUDIO_EXTENSIONS: List[str] = []  # definition at the bottom of the script
+
+    def _info(self):
+        return datasets.DatasetInfo(
+            features=datasets.Features(
+                {
+                    "path": datasets.Value("string"),
```
lhoestq, line 34: Should we keep the "path" field ? We don't have it for imagefolder.

It's fine if we don't have it and if later we add a parameter for users that would like to have it.

lhoestq (commented):
I merged the ASR task template PR :) following this, here are two comments/suggestions:

docs/source/audio_process.mdx
```diff
@@ -63,6 +63,53 @@ The `path` is useful if you want to load your own audio dataset. In this case, p
...
+```
+
+Load remote datasets from their URLs with the `data_files` parameter:
+
+```py
+>>> dataset = load_dataset("audiofolder", data_files="https://huggingface.co/datasets/polinaeterna/test_audiofolder/resolve/main/MLSPolishSample.zip", sampling_rate=16_000)
+```
+
+Note that `sampling_rate` is a required argument.
+Only `.zip` archives are supported now.
```
lhoestq, line 110:
```suggestion
Audio files can also be in `.zip` archives.
```

lhoestq (approved):
Thanks ! It looks all good to me :)

I'd let @mariosasko give his review as well

docs/source/audio_process.mdx
polinaeterna: Actually I meant that for example tar archives are not supported yet. Removed this line to avoid confusion :)

mariosasko (commented):
As I said in the previous comment, I firmly believe this loader should be aligned with the Image Folder to allow our users to easily switch modalities, from audio to image or the other way around.

The current problem with this loader is that it supports the ASR task by default, which could be surprising for the users thinking that this is the Image Folder counterpart for audio. To avoid this, we should support the audio classification task by default instead (we can add a template for it in this PR), where the `label` column is inferred from the directory structure. A quick search on the internet gives a number of tutorials where an audio dataset follows that structure. And for the ASR task, we can use the metadata file support from https://github.com/huggingface/datasets/pull/4069. 

So in this PR, IMO, it makes the most sense to add a new AudioClassification task template, copy&paste the Image Folder code, rename `Image`/`image` to `Audio`/`audio` and change the file extensions list.

lhoestq:
> The current problem with this loader is that it supports the ASR task by default, which could be surprising for the users thinking that this is the Image Folder counterpart for audio. To avoid this, we should support the audio classification task by default instead (we can add a template for it in this PR), where the label column is inferred from the directory structure.

Right indeed, good catch. It's better to keep polishing the API rather than pushing fast something that can be confusing for users. Let's go for maximum alignment between the two then @polinaeterna ?

polinaeterna:
@mariosasko sorry, I didn't understand from your previous message that by aligning with the ImageFolder you mean inferring labels from directories names. Sure, that's not a problem, I can add the corresponding code. Do you also mean that in this version we should get rid of transcription file and feature and add it in the future when the metadata support https://github.com/huggingface/datasets/pull/4069 will be merged?  
My understanding was that support for ASR task is more crucial than audio classification as it's more "common", but I would ask @anton-l and @patrickvonplaten about this. Anyway, it's not a problem to implement the classification task first, and the ASR one later.

src/datasets/packaged_modules/audiofolder/audiofolder.py
```diff
@@ -0,0 +1,161 @@
...
+        if not sampling_rate:
+            raise ValueError(
+                "To load an audio dataset, you must provide the `sampling_rate` parameter. "
+                "For example: "
+                '>>> load_dataset("audiofolder", data_dir="path/to/dir", sampling_rate=16_000)'
+            )
+        self.sampling_rate = sampling_rate
+        self.transcripts_filename = transcripts_filename
+
+        super(AudioFolderConfig, self).__init__(*args, **kwargs)
```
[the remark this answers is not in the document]
polinaeterna: My motivation was that we don't want users to accidentally mess up with sampling rate but your argument that it's not a required param in the Audio feature (so that way it's not consistent) seems valid to me.

mariosasko:
> Do you also mean that in this version we should get rid of transcription file and feature and add it in the future when the metadata support https://github.com/huggingface/datasets/pull/4069 will be merged?

We can wait for the linked PR to be merged first and then add the changes to this PR to have support for ASR from the get-go.

patrickvonplaten:
Don't follow 100% here, but as @polinaeterna said I think ASR is much more common than audio classification. Also, do you guys think a lot of users will use both the audio and image folder functionality ? Is it very important to have audio and image aligned here? Note that in Transformers while all models follow a common API, audio and vision models can be very different with respect to pre- and post-processing

lhoestq:
> I think ASR is much more common than audio classification

I agree, the main focus is ASR

> do you guys think a lot of users will use both the audio and image folder functionality ?

Yup I think so, people don't just use public academic datasets right ? `imagefolder` is almost used 1k times a week, and it's just the beginning.

> Is it very important to have audio and image aligned here?

If we can get some consistency for free, let's take it ^^ This way it will be easy for users to go from one modality to another, and documentation will be simpler.

> Note that in Transformers while all models follow a common API, audio and vision models can be very different with respect to pre- and post-processing

That make total sense. Here this is mainly about raw data loading (before preprocessing) so we just need to make something generic, no matter what task the data is used for. Even though actually we know that ASR will be the main usage for now :p

Let me know if it's clearer now or if you have other questions !

PR #3967, opened 2022-03-18
[feat] Add TextVQA dataset

apsdehal:
This would be the first classification-based vision-and-language dataset in the datasets library.

Currently, the dataset downloads everything you need beforehand. See the [paper](https://arxiv.org/abs/1904.08920) for more details.

Test Plan:
- Ran the full and the dummy data test locally


lhoestq:
Hey :) Have you had a chance to continue this PR ? Let me know if you have questions or if I can help

apsdehal:
Hey @lhoestq, let me wrap this up soon. I will resolve your comments in next PR.

src/datasets/utils/resources/tasks.json
```diff
@@ -77,7 +77,8 @@
         "subtasks": [
             "extractive-qa",
             "open-domain-qa",
-            "closed-domain-qa"
+            "closed-domain-qa",
+            "visual-question-answering",
```
lhoestq, line 81:
```suggestion
            "visual-question-answering"
```

lhoestq: To add this task on the Hub, please open a PR to update https://github.com/huggingface/hub-docs/blob/main/js/src/lib/interfaces/Types.ts first

datasets/textvqa/README.md
```diff
@@ -0,0 +1,203 @@
...
+- `image_width`: int, Width of the original image.
+- `image_height`: int, Height of the original image.
+- `question_tokens`:  List[str], A pre-tokenized list of question.
+- `answers`: List[str], List of 10 human-annotated answers for the question. These 10 answers are collected from 10 different users.
+- `question_id`: int, Unique id of the question.
+- `set_name`: string, the set to which this question belongs.
+
+### Data Splits
+
+There are three splits. `train`, `validation` and `test`. The `train` and `validation` sets share images with OpenImages `train` set and have their answers available. To get inference results and numbers on `test` set, you need to go to the [EvalAI leaderboard](https://eval.ai/web/challenges/challenge-page/874/overview) and upload your predictions there. Please see instructions at [https://textvqa.org/challenge/](https://textvqa.org/challenge/).
```
lhoestq, line 122: Does the test set have include answers as well ? This line in the code makes me think it may not have the answers:
```python
item["answers"] = item.get("answers", ["" for _ in range(_NUM_ANSWERS_PER_QUESTION)])
```

datasets/textvqa/README.md
```diff
@@ -0,0 +1,203 @@
...
+- cc-by-4.0
+multilinguality:
+- monolingual
+pretty_name: TextVQA
+size_categories:
+- 10K<n<100K
+source_datasets:
+- original
+task_categories:
+- question-answering
```
lhoestq, line 18: `question-answering` means extractive-qa for text in `transformers` and on the Hub. In particular each task category refers to a pipeline type in `transformers`.

Therefore I would create a new task category for `visual-question-answering` instead of having it under `question-answering`

Feel free to open a PR to add it to https://github.com/huggingface/hub-docs/blob/main/js/src/lib/interfaces/Types.ts :) cc @osanseviero would it make sense in your opinion ?

lhoestq (commented):
Cool Thanks ! more comments, especially one about the task category tag

datasets/textvqa/README.md
apsdehal: I just noticed that `visual-question-answering` is already there: https://github.com/huggingface/hub-docs/blob/f9e8ec8f882ddb2b0517e97afc998f9cf398953b/js/src/lib/interfaces/Types.ts#L549

datasets/textvqa/README.md
apsdehal: It is return -1 when the answer is missing.

apsdehal: Do you think we should return an empty string. That would be somewhat unintuitive. -1 clearly suggests that the answers are not present. Maybe I should mention this in the README explicitly.

apsdehal: Yes, the test set doesn't have answers. So we are returning the empty strings. Maybe we should mention this.

lhoestq (approved):
Thanks for fixing the task tag and for the comments about the missing answers !

LGTM :)

PR #3843, opened 2022-03-07
Fix Google Drive URL to avoid Virus scan warning in streaming mode

mariosasko:
The streaming version of https://github.com/huggingface/datasets/pull/3787.

Fix #3835 

CC: @albertvillanova

PR #3787, opened 2022-02-25
Fix Google Drive URL to avoid Virus scan warning

albertvillanova:
This PR fixes, in the datasets library instead of in every specific dataset, the issue of downloading the Virus scan warning page instead of the actual data file for Google Drive URLs.

Fix #3786.

PR #3979, opened 2022-03-21
Fix google drive streaming for small files

lhoestq:
Google drive did another change recently, following #3787 #3843 .

In particular Google Drive now returns 403 for GET requests with `confirm=t` when a files doesn't have a virus warning message. I fixed this by passing `confirm=t` if and only if when there is one (i.e. when status code is 200 for HEAD)


severo:
this fix will not be included?

lhoestq:
No we can't do anything except stop using google drive when possible

Issue #3960, opened 2022-03-18
Load local dataset error


TXacs:
When i used the datasets==1.11.0， it's all right. Util update the latest version, it get the error like this:
```
>>> from datasets import load_dataset
>>> data_files={'train': ['/ssd/datasets/imagenet/pytorch/train'], 'validation': ['/ssd/datasets/imagenet/pytorch/val']}
>>> ds = load_dataset('nateraw/image-folder', data_files=data_files, cache_dir='./', task='image-classification')
[] https://huggingface.co/datasets/nateraw/image-folder/resolve/main/ /dat/txacs/git/txacs/examples/image-classification/https:/huggingface.co/datasets/nateraw/image-folder/resolve/main
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/gf3/home/txacs/gv3/anaconda3/envs/txacs/lib/python3.6/site-packages/datasets/load.py", line 1671, in load_dataset
    **config_kwargs,
  File "/gf3/home/txacs/gv3/anaconda3/envs/txacs/lib/python3.6/site-packages/datasets/load.py", line 1521, in load_dataset_builder
    **config_kwargs,
  File "/gf3/home/txacs/gv3/anaconda3/envs/txacs/lib/python3.6/site-packages/datasets/builder.py", line 1031, in __init__
    super().__init__(*args, **kwargs)
  File "/gf3/home/txacs/gv3/anaconda3/envs/txacs/lib/python3.6/site-packages/datasets/builder.py", line 255, in __init__
    sanitize_patterns(data_files), base_path=base_path, use_auth_token=use_auth_token
  File "/gf3/home/txacs/gv3/anaconda3/envs/txacs/lib/python3.6/site-packages/datasets/data_files.py", line 584, in from_local_or_remote
    if not isinstance(patterns_for_key, DataFilesList)
  File "/gf3/home/txacs/gv3/anaconda3/envs/txacs/lib/python3.6/site-packages/datasets/data_files.py", line 546, in from_local_or_remote
    data_files = resolve_patterns_locally_or_by_urls(base_path, patterns, allowed_extensions)
  File "/gf3/home/txacs/gv3/anaconda3/envs/txacs/lib/python3.6/site-packages/datasets/data_files.py", line 196, in resolve_patterns_locally_or_by_urls
    for path in _resolve_single_pattern_locally(base_path, pattern, allowed_extensions):
[...]

PR #3980, opened 2022-03-21
Add tip on how to speed up loading with ImageFolder

mariosasko:
This PR does two things:
* adds a tip on how to speed up loading of a large number of files with ImageFolder (motivated by [this issue](https://github.com/huggingface/datasets/issues/3960))
* replaces the current references to the `Dataset` methods in the Image Processing doc with their fully qualified counterparts (to align it with the Audio Processing doc)

cc @stevhliu

### DIFF
--- a/docs/source/image_process.mdx
+++ b/docs/source/image_process.mdx
@@ -1,11 +1,11 @@
 # Process image data
 
-🤗 Datasets support loading and processing images with the [`datasets.Image`] feature. This guide will show you how to:
+🤗 Datasets support loading and processing images with the [`Image`] feature. This guide will show you how to:
 
 - Load an image dataset.
 - Load a generic image dataset with `ImageFolder`.
-- Use [`datasets.Dataset.map`] to quickly apply transforms to an entire dataset.
-- Add data augmentations to your images with [`datasets.Dataset.set_transform`].
+- Use [`~Dataset.map`] to quickly apply transforms to an entire dataset.
+- Add data augmentations to your images with [`Dataset.set_transform`].
 
 ## Image datasets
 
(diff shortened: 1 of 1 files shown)


stevhliu:
Thanks for adding that tip! 👍 

For the docs syntax, it might be better if we hide the package name/full path to the class or function and only show the name of it. I think it's easier for users to read the function name (eg,`cast_column`) instead of the full path which can be a bit lengthy for some functions like `datasets.IterableDataset.remove_columns` (and if we like this idea, we can align the rest of the docs on it).

mariosasko:
> For the docs syntax, it might be better if we hide the package name/full path to the class or function and only show the name of it. I think it's easier for users to read the function name (eg,cast_column) instead of the full path which can be a bit lengthy for some functions like datasets.IterableDataset.remove_columns (and if we like this idea, we can align the rest of the docs on it).

That's also OK, as long as we are consistent.

@lhoestq @albertvillanova @polinaeterna Which one of these two styles do you prefer?

lhoestq (approved):
Sounds good ! We should at least hide the package name IMO. And we could keep class name + method name to avoid ambiguities. For example for streaming datasets, I think it's better to write explicitly `IterableDataset.cast_column` and not just `cast_column`, since `Dataset.cast_column` also exists.

polinaeterna:
Agree on hiding `datasets` name. Not sure about hiding class name as it's anyway not visible for users if they use `Dataset.cast_column` or  `IterableDataset.cast_column` when working with their datasets. But I agree that the most important thing is to be consistent :)

stevhliu:
Good points! :)

I think it'll be good to show the class name since some functions have different parameters. For example, if users click on `IterableDataset.map` and then `Dataset.map`, they'll see different parameters and have to figure out why (which isn't too difficult I guess lol). But showing the class name avoids any confusion upfront.

PR #3981, opened 2022-03-21
Add TER metric card

emibaylor:
Add TER metric card

This card is still missing content for the following sections:
- **Limitations & Biases**
- **Values from Papers**
If anyone has any ideas for either of the above, feel free to either add them or point me to them and I'll add them!

### DIFF
--- a/metrics/ter/ter.py
+++ b/metrics/ter/ter.py
@@ -184,7 +184,12 @@ class Ter(datasets.Metric):
             raise ValueError("Sacrebleu requires the same number of references for each prediction")
         transformed_references = [[refs[i] for refs in references] for i in range(references_per_prediction)]
 
-        sb_ter = TER(normalized=normalized, no_punct=ignore_punct, asian_support=support_zh_ja_chars, case_sensitive=case_sensitive)
+        sb_ter = TER(
+            normalized=normalized,
+            no_punct=ignore_punct,
+            asian_support=support_zh_ja_chars,
+            case_sensitive=case_sensitive,
+        )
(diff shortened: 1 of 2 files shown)


metrics/ter/ter.py
```diff
@@ -118,16 +174,22 @@ def _compute(
         predictions,
         references,
         normalized: bool = False,
-        no_punct: bool = False,
-        asian_support: bool = False,
-        case_sensitive: bool = False,
+        ignore_punct: bool = False,
+        support_zh_ja_chars: bool = False,
+        ignore_case: bool = False,
```
lhoestq, line 179: Is it fine to have `ignore_case` to `False` here ? It used to ignore casing by default

lhoestq (commented):
Nice thanks ! Just one question about `ignore_case`:

metrics/ter/ter.py
emibaylor: I changed it because it seemed more consistent both with other metrics and with other inputs in this metric (in that, by default, the inputs are compared directly with no change to their form) if `ignore_case` was `False` by default. That being said, I don't think it's a big deal to keep it the way it was either, with `ignore_case` as `True` by default. Do you have a preference either way?

lhoestq: We can keep the default defined by the original library, i.e. `case_sensitive=False`. It seems common to do so, e.g. other similar metrics like WER or CER are not case sensitive either.

lhoestq (approved):
Thanks !

PR #3982, opened 2022-03-21
Mark Google Drive tests as slow

lhoestq:
These tests make the CI spam the Google Drive API, the CI now gets banned by Google Drive very often.

I think we can just skip these tests from the CI for now.

In the future we could have a CI job that runs only once a day or once a week for such cases

cc @albertvillanova @mariosasko @severo 

image

### DIFF
--- a/tests/test_streaming_download_manager.py
+++ b/tests/test_streaming_download_manager.py
@@ -32,7 +32,7 @@ from datasets.utils.streaming_download_manager import (
     xsplitext,
 )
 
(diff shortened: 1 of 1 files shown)


albertvillanova:
I was thinking exactly the same: running unit tests that request continuously a third-party API is not a good idea.

albertvillanova (approved):
Thanks. This PR gooes in the good direction for unit tests.

PR #3987, opened 2022-03-22
Fix Faiss custom_index device

albertvillanova:
Currently, if both `custom_index` and `device` are passed to `FaissIndex`, `device` is silently ignored.

This PR fixes this by raising a ValueError if both arguments are passed.

Alternatively, the `custom_index` could be transferred to the target `device`.


lhoestq (approved):
Good catch :)

PR #3989, opened 2022-03-22
Remove old wikipedia leftovers

albertvillanova:
After updating Wikipedia dataset, remove old wikipedia leftovers from doc.

### DIFF
--- a/docs/source/about_arrow.md
+++ b/docs/source/about_arrow.md
@@ -22,12 +22,12 @@ For example, loading the full English Wikipedia dataset only takes a few MB of R
 >>> from datasets import load_dataset
 
 # Process.memory_info is expressed in bytes, so convert to megabytes 
->>> mem_before = psutil.Process(os.getpid()).memory_info().rss  / (1024 * 1024)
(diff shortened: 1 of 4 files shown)


docs/source/about_arrow.md
```diff
@@ -22,12 +22,12 @@ For example, loading the full English Wikipedia dataset only takes a few MB of R
 >>> from datasets import load_dataset
 
 # Process.memory_info is expressed in bytes, so convert to megabytes 
->>> mem_before = psutil.Process(os.getpid()).memory_info().rss  / (1024 * 1024)
->>> wiki = load_dataset("wikipedia", "20200501.en", split='train')
+>>> mem_before = psutil.Process(os.getpid()).memory_info().rss >> 20
```
lhoestq, line 25: I think `/ (1024 * 1024)` makes more sense for non-python people :P

lhoestq (approved):
Nice thanks :)

This makes me think we shouldn't advise the use of `load_dataset` in dataset scripts, since it doesn't guarantee that the cache will work as expected (the cache directory is not set correctly, and the required disk space for downloaded files is not recorded)

PR #3995, opened 2022-03-23
Close `PIL.Image` file handler in `Image.decode_example`

### LINKED ISSUE #3985: [image feature] Too many files open error when image feature is returned as a path
## Describe the bug

PR in context: #3967. If I load the dataset in this PR (TextVQA), and do a simple list comprehension on the dataset, I get `Too many open files error`. This is happening due to the way we are loading the image feature when a str path is returned from the `_generate_examples`. Specifically at https://github.com/huggingface/datasets/blob/508eb4ab5d52f590baa677b4f64b1cc069139f7b/src/datasets/features/image.py#L110, we are open the file handle to the image but never closing it. This in my understanding is causing the issue.

## Steps to reproduce the bug
Pull the PR locally and run the following code
```python

dataset = load_dataset("./datasets/textvqa")["train"]
data = [item for item in dataset]

# Error happens
```

## Expected results

List comprehension should work smoothly

## Actual results

`Too many open files error`

## Environment info

- `datasets` version: 2.0.1.dev0
- Platform: macOS-12.2-arm64-arm-64bit
- Python version: 3.10.0
- PyArrow version: 7.0.0
- Pandas version: 1.4.1

mariosasko:
Closes the file handler of the PIL image object in `Image.decode_example` to avoid the `Too many open files` error.

To pass [the image equality checks](https://app.circleci.com/pipelines/github/huggingface/datasets/10774/workflows/d56670e6-16bb-4c64-b601-a152c5acf5ed/jobs/65825) in CI, `Image.decode_example` calls `image.load()` regardless of how the image object is created (not only for the `PIL.Image.open(local_path)` case). This is needed because `load()` sets the `readonly` attribute of a `PIL.Image` object to 0 (it's 1 after `PIL.Image.open(file_like)`), and in the older PIL versions (only fixed on main), that attribute is considered in `PIL.Image.__eq__`. More info can be found here: https://github.com/python-pillow/Pillow/issues/5926.

Fix #3985

### DIFF
--- a/src/datasets/features/image.py
+++ b/src/datasets/features/image.py
@@ -108,6 +108,7 @@ class Image:
             else:
                 if is_local_path(path):
                     image = PIL.Image.open(path)
+                    image.load()
                 else:
                     with xopen(path, "rb") as f:
                         bytes_ = BytesIO(f.read())
--- a/src/datasets/features/image.py
+++ b/src/datasets/features/image.py
@@ -108,13 +108,13 @@ class Image:
             else:
                 if is_local_path(path):
                     image = PIL.Image.open(path)
-                    image.load()
                 else:
                     with xopen(path, "rb") as f:
                         bytes_ = BytesIO(f.read())
                     image = PIL.Image.open(bytes_)
         else:
             image = PIL.Image.open(BytesIO(bytes_))
+        image.load()
         return image
 
     def flatten(self) -> Union["FeatureType", Dict[str, "FeatureType"]]:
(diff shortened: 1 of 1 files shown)


src/datasets/features/image.py
```diff
@@ -114,6 +114,7 @@ def decode_example(self, value: dict) -> "PIL.Image.Image":
                     image = PIL.Image.open(bytes_)
         else:
             image = PIL.Image.open(BytesIO(bytes_))
+        image.load()
```
lhoestq, line 117:
```suggestion
        image.load()  # to avoid "Too many open files" errors
```

lhoestq (approved):
thank you !

PR #3997, opened 2022-03-23
Sync Features dictionaries

mariosasko:
This PR adds a wrapper to the `Features` class to keep the secondary dict, `_column_requires_decoding`, aligned with the main dict (as discussed in https://github.com/huggingface/datasets/pull/3723#discussion_r806912731).

A more elegant approach would be to subclass `UserDict` and override `__setitem__` and `__delitem__`, but this PR doesn't implement it for the following reasons:
* it requires replacing all occurrences of `isinstance(obj, dict)` with `isinstance(obj, Mapping)`, which is five times slower than `isinstance(obj, dict)` on my machine, in `features.py`
* is a breaking change, i.e., `isinstance(Features(...), dict)` would return `False` after it
* IMO, it makes sense to be consistent in the user-facing API and subclass either `dict` or `UserDict`. The problem with the latter is that it can't be used for `DatasetDict` because `DatasetDict` exposes the `data` property, which is also used by `UserDict`, so this would result in a collision.


tests/features/test_features.py
```diff
@@ -233,6 +233,30 @@ def test_flatten_with_sequence(self):
...
+    def test_features_dicts_are_synced(self):
+        def assert_features_dicts_are_synced(features: Features):
+            assert (
+                hasattr(features, "_column_requires_decoding")
+                and features.keys() == features._column_requires_decoding.keys()
+            )
+
+        features = Features({"foo": Sequence({"bar": {"my_value": Value("int32")}})})
+        assert_features_dicts_are_synced(features)
+        features["barfoo"] = Value("binary")
```
lhoestq, line 245: maybe just add one line with an Image type just to be sure ?

lhoestq (approved):
Thanks ! That makes sense, and this solution is quite straightforward :)

PR #3999, opened 2022-03-23
Docs maintenance

stevhliu:
This PR links some functions to the API reference. These functions previously only showed up in code format because the path to the actual API was incorrect.

### DIFF
--- a/docs/source/faiss_es.mdx
+++ b/docs/source/faiss_es.mdx
@@ -32,7 +32,7 @@ FAISS retrieves documents based on the similarity of their vector representation
 >>> ds_with_embeddings.add_faiss_index(column='embeddings')
 ```
 
-4. Now you can query your dataset with the `embeddings` index. Load the DPR Question Encoder, and search for a question with [`Dataset.get_nearest_examples`]:
+4. Now you can query your dataset with the `embeddings` index. Load the DPR Question Encoder, and search for a question with [`~datasets.search.IndexableMixin.get_nearest_examples`]:
 
 ```py
 >>> from transformers import DPRQuestionEncoder, DPRQuestionEncoderTokenizer
@@ -46,13 +46,13 @@ FAISS retrieves documents based on the similarity of their vector representation
 '_that_ serious? It is not serious at all. It’s simply a fantasy to amuse\r\n'
 ```
 
-5. When you are done querying, save the index on disk with [`Dataset.save_faiss_index`]:
+5. When you are done querying, save the index on disk with [`~datasets.search.IndexableMixin.save_faiss_index`]:
 
 ```py
 >>> ds_with_embeddings.save_faiss_index('embeddings', 'my_index.faiss')
 ```
 
-6. Reload it at a later time with [`Dataset.load_faiss_index`]:
+6. Reload it at a later time with [`~datasets.search.IndexableMixin.load_faiss_index`]:
 
 ```py
 >>> ds = load_dataset('crime_and_punish', split='train[:100]')
@@ -78,7 +78,7 @@ Start ElasticSearch on your machine, or see the [ElasticSearch installation guid
 >>> squad.add_elasticsearch_index("context", host="localhost", port="9200")
 ```
 
-3. Then you can query the `context` index with [`Dataset.get_nearest_examples`]:
+3. Then you can query the `context` index with [`~datasets.search.IndexableMixin.get_nearest_examples`]:
 
 ```py
 >>> query = "machine"
@@ -97,7 +97,7 @@ Start ElasticSearch on your machine, or see the [ElasticSearch installation guid
 hf_squad_val_context
 ```
 
-5. Reload it later with the index name when you call [`Dataset.load_elasticsearch_index`]:
+5. Reload it later with the index name when you call [`~datasets.search.IndexableMixin.load_elasticsearch_index`]:
 
 ```py
 >>> from datasets import load_dataset
--- a/docs/source/stream.mdx
+++ b/docs/source/stream.mdx
@@ -115,9 +115,9 @@ The following methods allow you to modify the columns of a dataset. These method
 
 ### Rename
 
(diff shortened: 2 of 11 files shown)


docs/source/faiss_es.mdx
```diff
@@ -32,7 +32,7 @@ FAISS retrieves documents based on the similarity of their vector representation
 >>> ds_with_embeddings.add_faiss_index(column='embeddings')
 ```
 
-4. Now you can query your dataset with the `embeddings` index. Load the DPR Question Encoder, and search for a question with [`Dataset.get_nearest_examples`]:
+4. Now you can query your dataset with the `embeddings` index. Load the DPR Question Encoder, and search for a question with [`~datasets.search.IndexableMixin.get_nearest_examples`]:
```
lhoestq, line 35: I think you can point to the Dataset method here no ? Or the doc builder is not able to resolve it ?
Same for the other search and tf methods

```suggestion
4. Now you can query your dataset with the `embeddings` index. Load the DPR Question Encoder, and search for a question with [`~datasets.Dataset.get_nearest_examples`]:
```

docs/source/stream.mdx
```diff
@@ -137,7 +137,7 @@ When you need to remove one or more columns, give [`IterableDataset.remove_colum
 
 ### Cast
 
-[`IterableDataset.cast`] changes the feature type of one or more columns. This method takes your new `Features` as its argument. The following sample code shows how to change the feature types of `ClassLabel` and `Value`:
+[`IterableDataset.cast_column`] changes the feature type of one or more columns. This method takes your new `Features` as its argument. The following sample code shows how to change the feature types of `ClassLabel` and `Value`:
```
lhoestq, line 140: This part was correct: it explains `cast`, not `cast_column`

```suggestion
[`IterableDataset.cast`] changes the feature type of one or more columns. This method takes your new `Features` as its argument. The following sample code shows how to change the feature types of `ClassLabel` and `Value`:
```

docs/source/stream.mdx
```diff
@@ -151,7 +151,7 @@ When you need to remove one or more columns, give [`IterableDataset.remove_colum
 >>> new_features = dataset.features.copy()
 >>> new_features["label"] = ClassLabel(names=['negative', 'positive'])
 >>> new_features["idx"] = Value('int64')
->>> dataset = dataset.cast(new_features)
+>>> dataset = dataset.cast_column(new_features)
```
lhoestq, line 154:
```suggestion
>>> dataset = dataset.cast(new_features)
```

lhoestq (commented):
Thanks ! Some comments:

docs/source/stream.mdx
stevhliu: Ah, I see where I got confused! 😵‍💫  

`IterableDataset.cast` and `IterableDataset.rename_column` aren't shown in the current API reference because they aren't added [here](https://github.com/huggingface/datasets/blob/master/docs/source/package_reference/main_classes.mdx). I'll add these functions to `main_classes.mdx` and then they should be displayed properly (unless you left them out on purpose?)!

docs/source/faiss_es.mdx
stevhliu: It looks like neither format works even though the function - `get_nearest_examples` -  is in the list of `[[autodoc]] datasets.Dataset` and the [path](https://github.com/huggingface/datasets/blob/32e0e79aa57b36a269b6f2ee38449a043a393e50/src/datasets/search.py#L700) to it is correct.

lhoestq: I merged master into this branch to update the doc builder app. Hopefully it fixes the issue

stevhliu: Pinging @sgugger for help as well! 🙌

sgugger: Are you sure you have all dependencies installed?

lhoestq: We could still change to [`Dataset.get_nearest_examples`] which should work, but not sure why [`~datasets.Dataset.get_nearest_examples`] doesn't work

lhoestq: > Are you sure you have all dependencies installed?

I asked myself the same question, but for some reason linking to Dataset.add_faiss_index() does work, even if it's the same dependencies

sgugger: Could you maybe clarify what "doesn't work"? From a quick debugging, doc_builder is able to find the object from the two references (Dataset.get_nearest_examples and datasets.Dataset.get_nearest_examples).

stevhliu: Yeah, the functions are only displayed in code format instead of linking to the API docs.

sgugger: Looked into it more, and yes, you will need to link to `datasets.search.IndexableMixin.get_nearest_examples` because that's where the method actually is. Basically the `doc-builder` finds that when looking at `"Dataset.get_nearest_examples"` so it's the anchor name it gives to that method (you can check it's actually in the link to it: https://huggingface.co/docs/datasets/master/en/package_reference/main_classes#datasets.search.IndexableMixin.get_nearest_examples)

I can work on adding more references that point to it so `Dataset.get_nearest_examples` would also work as a reference, but not sure I will have time before next week :-)

sgugger: Disentangled the mess in [this PR](https://github.com/huggingface/doc-builder/pull/163). Once it's merged, you should have intra-links that work better for those kinds of methods :-)

lhoestq: Cool thanks ! I guess it means can have [`Dataset.get_nearest_examples`] and merge this PR :)

sgugger: Yes, you can even check it works on the doc build of this PR beforehand since the PR is merged.
On my tests locally, this particular reference was properly resolved.

stevhliu: Woo everything looks good now, thanks so much Sylvain! 🥇

lhoestq (approved):
Awesome thanks !

PR #4011, opened 2022-03-24
Fix SQuAD v2 metric docs on `references` format

bryant1410:
`references` it's not a list of dictionaries but a dictionary that has a list in its values.


albertvillanova:

PR #4019, opened 2022-03-25
Make yelp_polarity streamable

lhoestq:
It was using `dl_manager.download_and_extract` on a TAR archive, which is not supported in streaming mode. I replaced this by `dl_manager.iter_archive`


lhoestq:
The CI is failing because of the incomplete dataset card - this is unrelated to the goal of this PR so we can ignore it

PR #4023, opened 2022-03-25
Replace yahoo_answers_topics data url

lhoestq:
I replaced the Google Drive URL of the dataset by the FastAI one, since we've had some issues with Google Drive.


lhoestq:
The CI is failing because of issues in the dataset cards that are unrelated to this PR - merging

PR #4024, opened 2022-03-25
Doc: image_process small tip

FrancescoSaverioZuppichini:
I've added a small tip in the `image_process` doc


mariosasko:
This tip is unnecessary, i.e., Pillow will already be installed since the `Image` feature requires it for encoding and decoding. Thanks anyway.

cc @stevhliu I've noticed we are missing the installation section in the doc (`pip install datasets[vision]`). I can add it myself.

PR #4030, opened 2022-03-26
Use a constant for the articles regex in SQuAD v2

bryant1410:
The main reason for doing this is to be able to change the articles list if using another language, for example. It's not the most elegant solution but at least it makes the metric more extensible with no drawbacks.

BTW, what could be the best way to make this more generic (i.e., SQuAD in other languages)? Maybe receive a regex as an optional param, with the current value as the default? Similarly for SQuAD v1 (can't they re-use code?).

### DIFF
--- a/metrics/squad_v2/evaluate.py
+++ b/metrics/squad_v2/evaluate.py
@@ -16,6 +16,8 @@ import sys
 import numpy as np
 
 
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
Thanks !

I think it would be nice indeed to be able to change the normalization regex depending on the language indeed. cc @lvwerra

PR #4058, opened 2022-03-29
Updated annotations for nli_tr dataset

e-budur:
This PR adds annotation tags for `nli_tr` dataset so that the dataset can be searchable wrt. relevant query parameters.

The annotations in this PR are based on the existing annotations of `snli` and `multi_nli` datasets as `nli_tr` is a machine-generated extension of those datasets.

This PR is intended only for updating the annotation labels but a followup PR will focus on updating the missing sections in the `README.md` as well.

Thanks for all your time to review it.

### DIFF
--- a/datasets/nli_tr/README.md
+++ b/datasets/nli_tr/README.md
@@ -1,4 +1,31 @@
 ---
+annotations_creators:
+- expert-generated
+language_creators:
+- machine-generated
+languages:
+- tr
+licenses:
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
Thanks ! :)

e-budur:
Thank you so much @[lhoestq](https://github.com/lhoestq) for the time you take to your review the PR!

Issue #4051, opened 2022-03-29
ConnectionError: Couldn't reach https://raw.githubusercontent.com/huggingface/datasets/2.0.0/datasets/glue/glue.py


klyuhang9:
Hi, I meet a problem.
When I run the code:
`dataset = load_dataset('glue','sst2')`

There is a issue raising:
ConnectionError: Couldn't reach https://raw.githubusercontent.com/huggingface/datasets/2.0.0/datasets/glue/glue.py

I don't know why; it is ok when I use Google Chrome to view this url. 
Thanks for your help!

albertvillanova:
Hi @klyuhang9,

I'm sorry but I can't reproduce your problem:
```python
In [4]: ds = load_dataset("glue", "sst2", download_mode="force_redownload")
Downloading builder script: 28.8kB [00:00, 9.15MB/s]                                                                                                                                                       
Downloading metadata: 28.7kB [00:00, 10.7MB/s]                                                                                                                                                             
Downloading and preparing dataset glue/sst2 (download: 7.09 MiB, generated: 4.78 MiB, post-processed: Unknown size, total: 11.88 MiB) to .../.cache/huggingface/datasets/glue/sst2/1.0.0/dacbe3125aa31d7f70367a07a8a9e72a5a0bfeb5fc42e75c9db75b96da6053ad...
Downloading data: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 7.44M/7.44M [00:01<00:00, 4.12MB/s]
Dataset glue downloaded and prepared to .../.cache/huggingface/datasets/glue/sst2/1.0.0/dacbe3125aa31d7f70367a07a8a9e72a5a0bfeb5fc42e75c9db75b96da6053ad. Subsequent calls will reuse this data.  
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 3/3 [00:00<00:00, 1047.96it/s]

In [5]: ds
Out[5]: 
DatasetDict({
    train: Dataset({
        features: ['sentence', 'label', 'idx'],
        num_rows: 67349
    })
    validation: Dataset({
        features: ['sentence', 'label', 'idx'],
        num_rows: 872
    })
    test: Dataset({
        features: ['sentence', 'label', 'idx'],
        num_rows: 1821
    })
})
```

Please, note that sometimes GitHub has some temporary connectivity issues. Feel free to retry and re-open this issue if the problem persists.

[...]

Issue #3210, opened 2021-11-04
ConnectionError: Couldn't reach https://raw.githubusercontent.com/huggingface/datasets/1.15.1/datasets/wmt16/wmt16.py


xiuzhilu:
when I use python examples/pytorch/translation/run_translation.py     --model_name_or_path examples/pytorch/translation/opus-mt-en-ro     --do_train     --do_eval     --source_lang en     --target_lang ro     --dataset_name wmt16     --dataset_config_name ro-en     --output_dir /tmp/tst-translation     --per_device_train_batch_size=4     --per_device_eval_batch_size=4     --overwrite_output_dir     --predict_with_generate to finetune translation model on huggingface, I get the issue"ConnectionError: Couldn't reach https://raw.githubusercontent.com/huggingface/datasets/1.15.1/datasets/wmt16/wmt16.py".But I can open the https://raw.githubusercontent.com/huggingface/datasets/1.15.1/datasets/wmt16/wmt16.py by using website. What should I do to solve the issue?

lhoestq:
Hi ! Do you have some kind of proxy in your browser that gives you access to internet ?

Maybe you're having this error because you don't have access to this URL from python ?

xyx361100238:
Hi，do you fixed this error?
I still have this issue when use "use_auth_token=True"

lhoestq:
You don't need authentication to access those github hosted files
Please check that you can access this URL from your browser and also from your terminal

Issue #2787, opened 2021-08-11
ConnectionError: Couldn't reach https://raw.githubusercontent.com


jinec:
Hello,
I am trying to run run_glue.py and it gives me this error -

Traceback (most recent call last):
  File "E:/BERT/pytorch_hugging/transformers/examples/pytorch/text-classification/run_glue.py", line 546, in <module>
    main()
  File "E:/BERT/pytorch_hugging/transformers/examples/pytorch/text-classification/run_glue.py", line 250, in main
    datasets = load_dataset("glue", data_args.task_name, cache_dir=model_args.cache_dir)
  File "C:\install\Anaconda3\envs\huggingface\lib\site-packages\datasets\load.py", line 718, in load_dataset
    use_auth_token=use_auth_token,
  File "C:\install\Anaconda3\envs\huggingface\lib\site-packages\datasets\load.py", line 320, in prepare_module
    local_path = cached_path(file_path, download_config=download_config)
  File "C:\install\Anaconda3\envs\huggingface\lib\site-packages\datasets\utils\file_utils.py", line 291, in cached_path
    use_auth_token=download_config.use_auth_token,
  File "C:\install\Anaconda3\envs\huggingface\lib\site-packages\datasets\utils\file_utils.py", line 623, in get_from_cache
    raise ConnectionError("Couldn't reach {}".format(url))
ConnectionError: Couldn't reach https://raw.githubusercontent.com/huggingface/datasets/1.7.0/datasets/glue/glue.py

Trying to do python run_glue.py  --model_name_or_path
bert-base-cased
--task_name
mrpc
--do_train
--do_eval
--max_seq_length
128
--per_device_train_batch_size
32
--learning_rate
2e-5
--num_train_epochs
3
--output_dir
./tmp/mrpc/

Is this something on my end? From what I can tell, this was re-fixeded by @fullyz a few months ago.
Thank you!

jinec:
the bug code locate in ：
    if data_args.task_name is not None:
        # Downloading and loading a dataset from the hub.
        datasets = load_dataset("glue", data_args.task_name, cache_dir=model_args.cache_dir)

albertvillanova:
Hi @jinec,

From time to time we get this kind of `ConnectionError` coming from the github.com website: https://raw.githubusercontent.com

Normally, it should work if you wait a little and then retry.

jinec:
cannot connect，even by Web browser，please check that  there is some  problems。

albertvillanova:
[...]

Issue #2075, opened 2021-03-18
ConnectionError: Couldn't reach common_voice.py


LifaSun:
When I run: 
from datasets import load_dataset, load_metric

common_voice_train = load_dataset("common_voice", "zh-CN", split="train+validation")
common_voice_test = load_dataset("common_voice", "zh-CN", split="test")

Got:
ConnectionError: Couldn't reach https://raw.githubusercontent.com/huggingface/datasets/master/datasets/common_voice/common_voice.py

Version:
1.4.1

Thanks!  @lhoestq @LysandreJik @thomwolf

albertvillanova:
Hi @LifaSun, thanks for reporting this issue.

Sometimes, GitHub has some connectivity problems. Could you confirm that the problem persists?

LifaSun:
@albertvillanova Thanks! It works well now.

PR #4059, opened 2022-03-30
Load GitHub datasets from Hub

### LINKED ISSUE #2048: github is not always available - probably need a back up
Yesterday morning github wasn't working:

```
:/tmp$ wget https://raw.githubusercontent.com/huggingface/datasets/1.4.1/metrics/sacrebleu/sacrebleu.py--2021-03-12 18:35:59-- https://raw.githubusercontent.com/huggingface/datasets/1.4.1/metrics/sacrebleu/sacrebleu.py
Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 185.199.108.133, 185.199.111.133, 185.199.109.133, ...
Connecting to raw.githubusercontent.com (raw.githubusercontent.com)|185.199.108.133|:443... connected.
HTTP request sent, awaiting response... 500 Internal Server Error
2021-03-12 18:36:11 ERROR 500: Internal Server Error.
```

Suggestion: have a failover system and replicate the data on another system and reach there if gh isn't reachable? perhaps gh can be a master and the replicate a slave - so there is only one true source.

albertvillanova:
We have recurrently had connection errors when requesting GitHub because sometimes the site is not available.

This PR requests the Hub instead, once all GitHub datasets are mirrored on the Hub.

Fix #2048

Related to:
- #4051 
- #3210
- #2787 
- #2075
- #2036


lhoestq:
Cool ! Maybe add a note in setup.py to push to PyPI only after the `Update Hub repositories` CI job is done ? It can take a few minutes to add the version tag to all the dataset repositories on the Hub. If we push to PyPI before the tags are pushed, then some users might get some 404 if at the same time they installed `datasets` and run `load_dataset`.

albertvillanova:
@lhoestq I was going to increase the `max_retries` as done for metrics:
- #4063 

But then I realized that loading from the Hub would work as well. That is why I opened this PR.

Definitely, we should decide which behavior we want:
- We have been working in the direction of eliminating the distinctions between canonical/community datasets
- If we continue to go in that direction, then passing `revision` should have the same behavior for canonical/community
- If we want to continue to tight the library version with the canonical datasets version, that is definitely a difference between canonical and community datasets

Not sure what could be better in the long term...

albertvillanova:
> We could stop having this behavior and always use the latest version of the dataset, but when we do a breaking change it will break github datasets for previous versions of the library. 

Not sure of understanding this. Previous versions of the `datasets` library will continue to download GitHub datasets from GitHub, syncing library/dataset versions...

lhoestq:
Yes you're right, previous versions of `datasets` will still continue to download from github, but not future versions.
If we release `datasets` 2.1 by removing this behavior and if one day we release `datasets` 3.0 with a breaking change in the dataset scripts, then all version >=2.1 will break.

albertvillanova:
Normally in open source projects, when there is a duplicate PR, the latter is tagged as "duplicate" and closed. :stuck_out_tongue_winking_eye: 

Let me make things clear my mind: so you say that the blocking point that was preventing this PR from merging now is no longer a blocking point and could be addresses in a subsequent PR?

lhoestq (approved):
Cool ! LGTM :)

Finally we'll remove the differences between Hub datasets and GitHub datasets ^^

(Note that after this PR, all the changes made to a dataset will affect all the `datasets` version from now on)

albertvillanova:
> Note that after this PR, all the changes made to a dataset will affect all the datasets version from now on
Yes, we have aligned this behavior with Hub datasets, as this is already the case for Hub datasets.

PR #4060, opened 2022-03-30
Deprecate canonical Multilingual Librispeech

polinaeterna:
Deprecate canonical Multilingual Librispeech in favor of [the community one](https://huggingface.co/datasets/facebook/multilingual_librispeech) which supports streaming.

However, there is a problem regarding new ASR template schema: since it's changed, I guess all community datasets that use this template do not work with new version of the library, including MLS. Should we somehow notify users about that or is it possible to change this line ourselves? For MLS specifically, I cannot change the code directly as I'm not the member of the Facebook org. 

Hm, and the code should be change after the release, no?


lhoestq:
Yes, as discussed in #4006 we should update facebook/multilingual_librispeech indeed before we do a release. @anton-l could you help taking care of updating facebook/multilingual_librispeech ? We need to update the task template
```python
task_templates=[AutomaticSpeechRecognition(audio_column="audio", transcription_column="text")],
```
and write that `datasets>=2.1` is necessary to load it in the dataset card.

Once the change is done we can merge this PR and do the release I think

anton-l:
@polinaeterna @lhoestq 
Updated the script and the dataset card: https://huggingface.co/datasets/facebook/multilingual_librispeech

polinaeterna:
@anton-l @lhoestq now previewer doesn't work for this datasets as it cannot recognize new `audio_column` argument:
image

I'm not an expert in previewer things, where should I look into the corresponding code?

severo:
Yes, there are several datasets with the same error, eg https://github.com/huggingface/datasets-preview-backend/issues/188. I'm not sure what I should do to fix this? Upgrade datasets to master?

lhoestq:
@anton-l ended up removing the task template in facebook/multilingual_librispeech to make it work for old version of `datasets` and fix the viewer :) thanks !

polinaeterna:
@lhoestq can we merge now? ^^

Issue #3134, opened 2021-10-22
Couldn't reach https://raw.githubusercontent.com/huggingface/datasets/1.11.0/metrics/rouge/rouge.py


yanan1116:
datasets version: 1.12.1

`metric = datasets.load_metric('rouge')`

The error:

> ConnectionError                           Traceback (most recent call last)
> <ipython-input-3-dd10a0c5212f> in <module>
> ----> 1 metric = datasets.load_metric('rouge')
> 
> /usr/local/lib/python3.6/dist-packages/datasets/load.py in load_metric(path, config_name, process_id, num_process, cache_dir, experiment_id, keep_in_memory, download_config, download_mode, script_version, **metric_init_kwargs)
>     613         download_config=download_config,
>     614         download_mode=download_mode,
> --> 615         dataset=False,
>     616     )
>     617     metric_cls = import_main_class(module_path, dataset=False)
> 
> /usr/local/lib/python3.6/dist-packages/datasets/load.py in prepare_module(path, script_version, download_config, download_mode, dataset, force_local_path, dynamic_modules_path, return_resolved_file_path, **download_kwargs)
>     328                 file_path = hf_github_url(path=path, name=name, dataset=dataset, version=script_version)
>     329                 try:
> --> 330                     local_path = cached_path(file_path, download_config=download_config)
>     331                 except FileNotFoundError:
>     332                     if script_version is not None:
> 
> /usr/local/lib/python3.6/dist-packages/datasets/utils/file_utils.py in cached_path(url_or_filename, download_config, **download_kwargs)
>     296             use_etag=download_config.use_etag,
>     297             max_retries=download_config.max_retries,
> --> 298             use_auth_token=download_config.use_auth_token,
>     299         )
>     300     elif os.path.exists(url_or_filename):
> 
> /usr/local/lib/python3.6/dist-packages/datasets/utils/file_utils.py in get_from_cache(url, cache_dir, force_download, proxies, etag_timeout, resume_download, user_agent, local_files_only, use_etag, max_retries, use_auth_token)
>     603             raise FileNotFoundError("Couldn't find file at {}".format(url))
>     604         _raise_if_offline_mode_is_enabled(f"Tried to reach {url}")
> --> 605         raise ConnectionError("Couldn't reach {}".format(url))
>     606
>     607     # Try a second time
> 
> ConnectionError: Couldn't reach https://raw.githubusercontent.com/huggingface/datasets/1.11.0/metrics/rouge/rouge.py

[...]

PR #4063, opened 2022-03-30
Increase max retries for GitHub metrics

albertvillanova:
As GitHub recurrently raises connectivity issues, this PR increases the number of max retries to request GitHub metrics.

Related to:
- #3134

Also related to:
- #4059

### DIFF
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -581,6 +581,7 @@ class GithubMetricModuleFactory(_MetricModuleFactory):
         self.name = name
         self.revision = revision
         self.download_config = download_config or DownloadConfig()
+        self.download_config.max_retries = 3
         self.download_mode = download_mode
         self.dynamic_modules_path = dynamic_modules_path
         assert self.name.count("/") == 0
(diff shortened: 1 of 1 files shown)


src/datasets/load.py
```diff
@@ -581,6 +581,7 @@ def __init__(
         self.name = name
         self.revision = revision
         self.download_config = download_config or DownloadConfig()
+        self.download_config.max_retries = 3
```
lhoestq, line 584: Maybe just change it only if it's < 3 ?

src/datasets/load.py
```diff
@@ -581,6 +581,7 @@ def __init__(
         self.name = name
         self.revision = revision
         self.download_config = download_config or DownloadConfig()
```
lhoestq, line 583: you can copy the config here to not modify the user's object if they explicitly passed a DownloadConfig.
```suggestion
        self.download_config = download_config.copy() or DownloadConfig()
```

lhoestq (approved):
thanks :)

PR #4066, opened 2022-03-30
Tasks alignment with models

lhoestq:
I updated our `tasks.json` file with the new task taxonomy that is aligned with models.

The rule that defines a task is the following:

**Two tasks are different if and only if the steps of their pipelines** are different, i.e. if they can’t reasonably be implemented using the same coherent code (level of granularity/complexity of the code to be defined - ideally I’d like to say “HF user’s level”) - this is the same definition in `transformers` 

I will update the tags of all the datasets in this repository [in another PR](https://github.com/huggingface/datasets/pull/4067) for readability.

Main changes:
- conditional-text-generation is split between summarization, translation, text-generation and text2text-generation
- speech-processing is split into automatic-speech-recognition, audio-classification, etc.
- structure-prediction is renamed token-classification
- abstractive-qa now belongs to text2text-generation

Here is just a simplified YAML dump of `tasks.json`:

```yaml
audio-classification:
- keyword-spotting
- speaker-identification
- speaker-intent-classification
- emotion-recognition
- speaker-language-identification
audio-to-audio: []
automatic-speech-recognition: []
conversational:
- dialogue-generation
feature-extraction: []
fill-mask:
- slot-filling
- masked-language-modeling
image-classification:
- multi-label-image-classification
- multi-class-image-classification
image-segmentation:
- instance-segmentation
- semantic-segmentation
- panoptic-segmentation
image-to-text:
- image-captioning
multiple-choice:
- multiple-choice-qa
- multiple-choice-coreference-resolution
object-detection:
- face-detection
- vehicle-detection
question-answering:
- extractive-qa
- open-domain-qa
- closed-domain-qa
sentence-similarity: []
tabular-classification: []
tabular-to-text:
- rdf-to-text
summarization:
- news-articles-summarization
- news-articles-headline-generation
table-to-text: []
table-question-answering: []
text-classification:
- acceptability-classification
- entity-linking-classification
- fact-checking
- intent-classification
- multi-class-classification
- multi-label-classification
- natural-language-inference
- semantic-similarity-classification
- sentiment-classification
- topic-classification
- semantic-similarity-scoring
- sentiment-scoring
- sentiment-analysis
- hate-speech-detection
- text-scoring
text-generation:
- dialogue-modeling
- language-modeling
text-retrieval:
- document-retrieval
- utterance-retrieval
- entity-linking-retrieval
- fact-checking-retrieval
text-to-image: []
text-to-tabular:
- relation-extraction
- semantic-role-labeling
text-to-speech: []
[...]

PR #4067, opened 2022-03-30
Update datasets task tags to align tags with models

lhoestq:
**Requires https://github.com/huggingface/datasets/pull/4066 to be merged first**

Following https://github.com/huggingface/datasets/pull/4066 we need to update many dataset tags to use the new ones. This PR takes case of this and is quite big - feel free to review only certain tags if you don't want to spend too much time on it.

Note that the CI will never be green for this PR, because many dataset cards have missing tags or sections, and fixing them is out of scope of this PR (the CI on master will be green anyway)


lhoestq:
Just updated the tags of vision datasets :)
We can figure out one for image datasets without labels like PASS - not sure how to name the task for this, maybe `image-fill-mask` (for consistency with language modeling for pretraining) / `masked-auto-encoding` (from ViT). Let's see that in another PR later

datasets/svhn/README.md
```diff
@@ -18,7 +18,7 @@ task_categories:
 - image-classification
 - object-detection
 task_ids:
-- single-label-image-classification
+- object-detection-other-digit-detection
```
mariosasko, line 21: Would remove this and define task ids as `task_ids: []`

mariosasko (commented):
IMO we should use `multi-class image-classification` as `task_id` for the updated vision datasets (beans, cats_vs_dogs, cifar10, cifar100) and not `image-classification-other ...`.

datasets/svhn/README.md
lhoestq: done !

mariosasko (approved):
LGTM!

PR #4076, opened 2022-03-31
Add ROUGE Metric Card

emibaylor:
Add ROUGE metric card.

I've left the 'Values from popular papers' section empty for the time being because I don't know the summarization literature very well and am therefore not sure which paper(s) to pull from (note that the original rouge paper does not seem to present specific values, just correlations with human judgements). Any suggestions on which paper(s) to pull from would be helpful! :)

### DIFF
--- a/metrics/rouge/README.md
+++ b/metrics/rouge/README.md
@@ -1,0 +1,78 @@
+# Metric Card for ROUGE
+
+## Metric Description
+ROUGE, or Recall-Oriented Understudy for Gisting Evaluation, is a set of metrics and a software package used for
+evaluating automatic summarization and machine translation software in natural language processing.
+The metrics compare an automatically produced summary or translation against a reference or a set of references (human-produced) summary or translation.
+
+Note that ROUGE is case insensitive, meaning that upper case letters are treated the same way as lower case letters.
+
+This metrics is a wrapper around [Google Research reimplementation of ROUGE](https://github.com/google-research/google-research/tree/master/rouge)
+
+## How to Use
+At minimum, this metric takes as input a list of predictions and a list of references:
+```python
+rouge = datasets.load_metric('rouge')
+predictions = ["hello there", "general kenobi"]
+references = ["hello there", "general kenobi"]
+results = rouge.compute(predictions=predictions,
+                        references=references)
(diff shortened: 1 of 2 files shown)


metrics/rouge/README.md
```diff
@@ -0,0 +1,121 @@
+# Metric Card for ROUGE
+
+## Metric Description
+ROUGE, or Recall-Oriented Understudy for Gisting Evaluation, is a set of metrics and a software package used for evaluating automatic summarization and machine translation software in natural language processing. The metrics compare an automatically produced summary or translation against a reference or a set of references (human-produced) summary or translation.
+
+Note that ROUGE is case insensitive, meaning that upper case letters are treated the same way as lower case letters.
```
lhoestq, line 6: (nit) maybe clearer this way ?

```suggestion
Note that ROUGE is case insensitive, meaning that upper case letters are considered to be the same as lower case letters.
```

metrics/rouge/README.md
```diff
@@ -0,0 +1,121 @@
...
+        - `"rouge1"`: unigram (1-gram) based scoring
+        - `"rouge2"`: bigram (2-gram) based scoring
+        - `"rougeL"`: Longest common subsequence based scoring.
+        - `"rougeLSum"`: splits text using `"\n"`
+        - See [here](https://github.com/huggingface/datasets/issues/617) for more information
+- **use_aggregator** (`boolean`): If True, returns aggregates. Defaults to `True`.
+- **use_stemmer** (`boolean`): If `True`, uses Porter stemmer to strip word suffixes. Defaults to `False`.
+
+### Output Values
+The output is a dictionary with one entry for each rouge type in the input list `rouge_types`. If `use_aggregator=False`, each dictionary entry is a list of Score objects, with one score for each sentence. Each Score object includes the `precision`, `recall`, and `fmeasure`. E.g. if `rouge_types=['rouge1', 'rouge2']` and `use_aggregator=False`, the output is:
```
lhoestq, line 42: Maybe mention that by default the aggregator randomly samples 1000 elements to compute the aggregated metrics ?

lhoestq (approved):
Thank you :)

PR #4078, opened 2022-04-01
Fix GithubMetricModuleFactory instantiation with None download_config

albertvillanova:
Recent PR:
- #4063

introduced a potential bug if `GithubMetricModuleFactory` is instantiated with None `download_config`.

This PR add instantiation tests and fix that potential issue.

CC: @lhoestq


lhoestq (approved):
good catch !

Issue #4077, opened 2022-04-01
ArrowInvalid: Parquet magic bytes not found in footer. Either the file is corrupted or this is not a parquet file.


NielsRogge:
## Describe the bug

When uploading a relatively large image dataset of > 1GB, reloading doesn't work for me, even though pushing to the hub went just fine.

Basically, I do:

```
dataset.push_to_hub("dataset_name") # works fine, no errors

reloaded_dataset = load_dataset("dataset_name")
```

and it returns:

```
/usr/local/lib/python3.7/dist-packages/pyarrow/error.pxi in pyarrow.lib.check_status()

ArrowInvalid: Parquet magic bytes not found in footer. Either the file is corrupted or this is not a parquet file.
```

I created a Colab notebook to reproduce my error: https://colab.research.google.com/drive/141LJCcM2XyqprPY83nIQ-Zk3BbxWeahq?usp=sharing

PR #4081, opened 2022-04-01
Close parquet writer properly in `push_to_hub`

lhoestq:
We don’t call writer.close(), which causes https://github.com/huggingface/datasets/issues/4077. I can happen that we upload the file before the writer is garbage collected and writes the footer.

I fixed this by explicitly closing the parquet writer.

Close https://github.com/huggingface/datasets/issues/4077.


mariosasko (approved):
Thanks! LGTM!

albertvillanova (approved):
Good catch!

PR #4082, opened 2022-04-01
Add chrF(++) Metric Card

emibaylor:


metrics/chrf/chrf.py
```diff
@@ -73,31 +73,52 @@
...
-    lowercase: Enable case-insensitivity.
-    whitespace: If `True`, include whitespaces when extracting character n-grams.
-    eps_smoothing: If `True`, applies epsilon smoothing similar
-    to reference chrF++.py, NLTK and Moses implementations. Otherwise,
-    it takes into account effective match order similar to sacreBLEU < 2.0.0.
+    predictions (list of str): The predicted sentences.
+    references (list of list of str): The references. There should be one reference sub-list for each prediction sentence.
+    char_order (int): Character n-gram order. Defaults to `6`.
+    word_order (int): Word n-gram order. If equals to `2`, the metric is referred to as chrF++. Defaults to `0`.
+    beta (int): Determine the importance of recall w.r.t precision. Defaults to `2`.
```
emibaylor, line 80: One thing to note here is that the defaults in our code actually just reference the defaults in the source repo, instead of inputting these default values. For the time being, I've written what those referenced defaults are, but I think it would either be good to do one of the following:
- change the documentation here to explain that the defaults are set in another repo
- set the defaults explicitly here
Let me know which one (if either) you think is best!

lhoestq: You can just redirect to their repository or their documentation I think.

Otherwise if they update their repository, our documentation will be outdated and can be misleading.

metrics/chrf/README.md
```diff
@@ -0,0 +1,133 @@
...
+See the [sacreBLEU README.md](https://github.com/mjpost/sacreBLEU#chrf--chrf) for more information.
+
+
+## How to Use
+At minimum, this metric requires a `list` of predictions and a `list` of `list`s of references:
+```python
+>>> prediction = ["The relationship between cats and dogs is not exactly friendly.", "a good bookshop is just a genteel black hole that knows how to read."]
+>>> reference = [["The relationship between dogs and cats is not exactly friendly.", ], ["A good bookshop is just a genteel Black Hole that knows how to read."]]
+>>> chrf = datasets.load_metric("chrf")
+>>> results = chrf.compute(predictions=prediction, references=reference)
```
lhoestq, line 18:
```suggestion
>>> predictions = ["The relationship between cats and dogs is not exactly friendly.", "a good bookshop is just a genteel black hole that knows how to read."]
>>> references = [["The relationship between dogs and cats is not exactly friendly.", ], ["A good bookshop is just a genteel Black Hole that knows how to read."]]
>>> chrf = datasets.load_metric("chrf")
>>> results = chrf.compute(predictions=predictions, references=references)
```

can you also do the same for the others (both in the readme and in the docstring) ?

lhoestq (approved):
Thank you ! :)

PR #4087, opened 2022-04-04
Fix BeamWriter output Parquet file

albertvillanova:
Since now, the `BeamWriter` saved a Parquet file with a simplified schema, where each field value was serialized to JSON. That resulted in Parquet files larger than Arrow files.

This PR:
- writes Parquet file preserving original schema and without serialization, thus avoiding serialization overhead and resulting in a smaller output file size.
- fixes `parquet_to_arrow` function


lhoestq (approved):
Thanks for removing this ugly stuff x)

PR #4095, opened 2022-04-05
fix typo in rename_column error message

hunterlang:
I feel bad submitting such a tiny change as a PR but it confused me today 😄


albertvillanova (approved):
Thanks, tiny but necessary! ;)

Issue #4105, opened 2022-04-06
push to hub fails with huggingface-hub 0.5.0


frascuchon:
## Describe the bug
`ds.push_to_hub` is failing when updating a dataset in the form "org_id/repo_id"

## Steps to reproduce the bug
```python
from datasets import load_dataset
ds = load_dataset("rubrix/news_test")
ds.push_to_hub("<your-user>/news_test", token="<your-token>")
```

## Expected results
The dataset is successfully uploaded

## Actual results
An error validation is raised:

```bash
if repo_id and (name or organization):
>           raise ValueError(
                "Only pass `repo_id` and leave deprecated `name` and "
                "`organization` to be None."
E               ValueError: Only pass `repo_id` and leave deprecated `name` and `organization` to be None.
```

## Environment info

- `datasets` version: 1.18.1
- Platform: macOS
- Python version: 3.8.12
- PyArrow version: 6.0.0

lhoestq:
Hi ! Indeed there was a breaking change in `huggingface_hub` 0.5.0 in `HfApi.create_repo`, which is called here in `datasets` by passing the org name in both the `repo_id` and the `organization` arguments:

https://github.com/huggingface/datasets/blob/2230f7f7d7fbaf102cff356f5a8f3bd1561bea43/src/datasets/arrow_dataset.py#L3363-L3369

I think we should fix that in `huggingface_hub`, will keep you posted. In the meantime please use `huggingface_hub` 0.4.0

adrinjalali:
I'll be sending a fix for this later today on the `huggingface_hub` side.

The error would be converted to a `FutureWarning` if `datasets` uses kwargs instead of positional, for example here: 

https://github.com/huggingface/datasets/blob/2230f7f7d7fbaf102cff356f5a8f3bd1561bea43/src/datasets/arrow_dataset.py#L3363-L3369

to be:

``` python
            api.create_repo(
                name=dataset_name,
                token=token,
                repo_type="dataset",
                organization=organization,
                private=private,
            )
```

But `name` and `organization` are deprecated in `huggingface_hub=0.5`, and people should pass `repo_id='org/name` instead. Note that `repo_id` was introduced in 0.5 and if `datasets` wants to support older `huggingface_hub` versions (which I encourage it to do), there needs to be a helper function to do that. It can be something like:


```python
def create_repo(
    client,
    name: str,
    token: Optional[str] = None,
    organization: Optional[str] = None,
    private: Optional[bool] = None,
[...]

PR #4106, opened 2022-04-06
Support huggingface_hub 0.5

lhoestq:
Following https://github.com/huggingface/datasets/issues/4105

`huggingface_hub` deprecated some parameters in `HfApi` in 0.5. This PR updates all the calls to HfApi to remove all the deprecations, <s>and I set the `hugginface_hub` requirement to `>=0.5.0`</s>

cc @adrinjalali @LysandreJik

### DIFF
--- a/tests/test_upstream_hub.py
+++ b/tests/test_upstream_hub.py
@@ -14,6 +14,7 @@ from huggingface_hub.hf_api import HfFolder
 from packaging import version
 
 from datasets import Audio, ClassLabel, Dataset, DatasetDict, Features, Image, Value, load_dataset
+from datasets.utils._hf_hub_fixes import delete_repo
 from tests.utils import require_pil, require_sndfile
 
 
@@ -40,11 +41,8 @@ class TestPushToHub(TestCase):
     _api = HfApi(endpoint=ENDPOINT_STAGING)
 
     def cleanup_repo(self, ds_name):
-        if version.parse(huggingface_hub.__version__) < version.parse("0.5.0"):
-            organization, name = ds_name.split("/")
-            self._api.delete_repo(name, token=self._token, repo_type="dataset", organization=organization)
-        else:  # the `organization` parameter is deprecated in huggingface_hub>=0.5.0
-            self._api.delete_repo(ds_name, token=self._token, repo_type="dataset")
+        organization, name = ds_name.split("/")
+        delete_repo(hf_api=self._api, name=name, organization=organization, token=self._token, repo_type="dataset")
 
     @classmethod
     def setUpClass(cls):
--- a/tests/test_upstream_hub.py
+++ b/tests/test_upstream_hub.py
@@ -7,9 +7,11 @@ from pathlib import Path
 from unittest import TestCase
 from unittest.mock import patch
 
+import huggingface_hub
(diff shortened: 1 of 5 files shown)


lhoestq:
Looks like GH actions is not able to resolve `huggingface_hub` 0.5.0, I'm investivating

adrinjalali:
I'm glad to see changes in `huggingface_hub` are simplifying code here.

julien-c:
seems to supersede #4102, feel free to close mine :)

albertvillanova:
I found the issue:
- https://github.com/huggingface/huggingface_hub/pull/790

lhoestq:
Good catch, `huggingface_hub` doesn't support python 3.6 anymore indeed, therefore we should keep support for 0.4.0. I'm reverting the requirement version bump for now.

We can update the requirement once we drop support for python 3.6 in `datasets`

albertvillanova:
@lhoestq, I've opened this PR on `huggingface_hub`: 
- https://github.com/huggingface/huggingface_hub/pull/823

Is there a strong reason wht `huggingface_hub` no longer support Python 3.6?

lhoestq:
I think `datasets` can drop support for 3.6 soon. But for now maybe let's keep support for 0.4.0, python 3.6 users are not affected by https://github.com/huggingface/datasets/issues/4105 anyway.

`huggingface_hub` doesn't not have to support 3.6 again just for the CI IMO

albertvillanova:
@lhoestq I commented on the PR, that IMO it is not a good practice to drop support for Python 3.6 without a previous deprecation cycle.

lhoestq:
Re-added support for older versions. I ended up checking `huggingface_hub` version to use the old, deprecated API for <0.5.0

adrinjalali:
I find it good practice to have all dependency version related code in a single file so that when you decide to remove support for an old version of a dependency it's easy to find and remove them, hence suggesting `utils/_fixes.py` in https://github.com/huggingface/datasets/issues/4105#issuecomment-1090041204

lhoestq:
good idea, thanks !

lhoestq:
I used your suggestion @adrinjalali , I just replace the try/except with a check on the version of `huggingface_hub`

tests/test_upstream_hub.py
```diff
@@ -25,34 +27,25 @@
...
 @with_staging_testing
 class TestPushToHub(TestCase):
     _api = HfApi(endpoint=ENDPOINT_STAGING)
 
+    def cleanup_repo(self, ds_name):
+        if version.parse(huggingface_hub.__version__) < version.parse("0.5.0"):
+            organization, name = ds_name.split("/")
+            self._api.delete_repo(name, token=self._token, repo_type="dataset", organization=organization)
+        else:  # the `organization` parameter is deprecated in huggingface_hub>=0.5.0
+            self._api.delete_repo(ds_name, token=self._token, repo_type="dataset")
```
adrinjalali, line 47: same with this one?

lhoestq: done ! thanks

adrinjalali (approved):

albertvillanova (approved):
Thanks for the fix.

PR #4116, opened 2022-04-06
Pretty print dataset info files

mariosasko:
Adds indentation to the `dataset_infos.json` file when saving for nicer diffs.

(suggested by @julien-c)

This PR also updates the info files of the GH datasets. Note that this change adds more than **10 MB** to the repo size (the total file size before the change: 29.672298 MB, after: 41.666475 MB), so I'm not sure this change is a good idea.

`src/datasets/info.py` is the only relevant file for reviewers.


julien-c:
maybe just do it from now on no? (i.e. not for existing `dataset_infos.json` files)

julien-c:
yep sounds good too on my side!

lhoestq (approved):
Thanks :)

PR #4137, opened 2022-04-10
Add single dataset citations for TweetEval

gchhablani:
This PR adds single data citations as per request of the original creators of the TweetEval dataset.

This is a recent email from the creator:

> Could I ask you a favor? Would you be able to add at the end of the README the citations of the single datasets as well? You can just copy our readme maybe? https://github.com/cardiffnlp/tweeteval#citing-tweeteval
(just to be sure that the creator of the single datasets also get credits when tweeteval is used)

Please let me know if this looks okay or if any changes are needed.

Thanks,
Gunjan

### DIFF
--- a/datasets/tweet_eval/README.md
+++ b/datasets/tweet_eval/README.md
@@ -464,6 +464,90 @@ year={2020}
 }
 ```
 
+If you use any of the TweetEval datasets, please cite their original publications:
+
+#### Emotion Recognition:
+```
+@inproceedings{mohammad2018semeval,
+  title={Semeval-2018 task 1: Affect in tweets},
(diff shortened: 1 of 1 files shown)


gchhablani:
The `test_dataset_cards` method is failing with the error:

```
if error_messages:
>           raise ValueError("\n".join(error_messages))
E           ValueError: The following issues have been found in the dataset cards:
E           YAML tags:
E           The following typing errors are found: {'annotations_creators': "(Expected `typing.List` with length > 0. Found value of type: `<class 'list'>`, with length: 0.\n)\nOR\n(Expected `typing.Dict` with length > 0. Found value of type: `<class 'list'>`, with length: 0.\n)"}
```

Should this case be allowed? @lhoestq

lhoestq (approved):
Thanks a lot :)

PR #4144, opened 2022-04-11
Fix splits in local packaged modules, local datasets without script and hub datasets without script

### LINKED ISSUE #4150: Inconsistent splits generation for datasets without loading script (packaged dataset puts everything into a single split)
## Describe the bug
Splits for dataset loaders without scripts are prepared inconsistently. I think it might be confusing for users.

## Steps to reproduce the bug
* If you load a packaged datasets from Hub, it infers splits from directory structure / filenames (check out the data [here](https://huggingface.co/datasets/nateraw/test-imagefolder-dataset)):
```python
ds = load_dataset("nateraw/test-imagefolder-dataset")
print(ds)
### Output:
DatasetDict({
    train: Dataset({
        features: ['image', 'label'],
        num_rows: 6
    })
    test: Dataset({
        features: ['image', 'label'],
        num_rows: 4
    })
})
```
* If you do the same from locally stored data specifying only directory path you'll get the same:
```python
ds = load_dataset("/path/to/local/data/test-imagefolder-dataset")
print(ds)
### Output:
DatasetDict({
    train: Dataset({
        features: ['image', 'label'],
        num_rows: 6
    })
    test: Dataset({
        features: ['image', 'label'],
        num_rows: 4
    })
})
```
* However, if you explicitely specify package name (like `imagefolder`, `csv`, `json`), all the data is put into a single split:
```python
ds = load_dataset("imagefolder", data_dir="/path/to/local/data/test-imagefolder-dataset")
print(ds)
### Output:
DatasetDict({
    train: Dataset({
        features: ['image', 'label'],
        num_rows: 10
    })
})
```

## Expected results
For `load_dataset("imagefolder", data_dir="/path/to/local/data/test-imagefolder-dataset")` I expect the same output as of the two first options.

polinaeterna:
fixes #4150

I suggest to infer splits structure from files when `data_dir` is passed with `get_patterns_locally`, analogous to what's done in `LocalDatasetModuleFactoryWithoutScript` with `self.path`, instead of generating files with `data_dir/**` patterns and putting them all into a single default (train) split.

I would also suggest to align `HubDatasetModuleFactoryWithoutScript` and `LocalDatasetModuleFactoryWithoutScript` with this logic (remove `data_files = os.path.join(data_dir, "**")`). It's not reflected in the current code now as I'd like to discuss it cause I might be unaware of some use cases. @lhoestq @mariosasko @albertvillanova  WDYT?


lhoestq:
Thanks !
I'm in favor of this change, even though it's a breaking change:

if you had a dataset
```
data/
  train.csv
  test.csv
```

then running this code would now return both train and test splits:
```python
load_dataset("csv", data_dir="data/")
```
whereas right now it returns only a train split with the data from both CSV files.

In my opinion it's ok do do this breaking change because:
- it makes this behavior consistent with `load_dataset("path/to/data")` that also returns both splits: data_files resolution must be the same
- I don't expect too many affected users (unless people really wanted to group train and test images in the train split on purpose ?) compared to the many new users to come (especially with #4069 )
- this usage will become more and more common as we add packaged builder and imagefolder/audiofolder usage grows, so it may be better to do this change early

Let me know if you think this is acceptable @mariosasko or not, and if you think we need to first have a warning for some time before switching to this new behavior

polinaeterna:
Also, if people really want to put train and test images in a single train split they could do 
`load_dataset("imagefolder", data_files={"train": "/path/to/data/**})`. Probably (arguably :)), if this is a more counterintuitive case, then it should require manual files specification, not a default one (in which we expect that users do want to infer splits from filenames / dir structure but currently they have to pass smth like `{"train": "train.csv", "test": "test.csv"}` explicitly as `data_files`)

mariosasko:
I also like this change, and  I don't think we even need a warning during the transition period, considering I've been asked several times since the release of `imagefolder` why splits are not correctly inferred if the directory structure is as follows:
```
data_dir
    train
        label_a
            0.jpg
            ...
        label_b 
            0.jpg
            ...
    test
        label_a
            0.jpg
            ...
        label_b 
            0.jpg
            ...
```

lhoestq:
Cool ! Feel free to add a test (maybe something similar to `test_PackagedDatasetModuleFactory_with_data_dir` but with a data_dir that contains several splits) and mark this PR as ready for review then @polinaeterna :)

polinaeterna:
@lhoestq @mariosasko do you think it's a good idea to do the same with `HubDatasetModuleFactoryWithoutScript` and `LocalDatasetModuleFactoryWithoutScript` (see current changes). **I'm not sure I understand use cases when Hub dataset is used with `data_dir`** but if we agree on the current change, doing 
```python
ds = load_dataset("polinaeterna/jsonl_test", data_dir="data/")
```
on dataset with the following structure:
```
train.jsonl
test.jsonl
data/
   train.jsonl
   test.jsonl
```
will result in having two splits from files under `data/` dir in specified repo, while master version returns a single train split. 
The same would be for local dataset without script if doing smth like:
```python
ds = load_dataset("/home/polina/workspace/repos/jsonl_test", data_dir="/home/polina/workspace/repos/jsonl_test/)
```
Let me know if you think we should preserve the same logic for all factories or if I should roll back this change.

polinaeterna:
@lhoestq @mariosasko do you think it's a good idea to do the same with `HubDatasetModuleFactoryWithoutScript` and `LocalDatasetModuleFactoryWithoutScript` (see current changes). **I'm not sure I understand use cases when Hub dataset is used with `data_dir`** but if we agree on the current change, doing 
```python
ds = load_dataset("polinaeterna/jsonl_test", data_dir="data/")
```
on dataset with the following structure:
```
train.jsonl
test.jsonl
data/
   train.jsonl
   test.jsonl
```
will result in having two splits from files under `data/` dir in specified repo, while master version returns a single train split. 
The same would be for local dataset without script if doing smth like:
```python
ds = load_dataset("/home/polina/workspace/repos/jsonl_test", data_dir="/home/polina/workspace/repos/jsonl_test/)
```
Let me know if you think we should preserve the same logic for all factories or if I should roll back this change.

src/datasets/data_files.py
```diff
@@ -296,10 +296,13 @@ def get_patterns_locally(base_path: str) -> Dict[str, List[str]]:
 def _resolve_single_pattern_in_dataset_repository(
     dataset_info: huggingface_hub.hf_api.DatasetInfo,
     pattern: str,
+    base_path: Optional[str] = None,
     allowed_extensions: Optional[list] = None,
 ) -> List[PurePath]:
     data_files_ignore = FILES_TO_IGNORE
     fs = HfFileSystem(repo_info=dataset_info)
+    if base_path:
+        pattern = os.path.join(base_path, pattern)
```
lhoestq, line 305: The `HfFileSystem` expects paths with a `/` separator. However `os.path.join` on windows uses the `\` separator.

```suggestion
        pattern = f"{base_path}/{pattern}"
```

src/datasets/load.py
```diff
@@ -733,20 +733,25 @@ def __init__(
         data_files: Optional[Union[str, List, Dict]] = None,
         download_mode: Optional[DownloadMode] = None,
     ):
-        if data_files is None and data_dir is not None:
-            data_files = os.path.join(data_dir, "**")
 
         self.path = path
         self.name = Path(path).stem
         self.data_files = data_files
+        self.data_dir = data_dir if data_dir else None
```
lhoestq, line 740:
```suggestion
        self.data_dir = data_dir
```

src/datasets/load.py
```diff
@@ -782,22 +787,21 @@ def __init__(
         download_config: Optional[DownloadConfig] = None,
         download_mode: Optional[DownloadMode] = None,
     ):
-        if data_files is None and data_dir is not None:
-            data_files = os.path.join(data_dir, "**")
 
         self.name = name
         self.data_files = data_files
-        self.downnload_config = download_config
+        self.data_dir = str(Path(data_dir).resolve()) if data_dir else str(Path().resolve())
```
lhoestq, line 793: (nit) for consistency with the other module factories, I would just leave
```suggestion
        self.data_dir = data_dir
```

and check whether `self.data_dir` is None in `get_module()`

src/datasets/load.py
```diff
@@ -782,22 +787,21 @@ def __init__(
...
+        self.download_config = download_config
         self.download_mode = download_mode
         increase_load_count(name, resource_type="dataset")
 
     def get_module(self) -> DatasetModule:
         patterns = (
-            sanitize_patterns(self.data_files)
-            if self.data_files is not None
-            else get_patterns_locally(str(Path().resolve()))
+            sanitize_patterns(self.data_files) if self.data_files is not None else get_patterns_locally(self.data_dir)
```
lhoestq, line 800: (following previous comment)

in that case, you can check if `self.data_dir` is None here:

```suggestion
            sanitize_patterns(self.data_files)
            if self.data_files is not None
            else get_patterns_locally(self.data_dir)
            if self.data_dir is not None
            else get_patterns_locally(str(Path().resolve()))
        )
```

tests/test_load.py
```diff
@@ -231,6 +245,21 @@ def test_LocalDatasetModuleFactoryWithoutScript(self):
         assert importlib.import_module(module_factory_result.module_path) is not None
         assert os.path.isdir(module_factory_result.builder_kwargs["base_path"])
 
+    def test_LocalDatasetModuleFactoryWithoutScript_with_data_dir(self):
+        factory = LocalDatasetModuleFactoryWithoutScript(self._data_dir, data_dir=self._sub_data_dir)
```
lhoestq, line 249: I think a more probable case is when data_dir is a subdirectory of the base path:
```python
LocalDatasetModuleFactoryWithoutScript(data_dir_with_two_levels, data_dir="subdir1")
```

where data_dir_with_two_levels is like
```
data_dir2/
  subdir1/
    train.json
    test.json
  subdir2
    train.json
    test.json
```

We would also need the same test for `HubDatasetModuleFactoryWithoutScript`

src/datasets/load.py
```diff
@@ -733,20 +733,25 @@ def __init__(
...
         self.data_files = data_files
+        self.data_dir = data_dir if data_dir else None
         self.download_mode = download_mode
 
     def get_module(self) -> DatasetModule:
         patterns = (
-            sanitize_patterns(self.data_files) if self.data_files is not None else get_patterns_locally(self.path)
+            sanitize_patterns(self.data_files)
+            if self.data_files is not None
+            else get_patterns_locally(self.data_dir)
```
lhoestq, line 747: If `self.data_dir` is a relative path, I think it should be relative to the dataset repository's root `self.path`, for consistency with HubDatasetModuleFactoryWithoutScript

src/datasets/load.py
```diff
@@ -733,20 +733,25 @@ def __init__(
...
+            if self.data_dir is not None
+            else get_patterns_locally(self.path)
         )
         data_files = DataFilesDict.from_local_or_remote(
-            patterns, base_path=self.path, allowed_extensions=ALL_ALLOWED_EXTENSIONS
+            patterns,
+            base_path=self.data_dir if self.data_dir else self.path,
```
lhoestq, line 753: same here

lhoestq (commented):
Awesome ! Yes that your change to `LocalDatasetModuleFactoryWithoutScript` and `HubDatasetModuleFactoryWithoutScript` are great thanks :)

Having `data_dir` consistent across the different factories is definitely welcome. I guess you can rename the PR to say that it also fixes the splits for the repositories without splits.

I left a few comments:

tests/test_load.py
polinaeterna: [23920a2](https://github.com/huggingface/datasets/pull/4144/commits/23920a290f53ba6845cc2bf880c00b8b9a09f0f5)
i've modified the test for `LocalDatasetModuleFactoryWithoutScript`

for `HubDatasetModuleFactoryWithoutScript` i think it's enough just to change checking if num files in each split is == 1 instead of > 0, i.e. only files from `data2` belong to dataset, we know in advance that there are one file per split (see dir structure in a [repo](https://huggingface.co/datasets/mariosasko/test_multi_dir_dataset/tree/main)

src/datasets/load.py
polinaeterna: [0840185](https://github.com/huggingface/datasets/pull/4144/commits/08401858ba7660d77b2f4a7bea3f8c55ef9a9931)

do you think there is a need to do `sanitize_patterns(str(Path(self.data_dir).resolve())` here or simply `sanitize_patterns(self.data_dir)` is enough, as you suggested? `sanitize_patterns(self.data_dir)` works correctly but then I don't see a need for using `str(Path().resolve()` in the next line, when self.data_dir is None.

src/datasets/load.py
polinaeterna: [6fb34ff](https://github.com/huggingface/datasets/pull/4144/commits/6fb34ff72fd05046897fd6a17399fd4d7709863e)

[0fa3cd8](https://github.com/huggingface/datasets/pull/4144/commits/0fa3cd8a4d7a6d8947479e4f4196cdadc2a27aa5)

src/datasets/load.py
polinaeterna: [6fb34ff](https://github.com/huggingface/datasets/pull/4144/commits/6fb34ff72fd05046897fd6a17399fd4d7709863e)

src/datasets/load.py
lhoestq: Both are fine, but using absolute paths with `Path(self.data_dir).resolve()` is usually good practice so feel free to add it. Relative paths sometimes cause unwanted behaviors, so let's use resolve them to absolute paths when we can

lhoestq (commented):
Thanks ! As discussed offline let's also have tests for the new `base_path` parameter in `data_files.py` for hub repositories, and we can merge !

polinaeterna:
@lhoestq to test passing subdirectory (`base_path`) to data_files functions and methods, I extended the temporary test directory with data so that it contains subdirectory. Because of that the number of files in this directory increased, so I had to change some numbers and patterns to account for this change - [907ddf0](https://github.com/huggingface/datasets/pull/4144/commits/907ddf09d3afece5afbae18675c859d6e453f2bf)

Do you think it's ok? Another option is to create another tmp dir and do all the checks inside it.

lhoestq (approved):
Thank you ! It looks all good to me :)

PR #4145, opened 2022-04-11
Redirect TIMIT download from LDC

lhoestq:
LDC data is protected under US copyright laws and under various legal agreements between the Linguistic Data Consortium/the University of Pennsylvania and data providers which prohibit redistribution of that data by anyone other than LDC. Similarly, LDC's membership agreements, non-member user agreement and various corpus-specific license agreements specifically state that users cannot publish, retransmit, disclose, copy, reproduce or redistribute LDC databases to others outside their organizations.

LDC explicitly asked us to remove the download script for the TIMIT dataset. In this PR I remove all means to download the dataset, and redirect users to download the data from https://catalog.ldc.upenn.edu/LDC93S1


lhoestq:
CI is failing because some tags are outdated, but they're fixed in #4067

patrickvonplaten (approved):
Thanks! This might break some tests in Transformers, but we can fix after the merge by using a different dataset I think 

Just FYI @LysandreJik @sgugger

lhoestq:
We may do a release pretty soon (today ?), let me know if it's fine to include it in the new release

patrickvonplaten:
Fine to include this change!

PR #4158, opened 2022-04-12
Add AUC ROC Metric

emibaylor:


lhoestq (approved):
Awesome thank you ! Also cc @sashavor @lvwerra since this one is a good example of a metric that has different input types depending on the multiclass/multilabel setting

sorry for the delay reviewing it btw, I'm also spending some time figuring out issues with the CI lately and it's taking a lot of time ^^'

PR #4161, opened 2022-04-13
Add Visual Genome

thomasw21:


thomasw21:
Hum there seems to be some issues with tasks in test:
 - some tasks don't fit anything in `tasks.json`. Do I remove them in `task_categories`?
 - some tasks should exist, typically `visual-question-answering` (https://github.com/huggingface/datasets/blame/9f2ff14673cac1f1ad56d80221a793f5938b68c7/src/datasets/utils/resources/tasks.json#L195) yet the exception is failing on me. I'm guessing it's because my `master` is not up-to-date. However this means that the testing only tests my branch instead of the one merged with master?
 
 cc @mariosasko @lhoestq

datasets/visual_genome/visual_genome.py
```diff
@@ -0,0 +1,465 @@
...
+
+# ---- Visual Genome loading script ----
+
+
+class VisualGenomeConfig(datasets.BuilderConfig):
+    """BuilderConfig for Visual Genome."""
+
+    def __init__(self, name: str, version: Optional[str] = None, with_image: bool = True, **kwargs):
+        _version = _LATEST_VERSIONS[name] if version is None else version
+        _name = f"{name}_v{_version}"
```
thomasw21, line 307: After much thought into this, I don't think this makes sense anymore. If preview/autotrain rely on only names it's probably the issue of those components no? I think we should have a single name that has a default version (1.2.0) in our case.

lhoestq:
> some tasks don't fit anything in tasks.json. Do I remove them in task_categories?

You can keep them, but add `other-` as a prefix to those tasks to make the CI ignore it

> some tasks should exist, typically visual-question-answering (https://github.com/huggingface/datasets/blame/9f2ff14673cac1f1ad56d80221a793f5938b68c7/src/datasets/utils/resources/tasks.json#L195) yet the exception is failing on me. I'm guessing it's because my master is not up-to-date. However this means that the testing only tests my branch instead of the one merged with master?

Feel free to merge upstream/master into your branch ;)

datasets/visual_genome/visual_genome.py
lhoestq: What do you mean ? I think it's fine this way ?

lhoestq (approved):
Impressive ! Thanks for adding this dataset, and good job with the dataset card and the dataset script ! 

Can you remove huggingface.jpg please ? I would find it confusing to have it inside the dataset repository. We can keep the dummy data script though

Feel free to merge when it's good for you

thomasw21:
After offline discussions: will keep that image essentially it's necessary as I have a mapping that creates a mapping between url and local path (images are downloaded via a zip file) and dummy data needs to store that dummy image. The issue is when I read an annotation, I get a url, compute the local path, and basically I assume the local path exists since I've extracted all the images ... This isn't true if dummy data doesn't have all the images, so instead I've added a script that "fixes" the dummy data after using the CLI, it essentially adds the dummy image in the zip corresponding to the url.

datasets/visual_genome/visual_genome.py
thomasw21: I really don't like the use of versions in the naming, and I'd rather have it as a config argument (I usually don't care about the version, I usually only care about about the task ...)

PR #4162, opened 2022-04-13
Add Conceptual 12M

thomasw21:


lhoestq:
Looks like your dummy_data.zip file is not in the right location ;)
datasets/datasets/conceptual_12m/dummy/default/0.0.0/dummy_data.zip
->
datasets/conceptual_12m/dummy/default/0.0.0/dummy_data.zip

datasets/conceptual_12m/conceptual_12m.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def __init__(self):
+        super(Conceptual12MConfig, self).__init__()
+
+
+class Conceptual12M(datasets.GeneratorBasedBuilder):
+    """Conceptual 12M dataset."""
+
+    BUILDER_CONFIG_CLASS = Conceptual12MConfig
+    BUILDER_CONFIGS = [Conceptual12MConfig()]
```
mariosasko, line 59: This is not needed if a dataset script has only one config (with the default name). 

(please also remove the `Conceptual12MConfig` class definition)

mariosasko (commented):
Two nits. Other than that looks great.

PR #4167, opened 2022-04-13
Avoid rate limit in update hub repositories

lhoestq:
use http.extraHeader to avoid rate limit


lhoestq:
I also set GIT_LFS_SKIP_SMUDGE=1 to speed up git clones

PR #4170, opened 2022-04-14
to_tf_dataset rewrite

Rocketknight1:
This PR rewrites almost all of `to_tf_dataset()`, which makes it kind of hard to list all the changes, but the most critical ones are:

- Much better stability and no more dropping unexpected column names (Sorry @NielsRogge)
- Doesn't clobber custom transforms on the data (Sorry @NielsRogge again)
- Much better handling of the situation when the `collate_fn` adds columns that aren't in the dataset.
- Better inference of shapes and data types
- Lots of hacky special-casing code removed
- Can return string columns (as `tf.String`)
- Most arguments have default values, calling the method should be much simpler
- Can accept a `model` argument and only return columns that are valid inputs to that model

I still have a couple of TODOs remaining and some testing to do, so don't merge yet, but it should be mostly ready for review at this point!

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -42,6 +42,7 @@ from typing import (
     Iterator,
     List,
     Optional,
+    Sequence,
     Tuple,
     Union,
     overload,
@@ -347,20 +348,17 @@ class TensorflowDatasetMixin:
 
         return tf_columns_to_signatures, np_columns_to_dtypes
 
-    def tf_autodetect_drop_columns(self):
-        pass  # TODO Automatically find non-numerical columns - Quentin said he'd look into it!
-
     def to_tf_dataset(
         self,
         model: Optional[Any] = None,  # Should be a Keras model but the type isn't imported
-        columns: Optional[Union[str, List[str]]] = None,
+        columns: Optional[Union[str, Sequence[str]]] = None,
         batch_size: int = 8,
         shuffle: bool = True,
         collate_fn: Callable = numpy_default_data_collator,
         drop_remainder: Optional[bool] = None,
-        drop_columns: Optional[Union[str, List[str]]] = None,
+        drop_columns: Optional[Union[str, Sequence[str]]] = None,
         collate_fn_args: Optional[Dict[str, Any]] = None,
-        label_cols: Optional[Union[str, List[str]]] = None,
+        label_cols: Optional[Union[str, Sequence[str]]] = None,
         prefetch: bool = True,
     ):
         """Create a tf.data.Dataset from the underlying Dataset. This tf.data.Dataset will load and collate batches from
@@ -385,8 +383,7 @@ class TensorflowDatasetMixin:
                 labels. Note that many models compute loss internally rather than letting Keras do it, in which case it is
                 not necessary to actually pass the labels here, as long as they're in the input `columns`.
             drop_columns (:obj:`List[str]` or :obj:`str` or :obj:`bool`, optional): Dataset column(s) to drop before
-                passing to the `collate_fn`. By default, non-numeric columns will be dropped. This can be set to `False`
-                 to keep all columns.
+                passing to the `collate_fn`. By default, no columns are dropped.
             prefetch (:obj:`bool`, default ``True``): Whether to run the dataloader in a separate thread and maintain
                 a small buffer of batches for training. Improves performance by allowing data to be loaded in the
                 background while the model is training.
@@ -444,8 +441,10 @@ class TensorflowDatasetMixin:
         if drop_remainder is None:
             # We assume that if you're shuffling it's the train set, so we drop the remainder unless told not to
             drop_remainder = shuffle
-        if drop_columns is None:
-            drop_columns = self.tf_autodetect_drop_columns()
+        if not drop_columns:
+            drop_columns = []
+        elif isinstance(drop_columns, str):
+            drop_columns = [drop_columns]
         if drop_columns:
             pre_collate_cols_to_retain = set(self.features.keys()) - set(drop_columns)
         else:
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -27,9 +27,11 @@ from collections.abc import Mapping
 from copy import deepcopy
 from dataclasses import asdict
 from functools import partial, wraps
+from inspect import signature
 from io import BytesIO
 from math import ceil, floor
 from pathlib import Path
+from random import sample
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -58,8 +60,8 @@ from tqdm.auto import tqdm
 from . import config
 from .arrow_reader import ArrowReader
 from .arrow_writer import ArrowWriter, OptimizedTypedSequence
-from .features import Audio, ClassLabel, Features, Image, Sequence, Value
-from .features.features import FeatureType, _ArrayXD, decode_nested_example, pandas_types_mapper, require_decoding
+from .features import Audio, ClassLabel, Features, Image, Value
+from .features.features import FeatureType, decode_nested_example, pandas_types_mapper, require_decoding
 from .filesystems import extract_path_from_uri, is_remote_filesystem
 from .fingerprint import (
     fingerprint_transform,
@@ -97,6 +99,43 @@ if TYPE_CHECKING:
     from .dataset_dict import DatasetDict
 
 logger = logging.get_logger(__name__)
+
+
+def numpy_default_data_collator(features: List) -> Dict[str, Any]:
+    # Copied from Transformers
+    # Putting this here for now because I don't want to add a dependency
+    import numpy as np
+
+    if not hasattr(features[0], "keys"):  # Check for dict-like
+        features = [vars(f) for f in features]
+    first = features[0]
+    batch = {}
+
+    # Special handling for labels.
+    # Ensure that tensor is created with the correct type
+    # (it should be automatically the case, but let's make sure of it.)
+    if "label" in first and first["label"] is not None:
+        label = first["label"].item() if isinstance(first["label"], np.ndarray) else first["label"]
+        dtype = np.int64 if isinstance(label, int) else np.float32
+        batch["labels"] = np.array([f["label"] for f in features], dtype=dtype)
+    elif "label_ids" in first and first["label_ids"] is not None:
+        if isinstance(first["label_ids"], np.ndarray):
+            batch["labels"] = np.stack([f["label_ids"] for f in features])
+        else:
+            dtype = np.int64 if type(first["label_ids"][0]) is int else np.float32
+            batch["labels"] = np.array([f["label_ids"] for f in features], dtype=dtype)
+
+    # Handling of all other possible keys.
+    # Again, we will use the first element to figure out which key/values are not None for this model.
+
+    for k, v in first.items():
+        if k not in ("label", "label_ids") and v is not None and not isinstance(v, str):
+            if isinstance(v, np.ndarray):
+                batch[k] = np.stack([f[k] for f in features])
+            else:
+                batch[k] = np.array([f[k] for f in features])
+
+    return batch
 
 
 class LazyDict(UserDict):
@@ -211,9 +250,18 @@ class TensorflowDatasetMixin:
     _TF_DATASET_REFS = set()
 
     @staticmethod
-    def _get_output_signature(dataset: "Dataset", collate_fn: Callable, collate_fn_args: dict, batch_size: int):
+    def _get_output_signature(
+        dataset: "Dataset",
+        cols_to_retain: List,
+        collate_fn: Callable,
+        collate_fn_args: dict,
+        batch_size: int,
+        num_test_batches: int = 10,
+    ):
         """Private method used by `to_tf_dataset()` to find the shapes and dtypes of samples from this dataset
-           after being passed through the collate_fn.
+           after being passed through the collate_fn. Tensorflow needs an exact signature for tf.numpy_function, so
+           the only way to do this is to run test batches - the collator may add or rename columns, so we can't figure
+           it out just by inspecting the dataset.
 
         Args:
             dataset (:obj:`Dataset`): Dataset to load samples from.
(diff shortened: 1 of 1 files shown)


src/datasets/arrow_dataset.py
```diff
@@ -331,66 +382,108 @@ def to_tf_dataset(
...
+                raise ValueError("Either columns or model must be specified.")
+            autodetected_columns = True
+            columns = list(signature(model.call).parameters.keys())
+            model_name = type(model).__name__
+            # This is the same code as Sylvain's find_labels function in Transformers, but I didn't want the dependency
+            if not label_cols:
+                if "QuestionAnswering" in model_name:
+                    label_cols = [p for p in columns if "label" in p or p in ("start_positions", "end_positions")]
+                else:
+                    label_cols = [p for p in columns if "label" in p]
```
sgugger, line 432: This is very specific to Transformers models here, so you should check if the model is an instance of `TFPreTrainedModel` before doing the guessing.

sgugger (commented):
The fact your changes make this very entangled with Transformers code makes me think you are trying to do too much here. In particular the code for guessing the labels or the numpy data collator from Transformers are very aimed at Transformer models, and the method in Datasets would be expected to work on any model. Copy-pasting code from Transformers is kind of a red flag ;-)

I think we need to rethink the design and make the base `to_tf_dataset` more basic and less magic, then have another method in Transformers that will do the magic guessing of defaults using the model (and use a different default collate).

src/datasets/arrow_dataset.py
Rocketknight1: That's a very good point, yes! It's tricky, though, because these things are handled by `Trainer` for PyTorch users, where we have a lot of scope to add custom code, but if we want to use Keras for TF then these things have to be handled at the point of dataset creation. Maybe we could move this from a method on `Dataset` objects to a function in `transformers`, like `convert_to_tf_dataset()` that takes a HF `Dataset` as an argument?

sgugger: See my comment above.

Rocketknight1: Ah sorry, yes! Should have paid more attention to the second paragraph.

sgugger: No worries ;-)

Rocketknight1:
[Magic is now banned](https://www.youtube.com/watch?v=WIn58XoY728#t=36s) by decree of @sgugger. This is honestly much cleaner, and the functionality will make much more sense in `transformers` anyway!

src/datasets/arrow_dataset.py
Rocketknight1: Fixed!

src/datasets/arrow_dataset.py
```diff
@@ -331,66 +345,110 @@ def to_tf_dataset(
...
+            # Set a very simple default collator that just stacks things together
+            def default_collate_fn(features):
+                first = features[0]
+                batch = {}
+                for k, v in first.items():
+                    if isinstance(v, np.ndarray):
+                        batch[k] = np.stack([f[k] for f in features])
+                    else:
+                        batch[k] = np.array([f[k] for f in features])
+                return batch
```
sgugger, line 385: Can be defined outside of this function maybe?

sgugger (commented):
Better for me! Are there any real changes in the notebook? All I see in the diff is Pycharm adding useless metadata.

src/datasets/arrow_dataset.py
Rocketknight1: Sure, I'll move this! Also, the one change in the notebook is that I removed the `dummy_labels` arg in the calls to `to_tf_dataset()`, the rest is just useless metadata, yes.

Rocketknight1: Moved it to a new `tf_utils` file, protected by a check to see if TF is available.

gante (commented):
Looks good! Added a few minor comments, and I'd like to make a request as well: since we now have default arguments for everything, can we add a test case with no passed arguments?

(I don't want to approve because I don't have much `datasets` knowledge :) )

src/datasets/arrow_dataset.py
```diff
@@ -226,99 +237,102 @@ def _get_output_signature(dataset: "Dataset", collate_fn: Callable, collate_fn_a
...
-        shuffle: bool,
-        collate_fn: Callable,
-        drop_remainder: bool = None,
-        collate_fn_args: Dict[str, Any] = None,
-        label_cols: Union[str, List[str]] = None,
-        dummy_labels: bool = False,
+        columns: Optional[Union[str, Sequence[str]]] = None,
+        batch_size: int = 8,
+        shuffle: bool = True,
+        collate_fn: Callable = None,
```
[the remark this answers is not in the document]
Rocketknight1: Fixed!

src/datasets/arrow_dataset.py
```diff
@@ -226,99 +237,102 @@ def _get_output_signature(dataset: "Dataset", collate_fn: Callable, collate_fn_a
...
+        error_on_missing: bool = True,
     ):
         """Create a tf.data.Dataset from the underlying Dataset. This tf.data.Dataset will load and collate batches from
         the Dataset, and is suitable for passing to methods like model.fit() or model.predict().
 
         Args:
-            columns (:obj:`List[str]` or :obj:`str`): Dataset column(s) to load in the tf.data.Dataset. In general,
-                only columns that the model can use as input should be included here (numeric data only).
+            model (:obj:`tf.keras.Model`, optional): A Keras model object. Used to autodetect which columns to return.
+                Does nothing if `columns` is also passed.
```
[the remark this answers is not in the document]
Rocketknight1: Fixed!

src/datasets/arrow_dataset.py
```diff
@@ -226,99 +237,102 @@ def _get_output_signature(dataset: "Dataset", collate_fn: Callable, collate_fn_a
...
+            columns (:obj:`List[str]` or :obj:`str`, optional): Dataset column(s) to load in the tf.data.Dataset. It is acceptable
+                to include column names that are created by the `collate_fn` and that do not exist in the original
+                dataset.
             batch_size (:obj:`int`): Size of batches to load from the dataset.
```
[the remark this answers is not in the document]
Rocketknight1: Fixed!

src/datasets/arrow_dataset.py
```diff
@@ -331,66 +345,110 @@ def to_tf_dataset(
...
-            dummy_labels (:obj:`bool`, default ``False``): If no `label_cols` are set, output an array of "dummy" labels
-                with each batch. This can avoid problems with `fit()` or `train_on_batch()` that expect labels to be
-                a Tensor or np.ndarray, but should (hopefully) not be necessary with our standard train_step().
+            drop_columns (:obj:`List[str]` or :obj:`str` or :obj:`bool`, optional): Dataset column(s) to drop before
+                passing to the `collate_fn`. By default, no columns are dropped.
             prefetch (:obj:`bool`, default ``True``): Whether to run the dataloader in a separate thread and maintain
                 a small buffer of batches for training. Improves performance by allowing data to be loaded in the
                 background while the model is training.
+            error_on_missing (:obj:`bool`, default ``True``): Whether to return an error if one of the requested columns
+                is missing. Should be set to `False` if columns or label_cols are autodetected from a model.
```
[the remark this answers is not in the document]
Rocketknight1: I'm leaving this bit, even though we're not passing the model anymore, because it's advice on what to do if the autodetection happened in another function!

Rocketknight1:
@gante I renamed the default collator to `minimal_tf_collate_fn`!

src/datasets/arrow_dataset.py
Rocketknight1: Renamed!

src/datasets/arrow_dataset.py
```diff
@@ -226,171 +238,205 @@ def _get_output_signature(dataset: "Dataset", collate_fn: Callable, collate_fn_a
...
+                post_collate_cols_to_retain = columns
+            else:
+                post_collate_cols_to_retain = list(set(columns + label_cols))
         else:
-            cols_to_retain = columns
-        if "label" in cols_to_retain or "labels" in cols_to_retain or "label_ids" in cols_to_retain:
-            cols_to_retain += ["labels", "label", "label_ids"]  # Don't accidentally drop any labels with other names!
-        cols_to_retain = list(set(cols_to_retain))  # Remove any duplicates
-
+            post_collate_cols_to_retain = None
```
lhoestq, line 394: _get_output_signature expects a list of columns, not None

src/datasets/arrow_dataset.py
```diff
@@ -226,171 +238,205 @@ def _get_output_signature(dataset: "Dataset", collate_fn: Callable, collate_fn_a
...
+            elif np_arrays[0].dtype.kind == "U":  # Unicode strings
+                np_dtype = np.unicode_
+                tf_dtype = tf.string
             else:
-                # If this doesn't look like LM labels that got added by the collate_fn, let's not say anything
-                # about the dimensions we're unsure of
-                shape = [batch_size] + [None for dim in tensor.shape.as_list()[1:]]
-            signatures[column] = tf.TensorSpec(shape=shape, dtype=tensor.dtype)
-        return columns_to_dtypes, signatures
+                raise RuntimeError(f"Unrecognized array dtype {np_arrays[0].dtype}!")
```
lhoestq, line 296: Just FYI this error is currently raised if it's a nested type (dtype=object), which happens for audio datasets that have an "audio" column where each item looks like
```
{"array"::..., "sampling_rate":...., "path":>>>}
```
Let's discuss that in another PR after this one if you want ;)

lhoestq (commented):
Awesome thank you !

I added a few comments, and also found a bug when you don't specify the `columns` explicitly (a None is passed to _get_output_signature).

Feel free to also add some tests to make sure that the new default behavior works as expected !

src/datasets/arrow_dataset.py
Rocketknight1: Sorry, forgot to add the logic there. Fixing!

src/datasets/arrow_dataset.py
Rocketknight1: Yeah, that might need a separate PR! Right now I want to push this one over the line - I'm just focusing on how to handle column autodetection so users don't have to pass big awkward lists of arguments all the time, and then it should be ready to go.

src/datasets/arrow_dataset.py
```diff
@@ -226,171 +238,206 @@ def _get_output_signature(dataset: "Dataset", collate_fn: Callable, collate_fn_a
...
+                if dim == 0:
+                    static_shape.append(batch_size)
+                    continue
+                if len(sizes) == 1:  # This dimension looks constant
+                    static_shape.append(sizes.pop())
+                else:  # Use None for variable dimensions
+                    static_shape.append(None)
+            tf_columns_to_signatures[column] = tf.TensorSpec(shape=static_shape, dtype=tf_dtype)
+            np_columns_to_dtypes[column] = np_dtype
+        # endregion
```
lhoestq, line 311: This sounds specific to pycharm, can you remove them please ?

notebooks/Overview.ipynb
```diff
@@ -2630,19 +2827,21 @@
     "    return tokenizer.pad(examples, return_tensors='np')\n",
     "\n",
     "# to_tf_dataset() returns a tf.data.Dataset that we can pass straight to model.fit().\n",
-    "# Note the use of dummy_labels - this isn't always necessary, but can stop TF getting confused when you don't pass any labels.\n",
     "encoded_tf_dataset = encoded_dataset['train'].to_tf_dataset(\n",
     "    columns=columns,\n",
     "    collate_fn=collate_fn,\n",
     "    batch_size=8,\n",
     "    shuffle=True,\n",
-    "    dummy_labels=True\n",
```
lhoestq, line 2639: It was pretty hard to find these little changes in this big diff xD

src/datasets/utils/tf_utils.py
```diff
@@ -0,0 +1,39 @@
...
+            if isinstance(v, np.ndarray):
+                batch[k] = np.stack([f[k] for f in features])
+            elif isinstance(v, tf.Tensor):
+                batch[k] = np.stack([f[k].numpy() for f in features])
+            else:
+                batch[k] = np.array([f[k] for f in features])
+        return batch
+
+else:
+    minimal_tf_collate_fn = None
```
lhoestq, line 39: I guess you don't need to check `if config.TF_AVAILABLE` anymore and just define the function directly:

```suggestion
def minimal_tf_collate_fn(features):
    if config.TF_AVAILABLE:
        import tensorflow as tf
    else:
        raise ImportError("Called a Tensorflow-specific function but Tensorflow is not installed.")

    first = features[0]
    batch = {}
    for k, v in first.items():
        if isinstance(v, np.ndarray):
            batch[k] = np.stack([f[k] for f in features])
        elif isinstance(v, tf.Tensor):
            batch[k] = np.stack([f[k].numpy() for f in features])
        else:
            batch[k] = np.array([f[k] for f in features])
    return batch
```

src/datasets/arrow_dataset.py
```diff
@@ -226,171 +238,206 @@ def _get_output_signature(dataset: "Dataset", collate_fn: Callable, collate_fn_a
...
+                post_collate_cols_to_retain = columns
+            else:
+                post_collate_cols_to_retain = list(set(columns + label_cols))
         else:
-            cols_to_retain = columns
-        if "label" in cols_to_retain or "labels" in cols_to_retain or "label_ids" in cols_to_retain:
-            cols_to_retain += ["labels", "label", "label_ids"]  # Don't accidentally drop any labels with other names!
-        cols_to_retain = list(set(cols_to_retain))  # Remove any duplicates
-
+            post_collate_cols_to_retain = None  # Indicates keeping all columns
```
lhoestq, line 395: Later in the code you do `set(post_collate_cols_to_retain)`, which fails when it's None

lhoestq (commented):
I think it's still not working, see my comment. Can you please add some tests ?

src/datasets/arrow_dataset.py
Rocketknight1: This also works in several other IDEs like VSCode! Do you want me to remove it anyway?

gante: TBH, I also find the regions comments aesthetically unpleasing 🙈

Rocketknight1: Fiiiine. But look how good it can be! 
image

src/datasets/arrow_dataset.py
Rocketknight1: Fixed!

Rocketknight1:
@lhoestq @sgugger @gante 

I think this should now be ready, it looks good in testing! I'll try a few more notebooks today and tomorrow to be sure before I merge. Key changes are:

- No column autodetection magic (will make a separate PR to add this as a `transformers` function)
- Drops non-numerical features automatically (this is more of a 'DataLoader' method, we'll have a separate method to expose 'raw' datasets to `tf.data`)
- Better autodetection of numerical features.
- Shouldn't randomly crash mid-function :skull: 

We definitely have some questions still to resolve about how to handle making a 'DataLoader' dataset versus a 'raw' dataset - see [the Notion doc](https://www.notion.so/huggingface2/Splitting-to_tf_dataset-c2e0773c4bec484384064b30ed634383) if you're interested. Still, since this PR is just fixes/improvements to an existing method which never supported non-numerical features anyway, we can merge it before we've resolved those issues, and then think about how to name and split things afterwards.

src/datasets/utils/tf_utils.py
```diff
@@ -0,0 +1,62 @@
...
+    else:
+        raise ImportError("Called a Tensorflow-specific function but Tensorflow is not installed.")
+
+    first = features[0]
+    batch = {}
+    for k, v in first.items():
+        if isinstance(v, np.ndarray):
+            batch[k] = np.stack([f[k] for f in features])
+        elif isinstance(v, tf.Tensor):
+            batch[k] = np.stack([f[k].numpy() for f in features])
```
gante, line 35:
```suggestion
            batch[k] = tf.stack([f[k] for f in features])
```
(Or is there a particular reason to use numpy's stack? If the goal is to have a numpy object, we can call `.numpy()` after stacking, which should be faster :D)

Screenshot 2022-04-27 at 17 06 28

gante (commented):
Looks good! 👍 

Can we also add a test case for `.to_tf_dataset()`, i.e. using defaults for all arguments?

src/datasets/utils/tf_utils.py
Rocketknight1: You're right, this was an oversight. Thanks!

lhoestq:
> Drops non-numerical features automatically

Can you give more details on how this work ? This is not explained in the docs

Also why are you adding `error_on_missing` and `auto_fix_label_names ` ? The rationale is not clear to me. In particular I think it is sensible enough to expect users to not ask columns that don't exist, to rename a label column when required.

Rocketknight1:
@lhoestq I rewrote those parts - they were causing some other issues too! `error_on_missing` and `auto_fix_label_names` have been removed. The new logic is to simply drop (before batch collation) all columns the user doesn't ask for, but not to raise errors if the user asked for columns not in the dataset, as they may be added by the collator. Hopefully this cleans it up and matches the documentation better!

src/datasets/arrow_dataset.py
```diff
@@ -228,171 +239,201 @@ def _get_output_signature(dataset: "Dataset", collate_fn: Callable, collate_fn_a
...
+            # Following the logic in `transformers.Trainer`, we do not drop `label_ids` or `label` even if they
+            # are not in the list of requested columns, because the collator may rename them
+            # This might work better if moved to a method attached to our transformers Model objects, but doing so
+            # could break backward compatibility
+            unwanted_columns = [
+                col
+                for col in self.features.keys()
+                if col not in columns and col not in label_cols and col not in ("label_ids", "label")
+            ]
+            dataset = dataset.remove_columns(unwanted_columns)
```
Rocketknight1, line 409: One thing I'm still not sure about is that we special-case "label_ids" and "label", because those columns are frequently renamed to "labels" in `transformers` and we don't want to drop them. In PyTorch this special-casing occurs too, but it happens inside `Trainer`, so it doesn't mix things up between the `transformers` and `datasets` libraries. We might have to leave it here for now until I can make the 'magic method' on our `transformers` models!

Rocketknight1: The standard workflow for `transformers` is that the column in the dataset is called `label`, but the argument to the model is called `labels`, and the renaming is done by the `collate_fn`. Therefore, we must be careful not to drop columns called `label` or `label_ids`, even if the user doesn't request them, because the `collate_fn` might need them.

In PyTorch columns are dropped [by the Trainer](https://github.com/huggingface/transformers/blob/main/src/transformers/trainer.py#L610), and it makes sure not to drop these. However, in Tensorflow, columns are dropped by the `to_tf_dataset()` method, and therefore this code needs to be in there.

I think a good solution would be to make a method in `transformers` that calls `to_tf_dataset()` and converts datasets for training, and then we could move the `label` and `label_ids` special casing in there. However, until we do that, we'll need to keep the special casing in `to_tf_dataset()` or all our examples will break!

src/datasets/arrow_dataset.py
```diff
@@ -440,25 +476,24 @@ def split_features_and_labels(input_batch):
 
             tf_dataset = tf_dataset.map(split_features_and_labels)
 
-        elif len(columns) == 1:
+        elif isinstance(tf_dataset.element_spec, dict) and len(tf_dataset.element_spec) == 1:
```
lhoestq, line 479: (nit)

Not a big fan of changing the return type of a function based on the content of a dataset. I'd rather make the output type deterministic based on the input parameters only.

What if we do that only if the type of the `columns` argument of `to_tf_dataset` is a string, and if the user passes a list with a single element, then it should stay a dict with a single element ?

lhoestq (commented):
Cool, let me know if you need help for the tests :)

I think we need several additional tests, especially since the function is quite long and not trivial to maintain. In particular it would be nice to check the default behavior, and check the behavior of every single parameter.

src/datasets/arrow_dataset.py
Rocketknight1: I think that's a good idea - it would definitely have been better to do it that way, but it might break backward compatibility for some of our existing examples if we change now. Let me do some testing and see how bad it'd be!

Rocketknight1: Update: There are definitely some issues with this change, but my updates to train_step and test_step in another PR fix a lot of it. Can we leave it for now (since this method already exists, it's not like this PR makes it worse!), and maybe deprecate or change the behaviour in a follow-up PR?

lhoestq: Yup sounds good :)

Rocketknight1: Sure! I'll work on adding the tests you suggested and then ping you for another review.

Rocketknight1:
@lhoestq New tests are now in!

lhoestq:
Thanks for the ping, will take a look tomorrow :)

Maybe the rebase didn't go well for the code recently merged about label alignment from https://github.com/huggingface/datasets/pull/4277 ?

Rocketknight1:
It's very strange! The rebase looks fine to me. I might try to move my changes to a new branch from `master` and see if I can figure out which change causes this problem to appear.

Rocketknight1:
@lhoestq Got it! It was caused by a name collision - I was importing `typing.Sequence`, but the code also needed `features.Sequence`. The tests from that PR failed were expecting the latter but got the former, and then crashed.

src/datasets/arrow_dataset.py
```diff
@@ -228,114 +238,107 @@ def _get_output_signature(dataset: "Dataset", collate_fn: Callable, collate_fn_a
...
         """Create a tf.data.Dataset from the underlying Dataset. This tf.data.Dataset will load and collate batches from
         the Dataset, and is suitable for passing to methods like model.fit() or model.predict().
 
         Args:
-            columns (:obj:`List[str]` or :obj:`str`): Dataset column(s) to load in the tf.data.Dataset. In general,
-                only columns that the model can use as input should be included here (numeric data only).
+            columns (:obj:`List[str]` or :obj:`str`, optional): Dataset column(s) to load in the tf.data.Dataset. Column
+             names that are created by the `collate_fn` and that do not exist in the original dataset can be used.
+             If this argument is not set, we retain all non-numerical columns.
             batch_size (:obj:`int`): Size of batches to load from the dataset.
```
lhoestq, line 330: Can you update the docstring with all the new default values ?

src/datasets/arrow_dataset.py
```diff
@@ -228,114 +238,107 @@ def _get_output_signature(dataset: "Dataset", collate_fn: Callable, collate_fn_a
...
-        columns: Union[str, List[str]],
-        batch_size: int,
-        shuffle: bool,
-        collate_fn: Callable,
-        drop_remainder: bool = None,
-        collate_fn_args: Dict[str, Any] = None,
-        label_cols: Union[str, List[str]] = None,
-        dummy_labels: bool = False,
+        columns: Optional[Union[str, List[str]]] = None,
+        batch_size: int = 8,
```
lhoestq, line 315: Related to our discussions about the scope of this function, I think it should return a bare-bone tf.data.Dataset by default, and if users want to they can use it as their data loader (e.g. with batching and labels)

Therefore I was wondering what you think about  **not** batching by default ?

src/datasets/arrow_dataset.py
```diff
@@ -228,114 +238,107 @@ def _get_output_signature(dataset: "Dataset", collate_fn: Callable, collate_fn_a
             batch_size (:obj:`int`): The size of batches loaded from the dataset. Used for shape inference.
```
lhoestq, line 238: docstrings need to be updated here as well

src/datasets/arrow_dataset.py
```diff
@@ -354,56 +357,89 @@ def to_tf_dataset(
...
+        # are non-numeric!
+
+        # endregion
+
+        # region Compute dataset signature and verify columns
+        output_signature, columns_to_np_types = dataset._get_output_signature(
+            dataset,
+            collate_fn=collate_fn,
+            collate_fn_args=collate_fn_args,
+            batch_size=batch_size if drop_remainder else None,
```
lhoestq, line 430: why does `drop_remainder` matter here ?

lhoestq (commented):
Oh good catch ! And thanks for the tests :)

I think we just need to update the docstring, and I also added a question about the default `batch_size` (sorry for not raising the question earlier, I missed it)

src/datasets/arrow_dataset.py
Rocketknight1: Done!

src/datasets/arrow_dataset.py
Rocketknight1: If we disable `drop_remainder` then the dataset will output batches with variable length on the batch dimension. Therefore, the dataset's output shape should display that dimension as `None`. If we do use `drop_remainder` then that dimension will be constant, so we can include it in the dataset output shape.

src/datasets/arrow_dataset.py
Rocketknight1: Done!

src/datasets/arrow_dataset.py
Rocketknight1: I think we can explore a method that does that, but it shouldn't go in this PR! There are a number of challenges around parallel access + batching in `tf.data` that I'll need to explore, and I think performance could suffer a lot if I'm not careful. I do agree that eventually we want a unified system that can emit raw `tf.data.Dataset` objects, though.

lhoestq: Then maybe let's not have a default value for `batch_size` until we can implement batch_size=None ? Otherwise we will have to break the default behavior of `batch_size` from  `batch_size=8` to `batch_size=None`

src/datasets/arrow_dataset.py
Rocketknight1: Sure!

src/datasets/arrow_dataset.py
Rocketknight1: That's a good idea, will do!

src/datasets/arrow_dataset.py
Rocketknight1: Done!

src/datasets/arrow_dataset.py
Rocketknight1: Done!

src/datasets/arrow_dataset.py
```diff
@@ -225,117 +235,111 @@ def _get_output_signature(dataset: "Dataset", collate_fn: Callable, collate_fn_a
...
-        shuffle: bool,
-        collate_fn: Callable,
-        drop_remainder: bool = None,
-        collate_fn_args: Dict[str, Any] = None,
-        label_cols: Union[str, List[str]] = None,
-        dummy_labels: bool = False,
+        columns: Optional[Union[str, List[str]]] = None,
+        shuffle: bool = True,
+        collate_fn: Optional[Callable] = None,
+        drop_remainder: Optional[bool] = None,
```
lhoestq, line 319: I think it should be False by default, as `tf.data.Dataset.batch` does.

lhoestq (commented):
Thank you ! I added more comments about the default for `shuffle` and `drop_remainder`, and some suggestions:

src/datasets/arrow_dataset.py
Rocketknight1: If `shuffle` is `False` by default then `drop_remainder` will also follow it, because the default setting for `drop_remainder` is to use the same value as `shuffle`. The reason for that is that you usually only shuffle the training set, which is also the dataset where dropping the remainder batch is common.

lhoestq: Again by default we shouldn't so extra actions rather than what the user asked for, i.e. convert the dataset to a tf dataset.

Even tf.data.Dataset doesn't drop the remainder by default:
https://www.tensorflow.org/api_docs/python/tf/data/Dataset?version=nightly#batch

Your concern makes total sense but is specific to one use case. Feel free to explain it in the docstring though

Rocketknight1: That's fair! We might encounter some backward-compatibility problems if we changed it now, though.

I'm hoping we can move a lot of the more `transformers`-specific stuff to a method in that library, which would become the 'standard' conversion method for users. At that point, we could make these compatibility-breaking changes to `to_tf_dataset()` and make it more like the pure function it should have been in the beginning. Do you think that's okay?

lhoestq: Which backward compatibility problems ? Is there a place somewhere that requires XLA and expects to_tf_dataset to ignore some data just for XLA ?

Rocketknight1: Keras uses XLA whenever `jit_compile=True`, which a lot of users use by default, so we might see some issues there. I haven't tested it in detail - do you want me to run some tests, or are you okay leaving it unchanged for now?

lhoestq: It goes against the main rationale for this function so yea I'd like to discuss it more. I don't think testing really matters - XLA users should know that they should drop the remainder and asks for this explicitly since this is goes beyond a simple conversion from HF dataset to TF dataset.

My main concern now is is there a place somewhere when we say something like "use to_tf_dataset() without specifying drop_remainder=True explicitly and you can expect the best performance using XLA". Because if there is, this is an issue because it means this function has a bad API. And if not, then we can consider making the behavior of `drop_remainder` consistent with the rationale of this function

Rocketknight1: In general, we don't refer to XLA much in the docs, so I don't think we say anything like that explicitly. However, if we change this then users whose code has `jit_compile=True` may suddenly stop working after the patch. I'm willing to make the change if you think it's necessary, though!

Rocketknight1: After testing, I think most cases still work okay with `drop_remainder` defaulting to False, so I don't object to changing it anymore!

lhoestq (commented):
Cool ! This is a HUGE step !! Thanks a lot @Rocketknight1 :)

Here are my final comments, then I think we can merge 🚀

tests/test_arrow_dataset.py
```diff
@@ -2304,61 +2304,58 @@ def test_with_transform(self, in_memory):
...
             self.assertEqual(batch.shape.as_list(), [4])
             self.assertEqual(batch.dtype.name, "int64")
+        with self._create_dummy_dataset(in_memory, tmp_dir.name, multiple_columns=True) as dset:
+            # Check that it works with all default options (except batch_size because the dummy dataset only has 4)
+            tf_dataset = dset.to_tf_dataset(batch_size=4)
+            batch = next(iter(tf_dataset))
+            self.assertEqual(batch["col_1"].shape.as_list(), [4])
+            self.assertEqual(batch["col_2"].shape.as_list(), [4])
+            self.assertEqual(batch["col_1"].dtype.name, "int64")
+            self.assertEqual(batch["col_2"].dtype.name, "string")  # Assert that we're converting strings properly
```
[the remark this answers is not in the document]
Rocketknight1: Done!

src/datasets/arrow_dataset.py
```diff
@@ -225,117 +235,115 @@ def _get_output_signature(dataset: "Dataset", collate_fn: Callable, collate_fn_a
...
+
+        # region Generate test batches
+        test_batches = []
+        for _ in range(num_test_batches):
+            indices = sample(range(len(dataset)), test_batch_size)
+            test_batch = dataset[indices]
+            test_batch = [{key: value[i] for key, value in test_batch.items()} for i in range(test_batch_size)]
+            test_batch = collate_fn(test_batch, **collate_fn_args)
+            test_batches.append(test_batch)
+        # endregion
```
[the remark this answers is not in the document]
Rocketknight1: Done!

Rocketknight1:
@lhoestq Thanks! Also, when you're ready, don't merge it immediately! I'd like to do a quick round of manual testing with the very final build once you're happy to make sure it still works in our notebooks and examples.

lhoestq (approved):
Thanks ! Feel free to run all your tests and merge yourself and you're good :)

Rocketknight1:
@lhoestq Tests look good to me, merging now!

Issue #4139, opened 2022-04-11
Dataset viewer issue for Winoground


alcinos:
## Dataset viewer issue for 'Winoground'

**Link:** [*link to the dataset viewer page*](https://huggingface.co/datasets/facebook/winoground/viewer/facebook--winoground/train)

*short description of the issue*
Getting 401, message='Unauthorized'
The dataset is subject to authorization, but I can access the files from the interface, so I assume I'm granted to access it. I'd assume the permission somehow doesn't propagate to the dataset viewer tool.

Am I the one who added this dataset ? No

odellus:
I'm getting an error when I `load_dataset('facebook/winoground', use_auth_token=tok)`. Looks like there's some ISO-8859-1 encoding in the JSON for winograd's dataset.
```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
```
Probably related to why the dataset viewer isn't working.

severo:
related (same dataset): https://github.com/huggingface/datasets/issues/4149. But the issue is different. Looking at it

odellus:
I thought this issue was related to the error I was seeing, but upon consideration I'd think the dataset viewer would return a 500 (unable to create the split like me) or a 404 (unable to load split b/c it was never created) error if it was having the issue I was seeing in #4149. 401 message makes it look like dataset viewer isn't passing through the identity of the user who has signed the licensing agreement when making the request to GET [examples.jsonl](https://huggingface.co/datasets/facebook/winoground/resolve/a86a60456fbbd242e9a744199071a6bd3e7fd9de/examples.jsonl).

severo:
To replicate:

```python
>>> import datasets
>>> dataset= datasets.load_dataset('facebook/winoground', name='facebook--winoground', split='train', streaming=True, use_token="hf_app_...")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/slesage/hf/datasets-preview-backend/.venv/lib/python3.9/site-packages/datasets/load.py", line 1660, in load_dataset
    builder_instance = load_dataset_builder(
  File "/home/slesage/hf/datasets-preview-backend/.venv/lib/python3.9/site-packages/datasets/load.py", line 1486, in load_dataset_builder
    dataset_module = dataset_module_factory(
  File "/home/slesage/hf/datasets-preview-backend/.venv/lib/python3.9/site-packages/datasets/load.py", line 1238, in dataset_module_factory
[...]

PR #4173, opened 2022-04-14
Stream private zipped images

lhoestq:
As mentioned in https://github.com/huggingface/datasets/issues/4139 it's currently not possible to stream private zipped images from the Hub.

This is because `Image.decode_example` does not handle authentication. Indeed decoding requires to access and download the file from the private repository.

In this PR I added authentication to `Image.decode_example` via a `token_per_repo_id` optional argument. I first wanted to just pass `use_auth_token` but a single `Image` instance can be responsible of decoding images from a combination several datasets together (from `interleave_datasets` for example). Therefore I just used a dictionary `repo_id` -> `token` instead.

### DIFF
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -212,7 +212,7 @@ class DatasetBuilder:
         info: Optional[DatasetInfo] = None,
         features: Optional[Features] = None,
         use_auth_token: Optional[Union[bool, str]] = None,
-        namespace: Optional[str] = None,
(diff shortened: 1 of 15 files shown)


lhoestq:
oops looks like some tests are failing sorry, will fix them tomorrow

lhoestq:
cc @mariosasko this is ready for review, let me know what you think !

mariosasko (approved):
Thanks for fixing this!

PR #4174, opened 2022-04-15
Fix when map function modifies input in-place

thomasw21:
When `function` modifies input in-place, the guarantee that columns in `remove_columns` are contained in `input` doesn't hold true anymore. Therefore we need to relax way we pop elements by checking if that column exists.


lhoestq (approved):
Good catch ! and thanks for adding a test :)

PR #4177, opened 2022-04-18
Adding missing subsets to the dataset

micahcarroll:
This dataset for the [1st task of SemEval-2018](https://competitions.codalab.org/competitions/17751) competition was missing all subtasks except for subtask 5. I added another subtask (subtask 2), which is comprised of 12 additional data subsets: for each language in En, Es, Ar, there are 4 datasets, broken down by emotions (anger, fear, joy, sadness).

## Remaining questions

I wasn't able to find any documentation about how one should make PRs to modify datasets. Because of that, I just did my best to integrate the new data into the code, and tested locally that this worked. I'm sorry if I'm not respecting your contributing guidelines – if they are documented somewhere, I'd appreciate if you could send a pointer!

Not sure how `dataset_infos.json` and `dummy` should be updated. My understanding is that they were automatically generated at the time of the original dataset creation?

### DIFF
--- a/datasets/sem_eval_2018_task_1/README.md
+++ b/datasets/sem_eval_2018_task_1/README.md
@@ -77,7 +77,7 @@ Here, E refers to emotion, EI refers to emotion intensity, V refers to valence o
 
 Together, these tasks encompass various emotion and sentiment analysis tasks. You are free to participate in any number of tasks and on any of the datasets.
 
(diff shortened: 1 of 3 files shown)


osbm:
Datasets are not tracked in this repository anymore. You should move this PR to the [discussions page of this dataset](https://huggingface.co/datasets/sem_eval_2018_task_1/discussions)

PR #4178, opened 2022-04-19
[feat] Add ImageNet dataset

apsdehal:
To use the dataset download the tar file
[imagenet_object_localization_patched2019.tar.gz](https://www.kaggle.com/competitions/imagenet-object-localization-challenge/data?select=imagenet_object_localization_patched2019.tar.gz) from Kaggle and then point the datasets library to it by using:

```py

dataset = load_dataset("imagenet",
data_dir="/path/to/imagenet_object_localization_patched2019.tar.gz")
```

Currently train and validation splits are supported.


datasets/imagenet/imagenet.py
```diff
@@ -0,0 +1,148 @@
...
+ILSVRC 2012, commonly known as 'ImageNet' is an image dataset organized according to the WordNet hierarchy. Each meaningful concept in WordNet, possibly described by multiple words or word phrases, is called a "synonym set" or "synset". There are more than 100,000 synsets in WordNet, majority of them are nouns (80,000+). In ImageNet, we aim to provide on average 1000 images to illustrate each synset. Images of each concept are quality-controlled and human-annotated. In its completion, we hope ImageNet will offer tens of millions of cleanly sorted images for most of the concepts in the WordNet hierarchy.
+"""
+
+_VAL_PREP_SCRIPT = "https://raw.githubusercontent.com/soumith/imagenetloader.torch/master/valprep.sh"
+_DEFAULT_CONFIG_NAME = "2012"
+_NUM_IMAGENET1K_CLASSES = 1000
+
+
+class ImagenetConfig(datasets.BuilderConfig):
+    def __init__(self, name=_DEFAULT_CONFIG_NAME, num_classes=_NUM_IMAGENET1K_CLASSES, **kwargs):
```
apsdehal, line 32: There is a bug in datasets where if we pass `data_dir` to `load_dataset` it gets added to `config_kwargs` ultimately leading to a new config build as per https://github.com/huggingface/datasets/blob/cd3ce34ab1604118351e1978d26402de57188901/src/datasets/builder.py#L362. Is there a way to fix this so I can avoid having defaults and then also pop the `"version"` from the `kwargs`?

datasets/imagenet/imagenet.py
```diff
@@ -0,0 +1,148 @@
...
+            )
+        val_prep_script = dl_manager.download(_VAL_PREP_SCRIPT)
+        images_path = os.path.join(data_dir, "ILSVRC", "Data", "CLS-LOC")
+
+        self._prep_validation(images_path, val_prep_script)
+
+        label_classes = os.listdir(os.path.join(images_path, "train"))
+
+        self.info.features = datasets.Features(
+            {"image": datasets.Image(), "labels": datasets.ClassLabel(names=sorted(label_classes))}
```
apsdehal, line 113: Is there a way we can make labels optional? this will allow me to add `test` set to this.

datasets/imagenet/imagenet.py
```diff
@@ -0,0 +1,148 @@
...
+                name=datasets.Split.VALIDATION,
+                gen_kwargs={"files": dl_manager.iter_files(os.path.join(images_path, "val"))},
+            ),
+            # datasets.SplitGenerator(
+            #     name=datasets.Split.TEST,
+            #     gen_kwargs={"files": dl_manager.iter_files(os.path.join(images_path, "test")), "no_labels": True},
+            # ),
+        ]
+
+    def _generate_examples(self, files, no_labels=False):
```
apsdehal, line 131: `no_labels` is currently used for test set but I will remove it if we can't figure out a way to make `labels` optional.

datasets/imagenet/imagenet.py
```diff
@@ -0,0 +1,148 @@
...
+        """Yields examples."""
+        idx = 0
+        for file_or_dir in files:
+            if isinstance(file_or_dir, tuple):
+                _, f = file_or_dir
+            else:
+                f = file_or_dir
```
lhoestq, line 138: It's always a string here if I'm not mistaken, why check for tuples ?

lhoestq: Can you also not use the variable `f` for a string ? You can rename it `path` instead

lhoestq (commented):
Awesome thank you ! I left a few comments :)

Could you also try to make the dummy_data.zip file a bit smaller ?
For example `valprep.sh` can contain only the commands for the dummy images. You can also reduce the number of images per split (3-4 is fine).

datasets/imagenet/imagenet.py
apsdehal: But the problem is that it creates a new config with a new identifier each time, is that okay?

datasets/imagenet/imagenet.py
```diff
@@ -0,0 +1,148 @@
...
+_VAL_PREP_SCRIPT = "https://raw.githubusercontent.com/soumith/imagenetloader.torch/master/valprep.sh"
+_DEFAULT_CONFIG_NAME = "2012"
+_NUM_IMAGENET1K_CLASSES = 1000
+
+
+class ImagenetConfig(datasets.BuilderConfig):
+    def __init__(self, name=_DEFAULT_CONFIG_NAME, num_classes=_NUM_IMAGENET1K_CLASSES, **kwargs):
+        kwargs.pop("version", None)
+        super(ImagenetConfig, self).__init__(version=datasets.Version("1.0.0"), name=name, **kwargs)
+        self.num_classes = num_classes
```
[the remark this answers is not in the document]
apsdehal: Update the PR to remove this. Originally, I wanted to assert if number of labels we receive from folder are equal to ones specified in config but with tests it is not possible to update the `num_classes` on the fly unless you have suggestion.

datasets/imagenet/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+[Needs More Information]
+
+#### Who are the source language producers?
+
+WordNet Synsets further quality controlled by human annotators.
+
+### Annotations
+
+#### Annotation process
```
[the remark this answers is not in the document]
apsdehal: Added more details.

apsdehal:
Thanks for the comments. I believe I have addressed all of them and also decreased the size of the dummy data file, so it should be ready for a re-review. I also made a change to allow adding synset mapping and valprep script in config in case we add ImageNet 21k some time later.

datasets/imagenet/imagenet.py
```diff
@@ -0,0 +1,195 @@
...
+            "data_dir='path/to/folder/imagenet_object_localization_patched2019.tar.gz')`"
+        )
+
+    def _info(self):
+        return datasets.DatasetInfo(
+            description=_DESCRIPTION,
+            features=datasets.Features(
+                {
+                    "image": datasets.Image(),
+                }
```
lhoestq, line 83: Could you also add the `labels` field with a `ClassLabel` feature type here please ?

lhoestq (commented):
Thanks a lot ! cc @sashavor if you'd like to take a look at the dataset card

I just have one more comment about the feature types in `_info`, It seems that the `labels` field is missing

datasets/imagenet/imagenet.py
mariosasko: Each image from the dataset is mapped to a single label so maybe `label` is a better name than `labels`. Also, let's add `ImageClassification(image_column="image", label_column="label")` as a task template to the dataset info.

datasets/imagenet/README.md
```diff
@@ -0,0 +1,1221 @@
...
+
+See the section 3.1 in [1](https://arxiv.org/abs/1409.0575) for more details on data collection procedure and [2](https://ieeexplore.ieee.org/abstract/document/5206848) for general information on ImageNet.
+
+#### Who are the annotators?
+
+Images are automatically fetched from an image search engine based on the synsets and filtered using human annotators on Amazon Mechanical Turk. See [1](https://arxiv.org/abs/1409.0575) for more details.
+
+### Personal and Sensitive Information
+
+See the papers: [1](https://arxiv.org/abs/1409.0575) [2](https://ieeexplore.ieee.org/abstract/document/5206848).
```
sashavor, line 1150: I think we should add a comment that the images contain faces of people who did not consent to this usage, which is why a recent version of ImageNet has face obfuscation applied, which does not impact results: https://arxiv.org/abs/2103.06191

datasets/imagenet/README.md
```diff
@@ -0,0 +1,1221 @@
...
+
+## Considerations for Using the Data
+
+### Social Impact of Dataset
+
+ImageNet dataset has been very crucial in advancement of deep learning technology as being the standard benchmark for vision models. The dataset aims to probe models on their understanding of the objects and has successfully shouldered the responsibility of holding through the test of time for that purpose. ImageNet is still one of the major datasets on which models are evaluated for their generalization in computer vision capabilities as the field moves towards self-supervised algorithms. Please see the future section in [1](https://arxiv.org/abs/1409.0575).
+
+### Discussion of Biases
+
+1. A [study](https://image-net.org/update-sep-17-2019.php) of the history of the multiple layers (taxonomy, object classes and labeling) of ImageNet and WordNet in 2019 described how bias is deeply embedded in most classification approaches for of all sorts of images.
```
sashavor, line 1160: I may be wrong, but I think the numbering is weird here? all 1s instead of 1,2,3?

datasets/imagenet/README.md
```diff
@@ -0,0 +1,1221 @@
...
+1. A [study](https://arxiv.org/abs/1811.12231) has also shown that ImageNet trained models are biased towards texture rather than shapes which in contrast with how humans do object classification. Increasing the shape bias improves the accuracy and robustness.
+1. Another [study](https://arxiv.org/abs/2109.13228) more potential issues and biases with the ImageNet dataset and provides an alternative benchmark on top. The data collected contains humans without their consent.
+1. ImageNet data with face obfuscation is also provided at [this link](https://image-net.org/face-obfuscation/)
+
```
sashavor, line 1164: Maybe I missed something, but maybe we should add that the most recent version of ImageNet has "filtered 2,702 synsets in the "person" subtree that may cause problematic behaviors of the model." (see https://www.image-net.org/filtering-and-balancing/)

datasets/imagenet/README.md
```diff
@@ -0,0 +1,1221 @@
...
+- **Repository:**
+- **Paper:** https://arxiv.org/abs/1409.0575
+- **Leaderboard:** https://paperswithcode.com/sota/image-classification-on-imagenet
+- **Point of Contact:** mailto: imagenet.help.desk@gmail.com  
+
+### Dataset Summary
+
+ILSVRC 2012, commonly known as 'ImageNet' is an image dataset organized according to the WordNet hierarchy. Each meaningful concept in WordNet, possibly described by multiple words or word phrases, is called a "synonym set" or "synset". There are more than 100,000 synsets in WordNet, majority of them are nouns (80,000+). In ImageNet, we aim to provide on average 1000 images to illustrate each synset. Images of each concept are quality-controlled and human-annotated. In its completion, we hope ImageNet will offer tens of millions of cleanly sorted images for most of the concepts in the WordNet hierarchy.
+
+This dataset requires manual download of the ImageNet data tar file [imagenet_object_localization_patched2019.tar.gz](https://www.kaggle.com/competitions/imagenet-object-localization-challenge/data?select=imagenet_object_localization_patched2019.tar.gz) and point datasets library to it by:
```
sashavor, line 62: Just checking that this is the most recent version of ImageNet, with the problematic person categories removed?

datasets/imagenet/imagenet.py
apsdehal: The reason I didn't add labels is similar to why `ImageFolder` doesn't have labels initially but are added later as we don't know them ahead of the time. I will have to load the labels file before hand if we want to achieve this. Let me know what you think.

datasets/imagenet/README.md
apsdehal: That's just better way to writing numbered lists in markdown. It renders as incremental list but adding all as `1.` allows easy modifications later. Let's say I add `1.`, `2.`, `3.` and I want to add a new `2.` later. I will need to update all of the number lists afterwards to achieve that. With `1.` I can just add the line as it is.

datasets/imagenet/README.md
apsdehal: I think those were removed in later face obfuscation. There are 3 person categories that are pure person categories in ImageNet. From the balancing paper:

> The 1,000 categories selected for the challenge contain only 3 people categories (scuba diver, bridegroom, and baseball player) while the full ImageNet contains 2,832 people categories under the person subtree (accounting for roughly 8.3% of the total images). Their use can be problematic and raises important questions about fairness and representation.

datasets/imagenet/imagenet.py
apsdehal: I have adjusted the code to reflect the review comments.

datasets/imagenet/README.md
apsdehal: Added these.

apsdehal:
@lhoestq I have updated the PR to address all of the review comments.

lhoestq (approved):
Thanks ! It looks all good to me :)

pinging @sashavor to review the dataset card one more time and we should be good

datasets/imagenet2012/README.md
```diff
@@ -0,0 +1,1235 @@
...
+
+- **Homepage:** https://image-net.org/index.php
+- **Repository:**
+- **Paper:** https://arxiv.org/abs/1409.0575
+- **Leaderboard:** https://paperswithcode.com/sota/image-classification-on-imagenet
+- **Point of Contact:** mailto: imagenet.help.desk@gmail.com  
+
+### Dataset Summary
+
+ILSVRC 2012, commonly known as 'ImageNet' is an image dataset organized according to the WordNet hierarchy. Each meaningful concept in WordNet, possibly described by multiple words or word phrases, is called a "synonym set" or "synset". There are more than 100,000 synsets in WordNet, majority of them are nouns (80,000+). ImageNet aims to provide on average 1000 images to illustrate each synset. Images of each concept are quality-controlled and human-annotated. In its completion, ImageNet hopes to offer tens of millions of cleanly sorted images for most of the concepts in the WordNet hierarchy.
```
sashavor, line 60: > "There are more than 100,000 synsets in WordNet, majority of them are nouns (80,000+). ImageNet aims to provide on average 1000 images to illustrate each synset." 

This is confusing ImageNet 21k and ImagetNet 1k

datasets/imagenet2012/README.md
```diff
@@ -0,0 +1,1235 @@
...
+
+
+🛑 This dataset requires **manual download** of the ImageNet data tar file (which is about **155GB** in size) [imagenet_object_localization_patched2019.tar.gz](https://www.kaggle.com/competitions/imagenet-object-localization-challenge/data?select=imagenet_object_localization_patched2019.tar.gz) and point datasets library to it by:
+
+```python
+from datasets import load_dataset
+dataset = load_dataset("imagenet2012", data_dir="/path/to/imagenet_object_localization_patched2019.tar.gz")
+```
+
+💡 This dataset provides access to ImageNet (ILSVRC) 2012 which is the most commonly used **subset** of ImageNet. This dataset spans 1000 object classes and contains 1,281,167 training images, 50,000 validation images and 100,000 test images. The version also has the [patch](https://drive.google.com/file/d/16RYnHpVOW0XKCsn3G3S9GTHUyoV2-4WX/view) already applied. For full ImageNet dataset presented in [[2]](https://ieeexplore.ieee.org/abstract/document/5206848), please check the download section of the [main website](https://image-net.org/download-images.php).
```
sashavor, line 70: > The version also has the [patch](https://drive.google.com/file/d/16RYnHpVOW0XKCsn3G3S9GTHUyoV2-4WX/view) already applied.

What does 'the patch' refer to?

datasets/imagenet2012/README.md
```diff
@@ -0,0 +1,1235 @@
...
+
+### Supported Tasks and Leaderboards
+
+- `image-classification`: The goal of this task is to classify a given image into one of 1000 ImageNet classes. The leaderboard is available [here](https://paperswithcode.com/sota/image-classification-on-imagenet).
```
sashavor, line 74: Maybe use the link that filters results to `Imagenet 1k only`? https://paperswithcode.com/sota/image-classification-on-imagenet?tag_filter=171

The link currently provided is all results, 1k and 21k.

datasets/imagenet2012/README.md
```diff
@@ -0,0 +1,1235 @@
...
+
+## Considerations for Using the Data
+
+### Social Impact of Dataset
+
+ImageNet dataset has been very crucial in advancement of deep learning technology as being the standard benchmark for vision models. The dataset aims to probe models on their understanding of the objects and has successfully shouldered the responsibility of holding through the test of time for that purpose. ImageNet is still one of the major datasets on which models are evaluated for their generalization in computer vision capabilities as the field moves towards self-supervised algorithms. Please see the future section in [1](https://arxiv.org/abs/1409.0575).
```
sashavor, line 1168: "The ImageNet dataset"?

sashavor: "and has successfully shouldered the responsibility of holding through the test of time for that purpose." == "and has become the *de facto* dataset for this purpose?"

Not sure about successfully shouldering :)

sashavor: "Please see the future section in [1](https://arxiv.org/abs/1409.0575)." -- for what?

datasets/imagenet2012/README.md
```diff
@@ -0,0 +1,1235 @@
...
+
+### Discussion of Biases
+
+1. A [study](https://image-net.org/update-sep-17-2019.php) of the history of the multiple layers (taxonomy, object classes and labeling) of ImageNet and WordNet in 2019 described how bias is deeply embedded in most classification approaches for of all sorts of images.
+1. A [study](https://arxiv.org/abs/1811.12231) has also shown that ImageNet trained models are biased towards texture rather than shapes which in contrast with how humans do object classification. Increasing the shape bias improves the accuracy and robustness.
+1. Another [study](https://arxiv.org/abs/2109.13228) more potential issues and biases with the ImageNet dataset and provides an alternative benchmark on top. The data collected contains humans without their consent.
```
sashavor, line 1174: > more potential issues and biases with the ImageNet dataset and provides an alternative benchmark on top.

on top of what?

sashavor (changes requested):
Some minor comments. especially regarding studies done on ImageNet, and the importance of distinguishing Imagenet 21k vs 1k, especially if we plan on adding the 21k version later.

Otherwise, looks good to me!

datasets/imagenet2012/README.md
apsdehal: Thanks for the comment.
> There are more than 100,000 synsets in WordNet, majority of them are nouns (80,000+)
This is more of a generic statement about WordNet and not related to ImageNet directly?

> ImageNet aims to provide on average 1000 images to illustrate each synset.
This is true for ImageNet1k as well.

I think the following line is actually causing more confusion and I will remove it:
> ImageNet hopes to offer tens of millions of cleanly sorted images for most of the concepts in the WordNet hierarchy.

sashavor (approved):

PR #4184, opened 2022-04-19
[Librispeech] Add 'all' config

patrickvonplaten:
Add `"all"` config to Librispeech


patrickvonplaten:
Fix https://github.com/huggingface/datasets/issues/4179

datasets/librispeech_asr/librispeech_asr.py
```diff
@@ -117,61 +133,101 @@ def _info(self):
...
+                            audio_file = f"{id_}.flac"
+                            speaker_id, chapter_id = [int(el) for el in id_.split("-")[:2]]
+                            transcripts.append(
+                                {
+                                    "id": id_,
+                                    "speaker_id": speaker_id,
+                                    "chapter_id": chapter_id,
+                                    "file": audio_file,
+                                    "text": transcript,
+                                    "subset": subset,
```
patrickvonplaten, line 224: @albertz @lhoestq - this "subset" arg is one of "clean-100", "clean-360", "other-500" for:

```python
ds = load_dataset("librispeech_asr", "all", split="train")
ds[0]
```

This should allow the user to easily define which subsets are needed I think

albertz:
Just that I understand: With this change, simply doing `load_dataset("librispeech_asr")` is possible and returns the whole dataset?

datasets/librispeech_asr/librispeech_asr.py
```diff
@@ -69,6 +75,14 @@ def map_to_array(batch):
         "dev": _DL_URL + "dev-other.tar.gz",
         "train.500": _DL_URL + "train-other-500.tar.gz",
     },
+    "all": {
+        "dev": {retrieve_subset(u): _DL_URL + u for u in ["dev-clean.tar.gz", "dev-other.tar.gz"]},
```
albertz, line 79: I think it is more standard to split this into dev-clean and dev-other because you usually always treat and report them separately.

datasets/librispeech_asr/librispeech_asr.py
```diff
@@ -69,6 +75,14 @@ def map_to_array(batch):
...
+        "test": {retrieve_subset(u): _DL_URL + u for u in ["test-clean.tar.gz", "test-other.tar.gz"]},
```
albertz, line 80: Likewise.

datasets/librispeech_asr/librispeech_asr.py
```diff
@@ -69,6 +75,14 @@ def map_to_array(batch):
...
+        "train": {
```
albertz, line 81: But this is fine. I think it's easy to apply any custom filtering for curriculum learning for early stages in training and then later using the whole dataset.

datasets/librispeech_asr/librispeech_asr.py
patrickvonplaten: sounds good, we can make that change

patrickvonplaten:
> Just that I understand: With this change, simply doing `load_dataset("librispeech_asr")` is possible and returns the whole dataset?
> 
> And to get the subsets, I do sth like:
> 
> ```python
> ds = load_dataset("librispeech_asr")
> train_ds = ds["train"]
> dev_clean_ds = ds["dev-clean"]
> dev_other_ds = ds["dev-other"]
> test_clean_ds = ds["test-clean"]
> test_other_ds = ds["test-other"]
> ```
> 
> ?

You could do:


```python
ds = load_dataset("librispeech_asr", "all")  # <- note that we have to pass a config
train_ds = ds["train"]
dev_clean_ds = ds["dev-clean"]
dev_other_ds = ds["dev-other"]
test_clean_ds = ds["test-clean"]
test_other_ds = ds["test-other"]
```

albertz:
So, `load_dataset("librispeech_asr")` is not possible, it must be `load_dataset("librispeech_asr", "all")`?

Why is that?

The docs say:
```
name: `str` name, optional configuration for the dataset that affects the data generated on disk. Different
    `builder_config`s will have their own subdirectories and versions.
    If not provided, uses the first configuration in self.BUILDER_CONFIGS
```
https://github.com/huggingface/datasets/blob/cd3ce34ab1604118351e1978d26402de57188901/src/datasets/builder.py#L228

Or maybe you could just define `DEFAULT_CONFIG_NAME`?

lhoestq:
> If not provided, uses the first configuration in self.BUILDER_CONFIGS

Oh crap this is outdated documentation. No it doesn't take the first config by default.

albertz:
> No it doesn't take the first config by default.

But defining `DEFAULT_CONFIG_NAME` would work?

So should we define `DEFAULT_CONFIG_NAME = "all"` here as well? I think this is a reasonable default config.

Don't most datasets have some default config?

lhoestq:
> But defining DEFAULT_CONFIG_NAME would work?
>
> So should we define DEFAULT_CONFIG_NAME = "all" here as well? I think this is a reasonable default config.

Yes that would work, and I also find it reasonable to do it :)

> Don't most datasets have some default config?

Most datasets only have one configuration, so the single configuration is the default one. Then other datasets gave several configurations, and whether they have a default one is decided case-by-case.

e.g. `glue` is a benchmark and doesn't have a default task, one must choose which task of `glue` they want to use explicitely.

patrickvonplaten:
Thanks a lot for the feedback! 

Using `"all"` now as the default config. I changed the layout a bit so that there is not a single "train", but instead we have multiple "train.clean.100", "train.clean.360", "train.other.500". This way we don't even need to do filtering and it's also cleaner IMO.

@albertz - you should now be able to do the following:

```python
load_dataset("librispeech_asr")  # <- run this once to download, prepare dataset and cache everything

# The following operations will be very fast since all the downloading and processing is already cached
train_1 = load_dataset("librispeech_asr", split="train.clean.100")
print(train_1)
train_2 = load_dataset("librispeech_asr", split="train.clean.100+train.clean.360")
print(train_2)
train_full = load_dataset("librispeech_asr", split="train.clean.100+train.clean.360+train.other.500")
print(train_full)
dev_clean_ds = load_dataset("librispeech_asr", split="validation.clean")
print(dev_clean_ds)
dev_other_ds = load_dataset("librispeech_asr", split="validation.other")
print(dev_other_ds)
test_clean_ds = load_dataset("librispeech_asr", split="test.clean")
print(test_clean_ds)
test_other_ds = load_dataset("librispeech_asr", split="test.other")
print(test_other_ds)
```

patrickvonplaten:
Think this way we have the best of both worlds. Also @lhoestq, I think we could highlight better in the docs that it's possible to combine different splits. We do this actually quite a lot for speech. For Common Voice many people include "validation" in the training if the data is too small, e.g.: https://github.com/huggingface/transformers/blob/ff06b177917384137af2d9585697d2d76c40cdfc/examples/pytorch/speech-recognition/run_speech_recognition_ctc.py#L147

Should we maybe add a short section to the loading tutorial here: https://huggingface.co/docs/datasets/v2.1.0/en/loading#hugging-face-hub ? (Happy to do it)

albertz:
Is there any advantage or difference in calling `load_dataset` multiple times for each split? Or why not just call `load_dataset` once and then access each split?

Note in our case, we cannot really use the caching mechanism because we have a recipe pipeline used by multiple users (and I think a common cache dir for all users might end up in problems) and we basically would use `load_dataset("librispeech_asr").save_to_disk(...)` and then later `load_from_disk(...)`.

So with `load_from_disk`, we cannot really provide the split this way, so we anyway would do sth like:
```python
ds = datasets.load_from_disk(...)
train = ds["train"]
```
Or with your latest proposal, it would look like:
```python
ds = datasets.load_from_disk(...)
train_ds = datasets.concatenate_datasets(
  [ds["train.clean.100"], ds["train.clean.360"], ds["train.other.500"]])
```
right?

patrickvonplaten:
> Is there any advantage or difference in calling `load_dataset` multiple times for each split? Or why not just call `load_dataset` once and then access each split?
> 
> Note in our case, we cannot really use the caching mechanism because we have a recipe pipeline used by multiple users (and I think a common cache dir for all users might end up in problems) and we basically would use `load_dataset("librispeech_asr").save_to_disk(...)` and then later `load_from_disk(...)`. (See here: [rwth-i6/i6_core#253](https://github.com/rwth-i6/i6_core/pull/253))
> 
> So with `load_from_disk`, we cannot really provide the split this way, so we anyway would do sth like:
> 
> ```python
> ds = datasets.load_from_disk(...)
> train = ds["train"]
> ```
> 
> Or with your latest proposal, it would look like:
> 
> ```python
> ds = datasets.load_from_disk(...)
> train_ds = datasets.concatenate_datasets(
>   [ds["train.clean.100"], ds["train.clean.360"], ds["train.other.500"]])
> ```
> 
> right?

I see the use case! The only advantage by calling `datasets` multiple times is that one can easily "merge" splits with `"+"`, but yeah you can do the exact same with `concatenate`.

@lhoestq what do you think is the best approach with `load_from_disk`? 

@albertz, you could also define the `cache_dir` when doing `load_dataset(...)` which will then put all the relevant `arrow` files int the cache dir that you defined, e.g.:

```python
load_dataset("librispeech_asr", cache_dir="/easy/to/access/directory")
```

datasets/librispeech_asr/librispeech_asr.py
```diff
@@ -159,6 +271,9 @@ def _generate_examples(self, files):
                         id_, transcript = line.split(" ", 1)
                         audio_file = f"{id_}.flac"
                         speaker_id, chapter_id = [int(el) for el in id_.split("-")[:2]]
+                        audio_file = (
+                            os.path.join(local_extracted_archive, audio_file) if local_extracted_archive else None
```
patrickvonplaten, line 275: @lhoestq is the whole speech data now stored twice in non-streaming mode? Once as the original flac and once as an array of bytes? 

Should we change line 266: from `audio_data[id_] = f.read()` to just `audio_data[id_] = f` and change the line: `audio = {"path": transcript["file"], "bytes": audio_data[transcript["id"]]}` at the moment it looks to me like both the bytes and the file is saved in non-streaming mode which would create a huge overhead in memory no?

patrickvonplaten: Opening a new PR for this to change it internally in `src/datasets` ...

patrickvonplaten:
@albertz, I took a read through https://github.com/rwth-i6/i6_core/pull/253 . 

I think the best would be the following:

```
1. Do `ds = load_dataset(..., cache_dir="/dir/that/is/easy/to/access")` <- having merged this PR, this will save all the original `.flac` files in the `cache_dir`
2. Do `ds.save_to_disk("local/path")` this should then only save the `arrow.format` with a `path` string to the audio files which are located in `cache_dir` <- this won't require a lot of memory after https://github.com/huggingface/datasets/pull/4184#discussion_r854132740 is fixed and can be done for each person individually.
3. `ds = datasets.load_from_disk("local/path")` can the be used. An object of `ds` will then have a `path` variable that links to the original audio files in the `cache_dir`. You can change these audio files then easily to `.mp3. You could do this with the `.map(...)` function, e.g. define a function that maps through all audio files, load them and then save them on disk afterward.

**Comment by patrickvonplaten** [comment]
@lhoestq - I think this one is good to go

File: datasets/librispeech_asr/librispeech_asr.py (hunk 5 above)
Comment (lhoestq) [reply datasets/librispeech_asr/librispeech_asr.py:260]: Will be fixed in https://github.com/huggingface/datasets/pull/4187 :)

**Comment by dthulke** [comment]
@albertz 

> Oh, so you say that our current implementation in https://github.com/rwth-i6/i6_core/pull/253 is broken? Because our cache dir is just some temp directory which will be removed afterwards, and we just store what we get out of save_to_disk. I think it would be good to clarify that in the doc of save_to_disk, that this is not enough and can depend on files from the cache dir. (@dthulke)

Oh, I wasn't aware that audio files are handled this way. Then we should have the cache directory as an additional job output, so that we keep the audio files. 

> So, you say we anyway need to share the cache dir among users?

No, the cache dir can still be a directory in the job output folder. Then the audio paths in the corresponding dataset column correspond to the flac files in that directory. This way the "output" of the job is contained into the job directory and we don't write files to a global cache directory that is independent of the sisyphus graph.

If we want to share the audio data between different users, we can just link to a central instance of the job (similar to how we do it with the `DownloadLibriSpeechCorpusJob`).

**Comment by patrickvonplaten** [comment]
So what I would suggest here is to do the following:

1. Do `load_dataset(..., cache_dir=/a/read-only/folder)`
2. 
- Either just re-use `load_dataset(..., cache_dir=...)` which should always re-use the data in the `cache_dir` since the hash of the url matches - so there should never be any duplicated downloading 

or 

- If you want to store the files in MP3 locally, first convert the files to MP3 in the read-only folder, then take do `ds.save_to_disk(/some/path)` which will save the correct path to the read-only folder to MP3 and then you can easily re-use the small arrow dataset that is saved in `/some/path`

**Comment by patrickvonplaten** [comment]
> So what I would suggest here is to do the following:
> 
> 1. Do `load_dataset(..., cache_dir=/a/read-only/folder)`
> 
> * Either just re-use `load_dataset(..., cache_dir=...)` which should always re-use the data in the `cache_dir` since the hash of the url matches - so there should never be any duplicated downloading
> 
> or
> 
> * If you want to store the files in MP3 locally, first convert the files to MP3 in the read-only folder, then take do `ds.save_to_disk(/some/path)` which will save the correct path to the read-only folder to MP3 and then you can easily re-use the small arrow dataset that is saved in `/some/path`

Also relevant here: https://github.com/huggingface/datasets/issues/3663

**Comment by lhoestq** [comment]
I also added some documentation about how `save_to_disk` handles audio files here: https://github.com/huggingface/datasets/pull/4193

**Review by lhoestq — COMMENTED** [review]
This PR looks good thanks :)

@patrickvonplaten can you also update the dataset_infos.json file with this new config ?
```
datasets-cli test ./datasets/librispeech --save_infos --name all
```

To fix the CI you can also replace `audio-speaker-identification` by `speaker-identification` in the task tags of the datase tcard

**Comment by albertz** [comment]
> > So, you say we anyway need to share the cache dir among users?
> 
> No, the cache dir can still be a directory in the job output folder.

@dthulke But this is what I mean. When we share the job output folder, it means we share the cache dir among users.

I wonder if `load_dataset(..., cache_dir=job_output_cache_dir)` is always save to do then, that it really would not modify the `job_output_cache_dir`.

We could enforce that by making the `job_output_cache_dir` read-only afterwards. We currently don't do this.

@patrickvonplaten @dthulke But in any case, we actually prefer the data content to be inside the dataset (the arrow files). Lots of small files would be very problematic for our cache manager. We have one main copy of the data on NFS, but accessing the NFS directly by all computing nodes is not feasible, so the cache manager will have copies of the files on the nodes. So it means, whenever we access some file, we query the cache manager DB whether the file is already cached somewhere (some other computing node) and if so, it copies it from the other computing node and not from NFS. This works very well when there are not too many files (but the files can be big). So, we want to have only a few but big files. Even for NFS access this is much better.

I will maybe also comment in #3663.

**Comment by patrickvonplaten** [comment]
Hey @albertz @dthulke,

Thanks a lot for your input! 

We've discussed quite a bit with @lhoestq and we think the best approach is the following:


a)
`load_dataset(...)` will not store both bytes and the files because this would mean that 3x the size of the dataset would often be needed (1. the compressed `tar.gz` file, 2. the extracted file b, 3. the raw bytes in arrow format). 

For canonical datasets like librispeech and common voice I think we want to keep the dataset filenames because of i) no breaking changes and ii) reasons explained in #3663

However it's also trivial to write your own datasetset downloading script of librispeech and just not extract the folder e.g. this line: https://huggingface.co/datasets/common_voice/blob/main/common_voice.py#L671

And then it'll be allowed to save the bytes and the dataset will be self-contained out-of-the-box when using `load_dataset(...)`

b) Now, one major problem that you guys uncovered is that `save_to_disk(...)` is currently not necessarily saving a dataset to be self-contained. We will change that asap. This means that after we've corrected this when you do download the canonical librispeech dataset the following will work:

```python
ds = load_dataset("....")  # <- here we have a dependency on the filepathes
ds[0]["audio"]["bytes"] # <- will not work

ds.save_to_disk("/local/path")  # <- now we want to have a self-contained dataset in arrow format, so we load the files into bytes and save it in arrow format

# now you can delete everything besides "/local/path"

ds = load_from_disk("/local/path")  # <- this will work
```

So either option a) where you define your own librispeech data downloading script (you guys could just sign up here: https://huggingface.co/join) and upload a dataset loading script in private mode so that no one can see it and you would always store the audio as bytes or b) where you first load then save to disk then delete cache would work. 

Hope that fits in your vision :-)

cc @lhoestq @mariosasko

dthulke:
@albertz I would suggest to move the discussion on implementation details on our side to the following issue: rwth-i6/i6_core/issues/257

lhoestq:
I like the idea of adding `embed_external_files` and set it to True by default to `save_to_disk`.
It's indeed a kind of breaking change since some users will get bigger Arrow files when updating the lib, but the advantages are nice:
1. I like the idea of having it self contained, in case you want to delete your cache
2. users also upload these Arrow files to cloud storage via the `fs` parameter, and in this case they would expect to upload a self-contained dataset
3. consistency with `push_to_hub`

If it sounds good to you I'll open an issue to discuss this and track the advancements

lhoestq (approved):
Awesome thanks !

datasets/librispeech_asr/librispeech_asr.py
```diff
@@ -159,6 +256,11 @@ def _generate_examples(self, files):
                         id_, transcript = line.split(" ", 1)
                         audio_file = f"{id_}.flac"
                         speaker_id, chapter_id = [int(el) for el in id_.split("-")[:2]]
+                        audio_file = (
```
patrickvonplaten, line 259: Needed to have extracted files in non-streaming mode

albertvillanova:
Closed #4179.

wxlsummer:
> ```python
> load_dataset("librispeech_asr")
> ```

Hi when I run:

import pandas as pd
from multiprocessing import Pool
import os
data_dir = my own path
dataset=load_dataset("librispeech_asr", cache_dir=data_dir)

after downloading those files and the spliting process, here is a error like this:
ExpectedMoreSplits                        Traceback (most recent call last)
/tmp/ipykernel_815982/814767946.py in <module>
      4 import os
      5 data_dir = '/disk/scratch2/s1905792/librispeech'
----> 6 dataset=load_dataset("librispeech_asr", cache_dir=data_dir)
      7 #test = load_dataset("librispeech_asr", "all", "train", cache_dir=data_dir)

/disk/scratch2/s1905792/anaconda3/envs/py3.7-gpu/lib/python3.7/site-packages/datasets/load.py in load_dataset(path, name, data_dir, data_files, split, cache_dir, features, download_config, download_mode, verification_mode, ignore_verifications, keep_in_memory, save_infos, revision, use_auth_token, task, streaming, num_proc, storage_options, **config_kwargs)
   1813         try_from_hf_gcs=try_from_hf_gcs,
   1814         num_proc=num_proc,
-> 1815         storage_options=storage_options,
   1816     )
   1817 

/disk/scratch2/s1905792/anaconda3/envs/py3.7-gpu/lib/python3.7/site-packages/datasets/builder.py in download_and_prepare(self, output_dir, download_config, download_mode, verification_mode, ignore_verifications, try_from_hf_gcs, dl_manager, base_path, use_auth_token, file_format, max_shard_size, num_proc, storage_options, **download_and_prepare_kwargs)
    911                             verification_mode=verification_mode,
    912                             **prepare_split_kwargs,
--> 913                             **download_and_prepare_kwargs,
    914                         )
    915                     # Sync info

/disk/scratch2/s1905792/anaconda3/envs/py3.7-gpu/lib/python3.7/site-packages/datasets/builder.py in _download_and_prepare(self, dl_manager, verification_mode, **prepare_split_kwargs)
   1020 
   1021         if verification_mode == VerificationMode.BASIC_CHECKS or verification_mode == VerificationMode.ALL_CHECKS:
-> 1022             verify_splits(self.info.splits, split_dict)
   1023 
   1024         # Update the info object with the splits.

/disk/scratch2/s1905792/anaconda3/envs/py3.7-gpu/lib/python3.7/site-packages/datasets/utils/info_utils.py in verify_splits(expected_splits, recorded_splits)
     89         return
     90     if len(set(expected_splits) - set(recorded_splits)) > 0:
---> 91         raise ExpectedMoreSplits(str(set(expected_splits) - set(recorded_splits)))
     92     if len(set(recorded_splits) - set(expected_splits)) > 0:
     93         raise UnexpectedSplits(str(set(recorded_splits) - set(expected_splits)))

ExpectedMoreSplits: {'validation.other', 'test.other', 'test.clean', 'validation.clean', 'train.other.500', 'train.clean.360', 'train.clean.100'}

Could you please tell me where I was wrong and did I miss any files?
Best Regards,
Xiaoliang

albertvillanova:
Hi @wxlsummer,

Let's continue the discussion in the Community tab of the dataset: https://huggingface.co/datasets/openslr/librispeech_asr/discussions/11

PR #4193, opened 2022-04-21
Document save_to_disk and push_to_hub on images and audio files

lhoestq:
Following https://github.com/huggingface/datasets/pull/4187, I explained in the documentation of `save_to_disk` and `push_to_hub` how they handle image and audio data.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -976,6 +976,28 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         Saves a dataset to a dataset directory, or in a filesystem using either :class:`~filesystems.S3FileSystem` or
         any implementation of ``fsspec.spec.AbstractFileSystem``.
 
+        For :class:`Image` and :class:`Audio` data:
+
+        If your images and audio files are local files, then the resulting arrow file will store paths to these files.
+        If you want to include the bytes or your images or audio files instead, you must `read()` those files first.
+        This can be done by storing the "bytes" instead of the "path" of the images or audio files:
+
+        ```python
+        >>> def read_image_file(example):
+        ...     with open(example["image"].filename, "rb") as f:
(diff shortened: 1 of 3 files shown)


src/datasets/arrow_dataset.py
```diff
@@ -3502,9 +3524,12 @@ def push_to_hub(
         shard_size: Optional[int] = 500 << 20,
         embed_external_files: bool = True,
     ):
-        """Pushes the dataset to the hub.
+        """Pushes the dataset to the hub as a Parquet dataset.
         The dataset is pushed using HTTP requests and does not need to have neither git or git-lfs installed.
 
+        The resulting Parquet files are self-contained: if your dataset contains :class:`Image` or :class:`Audio`
+        data, the Parquet files will store the bytes of your images or audio files.
+
```
mariosasko, line 3532: We should mention that this is only true if `embed_external_files` is set to True (default behavior).

src/datasets/dataset_dict.py
```diff
@@ -853,11 +875,14 @@ def push_to_hub(
...
         embed_external_files: bool = True,
     ):
-        """Pushes the ``DatasetDict`` to the hub.
+        """Pushes the ``DatasetDict`` to the hub as a Parquet dataset.
         The ``DatasetDict`` is pushed using HTTP requests and does not need to have neither git or git-lfs installed.
 
         Each dataset split will be pushed independently. The pushed dataset will keep the original split names.
 
+        The resulting Parquet files are self-contained: if your dataset contains :class:`Image` or :class:`Audio`
+        data, the Parquet files will store the bytes of your images or audio files.
```
mariosasko, line 884: Same here

mariosasko (commented):
One nit

lhoestq:
Good catch, I updated the docstrings

mariosasko (approved):
Thanks! Looks good now.

PR #4197, opened 2022-04-21
Add remove_columns=True

thomasw21:
This should fix all the issue we have with in place operations in mapping functions. This is crucial as where we do some weird things like:
```
def apply(batch):
    batch_size = len(batch["id"])
    batch["text"] = ["potato" for _ range(batch_size)]
    return {}

# Columns are: {"id": int}
dset.map(apply, batched=True, remove_columns="text") # crashes because `text` is not in the original columns
dset.map(apply, batched=True) # mapped datasets has `text` column
```

In this PR we suggest to have `remove_columns=True` so that we ignore the input completely, and just use the output to generate mapped dataset. This means that inplace operations won't have any effects anymore.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -1847,7 +1847,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                 - `function(batch: Union[Dict[List], List[Any]]) -> Union[Dict, Any]` if `batched=True` and `with_indices=False` and `with_rank=False`
                 - `function(batch: Union[Dict[List], List[Any]], *extra_args) -> Union[Dict, Any]` if `batched=True` and `with_indices=True` and/or `with_rank=True` (one extra arg for each)
 
-                If no function is provided, default to identity function: ``lambda x: x``.
+                If no function is provided, default to identity function: ``lambda x: {}``.
             with_indices (:obj:`bool`, default `False`): Provide example indices to `function`. Note that in this case the
                 signature of `function` should be `def function(example, idx[, rank]): ...`.
             with_rank (:obj:`bool`, default `False`): Provide process rank to `function`. Note that in this case the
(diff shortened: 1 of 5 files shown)


src/datasets/iterable_dataset.py
```diff
@@ -206,12 +206,21 @@ def __iter__(self):
...
-                transformed_batch.update(self.function(*function_args))
+
+                processed_batch = self.function(*function_args)
                 # then remove the unwanted columns
-                if self.remove_columns:
+                if isinstance(self.remove_columns, Iterable):
                     for c in self.remove_columns:
-                        del transformed_batch[c]
+                        if c in transformed_batch:
+                            del transformed_batch[c]
```
thomasw21, line 217: This is the sidefix: remove has to be done on input, and then merged with processed_input similarly to original map function.

src/datasets/iterable_dataset.py
```diff
@@ -568,16 +585,21 @@ def map(
...
+                    - str: Single column removed. Columns from the output of `function` are preserved.
+                    - List[str]: List of columns removed. Columns from the output of `function` are preserved.
         """
         if isinstance(input_columns, str):
             input_columns = [input_columns]
         if isinstance(remove_columns, str):
             remove_columns = [remove_columns]
         if function is None:
-            function = lambda x: x  # noqa: E731
+            function = lambda x: {}  # noqa: E731
```
thomasw21, line 602: That's a direct consequence of the fix, this was an inplace operation, which means that if you remove a column, it would get removed from both `input` and `processed_input` which causes the test to work, but in fact given the API we have for `arrow_dataset` this should have returned an unchanged dataset.

tests/test_iterable_dataset.py
```diff
@@ -506,6 +513,23 @@ def test_iterable_dataset_map_batched(dataset: IterableDataset, generate_example
...
 
+    # Test that remove_column=True removes original column
+    dataset = dataset.map(func, batched=True, batch_size=batch_size, remove_columns=True)
+    assert isinstance(dataset._ex_iterable, MappedExamplesIterable)
+    assert dataset._ex_iterable.function is func
+    assert dataset._ex_iterable.batch_size == batch_size
+    assert next(iter(dataset)) == {"id+1": 1}
+
+
+def test_iterable_dataset_map_persist_with_remove_column(dataset: IterableDataset, generate_examples_fn):
```
lhoestq, line 524:
```suggestion
def test_iterable_dataset_map_persist_with_remove_column(dataset: IterableDataset):
```

lhoestq (commented):
Thanks for adding this and for the tests :)

src/datasets/arrow_dataset.py
```diff
@@ -1859,9 +1859,14 @@ def map(
                 `batch_size <= 0` or `batch_size == None`: Provide the full dataset as a single batch to `function`.
             drop_last_batch (:obj:`bool`, default `False`): Whether a last batch smaller than the batch_size should be
                 dropped instead of being processed by the function.
-            remove_columns (`Optional[Union[str, List[str]]]`, default `None`): Remove a selection of columns while doing the mapping.
+            remove_columns (`Optional[Union[bool,List[str]]]`, defaults to `None`): Remove a selection of columns while doing the mapping.
```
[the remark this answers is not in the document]
thomasw21: So I thought about this, and this would mean a breaking change. Right now, `False` and `None` do the same thing, but some people might be setting that value to `None`.

mariosasko:
Any reason why we can't just do `[inputs.copy()]` in this line for in-place operations to not have effects anymore:
https://github.com/huggingface/datasets/blob/bf432011ff9155a5bc16c03956bc63e514baf80d/src/datasets/arrow_dataset.py#L2232.

(in the `batched` case, we can also copy the inputs' values (list objects) to ignore in-place modifications to the inputs' columns)

I think `remove_columns=True` has no meaning, so I'm not a fan of this change.

thomasw21:
@mariosasko copy does have a cost associated with it ...

thomasw21:
Okay closing this PR for the following reasons:
 - `remove_columns=True` was expected to keep the `.update`-like operator for `.map`. I initially thought it would be a good way to ignore function side effects and only keep output of that function (cf. PR description).
 - expected `remove_columns=True` is a bad API according to @mariosasko and introduces unecessary changes for little gain (strictly equivalent to `remove_columns=dset.column_names`)

Issue #4192, opened 2022-04-21
load_dataset can't load local dataset,Unable to find ...


ahf876828330:
Traceback (most recent call last):
  File "/home/gs603/ahf/pretrained/model.py", line 48, in <module>
    dataset = load_dataset("json",data_files="dataset/dataset_infos.json")
  File "/home/gs603/miniconda3/envs/coderepair/lib/python3.7/site-packages/datasets/load.py", line 1675, in load_dataset
    **config_kwargs,
  File "/home/gs603/miniconda3/envs/coderepair/lib/python3.7/site-packages/datasets/load.py", line 1496, in load_dataset_builder
    data_files=data_files,
  File "/home/gs603/miniconda3/envs/coderepair/lib/python3.7/site-packages/datasets/load.py", line 1155, in dataset_module_factory
    download_mode=download_mode,
  File "/home/gs603/miniconda3/envs/coderepair/lib/python3.7/site-packages/datasets/load.py", line 800, in get_module
    data_files = DataFilesDict.from_local_or_remote(patterns, use_auth_token=self.downnload_config.use_auth_token)
  File "/home/gs603/miniconda3/envs/coderepair/lib/python3.7/site-packages/datasets/data_files.py", line 582, in from_local_or_remote
    if not isinstance(patterns_for_key, DataFilesList)
  File "/home/gs603/miniconda3/envs/coderepair/lib/python3.7/site-packages/datasets/data_files.py", line 544, in from_local_or_remote
    data_files = resolve_patterns_locally_or_by_urls(base_path, patterns, allowed_extensions)
  File "/home/gs603/miniconda3/envs/coderepair/lib/python3.7/site-packages/datasets/data_files.py", line 194, in resolve_patterns_locally_or_by_urls
    for path in _resolve_single_pattern_locally(base_path, pattern, allowed_extensions):
  File "/home/gs603/miniconda3/envs/coderepair/lib/python3.7/site-packages/datasets/data_files.py", line 144, in _resolve_single_pattern_locally
    raise FileNotFoundError(error_msg)
FileNotFoundError: Unable to find '/home/gs603/ahf/pretrained/dataset/dataset_infos.json' at /home/gs603/ahf/pretrained

![image](https://user-images.githubusercontent.com/33253979/164413285-84ea65ac-9126-408f-9cd2-ce4751a5dd73.png)
![image](https://user-images.githubusercontent.com/33253979/164413338-4735142f-408b-41d9-ab87-8484de2be54f.png)

[...]

PR #4200, opened 2022-04-22
Add to docs how to load from local script

albertvillanova:
This option was missing from the docs guide (it was only explained in the docstring of `load_dataset`). Although this is an infrequent use case, there might be some users interested in it.

Related to #4192

CC: @stevhliu

### DIFF
--- a/docs/source/loading.mdx
+++ b/docs/source/loading.mdx
@@ -5,6 +5,7 @@ You have already seen how to load a dataset from the Hugging Face Hub. But datas
 This guide will show you how to load a dataset from:
 
 - The Hub without a dataset loading script
+- Local loading script
 - Local files
 - In-memory data
 - Offline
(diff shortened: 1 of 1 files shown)


docs/source/loading.mdx
```diff
@@ -69,6 +70,18 @@ Specify a custom split with the `split` parameter:
 >>> c4_validation = load_dataset("allenai/c4", data_files=data_files, split="validation")
 ```
 
+## Local loading script
+
+You may have a 🤗 Datasets loading script locally on your computer. You can load this dataset by passing to [`load_dataset`] one of the following paths:
+
+- The local path to the loading script file
```
stevhliu, line 77:
```suggestion
- The local path to the loading script file.
```

docs/source/loading.mdx
```diff
@@ -69,6 +70,18 @@ Specify a custom split with the `split` parameter:
...
+- The local path to the directory containing the loading script file (only if the script file has the same name as the directory)
```
stevhliu, line 78:
```suggestion
- The local path to the directory containing the loading script file (only if the script file has the same name as the directory).
```

docs/source/loading.mdx
```diff
@@ -69,6 +70,18 @@ Specify a custom split with the `split` parameter:
...
+
+```py
+>>> dataset = load_dataset("path/to/local/loading_script/loading_script.py", split="train")
+>>> dataset = load_dataset("path/to/local/loading_script", split="train")  # equivalently because the file has the same name as the directory
```
stevhliu, line 82:
```suggestion
>>> dataset = load_dataset("path/to/local/loading_script", split="train")  # equivalent because the file has the same name as the directory
```

stevhliu (approved):
Looks great, thanks for adding this use case!

PR #4201, opened 2022-04-22
Update GH template for dataset viewer issues

albertvillanova:
Update template to use new issue forms instead.

With this PR we can check if this new feature is useful for us.

Once validated, we can update the other templates.

CC: @severo


albertvillanova:
You can see rendering at: https://github.com/huggingface/datasets/blob/6b48fedbdafe12a42c7b6edcecc32820af1a4822/.github/ISSUE_TEMPLATE/dataset-viewer.yml

severo (approved):
Nice, it will be easier for the users to report issues. Thanks

PR #4202, opened 2022-04-22
Fix some type annotation in doc

thomasw21:


src/datasets/arrow_dataset.py
```diff
@@ -1864,10 +1864,10 @@ def map(
...
-                - `function(batch: Union[Dict[List], List[Any]]) -> Union[Dict, Any]` if `batched=True` and `with_indices=False` and `with_rank=False`
-                - `function(batch: Union[Dict[List], List[Any]], *extra_args) -> Union[Dict, Any]` if `batched=True` and `with_indices=True` and/or `with_rank=True` (one extra arg for each)
+                - `function(example: Dict[str, Any]) -> Optional[Union[Dict[str, Any], pyarrow.Table]]` if `batched=False` and `with_indices=False` and `with_rank=False`
+                - `function(example: Dict[str, Any], *extra_args) -> Optional[Union[Dict[str, Any], pyarrow.Table]]` if `batched=False` and `with_indices=True` and/or `with_rank=True` (one extra arg for each)
+                - `function(batch: Dict[str, List]) -> Optional[Union[Dict[str, List], pyarrow.Table]]` if `batched=True` and `with_indices=False` and `with_rank=False`
+                - `function(batch: Dict[str, List], *extra_args) -> Optional[Union[Dict[str, List], pyarrow.Table]]` if `batched=True` and `with_indices=True` and/or `with_rank=True` (one extra arg for each)
```
lhoestq, line 1870: I would formulate it differently to make it easier to understand at first glance:

```suggestion
                - `function(example: Dict[str, Any]) -> Dict[str, Any]` if `batched=False` and `with_indices=False` and `with_rank=False`
                - `function(example: Dict[str, Any], *extra_args) -> Dict[str, Any]` if `batched=False` and `with_indices=True` and/or `with_rank=True` (one extra arg for each)
                - `function(batch: Dict[str, List]) -> Dict[str, List]` if `batched=True` and `with_indices=False` and `with_rank=False`
                - `function(batch: Dict[str, List], *extra_args) -> Dict[str, List]` if `batched=True` and `with_indices=True` and/or `with_rank=True` (one extra arg for each)
                
            For advanced usage, the function can also return a `pyarrow.Table`.
            Moreover if your function returns nothing (`None`), then `map` will run your function and return the dataset unchanged.
```

Let me know if you like it this way

thomasw21: Hum yeah that sounds reasonable (especially since we didn't cover `input_columns` usage)

lhoestq (approved):
Thanks !

PR #4204, opened 2022-04-22
Add Recall Metric Card

emibaylor:
What this PR mainly does:
- add metric card for recall metric
- update docs in recall python file

Note: I've also included a .json file with all of the metric card information. I've started compiling the relevant information in this type of .json files, and then using a script I wrote to generate the formatted metric card, as well as the docs to go in the .py file. I figured I'd upload the .json because it could be useful, especially if I also make a PR with the script I'm using (let me know if that's something you think would be beneficial!)


sashavor:
This looks good to me!

lhoestq (approved):
Cool ! Maybe let's not merge the JSON file unless we have the tool available yet ? In case we do some changes inside it. If it sounds good to you I think you can remove the JSON file before merging

metrics/recall/README.md
```diff
@@ -0,0 +1,114 @@
+# Metric Card for Recall
+
+
+## Metric Description
+
+Recall is the fraction of the positive examples that were correctly labeled by the model as positive. It can be computed with the equation:
+Recall = TP / (TP + FN)
+Where TP is the true positives and FN is the false negatives.
```
sashavor, line 8: 'number of true positives' and 'number of false negatives'?

metrics/recall/README.md
```diff
@@ -0,0 +1,114 @@
...
+- **zero_division** (): Sets the value to return when there is a zero division. Defaults to .
+    - `'warn'`: If there is a zero division, the return value is `0`, but warnings are also raised.
+    - `0`: If there is a zero division, the return value is `0`.
+    - `1`: If there is a zero division, the return value is `1`.
+
+
+### Output Values
+- **recall**(`float`, or `array` of `float`): Either the general recall score, or the recall scores for individual classes, depending on the values input to `labels` and `average`. Minimum possible value is 0. Maximum possible value is 1. A higher recall means that more of the positive examples have been labeled correctly. Therefore, a higher recall is generally considered better.
+
+Output Example(s):
```
sashavor, line 44: Can you explain why the two output structures are different?

sashavor (approved):
Minor things, otherwise all good!

PR #4205, opened 2022-04-22
Fix `convert_file_size_to_int` for kilobits and megabits

mariosasko:
Minor change to fully align this function with the recent change in Transformers (https://github.com/huggingface/transformers/pull/16891)


lhoestq (approved):
nice !

PR #4206, opened 2022-04-22
Add Nerval Metric

mdadda:
This PR adds readme.md and ner_val.py to metrics.
Nerval is a python package that helps evaluate NER models. It creates classification report and confusion matrix at entity level.


albertvillanova:

PR #4212, opened 2022-04-25
[Common Voice] Make sure bytes are correctly deleted if `path` exists

patrickvonplaten:
`path` should be set to local path inside audio feature if exist so that bytes can correctly be deleted.

### DIFF
--- a/datasets/common_voice/common_voice.py
+++ b/datasets/common_voice/common_voice.py
@@ -792,8 +792,9 @@ class CommonVoice(datasets.GeneratorBasedBuilder):
                     result = {key: value for key, value in zip(data_fields, field_values)}
 
                     # set audio feature
+                    path = os.path.join(local_extracted_archive, path) if local_extracted_archive else path
                     result["audio"] = {"path": path, "bytes": f.read()}
                     # set path to None if the audio file doesn't exist locally (i.e. in streaming mode)
-                    result["path"] = os.path.join(local_extracted_archive, path) if local_extracted_archive else None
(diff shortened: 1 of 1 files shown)


datasets/common_voice/common_voice.py
```diff
@@ -792,8 +792,9 @@ def _generate_examples(self, local_extracted_archive, archive_iterator, metadata
                     result = {key: value for key, value in zip(data_fields, field_values)}
 
                     # set audio feature
+                    path = os.path.join(local_extracted_archive, path) if local_extracted_archive else path
                     result["audio"] = {"path": path, "bytes": f.read()}
```
polinaeterna, line 796: maybe to be clearer and not to read bytes when it's non needed even smth like 
```
result["audio"] = {"path": path, "bytes": f.read() if not local_extracted_archive else None}
```
or
```
result["audio"] = {"path": path}
if not local_extracted_archive:
    result["audio"]["bytes"] = f.read()
```

not sure that it's clearer that way though

polinaeterna:
cool that you noticed that we store unnecessary bytes again :D

datasets/common_voice/common_voice.py
lhoestq: I would keep as it is right now.

Reading the bytes doesn't cost anything, since they are actually already in memory from `iter_archive`. And `bytes ` are discarded if the `path` is a local path to avoid duplicating data already (see #4187)

lhoestq (approved):
Thanks for the fix !

PR #4214, opened 2022-04-25
Skip checksum computation in Imagefolder by default

mariosasko:
Avoids having to set `ignore_verifications=True` in `load_dataset("imagefolder", ...)` to skip checksum verification and speed up loading.

The user can still pass `DownloadConfig(record_checksums=True)` to not skip this part.


src/datasets/utils/download_manager.py
```diff
@@ -159,18 +160,26 @@ def __init__(
                 download options
             base_path: `str`, base path that is used when relative paths are used to
                 download files. This can be a remote url.
-            record_checksums: `bool`, whether to record checksums of downloaded files.
+            record_checksums:
+                Deprecated: 'record_checksms' was replaced by 'DownloadConfig.record_checksums' in version 2.1.1 and will be removed in 2.4.0.
```
lhoestq, line 164: (no strong opinion on this)

Maybe we can keep it here ?

Users can still pass a `DownloadManager` to `download_and_prepare` with `record_checksums=True`.
Moreover having `record_checksums` in `DownloadConfig` is maybe not ideal since `DownloadConfig` is passed to `cached_path` that doesn't record checksums. Only the `DownloadManager` does that.

lhoestq (commented):
Thanks ! This will significantly speed up `imagefolder` :)

just one question about record_checksums:

src/datasets/utils/download_manager.py
mariosasko: Okay, that makes sense. Preferably there should be only one way to do things, so I removed `record_checksums` in `DownloadConfig`

lhoestq (approved):
Thanks !

Feel free to merge when the CI is green :)

PR #4216, opened 2022-04-25
Avoid recursion error in map if example is returned as dict value

mariosasko:
I noticed this bug while answering [this question](https://discuss.huggingface.co/t/correct-way-to-create-a-dataset-from-a-csv-file/15686/11?u=mariosasko). 

This code replicates the bug:
```python
from datasets import Dataset
dset = Dataset.from_dict({"en": ["aa", "bb"], "fr": ["cc", "dd"]})
dset.map(lambda ex: {"translation": ex})
```
and this is the fix for it (before this PR):
```python
from datasets import Dataset
dset = Dataset.from_dict({"en": ["aa", "bb"], "fr": ["cc", "dd"]})
dset.map(lambda ex: {"translation": dict(ex)})
```

Internally, this can be fixed by merging two dicts via dict unpacking (instead of `dict.update) `in `Dataset.map`, which avoids creating recursive dictionaries.

P.S. `{**a, **b}` is slightly more performant than `a.update(b)` in my bencmarks.

### DIFF
--- a/src/datasets/features/features.py
+++ b/src/datasets/features/features.py
@@ -339,7 +339,6 @@ def _cast_to_python_objects(obj: Any, only_1d_for_numpy: bool, optimize_list_cas
     elif isinstance(obj, pd.DataFrame):
         return obj.to_dict("list"), True
     elif isinstance(obj, Mapping):
-        obj = obj.data if isinstance(obj, UserDict) else obj
         has_changed = not isinstance(obj, dict)
         output = {}
         for k, v in obj.items():
--- a/src/datasets/features/features.py
+++ b/src/datasets/features/features.py
@@ -337,7 +337,7 @@ def _cast_to_python_objects(obj: Any, only_1d_for_numpy: bool, optimize_list_cas
         return obj.values.tolist(), True
     elif isinstance(obj, pd.DataFrame):
         return obj.to_dict("list"), True
-    elif isinstance(obj, Mapping):
+    elif isinstance(obj, Mapping):  # check for dict-like to handle nested LazyDict objects
         has_changed = not isinstance(obj, dict)
         output = {}
         for k, v in obj.items():
--- a/src/datasets/features/features.py
+++ b/src/datasets/features/features.py
@@ -18,7 +18,6 @@ import copy
 import json
 import re
 import sys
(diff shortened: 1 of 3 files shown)


src/datasets/features/features.py
```diff
@@ -337,9 +337,9 @@ def _cast_to_python_objects(obj: Any, only_1d_for_numpy: bool, optimize_list_cas
         return obj.values.tolist(), True
     elif isinstance(obj, pd.DataFrame):
         return obj.to_dict("list"), True
-    elif isinstance(obj, dict):
+    elif isinstance(obj, Mapping):
+        has_changed = not isinstance(obj, dict)
```
lhoestq, line 341: Can you add a comment to explain why we need this ? I'm not sure I understand why it's related to using `{**inputs, **processed_inputs}`

lhoestq (commented):
Good catch :o 

Just one question:

src/datasets/features/features.py
mariosasko: I think it's easier to understand this with an example. Let's say we are running the following map:
```python
dset.map(lambda ex: {"translation": ex})
```
In this case `inputs` is equal to `ex` and `processed_inputs` to `{"translation": ex}`. Running `update` in the next step, as we are currently doing, creates an object with infinite recursion:
```python
>>> inputs = {"en": "aa", "fr": "bb"}
>>> processed_inputs = {"translation": inputs}
>>> inputs.update(processed_inputs)
>>> inputs
{'en': 'aa', 'fr': 'bb', 'translation': {...}} <- infinite recursion
```
This PR avoids this with the dict unpacking syntax:
```python
>>> inputs = {"en": "aa", "fr": "bb"}
>>> processed_inputs = {"translation": inputs}
>>> {**inputs, **processed_inputs}
{'en': 'aa', 'fr': 'bb', 'translation': {'en': 'aa', 'fr': 'bb'}}
```
However, that's still not enough for scenarios where we turn the input dictionary (`ex` in my example) into the LazyDict object, a subclass of `UserDict`, and return it as a dict value (`translation` value in my example) because PyArrow only supports the built-in `dict` type hence the above check.

PR #4226, opened 2022-04-26
Add pearsonr mc, update functionality to match the original docs

emibaylor:
- adds pearsonr metric card
- adds ability to return p-value
    - p-value was mentioned in the original docs as a return value, but there was no option to return it. I updated the _compute function slightly to have an option to return the p-value.


metrics/pearsonr/README.md
```diff
@@ -0,0 +1,104 @@
+# Metric Card for Pearson Correlation Coefficient (pearsonr)
+
+
+## Metric Description
+
+Pearson correlation coefficient and p-value for testing non-correlation.
+The Pearson correlation coefficient measures the linear relationship between two datasets. The calculation of the p-value relies on the assumption that each dataset is normally distributed. Like other correlation coefficients, this one varies between -1 and +1 with 0 implying no correlation. Correlations of -1 or +1 imply an exact linear relationship. Positive correlations imply that as x increases, so does y. Negative correlations imply that as x increases, y decreases.
```
sashavor, line 7: I would put this part in the ### Output values section:

> Like other correlation coefficients, this one varies between -1 and +1 with 0 implying no correlation. Correlations of -1 or +1 imply an exact linear relationship. Positive correlations imply that as x increases, so does y. Negative correlations imply that as x increases, y decreases.

metrics/pearsonr/README.md
```diff
@@ -0,0 +1,104 @@
...
+['p-value', 'pearsonr']
+>>> print(round(results['pearsonr'], 2))
+-0.74
+>>> print(round(results['p-value'], 2))
+0.15
+```
+
+
+## Limitations and Bias
+
```
sashavor, line 73: I would suggest the following:

"As stated above, the calculation of the p-value relies on the assumption that each data set is normally distributed. This is not always the case, so verifying the true distribution of datasets is recommended."

sashavor (changes requested):
A few minor comments, otherwise LGTM!

sashavor (approved):

metrics/pearsonr/pearsonr.py
```diff
@@ -20,57 +20,64 @@
...
 """
 
+
 _KWARGS_DESCRIPTION = """
 Args:
-    predictions: Predicted labels, as returned by a model.
-    references: Ground truth labels.
+- **predictions** (`list` of `int`): Predicted class labels, as returned by a model.
+- **references** (`list` of `int`): Ground truth labels.
+- **return_pvalue** (`boolean`): If `True`, returns the p-value, along with the correlation coefficient. If `False`, returns only the correlation coefficient. Defaults to `False`.
```
lhoestq, line 32:
```suggestion
    predictions (`list` of `int`): Predicted class labels, as returned by a model.
    references (`list` of `int`): Ground truth labels.
    return_pvalue (`boolean`): If `True`, returns the p-value, along with the correlation coefficient. If `False`, returns only the correlation coefficient. Defaults to `False`.
```

metrics/pearsonr/pearsonr.py
```diff
@@ -20,57 +20,64 @@
...
+
 Returns:
-    pearsonr: Pearson correlation coefficient.
+- **pearsonr** (`float`): Pearson correlation coefficient. Minimum possible value is -1. Maximum possible value is 1. Values of 1 and -1 indicate exact linear positive and negative relationships, respectively. A value of 0 implies no correlation.
+- **p-value** (`float`): P-value, which roughly indicates the probability of an The p-value roughly indicates the probability of an uncorrelated system producing datasets that have a Pearson correlation at least as extreme as the one computed from these datasets. Minimum possible value is 0. Maximum possible value is 1. Higher values indicate higher probabilities.
```
lhoestq, line 36:
```suggestion
    pearsonr (`float`): Pearson correlation coefficient. Minimum possible value is -1. Maximum possible value is 1. Values of 1 and -1 indicate exact linear positive and negative relationships, respectively. A value of 0 implies no correlation.
    p-value (`float`): P-value, which roughly indicates the probability of an The p-value roughly indicates the probability of an uncorrelated system producing datasets that have a Pearson correlation at least as extreme as the one computed from these datasets. Minimum possible value is 0. Maximum possible value is 1. Higher values indicate higher probabilities.
```

lhoestq (approved):
Thanks :) as before, I just added suggestions for the docstrings formats:

sashavor:
thank you @lhoestq!! :hugs:

PR #4231, opened 2022-04-27
Fix invalid url to CC-Aligned dataset

juntang-zhuang:
The CC-Aligned dataset url has changed to  https://data.statmt.org/cc-aligned/, the old address http://www.statmt.org/cc-aligned/ is no longer valid


mariosasko (approved):
Thanks! LGTM!

(I've rerun the CI to pass the failing test.)

PR #4234, opened 2022-04-27
Autoeval config

nrajani:
Added autoeval config to imdb as pilot


lewtun:
Related to: https://github.com/huggingface/autonlp-backend/issues/414

datasets/imdb/README.md
```diff
@@ -3,6 +3,30 @@ pretty_name: IMDB
...
+train-eval-index:
+- config: plain_text
+  task: text-classification
+  task_id: binary_classification
+  splits:
+    train_split: train
+    eval_split: test
+  col_mapping:
+    text: text
+    label: label
```
lewtun, line 15: The column mapping goes from {source columns} -> {AutoTrain columns}, so we need to use `target` here:

```suggestion
    label: target
```

datasets/imdb/README.md
```diff
@@ -3,6 +3,30 @@ pretty_name: IMDB
...
+  metrics:
+  - type: accuracy
```
lewtun, line 17: I suggest adding a "pretty name" for all metrics so we can ultimately render this in AutoTrain
```suggestion
  - type: accuracy
  - name: Accuracy
```

datasets/imdb/README.md
```diff
@@ -3,6 +3,30 @@ pretty_name: IMDB
...
+  - type: f1
+    name: f1_macro
+    args:
+      average: macro
+  - type: f1
+    name: f1_micro
+    args:
+      average: micro  
+  - type: f1
```
lewtun, line 26: Would it make sense to also include:

- precision (macro, micro, weighted)
- recall (macro, micro, weighted)

datasets/imdb/README.md
```diff
@@ -3,6 +3,30 @@ pretty_name: IMDB
...
+  splits:
+    train_split: train
+    eval_split: test
+  col_mapping:
+    text: text
+    label: label
+  metrics:
+  - type: accuracy
+  - type: f1
+    name: f1_macro
```
lewtun, line 19: Perhaps it's better to use a "pretty name" here (and the other F1 metrics)?
```suggestion
    name: F1 macro
```

lewtun (approved):
Thanks for kicking off the metadata additions for evaluation @nrajani 🚀 ! I've left a few small nits, but otherwise this LGTM!

lewtun:
The tests are failing due to the changed metadata:

```
got an unexpected keyword argument 'train-eval-index'
```

I think you can fix this by updating the `DatasetMetadata` class and implementing an appropriate `validate_train_eval_index()` function

lewtun:
On the question of validating the `train-eval-index` metadata, I think the simplest approach would be to validate that the required fields exist and not worry about their values (which are open-ended).

For me, the required fields include:

* `config`
* `task`
* `task_id`
* `splits` (train / validation / eval)
* `col_mapping`
* `metrics` (checking that each one has `type`, `name`) 

Here I'm using the spec defined in https://github.com/huggingface/autonlp-backend/issues/414 as a guide.

WDYT @lhoestq ?

lhoestq:
Makes sense ! Currently the metadata type validator doesn't support subfields - let me open a PR to add it

lhoestq:
I ended up improving the metadata validation in this PR x)

In particular:
- I added support YAML keys with dashes instead of underscores for `train-eval-index`
- I added `train-eval-index` validation (nothing fancy, it just checks that it is a list if it exists in the YAML, but feel free to improve `validate_train_eval_index` if you want)

Let me know if it sounds good to you ! I think we can improve `validate_train_eval_index` in another PR and merge this one

lewtun (approved):
Thanks for add a special validator for `train-eval-index` @lhoestq ! I think this approach looks great 🤗

lewtun:
> Come on windows... I didn't do anything advanced...
> 
> Anyway, will try to fix this when I get back home x)

Hehe, thanks!

lhoestq:
Did I just fix it for windows and now it fails on linux ? xD

lewtun:
> Did I just fix it for windows and now it fails on linux ? xD

Looks like the Heisenberg uncertainty principle is at play here - you cannot simultaneously have unit tests passing in both Linux and Windows 😅

lhoestq:
The worst is that the tests pass locally both on my windows and my linux x)

lhoestq:
Ok fixed it, the issue came from python 3.6 that doesn't return the right `__origin__` for Dict and List types

datasets/imdb/README.md
```diff
@@ -19,6 +19,55 @@ task_categories:
 task_ids:
 - sentiment-classification
 paperswithcode_id: imdb-movie-reviews
+train-eval-index:
+- config: plain_text
+  task: text-classification
+  task_id: binary_classification
```
lhoestq, line 25: for reference, the task_id is the autotrain one, they're defined here (private) : https://github.com/huggingface/autonlp-backend/tree/main/zeus/zeus/models

lhoestq (approved):
Alright thanks for adding the first Autoeval config ! :D

lewtun:
> Alright thanks for adding the first Autoeval config ! :D

Woohoo! Thank you so much 🤗

fxmarty:
This is cool!

PR #4236, opened 2022-04-27
Replace data URL in big_patent dataset and support streaming

### LINKED ISSUE #4217: Big_Patent dataset broken
## Dataset viewer issue for '*big_patent*'

**Link:** *[link to the dataset viewer page](https://huggingface.co/datasets/big_patent/viewer/all/train)*

*Unable to view because it says FileNotFound, also cannot download it through the python API*

Am I the one who added this dataset ? No

albertvillanova:
This PR replaces the Google Drive URL with by our Hub one, once the data owners have approved to host their data on the Hub.

Moreover, this PR makes the dataset streamable.

Fix #4217.


albertvillanova:
I first uploaded the data files to the Hub: I think it is a good option because we have git lfs to track versions and changes. Moreover people will be able to make PRs to propose updates on the data files.
- I would have preferred to upload it it to the "data" org namespace, but it is already taken (although not used): might be possible to take it?

As an alternative (and to be consistent with previous datasets), I also uploaded the data files to our AWS bucket.

We should decide which to use (now and for future datasets) and set it here before merging. We should remove the data files for the non-chosen option.

CC: @lhoestq @mariosasko @polinaeterna

mariosasko:
I also think storing everything in one place/single repository is the best option.

@polinaeterna Canonical datasets also support data files (see the [`red_caps` repo](https://huggingface.co/datasets/red_caps/tree/main) for instance)

PR #4240, opened 2022-04-27
Fix yield for crd3

shanyas10:
Modified the `_generate_examples` function to consider all the turns for a chunk id as a single example
Modified the features accordingly

```
"turns": [
                        {
                            "names": datasets.features.Sequence(datasets.Value("string")),
                            "utterances": datasets.features.Sequence(datasets.Value("string")),
                            "number": datasets.Value("int32"),
                        }
                    ],
                }
```

I wasn't able to run `datasets-cli dummy_data datasets` command. Is there a workaround for this?


datasets/crd3/crd3.py
```diff
@@ -135,19 +135,20 @@ def _generate_examples(self, files_path):
...
-                            "turn_start": turn_start,
-                            "turn_end": turn_end,
-                            "alignment_score": score,
-                            "turn_num": turn_num,
-                            "turns": {
-                                "names": turn_names,
-                                "utterances": turn_utterances,
-                            },
-                        }
+                    for turn in enumerate(row["TURNS"]):
```
thomasw21, line 138:
```suggestion
                    for turn in row["TURNS"]:
```
I think this is it.

shanyas10: Thanks @thomasw21. I pushed a fix in the recent commit :)

PR #4244, opened 2022-04-27
task id update

nrajani:
changed multi input text classification as task id instead of category


nrajani:
Reverted the multi-input-text-classification tag from task_categories and added it as task_ids @lhoestq

lhoestq (approved):
Thanks ! :)

Issue #4238, opened 2022-04-27
Dataset caching policy


loretoparisi:
## Describe the bug
I cannot clean cache of my datasets files, despite I have updated the `csv` files on the repository [here](https://huggingface.co/datasets/loretoparisi/tatoeba-sentences). The original file had a line with bad characters, causing the following error

```
[/usr/local/lib/python3.7/dist-packages/datasets/features/features.py](https://localhost:8080/#) in str2int(self, values)
    852                 if value not in self._str2int:
    853                     value = str(value).strip()
--> 854                 output.append(self._str2int[str(value)])
    855             else:
    856                 # No names provided, try to integerize

KeyError: '\\N'
```

The file now is cleanup up, but I still get the error. This happens even if I inspect the local cached contents, and cleanup the files locally:

```python
from datasets import load_dataset_builder
dataset_builder = load_dataset_builder("loretoparisi/tatoeba-sentences")
print(dataset_builder.cache_dir)
print(dataset_builder.info.features)
print(dataset_builder.info.splits)
```

```
Using custom data configuration loretoparisi--tatoeba-sentences-e59b8ad92f1bb8dd
/root/.cache/huggingface/datasets/csv/loretoparisi--tatoeba-sentences-e59b8ad92f1bb8dd/0.0.0/433e0ccc46f9880962cc2b12065189766fbb2bee57a221866138fb9203c83519
None
None
```

and removing files located at `/root/.cache/huggingface/datasets/csv/loretoparisi--tatoeba-sentences-*`.
 Is there any remote file caching policy in place? If so, is it possibile to programmatically disable it? 
Currently it seems that the file `test.csv` on the repo [here](https://huggingface.co/datasets/loretoparisi/tatoeba-sentences/blob/main/test.csv) is cached remotely. In fact I download locally the file from raw link, the file is up-to-date; but If I use it within `datasets` as shown above, it gives to me always the first revision of the file, not the last.

Thank you.

## Steps to reproduce the bug
```python
from datasets import load_dataset,Features,Value,ClassLabel

[...]

PR #4246, opened 2022-04-28
Support to load dataset with TSV files by passing only dataset name

albertvillanova:
This PR implements support to load a dataset (w/o script) containing TSV files by passing only the dataset name (no need to pass `sep='\t'`):
```python
ds = load_dataset("dataset/name")
```

The refactoring allows for future builder kwargs customizations based on file extension.

Related to #4238.


lhoestq (approved):
Cool thanks !

Feel free to also update the type hints of `infer_module_for_data_files` and `infer_module_for_data_files_in_archives`, and maybe add some docstrings to explain that the second output corresponds to builder kwargs :)

PR #4260, opened 2022-04-30
Add mr_polarity movie review sentiment classification

mo6zes:
Add the MR (Movie Review) dataset. The original dataset contains sentences from Rotten Tomatoes labeled as either "positive" or  "negative". 

Homepage: [https://www.cs.cornell.edu/people/pabo/movie-review-data/](https://www.cs.cornell.edu/people/pabo/movie-review-data/)
paperswithcode: [https://paperswithcode.com/dataset/mr](https://paperswithcode.com/dataset/mr)

- [ ] I was not able to generate dummy data, the original dataset files have ".pos" and ".neg" as file extensions so the auto-generator does not work. Is it fine like this or should dummy data be added?


datasets/mr_polarity/mr_polarity.py
```diff
@@ -0,0 +1,159 @@
...
+                },
+            )
+        ]
+
+    # method parameters are unpacked from `gen_kwargs` as given in `_split_generators`
+    def _generate_examples(self, folder, split):
+        # TODO: This method handles input defined in _split_generators to yield (key, example) tuples from the dataset.
+        # The `key` is for legacy reasons (tfds) and is not important in itself, but must be unique for each example.
+
+        def clean_line(line):
```
mo6zes, line 126: The original dataset contains whitespaces before punctuation e.g. "simplistic , silly and tedious . ". I tried cleaning it with a regex, but I'm curious whether there is a better solution?

mo6zes:
whoops just found https://huggingface.co/datasets/rotten_tomatoes

PR #4262, opened 2022-05-01
Add YAML tags to Dataset Card rotten tomatoes

mo6zes:
The dataset card for the rotten tomatoes / MR movie review dataset had some missing YAML tags. Hopefully, this also improves the visibility of this dataset now that paperswithcode and huggingface link to eachother.


albertvillanova (approved):
Thank you.

PR #4263, opened 2022-05-02
Rename imagenet2012 -> imagenet-1k

lhoestq:
On the Hugging Face Hub, users refer to imagenet2012 (from #4178 ) as imagenet-1k in their model tags.

To correctly link models to imagenet, we should rename this dataset `imagenet-1k`.

Later we can add `imagenet-21k` as a new dataset if we want.

Once this one is merged we can delete the `imagenet2012` dataset repository on the Hub.

EDIT: to complete the rationale on why we should name it `imagenet-1k`:
If users specifically added the tag `imagenet-1k` , then it could be for two reasons (not sure which one is predominant), either they
- wanted to make it explicit that it’s not 21k -> the distinction is important for the community
- or they have been following this convention from other models -> the convention implicitly exists already


julien-c:
> Later we can add imagenet-21k as a new dataset if we want.

isn't it what models refer to as `imagenet` already?

lhoestq:
> isn't it what models refer to as imagenet already?

I wasn't sure, but it looks like it indeed. Therefore having a dataset `imagenet` for ImageNet 21k makes sense actually.

apsdehal (approved):
LGTM!

julien-c:
should we remove the repo mirror on the hub side or will you do it?

PR #4280, opened 2022-05-04
Add missing features to commonsense_qa dataset

### LINKED ISSUE #4275: CommonSenseQA has missing and inconsistent field names
## Describe the bug
In short, CommonSenseQA implementation is inconsistent with the original datasets.

More precisely, we need to:

1. Add a dataset matching the "id" field. The current dataset, instead, regenerates monotonically increasing id.  
2. The [“question”][“stem”] field is flattened into "question". We should match the original dataset and unflatten it
3. Add the missing "question_concept" field in the question tree node
4. Anything else? Go over the data structure of the newly repaired CommonSenseQA and make sure it matches the original

## Expected results
Every data item of the CommonSenseQA should structurally and data-wise match the original CommonSenseQA dataset.

## Actual results
TBD

## Environment info
- `datasets` version: 2.1.0
- Platform: macOS-10.15.7-x86_64-i386-64bit
- Python version: 3.8.13
- PyArrow version: 7.0.0
- Pandas version: 1.4.2

albertvillanova:
Fix partially #4275.


vblagoje:
@albertvillanova it adds question_concept and id which is great. I suppose we'll talk about staying true to the format on another PR.

albertvillanova:
Yes, let's merge this PR as it is: it adds missing features.

A subsequent PR may address the request on changing the dataset feature structure.

PR #4281, opened 2022-05-04
Remove a copy-paste sentence in dataset cards

albertvillanova:
Remove the following copy-paste sentence from dataset cards:
```
We show detailed information for up to 5 configurations of the dataset.
```


albertvillanova:
The non-passing tests have nothing to do with this PR.

PR #4286, opened 2022-05-05
Add Lahnda language tag

mariosasko:
This language is present in [Wikimedia's WIT](https://huggingface.co/datasets/wikimedia/wit_base) dataset.


lhoestq (approved):
thanks :)

Issue #4287, opened 2022-05-05
"NameError: name 'faiss' is not defined" on `.add_faiss_index` when `device` is not None


alvarobartt:
## Describe the bug

When using `datasets` to calculate the FAISS indices of a dataset, the exception `NameError: name 'faiss' is not defined` is triggered when trying to calculate those on a device (GPU), so `.add_faiss_index(..., device=0)` fails with that exception.

All that assuming that `datasets` is properly installed and `faiss-gpu` too, as well as all the CUDA drivers required.

## Steps to reproduce the bug

```python
# Sample code to reproduce the bug
from transformers import DPRContextEncoder, DPRContextEncoderTokenizer
import torch
torch.set_grad_enabled(False)
ctx_encoder = DPRContextEncoder.from_pretrained("facebook/dpr-ctx_encoder-single-nq-base")
ctx_tokenizer = DPRContextEncoderTokenizer.from_pretrained("facebook/dpr-ctx_encoder-single-nq-base")

from datasets import load_dataset
ds = load_dataset('crime_and_punish', split='train[:100]')
ds_with_embeddings = ds.map(lambda example: {'embeddings': ctx_encoder(**ctx_tokenizer(example["line"], return_tensors="pt"))[0][0].numpy()})

ds_with_embeddings.add_faiss_index(column='embeddings', device=0) # default `device=None`
```

## Expected results

A new column named `embeddings` in the dataset that we're adding the index to.

## Actual results

An exception is triggered with the following message `NameError: name 'faiss' is not defined`.

## Environment info

- `datasets` version: 2.1.0
- Platform: Linux-5.13.0-1022-azure-x86_64-with-glibc2.31
- Python version: 3.9.12
- PyArrow version: 7.0.0
- Pandas version: 1.4.2

alvarobartt:
Adding here the complete error traceback!

```
Traceback (most recent call last):
  File "/home/alvarobartt/lol.py", line 12, in <module>
    ds_with_embeddings.add_faiss_index(column='embeddings', device=0) # default `device=None`
  File "/home/alvarobartt/.local/lib/python3.9/site-packages/datasets/arrow_dataset.py", line 3656, in add_faiss_index
    super().add_faiss_index(
  File "/home/alvarobartt/.local/lib/python3.9/site-packages/datasets/search.py", line 478, in add_faiss_index
    faiss_index.add_vectors(self, column=column, train_size=train_size, faiss_verbose=True)
[...]

PR #4288, opened 2022-05-05
Add missing `faiss` import to fix https://github.com/huggingface/datasets/issues/4287

alvarobartt:
This PR fixes the issue recently mentioned in https://github.com/huggingface/datasets/issues/4287 🤗


lhoestq (approved):
good catch thanks !

PR #4292, opened 2022-05-06
Add API code examples for remaining main classes

stevhliu:
This PR adds API code examples for the remaining functions in the Main classes. I wasn't too familiar with some of the functions (`decode_batch`, `decode_column`, `decode_example`, etc.) so please feel free to add an example of usage and I can fill in the rest :)

### DIFF
--- a/src/datasets/features/image.py
+++ b/src/datasets/features/image.py
@@ -41,6 +41,16 @@ class Image:
     Args:
         decode (:obj:`bool`, default ``True``): Whether to decode the image data. If `False`,
             returns the underlying dictionary in the format {"path": image_path, "bytes": image_bytes}.
+
+    Examples:
+
+    ```py
+    >>> from datasets import load_dataset, Image
+    >>> ds = load_dataset("beans", split="train")
+    >>> ds = ds.cast_column('image', Image(decode=False))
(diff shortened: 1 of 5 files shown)


src/datasets/features/image.py
```diff
@@ -41,6 +41,16 @@ class Image:
     Args:
         decode (:obj:`bool`, default ``True``): Whether to decode the image data. If `False`,
             returns the underlying dictionary in the format {"path": image_path, "bytes": image_bytes}.
+
+    Examples:
+
+    ```py
+    >>> from datasets import load_dataset, Image
+    >>> ds = load_dataset("beans", split="train")
+    >>> ds = ds.cast_column('image', Image(decode=False))
```
lhoestq, line 50: I think we can also show the PIL Image object
```suggestion
    >>> ds.features["image"]
    Image(decode=True, id=None)
    >>> ds[0]["image"]
    <PIL.JpegImagePlugin.JpegImageFile image mode=RGB size=500x500 at 0x15E52E7F0>
    >>> ds = ds.cast_column('image', Image(decode=False))
    >>> ds[0]["image"]
```

lhoestq (approved):
Thanks !

PR #4293, opened 2022-05-08
Fix wrong map parameter name in cache docs

h4iku:
The `load_from_cache` parameter of `map` should be `load_from_cache_file`.


mariosasko (approved):
Thanks! The CI failure is unrelated to the changes, so feel free to ignore it.

lhoestq (approved):
thanks !

PR #4299, opened 2022-05-09
Remove manual download from imagenet-1k

mariosasko:
Remove the manual download code from `imagenet-1k` to make it a regular dataset.


datasets/imagenet-1k/imagenet-1k.py
```diff
@@ -41,141 +40,94 @@
...
+            if split in path:
+                file_root, file_ext = os.path.splitext(path)
+                if file_ext.lower() == self.IMAGE_EXTENSION:
+                    output = {"image": {"path": path, "bytes": file.read()}}
+
+                    if split == "train":
+                        label = os.path.basename(os.path.dirname(file_root))
+                        output["label"] = IMAGENET2012_CLASSES[label]
+                    elif split == "val":
+                        label = fixes.get(os.path.basename(file_root))
```
apsdehal, line 126: When will this actually be None. This will actually `KeyError` if the key is not present. Instead, I suggest you do this:
```suggestion
                        label = fixes.get(os.path.basename(file_root), None)
```

lhoestq: It already returns None by default ;) no need to specify it
```py
>>> {}.get(1) is None
True
```

datasets/imagenet-1k/imagenet-1k.py
```diff
@@ -41,141 +40,94 @@
 ILSVRC 2012, commonly known as 'ImageNet' is an image dataset organized according to the WordNet hierarchy. Each meaningful concept in WordNet, possibly described by multiple words or word phrases, is called a "synonym set" or "synset". There are more than 100,000 synsets in WordNet, majority of them are nouns (80,000+). ImageNet aims to provide on average 1000 images to illustrate each synset. Images of each concept are quality-controlled and human-annotated. In its completion, ImageNet hopes to offer tens of millions of cleanly sorted images for most of the concepts in the WordNet hierarchy. ImageNet 2012 is the most commonly used subset of ImageNet. This dataset spans 1000 object classes and contains 1,281,167 training images, 50,000 validation images and 100,000 test images
 """
 
-_IN1K_NUM_CLASSES = 1000
+_DATA_URL = (
+    "https://huggingface.co/datasets/imagenet-1k/resolve/main/data/imagenet_object_localization_patched2019.tar.gz"
+)
```
lhoestq, line 45: I think it would be nice to separate the train/val/test splits in different files, this way you don't have to download gigabytes of data before being able to stream the first image of certain splits.

lhoestq (commented):
Cool thanks :)

I think we can separate the data files for each split, it will be more convenient for streaming this way.

mariosasko:
@apsdehal I dismissed your review as it's no longer relevant after the data files changes suggested by @lhoestq.

lhoestq (approved):
Amazing thanks !

datasets/imagenet-1k/README.md
```diff
@@ -19,6 +19,16 @@ task_categories:
 - image-classification
 task_ids:
 - multi-class-image-classification
+extra_gated_prompt: |
+  By clicking on “Access repository” below, you also agree to ImageNet Terms of Access:
+  [RESEARCHER_FULLNAME] (the "Researcher") has requested permission to use the ImageNet database (the "Database") at Princeton University and Stanford University. In exchange for such permission, Researcher hereby agrees to the following terms and conditions:
+  1. Researcher shall use the Database only for non-commercial research and educational purposes.
+  2. Princeton University and Stanford University make no representations or warranties regarding the Database, including but not limited to warranties of non-infringement or fitness for a particular purpose.
```
lhoestq, line 26:
```suggestion
  2. Princeton University, Stanford University and Hugging Face make no representations or warranties regarding the Database, including but not limited to warranties of non-infringement or fitness for a particular purpose.
```

PR #4305, opened 2022-05-10
Fixes FrugalScore

moussaKam:
There are two minor modifications in this PR:
1) `predictions` and `references` are swapped. Basically Frugalscore is commutative, however some tiny differences can occur if we swap the references and the predictions. I decided to swap them just to obtain the exact results as reported in the paper.
2) I switched to dynamic padding that was was used in the training, forcing the padding to `max_length`  introduces errors for some reason that I ignore.

@lhoestq

### DIFF
--- a/metrics/frugalscore/frugalscore.py
+++ b/metrics/frugalscore/frugalscore.py
@@ -101,12 +101,15 @@ class FRUGALSCORE(datasets.Metric):
             no_cuda=(device == "cpu"),
             log_level="warning",
         )
-        dataset = {"sentence1": predictions, "sentence2": references}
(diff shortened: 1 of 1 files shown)


lhoestq:
> predictions and references are swapped. Basically Frugalscore is commutative, however some tiny differences can occur if we swap the references and the predictions. I decided to swap them just to obtain the exact results as reported in the paper.

What is the order of magnitude of the difference ? Do you know what causes this ?

> I switched to dynamic padding that was was used in the training, forcing the padding to max_length introduces errors for some reason that I ignore.

What error ?

moussaKam:
1) Here's an example:

PR #4308, opened 2022-05-10
Remove unused multiprocessing args from test CLI

albertvillanova:
Multiprocessing is not used in the test CLI.


lhoestq (approved):
Thanks ! this was indeed unnecessary

PR #4309, opened 2022-05-10
[WIP] Add TEDLIUM dataset

sanchit-gandhi:
Adds the TED-LIUM dataset https://www.tensorflow.org/datasets/catalog/tedlium#tedliumrelease3 

TODO:

- [x] Port `tedium.py` from TF datasets using `convert_dataset.sh` script
- [x] Make `load_dataset` work
- [ ] ~~Run `datasets-cli` command to generate `dataset_infos.json`~~
- [ ] ~~Create dummy data for continuous testing~~
- [ ] ~~Dummy data tests~~
- [ ] ~~Real data tests~~
- [ ] Create the metadata JSON
- [ ] Close PR and add directly to the Hub under LIUM org

### DIFF
--- a/datasets/tedlium/README.md
+++ b/datasets/tedlium/README.md
@@ -84,20 +84,20 @@ The audio and transcriptions are in English, as per the TED talks at http://www.
 
 ### Data Instances
 ```
-{'audio': {'path': '/home/sanchitgandhi/cache/downloads/extracted/6e3655f9e735ae3c467deed1df788e0dabd671c1f3e2e386e30aa3b571bd9761/TEDLIUM_release1/train/stm/PaulaScher_2008P.stm', 
+{'audio': {'path': '/home/sanchitgandhi/cache/downloads/extracted/6e3655f9e735ae3c467deed1df788e0dabd671c1f3e2e386e30aa3b571bd9761/TEDLIUM_release1/train/sph/PaulaScher_2008P.sph', 
   'array': array([-0.00048828, -0.00018311, -0.00137329, ...,  0.00079346,
           0.00091553,  0.00085449], dtype=float32),
   'sampling_rate': 16000},
 'text': '{COUGH} but <sil> i was so {COUGH} utterly unqualified for(2) this project and {NOISE} so utterly ridiculous {SMACK} and ignored the brief {SMACK} <sil>', 
 'speaker_id': 'PaulaScher_2008P', 
 'gender': 'female', 
-'file': '/home/sanchitgandhi/cache/downloads/extracted/6e3655f9e735ae3c467deed1df788e0dabd671c1f3e2e386e30aa3b571bd9761/TEDLIUM_release1/train/stm/PaulaScher_2008P.stm', 
+'file': '/home/sanchitgandhi/cache/downloads/extracted/6e3655f9e735ae3c467deed1df788e0dabd671c1f3e2e386e30aa3b571bd9761/TEDLIUM_release1/train/sph/PaulaScher_2008P.sph', 
 'id': 'PaulaScher_2008P-1003.35-1011.16-<o,f0,female>'}
 ```
 ### Data Fields
 
 - audio: A dictionary containing the path to the downloaded audio file, the decoded audio array, and the sampling rate. Note that when accessing the audio column: `dataset[0]["audio"]` the audio file is automatically decoded and resampled to `dataset.features["audio"].sampling_rate`. Decoding and resampling of a large number of audio files might take a significant amount of time. Thus it is important to first query the sample index before the `"audio"` column, *i.e.* `dataset[0]["audio"]` should **always** be preferred over `dataset["audio"][0]`. 
-- file: A path to the downloaded audio file in .sth format.
+- file: A path to the downloaded audio file in .sph format.
 - text: the transcription of the audio file.
 - gender: the gender of the speaker. One of: male, female or N/A.
 - id: unique id of the data sample.
@@ -106,7 +106,7 @@ The audio and transcriptions are in English, as per the TED talks at http://www.
 ### Data Splits
 There are three releases for the TED-LIUM corpus, progressively increasing the number of transcribed speech training data from 118 hours (Release 1), to 207 hours (Release 2), to 452 hours (Release 3).
 
-Release 1 (default config):
+Release 1:
 - 774 audio talks and automatically aligned transcriptions.
 - Contains 118 hours of speech audio data.
 - Homepage: https://www.openslr.org/7/
@@ -126,6 +126,10 @@ Release 3:
 - Selected monolingual data for language modeling from WMT12 publicly available corpora: these files come from the TED-LIUM 2 release, but have been modified to produce a tokenization more relevant for English language.
 - Homepage: https://www.openslr.org/51/
 
+Release 3 contains two different corpus distributions:
+- The ‘legacy’ one, on which the dev and test datasets are the same as in TED-LIUM 2 (and TED-LIUM 1).
+- The ‘speaker adaptation’ one, specially designed for experiments on speaker adaptation.
+
 Each release is split into a training, validation and test set:
 
 | Split      | Release 1 | Release 2 | Release 3 |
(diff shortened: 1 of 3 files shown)


sanchit-gandhi:
```python

ds = load_dataset('./datasets/tedlium', 'release1', cache_dir='/home/sanchitgandhi/cache')
```

```
Downloading and preparing dataset tedlium/release1 to /home/sanchitgandhi/cache/tedlium/release1/1.0.1/5a9fcb97b4b52d5a1c9dc7bde4b1d5994cd89c4a3425ea36c789bf6096fee4f0...
  File "/home/sanchit_huggingface_co/datasets/src/datasets/load.py", line 1703, in load_dataset
  File "/home/sanchit_huggingface_co/datasets/src/datasets/builder.py", line 605, in download_and_prepare
    self._download_and_prepare(
  File "/home/sanchit_huggingface_co/datasets/src/datasets/builder.py", line 1240, in _download_and_prepare
    raise MissingBeamOptions(
datasets.builder.MissingBeamOptions: Trying to generate a dataset using Apache Beam, yet no Beam Runner or PipelineOptions() has been provided in `load_dataset` or in the builder arguments. For big datasets it has to run on large-scale data processing tools like Dataflow, Spark, etc. More information about Apache Beam runners at https://beam.apache.org/documentation/runners/capability-matrix/
If you really want to run it locally because you feel like the Dataset is small enough, you can use the local beam runner called `DirectRunner` (you may run out of memory). 
Example of usage: 
        `load_dataset('tedlium', 'release1', beam_runner='DirectRunner')`
```
Specifying the `beam_runner='DirectRunner'` works:
```python

ds = load_dataset('./datasets/tedlium', 'release1', cache_dir='home/sanchitgandhi/cache', beam_runner='DirectRunner')
```

sanchit-gandhi:
Extra Python imports/Linux packages:
```
pip install pydub
sudo apt install ffmpeg
```

sanchit-gandhi:
Heavily inspired by the TF datasets script at: https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/audio/tedlium.py

The TF datasets script uses the module AudioSegment from the package `pydub` (https://github.com/jiaaro/pydub), which is used to to open the audio files (stored in .sph format):
https://github.com/huggingface/datasets/blob/61bf6123634bf6e7c7287cd6097909eb26118c58/datasets/tedlium/tedlium.py#L167-L170
This package requires the pip install of `pydub` and the system installation of `ffmpeg`: https://github.com/jiaaro/pydub#installation
Is it ok to use this import? Or do we tend to avoid introducing additional dependencies?

The TF datasets script also uses `_build_pcollection`:
https://github.com/huggingface/datasets/blob/8afbbb6fe66b40d05574e2e72e65e974c72ae769/datasets/tedlium/tedlium.py#L200-L206
However, I was advised against using `beam` logic. Thus, I have reverted to generating the examples file-by-file: https://github.com/huggingface/datasets/blob/61bf6123634bf6e7c7287cd6097909eb26118c58/datasets/tedlium/tedlium.py#L112-L138

I am now able to generate examples by running the `load_dataset` command:
```python

ds = load_dataset('./datasets/tedlium', 'release1', cache_dir='/home/sanchitgandhi/cache')
```

Generating the examples is **extremely** slow: it takes ~1 second per example, so ~60k seconds for the train set. Is there a way of paralleling this to make it faster?

lhoestq:
> This package requires the pip install of pydub and the system installation of ffmpeg: https://github.com/jiaaro/pydub#installation
Is it ok to use these packages? Or do we tend to avoid introducing additional dependencies?

It's ok, windows users will have have a bad time but I'm not sure we can do much about it.

> Here, generating examples is extremely slow: it takes ~1 second per example, so ~60k seconds for the train set (~16 hours). Is there a way of paralleling this to make it faster?

Not at the moment. For such cases we advise hosting the dataset ourselves in a processed format. The license doesn't allow this since the license is "NoDerivatives". Currently the only way to parallelize it is by keeping is as a beam dataset and let users pay Google Dataflow to process it (or use spark or whatever).

mariosasko:
> This package requires the pip install of pydub and the system installation of ffmpeg: https://github.com/jiaaro/pydub#installation
Is it ok to use these packages? Or do we tend to avoid introducing additional dependencies?

`soundfile`, which is a required audio dependency, should also work with `.sph` files, no?

sanchit-gandhi:
> `soundfile`, which is a required audio dependency, should also work with `.sph` files, no?

Awesome, thanks for the pointer @mariosasko! Switched `pydub` to `soundfile`, and having specifying the `dtype` argument in `soundfile.read` as `np.int16`, the arrays match with those from `pydub` ✅

I also did some heavy optimising of the script with the processing of the `.stm` and `.sph` files - it now runs 2000x faster than before, so there probably isn't a need to upload the data to the Hub @lhoestq. The total processing time is just ~2mins now 🚀

datasets/tedlium/tedlium.py
```diff
@@ -0,0 +1,271 @@
...
+            features=features,
+            supervised_keys=("audio", "text"),
+            homepage=self.config.url,
+            license=_LICENSE,
+            citation=self.config.citation,
+            task_templates=[AutomaticSpeechRecognition(audio_column="audio", transcription_column="text")],
+        )
+
+    def _split_generators(self, dl_manager):
+        data_dir = dl_manager.download_and_extract(self.config.download_url)
```
lhoestq, line 195: To make the dataset streaming work with TAR archives, you have to use `dl_manager.iter_archive`, not `download_and_extract`, see librispeech_asr for example:

https://github.com/huggingface/datasets/blob/095d12ff7414df118f60e00cd6494299a881743a/datasets/librispeech_asr/librispeech_asr.py#L116-L130

datasets/tedlium/README.md
```diff
@@ -0,0 +1,206 @@
...
+multilinguality:
+- monolingual
+size_categories:
+- 10K<n<100K
+source_datasets:
+- original
+task_categories:
+- automatic-speech-recognition
+task_ids:
+- automatic-speech-recognition
```
lhoestq, line 20: no need to be redundant with `automatic-speech-recognition` ;)
```suggestion
task_ids: []
```

lhoestq (commented):
That's super good news ! amazing :)

I added a comment in case you wanted to also make the dataset work in streaming mode.

Also you can add dummy data at `datasets/tedlium/dummy/release1/1.0.1/dummy_data.zip` if you'd like to have the dataset tested in the CI.

It must contain a directory named after the TAR archive that has to be downloaded (including the ".tgz" suffix), and in this directory you can place two or three dummy audio files (using the same name or structure as the original TAR archive).

datasets/tedlium/tedlium.py
sanchit-gandhi: Hey @lhoestq, I making good progress! The issue I’m facing is that the data for all three splits (train, dev and test) is downloaded from one `.tar` file at http://www.openslr.org/resources/19/TEDLIUM_release2.tar.gz. The subsets are separated into sub-directories when the file is extracted (`TEDLIUM_release1/train`, `TEDLIUM_release1/dev`, `TEDLIUM_release1/test`). Thus, in defining `archive_path` as the `dl_manger.download`, the variable points to **all three subsets**, and the iterator `filepath` a generator object over the data for all three subsets: 
https://github.com/huggingface/datasets/blob/18ff32a52a857af4cfa997d9e192844478e11531/datasets/tedlium/tedlium.py#L194-L208
This means that in `_generate_examples`, the iterator `filepath` has to go through every example and check if it belongs to the correct split:
https://github.com/huggingface/datasets/blob/18ff32a52a857af4cfa997d9e192844478e11531/datasets/tedlium/tedlium.py#L246-L249
Does the `dl_manager` have functionality to produce an iterator for a specific subset of the data extracted from the `.tar` file?

lhoestq: Feel free to host the data on the Hub and separate the splits in different files :)

sanchit-gandhi:
TEDLIUM completed and uploaded to the HF Hub: https://huggingface.co/datasets/LIUM/tedlium

lhoestq:
Awesome !

PR #4321, opened 2022-05-11
Adding dataset enwik8

HallerPatrick:
Because I regularly work with enwik8, I would like to contribute the dataset loader 🤗


lhoestq (commented):
Cool thanks for adding this dataset :)

Before merging it would be nice to include some dummy data for the dataset to be tested properly.

Here are a few steps to create them:
- create a directory `datasets/enwik8/dummy/enwik8/1.1.0/dummy_data`
- inside the directory, create another dir `enwik8.zip` (yes, a directory, not an actual ZIP archive)
- inside `enwik8.zip`, place two or three .raw files
- zip the `dummy_data` folder to `datasets/enwik8/dummy/enwik8/1.1.0/dummy_data.zip`

datasets/enwik8/enwik8.py
```diff
@@ -0,0 +1,95 @@
...
+        ]
+
+    # method parameters are unpacked from `gen_kwargs` as given in `_split_generators`
+    def _generate_examples(self, filepath, split):
+        with open(filepath, encoding="utf-8") as f:
+            if self.config.name.endswith("raw"):
+                yield 0, {"text": f.read()}
+            else:
+                lines = f.readlines()
+                for key, line in enumerate(lines):
```
lhoestq, line 91: you can read line by line, which is more efficient when you load the dataset in streaming mode :)
```suggestion
                for key, line in enumerate(f):
```

datasets/enwik8/enwik8.py
```diff
@@ -0,0 +1,95 @@
...
+                    if line.strip():
+                        yield key, {"text": line}
+                    else:
+                        yield key, {"text": ""}
```
lhoestq, line 95: you can just strip the line directly
```suggestion
                    yield key, {"text": line.strip()}
```

lhoestq (approved):
Thank you !

PR #4344, opened 2022-05-13
Fix docstring in DatasetDict::shuffle

felixdivo:
I think  due to #1626, the docstring contained this error ever since `seed` was added.


lhoestq (approved):
good catch ! thanks :)

PR #4347, opened 2022-05-13
Support remote cache_dir

albertvillanova:
This PR implements complete support for remote `cache_dir`. Before, the support was just partial.

This is useful to create datasets using Apache Beam (parallel data processing) builder with `cache_dir` in a remote bucket, e.g., for Wikipedia dataset.

### DIFF
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -287,7 +287,7 @@ class DatasetBuilder:
             self.info.features = features
 
         # prepare data dirs
-        cache_dir_root = str(cache_dir) or config.HF_DATASETS_CACHE
+        cache_dir_root = str(cache_dir or config.HF_DATASETS_CACHE)
         path_join = posixpath.join if is_remote_url(cache_dir_root) else os.path.join
         self._cache_dir_root = (
             os.path.expanduser(cache_dir_root) if not is_remote_url(cache_dir_root) else cache_dir_root
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -286,7 +286,8 @@ class DatasetBuilder:
         if features is not None:
             self.info.features = features
 
-        # prepare data dirs
+        # Prepare data dirs:
+        # cache_dir can be a remote bucket on GCS or S3 (when using BeamBasedBuilder for distributed data processing)
         self._cache_dir_root = str(cache_dir or config.HF_DATASETS_CACHE)
         self._cache_dir_root = (
             self._cache_dir_root if is_remote_url(self._cache_dir_root) else os.path.expanduser(self._cache_dir_root)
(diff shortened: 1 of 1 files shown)


src/datasets/builder.py
```diff
@@ -450,19 +462,26 @@ def _relative_data_dir(self, with_version=True, with_hash=True) -> str:
         builder_data_dir = self.name if namespace is None else f"{namespace}___{self.name}"
         builder_config = self.config
         hash = self.hash
+        path_join = os.path.join if is_local else posixpath.join
```
lhoestq, line 465: same here (and this way you don't need is_local)

src/datasets/builder.py
```diff
@@ -450,19 +462,26 @@ def _relative_data_dir(self, with_version=True, with_hash=True) -> str:
...
-            builder_data_dir = os.path.join(builder_data_dir, hash)
+            builder_data_dir = path_join(builder_data_dir, hash)
         return builder_data_dir
 
     def _build_cache_dir(self):
         """Return the data directory for the current version."""
-        builder_data_dir = os.path.join(self._cache_dir_root, self._relative_data_dir(with_version=False))
-        version_data_dir = os.path.join(self._cache_dir_root, self._relative_data_dir(with_version=True))
+        is_local = not is_remote_url(self._cache_dir_root)
+        path_join = os.path.join if is_local else posixpath.join
```
lhoestq, line 478: same here

lhoestq (commented):
Awesome ! Just one minor comment: you can use `xjoin` directly to support both URLs and local paths

albertvillanova:
@lhoestq thanks fr your review.

Please note that `xjoin` cannot be used in this context, as it always returns a POSIX path and this is not suitable on Windows machines.

lhoestq:
`xjoin` returns windows paths (not posix) on windows, since it just extends`os.path.join`

albertvillanova:
Also note that using `xjoin`, if we pass `cache_dir="C:\\Users\\Username\\.mycache"`, we get:
- `self._cache_dir_root = "C:\\Users\\Username\\.mycache"`
- `self._cache_downloaded_dir = "C:/Users/Username/.mycache/downloads"`

lhoestq:
It looks like it broke the CI on windows :/ maybe this was not a good idea, sorry

src/datasets/builder.py
```diff
@@ -286,24 +287,35 @@ def __init__(
             self.info.features = features
 
         # prepare data dirs
-        self._cache_dir_root = os.path.expanduser(cache_dir or config.HF_DATASETS_CACHE)
+        self._cache_dir_root = str(cache_dir or config.HF_DATASETS_CACHE)
```
lhoestq, line 290: I think we can add a comment here:
```suggestion
        # cache_dir specified by the user can be a remote bucket on GCS or S3
        # This can happen when using Apache Beam with Google Dataflow
        self._cache_dir_root = str(cache_dir or config.HF_DATASETS_CACHE)
```

PR #4350, opened 2022-05-13
Add a new metric: CTC_Consistency

YEdenZ:
Add CTC_Consistency metric
Do I also need to modify the `test_metric_common.py` file to make it run on test?


albertvillanova:
Thanks for your contribution, @YEdenZ.

Please note that our old `metrics` module is in the process of being incorporated to a separate library called `evaluate`: https://github.com/huggingface/evaluate

Therefore, I would ask you to transfer your PR to that repository. Thank you.

PR #4356, opened 2022-05-16
Fix dataset builder default version

albertvillanova:
Currently, when using a custom config (subclass of `BuilderConfig`), default version set at the builder level is ignored: we must set default version in the custom config class.

However, when loading a dataset with `config_kwargs` (for a configuration not present in `BUILDER_CONFIGS`), the default version set in the custom config is ignored and "0.0.0" is used instead:
```python
ds = load_dataset("wikipedia", language="co", date="20220501", beam_runner="DirectRunner")
```
generates the following config:
```python
WikipediaConfig(name='20220501.co', version=0.0.0, data_dir=None, data_files=None, description='Wikipedia dataset for co, parsed from 20220501 dump.')
```
with version "0.0.0" instead of "2.0.0".

See as a counter-example, when the config is present in `BUILDER_CONFIGS`:
```python
ds = load_dataset("wikipedia", "20220301.fr", beam_runner="DirectRunner")
```
generates the following config:
```python
WikipediaConfig(name='20220301.fr', version=2.0.0, data_dir=None, data_files=None, description='Wikipedia dataset for fr, parsed from 20220301 dump.')
```
with correct version "2.0.0", as set in the custom config class.

The reason for this is that `DatasetBuilder` has a default VERSION ("0.0.0") that overwrites the default version set at the custom config class.

This PR:
- Removes the default VERSION at `DatasetBuilder` (set to None, so that the class attribute exists but it does not override the custom config default version).
- Note that the `BuilderConfig` class already sets a default version = "0.0.0"; no need to pass this from the builder.


albertvillanova:
This PR requires one of these other PRs being merged first:
- #4359 
- huggingface/doc-builder#211

src/datasets/commands/dummy_data.py
```diff
@@ -295,14 +295,9 @@ def run(self):
...
-                if builder_config is None:
-                    name = None
-                    version = builder_cls.VERSION
-                else:
-                    version = builder_config.version
-                    name = builder_config.name
-
+                name = builder_config.name if builder_config else None
                 dataset_builder = builder_cls(name=name, hash=dataset_module.hash, cache_dir=tmp_dir)
+                version = builder_config.version if builder_config else dataset_builder.config.version
```
lhoestq, line 300: this is maybe simpler ?
```suggestion
                version = dataset_builder.config.version
```

tests/test_dataset_common.py
```diff
@@ -135,10 +135,7 @@ def check_load_dataset(self, dataset_name, configs, is_local=False, use_local_du
                     logger.info("Skip tests for this dataset for now")
                     return
 
-                if config is not None:
-                    version = config.version
-                else:
-                    version = dataset_builder.VERSION
+                version = config.version if config else dataset_builder.config.version
```
lhoestq, line 138: same
```suggestion
                version = dataset_builder.config.version
```

lhoestq (commented):
Oh good catch ! Maybe it's worth adding a test to make sure a custom config gets the expected version ?

tests/test_dataset_common.py
albertvillanova: @lhoestq I did a refactoring that strictly preserved previous functionality.

Your suggestion changes previous functionality:
- before: we could have configs with different versions and they would be handled properly (even configs with the same name but different versions would be handled properly)
- with your change: we assume all configs have the same version (and we also assume that all configs have the same version as the default config or first BUILDER_CONFIGS

Is this change in behavior intended?

lhoestq: Oh ok I see - please ignore my comment then

tests/test_builder.py
```diff
@@ -756,3 +756,66 @@ def test_custom_writer_batch_size(tmp_path, writer_batch_size, default_writer_ba
...
+        pass
+
+
+@pytest.mark.parametrize(
+    "builder_class, kwargs",
+    [
+        (DummyBuilderWithVersion, {}),
+        (DummyBuilderWithBuilderConfigs, dict(name="custom")),
+        (DummyBuilderWithCustomBuilderConfigs, dict(name="20220501.en")),
+        (DummyBuilderWithCustomBuilderConfigs, dict(date="20220501", language="ca")),
```
lhoestq, line 815: (nit) we usually define dicts using `{}`
```suggestion
        (DummyBuilderWithBuilderConfigs, {"name": "custom"}),
        (DummyBuilderWithCustomBuilderConfigs, {"name": "20220501.en"}),
        (DummyBuilderWithCustomBuilderConfigs, {"date": "20220501", "language": "ca"}),
```

albertvillanova: I had committed this change but forgot to push... I'm pushing it now. Thanks. :)

Issue #4361, opened 2022-05-17
`udhr` doesn't load, dataset checksum mismatch


leondz:
## Describe the bug
Loading `udhr` fails due to a checksum mismatch for some source files. Looks like both of the source files on unicode.org have changed:

size + checksum in datasets repo:
```
(hfdev) leon@blade:~/datasets/datasets/udhr$ jq .default.download_checksums < dataset_infos.json 
{
  "https://unicode.org/udhr/assemblies/udhr_xml.zip": {
    "num_bytes": 2273633,
    "checksum": "0565fa62c2ff155b84123198bcc967edd8c5eb9679eadc01e6fb44a5cf730fee"
  },
  "https://unicode.org/udhr/assemblies/udhr_txt.zip": {
    "num_bytes": 2107471,
    "checksum": "087b474a070dd4096ae3028f9ee0b30dcdcb030cc85a1ca02e143be46327e5e5"
  }
}
```

size + checksum regenerated from current source files:
```
(hfdev) leon@blade:~/datasets/datasets/udhr$ rm dataset_infos.json
(hfdev) leon@blade:~/datasets/datasets/udhr$ datasets-cli test --save_infos udhr.py
Using custom data configuration default
Testing builder 'default' (1/1)
Downloading and preparing dataset udhn/default (download: 4.18 MiB, generated: 6.15 MiB, post-processed: Unknown size, total: 10.33 MiB) to /home/leon/.cache/huggingface/datasets/udhn/default/0.0.0/ad74b91fa2b3c386e5751b0c52bdfda76d334f76731142fd432d4acc2e2fde66...
Dataset udhn downloaded and prepared to /home/leon/.cache/huggingface/datasets/udhn/default/0.0.0/ad74b91fa2b3c386e5751b0c52bdfda76d334f76731142fd432d4acc2e2fde66. Subsequent calls will reuse this data.
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00, 686.69it/s]
Dataset Infos file saved at dataset_infos.json
Test successful.
(hfdev) leon@blade:~/datasets/datasets/udhr$ jq .default.download_checksums < dataset_infos.json 
{
  "https://unicode.org/udhr/assemblies/udhr_xml.zip": {
    "num_bytes": 2389690,
    "checksum": "a3350912790196c6e1b26bfd1c8a50e8575f5cf185922ecd9bd15713d7d21438"
  },
[...]

PR #4362, opened 2022-05-17
Update dataset_infos for UDHN/udhr dataset

leondz:
Checksum update to `udhr` for issue #4361


albertvillanova:
Thanks for contributing @leondz.

The checksums of the files have changed because more languages have been added:
- the new language codes need to be added to the dataset card (README file)
- I think the dataset version number should also be increased, so that users who had previously cached it, get a new dataset download (with the additional languages)

leondz:
Yep! All done

albertvillanova (changes requested):

datasets/udhr/udhr.py
```diff
@@ -46,6 +46,8 @@
 class UDHN(datasets.GeneratorBasedBuilder):
     """Universal Declaration of Human Rights"""
 
+    VERSION = datasets.Version("1.2022.0519")
```
albertvillanova, line 49: Also note that we normally use semantic versioning.

leondz: Do you have a link to an acceptable scheme?

datasets/udhr/README.md
```diff
@@ -8,73 +8,102 @@ languages:
...
 - am
 - amc
 - ame
 - ami
 - amr
-- arb
+- ar
 - arl
 - arn
+- as
```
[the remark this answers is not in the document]
albertvillanova: Normally, we would like to be in sync with the data provided by the data owner.
- If we use their source URL, our checksum will raise an exception each time the data owner updates their data; that way we will know we need to update our version (and add the new languages to the dataset card)
- If we make a dump of current data, we will not notice there are updates unless we check periodically their website

I think it is better to have an exception that tells us to make an update on our side. Also note that users can circumvent this exception if they really want to load the dataset as it is, by passing `ignore_verifications=True`

datasets/udhr/udhr.py
leondz: This is the first version with a version number, so bumped to 1.0.0

datasets/udhr/README.md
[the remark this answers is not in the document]
leondz: OK, cool. Fixed now (stage 1+2 removed)

If anyone looks at this later, there's a repo for the data at https://github.com/unicode-org/udhr which seems to maintain a dated revision history of the source data, and that should allow post-hoc determination of which HF versions corresponded to what translations.

leondz:
I guess the language code CI failure is due to languages.json being a subset of bcp47 (see issue #4304), happy to contribute a solution here, e.g. autogeneration of the lang list from the relevant isos and the ietf bcp47 subtag register or full code for validation

albertvillanova (changes requested):
Thanks again for your contribution.

Yes, I think it is OK to set version 1.0.0 (as previous was 0.0.0).

One of the CI failures is related to dummy data: once you have updated the dataset version, the dummy_data ZIP file should be moved from "dummy/0.0.0/dummy_data.zip" to "dummy/1.0.0/dummy_data.zip".

Other CI failure is related to missing languages in our resources file. This has been addressed in this PR:
- #4371 

You should merge master branch into your feature branch to incorporate that fix.

leondz:
> Thanks again for your contribution, @leondz.
> 
> Yes, I think it is OK to set version 1.0.0 (as previous was 0.0.0).
> 
> One of the CI failures is related to dummy data: once you have updated the dataset version, the dummy_data ZIP file should be moved from "dummy/0.0.0/dummy_data.zip" to "dummy/1.0.0/dummy_data.zip".

Oh, thanks, I missed that one


> Other CI failure is related to missing languages in our resources file. This has been addressed in this PR:
> 
> * #4371
> 
> You should merge master branch into your feature branch to incorporate that fix.

Yeah, I saw this :) I already have the merge, thanks. I'm talking about the longer-term picture: every time another language code comes up (e.g. da-bornholm or es-VE), the json will need updating, because the current approach is non-exhaustive manual whitelisting instead of relying on the established bcp standard.

albertvillanova (changes requested):
Hi @leondz,

Could you please merge the master branch to see if the tests pass?
```
git fetch upstream master
git merge upstream/master
git push
```

albertvillanova:
@leondz you need to merge master branch again to fix the tests.

albertvillanova (approved):
Thank you!

Issue #3951, opened 2022-03-16
Forked streaming datasets try to `open` data urls rather than use network


dlwh:
## Describe the bug
Building on #3950, if you bypass the pickling problem you still can't use the dataset. Somehow something gets confused and the forked processes try to `open` urls rather than anything else.

## Steps to reproduce the bug
```python
from multiprocessing import freeze_support

import transformers
from transformers import Trainer, AutoModelForCausalLM, TrainingArguments
import datasets
import torch.utils.data

# work around #3950
class TorchIterableDataset(datasets.IterableDataset, torch.utils.data.IterableDataset):
    pass

def _ensure_format(v: datasets.IterableDataset) -> datasets.IterableDataset:
    return TorchIterableDataset(v._ex_iterable, v.info, v.split, "torch", v._shuffling)

if __name__ == '__main__':
    freeze_support()

    ds = datasets.load_dataset('oscar', "unshuffled_deduplicated_en", split='train', streaming=True)
    ds = _ensure_format(ds)

    model = AutoModelForCausalLM.from_pretrained("distilgpt2")
    Trainer(model, train_dataset=ds, args=TrainingArguments("out", max_steps=1000, dataloader_num_workers=4)).train()
```

## Expected results

I'd expect the dataset to load the url correctly and produce examples.

## Actual results
```
  warnings.warn(
***** Running training *****
  Num examples = 8000
  Num Epochs = 9223372036854775807
  Instantaneous batch size per device = 8
  Total train batch size (w. parallel, distributed & accumulation) = 8
  Gradient Accumulation steps = 1
  Total optimization steps = 1000
  0%|          | 0/1000 [00:00<?, ?it/s]Traceback (most recent call last):
  File "/Users/dlwh/src/mistral/src/stream_fork_crash.py", line 22, in <module>
    Trainer(model, train_dataset=ds, args=TrainingArguments("out", max_steps=1000, dataloader_num_workers=4)).train()
  File "/Users/dlwh/.conda/envs/mistral/lib/python3.8/site-packages/transformers/trainer.py", line 1339, in train
    for step, inputs in enumerate(epoch_iterator):
  File "/Users/dlwh/.conda/envs/mistral/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 521, in __next__
    data = self._next_data()
  File "/Users/dlwh/.conda/envs/mistral/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 1203, in _next_data
[...]

Issue #3950, opened 2022-03-16
Streaming Datasets don't work with Transformers Trainer when dataloader_num_workers>1


dlwh:
## Describe the bug
Streaming Datasets can't be pickled, so any interaction between them and multiprocessing results in a crash.

## Steps to reproduce the bug
```python
import transformers
from transformers import Trainer, AutoModelForCausalLM, TrainingArguments
import datasets

ds = datasets.load_dataset('oscar', "unshuffled_deduplicated_en", split='train', streaming=True).with_format("torch")
model = AutoModelForCausalLM.from_pretrained("distilgpt2")
Trainer(model, train_dataset=ds, args=TrainingArguments("out", max_steps=1000, dataloader_num_workers=4)).train()
```
## Expected results
For this code I'd expect a crash related to not having preprocessed the data, but instead we get a pickling error.

## Actual results
```
  0%|          | 0/1000 [00:00<?, ?it/s]Traceback (most recent call last):
  File "/Users/dlwh/src/mistral/src/stream_fork_crash.py", line 7, in <module>
    Trainer(model, train_dataset=ds, args=TrainingArguments("out", max_steps=1000, dataloader_num_workers=4)).train()
  File "/Users/dlwh/.conda/envs/mistral/lib/python3.8/site-packages/transformers/trainer.py", line 1339, in train
    for step, inputs in enumerate(epoch_iterator):
  File "/Users/dlwh/.conda/envs/mistral/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 359, in __iter__
    return self._get_iterator()
  File "/Users/dlwh/.conda/envs/mistral/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 305, in _get_iterator
    return _MultiProcessingDataLoaderIter(self)
  File "/Users/dlwh/.conda/envs/mistral/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 918, in __init__
    w.start()
  File "/Users/dlwh/.conda/envs/mistral/lib/python3.8/multiprocessing/process.py", line 121, in start
    self._popen = self._Popen(self)
  File "/Users/dlwh/.conda/envs/mistral/lib/python3.8/multiprocessing/context.py", line 224, in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
  File "/Users/dlwh/.conda/envs/mistral/lib/python3.8/multiprocessing/context.py", line 284, in _Popen
    return Popen(process_obj)
[...]

PR #4375, opened 2022-05-19
Support DataLoader with num_workers > 0 in streaming mode

lhoestq:
### Issue

It's currently not possible to properly stream a dataset using multiple `torch.utils.data.DataLoader` workers:

- the `TorchIterableDataset` can't be pickled and passed to the subprocesses: https://github.com/huggingface/datasets/issues/3950
- streaming extension is failing: https://github.com/huggingface/datasets/issues/3951
- `fsspec` doesn't work out of the box in subprocesses

### Solution in this PR

I fixed these to enable passing an `IterableDataset` to a  `torch.utils.data.DataLoader` with `num_workers > 0`.

I also had to shard the `IterableDataset` to give each worker a shard, otherwise data would be duplicated. This is implemented in `TorchIterableDataset.__iter__` and uses the new `IterableDataset._iter_shard(shard_idx)` method

I also had to do a few changes the patching that enable streaming in dataset scripts:
- the patches are now always applied - not just for streaming mode. They're applied when a builder is instantiated
- I improved it to also check for renamed modules or attributes (ex: pandas vs pd)
- I grouped all the patches of pathlib.Path into a class `xPath`, so that `Path` outside of dataset scripts stay unchanged - otherwise I didn't change the content of the extended Path methods for streaming
- I fixed a bug with the `pd.read_csv` patch, opening the file in "rb" mode was missing and causing some datasets to not work in streaming mode, and compression inference was missing

### A few details regarding `fsspec` in multiprocessing

From https://github.com/fsspec/filesystem_spec/pull/963#issuecomment-1131709948 :
> Non-async instances might be safe in the forked child, if they hold no open files/sockets etc.; I'm not sure any implementations pass this test!
> If any async instance has been created, the newly forked processes must:
> 1. discard references to locks, threads and event loops and make new ones
> 2. not use any async fsspec instances from the parent process
> 3. clear all class instance caches

Therefore in a DataLoader's worker, I clear the reference to the loop and thread (1). We should be fine for 2 and 3 already since we don't use fsspec class instances from the parent process.

Fix https://github.com/huggingface/datasets/issues/3950
Fix https://github.com/huggingface/datasets/issues/3951

TODO:
- [x] fix tests


lhoestq:
Alright this is finally ready for review ! It's quite long I'm sorry, but it's not easy to disentangle everything ^^'

The main additions are in
- src/datasets/formatting/dataset_wrappers/torch_iterable_dataset.py
- src/datasets/iterable_dataset.py
- src/datasets/utils/patching.py

src/datasets/download/streaming_download_manager.py
```diff
@@ -529,126 +505,153 @@ def xglob(urlpath, *, recursive=False, use_auth_token: Optional[Union[str, bool]
...
-        for globbed_path in globbed_paths:
-            yield type(path)("::".join([f"{fs.protocol}://{globbed_path}"] + rest_hops))
+        fs, *_ = fsspec.get_fs_token_paths(urlpath, storage_options=storage_options)
+        for dirpath, dirnames, filenames in fs.walk(main_hop):
+            yield "::".join([f"{fs.protocol}://{dirpath}"] + rest_hops), dirnames, filenames
 
 
-def xpathrglob(path, pattern, **kwargs):
-    """Rglob function for argument of type :obj:`~pathlib.Path` that supports both local paths end remote URLs.
+class xPath(type(Path())):
```
lhoestq, line 536: many changes in this file are just about moving functions inside this class.

For example I moved xpathrglob to xPath.rglob

src/datasets/formatting/dataset_wrappers/torch_iterable_dataset.py
```diff
@@ -0,0 +1,56 @@
...
+                )
+            # split workload
+            shards_indices = list(range(worker_info.id, self.n_shards, worker_info.num_workers))
+            if shards_indices:
+                logger.debug(
+                    f"dataloader worker#{worker_info.id}, ': Starting to iterate over {len(shards_indices)}/{self.n_shards} shards."
+                )
+                for shard_idx in shards_indices:
+                    for key, example in self._iter_shard(shard_idx):
+                        yield self._apply_feature_types(example)
```
lhoestq, line 49: this is where we shard the iterable dataset when it's passed to a DataLoader worker

src/datasets/iterable_dataset.py
```diff
@@ -499,19 +565,26 @@ def _iter(self):
             ex_iterable = self._ex_iterable
         yield from ex_iterable
 
+    def _iter_shard(self, shard_idx: int):
+        if self._shuffling:
+            ex_iterable = self._ex_iterable.shuffle_data_sources(self._effective_generator())
+        else:
+            ex_iterable = self._ex_iterable
+        yield from ex_iterable.shard_data_sources(shard_idx)
```
lhoestq, line 573: This is what is called when iterating in a DataLoader worker. The idea is to iterate only on one shard out of the `self.n_shards` available

src/datasets/streaming.py
```diff
@@ -85,20 +84,29 @@ def wrapper(*args, **kwargs):
...
-        patch.object(module.Path, "name", property(fget=xpathname)).start()
-        patch.object(module.Path, "stem", property(fget=xpathstem)).start()
-        patch.object(module.Path, "suffix", property(fget=xpathsuffix)).start()
-    patch_submodule(module, "pd.read_csv", wrap_auth(xpandas_read_csv), attrs=["__version__"]).start()
-    patch_submodule(module, "pd.read_excel", xpandas_read_excel, attrs=["__version__"]).start()
-    patch_submodule(module, "sio.loadmat", wrap_auth(xsio_loadmat), attrs=["__version__"]).start()
-    # xml.etree.ElementTree
-    for submodule in ["ElementTree", "ET"]:
-        patch_submodule(module, f"{submodule}.parse", wrap_auth(xet_parse)).start()
+    patch_submodule(module, "pathlib.Path", xPath).start()
```
lhoestq, line 87: Now we can just pass the source object to patch in the module, and it will patch it even the attribute or a parent module has been renamed (see test_patching.py for a list of all supported cases - I probably have to add a docstring to `patch_submodule` as well)

src/datasets/iterable_dataset.py
```diff
@@ -78,6 +83,11 @@ def _shuffle_kwargs(rng: np.random.Generator, kwargs: dict) -> dict:
     return shuffled_kwargs
 
 
+def _shard_kwargs(shard_idx: int, kwargs: dict) -> dict:
+    """Return a copy of the input kwargs but with only one shard"""
+    return {key: [value[shard_idx]] if isinstance(value, list) else value for key, value in kwargs.items()}
```
mariosasko, line 88: If I'm not mistaken, an index error can happen here in the case of different-sized lists in `gen_kwargs` (`n_shards` is equal to the max length of these lists), so we have to check if all the list sizes match, no?

src/datasets/utils/patching.py
```diff
@@ -39,19 +42,50 @@ def __init__(self, obj, target: str, new, attrs=None):
...
+        # Patch modules:
+        # it's used to patch attributes of submodules like "os.path.join";
+        # in this case we need to patch "os" and "os.path"
+        for i in range(len(submodules)):
+            submodule = import_module(".".join(submodules[: i + 1]))
+            for attr in self.obj.__dir__():
+                obj_attr = getattr(self.obj, attr)
+                if obj_attr is submodule or (
+                    (isinstance(obj_attr, _PatchedModuleObj) and obj_attr._original_module is submodule)
+                ):
```
mariosasko, line 59: A comment here would also be nice to explain that this allows us to patch renamed modules.

mariosasko (commented):
This is 🔥. Just two comments: 

(Also thanks for the the review instructions/code clarifications)

src/datasets/iterable_dataset.py
lhoestq: Maybe we should simply disallow users to use several lists of different sizes completely, to make things simple ? And make it clear in the docs that LISTS == SHARDS ?

I couldn't find a single dataset with lists of different sizes.

For now I'll just raise an error if lists are of different sizes in `_shard_kwargs`

lhoestq:
Added some comments and an error when lists have different lengths for sharding :)

src/datasets/iterable_dataset.py
mariosasko: Yes, that sounds good!

mariosasko:
Let's resolve the merge conflict and the CI error (if it's related to the changes), and I can review the PR again.

lhoestq:
Feel free to review again :) The CI fail is unrelated to this PR and will be fixed soon (the hub now returns 401 instead of 404 for unauthenticated requests to non-existing repos)

mariosasko (approved):
All looks good now! Thanks!

lhoestq:
CI failures are unrelated to this PR - merging :)

(CI fails are a mix of pip install fails and Hub fails)

justheuristic:
@lhoestq you're our hero :)

Issue #4379, opened 2022-05-20
Latest dill release raises exception


albertvillanova:
## Describe the bug
As reported by @sgugger, latest dill release is breaking things with Datasets.

```
______________ ExamplesTests.test_run_speech_recognition_seq2seq _______________


self = <multiprocess.pool.ApplyResult object at 0x7fa5981a1cd0>, timeout = None

    def get(self, timeout=None):
        self.wait(timeout)
        if not self.ready():
            raise TimeoutError
        if self._success:
            return self._value
        else:
>           raise self._value
E           TypeError: '>' not supported between instances of 'NoneType' and 'float'
```

albertvillanova:
Fixed by:
- #4380

gugarosa:
Just an additional insight, the latest dill (either 0.3.5 or 0.3.5.1) also broke the hashing/fingerprinting of any mapping function.

For example:
```
from datasets import load_dataset

d = load_dataset("rotten_tomatoes")
d.map(lambda x: x)
```

Returns the standard non-dillable error:
```
Parameter 'function'=<function <lambda> at 0x7fe7d18c9560> of the transform datasets.arrow_dataset.Dataset._map_single couldn't be hashed properly....
```

anivegesana:
@albertvillanova ExamplesTests.test_run_speech_recognition_seq2seq is in which file?

albertvillanova:
Thanks a lot @gugarosa for the insight: we will incorporate it in our CI as regression testing for future dill releases.

albertvillanova:
Hi @anivegesana, that test is in `transformers` library:
- https://github.com/huggingface/transformers/blob/main/examples/pytorch/test_pytorch_examples.py#L449
- https://github.com/huggingface/transformers/blob/main/examples/pytorch/speech-recognition/run_speech_recognition_seq2seq.py

anivegesana:
I did a deep dive into @gugarosa's problem and found the issue and it might be related to the one @sgugger discovered. In dill 0.3.5(.1), I created a new `save_function` that fixes a bug in dill that prevented the pickling of recursive inner functions. It was a more complete solution to the problem that `dill._dill.stack` tried to solve in the internal API of dill, so I removed it. Since datasets copies the `save_function` directly from the dill API, it stops working with the new dill version since `dill._dill.stack` is no longer present.

[...]

PR #4385, opened 2022-05-21
Test dill

albertvillanova:
Regression test for future releases of `dill`.

Related to #4379.


anivegesana:
I should point out that the hash will be the same if computed twice with the same code on the same version of dill (after adding huggingface's code that removes line numbers and file names, and sorts globals.) My changes in dill 0.3.5 and ones that I will make in 0.3.6 will result in different pickles than the ones dill 0.3.4 was making. This should still be fine for caching.

tests/test_fingerprint.py
```diff
@@ -0,0 +1,7 @@
+from datasets.fingerprint import Hasher
+
+
+def test_dill():
+    # AttributeError: module 'dill._dill' has no attribute 'stack'
+    hasher = Hasher()
+    hasher.update(lambda x: x)
```
lhoestq, line 7: I think you can just check this
```suggestion
    Hasher.hash(lambda x: x)
```

lhoestq (commented):
Thanks !
I think you can have this test in `test_caching.py` with the other tests on hashing :)

albertvillanova:
Just some comments @lhoestq:

The best practice for testing is to have a `test_<filename>.py` for each `<filename>.py`. Therefore in order to have the filenames aligned, I would propose:
- either renaming `fingerprint.py` to `caching.py`
- or renaming `test_caching.py` to `test_fingerprint.py`

On the other hand, my idea when implementing this test was not to test all the functionalities of the `Hasher`, but just to have a regression test that fails if dill version is > 0.3.4 and the pin in our `setup.py` is not present. Just recall that we had no failing test in our CI when the issue with dill was found on `transformers`.

The objective of this PR is just to have a regression test for that case: I tested and I got `AttributeError: module 'dill._dill' has no attribute 'stack'`

There was no equivalent test in `test_caching.py` because the our CI did not fail before pinning dill.

PR #4390, opened 2022-05-23
Fix metadata validation

albertvillanova:
Since Python 3.8, the typing module:
- raises an AttributeError when trying to access `__args__` on any type, e.g.: `List.__args__`
- provides the `get_args` function instead: `get_args(List)`

This PR implements this fix for Python >=3.8 whereas maintaining backward compatibility.


lhoestq (approved):
good catch thanks !

PR #4391, opened 2022-05-23
Refactor column mappings for question answering datasets

lewtun:
This PR tweaks the keys in the metadata that are used to define the column mapping for question answering datasets. This is needed in order to faithfully reconstruct column names like `answers.text` and `answers.answer_start` from the keys in AutoTrain.

As observed in https://github.com/huggingface/datasets/pull/4367 we cannot use periods `.` in the keys of the YAML tags, so a decision was made to use a flat mapping with underscores. For QA datasets, however, it's handy to be able to reconstruct the nesting -- hence this PR.

cc @sashavor

### DIFF
--- a/datasets/cuad/README.md
+++ b/datasets/cuad/README.md
@@ -30,8 +30,9 @@ train-eval-index:
   col_mapping:
     question: question
     context: context
-    answers__text: answers.text
-    answers__answer_start: answers.answer_start
+    answers:
+      text: text
+      answer_start: answer_start
   metrics:
     - type: cuad
       name: CUAD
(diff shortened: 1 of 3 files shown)


albertvillanova (approved):
Thanks.

I have no visibility about this, but if you say it is more useful for AutoTrain this way...

lewtun:
> Thanks.
> 
> I have no visibility about this, but if you say it is more useful for AutoTrain this way...

Thanks for the review @albertvillanova ! Yes, I need some way to reconstruct the original column names with a period because that's how they appear after we flatten the nested columns. In any case, we can adjust this later if needed :)

lewtun:
> Does that mean that we need to change the metadata?

Yes, but this PR takes care of it :)

PR #4392, opened 2022-05-23
remove int documentation from logging docs

lvwerra:
Removes the `int` documentation from the [logging section](https://huggingface.co/docs/datasets/package_reference/logging_methods#levels) of the docs.


albertvillanova (approved):
Thanks for spotting and fixing this!

PR #4395, opened 2022-05-23
Add Pascal VOC dataset

nateraw:
This PR adds the Pascal VOC dataset in the same way TFDS has it added. I believe we can iterate on this dataset and in future versions include more data, such as segmentation masks, but for now I think it is a good idea to just add it the same way as TFDS to get a solid first version out there.


datasets/voc/voc.py
```diff
@@ -0,0 +1,229 @@
...
+            description=_VOC_CONFIG_DESCRIPTION.format(year=2012, num_images=11540, num_objects=27450),
+            filenames={
+                "trainval": "VOCtrainval_11-May-2012.tar",
+                "test": "VOC2012test.tar",
+            },
+            has_test_annotations=False,
+        ),
+    ]
+
+    # DEFAULT_CONFIG_NAME = "2012"
```
nateraw, line 156: delete this

datasets/voc/voc.py
```diff
@@ -0,0 +1,229 @@
+# coding=utf-8
+# Copyright 2022 The TensorFlow Datasets Authors and the HuggingFace Datasets Authors
```
nateraw, line 2: make sure this copyright notice is correct

datasets/voc/voc.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+def _get_example_objects(annon_filepath):
+    """Function to get all the objects from the annotation XML file."""
+    with open(annon_filepath, "r") as f:
+        root = xml.etree.ElementTree.parse(f).getroot()
+
+        # Disable pytype to avoid attribute-error due to find returning
+        # Optional[Element]
+        # pytype: disable=attribute-error
```
nateraw, line 106: unnecessary, taken from TFDS. Remove.

nateraw (commented):
few comments for myself to clean up.

datasets/voc/README.md
```diff
@@ -0,0 +1,237 @@
...
+- crowdsourced
+language_creators:
+- crowdsourced
+languages:
+- en
+licenses:
+- other-inherited-from-flickr
+multilinguality:
+- monolingual
+pretty_name: voc
```
lhoestq, line 12:
```suggestion
pretty_name: PASCAL Visual Object Classes (VOC)
```

datasets/voc/README.md
```diff
@@ -0,0 +1,237 @@
...
+- 10K<n<100K
+source_datasets:
+- original
+task_categories:
+- object-detection
+task_ids:
+- other-object-detection
+---
+
+# Dataset Card for [Dataset Name]
```
lhoestq, line 23:
```suggestion
# Dataset Card for PASCAL Visual Object Classes (VOC)
```

datasets/voc/voc.py
```diff
@@ -0,0 +1,229 @@
...
+            citation=_VOC_CITATION.format(year=self.config.year),
+        )
+
+    def _split_generators(self, dl_manager):
+        paths = dl_manager.download_and_extract(
+            {k: os.path.join(_VOC_DATA_URL, v) for k, v in self.config.filenames.items()}
+        )
+        return [
+            datasets.SplitGenerator(
+                name=datasets.Split.TEST, gen_kwargs=dict(data_path=paths["test"], set_name="test")
```
lhoestq, line 187: (nit) we usually define dicts this way
```suggestion
                name=datasets.Split.TEST, gen_kwargs={"data_path": paths["test"], "set_name": "test"}
```

datasets/voc/voc.py
```diff
@@ -0,0 +1,229 @@
...
+            ),
+            datasets.SplitGenerator(
+                name=datasets.Split.TRAIN, gen_kwargs=dict(data_path=paths["trainval"], set_name="train")
```
lhoestq, line 190:
```suggestion
                name=datasets.Split.TRAIN, gen_kwargs={"data_path": paths["trainval"], "set_name": "train"}
```

datasets/voc/voc.py
```diff
@@ -0,0 +1,229 @@
...
+            ),
+            datasets.SplitGenerator(
+                name=datasets.Split.VALIDATION, gen_kwargs=dict(data_path=paths["trainval"], set_name="val")
```
lhoestq, line 193:
```suggestion
                name=datasets.Split.VALIDATION, gen_kwargs={"data_path": paths["trainval"], "set_name": "val"}
```

datasets/voc/voc.py
```diff
@@ -0,0 +1,229 @@
...
+                    "labels_no_difficult": datasets.Sequence(datasets.ClassLabel(names=_VOC_LABELS)),
+                }
+            ),
+            homepage=_VOC_URL.format(year=self.config.year),
+            citation=_VOC_CITATION.format(year=self.config.year),
+        )
+
+    def _split_generators(self, dl_manager):
+        paths = dl_manager.download_and_extract(
+            {k: os.path.join(_VOC_DATA_URL, v) for k, v in self.config.filenames.items()}
```
lhoestq, line 183: os.path.join would use a backslash on windows, which doesn't work for URLs ;)

You can just concatenate the strings instead
```suggestion
            {k: _VOC_DATA_URL + v  for k, v in self.config.filenames.items()}
```

datasets/voc/voc.py
```diff
@@ -0,0 +1,229 @@
...
+            ),
+            datasets.SplitGenerator(
+                name=datasets.Split.VALIDATION, gen_kwargs=dict(data_path=paths["trainval"], set_name="val")
+            ),
+        ]
+
+    def _generate_examples(self, data_path, set_name):
+        set_filepath = os.path.join(
+            data_path, os.path.normpath("VOCdevkit/VOC{}/ImageSets/Main/{}.txt".format(self.config.year, set_name))
+        )
```
lhoestq, line 200: `normpath` is not necessary here

```suggestion
        set_filepath = os.path.join(
            data_path, "VOCdevkit", "VOC" + self.config.year, "ImageSets", "Main", set_name + ".txt")
        )
```

datasets/voc/voc.py
```diff
@@ -0,0 +1,229 @@
...
+        load_annotations = self.config.has_test_annotations or set_name != "test"
+        with open(set_filepath, "r", encoding="utf-8") as f:
+            for line in f:
+                image_id = line.strip()
+                example = self._generate_example(data_path, image_id, load_annotations)
+                yield image_id, example
+
+    def _generate_example(self, data_path, image_id, load_annotations):
+        image_filepath = os.path.join(
+            data_path, os.path.normpath("VOCdevkit/VOC{}/JPEGImages/{}.jpg".format(self.config.year, image_id))
```
lhoestq, line 210: same
```suggestion
            data_path, "VOCdevkit", "VOC" + self.config.year, "JPEGImages", image_id + ".jpg")
```

datasets/voc/voc.py
```diff
@@ -0,0 +1,229 @@
...
+        )
+        annon_filepath = os.path.join(
+            data_path, os.path.normpath("VOCdevkit/VOC{}/Annotations/{}.xml".format(self.config.year, image_id))
```
lhoestq, line 213: same
```suggestion
            data_path, "VOCdevkit", "VOC" + self.config.year, "Annotations", image_id + ".xml")
```

datasets/voc/voc.py
```diff
@@ -0,0 +1,229 @@
...
+                    "labels": datasets.Sequence(datasets.ClassLabel(names=_VOC_LABELS)),
+                    "labels_no_difficult": datasets.Sequence(datasets.ClassLabel(names=_VOC_LABELS)),
+                }
+            ),
+            homepage=_VOC_URL.format(year=self.config.year),
+            citation=_VOC_CITATION.format(year=self.config.year),
+        )
+
+    def _split_generators(self, dl_manager):
+        paths = dl_manager.download_and_extract(
```
lhoestq, line 182: To be able to make this dataset work in streaming mode, you have to first `dl_manager.download` the TAR archive, and then iterate over the files using `dl_manager.iter_archive`.

Unfortunately it's not possible to `dl_manager.download_and_extract` TAR archives in streaming mode.

Streaming mode is also useful for the dataset preview on the Hub.

If you need certain files to be hosted on the Hub let me know and I can help you with it :)

lhoestq (commented):
Cool thanks ! Added a few comments

datasets/voc/README.md
nateraw: I ALWAYS miss this one 😅

datasets/voc/voc.py
lhoestq: Can the txt files fit in memory ? If so you can just load them all in a python dictionary for the lookup of your first proposition. I think 

(otherwise regarding option 2: the images come from Flickr (see terms of use [here](https://www.flickr.com/help/terms)), flickr users retain their copyright on the content of flickr unless they specify an open license like Creative Commons - but this is not the general case. As for ImageNet I think we could redistribute it only if we make this dataset a gated dataset)

mariosasko: I also think we should only go with the 2nd option + gated access if the txt files don't fit in memory.

datasets/voc/README.md
```diff
@@ -0,0 +1,237 @@
...
+    - `pose`: An `int` label related to an object's pose. The different poses are `frontal', `rear', `left' or `right'. The views are subjectively marked to indicate the view of the bulk of the object. Some objects have no view specified.
+    - `is_truncated`: A `bool` value. When True, it indicates that the bounding box specified for the object does not correspond to the full extent of the object (e.g. an image of a person from the waist up, or a view of a car extending outside the image.)
+    - `is_difficult`: A `bool` value. When True, it indicates that the object is considered difficult to recognize, for example an object which is clearly visible but unidentifiable without substantial use of context. Objects marked as difficult are currently ignored in the evaluation of the challenge (See [Supported Tasks And Leaderboards](#supported-tasks-and-leaderboards)).
+
+- `labels`: A list of `int` object class labels contained within the `objects` key.
+
+- `labels_no_difficult`: A list of `int` object class labels contained within the `objects` key where `is_difficult` is True.
```
mariosasko, line 145:
```suggestion
- `labels_no_difficult`: A list of `int` object class labels contained within the `objects` key where `is_difficult` is False.
```

datasets/voc/README.md
```diff
@@ -0,0 +1,237 @@
...
+#### Who are the source language producers?
+
+[More Information Needed]
+
+### Annotations
+
+#### Annotation process
+
+[More Information Needed]
+
```
mariosasko, line 177: This info can be found in the paper

mariosasko (commented):
Looks pretty good!

IMO it makes sense to split this dataset into two datasets, `voc2007` and `voc2012`, to have one dataset for each year. This will help us avoid maintaining one large script, which could be problematic especially when we add the other tasks/years, and allow us to reference the dataset [entries](https://paperswithcode.com/datasets?q=PASCAL+VOC&v=lst&o=match) by year on Papers With Code.

datasets/voc/voc.py
nateraw: The txt files would fit in memory, but it still doesn't solve the problem that the files I need to reference are in multiple directories, so iterating over the whole archive doesn't make sense I think. You can see here that I'd use the names in `train.txt` or `val.txt` to reference files in the `Annotations` and `JPEGImages` directories.

Screen Shot 2022-06-13 at 1 54 45 PM

lhoestq:
Some CI fails are unrelated to your PR and fixed on master, feel free to merge master into your branch :)

albertvillanova:
Thanks @nateraw for the addition of this dataset.

I would suggest to transfer it to the Hugging Face Hub, under a "pascal" organization namespace: "pascal/voc".

What do you think?

lhoestq:
FYI I think this dataset is also available at (internal) https://huggingface.co/datasets/HuggingFaceM4/pascal_voc

nateraw:
@lhoestq @albertvillanova what do you think best path forward is? No idea when I'll get to looking at this again, but would be nice to know plan so when I find time I can just get it done in one sitting.

albertvillanova:
My (not strong) opinion on this:
- as we are removing dataset scripts from GitHub, this dataset should be created directly on the Hub
- I proposed doing it under some kind of "official" org namespace, like pascal or pascal2; other suggestions are welcome
- the link given by @lhoestq might serve as inspiration for your implementation (I think yours misses action classification): their implementation comprises tasks: classification/detection, segmentation, action classification, person layout; it misses other tasks though

What do you think?

PR #4402, opened 2022-05-25
Skip identical files in `push_to_hub` instead of overwriting

mariosasko:
Skip identical files instead of overwriting them to save bandwidth and circumvent (user-side/server-side) errors, which can arise when working with large datasets due to long-lasting HTTP connections, by repeating calls to `push_to_hub` to resume an upload.

To be able to check if an upload can be resumed, this PR modifies the shard naming scheme from:
```
data/{split}-[0-9][0-9][0-9][0-9][0-9]-of-[0-9][0-9][0-9][0-9][0-9].parquet
```
to:
```
data/{split}-[0-9][0-9][0-9][0-9][0-9]-of-[0-9][0-9][0-9][0-9][0-9]-<SHARD_FINGERPRINT>.parquet
```
cc @LysandreJik

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -3949,60 +3949,63 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
             shards = shards_with_embedded_external_files(shards)
 
         files = api.list_repo_files(repo_id, repo_type="dataset", revision=branch, token=token)
-        files = [file for file in files if file.startswith("data/")]
+        data_files = [file for file in files if file.startswith("data/")]
 
-        def path_in_repo(_index):
-            return f"data/{split}-{_index:05d}-of-{num_shards:05d}.parquet"
-
-        # Only delete file shards that don't currently exist. Others will be overwritten if the content is different
-        # or will be left intact is the content is identical.
-        def should_delete_file(file_name):
-            file_to_overwrite = file_name in [path_in_repo(i) for i in range(num_shards)]
-            file_from_same_split = file_name.startswith(f"data/{split}-")
-
-            return file_from_same_split and not file_to_overwrite
-
-        file_shards_to_delete = [file for file in files if should_delete_file(file)]
-
-        def delete_file(file):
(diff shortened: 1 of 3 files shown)


tests/test_upstream_hub.py
```diff
@@ -77,7 +85,9 @@ def test_push_dataset_dict_to_hub_no_token(self):
 
             # Ensure that there is a single file on the repository that has the correct name
             files = sorted(self._api.list_repo_files(ds_name, repo_type="dataset"))
-            self.assertListEqual(files, [".gitattributes", "data/train-00000-of-00001.parquet", "dataset_infos.json"])
+            self.assertFilesMatch(
+                files, [".gitattributes", "data/train-00000-of-00001-<FINGERPRINT>.parquet", "dataset_infos.json"]
+            )
```
lhoestq, line 90: (nit)
It's maybe simpler to match it using a wildcard `*` (maybe with `fnmatch`) instead of implementing a custom logic that reads until the last dash ? And this way you don't have to have a different behavior for files inside `data/` and the others.
```suggestion
            self.assertFilesMatch(
                files, [".gitattributes", "data/train-00000-of-00001-*.parquet", "dataset_infos.json"]
            )
```

src/datasets/arrow_dataset.py
```diff
@@ -3949,60 +3949,63 @@ def shards_with_embedded_external_files(shards):
             shards = shards_with_embedded_external_files(shards)
 
         files = api.list_repo_files(repo_id, repo_type="dataset", revision=branch, token=token)
-        files = [file for file in files if file.startswith("data/")]
+        data_files = [file for file in files if file.startswith("data/")]
 
-        def path_in_repo(_index):
-            return f"data/{split}-{_index:05d}-of-{num_shards:05d}.parquet"
+        def path_in_repo(_index, shard):
+            return f"data/{split}-{_index:05d}-of-{num_shards:05d}-{shard._fingerprint}.parquet"
```
lhoestq, line 3955: Maybe you can check if the first shard is already on the Hub, and show a warning that says that it's resuming the push_to_hub operation ?

lhoestq (approved):
Love it thanks !

Issue #4323, opened 2022-05-12
Audio can not find value["bytes"]


YooSungHyun:
## Describe the bug
I wrote down _generate_examples like:
![image](https://user-images.githubusercontent.com/34292279/168027186-2fe8b255-2cd8-4b9b-ab1e-8d5a7182979b.png)

but where is the bytes?
![image](https://user-images.githubusercontent.com/34292279/168027330-f2496dd0-1d99-464c-b15c-bc57eee0415a.png)

## Expected results
value["bytes"] is not None, so i can make datasets with bytes, not path

## bytes looks like:
blah blah~~
[...]

PR #4409, opened 2022-05-26
Update: add using pcm bytes (#4323)

YooSungHyun:
first of all, please look #4323

why i can not use {"path","array","sampling_rate"}
because sf.write(format="wav") and sf.read(BytesIO) is change my pcm data value
maybe, i think wav got header but, pcm is not.
and variable naming, pcm data is "byte" type. so, "array" name is not fair i think

so, i use scipy lib and numpy (that is huggingface dependency)
and refer to @lhoestq answered,
1. encode -> using sampling_rate and pcm byte -> wav style byte (scipy.wavfile.write to byte)
2. byte converting using fairseq style pcm audio read  [FileAudioDataset](https://github.com/facebookresearch/fairseq/blob/main/fairseq/data/audio/raw_audio_dataset.py)
4. decode -> read wavfile.read

that way is not screw up my pcm byte to float data, and another audio type(wav) safety

please check!

### DIFF
--- a/src/datasets/features/audio.py
+++ b/src/datasets/features/audio.py
@@ -3,8 +3,10 @@ from dataclasses import dataclass, field
 from io import BytesIO
 from typing import TYPE_CHECKING, Any, ClassVar, Dict, Optional, Union
 
+import numpy as np
 import pyarrow as pa
 from packaging import version
+from scipy.io import wavfile
 
 from .. import config
 from ..download.streaming_download_manager import xopen
@@ -43,7 +45,7 @@ class Audio:
             channels.
         decode (:obj:`bool`, default ``True``): Whether to decode the audio data. If `False`,
             returns the underlying dictionary in the format {"path": audio_path, "bytes": audio_bytes}.
-
+    
     Example:
 
     ```py
@@ -92,7 +94,23 @@ class Audio:
             return {"bytes": buffer.getvalue(), "path": None}
         elif value.get("path") is not None and os.path.isfile(value["path"]):
             # we set "bytes": None to not duplicate the data if they're already available locally
-            return {"bytes": None, "path": value.get("path")}
+            if value.get("path").endswith("pcm"):
+                # "PCM" is only have byte RAW-Audio data style.
+                if value.get("sampling_rate") is None:
+                    # At least, If you want to convert "PCM-byte" to "WAV-byte", you have to know sampling rate
+                    raise KeyError("If you want to use pcm, you have to know your 'sampling_rate'!")
+                if value.get("bytes"):
+                    # If we already had PCM-byte, we don`t have to make "read file, make bytes" (just use it!)
+                    # I used fairseq's using PCM data (FileAudioDataset)
+                    bytes_value = np.frombuffer(value["bytes"], dtype=np.int16).astype(np.float32) / 32767
+                else:
+                    bytes_value = np.memmap(value["path"], dtype="h", mode="r").astype(np.float32) / 32767
+
+                buffer = BytesIO(bytes())
+                wavfile.write(buffer, value["sampling_rate"], bytes_value)
+                return {"bytes": buffer.getvalue(), "path": value.get("path")}
+            else:
+                return {"bytes": None, "path": value.get("path")}
         elif value.get("bytes") is not None or value.get("path") is not None:
             # store the audio bytes, and path is used to infer the audio format using the file extension
             return {"bytes": value.get("bytes"), "path": value.get("path")}
@@ -131,6 +149,8 @@ class Audio:
                 array, sampling_rate = self._decode_non_mp3_path_like(
                     path, "opus", token_per_repo_id=token_per_repo_id
                 )
+        elif path is not None and path.endswith("pcm"):
+            sampling_rate, array = wavfile.read(file)
         else:
             if file:
                 array, sampling_rate = self._decode_non_mp3_file_like(file)


YooSungHyun:
wefwef

src/datasets/features/audio.py
```diff
@@ -92,7 +94,23 @@ def encode_example(self, value: Union[str, dict]) -> dict:
             return {"bytes": buffer.getvalue(), "path": None}
         elif value.get("path") is not None and os.path.isfile(value["path"]):
             # we set "bytes": None to not duplicate the data if they're already available locally
-            return {"bytes": None, "path": value.get("path")}
+            if value.get("path").endswith("pcm"):
+                # "PCM" is only have byte RAW-Audio data style.
```
lhoestq, line 98:
```suggestion
                # "PCM" only has raw audio bytes
```

src/datasets/features/audio.py
```diff
@@ -92,7 +94,23 @@ def encode_example(self, value: Union[str, dict]) -> dict:
             return {"bytes": buffer.getvalue(), "path": None}
         elif value.get("path") is not None and os.path.isfile(value["path"]):
             # we set "bytes": None to not duplicate the data if they're already available locally
-            return {"bytes": None, "path": value.get("path")}
+            if value.get("path").endswith("pcm"):
```
lhoestq, line 97: you can access "path" directly, since we checked above that it exists
```suggestion
            if value["path"].endswith("pcm"):
```

src/datasets/features/audio.py
```diff
@@ -92,7 +94,23 @@ def encode_example(self, value: Union[str, dict]) -> dict:
...
+                # "PCM" is only have byte RAW-Audio data style.
+                if value.get("sampling_rate") is None:
+                    # At least, If you want to convert "PCM-byte" to "WAV-byte", you have to know sampling rate
+                    raise KeyError("If you want to use pcm, you have to know your 'sampling_rate'!")
```
lhoestq, line 101: small suggestion for the error message:
```suggestion
                    raise KeyError("To use PCM files, please specify a 'sampling_rate' in addition to the 'path'.")
```

src/datasets/features/audio.py
```diff
@@ -92,7 +94,23 @@ def encode_example(self, value: Union[str, dict]) -> dict:
...
+                if value.get("bytes"):
+                    # If we already had PCM-byte, we don`t have to make "read file, make bytes" (just use it!)
+                    # I used fairseq's using PCM data (FileAudioDataset)
```
lhoestq, line 104: I don't understand this comment, does this mean the code comes from fairseq ?
```suggestion
                    # I used fairseq's using PCM data (FileAudioDataset)
```

lhoestq (commented):
Awesome thanks ! Could you also add tests in `tests/features/test_audio.py` ?

Maybe add a small pcm file in `tests/features/data` and check that everything works as expected in tests cases like `test_audio_encode_example_pcm` and `test_audio_decode_example_pcm` for example.

mariosasko:
@lhoestq Maybe I'm missing something, but what's the reason to read and encode PCM files to WAV in `Audio.encode_example`. Isn't the whole purpose of the decodable types to operate on raw files whenever possible? IMO this PR should only modify `Audio.decode_example` to support PCM files/bytes decoding.

mariosasko:
But [`scipy.io.wavfile.read`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.io.wavfile.read.html), which is used for reading such files, returns a file's sampling rate. The only tricky part is [resampling](https://stackoverflow.com/questions/33682490/how-to-read-a-wav-file-using-scipy-at-a-different-sampling-rate) to a different sampling rate than the default one.

lhoestq:
How does it get the sampling rate of a PCM file then ? According to [SO](https://stackoverflow.com/a/57027667/17517845) it's not possible to infer it from the file alone

YooSungHyun:
@lhoestq 

> Because the PCM file is not enough, we also need the `sampling_rate` associated to it. Therefore the two alternatives are either:
> 
> * convert to WAV
> * add a `sampling_rate` field to the Audio arrow storage (not sure how it would behave for backward compatibility though)

"convert to WAV" way is need to many audio information (like channel, sampling rate, PCM type(16 or 32 etc...)  etc....)
i think that way gets higher complexity than "sampling_rate + audio float byte packed" way

YooSungHyun:
> Awesome thanks ! Could you also add tests in `tests/features/test_audio.py` ?
> 
> Maybe add a small pcm file in `tests/features/data` and check that everything works as expected in tests cases like `test_audio_encode_example_pcm` and `test_audio_decode_example_pcm` for example.

@lhoestq  how can i test test_audio.py? where is "__main__" func?
do you have some example or guideline?

YooSungHyun:
@lhoestq
maybe, if i write test code, i have to commit test_audio.py and send pr?
because, we need to keep `test_audio_encode_example_pcm` and `test_audio_decode_example_pcm` method after my pr merged?

YooSungHyun:
test complete & commit my test_audio.py

AND, some change in my code.

audio.py
i think "sampling_rate" is already Audio object initial variable. so, we don`t have to use input parameter.

test_audio.py
we can check "PCM" file to path (exactly, extenstion)
so, test case has to know `path`. if only have `bytes`, we don`t know that is "PCM" or not

YooSungHyun:
@lhoestq
hello????

src/datasets/features/audio.py
```diff
@@ -92,7 +94,22 @@ def encode_example(self, value: Union[str, dict]) -> dict:
...
+                    raise KeyError("To use PCM files, please specify a 'sampling_rate' in Audio object")
+                if value.get("bytes"):
+                    # If we already had PCM-byte, we don`t have to make "read file, make bytes" (just use it!)
+                    bytes_value = np.frombuffer(value["bytes"], dtype=np.int16).astype(np.float32) / 32767
+                else:
+                    bytes_value = np.memmap(value["path"], dtype="h", mode="r").astype(np.float32) / 32767
+
+                buffer = BytesIO(bytes())
+                wavfile.write(buffer, self.sampling_rate, bytes_value)
+                return {"bytes": buffer.getvalue(), "path": value.get("path")}
```
lhoestq, line 110: The path points to the pcm file, while here we are passing the new WAV data, so we need to set the path to None:
```suggestion
                return {"bytes": buffer.getvalue(), "path": None}
```

src/datasets/features/audio.py
```diff
@@ -131,6 +148,8 @@ def decode_example(self, value: dict, token_per_repo_id=None) -> dict:
                 array, sampling_rate = self._decode_non_mp3_path_like(
                     path, "opus", token_per_repo_id=token_per_repo_id
                 )
+        elif path is not None and path.endswith("pcm"):
+            sampling_rate, array = wavfile.read(file)
```
lhoestq, line 152: It's not possible to decode a pcm file, since we don't know the sampling rate, you can raise an error here:
```suggestion
        elif path is not None and path.endswith("pcm"):
            raise ValueError(f"Unable to decode PCM file at {path} since the sampling rate is unknown.")
```

lhoestq (commented):
Thanks for the tests ! This is awesome :)

I left two comments. Also to fix the CI feel free need to merge the `master` branch into yours

src/datasets/features/audio.py
YooSungHyun: oh! i think if we have converted pcm2wav file bytes, we can open to soundfile!
so we don`t need path good idea!

src/datasets/features/audio.py
YooSungHyun: how about don`t compare path? that is not necessary anymore i think.
so i will commit again

YooSungHyun:
@lhoestq
plz, check again!
we can merge to master?

YooSungHyun:
@lhoestq 
test_audio.py
if we don`t use path in pcm, test-case need to be changed
so, we check path just None

YooSungHyun:
@lhoestq oh, i pull master in my branch, cicleci is well

so glad! @lhoestq

src/datasets/features/audio.py
```diff
@@ -92,7 +94,22 @@ def encode_example(self, value: Union[str, dict]) -> dict:
...
+                    # At least, If you want to convert "PCM-byte" to "WAV-byte", you have to know sampling rate
+                    raise KeyError("To use PCM files, please specify a 'sampling_rate' in Audio object")
+                if value.get("bytes"):
+                    # If we already had PCM-byte, we don`t have to make "read file, make bytes" (just use it!)
+                    bytes_value = np.frombuffer(value["bytes"], dtype=np.int16).astype(np.float32) / 32767
+                else:
+                    bytes_value = np.memmap(value["path"], dtype="h", mode="r").astype(np.float32) / 32767
+
+                buffer = BytesIO(bytes())
+                wavfile.write(buffer, self.sampling_rate, bytes_value)
```
lhoestq, line 109: self.sampling_rate is the sampling rate we should obtain after decoding (and this is already implemented in decode_example so we're fine ^^)

When encoding, this is the input sampling rate that we should use:
```suggestion
                wavfile.write(buffer, value["sampling_rate"], bytes_value)
```

tests/features/test_audio.py
```diff
@@ -126,6 +146,18 @@ def test_audio_decode_example_opus(shared_datadir):
     assert decoded_example["sampling_rate"] == 48000
 
 
+def test_audio_decode_example_pcm(shared_datadir):
+    sampling_rate = 16_000
```
lhoestq, line 150: Maybe this test can make sure decoding uses the sampling rate of Audio(), not the sampling rate of the input example dictionary

```suggestion
@pytest.mark.parametrize("sampling_rate", [16_000, 48_000])
def test_audio_decode_example_pcm(shared_datadir, sampling_rate):
```

tests/features/test_audio.py
```diff
@@ -126,6 +146,18 @@ def test_audio_decode_example_opus(shared_datadir):
...
+    audio_path = str(shared_datadir / "test_audio_16000.pcm")
+    audio_input = {"path": audio_path, "sampling_rate": sampling_rate}
+    audio = Audio(sampling_rate=16_000)
```
lhoestq, line 153:
```suggestion
    audio_input = {"path": audio_path, "sampling_rate": 16_000}
    audio = Audio(sampling_rate=sampling_rate)
```

lhoestq (commented):
You can ignore the `multiprocess` error in the CI, this is unrelated to your PR and we're working on it ;)

Thanks a lot for the changes, I spotted one thing that requires a small change: when writing the WAV data, we should use the sampling rate from the input example dictionary (the same that contains "path") - and not the sampling rate from the Audio attribute.

Indeed the sampling rate of the Audio attribute is the expected sampling rate when decoding the example. When we encode it we must specify the sampling rate of the PCM file

YooSungHyun:
@lhoestq thx for comment!
test_audio.py test complete. it runs sucessfully
and, self.get("sampling_rate") -> value.get("sampling_rate") changed

and, some comment is not agreed to me, plz check my sub comment!

src/datasets/features/audio.py
```diff
@@ -3,8 +3,10 @@
 from io import BytesIO
 from typing import TYPE_CHECKING, Any, ClassVar, Dict, Optional, Union
 
+import numpy as np
 import pyarrow as pa
 from packaging import version
+from scipy.io import wavfile
```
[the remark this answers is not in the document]
lhoestq:
```suggestion
```

lhoestq (commented):
Thank you ! It looks all good now.

I just removed the scipy import, since we already import soundfile:

Issue #4398, opened 2022-05-24
Calling `cast_column`/`remove_columns` and a sequence of `map` operations ends up making `faiss` fail with `ValueError`


alvarobartt:
First of all, sorry in advance for the unclear title, but this bug is weird to explain (at least for me), so I tried my best to summarize all the information in this issue.

## Describe the bug

Calling a certain combination of operations over a 🤗 `Dataset` and then trying to calculate the `faiss` index with `.add_faiss_index` ends up throwing an exception while trying to set the format back of a previously removed column. But this just happens over certain conditions... I'll present some scenarios below!

## Steps to reproduce the bug

Assuming the following dataset named `sample.csv` with some IMDb data:

```csv
id,title,summary
1877830,"The Batman","When a sadistic serial killer begins murdering key political figures in Gotham, Batman is forced to investigate the city's hidden corruption and question his family's involvement."
9419884,"Doctor Strange in the Multiverse of Madness","Doctor Strange teams up with a mysterious teenage girl from his dreams who can travel across multiverses, to battle multiple threats, including other-universe versions of himself, which threaten to wipe out millions across the multiverse. They seek help from Wanda the Scarlet Witch, Wong and others."
11138512,"The Northman","From visionary director Robert Eggers comes The Northman, an action-filled epic that follows a young Viking prince on his quest to avenge his father's murder." 
1745960,"Top Gun: Maverick","After more than thirty years of service as one of the Navy's top aviators, Pete Mitchell is where he belongs, pushing the envelope as a courageous test pilot and dodging the advancement in rank that would ground him."
```

We'll be able to reproduce the bug using the following piece of code:

```python
# Sample code to reproduce the bug
from transformers import DPRContextEncoder, DPRContextEncoderTokenizer
import torch

torch.set_grad_enabled(False)
ctx_encoder = DPRContextEncoder.from_pretrained("facebook/dpr-ctx_encoder-single-nq-base")
ctx_tokenizer = DPRContextEncoderTokenizer.from_pretrained("facebook/dpr-ctx_encoder-single-nq-base")

from datasets import load_dataset, Value

ds = load_dataset("csv", data_files=["sample.csv"], split="train")
ds = ds.cast_column("id", Value("int32")) # from `int64` to `int32`
ds = ds.map(lambda x: {"inputs": f"{ctx_tokenizer.sep_token}".join(["title", "summary"])})
ds = ds.remove_columns(["title", "summary"])

def generate_embeddings(x):
[...]

PR #4411, opened 2022-05-26
Update `_format_columns` in `remove_columns`

alvarobartt:
As explained at #4398, when calling `dataset.add_faiss_index` under certain conditions when calling a sequence of operations `cast_column`, `map`, and `remove_columns`, it fails as it's trying to look for already removed columns.

So on, after testing some possible fixes, it seems that setting the dataset format right after removing the columns seems to be working fine, so I had to add a call to `.set_format` in the `remove_columns` function.

Hope this helps!


alvarobartt:
🤗 This PR closes https://github.com/huggingface/datasets/issues/4398

mariosasko:
I made a small change to make the fix easier to understand.

alvarobartt:
Hi, @mariosasko thanks! It makes sense, sorry I'm not that familiar with `datasets` code 😩

albertvillanova (changes requested):
Hi @alvarobartt thanks for the fix.

I think this PR would need a regression test, so that we are sure we will not reintroduce the bug in the future. Would you mind to add it?

Feel free to ping me if you need some help.

alvarobartt:
Sure @albertvillanova I'll do that later today and ping you once done, thanks! :hugs:

alvarobartt:
Hi again @albertvillanova! Let me know if those tests are fine 🤗

albertvillanova:
Hi @alvarobartt,

I had to edit edit your file using the GitHub UI in order to force the re-running of CI tests: I don't know why previously, after your last commit, it was not triggered. I did so because I think your tests are failing.

First I tried to do it using my terminal, but I don't have push right to your PR branch: I would ask you next time you open a PR, please mark the checkbox "Allow edits from maintainers": https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

alvarobartt:
Hi @albertvillanova, let me check those again! And regarding that checkbox I thought it was already checked so my bad there 😩

albertvillanova:
@lhoestq any idea why the CI is not triggered?

albertvillanova:
@alvarobartt I have tested locally and the tests continue failing.

I think there is an implementation error: `new_dset._format_columns` is always `None` in those cases.

alvarobartt:
You're right @albertvillanova I was indeed running the tests with `datasets==2.2.0` rather than with the branch version, I'll check it again! Sorry for the inconvenience...

alvarobartt:
> @alvarobartt I have tested locally and the tests continue failing.
> 
> I think there is a basis error: `new_dset._format_columns` is always `None` in those cases.

In order to have some regressions tests for the fixed scenario, I've manually updated the value of `_format_columns` in the `ArrowDataset` so as to check whether it's updated or not right after calling `remove_columns`, and it does behave as expected, so with the latest version of this branch the reported issue doesn't occur anymore.

alvarobartt (commented):
🤗 Regressions tests have already been included!

albertvillanova (changes requested):
Thanks.

I would remove the tests assertions: `self.assertEqual(..., None)`.
The regression tests of the fix you implement only apply when `self._format_columns` is not `None`.

What about adding an analogue regression test also in `tests/test_dataset_dict.py`?

alvarobartt:
Hi again @albertvillanova sorry I was on leave! I'll do that ASAP :hugs:

alvarobartt:
@albertvillanova, does it make sense to add regression tests for `DatasetDict`? As `DatasetDict` doesn't have the attribute `_format_columns`, when we call `remove_columns` over a `DatasetDict` it removes the columns and updates the attributes of each split which is an `ArrowDataset`. So on, we can either update first the `_format_columns` attribute of each split and then remove the columns over the `DatasetDict`, or just loop over the splits of `DatasetDict` and remove the columns right after updating `_format_columns` of each `ArrowDataset`.

I assume that the best regression test is the one implemented (mentioned first above), let me know if there's a better way to do that 👍🏻

lhoestq:
I think there's already a decorator to support transmitting the right `_format_columns`: `@transmit_format`, have you tried adding this decorator to `remove_columns` ?

alvarobartt:
> 

I can check now!

alvarobartt:
> I think there's already a decorator to support transmitting the right `_format_columns`: `@transmit_format`, have you tried adding this decorator to `remove_columns` ?

Hi @lhoestq I can check now!

alvarobartt:
It worked indeed @lhoestq, thanks for the proposal and the review! 🤗

mariosasko:
Oops, I forgot about `@transmit_format`'s existence. From what I see, we should also use this decorator in `flatten`, `rename_column` and  `rename_columns`. 

@alvarobartt Let me know if you'd like to work on this (in a subsequent PR).

lhoestq (approved):
Thanks ! Merging now :)

PR #4415, opened 2022-05-27
Update existing `dataset_infos.json` with new split info in `Dataset.push_to_hub`

mariosasko:
TODO:
- [ ] handle token + `{Audio, Image}.embed_storage`
- [ ] tests

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -17,7 +17,6 @@
 
 import contextlib
 import copy
-import itertools
 import json
 import os
 import shutil
@@ -60,6 +59,7 @@ from tqdm.auto import tqdm
 from . import config
 from .arrow_reader import ArrowReader
 from .arrow_writer import ArrowWriter, OptimizedTypedSequence
+from .download.download_config import DownloadConfig
 from .download.streaming_download_manager import xgetsize
 from .features import Audio, ClassLabel, Features, Image, Sequence, Value
 from .features.features import FeatureType, _ArrayXD, decode_nested_example, pandas_types_mapper, require_decoding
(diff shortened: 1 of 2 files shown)


src/datasets/arrow_dataset.py
```diff
@@ -4140,6 +4146,38 @@ def push_to_hub(
         info_to_dump.splits = {
             split: SplitInfo(split, num_bytes=dataset_nbytes, num_examples=len(self), dataset_name=dataset_name)
         }
+        if config.DATASETDICT_INFOS_FILENAME in repo_files:
```
lhoestq, line 4149: If the dataset_infos.json file already exists, then it seems that the the previous `info_to_dump` defined a few lines above is overwritten

Maybe move the code that creates `info_to_dump` when there's no `datasets_infos.json` in a `else` statement after this `if` ?

lhoestq (commented):
Thanks for the fix !

src/datasets/arrow_dataset.py
mariosasko: There is an additional nested `if` inside this `if` block, so we need two identical `else` blocks (one for this `if` and one for the nested `if`). Let me know which style you prefer.

lhoestq: Oh indeed, makes sense !

src/datasets/arrow_dataset.py
```diff
@@ -4140,6 +4146,37 @@ def push_to_hub(
...
+            dataset_infos_path = cached_path(
+                hf_hub_url(repo_id, config.DATASETDICT_INFOS_FILENAME),
+                download_config=download_config,
+            )
+            with open(dataset_infos_path, encoding="utf-8") as f:
+                dataset_infos: DatasetInfosDict = json.load(f)
+                repo_info = DatasetInfo.from_dict(dataset_infos[next(iter(dataset_infos))])
+            logger.warning("Updating downloaded metadata with the new split.")
+
+            if len(repo_info.splits) > 1 or (len(repo_info.splits) == 1 and next(iter(repo_info.splits)) != split):
```
lhoestq, line 4162: maybe simpler this way
```suggestion
            if not repo_info.splits or list(repo_info.splits) != [split]:
```

lhoestq (approved):
Looks all good, thanks :)

PR #4421, opened 2022-05-30
Add extractor for bzip2-compressed files

asivokon:
This change enables loading bzipped datasets, just like any other compressed dataset.


lhoestq (approved):
Awesome thanks ! Moreover bz2 was already supported in streaming mode, so I think we now fully support bz2 thanks to you :)

PR #4423, opened 2022-05-31
Add new dataset MMChat

silverriver:
Hi, I am adding a new dataset MMChat. 
It seems that all tests are passed


datasets/mmchat/README.md
```diff
@@ -0,0 +1,274 @@
+---
+annotations_creators:
+- other
```
lhoestq, line 3:
```suggestion
- no-annotation
```

datasets/mmchat/README.md
```diff
@@ -0,0 +1,274 @@
...
+language_creators:
+- other
```
lhoestq, line 5:
```suggestion
- found
```

datasets/mmchat/README.md
```diff
@@ -0,0 +1,274 @@
...
+languages:
+- zh
+licenses:
+- mit
```
lhoestq, line 9: Same comment as in https://github.com/huggingface/datasets/pull/4431

> The data has been crawled from weibo, therefore I don't think this is license is right.
I don't know much about Weibo, but if I recall correctly they claim the copyright of all the content on the platform, and if this is the case you are not allowed to sublicense the data unless you have an agreement with Weibo.
> 
> Maybe you can set this tag to "other-weibo" instead of "mit", and in the Licensing Information section you can redirect users to the terms of service of Weibo (and ideally summarize what people are allowed to do with this data) ?

lhoestq (commented):
Hi ! Thank for sharing this dataset :)

I added the same comments as in https://github.com/huggingface/datasets/pull/4431

Also it seems that you already have the dataset in https://huggingface.co/datasets/silver/mmchat, so it's unnecessary to add it here

lhoestq:
Thanks ! As for https://github.com/huggingface/datasets/pull/4431 please also update the licensing section in https://huggingface.co/datasets/silver/mmchat ;)

Then if it's fine for you feel free to close this PR

Issue #4422, opened 2022-05-30
Cannot load timit_asr data set


bhaddow:
## Describe the bug
I am trying to load the timit_asr data set. I have tried with a copy from the LDC, and a copy from deepai. In both cases they fail with a "duplicate key" error. With the LDC version I have to convert the file extensions all to upper-case before I can load it at all.

## Steps to reproduce the bug
```python
timit = datasets.load_dataset("timit_asr", data_dir = "/path/to/dataset")
# Sample code to reproduce the bug
```

## Expected results

The data set should load without error. It worked for me before the LDC url change.

## Actual results
```
datasets.keyhash.DuplicatedKeysError: FAILURE TO GENERATE DATASET !
Found duplicate Key: SA1
Keys should be unique and deterministic in nature
```

## Environment info

- `datasets` version:
- `datasets` version: 2.2.2
- Platform: Linux-5.4.0-90-generic-x86_64-with-glibc2.17
- Python version: 3.8.12
- PyArrow version: 8.0.0
- Pandas version: 1.4.2

albertvillanova:
Thanks for reporting, @bhaddow.

I'm fixing it.

bhaddow:
Thanks for the quick fix!

albertvillanova:
@bhaddow we have also made a fix so that you don't have to convert to uppercase the file extensions of the LDC data.

Would you mind checking if it works OK now for you and report if there are any issues? Thanks.

bhaddow:
Hi @albertvillanova  -It loads fine on a copy of the data from deepai - although I have to remove the copies of the .WAV files (with extension .WAV,wav). On a copy of the data that was obtained from the LDC, the glob still fails to find the files. The LDC copy looks like it was copied from CD, in 2004, so the structure may be different to a current download.

bhaddow:
Ah, if I change the train/ and test/ directories to TRAIN/ and TEST/  then it works!

albertvillanova:
Thanks for your investigation and report, @bhaddow. I'm adding another fix for the TRAIN/train and TEST/test directory names.

PR #4425, opened 2022-05-31
Make extensions case-insensitive in timit_asr dataset

albertvillanova:
Related to #4422.

### DIFF
--- a/datasets/timit_asr/timit_asr.py
+++ b/datasets/timit_asr/timit_asr.py
@@ -129,14 +129,18 @@ class TimitASR(datasets.GeneratorBasedBuilder):
     def _generate_examples(self, split, data_dir):
         """Generate examples from TIMIT archive_path based on the test/train csv information."""
         # Iterating the contents of the data to extract the relevant information
-        for wav_path in sorted(Path(data_dir).glob(f"**/{split.upper()}/**/*.WAV")):
+        wav_paths = sorted(Path(data_dir).glob(f"**/{split.upper()}/**/*.wav"))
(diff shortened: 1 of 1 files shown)


datasets/timit_asr/timit_asr.py
```diff
@@ -129,14 +129,18 @@ def _split_generators(self, dl_manager):
     def _generate_examples(self, split, data_dir):
         """Generate examples from TIMIT archive_path based on the test/train csv information."""
         # Iterating the contents of the data to extract the relevant information
-        for wav_path in sorted(Path(data_dir).glob(f"**/{split.upper()}/**/*.WAV")):
+        wav_paths = sorted(Path(data_dir).glob(f"**/{split.upper()}/**/*.wav"))
+        wav_paths = wav_paths if wav_paths else sorted(Path(data_dir).glob(f"**/{split.upper()}/**/*.WAV"))
```
lhoestq, line 133: This is fine if it can't have both .WAV and .wav files in the same folder - are we sure about this ?

lhoestq (commented):
Awesome thanks ! Do you think it's worth updating the dataset_infos.json file as well to check if it generates the right number of examples ?

datasets/timit_asr/timit_asr.py
albertvillanova: @lhoestq no, we are not sure... We need to have the data from LDC.

Otherwise, I could concatenate both globs, but I'm afraid it could give twice the filenames on Windows machines...

lhoestq: I think it's the same data on deepai than on LDC. I've only used the data from deepai to update the PR.

We can mergi this one for now and ask the user who has the data from LDC to run it and report if there's any issue I guess

lhoestq (approved):

PR #4427, opened 2022-05-31
Add HF.co for PRs/Issues for specific datasets

lhoestq:
As in https://github.com/huggingface/transformers/pull/17485, issues and PR for datasets under a namespace have to be on the HF Hub


albertvillanova (approved):
Good idea. Thanks.

julien-c (approved):

Issue #4428, opened 2022-05-31
Errors when building dummy data if you use nested _URLS


silverriver:
## Describe the bug
When making dummy data with the `datasets-cli dummy_data` tool,
an error will be raised if you use a nested _URLS in your dataset script.

Traceback (most recent call last):
  File "/home/name/LCCC/datasets/src/datasets/commands/datasets_cli.py", line 43, in <module>
    main()
  File "/home/name/LCCC/datasets/src/datasets/commands/datasets_cli.py", line 39, in main
    service.run()
  File "/home/name/LCCC/datasets/src/datasets/commands/dummy_data.py", line 311, in run
    self._autogenerate_dummy_data(
  File "/home/name/LCCC/datasets/src/datasets/commands/dummy_data.py", line 337, in _autogenerate_dummy_data
    dataset_builder._split_generators(dl_manager)
  File "/home/name/.cache/huggingface/modules/datasets_modules/datasets/personal_dialog/559332bced5eeafa7f7efc2a7c10ce02cee2a8116bbab4611c35a50ba2715b77/personal_dialog.py", line 108, in _split_generators
    data_dir = dl_manager.download_and_extract(urls)
  File "/home/name/LCCC/datasets/src/datasets/commands/dummy_data.py", line 56, in download_and_extract
    dummy_output = self.mock_download_manager.download(url_or_urls)
  File "/home/name/LCCC/datasets/src/datasets/download/mock_download_manager.py", line 130, in download
    return self.download_and_extract(data_url)
  File "/home/name/LCCC/datasets/src/datasets/download/mock_download_manager.py", line 122, in download_and_extract
    return self.create_dummy_data_dict(dummy_file, data_url)
  File "/home/name/LCCC/datasets/src/datasets/download/mock_download_manager.py", line 165, in create_dummy_data_dict
    if isinstance(first_value, str) and len(set(dummy_data_dict.values())) < len(dummy_data_dict.values()):
TypeError: unhashable type: 'list'

## Steps to reproduce the bug

You can use my dataset script implemented here:
https://github.com/silverriver/datasets/blob/2ecd36760c40b8e29b1137cd19b5bad0e19c76fd/datasets/personal_dialog/personal_dialog.py

```python
datasets_cli dummy_data datasets/personal_dialog --auto_generate
```

[...]

PR #4434, opened 2022-06-01
Fix dummy dataset generation script for handling nested types of _URLs

silverriver:
It seems that when user specify nested _URLs structures in their dataset script. An error will be raised when generating dummy dataset.

I think the types of all elements in `dummy_data_dict.values()` should be checked because they may have different types.

Linked to issue #4428 

PS: I am not sure whether my code fix this issue in a proper way.


lhoestq (approved):
Cool thanks ! LGTM :)

PR #4440, opened 2022-06-02
Update docs around audio and vision

stevhliu:
As part of the strategy to center the docs around the different modalities, this PR updates the quickstart to include audio and vision examples. This improves the developer experience by making audio and vision content more discoverable, enabling users working in these modalities to also quickly get started without digging too deeply into the docs.

Other changes include:

- Moved the installation guide to the Get Started section because it should be part of a user's onboarding to the library before exploring tutorials or how-to's.
- Updated the native TF code at creating a `tf.data.Dataset` because it was throwing an error. The `to_tensor()` bit was redundant and removing it fixed the error (please double-check me here!).
- Added some UI components to the quickstart so it's easier for users to navigate directly to the relevant section with context about what to expect.
- Reverted to the code tabs for content that don't have any framework-specific text. I think this saves space compared to the code blocks. We'll still use the code blocks if the `torch` text is different from the `tf` text.

Let me know what you think, especially if we should include some code samples for training a model in the audio/vision sections. I left this out since we already showed it in the NLP section. I want to keep the focus on using Datasets to load and process a dataset, and not so much the training part. Maybe we can add links to the Transformers docs instead?

### DIFF
--- a/docs/source/quickstart.mdx
+++ b/docs/source/quickstart.mdx
@@ -1,8 +1,8 @@
 # Quickstart
 
-This quickstart is intended for developers ready to dive into the code and see an end-to-end example of how to integrate 🤗 Datasets into their model training workflow. If you are a beginner or a non-developer, we recommend starting with our [tutorials](./tutorial), where you will get a more thorough introduction.
+This quickstart is intended for developers who are ready to dive into the code and see an example of how to integrate 🤗 Datasets into their model training workflow. If you're a beginner, we recommend starting with our [tutorials](./tutorial), where you'll get a more thorough introduction.
 
-Each dataset is unique, and depending on the task, some datasets may require additional steps to prepare the data. But you can always use 🤗 Datasets tools to load and process a dataset. Choose the type of dataset you want to work with, and let's get started!
+Each dataset is unique, and depending on the task, some datasets may require additional steps to prepare it for training. But you can always use 🤗 Datasets tools to load and process a dataset. The fastest and easiest way to get started is by loading an existing dataset from the [Hugging Face Hub](https://huggingface.co/datasets). There are thousands of datasets to choose from, spanning many tasks. Choose the type of dataset you want to work with, and let's get started!
 
 <div class="mt-4">
    <div class="w-full flex flex-col space-y-4 md:space-y-0 md:grid md:grid-cols-3 md:gap-y-4 md:gap-x-5">
@@ -16,14 +16,14 @@ Each dataset is unique, and depending on the task, some datasets may require add
       </a>
       <a class="!no-underline border dark:border-gray-700 p-5 rounded-lg shadow hover:shadow-lg" href="/docs/datasets/quickstart#nlp"
          ><div class="w-full text-center bg-gradient-to-r from-orange-300 via-red-400 to-violet-500 rounded-lg py-1.5 font-semibold mb-5 text-white text-lg leading-relaxed">NLP</div>
-         <p class="text-gray-700">Tokenize a dataset and train a model to determine whether a pair of sentences have the same meaning.</p>
+         <p class="text-gray-700">Tokenize a dataset and get it ready for a model to determine whether a pair of sentences have the same meaning.</p>
       </a>
    </div>
 </div>
 
 <Tip>
 
-Check out [Chapter 5](https://huggingface.co/course/chapter5/1?fw=pt) of the Hugging Face course to learn more about other important topics such as loading remote or local datasets, tools for cleaning up a dataset and creating your own dataset.
+Check out [Chapter 5](https://huggingface.co/course/chapter5/1?fw=pt) of the Hugging Face course to learn more about other important topics such as loading remote or local datasets, tools for cleaning up a dataset, and creating your own dataset.
 
 </Tip>
 
@@ -77,9 +77,9 @@ Audio datasets are loaded just like text datasets. However, an audio dataset is 
  'sampling_rate': 16000}
 ```
 
-**4**. Create a function that will preprocess the audio `array` with the feature extractor, and truncate and pad the sequences into tidy rectangular tensors. The most important thing to remember here is to call the audio `array` in the feature extractor since the `array` - the actual speech signal - is the model input.
+**4**. Create a function to preprocess the audio `array` with the feature extractor, and truncate and pad the sequences into tidy rectangular tensors. The most important thing to remember is to call the audio `array` in the feature extractor since the `array` - the actual speech signal - is the model input.
 
-Once you have the `preprocess_function`, use the [`~Dataset.map`] function to speed up processing by applying the function to batches of examples in the dataset.
+Once you have a preprocessing function, use the [`~Dataset.map`] function to speed up processing by applying the function to batches of examples in the dataset.
 
 ```py
 >>> def preprocess_function(examples):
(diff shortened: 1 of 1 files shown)


docs/source/quickstart.mdx
```diff
@@ -1,66 +1,83 @@
...
+      <a class="!no-underline border dark:border-gray-700 p-5 rounded-lg shadow hover:shadow-lg" href="/docs/datasets/quickstart#audio"
+         ><div class="w-full text-center bg-gradient-to-br from-violet-400 to-orange-500 rounded-lg py-1.5 font-semibold mb-5 text-white text-lg leading-relaxed">Audio</div>
+         <p class="text-gray-700">Resample an audio dataset and get it ready for a model to classify what type of banking issue a speaker is calling about.</p>
+      </a>
+      <a class="!no-underline border dark:border-gray-700 p-5 rounded-lg shadow hover:shadow-lg" href="/docs/datasets/quickstart#vision"
+         ><div class="w-full text-center bg-gradient-to-br from-cyan-400 to-pink-500 rounded-lg py-1.5 font-semibold mb-5 text-white text-lg leading-relaxed">Vision</div>
+         <p class="text-gray-700">Apply a data augmentation to an image dataset and get it ready for a model to diagnose disease in bean plants.</p>
+      </a>
+   </div>
+</div>
```
mariosasko, line 22: The NLP and Audio labels don't render [here](https://moon-ci-docs.huggingface.co/docs/datasets/pr_4440/quickstart#quickstart) in my browser due to color selection (emerald and violet). Maybe [this](https://stackoverflow.com/questions/70704377/default-colors-given-in-tailwind-documentation-are-not-working) SO thread can help resolve the issue.

mariosasko (commented):
Love the changes!

mariosasko:
> Let me know what you think, especially if we should include some code samples for training a model in the audio/vision sections. I left this out since we already showed it in the NLP section. I want to keep the focus on using Datasets to load and process a dataset, and not so much the training part. Maybe we can add links to the Transformers docs instead?

We plan to address this with end-to-end examples for each modality, which are more focused on preprocessing than the ones in the Transformers docs.

docs/source/quickstart.mdx
stevhliu: Oh weird! They render normally in the `doc-builder` preview, and we should have support for these colors since we're using Tailwind v3. @mishig25 do I need to add emerald to the color palette (I see violet is already there)?

Screen Shot 2022-06-07 at 9 41 23 AM

mishig25: oh indeed, that is expected. 
everytime, a new tailwind class gets used, it needs to be added to [internal link](https://github.com/huggingface/moon-landing/blob/main/server/views/pages/DocPage/TailwindWhitelist.svelte)
because tailwind [removes everything that is not whitelisted](https://v2.tailwindcss.com/docs/optimizing-for-production#removing-unused-css)

julien-c: i love this screenshot but i would suggest to change the order of modalities (for instance Audio, Vision, NLP) to "break" the perception that NLP is the first class citizen (which it is... but we're trying to change that perception 😆)

cc @osanseviero @gary149

docs/source/quickstart.mdx
```diff
@@ -73,107 +90,41 @@ The next step is to tokenize the text in order to build sequences of integers th
...
-   ...,
-   [1, 1, 1, ..., 0, 0, 0],
-   [1, 1, 1, ..., 0, 0, 0],
-   [1, 1, 1, ..., 0, 0, 0]], dtype=int32)>}, <tf.Tensor: shape=(32,), dtype=int64, numpy=
-array([1, 0, 1, 0, 1, 1, 0, 1, 0, 0, 0, 0, 1, 1, 0, 0, 0, 1, 0, 1, 1, 1,
-   0, 1, 1, 1, 0, 0, 1, 1, 1, 0])>)
+
+>>> dataset.set_format(type="tensorflow", columns=["input_ids", "token_type_ids", "attention_mask", "labels"])
+>>> features = {x: dataset[x] for x in ["input_ids", "token_type_ids", "attention_mask"]}
+>>> tf_dataset = tf.data.Dataset.from_tensor_slices((features, dataset["labels"])).batch(32)
```
lhoestq, line 122: We now have a memory efficient `to_tf_dataset`, maybe we may mention it in the quickstart instead.

`from_tensor_slices` brings everything in memory

lhoestq (commented):
Awesome thanks !

> Let me know what you think, especially if we should include some code samples for training a model in the audio/vision sections. I left this out since we already showed it in the NLP section.

I'd add the conversion to pytorch DataLoader and TF Dataset as well for audio and vision, if it's not too much information. But I think the training loop itself needs to be part of an end-to-end example as @mariosasko suggested.

As soon as there is a pytorch DataLoader or a TF Dataset, it means we're ready for training, I don't think it's necessary to show the complexity of one particular task/model and how it's trained in the quickstart.

Maybe the quickstart can redirect to its corresponding end-to-end example for those who would like to see a complete example with more context ? Something like "Want to see more in a concrete example ? See how a dataset can be prepared for speech recognition with transformers, etc."

docs/source/quickstart.mdx
```diff
@@ -7,16 +7,16 @@ Each dataset is unique, and depending on the task, some datasets may require add
...
+         ><div class="w-full text-center bg-gradient-to-r from-rose-300 via-purple-400 to-sky-500 rounded-lg py-1.5 font-semibold mb-5 text-white text-lg leading-relaxed">Vision</div>
+         <p class="text-gray-700">Apply data augmentation to an image dataset and get it ready for a model to diagnose disease in bean plants.</p>
       </a>
       <a class="!no-underline border dark:border-gray-700 p-5 rounded-lg shadow hover:shadow-lg" href="/docs/datasets/quickstart#vision"
-         ><div class="w-full text-center bg-gradient-to-br from-cyan-400 to-pink-500 rounded-lg py-1.5 font-semibold mb-5 text-white text-lg leading-relaxed">Vision</div>
-         <p class="text-gray-700">Apply a data augmentation to an image dataset and get it ready for a model to diagnose disease in bean plants.</p>
+         ><div class="w-full text-center bg-gradient-to-r from-orange-300 via-red-400 to-violet-500 rounded-lg py-1.5 font-semibold mb-5 text-white text-lg leading-relaxed">NLP</div>
+         <p class="text-gray-700">Tokenize a dataset and train a model to determine whether a pair of sentences have the same meaning.</p>
```
lhoestq, line 19: Maybe you can also re-order the actual content of the quick start ?

docs/source/quickstart.mdx
```diff
@@ -7,16 +7,16 @@ Each dataset is unique, and depending on the task, some datasets may require add
...
-         <p class="text-gray-700">Tokenize a dataset and train a model to determine whether a pair of sentences have the same meaning.</p>
+         ><div class="w-full text-center bg-gradient-to-r from-violet-300 via-sky-400 to-green-500 rounded-lg py-1.5 font-semibold mb-5 text-white text-lg leading-relaxed">Audio</div>
+         <p class="text-gray-700">Resample an audio dataset and get it ready for a model to classify what type of banking issue a speaker is calling about.</p>
       </a>
       <a class="!no-underline border dark:border-gray-700 p-5 rounded-lg shadow hover:shadow-lg" href="/docs/datasets/quickstart#audio"
-         ><div class="w-full text-center bg-gradient-to-br from-violet-400 to-orange-500 rounded-lg py-1.5 font-semibold mb-5 text-white text-lg leading-relaxed">Audio</div>
-         <p class="text-gray-700">Resample an audio dataset and get it ready for a model to classify what type of banking issue a speaker is calling about.</p>
+         ><div class="w-full text-center bg-gradient-to-r from-rose-300 via-purple-400 to-sky-500 rounded-lg py-1.5 font-semibold mb-5 text-white text-lg leading-relaxed">Vision</div>
```
lhoestq, line 14: I still don't see the color gradient for vision

lhoestq (commented):
Love the links you provided ! Later I think we can provide end-to-end examples in the `datasets` doc itself using a super simpler training loop instead of redirecting to the `transformers` doc that uses the `Trainer`

My last comments:

docs/source/quickstart.mdx
```diff
@@ -1,211 +1,285 @@
...
+# Quickstart
+
+This quickstart is intended for developers ready to dive into the code and see an end-to-end example of how to integrate 🤗 Datasets into their model training workflow. If you are a beginner or a non-developer, we recommend starting with our [tutorials](./tutorial), where you will get a more thorough introduction.
+
+Each dataset is unique, and depending on the task, some datasets may require additional steps to prepare the data. But you can always use 🤗 Datasets tools to load and process a dataset. Choose the type of dataset you want to work with, and let's get started!
+
+<div class="mt-4">
+   <div class="w-full flex flex-col space-y-4 md:space-y-0 md:grid md:grid-cols-3 md:gap-y-4 md:gap-x-5">
+      <a class="!no-underline border dark:border-gray-700 p-5 rounded-lg shadow hover:shadow-lg" href="/docs/datasets/quickstart#audio"
+         ><div class="w-full text-center bg-gradient-to-r from-violet-300 via-sky-400 to-green-500 rounded-lg py-1.5 font-semibold mb-5 text-white text-lg leading-relaxed">Audio</div>
```
mishig25, line 10: `via-sky-400` is still present here (which is not in Tailwind v2). Is it expected?

docs/source/quickstart.mdx
```diff
@@ -1,211 +1,285 @@
...
+         ><div class="w-full text-center bg-gradient-to-r from-orange-300 via-red-400 to-violet-500 rounded-lg py-1.5 font-semibold mb-5 text-white text-lg leading-relaxed">NLP</div>
+         <p class="text-gray-700">Tokenize a dataset and train a model to determine whether a pair of sentences have the same meaning.</p>
+      </a>
+   </div>
+</div>
 
 <Tip>
 
-For more detailed information on loading and processing a dataset, take a look at [Chapter 3](https://huggingface.co/course/chapter3/1?fw=pt) of the Hugging Face course! It covers additional important topics like dynamic padding, and fine-tuning with the Trainer API.
+For more detailed information about the 🤗 Datasets library, check out [Chapter 5](https://huggingface.co/course/chapter5/1?fw=pt) of the Hugging Face course! It covers other important topics such as loading remote or local datasets, tools for cleaning up a dataset and creating your own dataset.
```
mariosasko, line 26: Can we use a different wording here? As I understand, this would mean that our tutorial offers more info about the lib than the lib docs, which doesn't make sense 😄.

docs/source/quickstart.mdx
stevhliu: Yeah, I'd like to keep that if possible!

docs/source/quickstart.mdx
mishig25: should be good now https://moon-ci-docs.huggingface.co/docs/datasets/pr_4440/en/quickstart

lhoestq (approved):
Color gradients look all good now, thanks !

CI failures are unrelated to this PR btw - you can ignore them.

Feel free to merge if it's all good for you @stevhliu

PR #4451, opened 2022-06-06
Use newer version of multi-news with fixes

### LINKED ISSUE #4430: Add ability to load newer, cleaner version of Multi-News

The [Multi-News dataloader points to the original version of the Multi-News dataset](https://github.com/huggingface/datasets/blob/12540dd75015678ec6019f258d811ee107439a73/datasets/multi_news/multi_news.py#L47), but this has [known errors in it](https://github.com/Alex-Fabbri/Multi-News/issues/11). There exists a [newer version which fixes some of these issues](https://drive.google.com/open?id=1jwBzXBVv8sfnFrlzPnSUBHEEAbpIUnFq).

Unfortunately I don't think you can just replace this old URL with the new one, otherwise this could lead to issues with reproducibility.

Add a new version to the Multi-News dataloader that points to the updated dataset which has fixes for some known issues.

Replace the current URL to the original version to the dataset with the URL to the version with fixes.

**Additional context**

Would be happy to make a PR for this, could someone maybe point me to another dataloader that has multiple versions so I can see how this is handled in `datasets`?

JohnGiorgi:
Closes #4430.


mariosasko (approved):
I updated the dummy data and the info file. All looks good now. Thanks!

PR #4460, opened 2022-06-08
Drop Python 3.6 support

mariosasko:
Remove the fallback imports/checks in the code needed for Python 3.6 and update the requirements/CI files. Also, use Python types for the NumPy dtype wherever possible to avoid deprecation warnings in newer NumPy versions.


tests/test_arrow_dataset.py
```diff
@@ -3119,7 +3119,7 @@ def test_pickle_dataset_after_transforming_the_table(in_memory, method_and_param
 
 
 @pytest.mark.skipif(
-    os.name == "nt" and (os.getenv("CIRCLECI") == "true" or os.getenv("GITHUB_ACTIONS") == "true"),
+    os.name in ["nt", "posix"] and (os.getenv("CIRCLECI") == "true" or os.getenv("GITHUB_ACTIONS") == "true"),
```
lhoestq, line 3121: I don't think you need this ? The test currently passes on `main` on linux

src/datasets/utils/py_utils.py
```diff
@@ -515,19 +514,6 @@ class Pickler(dill.Pickler):
 
     dispatch = dill._dill.MetaCatchingDict(dill.Pickler.dispatch.copy())
 
-    def save_global(self, obj, name=None):
-        if sys.version_info[:2] < (3, 7) and _CloudPickleTypeHintFix._is_parametrized_type_hint(
```
lhoestq, line 519: I think you can also remove the `_CloudPickleTypeHintFix` class defined a bit later in this file

src/datasets/packaged_modules/text/dataset_infos.json
```diff
@@ -0,0 +1 @@
+{"bigscience": {"description": "", "citation": "", "homepage": "", "license": "", "features": {"text": {"dtype": "string", "id": null, "_type": "Value"}}, "post_processed": null, "supervised_keys": null, "task_templates": null, "builder_name": "text", "config_name": "bigscience", "version": {"version_str": "0.0.0", "description": null, "major": 0, "minor": 0, "patch": 0}, "splits": {"train": {"name": "train", "num_bytes": 938, "num_examples": 22, "dataset_name": "text"}}, "download_checksums": {"C:\\Users\\Mario\\Desktop\\bigscience\\biscience.txt": {"num_bytes": 892, "checksum": "1e1f85c9e2aefb6990dc6ec4a8805af1e5451ebecb7e9f50face10c83eed742e"}}, "download_size": 892, "post_processing_size": null, "dataset_size": 938, "size_in_bytes": 1830}}
```
lhoestq, line 1: I don't think this file is needed ^^

lhoestq (commented):
Thanks :) not sure why `torchaudio` fails to decode/resample in the CI though (it uses the latest version 0.12.0)

mariosasko:
[This comment](https://github.com/pytorch/audio/issues/2363#issuecomment-1179089175) explains the issue with MP3 decoding in `torchaudio` in the latest release (supports Python 3.7+). I fixed CI by pinning `torchaudio` to `<0.12.0`. Another way to fix this issue is by installing `ffmpeg` with conda or using the unofficial GH action. But I don't think it's worth making CI more complex, considering we can wait for the soundfile release, which should bring MP3 decoding, and drop the `torchaudio` dependency then.

julien-c:
Yay for dropping Python 3.6!

tests/test_arrow_dataset.py
mariosasko: It doesn't on Python 3.7. Notice the `test_dummy_dataset_serialize_s3` [failure](https://github.com/huggingface/datasets/runs/7467455443?check_suite_focus=true) in one of the previous test

lhoestq: Ok. This test is almost never run then, Maybe let's remove it completely in another PR, and make another one to make sure S3 export works as expected. Maybe we can use the `mockfs` fixture defined in https://github.com/huggingface/datasets/pull/4724 instead

mariosasko: This issue seems related to https://github.com/spulec/moto/issues/4750, but neither bumping `moto` to `3.0.0` (CI runs forever) nor pinning responses to `0.16` (the serialization fails) helps, so I agree we can disable these tests for now and replace them in another PR with the `mockfs` fixture

mariosasko:
I think we can merge in this state. Also, if an env has Python version < 3.7 installed, we raise a warning, so I don't think we even need to create (and pin) an issue to notify the contributors of this change.

lhoestq (approved):
Thanks ! LGTM :)

PR #4472, opened 2022-06-10
Fix 401 error for unauthticated requests to non-existing repos

lhoestq:
The hub now returns 401 instead of 404 for unauthenticated requests to non-existing repos.
This PR add support for the 401 error and fixes the CI fails on `master`


SBrandeis (approved):
👍

PR #4473, opened 2022-06-10
Add SST-2 dataset

albertvillanova:
Add SST-2 dataset.

Currently it is part of GLUE benchmark.

This PR adds it as a standalone dataset.

CC: @julien-c


julien-c:
on the hub this dataset is referenced as `sst-2` not `sst2` – is there a canonical orthography? If not, could we name it `sst-2`?

albertvillanova:
@julien-c, we normally do not use hyphens for dataset names: whenever the original dataset name contains a hyphen, we usually:
- either suppress it: CoNLL-2000 (`conll2000`), CORD-19 (`cord19`)
- or replace it with underscore: CC-News (`cc_news`), SQuAD-es (`squad_es`)

There are some exceptions though...

I think, the reason is there was a 1-to-1 relation with the corresponding Python module name.

I personally find confusing not having a rule and using both hyphens and underscores indistinctly: you never know which is the right orthography.

Whichever the decision we make, I would prefer to be applied consistently.

Also note that we already implemented this dataset as part of GLUE: https://github.com/huggingface/datasets/blob/master/datasets/glue/glue.py#L163
- dataset name: `glue`
- config name: `sst2`

On the other hand, let's see how other libraries name it:
- torchtext: `SST2` https://pytorch.org/text/stable/datasets.html#sst2
- OpenAI CLIP: `rendered-sst2` https://github.com/openai/CLIP/blob/main/data/rendered-sst2.md
- Kaggle: `SST2` https://www.kaggle.com/datasets/atulanandjha/stanford-sentiment-treebank-v2-sst2/version/22
- TensorFlow Datasets: `glue/sst2` https://www.tensorflow.org/datasets/catalog/glue#gluesst2

julien-c:
Ok, another option is to open PRs against the models in https://huggingface.co/models?datasets=sst-2 to change their dataset reference to `sst2`

(BTW some models refer to `sst2` already – but they're less popular: https://huggingface.co/models?datasets=sst2)

datasets/sst2/README.md
```diff
@@ -0,0 +1,179 @@
...
+source_datasets:
+- original
+task_categories:
+- text-classification
+task_ids:
+- sentiment-classification
+paperswithcode_id: sst
+pretty_name: Stanford Sentiment Treebank v2
+configs:
+- default
```
lhoestq, line 23: this is maybe not necessary, since it's the default anyway
```suggestion
```

lhoestq (approved):
Cool thanks ! I'm also in favor of opening PRs to the models referencing `sst-2` and use `sst2` instead

albertvillanova:
OK, I'm taking care of the subsequent PRs on models to align with this dataset name.

PR #4475, opened 2022-06-10
Improve error message for missing pacakges from inside dataset script

mariosasko:
Improve the error message for missing packages from inside a dataset script:

With this change, the error message for missing packages for `bigbench` looks as follows:
```
ImportError: To be able to use bigbench, you need to install the following dependencies:
        - 'bigbench' using 'pip install "bigbench @ https://storage.googleapis.com/public_research_data/bigbench/bigbench-0.0.1.tar.gz"'
```

And this is how it looked before:
```
ImportError: To be able to use bigbench, you need to install the following dependencies['bigbench', 'bigbench', 'bigbench', 'bigbench'] using 'pip install "bigbench @ https://storage.googleapis.com/public_research_data/bigbench/bigbench-0.0.1.tar.gz" bigbench bigbench bigbench' for instance'
```

### DIFF
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -288,17 +288,36 @@ def _download_additional_modules(
         local_imports.append((import_name, local_import_path))
 
     # Check library imports
-    needs_to_be_installed = []
+    needs_to_be_installed = {}
     for library_import_name, library_import_path in library_imports:
         try:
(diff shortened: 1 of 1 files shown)


mariosasko:
Closing in favor of #4484.

Issue #2767, opened 2021-08-06
equal operation to perform unbatch for huggingface datasets 


dorooddorood606:
Hi
I need to use "unbatch" operation in tensorflow on a huggingface dataset, I could not find this operation, could you kindly direct me how I can do it, here is the problem I am trying to solve:

I am considering "record" dataset in SuperGlue and I need to replicate each entery of the dataset for each answer, to make it similar to what T5 originally did:

https://github.com/google-research/text-to-text-transfer-transformer/blob/3c58859b8fe72c2dbca6a43bc775aa510ba7e706/t5/data/preprocessors.py#L925

Here please find an example:

  For example, a typical example from ReCoRD might look like
  {
      'passsage': 'This is the passage.',
      'query': 'A @placeholder is a bird.',
      'entities': ['penguin', 'potato', 'pigeon'],
      'answers': ['penguin', 'pigeon'],
  }
  and I need a prosessor which would turn this example into the following two examples:
  {
      'inputs': 'record query: A @placeholder is a bird. entities: penguin, '
                'potato, pigeon passage: This is the passage.',
      'targets': 'penguin',
  }
  and
  {
      'inputs': 'record query: A @placeholder is a bird. entities: penguin, '
                'potato, pigeon passage: This is the passage.',
      'targets': 'pigeon',
  }

For doing this, one need unbatch, as each entry can map to multiple samples depending on the number of answers, I am not sure how to perform this operation with  huggingface datasets library and greatly appreciate your help

@lhoestq 

Thank you very much.

dorooddorood606:
Hi @lhoestq 
Maybe this is clearer to explain like this, currently map function, map one example to a modified one, lets assume we want to map one example to multiple examples, in which we do not know in advance how many examples they would be for each entry. I greatly appreciate telling me how I can handle this operation, thanks a lot

jackfeinmann5:
Hi,
this is also my question on how to perform similar operation as "unbatch" in tensorflow in great huggingface dataset library. 
thanks.

mariosasko:
Hi,

`Dataset.map` in the batched mode allows you to map a single row to multiple rows. So to perform "unbatch", you can do the following:
```python
import collections

def unbatch(batch):
    new_batch = collections.defaultdict(list)
    keys = batch.keys()
    for values in zip(*batch.values()):
        ex = {k: v for k, v in zip(keys, values)}
[...]

PR #4479, opened 2022-06-12
Include entity positions as feature in ReCoRD

richarddwang:
https://huggingface.co/datasets/super_glue/viewer/record/validation

TLDR: We need to record entity positions, which are included in the source data but excluded by the loading script, to enable efficient and effective training for ReCoRD.

Currently, the loading script ignores the entity positions ("entity_start", "entity_end") and only records entity text. This might be because the training method of the official baseline is to make n training instance from a datapoint by replacing \"\@ placeholder\" in query with each entity individually.

But it increases the already heavy computation by multiple folds. So DeBERTa uses a method that take entity embeddings by their positions in the passage, and thus makes one training instance from one data point. It is way more efficient and proved effective for the ReCoRD task.

Can anybody help me with the dataset card rendering error? Maybe @lhoestq ?


lhoestq (commented):
Thanks @richarddwang ! Sorry for getting back to you after such a long delay.

Anyway it looks good to me ! Can you just remove the remaining jsonl files that are outside of the dummy  data zip files ? They are not needed.

Also can you update the `dataset_infos.json` file with the new column ?
```
datasets-cli test ./datasets/super_glue --name record --save_infos
```

richarddwang:
Thanks for the reply @lhoestq !

I have sucessed on `datasets-cli test ./datasets/super_glue --name record --save_infos`,
But as you can see, the check ran into `FAILED tests/test_dataset_cards.py::test_changed_dataset_card[super_glue] - V...`.
How can we solve it?

lhoestq (commented):
Thanks !

Actually I just noticed that this is a breaking change: the length of the "entities" field changes (current implementation uses a `set` to dedupe them). What about keeping "entities" deduplicated, and having a "entities_spans" field consisting of a list of `{"text": ..., "start": ..., "end": ...}` ? This way it won't break users code, e.g.
- https://github.com/google-research/text-to-text-transfer-transformer/blob/3c58859b8fe72c2dbca6a43bc775aa510ba7e706/t5/data/preprocessors.py#L925
- https://github.com/huggingface/datasets/issues/2767#issuecomment-894678553

What do you think ?

BTW the CI failure is unrelated to this PR - it appears that some tags are missing in the dataset cards ('annotations_creators', 'language_creators', 'license', 'multilinguality', 'size_categories', 'source_datasets', 'task_categories', and 'task_ids'), and so the dataset is not indexed properly on the Hugging Face website. This can be fixed in another PR

richarddwang:
That would be neat! Let me implement it.

lhoestq (approved):
Thank you ! LGTM

Merging since the CI errors are unrelated to this PR

PR #4481, opened 2022-06-13
Fix iwslt2017

lhoestq:
The files were moved to google drive, I hosted them on the Hub instead (ok according to the license)

I also updated the `datasets_infos.json`


lhoestq:
CI fails are just abut missing tags in the dataset card, merging !

nstylia:
FYI, 

The checksums have not been edited from the changes of .tgz to .zip files, and as a result a `ExpectedMoreDownloadedFiles` error occurs. Updating them in the `dataset_infos.json` should fix the error.

lhoestq:
Thanks for reporting and sorry for the delay, I opened https://huggingface.co/datasets/iwslt2017/discussions/2 to fix this

PR #4482, opened 2022-06-13
Test that TensorFlow is not imported on startup

lhoestq:
TF takes some time to be imported, and also uses some GPU memory.

I just added a test to make sure that in the future it's never imported by default when
```python
import datasets
```
is called.

Right now this fails because `huggingface_hub` does import tensorflow

### DIFF
--- a/tests/test_import.py
+++ b/tests/test_import.py
@@ -1,0 +1,10 @@
+import sys
+
+from .utils import require_tf
+
+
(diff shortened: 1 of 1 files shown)


albertvillanova:
Should we close this PR?

albertvillanova:
I'm closing this PR. Feel free to reopen it if necessary.

PR #4492, opened 2022-06-14
Pin the revision in imagenet download links

lhoestq:
Use the commit sha in the data files URLs of the imagenet-1k download script, in case we want to restructure the data files in the future. For example we may split it into many more shards for better paralellism.

cc @mariosasko


mariosasko (approved):
Thanks!

PR #4493, opened 2022-06-14
Add `@transmit_format` in `flatten`

alvarobartt:
As suggested by @mariosasko in https://github.com/huggingface/datasets/pull/4411, we should include the `@transmit_format` decorator to `flatten`, `rename_column`, and `rename_columns` so as to ensure that the value of `_format_columns` in an `ArrowDataset` is properly updated.

**Edit**: according to @mariosasko comment below, the decorator `@transmit_format` doesn't handle column renaming, so it's done manually for those instead.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -1427,6 +1427,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
 
         return dset
 
+    @transmit_format
     @fingerprint_transform(inplace=False)
     def flatten(self, new_fingerprint: Optional[str] = None, max_depth=16) -> "Dataset":
         """Flatten the table.
@@ -1635,6 +1636,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         return dataset
 
     @transmit_tasks
+    @transmit_format
     @fingerprint_transform(inplace=False)
     def rename_column(
         self, original_column_name: str, new_column_name: str, new_fingerprint: Optional[str] = None
@@ -1699,6 +1701,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         return dataset
 
     @transmit_tasks
+    @transmit_format
     @fingerprint_transform(inplace=False)
     def rename_columns(self, column_mapping: Dict[str, str], new_fingerprint: Optional[str] = None) -> "Dataset":
         """
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -1635,8 +1635,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         dataset._fingerprint = new_fingerprint
         return dataset
 
-    @transmit_tasks
-    @transmit_format
+    @transmit_tasks    
     @fingerprint_transform(inplace=False)
     def rename_column(
         self, original_column_name: str, new_column_name: str, new_fingerprint: Optional[str] = None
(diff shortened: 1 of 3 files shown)


alvarobartt:
@mariosasko please let me know whether we need to include some sort of tests to make sure that the decorator is working as expected. Thanks! 🤗

lhoestq (commented):
Thanks ! yes feel free to add a small test :)

src/datasets/arrow_dataset.py
```diff
@@ -1749,6 +1751,7 @@ def rename(columns):
             return [column_mapping[col] if col in column_mapping else col for col in columns]
 
         new_column_names = rename(self._data.column_names)
+        # `transmit_format` doesn't handle column renaming, so update the format columns manually
```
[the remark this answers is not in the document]
alvarobartt: Cool, I'll do that @lhoestq thanks :+1:

tests/test_arrow_dataset.py
```diff
@@ -750,9 +750,11 @@ def test_flatten(self, in_memory):
             ) as dset:
                 with self._to(in_memory, tmp_dir, dset) as dset:
                     fingerprint = dset._fingerprint
+                    dset._format_columns = ["a.b.c", "foo"]
```
lhoestq, line 753: I think it should be "a" - "a.b.c" is not a column yet at this point
```suggestion
                    dset._format_columns = ["a", "foo"]
```

lhoestq (commented):
Can you also add a test to make sure that columns created from an unformatted column are also unformatted ? I'm pretty sure @transmit_format doesn't support this case as in rename_columns though. So we might need to use a similar trick as in rename_columns to make it work (we force the values of _format_columns manually)

tests/test_arrow_dataset.py
alvarobartt: @lhoestq, you're right, that column is created on `dset.flatten()`, I've tested it and it works, so I'll commit your suggestion! :hugs:

tests/test_dataset_dict.py
```diff
@@ -46,13 +46,15 @@ def test_flatten(self):
             {"a": [{"b": {"c": ["text"]}}] * 10, "foo": [1] * 10},
             features=Features({"a": {"b": Sequence({"c": Value("string")})}, "foo": Value("int64")}),
         )
+        dset_split._format_columns = ["a.b.c", "foo"]
```
alvarobartt, line 49: As previously mentioned by @lhoestq, here it should also be `a` instead of `a.b.c` as that column is created in `flatten()`

```suggestion
        dset_split._format_columns = ["a", "foo"]
```

tests/test_arrow_dataset.py
```diff
@@ -750,9 +750,11 @@ def test_flatten(self, in_memory):
             ) as dset:
                 with self._to(in_memory, tmp_dir, dset) as dset:
                     fingerprint = dset._fingerprint
+                    dset._format_columns = ["a", "foo"]
```
lhoestq, line 753: Could you add a test where "a" is not formatted, and make sure "a.b.c" stays unformatted as well ?

I think right now this would fail, so we need to use the same trick as in `rename_columns`

https://github.com/huggingface/datasets/blob/ede72d3f9796339701ec59899c7c31d2427046fb/src/datasets/arrow_dataset.py#L1759-L1760

lhoestq: I think that if you set `dset._format_columns = ["foo"]` and then call `flatten`, then the new column `a.b.c` will end up being formatted (i.e. inside `_format_columns`). However I would expect `a.b.c` to be unformatted, since `a` was not part of the formatted columns before flatten.

We had a similar issue with `rename`: when an unformatted column "a" was renamed to "b", then "b" would have been formatted. To fix this issue we explicitly set `dataset._format_columns` inside `rename`, even though it already has `@transmit_columns`

I hope it's clearer this way, let me know if you have questions or if I can help

alvarobartt:
Oops, I thought this PR was already merged and deleted from the source repository, I'll be creating a new branch out of `main` so as to re-create this PR... My bad :weary:

PR #4497, opened 2022-06-15
Re-add download_manager module in utils

lhoestq:
https://github.com/huggingface/datasets/pull/4384 moved `datasets.utils.download_manager` to `datasets.download.download_manager`

This breaks `evaluate` which imports `DownloadMode` from `datasets.utils.download_manager` 

This PR re-adds `datasets.utils.download_manager` without circular imports.

We could also show a message that says that accessing it is deprecated, but I think we can do this in a subsequent PR, and just focus on doing a patch release for now

### DIFF
--- a/src/datasets/__init__.py
+++ b/src/datasets/__init__.py
@@ -69,12 +69,19 @@ from .splits import (
 )
 from .tasks import *
 from .utils import *
-from .utils import download_manager as _deprecated_download_manager
 from .utils import logging
 
 
 # deprecated modules
+from datasets import utils as _utils  # isort:skip
+from datasets.utils import download_manager as _deprecated_download_manager  # isort:skip
 
 
+_utils.DownloadConfig = DownloadConfig
+_utils.DownloadManager = DownloadManager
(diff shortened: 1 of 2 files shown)


albertvillanova:
Thanks for the fix.

I'm wondering how this fixes backward compatibility...

Executing this code:
```python
from datasets.utils.download_manager import DownloadMode
```
we will have
```python
DownloadMode = None
```

If afterwards we use something like:
```python
if download_mode == DownloadMode.FORCE_REDOWNLOAD
```
that will raise an exception.

lhoestq:
It works fine on my side:
```python
>>> from datasets.utils.download_manager import DownloadMode
>>> DownloadMode is not None
True
```

lhoestq:
As reported in https://github.com/huggingface/evaluate/pull/143
```
from datasets.utils import DownloadConfig
```
is also missing, I'm re-adding it

lhoestq:
Took the liberty of merging this one, to do a patch release soon. If we think of a better approach we can improve it later

PR #4499, opened 2022-06-15
fix ETT m1/m2 test/val dataset

kashif:
https://huggingface.co/datasets/ett/discussions/1


lhoestq:
Thansk for the fix ! Can you regenerate the datasets_infos.json please ? This way it will update the expected number of examples in the test and val splits

kashif:
ah yes!

lhoestq (approved):
Thanks !

PR #4412, opened 2022-05-26
Skip hidden files/directories in data files resolution and `iter_files`

mariosasko:
Fix #4115

PR #4505, opened 2022-06-15
Fix double dots in data files

lhoestq:
As mentioned in https://github.com/huggingface/transformers/pull/17715 `data_files` can't find a file if the path contains double dots `/../`. This has been introduced in https://github.com/huggingface/datasets/pull/4412, by trying to ignore hidden files and directories (i.e. if they start with a dot)

I fixed this and added a test

cc @sgugger @ydshieh


sgugger (approved):
Thanks for the quick fix!

mariosasko (approved):
Thanks!

PR #4509, opened 2022-06-16
Support skipping Parquet to Arrow conversion when using Beam

albertvillanova:


src/datasets/commands/run_beam.py
```diff
@@ -57,6 +58,9 @@ def register_subcommand(parser: ArgumentParser):
             "--ignore_verifications", action="store_true", help="Run the test without checksums and splits checks"
         )
         run_beam_parser.add_argument("--force_redownload", action="store_true", help="Force dataset redownload")
+        run_beam_parser.add_argument(
+            "--skip_parquet_to_arrow", action="store_true", help="Skip Parquet to Arrow conversion"
```
albertvillanova, line 62: Maybe a simpler name?

src/datasets/arrow_writer.py
```diff
@@ -597,20 +597,37 @@ def inc_num_examples(example):
             )
         )
 
-    def finalize(self, metrics_query_result: dict):
+    def finalize(self, metrics_query_result: dict, skip_parquet_to_arrow: bool = False):
```
lhoestq, line 600: I don't think we need to add an argument, the Beam writer could just always write to parquet only. Then the BeamBasedBuilder can be the one calling convert_parquet_to_arrow.
(this would also mean that the Beam writer won't need the cache_dir argument anymore as well, and be aligned with the Arrow writer)

Finally calling convert_parquet_to_arrow can only be done if the cache dir of the builder is a local cache dir. Otherwise if the cache dir is a bucket in a cloud storage there's no need to convert it back to arrow locally.

What do you think ?

albertvillanova: Thanks @lhoestq for your inspiring review.

Indeed, first I was thinking of doing a nicer refactoring (more or less aligned with your suggestion), but at the end I just went the easiest/fastest way to have Wikipedia as soon as possible...

I agree with your first paragraph: I'm proposing a modification in this PR.

On the other hand, I am not completely sure about your second paragraph suggestion: is that a generally expected behavior from our lib? 
- Think about the the natural_questions dataset preprocessing (still in progress): the cache_dir is remote and we want the arrow files
- That was also the case for the preprocessed wikipedia for big languages we did some months ago
I think maybe it would be better to allow the user to switch on/off the conversion, independently of the nature of the cache_dir

What do you think?

lhoestq: Yes right ! Sorry for the confusion :)

lhoestq:
When #4724 is merged, we can just pass `file_format="parquet"` to `download_and_prepare` and it will output parquet fiels without converting to arrow

lhoestq:
I think we can close this one

Issue #4502, opened 2022-06-15
Logic bug in arrow_writer?


cccntu:
https://github.com/huggingface/datasets/blob/88a902d6474fae8d793542d57a4f3b0d187f3c5b/src/datasets/arrow_writer.py#L475-L488

I got some error, and I found it's caused by `batch_examples` being `{}`. I wonder if the code should be as follows:
```
-        if batch_examples and len(next(iter(batch_examples.values()))) == 0:
+       if not batch_examples or len(next(iter(batch_examples.values()))) == 0:
            return
```
@lhoestq

cccntu:
Hi @alvarobartt ,
Thanks for answering. Do you know when and why an empty batch is passed to this function? This only happened to me when processing with multiple workers, while chunking examples, I think.

cccntu:
Thanks, I added a if-print and I found it does return an empty examples in the chunking function that is passed to `.map()`.

alvarobartt:
Glad I could help you out @cccntu :hugs:

lhoestq:
Hi ! We consider an empty batch to look like this:
```python
empty_batch = {
    "column_1": [],
    "column_2": [],
    ...
}
```

While `{}` corresponds to a batch with no columns.

Therefore calling this code should fail, because the two batches don't have the same columns:
```python
writer.write_batch({"a": [1, 2, 3]})
writer.write_batch({})
```

If you want to write an empty batch, you should do this instead:
```python
writer.write_batch({"a": [1, 2, 3]})
writer.write_batch({"a": []})
```

alvarobartt:
Makes sense, then the if-statement should remain the same or is it better to handle both cases separately using `if not batch_examples or len(next(iter(batch_examples.values()))) == 0: ...`?

Updating the regressions tests with an empty batch formatted as `{"col_1": [], "col_2": []}` instead of `{}` works fine with the current if, and also with the one proposed by @cccntu.

alvarobartt:
Great thanks for the response! So I'll just add that regression test and remove the current if-statement.

cccntu:
Hi @lhoestq ,

Thanks for your explanation. Now I get it that `{}` means the columns are different. But wouldn't it be nice if the code can ignore it, like it ignores `{"a": []}`?


--- 
BTW, 
> There's a check later in the code that makes sure that the columns are the right ones, so I don't think we need to check for {} here

I remember the error happens around here:
[...]

PR #4510, opened 2022-06-16
Add regression test for `ArrowWriter.write_batch` when batch is empty

alvarobartt:
As spotted by @cccntu in #4502, there's a logic bug in `ArrowWriter.write_batch` as the if-statement to handle the empty batches as detailed in the docstrings of the function ("Ignores the batch if it appears to be empty, preventing a potential schema update of unknown types."), the current if-statement is not handling properly `writer.write_batch({})` as an error is triggered.

Also, if we add a regression test in `test_arrow_writer.py::test_write_batch` before applying the fix, the test will fail as when trying to write an empty batch as follows:

```
=================================================================================== short test summary info ===================================================================================
FAILED tests/test_arrow_writer.py::test_write_batch[None-None] - ValueError: Schema and number of arrays unequal
FAILED tests/test_arrow_writer.py::test_write_batch[None-1] - ValueError: Schema and number of arrays unequal
FAILED tests/test_arrow_writer.py::test_write_batch[None-10] - ValueError: Schema and number of arrays unequal
FAILED tests/test_arrow_writer.py::test_write_batch[fields1-None] - ValueError: Schema and number of arrays unequal
FAILED tests/test_arrow_writer.py::test_write_batch[fields1-1] - ValueError: Schema and number of arrays unequal
FAILED tests/test_arrow_writer.py::test_write_batch[fields1-10] - ValueError: Schema and number of arrays unequal
FAILED tests/test_arrow_writer.py::test_write_batch[fields2-None] - ValueError: Schema and number of arrays unequal
FAILED tests/test_arrow_writer.py::test_write_batch[fields2-1] - ValueError: Schema and number of arrays unequal
FAILED tests/test_arrow_writer.py::test_write_batch[fields2-10] - ValueError: Schema and number of arrays unequal
======================================================================== 9 failed, 73 deselected, 7 warnings in 0.81s =========================================================================
```

So the batch is not ignored when empty, as `batch_examples={}` won't match the condition `if batch_examples: ...`.

### DIFF
--- a/src/datasets/arrow_writer.py
+++ b/src/datasets/arrow_writer.py
@@ -484,7 +484,7 @@ class ArrowWriter:
         Args:
             batch_examples: the batch of examples to add.
         """
-        if not batch_examples or len(next(iter(batch_examples.values()))) == 0:
+        if batch_examples and len(next(iter(batch_examples.values()))) == 0:
             return
         features = None if self.pa_writer is None and self.update_features else self._features
         try_features = self._features if self.pa_writer is None and self.update_features else None


alvarobartt:
As mentioned by @lhoestq, the current behavior is correct and we should not expect batches with different columns, in that case, the if should fail, as the values of the batch can be empty, but not the actual `batch_examples` value.

lhoestq (approved):
thanks !

Issue #4508, opened 2022-06-15
cast_storage method from datasets.features


romainremyb:
## Describe the bug
A bug occurs when mapping a function to a dataset object. I ran the same code with the same data yesterday and it worked just fine. It works when i run locally on an old version of datasets.

## Steps to reproduce the bug
Steps are:
- load whatever datset
- write a preprocessing function such as "tokenize_and_align_labels" written in https://huggingface.co/docs/transformers/tasks/token_classification
- map the function on dataset and get "ValueError: Class label -100 less than -1" from cast_storage method from datasets.features

# Sample code to reproduce the bug
def tokenize_and_align_labels(examples):
    tokenized_inputs = tokenizer(examples["tokens"], truncation=True, is_split_into_words=True, max_length=38,padding="max_length")

    labels = []
    for i, label in enumerate(examples[f"labels"]):
        word_ids = tokenized_inputs.word_ids(batch_index=i)  # Map tokens to their respective word.
        previous_word_idx = None
        label_ids = []
        for word_idx in word_ids:  # Set the special tokens to -100.
            if word_idx is None:
                label_ids.append(-100)
            elif word_idx != previous_word_idx:  # Only label the first token of a given word.
                label_ids.append(label[word_idx])
            else:
                label_ids.append(-100)
            previous_word_idx = word_idx
        labels.append(label_ids)

    tokenized_inputs["labels"] = labels
    return tokenized_inputs

tokenizer = AutoTokenizer.from_pretrained("distilbert-base-uncased")
dt = dataset.map(tokenize_and_align_labels, batched=True)

## Expected results
New dataset objects should load and do on older versions.

## Actual results
"ValueError: Class label -100 less than -1" from cast_storage method from datasets.features

## Environment info
everything works fine on older installations of datasets/transformers

Issue arises when installing datasets on google collab under python3.7
I can't manage to find the exact output you're requirering but version printed is datasets-2.3.2

mariosasko:
Hi! We've recently added a check to the `ClassLabel` type to ensure the values are in the valid label range `-1, 0, ..., num_classes-1` (-1 is used for missing values). The error in your case happens only if the `labels` column is of type `Sequence(ClassLabel(...))` before the `map` call and can be avoided by calling `dataset = dataset.cast_column("labels", Sequence(Value("int")))` beforehand. The token-classification examples in Transformers introduce a new `labels` column, so their type is also `Sequence(Value("int"))`, which doesn't lead to an error as this type unbounded.

PR #4511, opened 2022-06-16
Support all negative values in ClassLabel

lhoestq:
We usually use -1 to represent a missing label, but we should also support any negative values (some users use -100 for example). This is a regression from `datasets` 2.3

Fix https://github.com/huggingface/datasets/issues/4508

### DIFF
--- a/src/datasets/features/features.py
+++ b/src/datasets/features/features.py
@@ -1024,7 +1024,9 @@ class ClassLabel:
                     f"Class label {min_max['max']} greater than configured num_classes {self.num_classes}"
                 )
         elif isinstance(storage, pa.StringArray):
-            storage = pa.array(self.str2int(storage.to_pylist()))
+            storage = pa.array(
+                [self._strval2int(label) if label is not None else None for label in storage.to_pylist()]
+            )
         return array_cast(storage, self.pa_type)
 
     @staticmethod
--- a/tests/features/test_features.py
+++ b/tests/features/test_features.py
@@ -328,6 +328,11 @@ def test_classlabel_cast_storage():
     arr = pa.array(["__label_that_doesnt_exist__"])
     with pytest.raises(ValueError):
         classlabel.cast_storage(arr)
+    # from nulls
+    arr = pa.array([None])
+    result = classlabel.cast_storage(arr)
+    assert result.type == pa.int64()
+    assert result.to_pylist() == [None]
     # from empty
     arr = pa.array([])
     result = classlabel.cast_storage(arr)


src/datasets/features/features.py
```diff
@@ -1014,16 +1019,12 @@ def cast_storage(self, storage: Union[pa.StringArray, pa.IntegerArray]) -> pa.In
...
-                raise ValueError(f"Class label {min_max['min']} less than -1")
             if min_max["max"] >= self.num_classes:
                 raise ValueError(
                     f"Class label {min_max['max']} greater than configured num_classes {self.num_classes}"
                 )
         elif isinstance(storage, pa.StringArray):
-            storage = pa.array(
-                [self._strval2int(label) if label is not None else None for label in storage.to_pylist()]
-            )
+            storage = pa.array(self.str2int(storage.to_pylist()))
```
lhoestq, line 1027: Should we support `None` for the string representation of missing labels ? I think it could lead to silent errors in users code if we allow this

mariosasko: Yes, I think we should still support this (to be consistent with the behavior during generation - `encode_example` is not called if a label is `None`)

lhoestq: Ok ! `None` values can be used for N/A for example (which has a subtle difference compared to "unknown/missing label" represented by negative values).

Though to avoid silent error I would keep the current behavior of str2int and int2str, that is: raise an error if None is given as input

lhoestq: I reverted this to keep supporting None here :)

davidbenton:
Thanks for this fix! I'm not sure what the release timeline is, but FYI #4508 is a breaking issue for transformer token classification using Trainer and PyTorch. PyTorch defaults to -100 as the ignored label for [log loss](https://pytorch.org/docs/stable/generated/torch.nn.NLLLoss.html?highlight=nllloss#torch.nn.NLLLoss), so switching labels to -1 leads to index errors using Trainer defaults.

As a workaround, I'm using master branch directly (`pip install git+https://github.com/huggingface/datasets.git@master` for anyone who needs to do the same) until this gets released.

lhoestq:
The new release `2.4` fixes the issue, feel free to update `datasets` :) 
```
pip install -U datasets
```

l-uuz:
@lhoestq I hope it's OK to ping you here. I've noticed that `encode_example` does only work with -1. I already created #7645 to fix the documentation, but then I stumbled across your original changes to the docs text in this PR.

I am talking about this part in `ClassLabel -> encode_example`:

https://github.com/huggingface/datasets/blob/e71b0b19d79c7531f9b9bea7c09916b5f6157f42/src/datasets/features/features.py#L1126-L1129

PR #4518, opened 2022-06-16
Patch tests for hfh v0.8.0

LysandreJik:
This PR patches testing utilities that would otherwise fail with hfh v0.8.0.


lhoestq (approved):
thanks !

PR #4519, opened 2022-06-16
Create new sections for audio and vision in guides

stevhliu:
This PR creates separate sections in the guides for audio, vision, text, and general usage so it is easier for users to find loading, processing, or sharing guides specific to the dataset type they're working with. It'll also allow us to scale the docs to additional dataset types - like time series, tabular, etc. - while keeping our docs information architecture. 

Some other changes include:

- Experimented with decorating text with some CSS to highlight guides specific to each modality. Hopefully, it'll be easier for users to find and realize that these different docs exist!
- Added deprecation warning for Metrics and redirect to Evaluate.
- Updated `set_format` section to recommend using the new `to_tf_dataset` function if you need to convert to a TensorFlow dataset.
- Reorganized `toctree` to nest general usage, audio, vision, and text sections under the how-to guides.
- A quick review and edit to the Load and Process docs for clarity.

### DIFF
--- a/docs/source/audio_load.mdx
+++ b/docs/source/audio_load.mdx
@@ -1,0 +1,49 @@
+# Load audio data
+
+Audio datasets typically have an `audio` column which contains three important fields:
+
+* `array`: the decoded audio data represented as a 1-dimensional array.
+* `path`: the path to the downloaded audio file.
+* `sampling_rate`: the sampling rate of the audio data.
+
+When you load an audio dataset and call the `audio` column, the [`Audio`] feature automatically decodes and resamples the audio file:
+
+```py
+>>> from datasets import load_dataset, Audio
+
+>>> dataset = load_dataset("PolyAI/minds14", "en-US", split="train")
+>>> dataset[0]["audio"]
+{'array': array([ 0.        ,  0.00024414, -0.00024414, ..., -0.00024414,
+         0.        ,  0.        ], dtype=float32),
+ 'path': '/root/.cache/huggingface/datasets/downloads/extracted/f14948e0e84be638dd7943ac36518a4cf3324e8b7aa331c5ab11541518e9368c/en-US~JOINT_ACCOUNT/602ba55abb1e6d0fbce92065.wav',
+ 'sampling_rate': 8000}
+```
+
+<Tip warning={true}>
+
+Index into an audio dataset using the row index first and then the `audio` column - `dataset[0]["audio"]` - to avoid decoding and resampling all the audio files in the dataset. Otherwise, if you have a large dataset, this can be a slow and time-consuming process.
+
+</Tip>
+
+## Path
+
+The `path` is useful for loading your own dataset. Use the [`~Dataset.cast_column`] function to take a column of audio file paths, and decode it into `array`'s of the audio data with the [`Audio`] feature:
+
+```py
+>>> audio_dataset = audio_dataset.cast_column("paths_to_my_audio_files", Audio())
+```
+
+If you only want to load the underlying path to the audio dataset without decoding the audio file into an `array`, set `decode=False` in the [`Audio`] feature:
+
+```py
+>>> dataset = load_dataset("PolyAI/minds14", "en-US", split="train").cast_column('audio', Audio(decode=False))
+>>> dataset[0]
+{'audio': {'bytes': None,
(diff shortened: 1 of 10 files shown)


docs/source/loading.mdx
```diff
@@ -1,6 +1,6 @@
 # Load
 
-You have already seen how to load a dataset from the Hugging Face Hub. But datasets are stored in a variety of places, and sometimes you won't find the one you want on the Hub. A dataset can be on disk on your local machine, in a Github repository, and in in-memory data structures like Python dictionaries and Pandas DataFrames. Wherever your dataset may be stored, 🤗 Datasets provides a way for you to load and use it for training.
+Datasets are stored in various places; a dataset can be on your local machine's disk, in a Github repository, and in in-memory data structures like Python dictionaries and Pandas DataFrames. Wherever a dataset is stored, 🤗 Datasets can help you load it.
```
lhoestq, line 3: I'd make the distinction between the users' data and a dataset loaded with `datasets`, to make things clearer here:

```suggestion
Your data can be stored in various places; they can be on your local machine's disk, in a Github repository, and in in-memory data structures like Python dictionaries and Pandas DataFrames. Wherever a dataset is stored, 🤗 Datasets can help you load it.
```

docs/source/audio_load.mdx
```diff
@@ -0,0 +1,49 @@
...
+ 'sampling_rate': 8000}
+```
+
+<Tip warning={true}>
+
+Index into an audio dataset using the row index first and then the `audio` column - `dataset[0]["audio"]` - to avoid decoding and resampling all the audio files in the dataset. Otherwise, this can be a slow and time-consuming process if you have a large dataset.
+
+</Tip>
+
+## Path
```
lhoestq, line 28: I'd name this "Local files" instead

docs/source/audio_load.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+The `path` is useful for loading your own dataset. Use the [`~Dataset.cast_column`] function to take a column of audio file paths, and decode it into `array`'s with the [`Audio`] feature:
+
+```py
+>>> audio_dataset = audio_dataset.cast_column("paths_to_my_audio_files", Audio())
```
lhoestq, line 33: Maybe use something similar to what we have for images:
```python
>>> dataset = Dataset.from_dict({"image": ["path/to/image_1", "path/to/image_2", ..., "path/to/image_n"]}).cast_column("image", Image())
```

lhoestq (commented):
Nice thank you !

We can also add the "Load text data" page, currently it feels weird to not have it ;)

In particular users can do load_dataset("text", data_dir=...) or load_dataset("text", data_files=...) and they can use grep patterns to select several files.

The "text" loaders has a few parameters. The main parameter is `sample_by`. By default `sample_by="line"`, so one example = one line from the text files, but you can change it to "paragraph" or "document"

---------------

A bit unrelated, but I feel like those pages can also be grouped together in a "Dataset repository" section:
- Share
- Create a dataset loading script
- Create a dataset card
- Structure your repository

This way we can decouple "General usage" (how to use datasets) from "Dataset repositories" (how to create a repository)

stevhliu:
Ready for review!

The `toctree` is a bit longer now with the sections. I think if we keep the audio/vision/text/dataset repository sections collapsed by default, and keep the general usage expanded, it may look a little cleaner and not as overwhelming. Let me know what you think! 😄

mariosasko (commented):
Cool! 

Just one nit.

lhoestq (approved):
Thank you ! I think it's ok to leave the new sections uncollapsed though, as you want

docs/source/audio_process.mdx
```diff
@@ -1,93 +1,31 @@
...
-
-```bash
-pip install torchaudio
-```
-
-<Tip warning={true}>
-
-torchaudio's `sox_io` [backend](https://pytorch.org/audio/stable/backend.html#) supports decoding `mp3` files. Unfortunately, the `sox_io` backend is only available on Linux/macOS, and is not supported by Windows.
-
-</Tip>
```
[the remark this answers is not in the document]
stevhliu: I think it'll be better to keep the install instructions in the Installation section. What do you think about maybe adding a tip with a link to the installation section to make sure you have the `audio` dependency (same with `vision`)?

PR #4523, opened 2022-06-17
Update download url and improve card of `cats_vs_dogs` dataset

mariosasko:
Improve the download URL (reported here: https://huggingface.co/datasets/cats_vs_dogs/discussions/1), remove the `image_file_path` column (not used in Transformers, so it should be safe) and add more info to the card.


lhoestq (approved):
Thanks ! I couldn't find any use of the image_file_path field either. I think it's fine to remove it since most image datasets don't have it anyway

PR #4530, opened 2022-06-20
Add AudioFolder packaged loader

### LINKED ISSUE #3964: Add default Audio Loader

polinaeterna:
will close  #3964

AudioFolder is almost identical to ImageFolder except for inferring labels is not the default behavior (`drop_labels` is set to True in config), the option of inferring them is preserved though.

The weird thing is happening with the `test_data_files_with_metadata_and_archives` when `streaming` is `True`. Here is the log from the CI:
```

../.pyenv/versions/3.6.15/lib/python3.6/site-packages/datasets/features/audio.py:237: in _decode_non_mp3_path_like
    array, sampling_rate = librosa.load(f, sr=self.sampling_rate, mono=self.mono)
../.pyenv/versions/3.6.15/lib/python3.6/site-packages/librosa/util/decorators.py:88: in inner_f
    return f(*args, **kwargs)
../.pyenv/versions/3.6.15/lib/python3.6/site-packages/librosa/core/audio.py:176: in load
    raise (exc)
../.pyenv/versions/3.6.15/lib/python3.6/site-packages/librosa/core/audio.py:155: in load
    context = sf.SoundFile(path)
../.pyenv/versions/3.6.15/lib/python3.6/site-packages/soundfile.py:629: in __init__
    self._file = self._open(file, mode_int, closefd)
../.pyenv/versions/3.6.15/lib/python3.6/site-packages/soundfile.py:1184: in _open
    "Error opening {0!r}: ".format(self.name))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

err = 72
prefix = "Error opening <zipfile.ZipExtFile name='audio_file.wav' mode='r' compress_type=deflate>: "

    def _error_check(err, prefix=""):
        """Pretty-print a numerical error code if there is an error."""
        if err != 0:
            err_str = _snd.sf_error_number(err)
>           raise RuntimeError(prefix + _ffi.string(err_str).decode('utf-8', 'replace'))
E           RuntimeError: Error opening <zipfile.ZipExtFile name='audio_file.wav' mode='r' compress_type=deflate>: Error in WAV file. No 'data' chunk marker.
```
I hadn't been able to reproduce this locally until I created the same test environment (I mean with `pip install .[tests]`) with python3.6. The same env but with python3.8 passes the test! I didn't manage to figure out what's wrong, I also tried simply to replace the test wav file and still got the same error. Versions of `soundfile`, `librosa` and `libsndfile` are identical. Might it be something with zip compression? Sounds weird but I don't have any other ideas... 
 
TODO:

- [x] align with #4622
- [x] documentation
- [x] tests for AutoFolder?


polinaeterna:
@lhoestq @mariosasko I don't know what to do with the test, do you have any ideas :)

polinaeterna:
also i't passed in `pyarrow_latest_WIN`

lhoestq:
If the error only happens on 3.6, maybe #4460 can help ^^' It seems to work in 3.7 right ?

> inferring labels is not the default behavior (drop_labels is set to True in config)

I think it a missed opportunity to have a consistent API between imagefolder and audiofolder, since they do everything the same way. Can you give more details why you think we should drop the labels by default ?

mariosasko:
Considering audio classification in audio is not as common as image classification in image, I'm ok with having different config defaults as long as they are properly documented (check [Papers With Code](https://paperswithcode.com/datasets) for stats and compare the classification numbers to the other tasks, do this for both modalities)

Also, WDYT about creating a generic folder loader that ImageFolder and AudioFolder then subclass to avoid having to update both of them when there is something to update/fix?

polinaeterna:
@lhoestq I think it doesn't change the API itself, it's just don't infer labels by default, but you can **still** `drop_labels=False` to `load_dataset` and the labels will be inferred. 
Suppose that one has data structured as follows:
```
data/
   train/
      audio/
          file1.wav
          file2.wav
          file3.wav
      metadata.jsonl
   test/
      audio/
          file1.wav
          file2.wav
          file3.wav
      metadata.jsonl
```
If users load this dataset with `load_dataset("audiofolder", data_dir="data")` (the most native way), they will get a `label` feature that will always be equal to 0 (= "audio"). To mitigate this, they will have to always specify `load_dataset("audiofolder", data_dir="data", drop_labels=True)` explicitly and I believe it's not convenient. 

At the same time, `label` column can be added just as easy as adding one argument:` load_dataset("audiofolder", data_dir="data", drop_labels=False)`. As classification task is not as common, I think it should require more symbols to be added to the code :D  

But this is definitely should be explained in the docs, which I've forgotten to update... I'll add this section soon.

Also +to the generic loader, will work on it.

lhoestq:
If a metadata.jsonl file is present, then it doesn't have to infer the labels I agree. Note that this is already the case for imagefolder ;) in your case `load_dataset("audiofolder", data_dir="data")` won't return labels !

Labels are only inferred if there are no metadata.jsonl

lhoestq:
Feel free to merge the `main` branch into yours after updating your fork of `datasets`: https://github.com/huggingface/datasets/issues/4629

This should fix some errors in the CI

src/datasets/streaming.py
```diff
@@ -112,3 +112,9 @@ def extend_dataset_builder_for_streaming(builder: "DatasetBuilder"):
                 internal_import_name = imports[1]
                 internal_module_name = ".".join(builder.__module__.split(".")[:-1] + [internal_import_name])
                 extend_module_for_streaming(internal_module_name, use_auth_token=builder.use_auth_token)
+
+    # we need to patch autofolder.py manually because it's a parent of Image and Audio folder and it
+    # contains generation of examples (which uses `os` functions)
+    autofolder_module = [cls.__module__ for cls in type(builder).__mro__ if cls.__name__ == "AutoFolder"]
```
lhoestq, line 118: Just in case someone subclasses a class named "AutoFolder" that is not the one in `datasets`:
```suggestion
    # contains generation of examples (which uses `os` functions)
    from .packaged_modules.base.autofolder import AutoFolder

    autofolder_module = [cls.__module__ for cls in type(builder).__mro__ if cls is AutoFolder
```

polinaeterna:
@mariosasko  could you please review this PR again? :)

most of the tests for AutoFolder (base class for AudioFolder and ImageFolder) are now basically copied from Image/AudioFolder (their tests are also almost identical too) and adapted to test other methods. it should be refactored but i think this is not that important for now and might be done in the future PR, wdyt?

src/datasets/packaged_modules/base/autofolder.py
```diff
@@ -0,0 +1,398 @@
...
+                    check the config's `base_feature`.
+                    """
+                )
+
+
+class AutoFolder(datasets.GeneratorBasedBuilder):
+    BUILDER_CONFIG_CLASS = AutoFolderConfig
+    EXTENSIONS: List[str] = []
+    SKIP_CHECKSUM_COMPUTATION_BY_DEFAULT = True
+    METADATA_FILENAME: str = "metadata.jsonl"
```
mariosasko, line 76: Feel free to separate the abstract attributes from the non-abstract ones (and add some comments). Moreover, in Python 3.6+, you can do `EXTENSIONS: List[str]` without specifying the default value to make it more obvious this is an abstract attribute.

Additionally, I think this class can be abstract (`class AutoFolder(datasets.GeneratorBasedBuilder, abc.ABC)`) to make sure calling `load_dataset`/`load_dataset_builder` fails on it.

P.S. I think `folder_builder`/`FolderBuilder` would be a better name for this module/class.

src/datasets/packaged_modules/imagefolder/imagefolder.py
```diff
@@ -1,373 +1,50 @@
...
-                    },
-                )
+        # _prepare_split_generators() sets self.info.features,
+        # infers labels, finds metadata files if needed and returns splits
+        splits = self._prepare_split_generators(dl_manager)
+
+        add_labels = splits[0].gen_kwargs["add_labels"]
+        if self.config.features is None and add_labels:
+            task_template = ImageClassification(
+                image_column=self.config.base_feature_name, label_column=self.config.label_column
```
mariosasko, line 37: Feel free to remove this stuff (the task template API is deprecated, although not officially). I guess this is why you defined the `_prepare_` methods in the first place.

src/datasets/streaming.py
```diff
@@ -112,3 +111,12 @@ def extend_dataset_builder_for_streaming(builder: "DatasetBuilder"):
...
                 extend_module_for_streaming(internal_module_name, use_auth_token=builder.use_auth_token)
+
+    # we need to patch autofolder.py manually because it's a parent of ImageFolder and AudioFolder,`
+    # and it contains generation of examples for them (which uses `os` functions)
+    # but it's not patched automatically because it's not instantiated
+    from .packaged_modules.base.autofolder import AutoFolder
+
+    autofolder_module = [cls.__module__ for cls in type(builder).__mro__ if cls is AutoFolder]
+    if autofolder_module:
+        extend_module_for_streaming(autofolder_module[0], use_auth_token=builder.use_auth_token)
```
mariosasko, line 122: IMO it would be cleaner to remove this and allow patching internal imports in the packaged modules as long as these imports don't refer to something that's outside the `packaged_modules` subpackage.

This can be done by removing [this](https://github.com/huggingface/datasets/blob/18287ed524ac79f538abf77183f906e91d0de3de/src/datasets/streaming.py#L109) check and adding a test for internal imports to [this](https://github.com/huggingface/datasets/blob/main/tests/test_dataset_scripts.py) file

tests/packaged_modules/test_autofolder.py
```diff
@@ -0,0 +1,439 @@
+import importlib
```
mariosasko, line 1: It would be great if we could avoid test duplication, so perhaps we can have this file for the main "folder loader" tests and `test_imagefolder` and `test_audiofolder` for the ones specific to the loader (but it's ok to address this in another PR)

mariosasko (commented):
Sounds good! 

These are my comments:

polinaeterna:
@mariosasko thank you for the review! I'm sorry I accidentally asked for the review again, ignore it.

src/datasets/streaming.py
polinaeterna: that's what I tried to do in the first place but it appeared that [`get_import`](https://github.com/huggingface/datasets/blob/main/src/datasets/utils/py_utils.py#L514) do not work with imports starting with `..` and @lhoestq and I decided that it would be easier to hardcode it with a comment instead of modifying this function and everything that uses it. 

I tried to modify `get_import` in a non-breaking way here [bc1fb3d](https://github.com/huggingface/datasets/pull/4530/commits/bc1fb3d6937b194011a28b15df272801cc10187c), but I don't like it that way. Also, it breaks tests on windows CI and I can't figure out why. so maybe we can leave it hardcoded? or do you have any better ideas?

src/datasets/packaged_modules/base/autofolder.py
polinaeterna: `load_dataset` and `load_dataset_builder` won't work anyway because "autofolder" is not included in the `packaged_modules/__init__.py`. if it would, it wouldn't work with `config._base_feature` equals to `None`.

but I agree adding ABC makes the code more readable, but I'm not sure how to do it. just adding it as a base class doesn't make `inspect.isabstract()` return `True` as all abstract methods are implemented.

src/datasets/packaged_modules/base/autofolder.py
```diff
@@ -0,0 +1,398 @@
...
+                self.base_feature_name = self._base_feature._type.lower()
+            elif hasattr(self._base_feature, "__name__"):
+                self.base_feature_name = self._base_feature.__name__.lower()
+            else:
+                raise AttributeError(
+                    """
+                    `base_feature_name` is not provided and cannot be inferred form the base feature,
+                    check the config's `base_feature`.
+                    """
+                )
```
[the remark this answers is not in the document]
polinaeterna: it wouldn't work because neither `Image()` nor `Audio()` doesn't have `__name__` attribute. `__class__.__name__` would work but that way it works for anything (including `None`) which is probably not what we want.

src/datasets/packaged_modules/audiofolder/audiofolder.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+logger = datasets.utils.logging.get_logger(__name__)
+
+
+@dataclass
+class AudioFolderConfig(autofolder.AutoFolderConfig):
+    """Builder Config for AudioFolder."""
+
+    _base_feature: ClassVar = datasets.Audio()
```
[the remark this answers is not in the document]
polinaeterna: I think as `_base_feature` is marked as `ClassVar` in a dataclass it's not passed to it's `__init__`, and passing it to `load_dataset` would raise TypeError (unexpected keyword argument). 

but anyway I change it to builder.BASE_FEATURE with a corresponding comment. that way it doesn't need to be marked with a leading underscore to be excluded from documentation generation.

src/datasets/streaming.py
polinaeterna: ah tests shouldn't work this way either because in tests it would be imports from datasets library, not internal. so I would suggest to revert to hardcoded patching. cc @lhoestq

src/datasets/packaged_modules/base/autofolder.py
mariosasko: > just adding it as a base class doesn't make inspect.isabstract() return True as all abstract methods are implemented.

I wasn't aware of that. 

Not marking this one as resolved as I'd like to hear @lhoestq's opinion on the naming (but I'm also ok with AutoFolder)

polinaeterna: @mariosasko you mean the name of the folder? i'll change it now, i just incorrectly understand your comment first, `FolderBuilder` looks better for me too

src/datasets/packaged_modules/base/autofolder.py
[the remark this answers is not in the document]
polinaeterna: i'm not sure i understand correctly, you mean add `BASE_COLUMN_NAME` to the `FolderBuilder` as an abstract attribute and then set their values in `ImageFolder` and `AudioFolder`?

src/datasets/streaming.py
polinaeterna: @mariosasko are you ok with this solution? https://github.com/huggingface/datasets/pull/4530/commits/bfecab46c0aa2adea8adb014204ad8f26fd4c4bc

src/datasets/packaged_modules/base/autofolder.py
polinaeterna: [fe80766](https://github.com/huggingface/datasets/pull/4530/commits/fe80766ff7e58ff63d0a953c43536584a3b0ae78)

src/datasets/packaged_modules/base/autofolder.py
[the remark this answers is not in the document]
polinaeterna: [fe80766](https://github.com/huggingface/datasets/pull/4530/commits/fe80766ff7e58ff63d0a953c43536584a3b0ae78)

src/datasets/packaged_modules/base/autofolder.py
lhoestq: "Auto" meant that something was happening automatically, which is not what the classe it doing here, it's just an abstract/base class. `FolderBuilder` is maybe not ideal given that it doesn't build folders... ^^'

I find `FolderBasedBuilder` nice, and is consistent with the naming of the other abstract/base builders that we have: `GeneratorBasedBuilder`, `ArrowBasedBuilder`, `BeamBasedBuilder`. What do you think ?

docs/source/audio_load.mdx
```diff
@@ -55,3 +55,126 @@ If you only want to load the underlying path to the audio dataset without decodi
...
+}
+```
+
+Load remote datasets from their URLs with the `data_files` parameter:
+
+```py
+>>> dataset = load_dataset("audiofolder", data_files="https://s3.amazonaws.com/datasets.huggingface.co/SpeechCommands/v0.01/v0.01_test.tar.gz")
+```
+
+## AudioFolder with metadata
```
lhoestq, line 90: I think I would put this section first, since this is the main use case anyway.

docs/source/audio_load.mdx
```diff
@@ -55,3 +55,126 @@ If you only want to load the underlying path to the audio dataset without decodi
  'transcription': 'I would like to set up a joint account with my partner'}
 ```
 
+## AudioFolder
+
+You can also load a dataset with an `AudioFolder` dataset builder. It does not require writing a custom dataloader, making it useful for quickly loading audio data. Your dataset structure might look like:
```
lhoestq, line 60: I think you can remove this part, since it's already presented later in "AudioFolder with labels"

lhoestq (commented):
Thanks @polinaeterna and @mariosasko this is super cool ! IMO once the class name is settled we're good to go. We can re-organize the tests in a subsequent PR I think :)

I also added a comment about the default for `drop_labels` again. I think that if the documentation focuses first on  audio+metadata, it's ok to have the same default as ImageFolder

src/datasets/packaged_modules/audiofolder/audiofolder.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+logger = datasets.utils.logging.get_logger(__name__)
+
+
+@dataclass
+class AudioFolderConfig(folder_builder.FolderBuilderConfig):
+    """Builder Config for AudioFolder."""
+
+    drop_labels: bool = True  # usually we don't need labels as classification is not the main audio task
```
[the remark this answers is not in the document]
polinaeterna: well I don't have a strong opinion here anymore :D 
if we set `drop_labels=None` by default as you suggested, it might be confusing in cases when users provide only audio files, without metadata (or with broken metadata?). this is probably quite unlikely, so I'm ok with your suggestion.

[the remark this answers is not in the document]
polinaeterna: @mariosasko what do you think about that?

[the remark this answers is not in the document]
mariosasko: It's good to be consistent, so I agree with @lhoestq.

src/datasets/packaged_modules/base/autofolder.py
polinaeterna: @lhoestq I like `FolderBasedBuilder`. do you think we can preserve the name of directory and the script as `folder_builder/folder_builder.py` just to be shorter or is it not consistent and `folder_based_builder/folder_based_builder.py` is better?

src/datasets/streaming.py
mariosasko: I guess it's ok to hardcode this logic considering relative imports outside a builder package are only possible here, so a more generic approach is not needed at the moment.

src/datasets/streaming.py
```diff
@@ -112,3 +111,18 @@ def extend_dataset_builder_for_streaming(builder: "DatasetBuilder"):
...
+    # but these parents builders are not patched automatically as they are not instantiated, so we patch them here
+    from .builder import DatasetBuilder
+
+    parent_builder_modules = [
+        cls.__module__
+        for cls in type(builder).__mro__
+        if issubclass(cls, DatasetBuilder)
+        # check it's not a standard builder from datasets.builder and not the same module we've already patched
+        and cls.__module__ not in ["datasets.builder", builder.__module__]
+    ]
```
mariosasko, line 126: One nit to make this part cleaner:

```suggestion
    parent_builder_modules = [
        cls.__module__
        for cls in type(builder).__mro__[1:]
        if issubclass(cls, DatasetBuilder)
        and cls is not DatasetBuilder
    ]
```

src/datasets/packaged_modules/audiofolder/audiofolder.py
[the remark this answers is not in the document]
polinaeterna: i'm ok with that but it makes explaining things in documentation a bit more complicated...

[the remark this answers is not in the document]
polinaeterna: i've updated the docs, tried to make it simpler. still not sure that this logic with default None value of "drop_labels" is clear but I guess we'd better see what users say.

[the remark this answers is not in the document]
polinaeterna: @lhoestq @mariosasko what do you think about it now? 🤗 
also, don't you know what's happening with the CI? why it takes forever and finally some jobs are cancelled?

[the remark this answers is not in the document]
lhoestq: Looks all good to me ! :D 

Not sure what's happening with the CI though. I just re-launched one job to see if it was caused by a bug in github actions or the windows runners

src/datasets/streaming.py
```diff
@@ -112,3 +111,16 @@ def extend_dataset_builder_for_streaming(builder: "DatasetBuilder"):
...
+
+    # builders can inherit from other builders that might use streaming functionality
+    # (for example, ImageFolder and AudioFolder inherit from FolderBuilder which implements examples generation)
+    # but these parents builders are not patched automatically as they are not instantiated, so we patch them here
+    from .builder import DatasetBuilder
+
+    parent_builder_modules = [
+        cls.__module__
+        for cls in type(builder).__mro__[1:]
+        if issubclass(cls, DatasetBuilder) and cls is not DatasetBuilder
```
lhoestq, line 123: The windows CI fails when running the CSV builder, which is a GeneratorBasedBuilder. I took a look at the only changes that could affect a GeneratorBasedBuilder and I think this line caused the CI to hang on windows by patching `builder.py` for streaming and failing.

Maybe we can avoid patching `builder.py` for now ?
```suggestion
        if issubclass(cls, DatasetBuilder) and cls.__module__ != DatasetBuilder.__module__]
```

polinaeterna:
@mariosasko can you please take a look one more time? :) i hope we are ready to merge 

the last fix in streaming.py was not to patch builder.py, it caused lock on the Windows CI

lhoestq (approved):
LGTM !

PR #4532, opened 2022-06-20
Add Video feature

nateraw:
The following adds a `Video` feature for encoding/decoding videos on the fly from in memory bytes. It uses my own `encoded-video` library which is basically `pytorchvideo`'s encoded video but with all the `torch` specific stuff stripped out. Because of that, and because the tool I used under the hood is not very mature, I leave this as a draft idea that we can use to build off of.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -100,6 +100,7 @@ AUDIO_REQUIRE = [
 
 VISION_REQURE = [
     "Pillow>=6.2.1",
+    "encoded_video"
 ]
 
 BENCHMARKS_REQUIRE = [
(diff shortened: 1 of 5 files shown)


fcakyon:
@nateraw do you have any plans to continue this pr? Or should I write a custom loader script to use my video dataset from the hub?

nateraw:
@fcakyon I think we still want this feature in here, but my solution here isn't the right one, I'm afraid. Using my (very hacky) library is not the right move. Let's move to an issue to discuss the feature/workarounds for now.

PR #4535, opened 2022-06-21
Add `batch_size` parameter when calling `add_faiss_index` and `add_faiss_index_from_external_arrays`

alvarobartt:
Currently, even though the `batch_size` when adding vectors to the FAISS index can be tweaked in `FaissIndex.add_vectors()`, the function `ArrowDataset.add_faiss_index` doesn't have either the parameter `batch_size` to be propagated to the nested `FaissIndex.add_vectors` function or `*args, **kwargs`, so on, this PR adds the `batch_size` parameter to both `ArrowDataset.add_faiss_index` and `ArrowDataset.add_faiss_index_from_external_arrays`.

This is useful so as to tweak the `batch_size` according to the VM specifications.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -1119,7 +1119,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         # Get json serializable dataset info
         dataset_info = asdict(dataset._info)
 
-        # Save dataset + indices + state + info
+        # Save dataset + state + info
         fs.makedirs(dataset_path, exist_ok=True)
         with fs.open(Path(dataset_path, config.DATASET_ARROW_FILENAME).as_posix(), "wb") as dataset_file:
             with ArrowWriter(stream=dataset_file) as writer:
--- a/src/datasets/search.py
+++ b/src/datasets/search.py
@@ -447,6 +447,7 @@ class IndexableMixin:
         string_factory: Optional[str] = None,
         metric_type: Optional[int] = None,
         custom_index: Optional["faiss.Index"] = None,
+        batch_size: Optional[int] = 1000,
         train_size: Optional[int] = None,
         faiss_verbose: bool = False,
     ):
@@ -465,6 +466,7 @@ class IndexableMixin:
             string_factory (Optional :obj:`str`): This is passed to the index factory of Faiss to create the index. Default index class is IndexFlatIP.
             metric_type (Optional :obj:`int`): Type of metric. Ex: `faiss.METRIC_INNER_PRODUCT` or `faiss.METRIC_L2`.
             custom_index (Optional :obj:`faiss.Index`): Custom Faiss index that you already have instantiated and configured for your needs.
+            batch_size (Optional :obj:`int`): Size of the batch to use while adding vectors to the FaissIndex. Default value is 1000.
             train_size (Optional :obj:`int`): If the index needs a training step, specifies how many vectors will be used to train the index.
             faiss_verbose (:obj:`bool`, defaults to False): Enable the verbosity of the Faiss index.
         """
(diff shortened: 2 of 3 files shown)


alvarobartt:
Also, I had a doubt while checking the code related to the indices... 

@lhoestq, there's a value in `config.py` named `DATASET_INDICES_FILENAME` which has the arrow extension (which I assume it should be `indices.faiss`, as the Elastic Search indices are not stored in a file, but not sure), and it's just used before actually saving an `ArrowDataset` in disk, but since those indices are never stored AFAIK, is that actually required?

https://github.com/alvarobartt/datasets/blob/aec86ea4b790ccccc9b2e0376a496728b1c914cc/src/datasets/config.py#L183

https://github.com/alvarobartt/datasets/blob/aec86ea4b790ccccc9b2e0376a496728b1c914cc/src/datasets/arrow_dataset.py#L1079-L1092

So should I also remove that?

P.S. I also edited the following code comment which I found misleading as it's not actually storing the indices.

https://github.com/huggingface/datasets/blob/8ddc4bbeb1e2bd307b21f5d21f884649aa2bf640/src/datasets/arrow_dataset.py#L1122

lhoestq:
> @lhoestq, there's a value in config.py named DATASET_INDICES_FILENAME which has the arrow extension (which I assume it should be indices.faiss, as the Elastic Search indices are not stored in a file, but not sure), and it's just used before actually saving an ArrowDataset in disk, but since those indices are never stored AFAIK, is that actually required?

The arrow file is used to store an indices mapping (when you shuffle the dataset for example) - not for a faiss index ;)

src/datasets/search.py
```diff
@@ -447,6 +447,7 @@ def add_faiss_index(
         string_factory: Optional[str] = None,
         metric_type: Optional[int] = None,
         custom_index: Optional["faiss.Index"] = None,
+        batch_size: Optional[int] = 1000,
```
lhoestq, line 450: I think you need to make the same changes in arrow_dataset.py, where add_faiss_index and add_faiss_index_from_external_arrays are overriden ;)

lhoestq (commented):
Cool thank you ! there is one more change to do in arrow_dataset.py for it to work and we'll be good:

alvarobartt:
Ok cool thanks a lot for the explanation @lhoestq I was not sure about that :+1: I'll also add it there as you suggested!

lhoestq (approved):
Thank you !

src/datasets/search.py
```diff
@@ -483,6 +487,7 @@ def add_faiss_index_from_external_arrays(
...
```
lhoestq, line 490:
```suggestion
        batch_size: int = 1000,
```

src/datasets/search.py
lhoestq, line 450:
```suggestion
        batch_size: int = 1000,
```

src/datasets/arrow_dataset.py
```diff
@@ -4476,6 +4479,7 @@ def add_faiss_index_from_external_arrays(
         string_factory: Optional[str] = None,
         metric_type: Optional[int] = None,
         custom_index: Optional["faiss.Index"] = None,  # noqa: F821
+        batch_size: Optional[int] = 1000,
```
lhoestq, line 4482:
```suggestion
        batch_size: int = 1000,
```

src/datasets/arrow_dataset.py
```diff
@@ -4403,6 +4403,7 @@ def add_faiss_index(
...
```
lhoestq, line 4406:
```suggestion
        batch_size: int = 1000,
```

lhoestq (commented):
actually just one change in the type hint and we can merge:

lhoestq:
CI failures are unrelated to this PR and fixed on master, merging

PR #4536, opened 2022-06-21
Properly raise FileNotFound even if the dataset is private

lhoestq:
`tests/test_load.py::test_load_streaming_private_dataset` was failing because the hub now returns 401 when getting the HfApi.dataset_info of a dataset without authentication. `load_dataset` was raising ConnectionError, while it should be FileNoteFoundError since it first checks for local files before checking the Hub.

Moreover when use_auth_token is not set (default is False), we should not pass `token=None` to HfApi.dataset_info, or it will use the local token by default - instead it should use no token. It's currently not possible to ask for no token to be used, so as a workaround I simply set token="no-token"

### DIFF
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -783,7 +783,7 @@ class HubDatasetModuleFactoryWithoutScript(_DatasetModuleFactory):
         hfh_dataset_info = HfApi(config.HF_ENDPOINT).dataset_info(
             self.name,
             revision=self.revision,
-            token=token,
+            token=token if token else "no-token",
             timeout=100.0,
         )
         patterns = (
@@ -1140,7 +1140,7 @@ def dataset_module_factory(
                     dataset_info = hf_api.dataset_info(
                         repo_id=path,
                         revision=revision,
-                        token=token,
+                        token=token if token else "no-token",
                         timeout=100.0,
                     )
                 except Exception as e:  # noqa: catch any exception of hf_hub and consider that the dataset doesn't exist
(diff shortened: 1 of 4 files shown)


src/datasets/load.py
```diff
@@ -783,7 +783,7 @@ def get_module(self) -> DatasetModule:
         hfh_dataset_info = HfApi(config.HF_ENDPOINT).dataset_info(
             self.name,
             revision=self.revision,
-            token=token,
+            token=token if token else "no-token",
```
julien-c, line 786: would passing `token=False` work instead?

julien-c: also don't hesitate to ping @SBrandeis / @coyotte508 / @Pierrci on those kind of PRs =)

SBrandeis: `"no-token"` won't trigger the type checker so I think it's better

albertvillanova: If HFH maintainers prefer `"no-token"`, then this is OK! ;)

src/datasets/load.py
```diff
@@ -1159,7 +1159,10 @@ def dataset_module_factory(
                     elif "401" in str(e):
                         msg = f"Dataset '{path}' doesn't exist on the Hub"
                         msg = msg + f" at revision '{revision}'" if revision else msg
-                        raise FileNotFoundError(msg + ". If the repo is private, make sure you are authenticated.")
+                        raise FileNotFoundError(
+                            msg
+                            + ". If the repo is private, make sure you are authenticated with ``use_auth_token=True`` after logging in with ``huggingface-cli login```."
```
albertvillanova, line 1164: Two nits:
- I would not use double backticks here: this is written to the shell (no Markdown/reStructuredText rendering is going here)
- Anyway, there is a wrong triple backtick at the end.

albertvillanova (approved):
Thank you. Good catch! ;)

Issue #4533, opened 2022-06-20
Timestamp not returned as datetime objects in streaming mode


lhoestq:
As reported in (internal) https://github.com/huggingface/datasets-server/issues/397

```python
>>> from datasets import load_dataset
>>> dataset = load_dataset("ett", name="h2", split="test", streaming=True)
>>> d = next(iter(dataset))
>>> d['start']
Timestamp('2016-07-01 00:00:00')
```

while loading in non-streaming mode it returns `datetime.datetime(2016, 7, 1, 0, 0)`

PR #4541, opened 2022-06-22
Fix timestamp conversion from Pandas to Python datetime in streaming mode

lhoestq:
Arrow accepts both pd.Timestamp and datetime.datetime objects to create timestamp arrays.
However a timestamp array is always converted to datetime.datetime objects.

This created an inconsistency between streaming in non-streaming. e.g. the `ett` dataset outputs datetime.datetime objects in non-streaming but pd.timestamp in streaming.

I fixed this by always converting pd.Timestamp to datetime.datetime during the example encoding step.
I fixed the same issue for pd.Timedelta as well. Finally I added an extra step of conversion for Series and DataFrame to take this into account in case such data are passed as Series or DataFrame.

Fix https://github.com/huggingface/datasets/issues/4533
Related to https://github.com/huggingface/datasets-server/issues/397


lhoestq:
CI failures are unrelated to this PR, merging

PR #4543, opened 2022-06-22
[CI] Fix upstream hub test url

lhoestq:
Some tests were still using moon-stagign instead of hub-ci.

I also updated the token to use one dedicated to `datasets`


lhoestq:
Remaining CI failures are unrelated to this fix, merging

Issue #4115, opened 2022-04-06
ImageFolder add option to ignore some folders like '.ipynb_checkpoints'


cceyda:
**Is your feature request related to a problem? Please describe.**
I sometimes like to peek at the dataset images from jupyterlab. thus '.ipynb_checkpoints' folder appears where my dataset is and (just realized) leads to accidental duplicate image additions. I think this is an easy enough thing to miss especially if the dataset is very large.

**Describe the solution you'd like**
maybe have an option `ignore` or something .gitignore style
`dataset = load_dataset("imagefolder", data_dir="./data/original", ignore="regex?")`

**Describe alternatives you've considered**
Could filter out manually

nateraw:
Maybe it would be nice to ignore private dirs like this one (ones starting with `.`) by default. 

CC @mariosasko

mariosasko:
Maybe we can add a `ignore_hidden_files` flag to the builder configs of our packaged loaders (to be consistent across all of them), wdyt @lhoestq @albertvillanova?

lhoestq:
I think they should always ignore them actually ! Not sure if adding a flag would be helpful

lhoestq:
> @lhoestq But what if the user explicitly requests those files via regex?

Usually hidden files are meant to be ignored. If they are data files, they must be placed outside a hidden directory in the first place right ? I think it's more sensible to explain this than adding a flag.

> glob.glob ignores hidden files (files starting with ".") by default unless they are explicitly requested, but fsspec's glob doesn't follow this behavior, which is probably a bug, so maybe we can raise an issue or open a PR in their repo?

After globbing using `fsspec`, we already ignore files that start with a `.` in `_resolve_single_pattern_locally` and `_resolve_single_pattern_in_dataset_repository`, I guess we can just account for parent directories as well ?

We could open an issue on `fsspec` but I think they won't change this since it's an important breaking change for them.

PR #4551, opened 2022-06-23
Perform hidden file check on relative data file path

### LINKED ISSUE #4549: FileNotFoundError when passing a data_file inside a directory starting with double underscores
Bug experienced in the `accelerate` CI: https://github.com/huggingface/accelerate/runs/7016055148?check_suite_focus=true

This is related to https://github.com/huggingface/datasets/pull/4505 and the changes from https://github.com/huggingface/datasets/pull/4412

mariosasko:
Fix #4549


lhoestq (approved):
Thanks ! LGTM :)

Though it doesn't fix the issue if the relative path points to an actual file (not a pattern) and if it contains a double underscores (so it doesn't completely "Fix https://github.com/huggingface/datasets/issues/4549"):

```python
resolve_patterns_locally_or_by_urls(
    base_path="/Users/quentinlhoest/Desktop/hf/datasets/playground",
    patterns=[".yo/data.txt"]
)
```

^ this one should return the file

but we still have to make sure this one doesn't:
```python
resolve_patterns_locally_or_by_urls(
    base_path="/Users/quentinlhoest/Desktop/hf/datasets/playground",
    patterns=["*/data.txt"]
)
```

--------------

(and probably out of scope for now - not important)

maybe this one should return the file
```python
resolve_patterns_locally_or_by_urls(
    base_path="/Users/quentinlhoest/Desktop/hf/datasets/playground",
    patterns=[".yo/*"]
)
```

mariosasko:
I'm aware of this behavior, which is tricky to solve due to fsspec's hidden file handling (see https://github.com/huggingface/datasets/issues/4115#issuecomment-1108819538). I've tested some regex patterns to address this, and they seem to work (will push them on Monday; btw they don't break any of fsspec's tests, so maybe we can contribute this enhancement to them). Also, perhaps we should include the files starting with `__`  in the results again (we hadn't had issues with this pattern before). WDYT?

lhoestq:
I see. Feel free to merge this one if it's good for you btw :)

> Also, perhaps we should include the files starting with __ in the results again (we hadn't had issues with this pattern before)

The point was mainly to ignore __pycache__ directories for example. Also also for consistency with the iter_files/iter_archive which are already ignoring them

mariosasko:
Very elegant solution! Feel free to merge if the CI is green after adding the tests.

lhoestq:
CI failure is unrelated to this PR

PR #4553, opened 2022-06-23
Stop dropping columns in to_tf_dataset() before we load batches

Rocketknight1:
`to_tf_dataset()` dropped unnecessary columns before loading batches from the dataset, but this is causing problems when using a transform, because the dropped columns might be needed to compute the transform. Since there's no real way to check which columns the transform might need, we skip dropping columns and instead drop keys from the batch after we load it.

cc @amyeroberts and https://github.com/huggingface/notebooks/pull/202

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -265,9 +265,11 @@ class TensorflowDatasetMixin:
             indices = sample(range(len(dataset)), test_batch_size)
             test_batch = dataset[indices]
             if cols_to_retain is not None:
-                test_batch = {key: value for key, value in test_batch.items()
-                              if key in cols_to_retain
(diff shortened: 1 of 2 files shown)


Rocketknight1:
@lhoestq Rebasing fixed the test failures, so this should be ready to review now! There's still a failure on Win but it seems unrelated.

Rocketknight1:
Gentle ping @lhoestq ! This is a simple fix (dropping columns after loading a batch from the dataset rather than with `.remove_columns()` to make sure we don't break transforms), and tests are green so we're ready for review!

lhoestq (commented):
Sounds good ! Can you write a test please ?

Rocketknight1:
@lhoestq Test is in!

lhoestq (approved):
thank you :)

Issue #4354, opened 2022-05-15
Problems with WMT dataset


eldarkurtic:
## Describe the bug
I am trying to load WMT15 dataset and to define which data-sources to use for train/validation/test splits, but unfortunately it seems that the official documentation at [https://huggingface.co/datasets/wmt15#:~:text=Versions%20exists%20for,wmt_translate%22%2C%20config%3Dconfig)](https://huggingface.co/datasets/wmt15#:~:text=Versions%20exists%20for,wmt_translate%22%2C%20config%3Dconfig)) doesn't work anymore.

## Steps to reproduce the bug
```shell
>>> import datasets
>>> a = datasets.translate.wmt.WmtConfig()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'datasets' has no attribute 'translate'
>>> a = datasets.wmt.WmtConfig()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'datasets' has no attribute 'wmt'
```

## Expected results
To load WMT15 with given data-sources.

## Environment info

- `datasets` version: 2.0.0
- Platform: Linux-5.10.0-10-amd64-x86_64-with-glibc2.17
- Python version: 3.8.12
- PyArrow version: 7.0.0
- Pandas version: 1.4.1

mariosasko:
Hi! Yes, the docs are outdated. Expect this to be fixed soon. 

In the meantime, you can try to fix the issue yourself.

These are the configs/language pairs supported by `wmt15` from which you can choose:
* `cs-en` (Czech - English)
* `de-en` (German - English)
* `fi-en` (Finnish- English)
* `fr-en` (French - English)
* `ru-en` (Russian - English)

And the current implementation always uses all the subsets available for a language, so to define custom subsets, you'll have to clone the repo from the Hub and replace the line https://huggingface.co/datasets/wmt15/blob/main/wmt_utils.py#L688 with:
`for split, ss_names in (self._subsets if self.config.subsets is None else self.config.subsets).items()`

Then, you can load the dataset as follows:
```python
from datasets import load_dataset
dset = load_dataset("path/to/local/wmt15_folder", "<one of 5 available configs>", subsets=...)

eldarkurtic:
@mariosasko thanks a lot for the suggested fix!

khushmeeet:
Hi @mariosasko 

[...]

PR #4554, opened 2022-06-24
Fix WMT dataset loading issue and docs update (Re-opened)

khushmeeet:
This PR is a fix for #4354 

Changes are made for `wmt14`, `wmt15`, `wmt16`, `wmt17`, `wmt18`, `wmt19` and `wmt_t2t`. And READMEs are updated for the corresponding datasets.

Let me know, if any additional changes are required.

Thanks

### DIFF
--- a/datasets/wmt14/wmt_utils.py
+++ b/datasets/wmt14/wmt_utils.py
@@ -25,7 +25,6 @@ import itertools
 import os
 import re
 import xml.etree.cElementTree as ElementTree
-from abc import ABC, abstractmethod
 
 import datasets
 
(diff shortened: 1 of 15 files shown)


datasets/wmt14/wmt_utils.py
```diff
@@ -34,22 +33,31 @@
...
-sources. The base `wmt_translate` allows you to create your own config to choose
-your own data/language pair by creating a custom `datasets.translate.wmt.WmtConfig`.
+Versions exist for different years using a combination of data
+sources. The base `wmt` allows you to create a custom dataset by choosing
+your own data/language pair. This can be done as follows:
 
 ```
-config = datasets.wmt.WmtConfig(
-    version="0.0.1",
+from datasets import inspect_dataset, load_dataset_builder
```
khushmeeet, line 43: Should I make similar changes here as well, As I did in the README file?

mariosasko: Yes, these code snippets must match the ones from the READMEs.

src/datasets/load.py
```diff
@@ -111,11 +111,13 @@ def import_main_class(module_path, dataset=True) -> Optional[Union[Type[DatasetB
...
     for name, obj in module.__dict__.items():
-        if isinstance(obj, type) and issubclass(obj, main_cls_type):
+        if inspect.isclass(obj) and issubclass(obj, main_cls_type):
             if inspect.isabstract(obj):
                 continue
             module_main_cls = obj
-            break
+            obj_module = inspect.getmodule(obj)
+            if obj_module is not None and module == obj_module:
+                break
```
lhoestq, line 120: This change sounds unrelated to the goal of this PR.

Can you remove this change ? If this is something useful, please open a new PR and explain why you think this is an improvement

mariosasko: I proposed this change to make it possible to load the `Wmt` class with `load_dataset_builder("path/to/wmt_utils.py")` and not break `load_dataset_builder("wmt<year>")` due to the `Wmt` class no longer being abstract (to support instantiation). And the way we do this is by giving higher precedence to the builder class defined in `module` than to the imported builder classes. 

But feel free to propose a cleaner solution.

lhoestq (approved):
It looks all good then, thanks. It's supernice to allow users to filter by subsets :)

I think we can merge unless @mariosasko has some comments
The CI failures are only about missing content in the dataset tags. It would be nice to add them, but this can be done in another PR

mariosasko (approved):
Yes, all looks good. Thanks @khushmeeet.

PR #4557, opened 2022-06-24
Add evaluation metadata to wmt16

lewtun:
Just to confirm: we should add this metadata via GitHub and not Hub PRs for canonical datasets right?


albertvillanova:

PR #4558, opened 2022-06-24
Add evaluation metadata to wmt14

lewtun:


datasets/wmt14/README.md
```diff
@@ -6,6 +6,15 @@ multilinguality:
...
 task_ids: []
+train-eval-index:
+- config: de-en
+  task: translation
+  task_id: translation
+  splits:
+    eval_split: test
+  col_mapping:
+    translation.de: source
+    translation.en: target
```
lhoestq, line 17: it could also be the other way around no ?

albertvillanova:

PR #4560, opened 2022-06-24
Add evaluation metadata to imagenet-1k

lewtun:


albertvillanova:

PR #4576, opened 2022-06-27
Include `metadata.jsonl` in resolved data files

### LINKED ISSUE #4548: Metadata.jsonl for Imagefolder is ignored if it's in a parent directory to the splits directories/do not have "{split}_" prefix
If data contains a single `metadata.jsonl` file for several splits, it won't be included in a dataset's `data_files` and therefore ignored. 
This happens when a directory is structured like as follows:
```
train/
  file_1.jpg
  file_2.jpg
test/
  file_3.jpg
  file_4.jpg
metadata.jsonl
```
or like as follows:
```
train_file_1.jpg
train_file_2.jpg
test_file_3.jpg
test_file_4.jpg
metadata.jsonl
```
The same for HF repos.

because it's ignored by the patterns [here](https://github.com/huggingface/datasets/blob/master/src/datasets/data_files.py#L29)

@lhoestq @mariosasko Do you think it's better to add this functionality in `data_files.py` or just specifically in imagefolder/audiofolder code? In `data_files.py` would me more general but I don't know if there are any other cases when that might be needed.

mariosasko:
Include `metadata.jsonl` in resolved data files.

Fix #4548 

@lhoestq ~~https://github.com/huggingface/datasets/commit/d94336d30eef17fc9abc67f67fa1c139661f4e75 adds support for metadata files placed at the root, and https://github.com/huggingface/datasets/commit/4d20618ea7a19bc143ddc5fdff9d79e671fcbb95 accounts for nested metadata files also, but this results in more complex code. Let me know which one of these two approaches you prefer.~~ Maybe https://github.com/huggingface/datasets/commit/d94336d30eef17fc9abc67f67fa1c139661f4e75 is good enough for now (for the sake of simplicity). https://github.com/huggingface/datasets/commit/4d20618ea7a19bc143ddc5fdff9d79e671fcbb95 breaks the imagefolder tests due to duplicates in the resolved metadata files. One way to fix this would be to resolve the metadata pattern only on parent directories, but this adds even more logic to `_get_data_files_patterns`, so not sure if this is what we should do.

### DIFF
--- a/src/datasets/data_files.py
+++ b/src/datasets/data_files.py
@@ -48,6 +48,7 @@ ALL_DEFAULT_PATTERNS = [
     DEFAULT_PATTERNS_SPLIT_IN_DIR_NAME,
     DEFAULT_PATTERNS_ALL,
 ]
+METADATA_PATTERN = "metadata.jsonl"  # metadata file for ImageFolder and AudioFolder
 WILDCARD_CHARACTERS = "*[]"
 FILES_TO_IGNORE = ["README.md", "config.json", "dataset_infos.json", "dummy_data.zip", "dataset_dict.json"]
 
@@ -103,7 +104,19 @@ def _get_data_files_patterns(pattern_resolver: Callable[[str], List[PurePath]]) 
             except FileNotFoundError:
                 pass
         if non_empty_splits:
-            return {split: patterns_dict[split] for split in non_empty_splits}
+            include_metadata_pattern = False
+            if patterns_dict != DEFAULT_PATTERNS_ALL:
+                try:
+                    pattern_resolver(METADATA_PATTERN)
+                except FileNotFoundError:
+                    pass
+                else:
+                    include_metadata_pattern = True
+
+            return {
+                split: patterns_dict[split] + [METADATA_PATTERN] if include_metadata_pattern else patterns_dict[split]
+                for split in non_empty_splits
+            }
     raise FileNotFoundError(f"Couldn't resolve pattern {pattern} with resolver {pattern_resolver}")
 
 
--- a/src/datasets/data_files.py
+++ b/src/datasets/data_files.py
@@ -48,7 +48,7 @@ ALL_DEFAULT_PATTERNS = [
     DEFAULT_PATTERNS_SPLIT_IN_DIR_NAME,
     DEFAULT_PATTERNS_ALL,
 ]
-METADATA_PATTERNS = ["metadata.jsonl", "**/metadata.jsonl"]  # metadata file for ImageFolder and AudioFolder
+METADATA_PATTERN = "metadata.jsonl"  # metadata file for ImageFolder and AudioFolder
 WILDCARD_CHARACTERS = "*[]"
 FILES_TO_IGNORE = ["README.md", "config.json", "dataset_infos.json", "dummy_data.zip", "dataset_dict.json"]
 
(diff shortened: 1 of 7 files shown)


mariosasko:
Yes, that's indeed the problem. My solution in https://github.com/huggingface/datasets/commit/4d20618ea7a19bc143ddc5fdff9d79e671fcbb95 that accounts for that (include metadata files only if image files are present; not ideal): https://github.com/huggingface/datasets/blob/4d20618ea7a19bc143ddc5fdff9d79e671fcbb95/src/datasets/data_files.py#L119-L125.
Perhaps a cleaner approach would be to check for metadata files after the packaged module type is inferred as `imagefolder` and append metadata files to already resolved data files (if there are any).

tests/test_load.py
```diff
@@ -76,6 +76,9 @@ def _generate_examples(self, filepath, **kwargs):
 SAMPLE_DATASET_IDENTIFIER = "lhoestq/test"  # has dataset script
 SAMPLE_DATASET_IDENTIFIER2 = "lhoestq/test2"  # only has data files
 SAMPLE_DATASET_IDENTIFIER3 = "mariosasko/test_multi_dir_dataset"  # has multiple data directories
+SAMPLE_DATASET_IDENTIFIER4 = (
+    "mariosasko/test_imagefolder_with_metadata"  # imagefolder with a metadata file outside of it
```
lhoestq, line 80:
```suggestion
    "mariosasko/test_imagefolder_with_metadata"  # imagefolder with a metadata file outside of the train/test directories
```

tests/test_load.py
```diff
@@ -311,6 +356,31 @@ def test_PackagedDatasetModuleFactory_with_data_dir(self):
...
+        assert Path(module_factory_result.builder_kwargs["data_files"]["train"][0]).parent.samefile(
+            self._data_dir_with_metadata
+        )
+        assert Path(module_factory_result.builder_kwargs["data_files"]["test"][0]).parent.samefile(
+            self._data_dir_with_metadata
+        )
+        assert any(
+            data_file.name == "metadata.jsonl"
+            for data_file in module_factory_result.builder_kwargs["data_files"]["train"]
+        ) and any(
```
lhoestq, line 379:
```suggestion
        )
        assert any(
```

tests/test_load.py
```diff
@@ -292,6 +320,23 @@ def test_LocalDatasetModuleFactoryWithoutScript_with_data_dir(self):
...
+        assert importlib.import_module(module_factory_result.module_path) is not None
+        assert (
+            module_factory_result.builder_kwargs["data_files"] is not None
+            and len(module_factory_result.builder_kwargs["data_files"]["train"]) > 0
+            and len(module_factory_result.builder_kwargs["data_files"]["test"]) > 0
+        )
+        assert any(
+            data_file.name == "metadata.jsonl"
+            for data_file in module_factory_result.builder_kwargs["data_files"]["train"]
+        ) and any(
```
lhoestq, line 335:
```suggestion
        )
        assert any(
```

lhoestq (approved):
I love this approach thanks !

Feel free to merge if the CI is ok

lhoestq:
The CI still struggles but you can merge since at least one of the two WIN CI succeeded

src/datasets/data_files.py
```diff
@@ -393,7 +422,7 @@ def resolve_patterns_in_dataset_repository(
     return data_files_urls
 
 
-def get_patterns_in_dataset_repository(
+def get_data_patterns_in_dataset_repository(
```
albertvillanova, line 425: This is a breaking change. 

See in:
- #4610 

There might be Hub dataset scripts that use `get_patterns_in_dataset_repository`.

CC: @mariosasko @lhoestq

PR #4577, opened 2022-06-27
Add authentication tip to `load_dataset`

mariosasko:
Add an authentication tip similar to the one in transformers' `PreTrainedModel.from_pretrained` to `load_dataset`/`load_dataset_builder`.


lhoestq (approved):
thanks !

PR #4579, opened 2022-06-27
Support streaming cfq dataset

albertvillanova:
Support streaming cfq dataset.


datasets/cfq/cfq.py
```diff
@@ -108,21 +107,16 @@ def _info(self):
 
     def _split_generators(self, dl_manager):
         """Returns SplitGenerators."""
-        data_dir = dl_manager.download_and_extract(_DATA_URL)
-        data_dir = os.path.join(data_dir, "cfq")
+        archive_path = dl_manager.download(_DATA_URL)
         return [
             datasets.SplitGenerator(
                 name=datasets.Split.TRAIN,
```
lhoestq, line 113: small error here:
```suggestion
                name=split,
```

albertvillanova: Thanks for the catch!

datasets/cfq/cfq.py
```diff
@@ -136,17 +130,20 @@ def _scrub_json(self, content):
...
-            with open(splits_path, encoding="utf-8") as splits_file:
+        samples_path = "cfq/dataset.json"
+        splits = samples = None
+        for path, file in data_files:
+            if path == self.config.splits_path:
+                splits = json.load(file)
+            elif path == samples_path:
                 logger.info("Reading json from %s into memory...", samples_path)
-                samples = json.loads(self._scrub_json(samples_file.read()))
+                samples = json.loads(self._scrub_json(file.read().decode("utf-8")))
```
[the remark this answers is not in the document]
albertvillanova: @lhoestq any comment on this?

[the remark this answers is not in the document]
albertvillanova: @lhoestq I've refactored the parsing of the file to reduce RAM usage.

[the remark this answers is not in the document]
lhoestq: Awesome thanks !

datasets/cfq/cfq.py
```diff
@@ -131,22 +130,36 @@ def _scrub_json(self, content):
...
+        for line in content:
+            line = line.decode("utf-8")
+            if not question_match:
+                question_match = question_regex.match(line)
+            else:
+                query_match = query_regex.match(line)
+                if query_match:
+                    result.append((question_match.group(1), query_match.group(1)))
+                    question_match = None
+        return "[" + ",".join(["{" + match[0] + "," + match[1] + "}" for match in result]) + "]"
```
lhoestq, line 147: maybe this can be done iteratively to not have all the dataset in RAM
```suggestion
                    yield json.loads("{" + question_match.group(1) + "," + query_match.group(1) + "}")
                    question_match = None
```

lhoestq: Ah good catch. I think it's fine to merge it as is then for now

Though ideally the dataset needs to be restructured to allow a practical streaming and lower RAM usage.

albertvillanova:
@lhoestq I've been refactoring a little the code:
- Use less RAM by loading only the required samples: only if its index is in the splits file
- Start yielding "earlier" in streaming mode: for each `split_idx`:
  - either yield from buffer
  - or iterate over samples and either yield or buffer the sample
  
  The speed gain obviously depends on how the indexes are sorted in the split file:
    - Best case: indices are [1, 2, 3]
    - Worst case (no speed gain): indices are [3, 1, 2] or [3, 2, 1]

Let me know what you think.

albertvillanova:
I have to update the dummy data so that it aligns with the real data (inside the archive, the samples file `dataset.json` is the last member).

albertvillanova:
There is an issue when testing `test_load_dataset_cfq` with dummy data:
- `MockDownloadManager.iter_archive` yields FIRST `'cfq/dataset.json'`
- [`Streaming`]`DownloadManager.iter_archive` yields LAST `'cfq/dataset.json'` when using real data tar.gz archive

Note that this issue arises only with dummy data: loading the real dataset works smoothly for all configurations: I recreated the `dataset_infos.json` file to check it (it generated the same file).

albertvillanova:
This PR should be merged first:
- #4611

Issue #3769, opened 2022-02-21
assign my_dataset.map() result to my_dataset cause faiss index lost


Oaklight:
## Describe the bug
assigning the resulted dataset to original dataset causes lost of the faiss index

## Steps to reproduce the bug
`my_dataset` is a regular loaded dataset. It's a part of a customed dataset structure
```python
self.my_dataset.add_faiss_index('embeddings')
self.my_dataset.list_indexes()
# ['embeddings']

my_dataset2 = my_dataset.map(
    lambda x: _get_nearest_examples_batch(x['text']), batch=True
)

# the unexpected result:
my_dataset2.list_indexes()
# []

self.my_dataset.list_indexes()
# ['embeddings']
```

in case something wrong with my `_get_nearest_examples_batch()`, it's like this
```python
def _get_nearest_examples_batch(self, examples, k=5):
    queries = embed(examples)
    scores_batch, retrievals_batch = self.dataset.get_nearest_examples_batch(self.faiss_column, queries, k)
    return {
        'neighbors': [batch['text'] for batch in retrievals_batch],
        'scores': scores_batch
    }
```

## Expected results
reassignment shouldn't intact the indexes

## Actual results
reassignment drops the indexes

## Environment info

- `datasets` version: 1.18.3
- Platform: Ubuntu 20.04.3 LTS
- Python version: 3.8.12
- PyArrow version: 7.0.0

lhoestq:
Hi ! Indeed `map` is dropping the index right now, because one can create a dataset with more or fewer rows using `map` (and therefore the index might not be relevant anymore)

I guess we could check the resulting dataset length, and if the user hasn't changed the dataset size we could keep the index, what do you think ?

cceyda:
doing `.add_column("x",x_data)` also removes the index. the new column might be irrelevant to the index so I don't think it should drop.

lhoestq:
I agree `add_column` shouldn't drop the index indeed ! Is it something you'd like to contribute ? I think it's just a matter of copying the `self._indexes` dictionary to the output dataset

PR #4582, opened 2022-06-27
add_column should preserve _indexes

cceyda:
https://github.com/huggingface/datasets/issues/3769#issuecomment-1167146126

doing `.add_column("x",x_data)` also removed any `_indexes` on the dataset, decided this shouldn't be the case.

This was because `add_column` was creating a new `Dataset(...)` and wasn't possible to pass indexes on init.
with this PR now can pass 'indexes' on init through `IndexableMixin`

- [ ] Added test

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -650,11 +650,12 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         info: Optional[DatasetInfo] = None,
         split: Optional[NamedSplit] = None,
         indices_table: Optional[Table] = None,
+        indexes: Optional[dict] = None,
         fingerprint: Optional[str] = None,
     ):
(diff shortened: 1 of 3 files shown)


src/datasets/arrow_dataset.py
```diff
@@ -4393,7 +4394,9 @@ def add_column(self, name: str, column: Union[list, np.array], new_fingerprint:
         info = dataset.info.copy()
         info.features.update(Features.from_arrow_schema(column_table.schema))
         table = update_metadata_with_features(table, info.features)
-        return Dataset(table, info=info, split=self.split, indices_table=None, fingerprint=new_fingerprint)
+        return Dataset(
```
cceyda, line 4397: Could just do it like below without touching the classes, but I don't know this looked better.
```
new_dataset= Dataset(table, info=info, split=self.split, indices_table=None, fingerprint=new_fingerprint)
        new_dataset._indexes = self._indexes
        return new_dataset
```

Issue #1718, opened 2021-01-11
Possible cache miss in datasets


ofirzaf:
Hi,

I am using the datasets package and even though I run the same data processing functions, datasets always recomputes the function instead of using cache.
I have attached an example script that for me reproduces the problem.
In the attached example the second map function always recomputes instead of loading from cache.
Is this a bug or am I doing something wrong?
Is there a way for fix this and avoid all the recomputation?

Thanks

Edit:
transformers==3.5.1
datasets==1.2.0

```
from datasets import load_dataset
from transformers import AutoTokenizer

datasets = load_dataset('wikitext', 'wikitext-103-raw-v1')
tokenizer = AutoTokenizer.from_pretrained('bert-base-uncased', use_fast=True)


column_names = datasets["train"].column_names
text_column_name = "text" if "text" in column_names else column_names[0]
def tokenize_function(examples):
    return tokenizer(examples[text_column_name], return_special_tokens_mask=True)

tokenized_datasets = datasets.map(
    tokenize_function,
    batched=True,
    num_proc=60,
    remove_columns=[text_column_name],
    load_from_cache_file=True,
)
max_seq_length = tokenizer.model_max_length
def group_texts(examples):
    # Concatenate all texts.
    concatenated_examples = {
        k: sum(examples[k], []) for k in examples.keys()}
    total_length = len(concatenated_examples[list(examples.keys())[0]])
    # We drop the small remainder, we could add padding if the model supported it instead of this drop, you can
    # customize this part to your needs.
    total_length = (total_length // max_seq_length) * max_seq_length
    # Split by chunks of max_len.
    result = {
        k: [t[i: i + max_seq_length]
            for i in range(0, total_length, max_seq_length)]
        for k, t in concatenated_examples.items()
    }
    return result

tokenized_datasets = tokenized_datasets.map(
    group_texts,
    batched=True,
    num_proc=60,
    load_from_cache_file=True,
)
print(tokenized_datasets)

print('finished')
```

lhoestq:
Thanks for reporting !
I was able to reproduce thanks to your code and find the origin of the bug.
The cache was not reusing the same file because one object was not deterministic. It comes from a conversion from `set` to `list` in the `datasets.arrrow_dataset.transmit_format` function, where the resulting list would not always be in the same order and therefore the function that computes the hash used by the cache would not always return the same result.
I'm opening a PR to fix this.

Also we plan to do a new release in the coming days so you can expect the fix to be available soon.
[...]

PR #4587, opened 2022-06-28
Validate new_fingerprint passed by user

lhoestq:
Users can pass the dataset fingerprint they want in `map` and other dataset transforms.

However the fingerprint is used to name cache files so we need to make sure it doesn't contain bad characters as mentioned in https://github.com/huggingface/datasets/issues/1718, and that it's not too long

### DIFF
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -105,11 +105,11 @@ class BuilderConfig:
 
     def __post_init__(self):
         # The config name is used to name the cache directory.
-        invalid_windows_characters = r"<>:/\|?*"
-        for invalid_char in invalid_windows_characters:
+        invalid_windows_characters_in_path = r"<>:/\|?*"
+        for invalid_char in invalid_windows_characters_in_path:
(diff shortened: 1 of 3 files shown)


src/datasets/builder.py
```diff
@@ -105,11 +105,11 @@ class BuilderConfig:
 
     def __post_init__(self):
         # The config name is used to name the cache directory.
-        invalid_windows_characters = r"<>:/\|?*"
-        for invalid_char in invalid_windows_characters:
+        invalid_windows_characters_in_path = r"<>:/\|?*"
```
albertvillanova, line 108: Maybe worth defining it as a constant at the module level? It is used again afterwards.

albertvillanova (approved):
Thank you.

src/datasets/builder.py
lhoestq: I moved `INVALID_WINDOWS_CHARACTERS_IN_PATH` to `naming.py` :)

PR #4599, opened 2022-06-29
Smooth-BLEU bug fixed

Aktsvigun:
Hi,

the current implementation of smooth-BLEU contains a bug: it smoothes unigrams as well. Consequently, when both the reference and translation consist of totally different tokens, it anyway returns a non-zero value (please see the attached image). 

This however contradicts the source paper suggesting the smooth-BLEU _(Chin-Yew Lin, Franz Josef Och. ORANGE: a method for evaluating automatic evaluation metrics for machine translation. COLING 2004.)_ :

> Add one count to the n-gram hit and total ngram count for n > 1. Therefore, for candidate translations with less than n words, they can still get a positive smoothed BLEU score from shorter n-gram matches; however if nothing matches then they will get zero scores. 

This pull request aims at fixing this bug.

I made a pull request in the target repository `tensorflow/nmt`, which implements this script, yet the last commit there is dating 19.02.2019 and I doubt whether this will be fixed promptly. Yet, this bug is critical, for instance for summarization datasets with short summaries (e.g. AESLC), since smoothing needs to be applied there. Therefore, the easiest solution that I found is to fork the repo and download this script directly from the forked fixed repo.

Kind,
Akim Tsvigun

Снимок экрана 2022-06-29 в 17 49 27

### DIFF
--- a/metrics/bleu/bleu.py
+++ b/metrics/bleu/bleu.py
@@ -104,7 +104,7 @@ class Bleu(datasets.Metric):
                     ),
                 }
             ),
-            codebase_urls=["https://github.com/tensorflow/nmt/blob/master/nmt/scripts/bleu.py"],
+            codebase_urls=["https://github.com/Aktsvigun/nmt/blob/master/nmt/scripts/bleu.py"],
             reference_urls=[
                 "https://en.wikipedia.org/wiki/BLEU",
                 "https://towardsdatascience.com/evaluating-text-output-in-nlp-bleu-at-your-own-risk-e8609665a213",


lhoestq (commented):
Thanks for the fix ! Though I think this has to be fixed in tensorflow/nmt directly. Once it's merged on their side, we can pin the right version here
```diff
- from .nmt_bleu import compute_bleu  # From: https://github.com/tensorflow/nmt/blob/master/nmt/scripts/bleu.py
+ from .nmt_bleu import compute_bleu  # From: https://github.com/tensorflow/nmt/blob/<commit-hash>/nmt/scripts/bleu.py
```

In the meantime I'd suggest mentioning you opened a pull request to fix a bug with smoothed bleu in the BLEU metric card: https://github.com/huggingface/evaluate/blob/main/metrics/bleu/README.md

albertvillanova:
Thanks @Aktsvigun for your fix.

However, metrics in `datasets` are in deprecation mode:
- #4739

You should transfer this PR to the `evaluate` library: https://github.com/huggingface/evaluate

Just for context, here the link to the PR by @Aktsvigun on tensorflow/nmt:
- https://github.com/tensorflow/nmt/pull/488

PR #4607, opened 2022-06-30
Align more metadata with other repo types (models,spaces)

julien-c:
see also associated PR on the `datasets-tagging` Space: https://huggingface.co/spaces/huggingface/datasets-tagging/discussions/2 (to merge after this one is merged)


src/datasets/utils/metadata.py
```diff
@@ -218,8 +218,10 @@ class _nothing:
 class DatasetMetadata:
     annotations_creators: List[str]
     language_creators: Union[EmptyList, List[str]]
-    languages: Union[EmptyList, List[str]]
-    licenses: List[str]
+    language: Union[EmptyList, List[str]]
+    languages: Union[EmptyList, List[str]]  # deprecated
+    license: List[str]
+    licenses: List[str]  # deprecated
```
lhoestq, line 224: I think you can simply give default values (an empty list should do the job) to `languages` and `licenses` to fix the CI

julien-c:
thanks so much @lhoestq !!

albertvillanova (approved):
Thank you @julien-c, really necessary to align as much as possible all projects.

Indeed I had already noticed that when creating a new dataset using the Hub UI, if a license was chosen from the dropdown menu, the generated tag was `license` instead of `licenses`. This fixes this issue as well.

julien-c:
There's also a follow-up PR to this one, in #4613 – I would suggest to merge all of them at the same time and hope not too many things are broken 🙀 🙀

lhoestq:
Alright merging this one now, let's see how broken things get

Issue #4605, opened 2022-06-30
Dataset Viewer issue for boris/gis_filtered


WaterKnight1998:
### Link

https://huggingface.co/datasets/boris/gis_filtered/viewer/boris--gis_filtered/train

### Description

When I try to access this from the website I get this error:

Status code:   400
Exception:     ClientResponseError
Message:       401, message='Unauthorized', url=URL('https://huggingface.co/datasets/boris/gis_filtered/resolve/80b805053ce61d4eb487b6b8d9095d775c2c466e/data/train/0000.parquet')

If I try to load with code I also get the same issue: 
```python
dataset2_train=load_dataset("boris/gis_filtered", use_auth_token=os.environ["HF_TOKEN"],split="train",streaming=True)
dataset2_validation=load_dataset("boris/gis_filtered", use_auth_token=os.environ["HF_TOKEN"], split="validation",streaming=True)
```

### Owner

No

severo:
Yes, this dataset is "gated": you first have to go to https://huggingface.co/datasets/boris/gis_filtered and click "Access repository" (if you accept to share your contact information with the repository authors).

WaterKnight1998:
I already did that, it returns error when using streaming

severo:
Oh, sorry, I misread. Looking at it. Maybe @huggingface/datasets or @SBrandeis

lhoestq:
I could reproduce the error, even though I provided my token and accepted the gate form

lhoestq:
This is indeed a bug in `datasets`. Parquet datasets in gated/private repositories can't be streamed properly, which caused the viewer to fail. I opened a PR at https://github.com/huggingface/datasets/pull/4608

PR #4608, opened 2022-06-30
Fix xisfile, xgetsize, xisdir, xlistdir in private repo

lhoestq:
`xisfile` is working in a private repository when passing a chained URL to a file inside an archive, e.g. `zip://a.txt::https://huggingface/datasets/username/dataset_name/resolve/main/data.zip`. However it's not working when passing a simple file `https://huggingface/datasets/username/dataset_name/resolve/main/data.zip`.

This is because the authentication headers are not passed correctly in this case.

This is causing dataset streaming to fail in private parquet repositories, as noted in https://github.com/huggingface/datasets/issues/4605

I fixed `xisfile` and the other functions that behave the same way: xgetsize, xisdir and xlistdir

TODO:
- [x] tests

fix https://github.com/huggingface/datasets/issues/4605

### DIFF
--- a/src/datasets/download/streaming_download_manager.py
+++ b/src/datasets/download/streaming_download_manager.py
@@ -294,8 +294,7 @@ def xisdir(path, use_auth_token: Optional[Union[str, bool]] = None) -> bool:
         return os.path.isdir(path)
     else:
         if not rest_hops and (main_hop.startswith("http://") or main_hop.startswith("https://")):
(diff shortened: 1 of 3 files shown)


lhoestq:
Added tests for xisfile, xgetsize, xlistdir and xglob for private repos, and also tests for xwalk that was untested

mariosasko (approved):
Thanks for fixing this!

PR #4611, opened 2022-07-01
Preserve member order by MockDownloadManager.iter_archive

albertvillanova:
Currently, `MockDownloadManager.iter_archive` yields paths to archive members in an order given by `path.rglob("*")`, which migh not be the same order as in the original archive.

See issue in:
- https://github.com/huggingface/datasets/pull/4579#issuecomment-1172135027

This PR fixes the order of the members yielded by `MockDownloadManager.iter_archive` so that it is the same as in the original archive.


lhoestq (approved):
That's clever ! LGTM :)

PR #4613, opened 2022-07-01
Align/fix license metadata info

julien-c:
fix bad "other-*" licenses and add the corresponding "license_details" when relevant


lhoestq (approved):
I fixed the merge conflicts and made load_tsv_licenses a bit more pythonic.

I also updated the license validator to no longer accept "other-*"

julien-c:
Thank you thank you! Let's merge and pray? 😱

lhoestq:
I just need to add `license_details` to the validator and yup we can merge

datasets/c3/README.md
```diff
@@ -6,7 +6,7 @@ language_creators:
 language:
 - zh
 license:
-- other-non-commercial-research
+- other
```
albertvillanova, line 9: When removing the text after "other", shouldn't we add that information with the `license-details` tag (like done e.g. in "afrikaans_ner_corpus"), to avoid losing that information?

In this particular example, I think it is important to keep that the license is only for non-commercial research purpose.

albertvillanova (commented):
Thanks again!

datasets/c3/README.md
lhoestq: Custom licenses can be written in the `### Licensing Information` section I think.

I think we can re-add this info in a subsequent PR

PR #4614, opened 2022-07-01
Ensure ConcatenationTable.cast uses target_schema metadata

dtuit:
Currently, `ConcatenationTable.cast` does not use target_schema metadata when casting subtables. This causes an issue when using cast_column and the underlying table is a ConcatenationTable.

Code example of where issue arrises:
```
from datasets import Dataset, Image

column1 = [0, 1]
image_paths = ['/images/image1.jpg', '/images/image2.jpg']

ds = Dataset.from_dict({"column1": column1})
ds = ds.add_column("image", image_paths)
ds.cast_column("image", Image()) # Fails here
```
Output
```
...
string
to
{'bytes': Value(dtype='binary', id=None), 'path': Value(dtype='string', id=None)}
```


lhoestq (commented):
Hi ! Thanks for opening a PR :)

The `target_schema.metadata` contains metadata about all the columns, while here the `subschema` may only contain a subset of the columns.

Therefore using the same metadata for the `subschema` can lead to issues:
```python
Features.from_arrow_schema(subschema)
```
would return the features of all the columns, instead of only the columns of the subschema.

Because of this, I think it would be more robust to:
1. convert the `target_schema` to `target_features = Features.from_arrow_schema(target_schema)`
2. then pick the right subfields `subfeatures = Features({subfield.name: target_features[subfield.name] for subfield in subfields})`
3. and finally convert the `subfeatures` to `subschema` with `subschema = subfeatures.arrow_schema`

Could you try that and let me know if it works ?

dtuit:
Hi @lhoestq, Thanks for the detailed comment. I've tested the suggested approach and can confirm it works for the testcase outlined above! The PR is updated with the changes.

Issue #2832, opened 2021-08-24
Logging levels not taken into account


LysandreJik:
## Describe the bug

The `logging` module isn't working as intended relative to the levels to set.

## Steps to reproduce the bug

```python
from datasets import logging

logging.set_verbosity_debug()
logger = logging.get_logger()

logger.error("ERROR")
logger.warning("WARNING")
logger.info("INFO")
logger.debug("DEBUG"
```

## Expected results

I expect all logs to be output since I'm putting a `debug` level.

## Actual results

Only the two first logs are output.

## Environment info

- `datasets` version: 1.11.0
- Platform: Linux-5.13.9-arch1-1-x86_64-with-glibc2.33
- Python version: 3.9.6
- PyArrow version: 5.0.0

## To go further

This logging issue appears in `datasets` but not in `transformers`. It happens because there is no handler defined for the logger. When no handler is defined, the `logging` library will output a one-off error to stderr, using a `StderrHandler` with level `WARNING`.

`transformers` sets a default `StreamHandler` [here](https://github.com/huggingface/transformers/blob/5c6eca71a983bae2589eed01e5c04fcf88ba5690/src/transformers/utils/logging.py#L86)

yweweler:
I just take a look at all the outputs produced by `datasets` using the different log-levels.
As far as i can tell using `datasets==1.17.0` they overall issue seems to be fixed.

However, I noticed that there is one tqdm based progress indicator appearing on STDERR that I can simply not suppress.
```
Resolving data files: 100%|██████████| 652/652 [00:00<00:00, 1604.52it/s]
```

According to _get_origin_metadata_locally_or_by_urls it shold be supressable by using the `NOTSET` log-level
https://github.com/huggingface/datasets/blob/1406a04c3e911cec2680d8bc513653e0cafcaaa4/src/datasets/data_files.py#L491-L501
Sadly when specifiing the log-level `NOTSET` it seems to has no effect.

But appart from it not having any effect I must admit that it seems unintuitive to me.
I would suggest changing this such that it is only shown when the log-level is greater or equal to INFO.

This would conform better to INFO according to the [documentation](https://huggingface.co/docs/datasets/v1.0.0/package_reference/logging_methods.html#datasets.logging.set_verbosity_info).
> This will display most of the logging information and tqdm bars.

Any inputs on this?
I will be happy to supply a PR if desired 👍

mariosasko:
Hi! This should disable the tqdm output:
```python
import datasets
[...]

PR #4622, opened 2022-07-04
Fix ImageFolder with parameters drop_metadata=True and drop_labels=False (when metadata.jsonl is present)

### LINKED ISSUE #4621: ImageFolder raises an error with parameters drop_metadata=True and drop_labels=False when metadata.jsonl is present
## Describe the bug

If you pass `drop_metadata=True` and `drop_labels=False` when a `data_dir` contains at least one `matadata.jsonl` file, you will get a KeyError. This is probably not a very useful case but we shouldn't get an error anyway. Asking users to move metadata files manually outside `data_dir` or pass features manually (when there is a tool that can infer them automatically) don't look like a good idea to me either.

### Clone an example dataset from the Hub
```bash
git clone https://huggingface.co/datasets/nateraw/test-imagefolder-metadata
```
### Try to load it
```python
ds = load_dataset("test-imagefolder-metadata", drop_metadata=True, drop_labels=False)
```
or even just
```python
ds = load_dataset("test-imagefolder-metadata", drop_metadata=True)
```
as `drop_labels=False` is a default value.

## Expected results
A DatasetDict object with two features: `"image"` and `"label"`.

## Actual results
```
  File "/home/polina/workspace/datasets/debug.py", line 18, in <module>
    ds = load_dataset(
  File "/home/polina/workspace/datasets/src/datasets/load.py", line 1732, in load_dataset
  File "/home/polina/workspace/datasets/src/datasets/builder.py", line 704, in download_and_prepare
    self._download_and_prepare(
  File "/home/polina/workspace/datasets/src/datasets/builder.py", line 1227, in _download_and_prepare
    super()._download_and_prepare(dl_manager, verify_infos, check_duplicate_keys=verify_infos)
  File "/home/polina/workspace/datasets/src/datasets/builder.py", line 793, in _download_and_prepare
  File "/home/polina/workspace/datasets/src/datasets/builder.py", line 1218, in _prepare_split
    example = self.info.features.encode_example(record)
  File "/home/polina/workspace/datasets/src/datasets/features/features.py", line 1596, in encode_example
    return encode_nested_example(self, example)
  File "/home/polina/workspace/datasets/src/datasets/features/features.py", line 1165, in encode_nested_example
    {
  File "/home/polina/workspace/datasets/src/datasets/features/features.py", line 1165, in <dictcomp>
    {
  File "/home/polina/workspace/datasets/src/datasets/utils/py_utils.py", line 249, in zip_dict
    yield key, tuple(d[key] for d in dicts)
  File "/home/polina/workspace/datasets/src/datasets/utils/py_utils.py", line 249, in <genexpr>
    yield key, tuple(d[key] for d in dicts)
KeyError: 'label'
```

## Environment info
`datasets` master branch 

- `datasets` version: 2.3.3.dev0
- Platform: Linux-5.14.0-1042-oem-x86_64-with-glibc2.17
- Python version: 3.8.12
- PyArrow version: 6.0.1
- Pandas version: 1.4.1

polinaeterna:
Will fix #4621 

ImageFolder raises `KeyError: 'label'` with params `drop_metadata=True` and `drop_labels=False` (if there is at least one metadata.jsonl file a data directory). This happens because metadata files are collected inside `analyze()` function regardless of `drop_metadata` value. And then the following condition doesn't pass: https://github.com/huggingface/datasets/blob/master/src/datasets/packaged_modules/imagefolder/imagefolder.py#L167

So I suggest to double check it inside `analyze()` not to collect metadata files if they are not needed. (and labels too, to be consistent)

---
Also, I added a test to check if labels are inferred correctly from directories names in general (because we didn't have it) :)


polinaeterna:
@mariosasko @lhoestq if you agree with this, I'll add a test to check this case

lhoestq (approved):
Awesome thanks ! cc @mariosasko can you take a look as well ?

tests/packaged_modules/test_imagefolder.py
```diff
@@ -17,6 +17,24 @@ def cache_dir(tmp_path):
     return str(tmp_path / "imagefolder_cache_dir")
 
 
+@pytest.fixture
+def data_dir_with_labels(tmp_path, image_file):
```
mariosasko, line 21: One nit for consistent fixture names in this test module

```suggestion
def data_files_with_no_metadata(tmp_path, image_file):
```

tests/packaged_modules/test_imagefolder.py
```diff
@@ -195,6 +233,11 @@ def test_generate_examples_drop_metadata(image_file_with_metadata, drop_metadata
             example.keys() == {"image", "caption"} and all(val is not None for val in example.values())
             for _, example in generator
         )
+    elif not drop_metadata:
```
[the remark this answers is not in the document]
polinaeterna: wow that was incorrect and I missed it, thank you!

tests/packaged_modules/test_imagefolder.py
mariosasko: You also need to update the fixture name in the test

polinaeterna:
@lhoestq it wouldn't raise an error if we check `example.keys() == {"image", "label"}` as test checks only `_generate_examples`, not `encode_example`. and in the implementation of this PR `_generate_examples` would return both `image` and `label` key in the case when `drop_metadata=True` and `drop_labels=False` (default) as it seems that we agreed on that :)

polinaeterna:
and on the other hand it would raise an error if `label` column is missing in _generate_examples when `drop_metadata=True` and `drop_labels=False`

but "it" i mean tests :D (`test_generate_examples_with_metadata_that_misses_one_image`, `test_generate_examples_with_metadata_in_wrong_location` and `test_generate_examples_drop_metadata`)

mariosasko:
Perhaps we could make `self.config.drop_metadata = None` and `self.config.drop_labels = None` the defaults to see explicitly what the user wants. This would then turn into  `self.config.drop_metadata = False` and `self.config.drop_labels = True` if metadata files are present and `self.config.drop_metadata = True` and `self.config.drop_labels = False` if not. And if the user wants to have the `label` column alongside metadata columns, it can do so by passing `drop_labels = False` explicitely (in that scenario we have to check that the `label` column is not already present in metadata files). And maybe we can also improve the logging messages.

I find it problematic that the current implementation drops labels in some scenarios even if `self.config.drop_labels = False`, and the user doesn't have control over this behavior.

Let me know what you think.

src/datasets/packaged_modules/imagefolder/imagefolder.py
```diff
@@ -124,19 +120,36 @@ def analyze(files_or_archives, downloaded_files_or_dirs, split):
             if do_analyze:
                 analyze(files, downloaded_files, split_name)
                 analyze(archives, downloaded_dirs, split_name)
+
+                if metadata_files:
+                    add_metadata = not (self.config.drop_metadata is True)
+                    add_labels = self.config.drop_labels is False
+                else:
+                    add_metadata = False
+                    add_labels = not (self.config.drop_labels is True)
```
lhoestq, line 129: I just added this logic, I hope it helps making things clearer.

I also updated the tests you mentioned @polinaeterna : 
- test_generate_examples_with_metadata_that_misses_one_image
- test_generate_examples_with_metadata_in_wrong_location
- test_generate_examples_drop_metadata

Feel free to revert if you think it's not the right solution.

polinaeterna: I like it that way, thank you! 

So the default behavior (both `drop_metadata` and `drop_labels` are None) is: 
parse metadata and labels, if `metadatata_files` are found - add metadata and do not add inferred labels even if they are found. Add labels only if it’s explicitly specified by a user (`drop_labels` is set to False)

If metadata files are not found, default behavior is to add labels if they are found.

And if we add both inferred labels and metadata, if "label" column is found in metadata, it replaces the inferred labels. 

Right?

polinaeterna: ah no, I see that it raises an error if "label" feature is present both in `metadata.jsonl` and added if `add_labels` is True here: https://github.com/huggingface/datasets/pull/4622/files#diff-8bb31057be49645ccb98250431e4ca7e663d1b3413c017a4c932f0c8b77b8188R194

do you think we should preserve this behavior? if so, ignore my comments below

lhoestq: Yea the rule is:
- if there are metadata, use the metadata
- otherwise, infer labels from the file names
- if the user is not interested by the provided metadata or the labels, they can explicitly drop them

Then there how to get labels from a dataset with metadata, and in this case the user can again explicitly ask for the labels with drop_labels=False (but this is a rare case IMO). And in this case, I think the "label" column should override the "label" column in the metadata (otherwise the user can simply not ask for the labels).

Therefore I would keep the current behavior:
```python
{..., **image_metadata, **image_label}
```

polinaeterna: ok, agree! but current behavior in cases when "label" is both inferred from data files (and `drop_labels`=False) and is present in metadata features (and `drop_metadata` is None or False) is to raise an error:
```
ValueError: Metadata feature keys ['label'] are already present as the image features
```

Another option would be to give a warning in this case and add to `self.config.features` only those features that are not already present in `self.config.features` (because we want "label" column to be for sure a ClassLabel but it might be inferred from metadata as a string). But I agree it seems to be a corner case (though I'm not an expert in image processing at all)

polinaeterna: @mariosasko what do you think about the case described above? do you think we should preserve the current behavior and raise error when “label” is presented both in metadata and inferred from filenames (while drop_labels=False)? I personally would prefer to log a warning and chose labels from filenames but it’s not that important at all and we can leave as is if you agree on that.
I added some comments to the code and modified ImageFolder documentation, if you think they’re ok I guess we can merge?

mariosasko: `drop_labels=False` means the labels from filenames are explicitly requested, so a warning sounds good. Please also add a test for that.

mariosasko (commented):
Looks good already! I think we can merge after we address the following comments + warning for the duplicated "label" column.

src/datasets/packaged_modules/imagefolder/imagefolder.py
```diff
@@ -174,14 +194,18 @@ def analyze(files_or_archives, downloaded_files_or_dirs, split):
...
-                # Verify that there are no duplicated keys when compared to the existing features ("image", optionally "label")
+            if add_metadata:
+                # Warn if there are duplicated keys in metadata compared to the existing features ("image", optionally "label")
                 duplicated_keys = set(self.info.features) & set(metadata_features)
                 if duplicated_keys:
-                    raise ValueError(
-                        f"Metadata feature keys {list(duplicated_keys)} are already present as the image features"
+                    warnings.warn(
+                        f"Metadata feature keys {list(duplicated_keys)} are already present as the image features, "
+                        f"using image {list(duplicated_keys)} features instead of the metadata ones."
```
mariosasko, line 203: We prefer warnings from the logger in dataset scripts over `warnings.warn` (+ sentence reformulation to make it easier to understand):
```suggestion
                    logger.warning(
                        f"Ignoring metadata columns {list(duplicated_keys)} as they are already present in the features dictionary."
```

mariosasko (commented):
Two nits.

src/datasets/packaged_modules/imagefolder/imagefolder.py
polinaeterna: @mariosasko I sometimes encounter problems with the datasets logger even if I set logging level manually to info or even debug. so I decided that warning would be better as it is shown independently of logger settings.

polinaeterna: btw thank you for reformulation!! I'm so bad at this :)

mariosasko: Could this issue be related to https://github.com/huggingface/datasets/issues/2832?

PS: You can use the `caplog` fixture in the tests to test the logger's output

polinaeterna: yes, adding `library_root_logger.addHandler(logging.StreamHandler())` to https://github.com/huggingface/datasets/blob/main/src/datasets/utils/logging.py#L67 fixed the issue!

PR #4624, opened 2022-07-04
Remove all paperswithcode_id: null

lhoestq:
On the Hub there is a validation error on the `paperswithcode_id` tag when the value is `null`:

image

We've been using `null` to specify that we checked on pwc but the dataset doesn't exist there.

To have the validation working again we can simply remove all the `paperswithcode_id: null`.

cc @julien-c


lhoestq:
Yup it's maybe better to support it on the Hub side then indeed, thanks ! Closing this one

PR #4625, opened 2022-07-04
Unpack `dl_manager.iter_files` to allow parallization

mariosasko:
Iterate over data files outside `dl_manager.iter_files` to allow parallelization in streaming mode.

(The issue reported [here](https://discuss.huggingface.co/t/dataset-only-have-n-shard-1-when-has-multiple-shards-in-repo/19887))

PS: Another option would be to override `FilesIterable.__getitem__` to make it indexable and check for that type in `_shard_kwargs` and `n_shards,` but IMO this solution adds too much unnecessary complexity.

### DIFF
--- a/src/datasets/packaged_modules/csv/csv.py
+++ b/src/datasets/packaged_modules/csv/csv.py
@@ -1,3 +1,4 @@
+import itertools
 from dataclasses import dataclass
 from typing import List, Optional, Union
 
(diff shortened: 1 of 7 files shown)


lhoestq:
Cool thanks ! Yup it sounds like the right solution.

It looks like `_generate_tables` seem to be updated as well to fix the CI

PR #4627, opened 2022-07-04
fixed duplicate calculation of spearmanr function in metrics wrapper.

benlipkin:
During _compute, the scipy.stats spearmanr function was called twice, redundantly, once for calculating the score and once for calculating the p-value, under the conditional branch where return_pvalue=True. I adjusted the _compute function to execute the spearmanr function once, store the results tuple in a temporary variable, and then pass the indexed contents to the expected keys of the returned dictionary.

### DIFF
--- a/metrics/spearmanr/spearmanr.py
+++ b/metrics/spearmanr/spearmanr.py
@@ -112,12 +112,8 @@ class Spearmanr(datasets.Metric):
         )
 
     def _compute(self, predictions, references, return_pvalue=False):
+        results = spearmanr(references, predictions)
         if return_pvalue:
-            return {
-                "spearmanr": spearmanr(references, predictions)[0],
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
Thank you !

I think the same improvement can be applied in the `evaluate` library, which is the new library we're using to add new metrics :)

https://github.com/huggingface/evaluate/blob/main/metrics/spearmanr/spearmanr.py

benlipkin:
PR opened in`evaluate` library with same minor adjustment: https://github.com/huggingface/evaluate/pull/176

lhoestq:
> If I were to open a PR with the wrapper, description, citation, docstrings, readme, etc. would it make more sense to do that in the datasets or evaluate repo (or both)?

I think you could just add it to `evaluate`, we're not adding new metrics in this repo anymore

PR #4631, opened 2022-07-04
Update WinoBias README

sashavor:
I'm adding some information about Winobias that I got from the paper :smile:

I think this makes it a bit clearer!


lhoestq (approved):
Cool thanks ! :)

Issue #4477, opened 2022-06-11
Dataset Viewer issue for fgrezes/WIESP2022-NER


AshTayade:

severo:
https://huggingface.co/datasets/fgrezes/WIESP2022-NER

The error:

```
Message:       Couldn't find a dataset script at /src/services/worker/fgrezes/WIESP2022-NER/WIESP2022-NER.py or any data file in the same directory. Couldn't find 'fgrezes/WIESP2022-NER' on the Hugging Face Hub either: FileNotFoundError: Unable to resolve any data file that matches ['**test*', '**eval*'] in dataset repository fgrezes/WIESP2022-NER with any supported extension ['csv', 'tsv', 'json', 'jsonl', 'parquet', 'txt', 'blp', 'bmp', 'dib', 'bufr', 'cur', 'pcx', 'dcx', 'dds', 'ps', 'eps', 'fit', 'fits', 'fli', 'flc', 'ftc', 'ftu', 'gbr', 'gif', 'grib', 'h5', 'hdf', 'png', 'apng', 'jp2', 'j2k', 'jpc', 'jpf', 'jpx', 'j2c', 'icns', 'ico', 'im', 'iim', 'tif', 'tiff', 'jfif', 'jpe', 'jpg', 'jpeg', 'mpg', 'mpeg', 'msp', 'pcd', 'pxr', 'pbm', 'pgm', 'ppm', 'pnm', 'psd', 'bw', 'rgb', 'rgba', 'sgi', 'ras', 'tga', 'icb', 'vda', 'vst', 'webp', 'wmf', 'emf', 'xbm', 'xpm', 'zip']
```

I understand the issue is not related to the dataset viewer in itself, but with the autodetection of the data files without a loading script in the datasets library. cc @lhoestq @albertvillanova @mariosasko

lhoestq:
Apparently it finds `scoring-scripts/compute_seqeval.py` which matches `**eval*`, a regex that detects a test split. We should probably improve the regex because it's not supposed to catch this kind of files. It must also only check for files with supported extensions: txt, csv, etc.

PR #4633, opened 2022-07-05
[data_files] Only match separated split names

lhoestq:
As reported in https://github.com/huggingface/datasets/issues/4477, the current pattern matching to infer which file goes into which split is too permissive. For example a file "contest.py" would be considered part of a test split (it contains "test") and "seqeval.py" as well (it contains "eval").

In this PR I made the pattern matching more robust by only matching split names **between separators**. The supported separators are dots, dashes, spaces and underscores.

I updated the docs accordingly.

One detail about the tests: I had to update one test because it was using `PurePath.match` as a reference for globbing, but it doesn't support the `[..]` glob pattern. Therefore I added a `mock_fs` context manager that can be used to easily define a dummy filesystem with certain files in it and run pattern matching tests. Its code comes mostly from test_streaming_download_manager.py

Close https://github.com/huggingface/datasets/issues/4477


lhoestq:
I ran a script to find affected datasets. Adding "testing" and "evaluation" fixes all of of them except one:
- projecte-aina/cat_manynames:	human_annotated_testset.tsv

Let me open a PR on their repository to fix it

mariosasko (approved):
All looks good now. Thanks!

tests/test_data_files.py
```diff
@@ -491,6 +493,69 @@ def test_DataFilesDict_from_hf_local_or_remote_hashing(text_file):
...
+@contextmanager
+def mock_fs(file_paths: List[str]):
+    """context manager to set up a mock:// filesystem in sfspec containing the provided files"""
+
+    dir_paths = {file_path.rsplit("/")[0] for file_path in file_paths if "/" in file_path}
+    fs_contents = [{"name": dir_path, "type": "directory"} for dir_path in dir_paths] + [
+        {"name": file_path, "type": "file", "size": 10} for file_path in file_paths
+    ]
+
+    class DummyTestFS(AbstractFileSystem):
```
albertvillanova, line 505: Awesome that you use a mock file system. I personally think this goes in the right direction.

tests/test_data_files.py
```diff
@@ -491,6 +493,69 @@ def test_DataFilesDict_from_hf_local_or_remote_hashing(text_file):
...
+                path,
+                mode,
+                block_size,
+                autocommit,
+                cache_options=cache_options,
+                **kwargs,
+            )
+
+    with patch.dict(fsspec.registry.target, {"mock": DummyTestFS}):
+        yield DummyTestFS()
```
albertvillanova, line 556: Just a comment (feel free to ignore it): you use here `unittest.mock.patch`, but you could use `pytest.monkeypatch` instead.

tests/test_data_files.py
```diff
@@ -491,6 +493,69 @@ def test_DataFilesDict_from_hf_local_or_remote_hashing(text_file):
         assert Hasher.hash(data_files1) != Hasher.hash(data_files2)
 
 
+@contextmanager
+def mock_fs(file_paths: List[str]):
+    """context manager to set up a mock:// filesystem in sfspec containing the provided files"""
```
albertvillanova, line 498: Typo
```suggestion
    """context manager to set up a mock:// filesystem in fsspec containing the provided files"""
```

tests/test_data_files.py
```diff
@@ -541,25 +606,30 @@ def test_DataFilesDict_from_hf_local_or_remote_hashing(text_file):
         {"validation": "dev/dataset.txt"},
         # With other extensions
         {"train": "train.parquet", "test": "test.parquet", "validation": "valid.parquet"},
+        # With "dev" or "eval" without separators
+        {"train": "developers_list.txt"},
+        {"train": "data/seqeval_results.txt"},
```
albertvillanova, line 611: Maybe also adding a test for "test": "contest.txt"?

tests/test_data_files.py
```diff
@@ -541,25 +606,30 @@ def test_DataFilesDict_from_hf_local_or_remote_hashing(text_file):
...
+    ) as fs:
+
+        def resolver(pattern):
+            return [PurePath(file_path) for file_path in fs.glob(pattern) if fs.isfile(file_path)]
+
+        patterns_per_split = _get_data_files_patterns(resolver)
+        assert sorted(patterns_per_split.keys()) == sorted(data_file_per_split.keys())
+        for split, patterns in patterns_per_split.items():
+            matched = [file_path.as_posix() for pattern in patterns for file_path in resolver(pattern)]
+            assert matched == data_file_per_split[split]
```
albertvillanova, line 632: Maybe better moving the context manager to the specific position where it is necessary?
```suggestion

    def resolver(pattern):
        with mock_fs(
            [file_path for split_file_paths in data_file_per_split.values() for file_path in split_file_paths]
        ) as fs:
            return [PurePath(file_path) for file_path in fs.glob(pattern) if fs.isfile(file_path)]

    patterns_per_split = _get_data_files_patterns(resolver)
    assert sorted(patterns_per_split.keys()) == sorted(data_file_per_split.keys())
    for split, patterns in patterns_per_split.items():
        matched = [file_path.as_posix() for pattern in patterns for file_path in resolver(pattern)]
        assert matched == data_file_per_split[split]
```

albertvillanova (commented):
Awesome!! This will definitely make the user experience much more pleasant and aligned with their expectations.

tests/test_data_files.py
lhoestq: This function is a context manager that doesn't have access to the `monkeypatch` fixture of pytest, so I used `unittest.mock.patch` instead.

albertvillanova (approved):
Thank you!! Nice job.

tests/test_data_files.py
albertvillanova: @lhoestq again not important: but indeed you are not using the patching (neither `unittest.mock.patch` nor `pytest.monkeypatch`. You are just using the returned instance `DummyTestFS()`.

So I guess you could just remove the patching and the test will pass anyway.

albertvillanova:
```suggestion
    yield DummyTestFS()
```

PR #4638, opened 2022-07-06
The speechocean762 dataset

jimbozhang:
[speechocean762](https://www.openslr.org/101/) is a non-native English corpus for pronunciation scoring tasks. It is free for both commercial and non-commercial use.

I believe it will be easier to use if it could be available on Hugging Face.


datasets/speechocean762/speechocean762.py
```diff
@@ -0,0 +1,174 @@
...
+        root_path = Path(local_extracted_archive) / "speechocean762"
+        meta_path = root_path / f"{split}.json"
+
+        with open(meta_path, encoding="utf-8") as f:
+            info = json.load(f)
+            for utt_id in info:
+                audio = info[utt_id]
+                audio["id"] = utt_id
+                audio["original_full_path"] = f"{root_path}/WAVE/SPEAKER{audio['speaker']}/{utt_id}.WAV"
+                audio_data, _ = librosa.load(audio["original_full_path"], sr=_SAMPLE_RATE)
```
patrickvonplaten, line 171: I don't think we should load the audio data here - this is done automatically when the feature "audio" is being called. 

Could we maybe instead do:
```suggestion
                audio_bytes = audio["original_full_path"].read()
                ...
                audio["audio"] = {"path": audio["original_full_path"], "bytes": audio_bytes}
```

patrickvonplaten:
@sanchit-gandhi could you also maybe take a quick look? :-)

datasets/speechocean762/speechocean762.py
jimbozhang: Thanks for your kind suggestion. I think `audio["original_full_path"].read()` would not work here because `audio["original_full_path"]` is a str without the `read()` method:

```python
# Line 170 of datasets/speechocean762/speechocean762.py
audio["original_full_path"] = f"{root_path}/WAVE/SPEAKER{audio['speaker']}/{utt_id}.WAV"
```

I didn't find related example codes in `datasets/librispeech_asr/librispeech_asr.py`. So I'm not sure how to make the feature "audio" with `read()` :flushed:

datasets/speechocean762/speechocean762.py
```diff
@@ -0,0 +1,174 @@
...
+            ),
+        ]
+
+    def _generate_examples(self, local_extracted_archive, split):
+        root_path = Path(local_extracted_archive) / "speechocean762"
+        meta_path = root_path / f"{split}.json"
+
+        with open(meta_path, encoding="utf-8") as f:
+            info = json.load(f)
+            for utt_id in info:
```
lhoestq, line 167: and here you need to iterate over the files of the TAR archive
```python
for path, f in files:
    utt_id = path.split("/")[-1].split(".")[0]
    audio_info = info[utt_id]
    audio_info["audio"] = {"path": path, "bytes": f.read()}
    yield utt_id, audio_info
```

note that you may have to host the JSON file somewhere else, to load it separately.

datasets/speechocean762/README.md
```diff
@@ -0,0 +1,260 @@
...
+English. Note that the mother tongue of all the speakers is Mandarin. 
+
+## Dataset Structure
+
+### Data Instances
+
+Below is a typical meta data of an instance of the dataset:
+
+```
+"014350017": {    ,
```
lhoestq, line 81: Can you provide one single example here instead ? This way users know exactly what the content of the dataset looks like.

You can get one using `load_dataset` and `ds["train"][0]`

datasets/speechocean762/README.md
```diff
@@ -0,0 +1,260 @@
...
+        }
+   ]
+}
+```
+
+### Data Fields
+
+- text: transcript text
+
+- accuracy: the pronunciation accuracy
```
lhoestq, line 171:
```suggestion
- accuracy: the pronunciation accuracy (0-10)
```

datasets/speechocean762/README.md
```diff
@@ -0,0 +1,260 @@
...
+
+- completeness: the percentage of the words in the target text that are actually pronounced
+
+- fluency: intonation, stable speaking speed and rhythm
+
+- stress: whether the stress position is correct
+
+- mispronunciation: only available when phone accuracy is lower than 0.5. Indicates which phoneme the current phone was actually pronounced.
+
```
lhoestq, line 180: Can you also describe all the other fields here please ? Especially the "words" one if possible

lhoestq (commented):
Thanks for adding this dataset :)

I added a few comments, mainly about dataset streaming (which is required for the dataset preview to work on huggingface.co)

Moreover I think it would be nice to create the `speechocean` organization (feel free to use a more appropriate name though if needed) on huggingface.co and put this dataset under the `speechocean` namespace, instead of having it as a standalone dataset.

datasets/speechocean762/speechocean762.py
sanchit-gandhi: An example from Gigaspeech on how this might be achieved: https://huggingface.co/datasets/speechcolab/gigaspeech/blob/main/gigaspeech.py#L345

datasets/speechocean762/speechocean762.py
sanchit-gandhi: Once you iterate over the files of the TAR archive as per @lhoestq's suggestion you will be able to `.read()` the `audio_file`!

sanchit-gandhi (commented):
Thanks @jimbozhang for working on this dataset and making it accessible to the speech community 🤗 Addressing @lhoestq's suggestion of iterating over TAR files should also then resolve @patrickvonplaten's suggestion of directly reading bytes and dropping the `librosa` requirement

datasets/speechocean762/speechocean762.py
polinaeterna: Also make sure you get a copy of audio sample metadata with `audio_info = dict(info[utt_id])` instead of `audio_info = info[utt_id]` in order not to fill the original `info` metadata dict with audio bytes to avoid increasing RAM usage.

albertvillanova:
Thanks for your contribution, @jimbozhang. Are you still interested in adding this dataset?

jimbozhang:
> Thanks for your contribution, @jimbozhang. Are you still interested in adding this dataset?
> 
> We are removing the dataset scripts from this GitHub repo and moving them to the Hugging Face Hub: https://huggingface.co/datasets
> 
> We would suggest you create this dataset there. Please, feel free to tell us if you need some help.

Yes, I just planned to complete this dataset these days, and this suggestion is just in time! Thanks a lot!
I will create this dataset to Hugging Face Hub soon, maybe this week.

PR #4663, opened 2022-07-08
Add text decorators

stevhliu:
This PR adds some decoration to text about different modalities to make it more obvious separate guides exist for audio, vision, and text. The goal is to make it easier for users to discover these guides!

underline

TODO:

- [x] Open PR to support new Tailwind classes


lhoestq (approved):
Love it, thanks :)

PR #4679, opened 2022-07-13
Added method to remove excess nesting in a DatasetDict

CakeCrusher:
Added the ability for a DatasetDict to remove additional nested layers within its features to avoid conflicts when collating. It is meant to accompany [this PR](https://github.com/huggingface/transformers/pull/18119) to resolve the same issue [#15505](https://github.com/huggingface/transformers/issues/15505).

@stas00 @lhoestq

### DIFF
--- a/src/datasets/dataset_dict.py
+++ b/src/datasets/dataset_dict.py
@@ -1025,6 +1025,24 @@ class DatasetDict(dict):
             }
         )
 
+    def remove_excess_nesting(self) -> "DatasetDict":
+        """
+        Flattens any leading dimensions of shape 1 for all features in the dataset.
+        """
+
+        def clean_row_nesting(row):
+            row_keys = row.keys()
+            for key in row_keys:
+                arr = np.array(row[key])
+                for _ in range(arr.ndim - 1):
+                    arr = arr[0] if arr.shape[0] == 1 else arr
+                row[key] = arr.tolist()
+            return row
+
+        claned_dataset = self.map(clean_row_nesting)
+
+        return claned_dataset
+
     def save_to_disk(self, dataset_dict_path: str, fs=None):
         """
         Saves a dataset dict to a filesystem using either :class:`~filesystems.S3FileSystem` or


lhoestq:
Hi ! I think the issue you linked is closed and suggests to use `remove_columns`.

Moreover if you end up with a dataset with an unnecessarily nested data, please modify your processing functions to not output nested data, or use `map(..., batched=True)` if you function take batches as input

CakeCrusher:
Hi @lhoestq , you are right about the issues this pull has steered beyond that issue. I created this [colab notebook](https://colab.research.google.com/drive/16aLu6QrDSV_aUYRdpufl5E4iS08qkUGj?usp=sharing) to present the error. I tried using batch and that won't resolve it either. I'm looking into that error right now.

CakeCrusher:
Yes that is correct, the purpose of this pull is to advise of a more general solution like with `def remove_excess_nesting(self)` or maybe automate the solution (stas00 advised not to do that as it could "not be backwards compatible").

lhoestq:
I'm not sure I understand how having `remove_excess_nesting` would make more sense than just fixing the preprocessFunction to simply not return nested samples

CakeCrusher:
Figuring out the issue can be a bit difficult to figure out. Only until I added batch does it make a little more sense with the error

> sequence item 0: expected str instance, list found

but batch was never intended.

When you run the colab you will notice that only until collating do you learn there is this error. So i figured it would be better to address it during at the `DatasetDict` level.
I think it would be ideal if the user could be notified at the preprocess function.

CakeCrusher:
I'm not arguing that `remove_excess_nesting` is the right solution but what I aim to address is dealing with unnecessary nesting as early as possible.

lhoestq:
> When you run the colab you will notice that only until collating do you learn there is this error.

I think users can just check the `dataset.features` and they would notice that the data are nested
```python
{
   'input_ids': Sequence(Sequence(feature=Value(dtype='int32', id=None), length=-1, id=None), length=-1, id=None)
   ...
}
```

Sometime nested data are intentional, so you can't know in advance if it's a user's mistake or something planned.

CakeCrusher:
Yes, I understand, it could be intentional and only the collator has problems with it. So, it is not worth handling it any differently in any other non-erroneous data. 

That being said do you think there is any use for the `remove_excess_nesting` method? Or maybe it should be applied in a different way? If not feel free to close this PR.

lhoestq:
I think users can write it and use `map` themselves if needed, it is pretty straightforward to implement.

I'm closing this PR if you don't mind, and thank you for the discussion :)

CakeCrusher:
No problem @lhoestq , thanks for walking me through it.

PR #4683, opened 2022-07-15
Update create dataset card docs

stevhliu:
This PR proposes removing the [online dataset card creator](https://huggingface.co/datasets/card-creator/) in favor of simply copy/pasting a template and using the [Datasets Tagger app](https://huggingface.co/spaces/huggingface/datasets-tagging) to generate the tags. The Tagger app provides more guidance by showing all possible values a user can select in the dropdown menus, whereas the online dataset card creator doesn't, which can make it difficult to know what tag values to input.

Let me know what you think! :)

### DIFF
--- a/docs/source/dataset_card.mdx
+++ b/docs/source/dataset_card.mdx
@@ -1,25 +1,24 @@
 # Create a dataset card
 
-Each dataset should be accompanied with a Dataset card to promote responsible usage, and alert the user to any potential biases within the dataset.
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
Thanks ! I think it's better now that users can edit their dataset cards directly on hf.co easily. We don't maintain the online card creator, it's been contributed by someone super nice from the community

PR #4686, opened 2022-07-15
Align logging with Transformers (again)

### LINKED ISSUE #2832: Logging levels not taken into account

mariosasko:
Fix #2832

### DIFF
--- a/docs/source/package_reference/logging_methods.mdx
+++ b/docs/source/package_reference/logging_methods.mdx
@@ -50,6 +50,14 @@ By default, `tqdm` progress bars will be displayed during dataset download and p
 
 [[autodoc]] datasets.logging.get_logger
 
+[[autodoc]] datasets.logging.enable_default_handler
+
+[[autodoc]] datasets.logging.disable_default_handler
+
+[[autodoc]] datasets.logging.enable_explicit_format
+
+[[autodoc]] datasets.logging.reset_format
+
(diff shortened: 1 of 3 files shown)


mariosasko:
I wasn't aware of https://github.com/huggingface/datasets/pull/1845 before opening this PR. This issue seems much more complex now ...

PR #4672, opened 2022-07-11
Support extract 7-zip compressed data files

albertvillanova:
Fix partially #3541, fix #4670.

PR #4693, opened 2022-07-16
update `samsum` script

bhavitvyamalik:
update `samsum` script after #4672 was merged (citation is also updated)


albertvillanova (commented):
Thank you @bhavitvyamalik.

This change should definitely be done, but I think not yet: current integration of support to "7z" files does not support streaming mode. See e.g.:

```python
from datasets import load_dataset; ds = load_dataset("datasets/samsum", split="train", streaming=True); item = next(iter(ds)); item
```
gives
```
FileNotFoundError: https://huggingface.co/datasets/samsum/resolve/main/data/corpus.7z/train.json
```

IMO, support for streaming "7z" files should be addressed before merging this PR.

albertvillanova:
We are closing PRs to dataset scripts because we are moving them to the Hub.

Thanks anyway.

PR #4698, opened 2022-07-18
Enable streaming dataset to use the "all" split

### LINKED ISSUE #4637: The "all" split breaks streaming
## Describe the bug

Not sure if this is a bug or just the way streaming works, but setting `streaming=True` did not work when setting `split="all"`

## Steps to reproduce the bug
The following works:
```python
ds = load_dataset('super_glue', 'wsc.fixed', split='all')
```
The following throws `ValueError: Bad split: all. Available splits: ['train', 'validation', 'test']`:

```python
ds = load_dataset('super_glue', 'wsc.fixed', split='all', streaming=True)
```

## Expected results
An iterator over all splits.

## Actual results
I had to do the following to achieve the desired result:
```python
from itertools import chain
ds = load_dataset('super_glue', 'wsc.fixed', streaming=True)
it =  chain.from_iterable(ds.values())
```

## Environment info

- `datasets` version: 2.3.2
- Platform: Linux-4.15.0-176-generic-x86_64-with-glibc2.31
- Python version: 3.10.5
- PyArrow version: 8.0.0
- Pandas version: 1.4.3

cakiki:
Fixes #4637

### DIFF
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -48,7 +48,7 @@ from .download.streaming_download_manager import StreamingDownloadManager
 from .features import Features
 from .fingerprint import Hasher
 from .info import DatasetInfo, DatasetInfosDict, PostProcessedInfo
-from .iterable_dataset import ExamplesIterable, IterableDataset, _generate_examples_from_tables_wrapper
+from .iterable_dataset import ExamplesIterable, IterableDataset, _concatenate_iterable_datasets, _generate_examples_from_tables_wrapper
 from .keyhash import DuplicatedKeysError
 from .naming import INVALID_WINDOWS_CHARACTERS_IN_PATH, camelcase_to_snakecase
 from .splits import Split, SplitDict, SplitGenerator
@@ -1024,7 +1024,7 @@ class DatasetBuilder:
         self._check_manual_download(dl_manager)
         splits_generators = {sg.name: sg for sg in self._split_generators(dl_manager)}
         # By default, return all splits
-        if split is None:
+        if split is None or split == "all":
             splits_generator = splits_generators
         elif split in splits_generators:
             splits_generator = splits_generators[split]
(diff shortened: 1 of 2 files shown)


cakiki:
@albertvillanova 
Adding the validation split causes these two `assert_called_once` assertions to fail:
https://github.com/huggingface/datasets/blob/main/tests/test_builder.py#L548-L562

It might be better to create a new dummy generator for the streaming tests, WDYT?

albertvillanova:
@cakiki have you read my comment in the issue page?
https://github.com/huggingface/datasets/issues/4637#issuecomment-1175984812

cakiki:
Streaming with `split=all` seems to be working, will fix the failing test next

cakiki:
Not sure if marking the PR as "ready for review" actually notified you, so tagging @albertvillanova just in case :smiley_cat:

cakiki:
cc @lhoestq

src/datasets/builder.py
```diff
@@ -1039,6 +1044,11 @@ def as_streaming_dataset(
         )
         if isinstance(datasets, dict):
             datasets = IterableDatasetDict(datasets)
+            if split == "all":
+                datasets = _concatenate_iterable_datasets(
+                    datasets.values(), info=self.info, split="+".join(splits_generator), axis=0
+                )
```
lhoestq, line 1050: Nice thanks !

I think you can place this inside _as_streaming_dataset_single instead for consistency with as_dataset/_build_single_dataset

albertvillanova:
Hi @cakiki, still interested in working on this? :)

cakiki:
@albertvillanova So sorry; I have no idea how this slipped through the cracks. Yes, I'd still like to work on this. Is it okay if I DM you on slack?

albertvillanova:
Sure!! And nevermind!

PR #4706, opened 2022-07-18
Fix empty examples in xtreme dataset for bucc18 config

lhoestq:
As reported in https://huggingface.co/muibk, there are empty examples in xtreme/bucc18.de

I applied your fix @mustaszewski

I also used a dict to make the dataset generation much faster


albertvillanova:
I guess the report link is this instead: https://huggingface.co/datasets/xtreme/discussions/1

albertvillanova (approved):
Thanks @mustaszewski for reporting and fixing the issue of empty examples. We are giving credit to your contribution as co-author.
And thanks @lhoestq for the performance improvement and the fix of the tag in the dataset card.

PR #4714, opened 2022-07-19
Fix named split sorting and remove unnecessary casting

albertvillanova:
This PR:
- makes `NamedSplit` sortable: so that `sorted()` can be called on them
- removes unnecessary `sorted()` on `dict.keys()`: `dict_keys` view is already like a `set`
- removes unnecessary casting of `NamedSplit` to `str`

### DIFF
--- a/src/datasets/data_files.py
+++ b/src/datasets/data_files.py
@@ -14,7 +14,7 @@ from .utils.file_utils import hf_hub_url, is_relative_path, is_remote_url, reque
 from .utils.py_utils import string_to_dict
 
 
-DEFAULT_SPLIT = str(Split.TRAIN)
+DEFAULT_SPLIT = Split.TRAIN
 
 
 logger = logging.get_logger(__name__)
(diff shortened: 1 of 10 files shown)


src/datasets/data_files.py
```diff
@@ -14,7 +14,7 @@
 from .utils.py_utils import string_to_dict
 
 
-DEFAULT_SPLIT = str(Split.TRAIN)
+SANITIZED_DEFAULT_SPLIT = str(Split.TRAIN)
```
lhoestq, line 17: (nit) I think you can keep DEFAULT_SPLIT and call str() in sanitize_patterns, as sanitize_patterns already calls str() to the users requested splits.

lhoestq (approved):
sounds good to me !

albertvillanova:
Sorry, just too sync... :sweat_smile:

src/datasets/data_files.py
albertvillanova: @lhoestq that makes sense if `DEFAULT_SPLIT` is used anywhere else than in `sanitize_patterns`, but this is not currently the case.

PR #4718, opened 2022-07-19
Make Extractor accept Path as input

albertvillanova:
This PR:
- Makes `Extractor` accept instance of `Path` as input
- Removes unnecessary castings of `Path` to `str`


src/datasets/utils/extract.py
```diff
@@ -248,7 +248,7 @@ def infer_extractor_format(cls, path):
     @classmethod
     def extract(cls, input_path, output_path, extractor_format=None, extractor="deprecated"):
```
lhoestq, line 249: maybe add some typing to make it explicit that it can accept Path objects ?

lhoestq (approved):
thanks :)

PR #4724, opened 2022-07-20
Download and prepare as Parquet for cloud storage

lhoestq:
Download a dataset as Parquet in a cloud storage can be useful for streaming mode and to use with spark/dask/ray.

This PR adds support for `fsspec` URIs like `s3://...`, `gcs://...` etc. and ads the `file_format` to save as parquet instead of arrow:
```python
from datasets import *

cache_dir = "s3://..."
builder = load_dataset_builder("crime_and_punish", cache_dir=cache_dir)
builder.download_and_prepare(file_format="parquet")
```

EDIT: actually changed the API to

```python
from datasets import *

builder = load_dataset_builder("crime_and_punish")
builder.download_and_prepare("s3://...", file_format="parquet")
```

credentials to cloud storage can be passed using the `storage_options` argument in

For consistency with the BeamBasedBuilder, I name the parquet files `{builder.name}-{split}-xxxxx-of-xxxxx.parquet`. I think this is fine since we'll need to implement parquet sharding after this PR, so that a dataset can be used efficiently with dask for example.

Note that images/audio files are not embedded yet in the parquet files, this will added in a subsequent PR

TODO:
- [x] docs
- [x] tests


tests/fixtures/fsspec.py
```diff
@@ -0,0 +1,84 @@
+import posixpath
+from pathlib import Path
+
+import fsspec
+import pytest
+from fsspec.implementations.local import AbstractFileSystem, LocalFileSystem, stringify_path
+
+
+class MockFileSystem(AbstractFileSystem):
```
lhoestq, line 9: I created this mockfs fixture that is not only read but also **write**, this way we can do all our tests based on fsspec using this one :) @albertvillanova

src/datasets/utils/py_utils.py
```diff
@@ -188,7 +187,19 @@ def _asdict_inner(obj):
...
+@contextmanager
+def nullcontext():
+    """Context manager that does no additional processing.
+    Used as a stand-in for a normal context manager, when a particular
+    block of code is only sometimes used with a normal context manager:
+    cm = optional_cm if condition else nullcontext()
+    with cm:
+        # Perform operation, using optional_cm if condition is True
+    """
+    yield
```
mariosasko, line 199: 3.7 is a minimal supported Python version as of yesterday :), so this can be replaced with `contextlib.nullcontext`

mariosasko (commented):
Good job! Some nits:

src/datasets/builder.py
```diff
@@ -315,35 +321,37 @@ def __init__(
...
-            self._cache_dir_root if is_remote_url(self._cache_dir_root) else os.path.expanduser(self._cache_dir_root)
+
+        fs_token_paths = fsspec.get_fs_token_paths(
+            cache_dir or os.path.expanduser(config.HF_DATASETS_CACHE), storage_options=storage_options
         )
-        path_join = posixpath.join if is_remote_url(self._cache_dir_root) else os.path.join
+        self._fs: fsspec.AbstractFileSystem = fs_token_paths[0]
+
+        is_local = not is_remote_filesystem(self._fs)
+        path_join = os.path.join if is_local else os.path.join
```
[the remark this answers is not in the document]
lhoestq: thanks, good catch !

src/datasets/utils/py_utils.py
lhoestq: Nice, just did the change :)

src/datasets/builder.py
```diff
@@ -790,8 +806,9 @@ def _download_and_prepare(self, dl_manager, verify_infos, **prepare_split_kwargs
 
             try:
                 # Prepare split will record examples associated to the split
-                self._prepare_split(split_generator, **prepare_split_kwargs)
+                self._prepare_split(split_generator, file_format=file_format, **prepare_split_kwargs)
             except OSError as e:
+                raise
```
mariosasko, line 811: This was added for debugging? It makes the raise below unreachable

src/datasets/builder.py
```diff
@@ -654,19 +664,20 @@ def download_and_prepare(
...
+                self._fs.makedirs(tmp_dir, exist_ok=True)
+                try:
+                    yield tmp_dir
+                    if self._fs.isdir(dirname):
+                        self._fs.rm(dirname, recursive=True)
+                    if is_local:
+                        # LocalFileSystem.mv does copy + rm, it is more efficient to simply rename a local directory
+                        os.rename(self._fs._strip_protocol(tmp_dir), self._fs._strip_protocol(dirname))
+                    else:
+                        self._fs.mv(tmp_dir, dirname, recursive=True)
```
mariosasko, line 677: `mv` in `s3fs` and `hffs` (I expect these two filesystems + `gcsfs` to be the most used ones with this feature) reuploads the files, which is super expensive, so it would be great if we could replace this with some cheaper logic.

src/datasets/builder.py
lhoestq: yea sorry, forgot to remove it

src/datasets/builder.py
lhoestq: I don't see an easy way to rename the files without reuploading in the general case, I guess we can just not use the `incomplete_dir` for cloud storage, what do you think ?

mariosasko: Yes, I think this is the only option we have. (We just need to make sure we delete all the uploaded artifacts on error.)

mariosasko: `mv` calls `cp_file` and `cp_file` is also optimized in `s3fs` (it calls [this API](https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html)), so I think it's also ok to keep this as-is for now and instead we should optimize `cp_file` in `hffs`

lhoestq: oh that's good to know for s3fs, I'm down to keeping it as is then for simplicity, as we have a solution for hffs

lhoestq:
Added some docs for dask and took your comments into account

cc @philschmid if you also want to take a look :)

docs/source/filesystems.mdx
```diff
@@ -10,175 +12,187 @@
...
->>> dataset = load_from_disk('s3://my-private-datasets/imdb/train',fs=s3)  
+Load your own data files (see [how to load local and remote files](./loading#local-and-remote-files)):
 
->>> print(len(dataset))
->>> # 25000
+```py
+>>> data_files = {"train": ["path/to/train.csv"]}
+>>> cache_dir = "s3://my-bucket/datasets-cache"
+>>> builder = load_dataset_builder("csv", data_files=data_files, cache_dir=cache_dir, storage_options=storage_options)
+>>> builder.download_and_prepare(file_format="parquet")
```
philschmid, line 133: From a UX perspective it would be cool if i could do 
```python
# singe file
ds = load_dataset("s3://my-bucket/datasets-cache/train.csv",storage_options=storage_options)
# multi
ds = load_dataset("s3://my-bucket/datasets-cache",data_files={"train":["path/to/train.csv"]},storage_options=storage_options)
```
But not sure if this is possible to implement. Thats for example also how `pandas` is doing it 
`data = pd.read_csv('s3://bucket....csv')`

docs/source/filesystems.mdx
```diff
@@ -10,175 +12,187 @@
...
 
-# create GCSFileSystem instance using default gcloud credentials with project
->>> gcs = gcsfs.GCSFileSystem(project='my-google-project')
+After you have processed your dataset, you can save it to your cloud storage with [`Dataset.save_to_disk`]:
 
-# loads encoded_dataset from your gcs bucket
->>> dataset = load_from_disk('gcs://my-private-datasets/imdb/train', fs=gcs)
+```py
+# saves encoded_dataset to amazon s3
+>>> encoded_dataset.save_to_disk("s3://my-private-datasets/imdb/train", fs=fs)
```
philschmid, line 166: Similar to my comment above it would be nice to use the more "native" save methods, like `to_csv`, `to_parquet`etc... 
,e.g. 
```python
encoded_dataset.to_csv("s3://my-bucket/dataset.csv",storage_options=storage_options)
```
similar to pandas as well.

philschmid (commented):
I haven't reviewed the code yet but left some comments regarding the DX. I think to increase the usage of `datasets` within DS/Research (who are currently using other libraries because of better cloud storage integrations) we have to provide a similar API to what they are used to/datasets have from Hub datasets, e.g. loading datasets from cloud storage with `pandas` is similar to loading datasets with pandas from a local disk.

docs/source/filesystems.mdx
lhoestq: Yup this is definitely doable :)
It sounds super intuitive and practical to use, thanks !

lhoestq:
Just noticed that it would be more convenient to pass the output dir to download_and_prepare directly, to bypass the caching logic which prepares the dataset at `<cache_dir>/<name>/<version>/<hash>/`. And this way the cache is only used for the downloaded files. What do you think ?

philschmid:
Might be of interest: 
PyTorch and AWS introduced better support for S3 streaming in `torchtext`. 
image

lhoestq:
Alright I did the last change I wanted to do, here is the final API:

```python
builder = load_dataset_builder(...)
builder.download_and_prepare("s3://...", storage_options={"token": ...})
```

and it creates the arrow files directly in the specified directory, not in a nested subdirectory structure as we do in the cache !

> this approach still downloads the files needed to generate a dataset locally. Considering one of our goals is to align the streaming API with the non-streaming one, this could be avoided by running to_<format> on streamed/iterable datasets

Yup this can be explored in some future work I think. Though to keep things simple and clear I would keep the streaming behaviors only when you load a dataset in streaming mode, and not include it in `download_and_prepare` (because it wouldn't be aligned with the name of the function, which imply to 1. download and 2. prepare ^^). Maybe an API like that can make sense for those who need full streaming

```python
ds = load_dataset(..., streaming=True)
ds.to_parquet("s3://...")
```

docs/source/filesystems.mdx
```diff
@@ -10,175 +12,191 @@
...
+
+To use an anonymous connection, use `anon=True`.
+Otherwise, include your `aws_access_key_id` and `aws_secret_access_key` whenever you are interacting with a private S3 bucket.
 
 ```py
->>> import datasets
->>> s3 = datasets.filesystems.S3FileSystem(anon=True)  
->>> s3.ls('public-datasets/imdb/train')
-['dataset_info.json.json','dataset.arrow','state.json']
+>>> storage_options = {"anon": True}  # for anynonous connection
```
albertvillanova, line 34: Typo:
```suggestion
>>> storage_options = {"anon": True}  # for anonymous connection
```

docs/source/filesystems.mdx
```diff
@@ -10,175 +12,191 @@
...
+```py
+>>> data_files = {"train": ["path/to/train.csv"]}
+>>> output_dir = "s3://my-bucket/imdb"
+>>> builder = load_dataset_builder("csv", data_files=data_files)
+>>> builder.download_and_prepare(output_dir, storage_options=storage_options, file_format="parquet")
 ```
 
-## Google Cloud Storage
+It is highly recommended to save the files as compressed Parquet files to optimize I/O by specifying `file_format="parquet"`.
+Otherwize the dataset is saved as an uncompressed Arrow file.
```
albertvillanova, line 141: Typo:
```suggestion
Otherwise the dataset is saved as an uncompressed Arrow file.
```

albertvillanova (commented):
Thanks @lhoestq, awesome work!!! It is really important that we support multiple cloud storage.

Just a general comment about the documentation. Feel free to tell me if you don't agree.

I think the use of the word "loading" can be misleading: we are using it with different meanings:
- until now, "loading" was meaning "get" (load into memory or memory-map), contrary to "save" or "share"
- now we are using "loading"  as "saving": "loading into a cloud storage", differently from "load a dataset from the HF Hub"

I think we should be clear and make the difference between:
- loading a dataset FROM a cloud storage, with an API as suggested by @philschmid
  ```python
  ds = load_dataset("s3://my-bucket/datasets-cache/train.csv",storage_options=storage_options)
  ```
- and saving it TO a cloud storage, with your proposal

src/datasets/builder.py
```diff
@@ -314,36 +318,30 @@ def __init__(
...
-            with FileLock(lock_path):
-                if os.path.exists(self._cache_dir):  # check if data exist
-                    if len(os.listdir(self._cache_dir)) > 0:
-                        logger.info("Overwrite dataset info from restored data version.")
-                        self.info = DatasetInfo.from_directory(self._cache_dir)
-                    else:  # dir exists but no data, remove the empty dir as data aren't available anymore
-                        logger.warning(
-                            f"Old caching folder {self._cache_dir} for dataset {self.name} exists but not data were found. Removing it. "
-                        )
-                        os.rmdir(self._cache_dir)
```
albertvillanova, line 346: I see you are reverting the changes introduced by:
- #4347

Some concerns:
- What about if the user sets a remote `config.HF_DATASETS_CACHE` or `config.DOWNLOADED_DATASETS_DIR`?
- The `FileLock` was raising errors when working with a remote cache dir (BeamBasedBuilder)

albertvillanova (commented):
Just some additional comments below.

src/datasets/builder.py
lhoestq: Good catch thanks !

The FileLock is should only be applied when the output directory is a local directory indeed

lhoestq:
totally agree with your comment on the meaning of "loading", I'll update the docs

lhoestq:
I took your comments into account and reverted all the changes related to `cache_dir` to keep the support for remote `cache_dir` for beam datasets. I also updated the wording in the docs to not use "load" when it's not appropriate :)

mariosasko (approved):
Looks all good now! Thanks!

src/datasets/builder.py
```diff
@@ -570,18 +578,23 @@ def get_imported_module_dir(cls):
...
         use_auth_token: Optional[Union[bool, str]] = None,
+        file_format: Optional[str] = None,
+        storage_options: Optional[dict] = None,
         **download_and_prepare_kwargs,
     ):
         """Downloads and prepares dataset for reading.
 
         Args:
+            output_dir (:obj:`str`, optional): output directory for the dataset.
+                Default to this builder's ``cache_dir``, which is inside ~/.cache/huggingface/datasets by default.
```
albertvillanova, line 597: We might add the docs directive:
```
<Added version="2.5.0"/>
```

The same for the other new params in this docstring (`file_format` and `storage_options`, besides `output_dir`) and the `fs` params in the `DatasetInfo` docstring.

albertvillanova (approved):
Thanks @lhoestq, awesome job!

Just a few cosmetic comments on the docs...

PR #4726, opened 2022-07-20
Fix broken link to the Hub

stevhliu:
The Markdown link fails to render if it is in the same line as the `<span>`. This PR implements @mishig25's fix by using `<a href=" ">` instead.

Screen Shot 2022-07-20 at 3 53 05 PM


mishig25 (approved):
Lgtm!

albertvillanova (approved):
Thanks @stevhliu and @mishig25.

PR #4738, opened 2022-07-22
Use CI unit/integration tests

albertvillanova:
This PR:
- Implements separate unit/integration tests
- A fail in integration tests does not cancel the rest of the jobs
  - We should implement more robust integration tests: work in progress in a subsequent PR
- For the moment, test involving network requests are marked as integration: to be evolved

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,10 +35,11 @@ jobs:
   test:
     needs: check_code_quality
     strategy:
-      fail-fast: false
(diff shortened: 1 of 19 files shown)


albertvillanova:
I think this PR can be merged. Willing to see it in action.

CC: @lhoestq

lhoestq (approved):
Nice ! And this way we can also split the test dependencies to make the CI run even faster <3

PR #4740, opened 2022-07-25
Fix multiprocessing in map_nested 

### LINKED ISSUE #4636: Add info in docs about behavior of download_config.num_proc

I went to override `download_config.num_proc` and was confused about what was happening under the hood. It would be nice to have the behavior documented a bit better so folks know what's happening when they use it.

- Add note about how the default number of workers is 16. Related code:

https://github.com/huggingface/datasets/blob/7bcac0a6a0fc367cc068f184fa132b8de8dfa11d/src/datasets/download/download_manager.py#L299-L302

- Add note that if the number of workers is higher than the number of files to download, it won't use multiprocessing.

maybe it would also be nice to set `num_proc` = `num_files` when `num_proc` > `num_files`. 

**Additional context**

...

albertvillanova:
As previously discussed:

Before, multiprocessing was not used in `map_nested` if `num_proc` was greater than or equal to `len(iterable)`.
- Multiprocessing was not used e.g. when passing `num_proc=20` but having 19 files to download
- As by default, `DownloadManager` sets `num_proc=16`, before multiprocessing was only used when `len(iterable)>16` by default

Now, if `num_proc` is greater than or equal to ``len(iterable)``, `num_proc` is set to ``len(iterable)`` and multiprocessing is used.
- We pass the variable `multiprocessing_min_length=16`, so that multiprocessing is only used if at least 16 files to be downloaded
- ~As by default, `DownloadManager` sets `num_proc=16`, now multiprocessing is used when `len(iterable)>1` by default~

See discussion below.

~After having had to fix some tests (87602ac), I am wondering:~
- ~do we want to have multiprocessing by default?~
  - ~please note that `DownloadManager.download` sets `num_proc=16` by default~
- ~or would it be better to ask the user to set it explicitly if they want multiprocessing (and default to `num_proc=1`)?~

Fix #4636.

CC: @nateraw


src/datasets/utils/py_utils.py
```diff
@@ -352,12 +381,13 @@ def map_nested(
...
     if num_proc is None:
         num_proc = 1
-    if num_proc <= 1 or len(iterable) <= num_proc:
+    if num_proc <= 1 or len(iterable) <= 1:
         mapped = [
             _single_map_nested((function, obj, types, None, True, None))
             for obj in logging.tqdm(iterable, disable=disable_tqdm, desc=desc)
         ]
     else:
+        num_proc = num_proc if num_proc <= len(iterable) else len(iterable)
```
nateraw, line 390: Should we add a test for this logic? I'm not sure exactly how one would do that, but just a thought.

src/datasets/utils/py_utils.py
```diff
@@ -325,12 +325,41 @@ def map_nested(
...
+
+    Now, if `num_proc` is greater than or equal to ``len(iterable)``, `num_proc` is set to ``len(iterable)`` and
+    multiprocessing is used.
+
+    </Changed>
+
+    Args:
+        function (Callable): Function to be applied to `data_struct`.
+        data_struct: Data structure to apply `function` to.
+        dict_only (bool, default False): Whether only apply `function` recursively to `dict` values in `data_struct`.
```
nateraw, line 348: Nit - Maybe something like this for the args listed here? Not sure we have a standard across the package.

```suggestion
        dict_only (`bool`, default `False`): Whether only apply `function` recursively to `dict` values in `data_struct`.
```

Otherwise I would put "Defaults to `False` a the end of the description itself and leave the annotation as (`bool`).

nateraw (commented):
This is awesome, thank you!! The docstring is a very welcome addition.

Unrelated, but this PR makes me think that we have wide variety of different docstring formats across the repo.

src/datasets/utils/py_utils.py
albertvillanova: Definitely, you are right. I didn't pay too much attention to the syntax because this docstring is not published in our docs. But I agree it is better to do it right in case it will be published some day. I'm fixing it.

tests/test_dataset_common.py
```diff
@@ -322,7 +322,9 @@ def test_builder_configs(self, dataset_name):
 
 def distributed_load_dataset(args):
     data_name, tmp_dir, datafiles = args
-    dataset = load_dataset(data_name, cache_dir=tmp_dir, data_files=datafiles)
+    # Disable multiprocessing in download to avoid AssertionError: daemonic processes are not allowed to have children
+    download_config = DownloadConfig(num_proc=1)
+    dataset = load_dataset(data_name, cache_dir=tmp_dir, data_files=datafiles, download_config=download_config)
```
lhoestq, line 327: I think this could break some example scripts in `transformers` that use pytorch DDP. Switching to multithreading could help maybe, which is more appropriate to download files anyway.

src/datasets/utils/py_utils.py
albertvillanova: Tests added.

tests/test_dataset_common.py
albertvillanova: @lhoestq the intention of this PR was not to change the downloading mechanism, but to address the request by @nateraw:
```
nateraw
Maybe it would be nice if num_proc > num_files then you set num_proc = num_files.

Quentin Lhoest
Yep indeed
```

The specific use case was that when passing `num_proc=20` but having 19 files to download, multiprocessing was not used (which is counter-intuitive).

From the user point of view, I have raised my concerns in this PR description: https://github.com/huggingface/datasets/pull/4740#issue-1316478007

lhoestq: I see, I'm just a bit concerned with the regression it might cause for users that use multiprocessing already e.g. to train models.

albertvillanova: @lhoestq I agree this is a breaking change, thus my doubts as well, as explained at the end of the PR description: https://github.com/huggingface/datasets/pull/4740#issue-1316478007

My usual preference would have been being explicit (as the Zen of Python `Explicit is better than implicit.`), thus I would have recommended not doing multiprocessing unless explicitly required by the user.

However, given that this implicit multiprocessing is already in place (when more than 16 files to download), IMO is preferable to keep backward compatibility (thus not introducing breaking changes) as much as possible.

I'm thinking of a workaround...

albertvillanova:
@lhoestq as a workaround to preserve previous behavior, the parameter `multiprocessing_min_length=16` is passed from `download` to `map_nested`, so that multiprocessing is only used if at least 16 files to be downloaded.

Note that there is a small breaking change (I think previously it was unintended behavior, so that I have fixed it):
- Before (with default `num_proc=16`) if there were 16 files to be downloaded, multiprocessing was not used
- Now (with default `num_proc=16`) if there are 16 files to be downloaded, multiprocessing is used

tests/test_dataset_common.py
lhoestq: > My usual preference would have been being explicit (as the Zen of Python Explicit is better than implicit.), thus I would have recommended not doing multiprocessing unless explicitly required by the user.

100% agree, though using multithreading to download files in parallel is something that doesn't need user's permission IMO. Multiprocessing has some side effects so I agree it's better if users ask for it explicitly. That's why I think we should switch to multithreading at one point

lhoestq:
Thanks for the workaround !

src/datasets/utils/py_utils.py
```diff
@@ -318,19 +318,52 @@ def _single_map_nested(args):
...
+        dict_only (`bool`, default `False`): Whether only apply `function` recursively to `dict` values in
+            `data_struct`.
+        map_list (`bool`, default `True`): Whether also apply `function` recursively to `list` elements (besides `dict`
+            values).
+        map_tuple (`bool`, default `False`): Whether also apply `function` recursively to `tuple` elements (besides
+            `dict` values).
+        map_numpy (`bool, default `False`): Whether also apply `function` recursively to `numpy.array` elements (besides
+            `dict` values).
+        num_proc (`int`, *optional*): Number of processes.
+        multiprocessing_min_length (`int`, default `2`): Minimum length of `data_struct` required for multiprocessing.
```
lhoestq, line 359: Maybe name it `min_length_for_parallelism` in case we ever want to add support for multithreading in map_nested ?

lhoestq (approved):
Thanks !

src/datasets/utils/py_utils.py
albertvillanova: @lhoestq I renamed it to `parallel_min_length` so that its naming aligns with other variable names:
- `IN_MEMORY_MAX_SIZE`
- `STREAMING_READ_MAX_RETRIES`

I just replaced `parallelism` with `parallel` to make it simpler. Do you think it is OK? Otherwise it would have been `parallelism_min_length`.

lhoestq: Yup that's perfect IMO, thanks !

PR #4747, opened 2022-07-26
Shard parquet in `download_and_prepare`

lhoestq:
Following https://github.com/huggingface/datasets/pull/4724 (needs to be merged first)

It's good practice to shard parquet files to enable parallelism with spark/dask/etc.

I added the `max_shard_size` parameter to `download_and_prepare` (default to 500MB for parquet, and None for arrow).

```python
from datasets import *

output_dir = "./output_dir"  # also supports "s3://..."
builder = load_dataset_builder("squad")
builder.download_and_prepare(output_dir, file_format="parquet", max_shard_size="5MB")
```

### Implementation details

The examples are written to a parquet file until `ParquetWriter._num_bytes > max_shard_size`. When this happens, a new writer is instantiated to start writing the next shard. At the end, all the shards are renamed to include the total number of shards in their names: `{builder.name}-{split}-{shard_id:05d}-of-{num_shards:05d}.parquet`

I also added the `MAX_SHARD_SIZE` config variable (default to 500MB)

TODO:
- [x] docstrings
- [x] docs
- [x] tests

cc @severo


lhoestq:
This is ready for review cc @mariosasko let me know what you think

src/datasets/builder.py
```diff
@@ -1251,53 +1318,97 @@ def _generate_examples(self, **kwargs):
...
+        try:
+            for key, record in logging.tqdm(
+                generator,
+                unit=" examples",
+                total=split_info.num_examples,
+                leave=False,
+                disable=not logging.is_progress_bar_enabled(),
+                desc=f"Generating {split_info.name} split",
+            ):
+                if max_shard_size is not None and writer._num_bytes > max_shard_size:
```
mariosasko, line 1372: The final shard size can easily be off (significantly) using this logic, no? By default, `writer._num_bytes` is only updated every 10k examples (`WRITER_BATCH_SIZE`), so if the `max_shard_size` is small...

mariosasko (commented):
Thanks!

src/datasets/builder.py
lhoestq: I think right now this is the builder's responsibility to specify if 10,000 is too much (`DEFAULT_WRITER_BATCH_SIZE` class attribute).

Though I agree this is not ideal. I think it's ok to have it this way in this PR (since it would be off if the max_shard_size is very very small), but it would be nice to have something smarter in general

src/datasets/builder.py
```diff
@@ -1496,12 +1653,19 @@ def _save_info(self):
...
+    def _prepare_split(
+        self, split_generator, pipeline, file_format="arrow", max_shard_size: Optional[Union[str, int]] = None
+    ):
         import apache_beam as beam
 
+        if max_shard_size is not None:
+            raise NotImplementedError(
+                "max_shard_size is not supported for Beam datasets, please."
+                "Set it to None to use the default Apache Beam sharding and get the best performance."
+            )
```
mariosasko, line 1665:
```suggestion
            raise NotImplementedError(
                "max_shard_size is not supported for Beam datasets."
                "Please set it to None to use the default Apache Beam sharding and get the best performance."
            )
```

mariosasko (approved):
One nit. Besides that, all looks good!

PR #4748, opened 2022-07-27
Add image classification processing guide

stevhliu:
This PR follows up on #4710 to separate the object detection and image classification guides. It expands a little more on the original guide to include a more complete example of loading and transforming a whole dataset.

### DIFF
--- a/docs/source/_toctree.yml
+++ b/docs/source/_toctree.yml
@@ -56,6 +56,10 @@
       title: Load image data
     - local: image_process
       title: Process image data
+    - local: image_classification
+      title: Image classification
+    - local: object_detection
+      title: Object detection
     title: "Vision"
   - sections:
     - local: nlp_load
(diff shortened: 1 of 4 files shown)


lhoestq (approved):
Thanks !

docs/source/_toctree.yml
```diff
@@ -56,6 +56,10 @@
       title: Load image data
     - local: image_process
       title: Process image data
+    - local: image_classification
+      title: Image classification
+    - local: object_detection
+      title: Object detection
```
mariosasko, line 62: Perhaps a table at the end of the `image_process` page that redirects to a specific task (processing) would be better than having each task in the index to keep the index small.

mariosasko (commented):
Thanks! One nit from me:

docs/source/_toctree.yml
lhoestq: Small pref to have them in the index, this is the kind of things that users should be able to find in very few clicks IMO

stevhliu: I think keeping it in the index would be more discoverable for users since not everyone may necessarily read the `image_process` page :)

PR #4762, opened 2022-07-28
Improve features resolution in streaming

lhoestq:
`IterableDataset._resolve_features` was returning the features sorted alphabetically by column name, which is not consistent with non-streaming. I changed this and used the order of columns from the data themselves. It was causing some inconsistencies in the dataset viewer as well.

I also fixed `interleave_datasets` that was not filling missing columns with None, because it was not using the columns from `IterableDataset._resolve_features`

cc @severo


src/datasets/iterable_dataset.py
```diff
@@ -1350,8 +1351,10 @@ def _concatenate_iterable_datasets(
     else:
         _check_column_names([col_name for dset in dsets for col_name in dset.features])
 
+    # TODO: improve this to account for a mix of ClassLabel and Value for example
+    # right now it would keep the type of the first dataset in the list
     features = Features()
-    for dset in dsets:
+    for dset in dsets[::-1]:
```
mariosasko, line 1357: We should use `_align_features` (defined in `arrow_dataset.py`) here to keep this logic aligned with the arrow-based dataset.

src/datasets/iterable_dataset.py
```diff
@@ -1396,13 +1399,19 @@ def _interleave_iterable_datasets(
...
+    datasets = [d._resolve_features() for d in datasets]
+
+    # Perform checks
+    _check_if_features_can_be_aligned([dset.features for dset in datasets])
+
+    # TODO: improve this to account for a mix of ClassLabel and Value for example
+    # right now it would keep the type of the first dataset in the list
+    features = Features()
+    for dset in datasets[::-1]:
+        features.update(dset.features)
```
mariosasko, line 1411: Same here

mariosasko (commented):
Thanks! One nit:

lhoestq:
Just took your comment into account @mariosasko , let me know if it's good for you now :)

src/datasets/iterable_dataset.py
```diff
@@ -9,7 +9,7 @@
 import numpy as np
 import pyarrow as pa
 
-from .arrow_dataset import DatasetInfoMixin
+from .arrow_dataset import DatasetInfoMixin, _align_features
```
mariosasko, line 12: I think we can move `_align_features` (and `_check_if_features_can_be_aligned`) from `iterable_dataset.py`/`arrow_dataset.py` to `features.py` to have everything in one place

src/datasets/iterable_dataset.py
```diff
@@ -1350,9 +1351,11 @@ def _concatenate_iterable_datasets(
...
         _check_column_names([col_name for dset in dsets for col_name in dset.features])
 
-    features = Features()
-    for dset in dsets:
-        features.update(dset.features)
+    # TODO: improve this to account for a mix of ClassLabel and Value for example
+    # right now it would keep the type of the first dataset in the list
+    features = Features(
+        {k: v for features in _align_features([dset.features for dset in dsets[::-1]]) for k, v in features.items()}
+    )
```
mariosasko, line 1358: I believe `[::-1]` is not needed here since we use `_check_if_features_can_be_aligned` + `_align_features`:
```suggestion
    features = Features(
        {k: v for features in _align_features([dset.features for dset in dsets]) for k, v in features.items()}
    )
```

src/datasets/iterable_dataset.py
```diff
@@ -1396,13 +1399,19 @@ def _interleave_iterable_datasets(
...
+    datasets = [d._resolve_features() for d in datasets]
+
+    # Perform checks
+    _check_if_features_can_be_aligned([dset.features for dset in datasets])
+
+    # TODO: improve this to account for a mix of ClassLabel and Value for example
+    # right now it would keep the type of the first dataset in the list
+    features = Features(
+        {k: v for features in _align_features([dset.features for dset in datasets[::-1]]) for k, v in features.items()}
+    )
```
mariosasko, line 1411: Same here:

```suggestion
    features = Features(
        {k: v for features in _align_features([dset.features for dset in datasets]) for k, v in features.items()}
    )
```

mariosasko (approved):
Some nits. Other than that, looks good!

PR #4763, opened 2022-07-28
More rigorous shape inference in to_tf_dataset

Rocketknight1:
`tf.data` needs to know the shape of tensors emitted from a `tf.data.Dataset`. Although `None` dimensions are possible, overusing them can cause problems - Keras uses the dataset tensor spec at compile-time, and so saying that a dimension is `None` when it's actually constant can hurt performance, or even cause training to fail for dimensions that are needed to determine the shape of weight tensors!

The compromise I used here was to sample several batches from the underlying dataset and apply the `collate_fn` to them, and then to see which dimensions were "empirically variable". There's an obvious problem here, though - if you sample 10 batches and they all have the same shape on a certain dimension, there's still a small chance that the 11th batch will be different, and Keras will throw an error  if a dataset tries to emit a tensor whose shape doesn't match the spec.

I encountered this bug in practice once or twice for datasets that were mostly-but-not-totally constant on a given dimension, and I still don't have a perfect solution, but this PR should greatly reduce the risk. It samples many more batches, and also samples very small batches (size 2) - this increases the variability, making it more likely that a few outlier samples will be detected.

Ideally, of course, we'd determine the full output shape analytically, but that's surprisingly tricky when the `collate_fn` can be any arbitrary Python code!

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -299,24 +299,17 @@ class TensorflowDatasetMixin:
                     f"Unrecognized array dtype {np_arrays[0].dtype}. \n"
                     "Nested types and image/audio types are not supported yet."
                 )
-            if (
-                column in dataset
-                and isinstance(dataset.features[column], Sequence)
-                and dataset.features[column].length != -1
-            ):
-                static_shape = [batch_size, dataset.features[column].length]
-            else:
-                shapes = [array.shape for array in np_arrays]
-                static_shape = []
-                for dim in range(len(shapes[0])):
-                    sizes = set([shape[dim] for shape in shapes])
-                    if dim == 0:
-                        static_shape.append(batch_size)
-                        continue
-                    if len(sizes) == 1:  # This dimension looks constant
-                        static_shape.append(sizes.pop())
-                    else:  # Use None for variable dimensions
-                        static_shape.append(None)
+            shapes = [array.shape for array in np_arrays]
+            static_shape = []
+            for dim in range(len(shapes[0])):
+                sizes = set([shape[dim] for shape in shapes])
+                if dim == 0:
+                    static_shape.append(batch_size)
+                    continue
+                if len(sizes) == 1:  # This dimension looks constant
+                    static_shape.append(sizes.pop())
+                else:  # Use None for variable dimensions
+                    static_shape.append(None)
             tf_columns_to_signatures[column] = tf.TensorSpec(shape=static_shape, dtype=tf_dtype)
             np_columns_to_dtypes[column] = np_dtype
 
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -224,7 +224,7 @@ class TensorflowDatasetMixin:
         collate_fn_args: dict,
         cols_to_retain: Optional[List[str]] = None,
         batch_size: Optional[int] = None,
-        num_test_batches: int = 10,
+        num_test_batches: int = 200,
     ):
         """Private method used by `to_tf_dataset()` to find the shapes and dtypes of samples from this dataset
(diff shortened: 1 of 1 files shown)


src/datasets/arrow_dataset.py
```diff
@@ -420,6 +420,26 @@ def to_tf_dataset(
             batch_size=batch_size if drop_remainder else None,
         )
 
+        shape_verification_signature, _ = dataset._get_output_signature(
```
lhoestq, line 423: why do you need to call it a second time ? can't this logic be inside _get_output_signature ?

Rocketknight1: That would make sense, actually! I'll move it.

Rocketknight1: @lhoestq I cleaned things up a lot based on your feedback - `_get_output_signature` is only called once, and it now immediately samples 200 batches of size 2 to infer the shape, but then overwrites the batch size element of the inferred shape with the actual batch size.

lhoestq: Cool! :)

I also think 10 batches is good by default, going to 200 batches can take too much time for some datasets IMO

Rocketknight1: I actually specifically had problems with incorrect inferences when using 10! I think it's preferable for `to_tf_dataset()` to be a little slow sometimes (it's only called once at dataset creation time) than to infer wrong shapes and create tricky bugs for users.

If you want, though, I can make `num_test_batches` an argument to `to_tf_dataset`?

gante (approved):
LGTM 👍

src/datasets/arrow_dataset.py
Rocketknight1: Gentle ping @lhoestq!

Rocketknight1: Pinging @albertvillanova instead because Quentin is on holiday and I didn't realize >.>;

lhoestq: > I actually specifically had problems with incorrect inferences when using 10!

Can you explain what problems ?

Rocketknight1: In some cases, sampling 10 batches from the dataset makes it look like the dataset has a constant shape, but actually it doesn't. This is particularly common when datasets have been truncated. For example, if the average length in a dataset before truncation is >> 512, but we truncate at 512, then most batches will have length 512, but if some samples in the dataset have length < 512, then there will occasionally be batches with length < 512 too.

By reducing the batch size for shape inference and increasing the number of batches sampled, this problem is resolved in all the cases I know about!

lhoestq: What about adding a way for users to specify if the shapes are fixed or not ? Could be via a new parameter, or by checking if the feature type is `Sequence(..., length=512)`

Rocketknight1: I think that's a good idea! We'll still need shape inference but it might be useful, and I can look into adding it when I get back!

src/datasets/arrow_dataset.py
```diff
@@ -301,17 +299,24 @@ def _get_output_signature(
...
-                if len(sizes) == 1:  # This dimension looks constant
-                    static_shape.append(sizes.pop())
-                else:  # Use None for variable dimensions
-                    static_shape.append(None)
+            if (
+                column in dataset
+                and isinstance(dataset.features[column], Sequence)
+                and dataset.features[column].length != -1
+            ):
+                static_shape = [batch_size, dataset.features[column].length]
```
lhoestq, line 307: I think this is not needed if we do shape inference. My comment about `Sequence` was just an idea if shape inference was not here, sorry for the confusion

You can remove this part

Rocketknight1: Alright!

PR #4770, opened 2022-07-29
fix typo

xwwwwww:
By defaul -> By default


lhoestq:
good catch thanks ! Can you check if the same typo is also present in `add_elasticsearch_index` ? It has a very similar signature

xwwwwww:
> good catch thanks ! Can you check if the same typo is also present in `add_elasticsearch_index` ? It has a very similar signature

fixed

lhoestq (approved):
Thanks !

Issue #4757, opened 2022-07-28
Document better when relative paths are transformed to URLs


albertvillanova:
As discussed with @ydshieh, when passing a relative path as `data_dir` to `load_dataset` of a dataset hosted on the Hub, the relative path is transformed to the corresponding URL of the Hub dataset.

Currently, we mention this in our docs here: [Create a dataset loading script > Download data files and organize splits](https://huggingface.co/docs/datasets/v2.4.0/en/dataset_script#download-data-files-and-organize-splits)
> If the data files live in the same folder or repository of the dataset script, you can just pass the relative paths to the files instead of URLs.

Maybe we should document better how relative paths are handled, not only when creating a dataset loading script, but also when passing to `load_dataset`:
- `data_dir`
- `data_files`

CC: @stevhliu

PR #4773, opened 2022-07-29
Document loading from relative path

stevhliu:
This PR describes loading a dataset from the Hub by specifying a relative path in `data_dir` or `data_files` in `load_dataset` (see #4757).

### DIFF
--- a/docs/source/loading.mdx
+++ b/docs/source/loading.mdx
@@ -54,19 +54,13 @@ If you don't specify which data files to use, [`load_dataset`] will return all t
 
 </Tip>
 
-You can also load a specific subset of the files with the `data_files` or `data_dir` parameter. Datasets are loaded from the Hub if you pass a relative path to `data_dir`. The relative path is transformed to a URL corresponding to where the dataset is loaded from. If you want to load a dataset from a local directory, you need to pass an absolute path instead:
+You can also load a specific subset of the files with the `data_files` or `data_dir` parameter. These parameters can accept a relative path which resolves to the base path corresponding to where the dataset is loaded from.
 
 ```py
 >>> from datasets import load_dataset
 
 # load files that match the grep pattern
 >>> c4_subset = load_dataset("allenai/c4", data_files="en/c4-train.0000*-of-01024.json.gz")
-
-# load dataset from the en directory on the Hub
->>> c4_subset = load_dataset("allenai/c4", data_dir="./en/")
-
-# load dataset from the en directory on your computer
->>> c4_subset = load_dataset("allenai/c4", data_dir="path/to/local/allenai/c4/en/")
 ```
 
 The `split` parameter can also map a data file to a specific split:
@@ -86,13 +80,6 @@ You may have a 🤗 Datasets loading script locally on your computer. In this ca
 ```py
 >>> dataset = load_dataset("path/to/local/loading_script/loading_script.py", split="train")
 >>> dataset = load_dataset("path/to/local/loading_script", split="train")  # equivalent because the file has the same name as the directory
-```
-
-You can load a specific subset of files with the `data_files` or `data_dir` parameter. Passing a relative path to these parameters will load a dataset from your local directory:
-
-```py
->>> from datasets import load_dataset
->>> c4_subset = load_dataset("path/to/local/loading_script", data_dir="./train-v1/")
 ```
 
 ## Local and remote files
--- a/docs/source/loading.mdx
+++ b/docs/source/loading.mdx
@@ -54,11 +54,19 @@ If you don't specify which data files to use, [`load_dataset`] will return all t
 
 </Tip>
 
-You can also load a specific subset of the files with the `data_files` parameter. The example below only loads files that match the grep pattern:
+You can also load a specific subset of the files with the `data_files` or `data_dir` parameter. Datasets are loaded from the Hub if you pass a relative path to `data_dir`. The relative path is transformed to a URL corresponding to where the dataset is loaded from. If you want to load a dataset from a local directory, you need to pass an absolute path instead:
 
 ```py
 >>> from datasets import load_dataset
+
+# load files that match the grep pattern
 >>> c4_subset = load_dataset("allenai/c4", data_files="en/c4-train.0000*-of-01024.json.gz")
+
+# load dataset from the en directory on the Hub
+>>> c4_subset = load_dataset("allenai/c4", data_dir="./en/")
+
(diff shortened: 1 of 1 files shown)


docs/source/loading.mdx
```diff
@@ -68,6 +68,13 @@ The `split` parameter can also map a data file to a specific split:
 >>> c4_validation = load_dataset("allenai/c4", data_files=data_files, split="validation")
 ```
 
+Datasets hosted on the Hub can also be loaded from a relative path through the `data_dir` or `data_files` parameters. The dataset's relative path is resolved to its URL in the repository where the loading script is stored.
```
lhoestq, line 71: Thanks !

I think I would simply mention `data_dir` in the previosu paragraph:
```
You can also load a specific subset of the files with the `data_files` or `data_dir` parameter.
```

This way it is more explicit that they work the same way.

The example is since as well :)

albertvillanova: I agree this sentence here may lead to confusion:
- the sentence mentions a repository with a loading script
- but this section is about Hub datasets without loading script

albertvillanova (changes requested):
Thanks for addressing this issue, @stevhliu.

However, I think more explanation is needed. The original issue reported by @ydshieh was that:
- he wanted to use a Hub dataset containing a script: "ydshieh/coco_dataset_script"
- but he wanted to pass as `data_dir` a local directory on his computer (not a directory on the repo)

```python
ds = datasets.load_dataset("ydshieh/coco_dataset_script", "2017", data_dir="./local_real_data/")
```

However, this does not use the "./local_real_data" on his computer, but instead tries to look for this directory on the Hub repo.

He had to change the `data_dir` from relative to absolute, so that his computer local directory was considered.

The behavior of `load_dataset` is:
- if you pass a repo ID (`load_dataset("repo_id"...`), then `data_dir` and `data_files`, if they are relative, are transformed to URLs corresponding to the repo ID URL 
  - if you pass absolute paths instead, local paths (in the calling computer) are used
- if you pass a local path instead of repo ID (`load_dataset("local/path"...`), then relative `data_dir` and `data_files` are resolved locally in the calling computer

As @lhoestq mentioned in the discussion:
> basically the base path used to resolve relative path corresponds to where the dataset is loaded from

docs/source/loading.mdx
```diff
@@ -54,11 +54,19 @@ If you don't specify which data files to use, [`load_dataset`] will return all t
 
 </Tip>
 
-You can also load a specific subset of the files with the `data_files` parameter. The example below only loads files that match the grep pattern:
+You can also load a specific subset of the files with the `data_files` or `data_dir` parameter. Datasets are loaded from the Hub if you pass a relative path to `data_dir`. The relative path is transformed to a URL corresponding to where the dataset is loaded from. If you want to load a dataset from a local directory, you need to pass an absolute path instead:
```
albertvillanova, line 57: I think this may need some rephrasing: I would say it makes not much sense to pass an absolute local directory to load a Hub dataset *without a loading script*: I think this specific use case is only applicable to Hub datasets *with a loading script*.

I mean, passing `load_dataset("allenai/c4",...` just analyzes all the files in the repo to see that they are mainly JSON files and then choose the "json" builder.

Therefore, 
```python
load_dataset("allenai/c4", data_dir="/absolute/path/to/local/allenai/c4/en")
```
is equivalent to:
```python
load_dataset("json", data_dir="/absolute/path/to/local/allenai/c4/en")
```

albertvillanova (commented):
I'm not sure if we should talk about absolute local directories/files when explaining how to load a Hub dataset without a script...

IMO, this would be applicable to these cases:
- Hub dataset without script
  - data_dir/data_files relative paths to data files in the corresponding Hub dataset repo
- Hub dataset with script
  - data_dir/data_files RELATIVE paths to data files in the corresponding Hub dataset repo
  - data_dir/data_files ABSOLUTE paths to data files in the local computer
- local dataset with script
  - data_dir/data_files relative paths to data files in the local computer


What do you think?

docs/source/loading.mdx
```diff
@@ -54,19 +54,13 @@ If you don't specify which data files to use, [`load_dataset`] will return all t
...
+You can also load a specific subset of the files with the `data_files` or `data_dir` parameter. These parameters can accept a relative path which resolves to the base path corresponding to where the dataset is loaded from.
 
 ```py
 
 # load files that match the grep pattern
 >>> c4_subset = load_dataset("allenai/c4", data_files="en/c4-train.0000*-of-01024.json.gz")
-
-# load dataset from the en directory on the Hub
->>> c4_subset = load_dataset("allenai/c4", data_dir="./en/")
```
lhoestq: I think you can keep this one 
```python
c4_subset = load_dataset("allenai/c4", data_dir="en")
```

lhoestq (approved):
Thanks ! Sounds good to me

albertvillanova (commented):
Thanks again @stevhliu for all your changes after so many iteration and sorry for my late feedback.

I think it is clear now the use of relative paths when loading a Hub dataset without a script.

However, I'm afraid we have not clarified the issue raised by @ydshieh (see comment https://github.com/huggingface/datasets/pull/4773#pullrequestreview-1057120994): difference between passing a relative or an absolute path as data_dir/data_files when loading a Hub dataset **with a script**.

Hub datasets **with a loading script** are not discussed in this page of the docs (`loading.mdx`). They are discussed in `load_hub.mdx`, although from quite a general perspective. Do you think an explanation about this issue on that docs file would make sense? Maybe in an additional section or in a Tip box? Or maybe this issue is not general enough to fit on that doc page...

Honestly, I'm not sure... @lhoestq any clever advice? :)

stevhliu:
Thanks for the feedback!

I agree that adding it to `load_hub.mdx` is probably a bit too specific, especially for beginners reading the tutorials. Since this clarification is closely related to loading from the Hub (the only difference being the presence/absence of a loading script), I think it makes the most sense to keep it somewhere in `loading.mdx`. What do you think about adding a Warning in Loading >>> Hugging Face Hub that explains the difference between relative/absolute paths when there is a script?

lhoestq:
What about updating the section about "manual download" ? I think it does there no ?

https://huggingface.co/docs/datasets/v2.4.0/en/loading#manual-download

stevhliu:
Updated the manual download section :)

lhoestq:
Thanks ! Pinging @albertvillanova to review this change, and then I think we're good to merge

PR #4778, opened 2022-08-01
Update local loading script docs

### LINKED ISSUE #4732: Document better that loading a dataset passing its name does not use the local script
As reported by @TrentBrick here https://github.com/huggingface/datasets/issues/4725#issuecomment-1191858596, it could be more clear that loading a dataset by passing its name does not use the (modified) local script of it.

What he did:
- he installed `datasets` from source
- he modified locally `datasets/the_pile/the_pile.py` loading script
- he tried to load it but using `load_dataset("the_pile")` instead of `load_dataset("datasets/the_pile")`
  - as explained here https://github.com/huggingface/datasets/issues/4725#issuecomment-1191040245:
    - the former does not use the local script, but instead it downloads a copy of `the_pile.py` from our GitHub, caches it locally (inside `~/.cache/huggingface/modules`) and uses that.

He suggests adding a more clear explanation about this. He suggests adding it maybe in [Installation > source](https://huggingface.co/docs/datasets/installation))

CC: @stevhliu

stevhliu:
This PR clarifies the local loading script section to include how to load a dataset after you've modified the local loading script (closes #4732).

### DIFF
--- a/docs/source/loading.mdx
+++ b/docs/source/loading.mdx
@@ -92,7 +92,7 @@ Make your edits to the loading script and then load it by passing its local path
 
 ```py
 >>> from datasets import load_dataset
->>> eli5 = load_dataset("/local/path/to/datasets/eli5")
+>>> eli5 = load_dataset("path/to/local/eli5")
 ```
 
 ## Local and remote files
--- a/docs/source/loading.mdx
+++ b/docs/source/loading.mdx
@@ -80,6 +80,14 @@ You may have a 🤗 Datasets loading script locally on your computer. In this ca
 >>> dataset = load_dataset("path/to/local/loading_script", split="train")  # equivalent because the file has the same name as the directory
 ```
 
+The [`~Dataset.load_dataset`] function pulls a loading script from GitHub or the Hub unless you give it a path to the script. This means if you edit a local loading script and go to load the dataset, you need to load the dataset from its local path; otherwise, it'll just download and use a copy of the loading script from GitHub. Prefix the dataset name in your [`~Dataset.load_dataset`] call with `datasets` to use the local GitHub dataset script stored inside the `datasets` directory:
+
+```py
+>>> from datasets import load_dataset
+# edit local loading script and now load the dataset
+>>> eli5 = load_dataset("datasets/eli5")
+```
+
 ## Local and remote files
 
 Datasets can be loaded from local files stored on your computer and from remote files. The datasets are most likely stored as a `csv`, `json`, `txt` or `parquet` file. The [`load_dataset`] function can load each of these file types.
--- a/docs/source/loading.mdx
+++ b/docs/source/loading.mdx
@@ -82,11 +82,10 @@ You may have a 🤗 Datasets loading script locally on your computer. In this ca
 
 ### Edit loading script
 
-You can also edit a loading script from the Hub to add your own modifications. Use the [`~datasets.inspect_dataset`] function to download and inspect a loading script:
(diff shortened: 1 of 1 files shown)


mariosasko:
I would rather have a section in the docs that explains how to modify the script of an existing dataset (`inspect_dataset` + modification + `load_dataset`) instead of focusing on the GH datasets bundled with the source (only applicable for devs).

docs/source/loading.mdx
```diff
@@ -80,6 +80,14 @@ You may have a 🤗 Datasets loading script locally on your computer. In this ca
 >>> dataset = load_dataset("path/to/local/loading_script", split="train")  # equivalent because the file has the same name as the directory
 ```
 
+The [`~Dataset.load_dataset`] function pulls a loading script from GitHub or the Hub unless you give it a path to the script. This means if you edit a local loading script and go to load the dataset, you need to load the dataset from its local path; otherwise, it'll just download and use a copy of the loading script from GitHub. Prefix the dataset name in your [`~Dataset.load_dataset`] call with `datasets` to use the local GitHub dataset script stored inside the `datasets` directory:
```
lhoestq, line 83:
````suggestion
## Edit an existing script

To edit a dataset, first download the dataset repository locally, for example using
```
git clone https://huggingface.co/datasets/eli5
```
then apply your modifications and pass the local path to the repository to [`~Dataset.load_dataset`] to try your modifications:
````

File: docs/source/loading.mdx [hunk 2]
```diff
@@ -80,6 +80,14 @@ You may have a 🤗 Datasets loading script locally on your computer. In this ca
...
+
+```py
+>>> from datasets import load_dataset
+# edit local loading script and now load the dataset
+>>> eli5 = load_dataset("datasets/eli5")
```
lhoestq, line 88:
```suggestion
>>> eli5 = load_dataset("/local/path/to/datasets/eli5")
```

lhoestq (commented):
I added a suggestion to present a more general way to edit a dataset from the Hub, WDYT ?

stevhliu:
Good idea! I went with @mariosasko's suggestion to use `inspect_dataset` instead of cloning a dataset repository since it's a good opportunity to show off more of the library's lesser-known functions if that's ok with everyone :)

lhoestq (dismissed):
Love it !

mariosasko:
One advantage of cloning the repo is that it fetches potential data files referenced inside a script using relative paths, so if we decide to use `inspect_dataset`, we should at least add a tip to explain this limitation and how to circumvent it.

lhoestq:
Oh you're right. Calling `load_dataset` on the modified script without having the files that come with it is not ideal. I agree it should be `git clone` instead - and inspect is for inspection only ^^'

docs/source/loading.mdx
```diff
@@ -80,6 +80,21 @@ You may have a 🤗 Datasets loading script locally on your computer. In this ca
...
+
+```bash
+git clone https://huggingface.co/datasets/eli5
+```
+
+Make your edits to the loading script and then load it by passing its local path to [`~datasets.load_dataset`]:
+
+```py
+>>> from datasets import load_dataset
+>>> eli5 = load_dataset("/local/path/to/datasets/eli5")
```
albertvillanova, line 95: To be consistent with the paragraph above:
```suggestion
>>> eli5 = load_dataset("path/to/local/eli5")
```

Also note that there isn't necessarily a "datasets" directory once cloned from the Hub (and not from GitHub).

albertvillanova (approved):
Thanks @stevhliu, much clearer now! :)

PR #4797, opened 2022-08-05
Torgo dataset creation

YingLi001:


albertvillanova:
Hi @YingLi001, thanks for your proposal to add this dataset.

However, now we add datasets directly to the Hub (instead of our GitHub repository). You have the instructions in our docs: 
- [Create a dataset loading script](https://huggingface.co/docs/datasets/dataset_script)
- [Create a dataset card](https://huggingface.co/docs/datasets/dataset_card)
- [Share](https://huggingface.co/docs/datasets/share)

PR #4800, opened 2022-08-08
support LargeListArray in pyarrow

Jiaxin-Wen:
```python
import numpy as np
import datasets
a = np.zeros((5000000, 768))
res = datasets.Dataset.from_dict({'embedding': a})

'''
  File '/home/wenjiaxin/anaconda3/envs/data/lib/python3.8/site-packages/datasets/arrow_writer.py', line 178, in __arrow_array__
    out = numpy_to_pyarrow_listarray(data)
  File "/home/wenjiaxin/anaconda3/envs/data/lib/python3.8/site-packages/datasets/features/features.py", line 1173, in numpy_to_pyarrow_listarray
    offsets = pa.array(np.arange(n_offsets + 1) * step_offsets, type=pa.int32())
  File "pyarrow/array.pxi", line 312, in pyarrow.lib.array
  File "pyarrow/array.pxi", line 83, in pyarrow.lib._ndarray_to_array
  File "pyarrow/error.pxi", line 100, in pyarrow.lib.check_status
pyarrow.lib.ArrowInvalid: Integer value 2147483904 not in range: -2147483648 to 2147483647
'''
```

Loading a large numpy array currently raises the error above as the type of offsets is `int32`. 
And pyarrow has supported [LargeListArray](https://arrow.apache.org/docs/python/generated/pyarrow.LargeListArray.html) for this case.


mariosasko:
Hi, thanks for working on this! Can you run `make style` at the repo root to fix the code quality error in CI and add a test?

xwwwwww:
Hi, I have fixed the code quality error and added a test

xwwwwww:
It seems that CI fails due to the lack of memory for allocating a large array, while I pass the test locally.

tests/features/test_array_xd.py
```diff
@@ -359,6 +359,17 @@ def test_array_xd_with_np(data, feature, expected):
     assert ds[0]["col"] == expected
 
 
+@pytest.mark.parametrize(
+    "data, feature, expected",
+    [
+        (np.zeros((5000000, 768), dtype=np.float32), None, [0] * 768),
```
mariosasko, line 365: This should be enough to pass the tests:

```suggestion
        (np.zeros((5000000, 768), dtype=np.uint8), None, [0] * 768),
```

(According to the GH docs, the Windows/Linux runner has 7GB of RAM)

mariosasko:
Also, the current implementation of the NumPy-to-PyArrow conversion creates a lot of copies, which is not ideal for large arrays.

We can improve performance significantly if we rewrite this part:
https://github.com/huggingface/datasets/blob/83f695c14507a3a38e9f4d84612cf49e5f50c153/src/datasets/features/features.py#L1322-L1323

as
```python
 values = pa.array(arr.ravel(), type=type) 
```

tests/features/test_array_xd.py
xwwwwww: still failed while using uint8 :(

mariosasko: Okay, let's try with this array: `np.zeros((46341, 46341), dtype=np.uint8)`  (it goes barely over the limit)

mariosasko:
@xwwwwww Feel free to ignore https://github.com/huggingface/datasets/pull/4800#issuecomment-1212280549 and revert the changes you've made to address it. 

Without copying the array, this would be possible:
```python
arr = np.array([
    [1, 2, 3],
    [4, 5, 6]
])

dset = Dataset.from_dict({"data": [arr]})

arr[0][0] = 100 # this change would be reflected in dset's PyArrow table -> a breaking change and also probably unexpected by the user 
```

xwwwwww:
> @xwwwwww Feel free to ignore [#4800 (comment)](https://github.com/huggingface/datasets/pull/4800#issuecomment-1212280549) and revert the changes you've made to address it.
> 
> Without copying the array, this would be possible:
> 
> ```python
> arr = np.array([
>     [1, 2, 3],
>     [4, 5, 6]
> ])
> 
> dset = Dataset.from_dict({"data": [arr]})
> 
> arr[0][0] = 100 # this change would be reflected in dset's PyArrow table -> a breaking change and also probably unexpected by the user 
> ```

Oh, that makes sense.

xwwwwww:
passed tests in ubuntu while failed in windows

xwwwwww:
@mariosasko Hi, do you have any clue about this failure in windows?

mariosasko:
Perhaps we can skip the added test on Windows then.

Not sure if this can help, but the ERR tool available on Windows outputs the following for the returned error code `-1073741819`:
```
# for decimal -1073741819 / hex 0xc0000005
  ISCSI_ERR_SETUP_NETWORK_NODE                                   iscsilog.h
# Failed to setup initiator portal. Error status is given in
# the dump data.
  STATUS_ACCESS_VIOLATION                                        ntstatus.h
# The instruction at 0x%p referenced memory at 0x%p. The
# memory could not be %s.
  USBD_STATUS_DEV_NOT_RESPONDING                                 usb.h
# as an HRESULT: Severity: FAILURE (1), FACILITY_NONE (0x0), Code 0x5
# for decimal 5 / hex 0x5
  WINBIO_FP_TOO_FAST                                             winbio_err.h
# Move your finger more slowly on the fingerprint reader.
# as an HRESULT: Severity: FAILURE (1), FACILITY_NULL (0x0), Code 0x5
  ERROR_ACCESS_DENIED                                            winerror.h
# Access is denied.
# 5 matches found for "-1073741819"
```

xwwwwww:
What's the proper way to skip the added test in windows?
I tried `if platform.system() == 'Linux'`, but the CI test seems stuck

xwwwwww:
@mariosasko Hi, any idea about this :)

mariosasko:
Hi again! We want to skip the test on Windows but not on Linux. You can use this decorator to do so: 
```python
@pytest.mark.skipif(os.name == "nt" and (os.getenv("CIRCLECI") == "true" or os.getenv("GITHUB_ACTIONS") == "true"), reason="The Windows CI runner does not have enough RAM to run this test")
@pytest.mark.parametrize(...)
def test_large_array_xd_with_np(...):
    ...
```

xwwwwww:
> Hi again! We want to skip the test on Windows but not on Linux. You can use this decorator to do so:
> 
> ```python
> @pytest.mark.skipif(os.name == "nt" and (os.getenv("CIRCLECI") == "true" or os.getenv("GITHUB_ACTIONS") == "true"), reason="The Windows CI runner does not have enough RAM to run this test")
> @pytest.mark.parametrize(...)
> def test_large_array_xd_with_np(...):
>     ...
> ```

CI on windows still stucks :(

xwwwwww:
@mariosasko Hi, could you please take a look at this issue

xwwwwww:
@mariosasko Hi, all checks have passed, and we are finally ready to merge this PR :)

xwwwwww:
@lhoestq @albertvillanova Perhaps other maintainers can take a look and merge this PR :)

src/datasets/features/features.py
```diff
@@ -1324,8 +1342,12 @@ def numpy_to_pyarrow_listarray(arr: np.ndarray, type: pa.DataType = None) -> pa.
...
         n_offsets = reduce(mul, arr.shape[: arr.ndim - i - 1], 1)
         step_offsets = arr.shape[arr.ndim - i - 1]
-        offsets = pa.array(np.arange(n_offsets + 1) * step_offsets, type=pa.int32())
-        values = pa.ListArray.from_arrays(offsets, values)
+        if n_offsets * step_offsets < (1 << 31):
+            offsets = pa.array(np.arange(n_offsets + 1) * step_offsets, type=pa.int32())
+            values = pa.ListArray.from_arrays(offsets, values)
+        else:
+            offsets = pa.array(np.arange(n_offsets + 1) * step_offsets, type=pa.int64())
+            values = pa.LargeListArray.from_arrays(offsets, values)
```
lhoestq, line 1350: Have you tried using pa.chunked_array instead of pa.LargeListArray ? (i.e. chunking the input numpy array into small arrays to have a list of pa.ListArray that you concatenate with pa.chunked_array)

In the rest of the code base we don't support LargeListArray and so it could lead to issues when doing type inference or type casting

lhoestq (commented):
Thanks for fixing this ! I left a few comments

src/datasets/features/features.py
```diff
@@ -372,12 +387,15 @@ def _cast_to_python_objects(obj: Any, only_1d_for_numpy: bool, optimize_list_cas
...
-                ], True
+                return (
+                    [
+                        _cast_to_python_objects(
+                            elmt, only_1d_for_numpy=only_1d_for_numpy, optimize_list_casting=optimize_list_casting
+                        )[0]
+                        for elmt in obj
+                    ],
+                    True,
+                )
```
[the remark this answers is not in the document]
xwwwwww: So it is not recommended to run `make style`?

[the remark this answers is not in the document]
lhoestq: It is ! But maybe you didn't have the right versions of black and isort ;)

You can get the right versions with `pip install -e .[quality]` or `pip install datasets[quality]`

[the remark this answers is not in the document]
xwwwwww: emm I currently have `black==22.6.0` and `isort==5.10.1`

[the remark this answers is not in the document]
lhoestq: it sounds good now thanks !

src/datasets/features/features.py
xwwwwww: Do you mean that it will be ok to return a pa.ChunkedArray?

lhoestq: I think so but I haven't tested

lhoestq: Oh yea indeed. I'm not sure it would even work if the shape was (1, 50000, 50000) since I believe a ListArray can't have a ChunkedArray as values.

Let's go for LargeListArray then, though we'll have to update a few things in table.py but this can be handled later I think (let me know if that's something you'd like to help with !)

xwwwwww: Sure! Could you provide some guidance?

xwwwwww: @lhoestq Hi, I checked `table.py` and found that `pa.ListArray` is only called in the `array_cast` function. Does that mean I only need to update this function for the adaption of LargeListArray?

lhoestq: Yes indeed, in array_cast, cast_array_to_feature and embed_array_storage.

Then in features.py there's also generate_from_arrow_type to update

Finally there's still one situation that needs to be improved: when you concatenate a dataset with a list and a dataset with a large list. I think in this case the regular list should be promoted to a large list but this is maybe less trivial

xwwwwww: Do you mean the case where the two datasets to be concatenated share the same column? However, I think it would raise a ValueError for duplicated columns.

xwwwwww: And I've updated all these three functions, i.e., array_cast, cast_array_to_feature, and embed_array_storage.
Please review these changes as I am not familiar with these functions :)

lhoestq: > Do you mean the case where the two datasets to be concatenated share the same column? However, I think it would raise a ValueError for duplicated columns.

When you have two datasets with the same column "embedding" but one has 50,000 elements (so it uses LargeListArray) and the second dataset has 1 element (so it uses a ListArray)

xwwwwww: take this as an example
```python
a = np.zeros((3,3))
da = datasets.Dataset.from_dict({"col": a})
b = np.ones((3,3))
db = datasets.Dataset.from_dict({"col": b})
res = datasets.concatenate_datasets([da, db], axis=1)
```

it would raise the following error
```none
  File "tmp.py", line 35, in <module>
    res = datasets.concatenate_datasets([da, db], axis=1)
  File "/data/wenjiaxin/home/datasets/src/datasets/combine.py", line 176, in concatenate_datasets
    return _concatenate_map_style_datasets(dsets, info=info, split=split, axis=axis)
  File "/data/wenjiaxin/home/datasets/src/datasets/arrow_dataset.py", line 4773, in _concatenate_map_style_datasets
    _check_column_names([col_name for dset in dsets for col_name in dset._data.column_names])
  File "/data/wenjiaxin/home/datasets/src/datasets/arrow_dataset.py", line 606, in _check_column_names
    raise ValueError(f"The table can't have duplicated columns but columns {duplicated_columns} are duplicated.")
ValueError: The table can't have duplicated columns but columns ['col'] are duplicated.
```

lhoestq: I mean concatenate on axis=0 ;)

xwwwwww: @mariosasko @lhoestq Hi, do we have conclusions about this issue :)

lhoestq: Maybe there's another way, what about using FixedSizeListArray instead ? It doesn't have an offsets array so it should fix the issue ?

xwwwwww:
```python
a = np.zeros((50000, 50000))
da = datasets.Dataset.from_dict({"col": a}, features=datasets.Features({"col": Sequence(feature=Value(dtype='float64', id=None))}))
```

I test this case, and it successfully passes the sanity check. 

Both `self.info.features.type` and `inferred_features.type` is `struct<col: list<item: double>>`

lhoestq: Oh yea right ! It returns an arrow type using `list` every time so it should be alright ! Good catch :)

xwwwwww: 😄okay

xwwwwww: Considering that @mariosasko may be too busy these days, shall we merge this PR and support this important feature? @lhoestq

lhoestq: we're discussing this internally and doing some tests to make sure - will keep you posted ;)

lhoestq: From internal discussions, this code would fail because the ArrowWriter can't start writing large lists to disk if the schema has been determined to contain regular lists:
```python
import numpy as np
import datasets
a = np.zeros((5000000, 768), np.uint8)

datasets.Dataset.from_dict({"id": range(2)}).map(lambda _: {"a": a}, writer_batch_size=1, new_fingerprint="foo")
# ArrowInvalid: Array of type large_list<item: uint8> too large to convert to list<item: uint8>
```

This means that whenever someone runs `map` with varying lengths arrays and at one point an array doesn't fit into a regular ListArray, this will fail.

In this case we expect to let the user specify in advance that the largelist type must be used.
Therefore I think we need to add a parameter to `Sequence` to differentiate between regular lists and large lists. Maybe `Sequence(..., large=True)` ?
and this way we can also always verify
```python
schema == Features.from_arrow_schema(schema).arrow_schema
```

I hope that makes sense ^^'

xwwwwww: got it! Could you please provide some instructions on how to support this, e.g., functions I have to update

lhoestq: I think the key is to add the `large` parameter to `Sequence` and update the functions you modified in this PR to use pa.list_() if large is False, and pa.large_list otherwise

one-matrix:
same issus come from pyarrow.Is there a solution for this?
file parquet:50GB
datasets version: 2.14.4
pyarrow :12.0.1

Generating train split: 0 examples [01:22, ? examples/s]
  File "/opt/conda/lib/python3.10/site-packages/datasets/builder.py", line 1925, in _prepare_split_single
    for _, table in generator:
  File "/opt/conda/lib/python3.10/site-packages/datasets/packaged_modules/parquet/parquet.py", line 79, in _generate_tables
    for batch_idx, record_batch in enumerate(
  File "pyarrow/_parquet.pyx", line 1315, in iter_batches
  File "pyarrow/error.pxi", line 115, in pyarrow.lib.check_status
OSError: List index overflow.

lkfo415579:
when this feature adds to the newest version?

lhoestq:
LargeListArray support is not ready yet, there is one remaining change:

> I think the key is to add the large parameter to Sequence and update the functions you modified in this PR to use pa.list_() if large is False, and pa.large_list otherwise

thusinh1969:
Gents, any move on this. Convert largse list of dicts to Datasets is a nightmare and took all RAM possible. Is there any other alternative?

Thanks,
Steve

albertvillanova:
Arrow large_list is supported since datasets 2.21.0. See: https://github.com/huggingface/datasets/releases/tag/2.21.0
- #7019

PR #4818, opened 2022-08-10
Add add cc-by-sa-2.5 license tag

polinaeterna:
- [ ] add it to moon-landing
- [ ] add it to hub-docs


albertvillanova:
I think we can close this PR because the `standard_licenses.tsv` file was removed from this repo and we no longer perform any dataset card validation.

PR #4322, opened 2022-05-12
Added stratify option to train_test_split function.

nandwalritik:
This PR adds `stratify` option to `train_test_split` method. I took reference from scikit-learn's `StratifiedShuffleSplit` class for implementing stratified split and integrated the changes as were suggested by @lhoestq.

It fixes #3452.

@lhoestq Please review and let me know, if any changes are required.

PR #4821, opened 2022-08-11
Fix train_test_split docs

NielsRogge:
I saw that `stratify` is added to the `train_test_split` method as per #4322, hence the docs can be updated.


albertvillanova (approved):
Thanks for the catch.

PR #4825, opened 2022-08-11
[Windows] Fix Access Denied when using os.rename()

### LINKED ISSUE #2937: load_dataset using default cache on Windows causes PermissionError: [WinError 5] Access is denied
## Describe the bug
Standard process to download and load the wiki_bio dataset causes PermissionError in Windows 10 and 11.

## Steps to reproduce the bug
```python
ds = load_dataset('wiki_bio')
```

## Expected results
It is expected that the dataset downloads without any errors.

## Actual results
PermissionError see trace below:
```
Using custom data configuration default
Downloading and preparing dataset wiki_bio/default (download: 318.53 MiB, generated: 736.94 MiB, post-processed: Unknown size, total: 1.03 GiB) to C:\Users\username\.cache\huggingface\datasets\wiki_bio\default\1.1.0\5293ce565954ba965dada626f1e79684e98172d950371d266bf3caaf87e911c9...
  File "C:\Users\username\.conda\envs\hf\lib\site-packages\datasets\load.py", line 1112, in load_dataset
  File "C:\Users\username\.conda\envs\hf\lib\site-packages\datasets\builder.py", line 644, in download_and_prepare
    self._save_info()
  File "C:\Users\username\.conda\envs\hf\lib\contextlib.py", line 120, in __exit__
    next(self.gen)
  File "C:\Users\username\.conda\envs\hf\lib\site-packages\datasets\builder.py", line 598, in incomplete_dir
    os.rename(tmp_dir, dirname)
PermissionError: [WinError 5] Access is denied: 'C:\\Users\\username\\.cache\\huggingface\\datasets\\wiki_bio\\default\\1.1.0\\5293ce565954ba965dada626f1e79684e98172d950371d266bf3caaf87e911c9.incomplete' -> 'C:\\Users\\username\\.cache\\huggingface\\datasets\\wiki_bio\\default\\1.1.0\\5293ce565954ba965dada626f1e79684e98172d950371d266bf3caaf87e911c9'
```
By commenting out the os.rename() [L604](https://github.com/huggingface/datasets/blob/master/src/datasets/builder.py#L604) and the shutil.rmtree() [L607](https://github.com/huggingface/datasets/blob/master/src/datasets/builder.py#L607) lines, in my virtual environment, I was able to get the load process to complete, rename the directory manually and then rerun the `load_dataset('wiki_bio')` to get what I needed.

It seems that os.rename() in the `incomplete_dir` content manager is the culprit. Here's another project [Conan](https://github.com/conan-io/conan/issues/6560) with similar issue with os.rename() if it helps debug this issue.

## Environment info
- `datasets` version: 1.12.1
- Platform: Windows-10-10.0.22449-SP0
- Python version: 3.8.12
- PyArrow version: 5.0.0

DougTrajano:
In this PR, we are including an additional step when `os.rename()` raises a PermissionError.

Basically, we will use `shutil.move()` on the temp files.

Fix #2937

### DIFF
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -664,6 +664,8 @@ class DatasetBuilder:
                         if os.path.isdir(dirname):
                             shutil.rmtree(dirname)
                         os.rename(tmp_dir, dirname)
+                    except PermissionError:
+                        shutil.move(tmp_dir, dirname)
                     finally:
                         if os.path.exists(tmp_dir):
                             shutil.rmtree(tmp_dir)
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -663,8 +663,6 @@ class DatasetBuilder:
                         yield tmp_dir
                         if os.path.isdir(dirname):
                             shutil.rmtree(dirname)
-                        os.rename(tmp_dir, dirname)
-                    except PermissionError:
                         shutil.move(tmp_dir, dirname)
                     finally:
                         if os.path.exists(tmp_dir):


lhoestq:
Cool thank you ! Maybe we can just replace `os.rename` by `shutil.move` instead ?

DougTrajano:
> Cool thank you ! Maybe we can just replace `os.rename` by `shutil.move` instead ?

Yes, I think that could be a better solution, but I didn't test it in Linux (e.g. Ubuntu) to guarantee that `os.rename()` could be completely replaced by `shutil.move()`.

lhoestq:
`shutil.move` does call `os.rename` first before doing extra work to make it work on windows, so this is a safe change for linux ;)

DougTrajano:
> AFAIK `shutil.move` does call `os.rename` first before doing extra work to make it work on windows, so this is should be a safe safe change for linux ;)

alright, let me change the PR then.

DougTrajano:
Hi @lhoestq looks like one of the tests failed, but is not related to this change, do I need to do something from my side?

lhoestq (approved):
It's all good thanks :)

the error is unrelated to your change

PR #4826, opened 2022-08-11
Fix language tags in dataset cards

albertvillanova:
Fix language tags in all dataset cards, so that they are validated (aligned with our `languages.json` resource).


albertvillanova:

PR #4827, opened 2022-08-11
Add license metadata to pg19

julien-c:
As reported over email by Roy Rijkers

### DIFF
--- a/datasets/pg19/README.md
+++ b/datasets/pg19/README.md
@@ -1,7 +1,8 @@
 ---
 language:
 - en
-license: apache-2.0
+license:
+- apache-2.0
 paperswithcode_id: pg-19
 pretty_name: PG-19
 ---
--- a/datasets/pg19/README.md
+++ b/datasets/pg19/README.md
@@ -1,6 +1,7 @@
 ---
 language:
 - en
+license: apache-2.0
 paperswithcode_id: pg-19
 pretty_name: PG-19
 ---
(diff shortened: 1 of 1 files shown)


albertvillanova (approved):
Thanks Roy and Julien for the addendum.

datasets/pg19/README.md
```diff
@@ -1,6 +1,7 @@
 ---
 language:
 - en
+license: apache-2.0
```
albertvillanova, line 4: The license tag should be an array.

albertvillanova (changes requested):
Fix needed to pass the validation.

datasets/pg19/README.md
julien-c: hmm that's not what's documented on https://huggingface.co/docs/hub/datasets-cards =) 

Screenshot 2022-08-11 at 16 12 20

julien-c: but ok i'll push this for consistency

albertvillanova (approved):
Then non-passing tests have nothing to do with this PR, but with other missing information in the dataset card.

datasets/pg19/README.md
albertvillanova: Good catch: we should align docs and validation implementation:

https://github.com/huggingface/datasets/blob/cfae8ac7c896772af537441508aaae2391aea53a/src/datasets/utils/metadata.py#L232

PR #4828, opened 2022-08-11
Support PIL Image objects in `add_item`/`add_column`

### LINKED ISSUE #4796: ArrowInvalid: Could not convert <PIL.Image.Image image mode=RGB when adding image to Dataset
## Describe the bug

When adding a Pillow image to an existing Dataset on the hub, `add_item` fails due to the Pillow image not being automatically converted into the Image feature.

## Steps to reproduce the bug
```python
from PIL import Image

dataset = load_dataset("hf-internal-testing/example-documents")

image = Image.open("/content/cord_example.png").convert("RGB")

new_image = {'image': image}
dataset['test'] = dataset['test'].add_item(new_image)
```

## Expected results
The image should be automatically casted to the Image feature when using `add_item`. For now, this can be fixed by using `encode_example`:

```
import datasets

feature = datasets.Image(decode=False)
new_image = {'image': feature.encode_example(image)}
dataset['test'] = dataset['test'].add_item(new_image)
```

## Actual results

```
ArrowInvalid: Could not convert <PIL.Image.Image image mode=RGB size=576x864 at 0x7F7CCC4589D0> with type Image: did not recognize Python value type when inferring an Arrow data type
```

mariosasko:
Fix #4796 

PS: We should also improve the type inference in `OptimizedTypeSequence` to make it possible to also infer the complex types (only `Image` currently) in nested arrays (e.g.  `[[pil_image], [pil_image, pil_image]]` or `[{"img": pil_image}`]), but I plan to address this in a separate PR.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -4388,14 +4388,15 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         })
         ```
         """
-        column_table = InMemoryTable.from_pydict({name: column})
+        ts = OptimizedTypedSequence(column, type=None, col=name)
+        column_table = InMemoryTable.from_pydict({name: ts})
         _check_column_names(self._data.column_names + column_table.column_names)
         dataset = self.flatten_indices() if self._indices is not None else self
         # Concatenate tables horizontally
         table = concat_tables([dataset._data, column_table], axis=1)
         # Update features
         info = dataset.info.copy()
-        info.features.update(Features.from_arrow_schema(column_table.schema))
+        info.features.update(Features({name: ts.get_inferred_type()}))
         table = update_metadata_with_features(table, info.features)
         return Dataset(table, info=info, split=self.split, indices_table=None, fingerprint=new_fingerprint)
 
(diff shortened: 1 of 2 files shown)


src/datasets/arrow_dataset.py
```diff
@@ -4388,14 +4388,15 @@ def add_column(self, name: str, column: Union[list, np.array], new_fingerprint:
...
+        ts = OptimizedTypedSequence(column, type=None, col=name)
+        column_table = InMemoryTable.from_pydict({name: ts})
         _check_column_names(self._data.column_names + column_table.column_names)
         dataset = self.flatten_indices() if self._indices is not None else self
         # Concatenate tables horizontally
         table = concat_tables([dataset._data, column_table], axis=1)
         # Update features
         info = dataset.info.copy()
-        info.features.update(Features.from_arrow_schema(column_table.schema))
+        info.features.update(Features({name: ts.get_inferred_type()}))
```
lhoestq, line 4399: Why not keep `Features.from_arrow_schema` here ? `get_inferred_type` instantiates an extra array to infer the types.

src/datasets/arrow_dataset.py
```diff
@@ -4626,9 +4627,12 @@ def add_item(self, item: dict, new_fingerprint: str):
         {'label': 0, 'text': 'this movie is the absolute worst thing I have ever seen'}
         ```
         """
-        item_table = InMemoryTable.from_pydict({k: [v] for k, v in item.items()})
+        item = {k: OptimizedTypedSequence([v], type=None, col=k) for k, v in item.items()}
+        item_table = InMemoryTable.from_pydict(item)
         # We don't call _check_if_features_can_be_aligned here so this cast is "unsafe"
-        dset_features, item_features = _align_features([self.features, Features.from_arrow_schema(item_table.schema)])
+        dset_features, item_features = _align_features(
+            [self.features, Features({k: v.get_inferred_type() for k, v in item.items()})]
```
lhoestq, line 4634: same here

lhoestq (approved):
Nice thanks !

src/datasets/arrow_dataset.py
mariosasko: `get_inferred_type` is needed for the type inference (e.g. for a sequence of PIL Images `Features.from_arrow_schema` would return `{'bytes': Value(dtype='binary', id=None), 'path': Value(dtype='string', id=None)}` instead of `Image(decode=True, id=None)`).

lhoestq: Oh ok I see ! Anyway later we can cache the inferred type when `__arrow_array__` is called for the first time, this way `get_inferred_type` doesn't recreate an arrow array

SamuelzXu:
Hey @mariosasko could we please merge this? I'm still getting the original error at #4796 .

albertvillanova:
Are you planning to continue working on this?

PR #4837, opened 2022-08-12
Add support for CSV metadata files to ImageFolder

### LINKED ISSUE #4814: Support CSV as metadata file format in AudioFolder/ImageFolder
Requested here: https://discuss.huggingface.co/t/how-to-structure-an-image-dataset-repo-using-the-image-folder-approach/21004. CSV is also used in AutoTrain for specifying metadata in image datasets.

mariosasko:
Fix #4814


lhoestq:
Cool thanks ! Maybe let's include this change after the refactoring from FolderBasedBuilder in #3963 to avoid dealing with unpleasant conflicts ?

mariosasko:
@lhoestq I resolved the conflicts (AudioFolder also supports CSV metadata now). Let me know what you think.

lhoestq (approved):
Awesome thanks !

I think we can see if we can stop having to duplicate the tests in both imagefolder and audiofolder later

mariosasko:
@lhoestq Thanks for the suggestion! Indeed it makes more sense to use CSV as the default format in the folder-based builders.

PR #4841, opened 2022-08-12
Update ted_talks_iwslt license to include ND

cakiki:
Excerpt from the paper's abstract: "Aside from its cultural and social relevance, this content, which is published under the Creative Commons BY-NC-ND license, also represents a precious language resource for the machine translation research community"


albertvillanova (approved):
Good catch. Thanks.

PR #4842, opened 2022-08-12
Update stackexchange license

cakiki:
The correct license of the stackexchange subset of the Pile is `cc-by-sa-4.0`, as can for example be seen here: https://stackoverflow.com/help/licensing


albertvillanova (approved):
Thaks for the fix.

PR #4866, opened 2022-08-19
amend docstring for dunder

schafsam:
display dunder method in docsting with underlines an not bold markdown.


albertvillanova (changes requested):
Hi @schafsam , good catch!

I think this should be surrounded with single backticks though...

Also I think this should be fixed everywhere in the docstrings.

PR #4867, opened 2022-08-19
Complete tags of superglue dataset card

richarddwang:
Related to #4479 .


albertvillanova (approved):
Thanks, @richarddwang. These fixes on the dataset card are really useful to improve discoverability and quality.

PR #4880, opened 2022-08-23
Added names of less-studied languages

BenjaminGalliot:
Added names of less-studied languages (nru – Narua and jya – Japhug) for existing datasets.


src/datasets/utils/resources/languages.json
```diff
@@ -682,6 +682,7 @@
     "jv-ID": "Javanese (Indonesia)",
     "jv-Java": "Javanese (Javanese)",
     "jv-x-bms": "Banyumasan",
+    "jya": "Japhug",
```
albertvillanova, line 685: Please note the official name corresponding to the tag "jya" is "Jiarong": Japhug is one of the Jiarong languages: 
- https://iso639-3.sil.org/code/jya
- https://en.wikipedia.org/wiki/Gyalrong_languages#Languages

```suggestion
    "jya": "Jiarong",
```

albertvillanova (changes requested):
Thanks for the addition, @BenjaminGalliot.

Just a suggested fix below.

src/datasets/utils/resources/languages.json
BenjaminGalliot: After a discussion with the linguistic experts, we decided to use the Glottolog codes, which are more precise and therefore more robust because they avoid potential future ambiguities...

alexis-michaud: If I may 'jump in': a more general topic is, Which way to go for identifying language varities. It's a big topic. Glottolog and Ethnologue both have information on language groupings. Glottolog is more fine-grained and the database maintainers are more transparent about how decisions are made, so, strange as the codes may seem ('japh1234' looks more unwieldy than 'jya'), they look like a good way to go. An issue was created [here](https://github.com/huggingface/datasets/issues/4881) in the hope of bringing attention to this topic.

BenjaminGalliot: And I just added the ISO 639-3 codes too since they are requested in the dataset card.

src/datasets/utils/resources/languages.json
```diff
@@ -670,6 +670,7 @@
     "ja-Latn": "Japanese (Latin)",
     "jak": "Jakun",
     "jam": "Jamaican Patois",
+    "japh1234": "Japhug",
```
albertvillanova, line 673: For the moment, we are not using Glottolog codes in our `languages.json` list. We use IANA language codes.

If you would like to propose another approach, it would be better to open an Issue page (or participate in some already opened, like #4881) to discuss and agree on it before being included to the library. Thanks.

src/datasets/utils/resources/languages.json
```diff
@@ -1052,7 +1055,7 @@
     "pt-MZ": "Portuguese (Mozambique)",
     "pt-PT": "Portuguese (Portugal)",
     "pt-ST": "Portuguese (São Tomé & Príncipe)",
-    "pt-TL": "Portuguese (Timor-Leste)", 
+    "pt-TL": "Portuguese (Timor-Leste)",
```
albertvillanova, line 1058: This trailing blank space should be removed.

src/datasets/utils/resources/languages.json
```diff
@@ -1379,6 +1382,7 @@
     "yo": "Yoruba",
     "yo-BJ": "Yoruba (Benin)",
     "yo-NG": "Yoruba (Nigeria)",
+    "yong1288": "Yongning Na",
```
albertvillanova, line 1385: The same as above.

albertvillanova (changes requested):
As pointed out in my comment below, currently we are using IANA language codes, not Glottolog codes.

src/datasets/utils/resources/languages.json
albertvillanova: Yes, I agree the issue page #4881 is the right approach to propose this change to discuss and agree on it before incorporating to the library. :)

src/datasets/utils/resources/languages.json
BenjaminGalliot: I don’t see any trailing blank space… I configured my IDE to remove them on file save, that is why this unrelated line had been modified…

src/datasets/utils/resources/languages.json
BenjaminGalliot: Same!

src/datasets/utils/resources/languages.json
BenjaminGalliot: Yes, I noticed it was checked, that is why I modified the corpus card to include both ISO 639-3 and Glottolog codes (the latter are used as sub-corpus names)

src/datasets/utils/resources/languages.json
albertvillanova: @BenjaminGalliot are you working on a specific dataset addition?

src/datasets/utils/resources/languages.json
BenjaminGalliot: So, if I understand well, I should only add the IANA/ISO 639-3 codes in this file, and the Glottolog codes in my corpus `language_details` field, right?

src/datasets/utils/resources/languages.json
BenjaminGalliot: Yes, called [pangloss](https://huggingface.co/datasets/Lacito/pangloss)

albertvillanova: Great dataset, @BenjaminGalliot! 

Please note that our `languages.json` list is used to validate only legacy datasets present on this GitHub repo.

Currently, datasets (as your case) are added to the Hub instead, and their validation is done by the Hub team: our `languages.json` file is not used by them, although we try to be on sync.
- I have pinged that team on the issue discussion page: #4881

In the meantime (the change of language codes could take some time), I would recommend you to used the `language_details` tag instead, something like:
```
language_details: 
```
language_details: "'jya' consists of regular 'japh1234'"
```

src/datasets/utils/resources/languages.json
albertvillanova: Yes, for the moment I think that would be the simplest/fastest thing...

albertvillanova: The change of the coding schema has to be agreed with the rest of the teams and will take some time...

src/datasets/utils/resources/languages.json
BenjaminGalliot: OK, I removed the `language_bcp47` field and added the `language_details` one!

BenjaminGalliot:
OK, I removed Glottolog codes and only added ISO 639-3 ones. The former are for the moment in corpus card description, language details, and in subcorpora names.

albertvillanova (approved):
Thanks.

Issue #4885, opened 2022-08-24
Create dataset from list of dicts


sanderland:
I often find myself with data from a variety of sources, and a list of dicts is very common among these.
However, converting this to a Dataset is a little awkward, requiring either

```Dataset.from_pandas(pd.DataFrame(formatted_training_data))```
Which can error out on some more exotic values as 2-d arrays for reasons that are not entirely clear
> ArrowInvalid: ('Can only convert 1-dimensional array values', 'Conversion failed for column labels with type object')

Alternatively:
```Dataset.from_dict({k: [s[k] for s in formatted_training_data] for k in  formatted_training_data[0].keys()})```
Which works, but is a little ugly.

**Describe the solution you'd like**
Either `.from_dict` accepting a list of dicts, or a `.from_records` function accepting such.

I am happy to PR this, just wanted to check you are happy to accept this I haven't missed something obvious, and which of the solutions would be prefered.

albertvillanova:
Hi @sanderland, thanks for your enhancement proposal.

I agree with you that this would be useful.

Please note that under the hood, we use PyArrow tables as backend:
- The implementation  of `Dataset.from_dict` uses the PyArrow `Table.from_pydict`

Therefore, I would suggest:
- Implementing `Dataset.from_list` using the PyArrow `Table.from_pylist`

What do you think?
Let's see if other people have other suggestions...

sanderland:
Thanks for the quick and positive reply @albertvillanova! 
`from_list` seems sensible. Have opened a PR so we can discuss details there.

mariosasko:
Resolved via #4890.

PR #4890, opened 2022-08-25
add Dataset.from_list

sanderland:
As discussed in #4885 

I initially added this bit at the end, thinking filling this field was necessary as it is done in from_dict. 
However, it seems the constructor takes care of filling info when it is empty.
```
if info.features is None:
  info.features = Features(
      {
          col: generate_from_arrow_type(coldata.type)
          for col, coldata in zip(pa_table.column_names, pa_table.columns)
      }
  )
```

### DIFF
--- a/docs/source/loading.mdx
+++ b/docs/source/loading.mdx
@@ -211,7 +211,7 @@ Load a list of Python dictionaries with [`~Dataset.from_list`]:
 
 ```py
 >>> from datasets import Dataset
->>> my_list = [{"a": 1}, {"a": 2}]
+>>> my_list = [{"a": 1}, {"a": 2}, {"a": 3}]
 >>> dataset = Dataset.from_list(my_list)
 ```
 
--- a/docs/source/loading.mdx
+++ b/docs/source/loading.mdx
@@ -205,6 +205,16 @@ Load Python dictionaries with [`~Dataset.from_dict`]:
 >>> dataset = Dataset.from_dict(my_dict)
 ```
 
+### Python list of dictionaries
+
+Load a list of Python dictionaries with [`~Dataset.from_list`]:
+
+```py
+>>> from datasets import Dataset
+>>> my_list = [{"a": 1}, {"a": 2}]
+>>> dataset = Dataset.from_list(my_list)
+```
+
 ### Pandas DataFrame
 
 Load Pandas DataFrames with [`~Dataset.from_pandas`]:
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -63,7 +63,13 @@ from .arrow_writer import ArrowWriter, OptimizedTypedSequence
 from .download.download_config import DownloadConfig
 from .download.streaming_download_manager import xgetsize
 from .features import Audio, ClassLabel, Features, Image, Sequence, Value
-from .features.features import FeatureType, decode_nested_example, pandas_types_mapper, require_decoding
+from .features.features import (
+    FeatureType,
+    decode_nested_example,
+    generate_from_arrow_type,
+    pandas_types_mapper,
+    require_decoding,
+)
 from .filesystems import extract_path_from_uri, is_remote_filesystem
 from .fingerprint import (
     fingerprint_transform,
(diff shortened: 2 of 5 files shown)


chatdesk:
@albertvillanova it seems PyArrow 6 does not have from_list while PyArrow 7 does, causing tests on v6 to fail. How do you usually handle such version differences?

sanderland:
@albertvillanova it seems tests fail on pyarrow 6, perhaps from_pylist is a v7 method? How do you usually handle these version differences?

src/datasets/table.py
```diff
@@ -749,6 +749,28 @@ def from_pydict(cls, *args, **kwargs):
...
+            metadata (:obj:`Union[dict, Mapping]`, default None):
+                Optional metadata for the schema (if inferred).
+
+        Returns:
+            :class:`datasets.table.Table`:
+        """
+        try:
+            return cls(pa.Table.from_pylist(mapping, *args, **kwargs))
+        except AttributeError:  # pyarrow <7 does not have from_pylist, so we convert and use from_pydict
+            mapping = {k: [r.get(k) for r in mapping] for k in mapping[0]}
```
lhoestq, line 771: for empty lists we can just return an empty mapping
```suggestion
            mapping = {k: [r.get(k) for r in mapping] for k in mapping[0]} if mapping else {}
```

Can you also add a test in `test_table.py` please ?

src/datasets/arrow_dataset.py
```diff
@@ -872,6 +872,40 @@ def from_dict(
...
+        """
+        if info is not None and features is not None and info.features != features:
+            raise ValueError(
+                f"Features specified in `features` and `info.features` can't be different:\n{features}\n{info.features}"
+            )
+        features = features if features is not None else info.features if info is not None else None
+        if info is None:
+            info = DatasetInfo()
+        info.features = features
+        pa_table = InMemoryTable.from_pylist(mapping=mapping)
```
lhoestq, line 906: We need to use `OptimizedTypedSequence` here as in `from_dict` for consistency.

I think it's ok to simply transform the data into a dict and use `cls.from_dict` here to keep things simple.

tests/test_dataset_list.py
```diff
@@ -0,0 +1,42 @@
+from unittest import TestCase
+
+from datasets import Sequence, Value
+from datasets.arrow_dataset import Dataset
+
+
+class DatasetListTest(TestCase):
```
lhoestq, line 7: You can put these tests in `test_arrow_dataset.py` with the `from_dict` tests :)

docs/source/loading.mdx
```diff
@@ -205,6 +205,16 @@ Load Python dictionaries with [`~Dataset.from_dict`]:
 >>> dataset = Dataset.from_dict(my_dict)
 ```
 
+### Python list of dictionaries
+
+Load a list of Python dictionaries with [`~Dataset.from_list`]:
+
+```py
+>>> from datasets import Dataset
+>>> my_list = [{"a": 1}, {"a": 2}]
```
lhoestq, line 214: (nit) I find it nice if we have the sample data as the `from_dict` example
```suggestion
>>> my_list = [{"a": 1}, {"a": 2}, {"a": 3}]
```

lhoestq (commented):
Cool ! Thanks a lot :)

I added a few comments:

tests/test_dataset_list.py
sanderland: A bit confused about this, since there is also "test_dataset_dict.py"

src/datasets/arrow_dataset.py
sanderland: Sure. Do you still want `InMemoryTable.from_pylist` even though it is not used anymore?

tests/test_dataset_list.py
lhoestq: oh right ! This way is fine then sorry

src/datasets/arrow_dataset.py
lhoestq: I think it's a great addition, let's keep it ^^

tests/test_dataset_list.py
sanderland: Alright, I think it's all set then :)

lhoestq (approved):
Thanks ! LGTM :)

PR #4909, opened 2022-08-29
Update GLUE evaluation metadata

lewtun:
This PR updates the evaluation metadata for GLUE to:

* Include defaults for all configs except `ax` (which only has a `test` split with no known labels)
* Fix the default split from `test` to `validation` since `test` splits in GLUE have no labels (they're private)
* Fix the `task_id` for some existing defaults

cc @sashavor @douwekiela


albertvillanova (approved):
Thank you.

Issue #4776, opened 2022-08-01
RuntimeError when using torchaudio 0.12.0 to load MP3 audio file


albertvillanova:
Current version of `torchaudio` (0.12.0) raises a RuntimeError when trying to use `sox_io` backend but non-Python dependency `sox` is not installed:
https://github.com/pytorch/audio/blob/2e1388401c434011e9f044b40bc8374f2ddfc414/torchaudio/backend/sox_io_backend.py#L21-L29
```python
def _fail_load(
    filepath: str,
    frame_offset: int = 0,
    num_frames: int = -1,
    normalize: bool = True,
    channels_first: bool = True,
    format: Optional[str] = None,
) -> Tuple[torch.Tensor, int]:
    raise RuntimeError("Failed to load audio from {}".format(filepath))
```

Maybe we should raise a more actionable error message so that the user knows how to fix it.

UPDATE:
- this is an incompatibility of latest torchaudio (0.12.0) and the sox backend

TODO:
- [x] as a temporary solution, we should recommend installing torchaudio<0.12.0
  - #4777 
  - #4785
- [ ] however, a stable solution must be found for torchaudio>=0.12.0

Related to: 
- https://github.com/huggingface/transformers/issues/18379

fxtentacle:
Requiring torchaudio<0.12.0 isn't really a viable solution because that implies torch<0.12.0 which means no sm_86 CUDA support which means no RTX 3090 support.

But in my case, the error only occurs if `_fallback_load` resolves to `_fail_load` inside torchaudio 0.12.0 which is only the case if FFMPEG initialization failed:  https://github.com/pytorch/audio/blob/b1f510fa5681e92ee82bdc6b2d1ed896799fc32c/torchaudio/backend/sox_io_backend.py#L36-L47

That means the proper solution for torchaudio>=0.12.0 is to check `torchaudio._extension._FFMPEG_INITIALIZED` and if it is False, then we need to remind the user to install a dynamically linked ffmpeg 4.1.8 and then maybe call `torchaudio._extension._init_ffmpeg()` to force a user-visible exception showing the missing ffmpeg dynamic library name.

On my system, installing 
- libavcodec.so.58 
- libavdevice.so.58  
- libavfilter.so.7  
- libavformat.so.58  
- libavutil.so.56  
- libswresample.so.3  
- libswscale.so.5
from ffmpeg 4.1.8 made HF datasets 2.3.2 work just fine with  torchaudio 0.12.1+cu116:

[...]

PR #4923, opened 2022-08-31
decode mp3 with librosa if torchaudio is > 0.12 as a temporary workaround 

polinaeterna:
`torchaudio>0.12` fails with decoding mp3 files if `ffmpeg<4`. currently we ask users to downgrade torchaudio, but sometimes it's not possible as torchaudio version is binded to torch version. as a temporary workaround we can decode mp3 with librosa (though it 60 times slower, at least it works)

another option would be to ask users to install the required version of `ffmpeg`, but is non-trivial on colab: it's not in apt packages in ubuntu 18 and `conda` is not preinstalled (with `conda` it would be easily installable)

- [x] decode with torchaudio anyway if the version of ffmpeg is correct? it's 60 times faster
- [x] tests 
- [x] DO NOT FORGET to get back all the tests

see https://github.com/huggingface/datasets/issues/4776 and https://github.com/huggingface/datasets/issues/3663#issuecomment-1225797165 (there is a Colab notebook to reproduce the error)


lhoestq:
Thanks ! Should we still support torchaudio>0.12 if it works ? And if it doesn't we can explain that downgrading is the right solution, or alternatively use librosa

polinaeterna:
@lhoestq 

> Should we still support torchaudio>0.12 if it works ? And if it doesn't we can explain that downgrading is the right solution, or alternatively use librosa

I'm not sure here, because from the one hand, if `torchaudio` works - it works 60 times faster then `librosa`.
But from the other hand, we will get inconsistent behavior (=different results of decoding) for users of `torchaudio>0.12`. 
I'd better go for using `librosa` only to avoid inconsistency then. wdyt?

lhoestq:
It seems a bit too constraining to not allow users who have a working torchaudio 0.12 setup to not use it. 

If the issue is about avoiding silent errors if the decoding changes, maybe we can log which back-end is used ? It can even be a warning with performance suggestions ("you're using librosa but torchaudio 0.xx is recommended").

Note that users can still have a requirements.txt or whatever in their projects if they really want full reproducibility (and it's the bare minimum imo)

There are multiple possible back-ends so it's maybe not reasonable to only allow one back-end, especially since each back-end has installation constrains and there's no "best" back-end.

src/datasets/features/audio.py
```diff
@@ -282,19 +287,44 @@ def _decode_non_mp3_file_like(self, file, format=None):
...
+                except ImportError as err:
+                    raise ImportError(
+                        "Your version of `torchaudio` (>=0.12.0) doesn't support decoding 'mp3' files on your machine. "
+                        "To support 'mp3' decoding with `torchaudio>=0.12.0`, please install `ffmpeg>=4` "
+                        'or downgrade `torchaudio` to <0.12: `pip install "torchaudio<0.12"`. '
+                        "To support decoding 'mp3' audio files without `torchaudio`, please install `librosa`: "
+                        "`pip install librosa`. Note that decoding will be extremely slow in that case."
+                    ) from err
+                # try to decode with librosa for torchaudio>=0.12.0 as a workaround
+                logger.warning("Decoding mp3 with `librosa` instead of `torchaudio`, decoding is slow.")
```
lhoestq, line 314: Maybe we should warn this only once ? You can use `warnings.warn` to do that

tests/utils.py
```diff
@@ -64,7 +64,16 @@ def parse_flag_from_env(key, default=False):
     find_library("sox") is None,
     reason="test requires sox OS dependency; only available on non-Windows: 'sudo apt-get install sox'",
 )
-require_torchaudio = pytest.mark.skipif(find_spec("torchaudio") is None, reason="test requires torchaudio")
+require_torchaudio = pytest.mark.skipif(
+    find_spec("torchaudio") is None
+    or version.parse(import_module("torchaudio").__version__) > version.parse("0.12.0"),
```
lhoestq, line 69:
```suggestion
    or version.parse(import_module("torchaudio").__version__) >= version.parse("0.12.0"),
```

lhoestq (approved):
Cool thank you !

src/datasets/features/audio.py
polinaeterna: I can't figure out how to do this. I changed `logger.warning` to `warnings.warn` here and tried setting `warnings.filterwarnings("once")` / `warnings.simplefilter("once")` but it didn't work, What am I missing?

polinaeterna: @

polinaeterna: @lhoestq

polinaeterna: ah lol I used warnings from logger to check if decoding was done with librosa, and now warnings from .warn are not captured in `caplog` in pytest, so these tests fail

lhoestq: I think it does the warning always once by default, you don't need to use an extra filter.

And I think `with pytest.warns(UserWarning):` should work in the test

polinaeterna: thanks! fixed it here https://github.com/huggingface/datasets/pull/4923/commits/ebf77d742db1e72386ae4b3d57c843d64cbefc6d

I don't know why but for some reason in librosa case warnings are shown at each decoding anyway, I checked it on Colab, see pic. Might it be because of that `librosa.load` itself gives warnings on decoding: `UserWarning: PySoundFile failed. Trying audioread instead.` 

Anyway, maybe not that important?

image

lhoestq: Users can still mute those warnings, I don't think we can do something about it.

lhoestq:
Woohoo all green ! Feel free to merge if it's all good for you :)

PR #4925, opened 2022-09-02
Add note about loading image / audio files to docs

lewtun:
This PR adds a small note about how to load image / audio datasets that have multiple splits in their dataset structure.

Related forum thread: https://discuss.huggingface.co/t/loading-train-and-test-splits-with-audiofolder/22447

cc @NielsRogge

### DIFF
--- a/docs/source/audio_load.mdx
+++ b/docs/source/audio_load.mdx
@@ -85,12 +85,14 @@ Load your dataset by specifying `audiofolder` and the directory of your dataset 
  'label': 1}
 ```
 
-If your directory contains multiple splits, use the `data_files` argument to load the images associated with each split:
+You can also use `audiofolder` to load datasets involving multiple splits. To do so, your dataset directory should have the following structure:
 
-```py
->>> from datasets import load_dataset
+```
+folder/train/electronic/01.mp3
+folder/train/punk/01.mp3
 
->>> dataset = load_dataset("audiofolder", data_files={"train":["train/electronic/01.mp3", ...], "test": ["test/electronic/01.mp3", ...]})
+folder/test/electronic/09.mp3
+folder/test/punk/09.mp3
 ```
 
 ## AudioFolder with metadata
@@ -140,6 +142,12 @@ You can load remote datasets from their URLs with the `data_files` parameter:
 ```py
 >>> dataset = load_dataset("audiofolder", data_files="https://s3.amazonaws.com/datasets.huggingface.co/SpeechCommands/v0.01/v0.01_test.tar.gz")
 ```
+
+<Tip>
+
+Some audio datasets, like those found in [Kaggle competitions](https://www.kaggle.com/), have separate metadata files for each split. Provided the metadata features are the same for each split, `audiofolder` can be used to load all splits at once. If the metadata features differ across each split, you should load them with separate `load_dataset()` calls.
+
+</Tip>
 
 ## AudioFolder with labels
 
--- a/docs/source/audio_load.mdx
+++ b/docs/source/audio_load.mdx
@@ -57,7 +57,41 @@ If you only want to load the underlying path to the audio dataset without decodi
 
 ## AudioFolder
 
-You can also load a dataset with an `AudioFolder` dataset builder. It does not require writing a custom dataloader, making it useful for quickly loading audio data.
+You can also load a dataset with an `AudioFolder` dataset builder. It does not require writing a custom dataloader, making it useful for quickly loading audio data. For classification tasks, your audio dataset structure should look like this:
+
+```
+folder/train/electronic/01.mp3
+folder/train/punk/01.mp3
+```
+
+Load your dataset by specifying `audiofolder` and the directory of your dataset in `data_dir`:
+
+```py
+>>> from datasets import load_dataset
+
+>>> dataset = load_dataset("audiofolder", data_dir="/path/to/folder")
+>>> dataset["train"][0]
+{'audio': {'path': '/path/to/electronic/01.mp3',
+  'array': array([ 3.9714024e-07,  7.3031038e-07,  7.5640685e-07, ...,
+         -1.1963668e-01, -1.1681189e-01, -1.1244172e-01], dtype=float32),
+  'sampling_rate': 44100},
+ 'label': 0}
+
+>>> dataset["train"][-1]
+{'audio': {'path': '/path/to/punk/01.mp3',
+  'array': array([0.15237972, 0.13222949, 0.10627693, ..., 0.41940814, 0.37578005,
+         0.33717662], dtype=float32),
+  'sampling_rate': 44100},
+ 'label': 1}
+```
+
+If your directory contains multiple splits, use the `data_files` argument to load the images associated with each split:
+
+```py
+>>> from datasets import load_dataset
+
+>>> dataset = load_dataset("audiofolder", data_files={"train":["train/electronic/01.mp3", ...], "test": ["test/electronic/01.mp3", ...)
+```
 
 ## AudioFolder with metadata
 
--- a/docs/source/audio_load.mdx
+++ b/docs/source/audio_load.mdx
@@ -90,7 +90,7 @@ If your directory contains multiple splits, use the `data_files` argument to loa
 ```py
 >>> from datasets import load_dataset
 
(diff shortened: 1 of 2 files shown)


docs/source/image_load.mdx
```diff
@@ -82,25 +90,22 @@ Load remote datasets from their URLs with the `data_files` parameter:
 
 ## ImageFolder with metadata
 
-Metadata associated with your dataset can also be loaded, extending the utility of `ImageFolder` to additional vision tasks like image captioning and object detection. Make sure your dataset has a `metadata.csv` file:
+Metadata associated with your dataset can also be loaded, extending the utility of `ImageFolder` to additional vision tasks like image captioning and object detection. Make sure your dataset has a `metadata.jsonl` file:
```
lewtun, line 93: Hmm, I'm not sure how these changes entered my PR - will take a look (maybe due to change of `master` to `main` a while ago)

docs/source/image_load.mdx
```diff
@@ -72,6 +72,14 @@ Load your dataset by specifying `imagefolder` and the directory of your dataset
 {"image": <PIL.PngImagePlugin.PngImageFile image mode=RGBA size=1200x215 at 0x15E8DAD30>, "label": 1}
 ```
 
+If your directory contains multiple splits, use the `data_files` argument to load the images associated with each split:
+
+```py
+
+>>> dataset = load_dataset("imagefolder", data_files={"train":["train/dog/golden_retriever.png", ...], "test": ["test/dog/german_shepherd.png", ...)
+```
```
NielsRogge, line 81: You can put this below "Load remote datasets from their URLs with the `data_files` parameter:"

docs/source/audio_load.mdx
```diff
@@ -57,7 +57,41 @@ If you only want to load the underlying path to the audio dataset without decodi
 
 ## AudioFolder
 
-You can also load a dataset with an `AudioFolder` dataset builder. It does not require writing a custom dataloader, making it useful for quickly loading audio data.
+You can also load a dataset with an `AudioFolder` dataset builder. It does not require writing a custom dataloader, making it useful for quickly loading audio data. For classification tasks, your audio dataset structure should look like this:
+
+```
+folder/train/electronic/01.mp3
+folder/train/punk/01.mp3
```
NielsRogge, line 64: As @polinaeterna indicated, if you have a "test" subfolder, it will work too, so we might include that for both audio and image folder?

docs/source/audio_load.mdx
```diff
@@ -57,7 +57,41 @@ If you only want to load the underlying path to the audio dataset without decodi
...
+
+>>> dataset["train"][-1]
+{'audio': {'path': '/path/to/punk/01.mp3',
+  'array': array([0.15237972, 0.13222949, 0.10627693, ..., 0.41940814, 0.37578005,
+         0.33717662], dtype=float32),
+  'sampling_rate': 44100},
+ 'label': 1}
+```
+
+If your directory contains multiple splits, use the `data_files` argument to load the images associated with each split:
```
polinaeterna, line 88: in fact it's not needed, the directory structure will be parsed if you provide data directory to `data_dir` too :) 
in case if different metadata files have different set of features, it would anyway throw an error (this is intentional), so the only way to load such cases is to load them separately (with different `load_dataset` i mean). if you think it's an incorrect behavior, feel free to open an issue and ping me and @mariosasko and @lhoestq, we'll discuss it :) 


note aside - you can also use `glob` style in data files: `data_files={"train":["train/electronic/**/*.mp3"]}`

docs/source/image_load.mdx
polinaeterna: no need for this paragraph here too, see my comment for audio_load.mdx :)

docs/source/audio_load.mdx
polinaeterna: yes, might be worth adding that 
```
folder/train/electronic/01.mp3
folder/train/punk/01.mp3
...
folder/test/electronic/09.mp3
folder/test/punk/09.mp3
```
would also work.

docs/source/audio_load.mdx
lewtun: Thanks for the explanation! I think the behaviour makes sense, but I've added a small note here to help future users understand what to do: https://github.com/huggingface/datasets/pull/4925/files#diff-80eb88d203f3d89e6416bf139bf3caef42021d5953261f526fc656b9871325c9R148

lewtun:
Thanks for the feedback @polinaeterna ! I've reworded the docs a bit to integrate your comments and this should be ready for another review :)

docs/source/image_load.mdx
```diff
@@ -72,6 +72,16 @@ Load your dataset by specifying `imagefolder` and the directory of your dataset
...
 ```
 
+You can also use `imagefolder` to load datasets involving multiple splits. To do so, your dataset directory should have the following structure:
+
+```
+folder/train/dog/golden_retriever.png
+folder/train/cat/maine_coon.png
+
+folder/test/dog/german_shepherd.png
+folder/test/cat/bengal.png
```
NielsRogge, line 82: Why not directly include this when talking about "Your image dataset structure should look like this:" above?

Just my two cents :)

docs/source/audio_load.mdx
polinaeterna: lol there is actually a separate page in the documentation about how to structure a repo to include many splits. sorry for forgetting about it 🤦‍♀️ 
I think it definitely makes sense to put the link to it somewhere here:
https://huggingface.co/docs/datasets/v2.4.0/en/repository_structure#structure-your-repository

polinaeterna (commented):
@lewtun thank you for your changes! 
I've just realized that there is another PR about audio documentation open: https://github.com/huggingface/datasets/pull/4872
and there the more detailed description on how to use AudioFolder is moved to another section (Create an audio dataset)
I feel like it might be a pain to integrate changes from both of your PR and that one in the future. Should we maybe first merge the other one and then integrate your note? or do you think it's reasonable to add your comments to that PR directly? I'm gonna review it today btw :)

lewtun:
> I've just realized that there is another PR about audio documentation open: #4872
> and there the more detailed description on how to use `audiofolder` is moved to another section ("Create an audio dataset")

Ah yes, let's add a comment to #4872 - that will be simpler than the alternatives :)

mariosasko:
@polinaeterna @lhoestq What do you think about adding support for the metadata format from Kaggle (one metadata file for each split with the name equal to the split name) to ImageFolder/AudioFolder? I also think we can relax some requirements a bit by:
* allowing `filename` as the name of the main metadata column (currently, only `file_path` is allowed)
* not requiring that the features of all the given metadata files are equal. Instead, we can have a soft check by using `_check_if_features_can_be_aligned` + `_align_features`. The rationale is that train/val metadata often has extra columns compared to test metadata.

These changes would allow us to load the Kaggle dataset linked in the forum thread without any "interventions".

lhoestq:
Can you give more details about the Kaggle format ? I'm down to discuss it in a separate issue if you don't mind.

> allowing filename as the name of the main metadata column (currently, only file_path is allowed)

`filename` refers to the name of the file, so there's no logic about relative path or directories. If I recall correctly this is what we're doing right now so why not

> not requiring that the features of all the given metadata files are equal. Instead, we can have a soft check by using _check_if_features_can_be_aligned + _align_features. The rationale is that train/val metadata often has extra columns compared to test metadata.

+1 and we can set to None the missing features

mariosasko:
I'm not sure if this is worth opening a new issue :).

What I mean by the Kaggle format is the structure like this one (the name of a metadata file is equal to the directory it "references"):
```
- train
    - img1.jpeg
    - img2.jpeg
    - ...
- test
    - img1.jpeg
    - img2.jpeg
    - ...   
- train.csv
- test.csv
```

polinaeterna:
@mariosasko +1 allowing different features set and metadata filenames corresponding to split names

Considering filename column - right now it's even called `file_name` now, which is not nice because in fact it's a relative file path indeed, so I think it should be `file_path` (and I don't know why I haven't thought about it before the release...)

polinaeterna:
@lewtun don't you mind if I close this pull request as I've integrated your changes in https://github.com/huggingface/datasets/pull/4872 ? (it doesn't have a link to a kaggle example though)

Issue #4876, opened 2022-08-23
Move DatasetInfo from `datasets_infos.json` to the YAML tags in `README.md`


lhoestq:
Currently there are two places to find metadata for datasets:
- datasets_infos.json, which contains **per dataset config**
  - description
  - citation
  - license
  - splits and sizes
  - checksums of the data files
  - feature types
  - and more
- YAML tags, which contain
  - license
  - language
  - train-eval-index
  - and more

It would be nice to have a single place instead. We can rely on the YAML tags more than the JSON file for consistency with models. And it would all be indexed by our back-end directly, which is nice to have. 

One way would be to move everything to the YAML tags except the checksums (there can be tens of thousands of them). The description/citation is already in the dataset card so we probably don't need to have them in the YAML card, it would be redundant.

Here is an example for SQuAD
```yaml

download_size: 35142551
dataset_size: 89789763
version: 1.0.0
splits:
- name: train
  num_examples: 87599
  num_bytes: 79317110
- name: validation
  num_examples: 10570
  num_bytes: 10472653
features:
- name: id
  dtype: string
- name: title
  dtype: string
- name: context
  dtype: string
- name: question
  dtype: string
- name: answers
  struct:
  - name: text
    list:
      dtype: string
  - name: answer_start
    list:
      dtype: int32
```

Since there is only one configuration for SQuAD, this structure is ok. For datasets with several configs we can see in a second step, but IMO it would be ok to have these fields per config using another syntax
```yaml
configs:
- config: unlabeled
  splits:
  - name: train
    num_examples: 10000
  features:
  - name: text
    dtype: string
- config: labeled
  splits:
  - name: train
    num_examples: 100
  features:
  - name: text
    dtype: string
  - name: label
    dtype: ClassLabel
    names:
    - negative
    - positive
```
So in the end you could specify a YAML tag either at the top level (for all configs) or per config in the `configs` field

Alternatively we could keep config specific stuff in the `dataset_infos.json` as it it today

Not sure yet what's the best approach here but cc @julien-c @mariosasko @albertvillanova @polinaeterna for feedback :)

julien-c:
also @osanseviero @Pierrci @SBrandeis potentially

SBrandeis:
Love this in principle 🚀 

Let's keep in mind users might rely on `dataset_infos.json` already.

[...]

PR #4926, opened 2022-09-02
Dataset infos in yaml

lhoestq:
To simplify the addition of new datasets, we'd like to have the dataset infos in the YAML and deprecate the dataset_infos.json file. YAML is readable and easy to edit, and the YAML metadata of the readme already contain dataset metadata so we would have everything in one place.

To be more specific, I moved these fields from DatasetInfo to the YAML:
- config_name (if there are several configs)
- download_size
- dataset_size
- features
- splits

Here is what I ended up with for `squad`:
```yaml
dataset_info:
  features:
  - name: id
    dtype: string
  - name: title
    dtype: string
  - name: context
    dtype: string
  - name: question
    dtype: string
  - name: answers
    sequence:
    - name: text
      dtype: string
    - name: answer_start
      dtype: int32
  splits:
  - name: train
    num_bytes: 79346360
    num_examples: 87599
  - name: validation
    num_bytes: 10473040
    num_examples: 10570
  config_name: plain_text
  download_size: 35142551
  dataset_size: 89819400
```

and it can be a list if there are several configs

I already did the change for `conll2000` and `crime_and_punish` as an example.

### Load/Read

This is done via `DatasetInfosDict.write_to_directory/from_directory`

I had to implement a custom the YAML export logic for `SplitDict`, `Version` and `Features`.
The first two are trivial, but the logic for `Features` is more complicated, because I added a simplification step (or the YAML would be too long and less readable): it's just a formatting step to remove unnecessary nesting of YAML data.

### Other changes

I had to update the DatasetModule factories to also download the README.md alongside the dataset scripts/data files, and not just the dataset_infos.json

## YAML validation

I removed the old validation code that was in metadata.py, now we can just use the Hub YAML validation

## Datasets-cli

The `datasets-cli test --save_infos` command now creates a README.md file with the dataset_infos in it, instead of a datasets_infos.json file

## Backward compatibility

`dataset_infos.json` files are still supported and loaded if they exist to have full backward compatibility.
Though I removed the unnecessary keys when the value is the default (like all the `id: null` from the Value feature types) to make them easier to read.

## TODO

- [x] add comments
- [x] tests
- [x] document the new YAML fields 
- [x] try to reload the new dataset_infos.json file content with an old version of `datasets`

## EDITS

- removed "config_name" when there's only one config
- removed "version" for now (?), because it's not useful in general
- renamed the yaml field "dataset_info" instead of "dataset_infos", since it only has one by default (and because "infos" is not english)

Fix https://github.com/huggingface/datasets/issues/4876


datasets/conll2000/README.md
```diff
@@ -3,6 +3,98 @@ language:
 - en
 paperswithcode_id: conll-2000-1
 pretty_name: CoNLL-2000
+dataset_infos:
+- config_name: conll2000
```
julien-c, line 7: do we need the (non-default) config_name for backward compatibility?

lhoestq: The dataset explicitly defines one configuration with this name, but since there's only one there's no ambiguity. Let me tweak the YAML loading a bit and we can remove this line :)

lhoestq: I removed config name for conll2000 in the YAML part, and for everything to connect smoothly I had to remove the definition of the config in the dataset script itself. I'll see if it's something I can do for all the datasets that have one configuration

lhoestq:
Alright this is ready for review :)
I mostly would like your opinion on the YAML structure and what we can do in the docs (IMO we can add the docs about those fields in the Hub docs). Other than that let me know if the changes in info.py and features.py look good to you

julien-c:
LGTM and looking forward to having this merged!! ❤️

lhoestq:
We plan to do a release today, we'll merge this after the release :)

src/datasets/utils/py_utils.py
```diff
@@ -167,7 +167,8 @@ def _asdict_inner(obj):
             result = {}
             for f in fields(obj):
                 value = _asdict_inner(getattr(obj, f.name))
-                result[f.name] = value
+                if not f.init or value != f.default or f.metadata.get("include_in_asdict_even_if_is_default", False):
+                    result[f.name] = value
```
lhoestq, line 171: To simplify the JSON and YAML dumps, I stop dumping the dataclass attributes that have are the default value.

e.g. decode=True for Image, or length=-1 for Sequence

lhoestq (commented):
I modified the class_label YAML dump structure to show the label ids. This is more practical this way, see conll2000.

This is ready for review now ! :)

lhoestq:
Created https://github.com/huggingface/datasets/pull/5018 where I added the YAML `dataset_info` of every single dataset in this repo

mariosasko (commented):
Great job! 

One nit: the metadata generation in `push_to_hub` also needs to be updated, no?

PS: We should also probably use `DatasetInfo` from `hugginface_hub` instead of having our own implementation in `info.py`, but this can be addressed later.

docs/source/dataset_script.mdx
```diff
@@ -274,17 +274,17 @@ def _generate_examples(self, filepath):
...
 
-Run the following command to generate your dataset metadata in `dataset_infos.json` and make sure your new dataset loading script works correctly:
+Run the following command to generate your dataset metadata in `README.md` and make sure your new dataset loading script works correctly:
 
 ```
 datasets-cli test path/to/<your-dataset-loading-script> --save_infos --all_configs
 ```
 
-If your dataset loading script passed the test, you should now have a `dataset_infos.json` file in your dataset folder.
+If your dataset loading script passed the test, you should now have a `README.md` file in your dataset folder containing a `dataset_infos` field with some metadata.
```
polinaeterna, line 287:
```suggestion
If your dataset loading script passed the test, you should now have a `README.md` file in your dataset folder containing a `dataset_info` field with some metadata.
```

.github/PULL_REQUEST_TEMPLATE/add_dataset.md
```diff
@@ -6,11 +6,9 @@
 
 ### Checkbox
 
-- [ ] Create the dataset script `/datasets/my_dataset/my_dataset.py` using the template
+- [ ] Create the dataset script `./my_dataset/my_dataset.py` using the template
 - [ ] Fill the `_DESCRIPTION` and `_CITATION` variables
 - [ ] Implement `_infos()`, `_split_generators()` and `_generate_examples()`
```
polinaeterna, line 11:
```suggestion
- [ ] Implement `_info()`, `_split_generators()` and `_generate_examples()`
```

polinaeterna (commented):
love this change! i added a few suggestions/fixes for documentation :)

src/datasets/commands/test.py
```diff
@@ -50,7 +50,9 @@ def register_subcommand(parser: ArgumentParser):
             help="Can be used to specify a manual directory to get the files from.",
         )
         test_parser.add_argument("--all_configs", action="store_true", help="Test all dataset configurations")
-        test_parser.add_argument("--save_infos", action="store_true", help="Save the dataset infos file")
+        test_parser.add_argument(
+            "--save_infos", action="store_true", help="Save the dataset infos in the dataset card (README.md)"
```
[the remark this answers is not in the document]
lhoestq: Good idea ! I changed to --save_info and kept --save_infos as an alias

lhoestq:
Tooks your comments into account and updated `push_to_hub` to push the dataset_info to the README.md instead of json :) Let me know if it sounds good to you now !

PR #4928, opened 2022-09-03
Add ability to read-write to SQL databases.

### LINKED ISSUE #3094: Support loading a dataset from SQLite files
As requested by @julien-c, we could eventually support loading a dataset from SQLite files, like it is the case for JSON/CSV files.

Dref360:
Fixes #3094 

Add ability to read/write to SQLite files and also read from any SQL database supported by SQLAlchemy.

I didn't add SQLAlchemy as a dependence as it is fairly big and it remains optional. 

I also recorded a Loom to showcase the feature.

https://www.loom.com/share/f0e602c2de8a46f58bca4b43333d541f


Dref360:
Ah CI runs with `pandas=1.3.5` which doesn't return the number of row inserted.

julien-c:
wow this is super cool!

src/datasets/arrow_dataset.py
```diff
@@ -1057,6 +1058,48 @@ def from_text(
...
+        split: Optional[NamedSplit] = None,
+        features: Optional[Features] = None,
+        cache_dir: str = None,
+        keep_in_memory: bool = False,
+        **kwargs,
+    ):
+        """Create Dataset from text file(s).
+
+        Args:
+            path_or_paths (path-like or list of path-like): Path(s) of the text file(s).
```
lhoestq, line 1074:
```suggestion
            path_or_paths (path-like or list of path-like): Path(s) of the SQLite file(s).
```

src/datasets/arrow_dataset.py
```diff
@@ -4051,6 +4094,38 @@ def to_parquet(
...
+                Defaults to :obj:`datasets.config.DEFAULT_MAX_BATCH_SIZE`.
+            **sql_writer_kwargs (additional keyword arguments): Parameters to pass to pandas's :function:`Dataframe.to_sql`
+
+        Returns:
+            int: The number of characters or bytes written
+
+        Example:
+
+        ```py
+        >>> ds.to_sql("path/to/dataset/directory")
```
lhoestq, line 4119: I think you can also provide an example with sqlalchemy here in the docstring

src/datasets/packaged_modules/sql/sql.py
```diff
@@ -0,0 +1,100 @@
...
+                # more expensive cast; allows str <-> int/float or str to Audio for example
+                pa_table = table_cast(pa_table, schema)
+        return pa_table
+
+    def _generate_tables(self, files):
+        for file_idx, file in enumerate(itertools.chain.from_iterable(files)):
+            with contextlib.closing(connect(file)) as conn:
+                sql_file_reader = pd.read_sql(
+                    f"SELECT * FROM `{self.config.table_name}`", conn, **self.config.read_sql_kwargs
+                )
```
lhoestq, line 89: Do you know if this loads the full database into memory before returning chunks from it ?

tests/io/test_sql.py
```diff
@@ -0,0 +1,176 @@
...
+        None,
+        {"col_1": "string", "col_2": "int64", "col_3": "float64"},
+        {"col_1": "string", "col_2": "string", "col_3": "string"},
+        {"col_1": "int32", "col_2": "int32", "col_3": "int32"},
+        {"col_1": "float32", "col_2": "float32", "col_3": "float32"},
+    ],
+)
+def test_dataset_from_sql_features(features, sql_path, tmp_path):
+    cache_dir = tmp_path / "cache"
+    # CSV file loses col_1 string dtype information: default now is "int64" instead of "string"
```
lhoestq, line 49:
```suggestion
```

tests/io/test_sql.py
```diff
@@ -0,0 +1,176 @@
...
+        None,
+        {"col_1": "string", "col_2": "int64", "col_3": "float64"},
+        {"col_1": "string", "col_2": "string", "col_3": "string"},
+        {"col_1": "int32", "col_2": "int32", "col_3": "int32"},
+        {"col_1": "float32", "col_2": "float32", "col_3": "float32"},
+    ],
+)
+def test_sql_datasetdict_reader_features(features, sql_path, tmp_path):
+    cache_dir = tmp_path / "cache"
+    # CSV file loses col_1 string dtype information: default now is "int64" instead of "string"
```
lhoestq, line 114:
```suggestion
```

lhoestq (commented):
Nice thank you !

Also feel free to add a section in the documentation about this amazing feature !
I think you can add it to `loading.mdx` right after the Parquet section (rendered [here](https://huggingface.co/docs/datasets/loading#parquet)) here:

https://github.com/huggingface/datasets/blob/5389985191b11b3b11871aac554d44a7b5ec438f/docs/source/loading.mdx#L180-L184

src/datasets/packaged_modules/sql/sql.py
Dref360: From what I see [here](https://github.com/pandas-dev/pandas/blob/v1.4.4/pandas/io/sql.py#L2050) it doesn't load everything, it moves the DB cursor for `chunksize` rows and yield the batch. So it should Just Work :tm: .

Dref360:
@lhoestq I'm getting error in integration tests, not sure if it's related to my PR. Any help would be appreciated :) 

```
if not self._is_valid_token(token):
>               raise ValueError("Invalid token passed!")
E               ValueError: Invalid token passed!
```

lhoestq:
I just relaunched the tests, it should be fixed now

mariosasko:
Thanks a lot for working on this!

I have some concerns with the current design:
* Besides SQLite, the loader should also work with the other engines supported by SQLAlchemy. (A better name for it in the current state would be `sqlite` :))
* It should support arbitrary queries/table names - only the latter currently works.
* Exposing this loader as a packaged builder (`load_dataset("sql", ...)`) is not a good idea for the following reasons:
    * Considering the scenario where a table with the same name is present in multiple files is very unlikely, the data files resolution is not needed here
    * `load_dataset("sql", ...)` also implies that streaming should work, but that's not the case.  And I don't think we can change that, considering how hard it is to make SQLite files streamable.

All this makes me think we shouldn't expose this builder as a packaged module and, instead, limit the API to `Dataset.from_sql`/`Dataset.to_sql` (with the signatures matching the ones in pandas as much as possible; regarding this, note that SQLAlchemy connections are not hashable/picklable, which is required for caching, but I think it's OK only to allow URI strings as connections to bypass that (Dask has the same limitation).

WDYT?

mariosasko:
Perhaps after we merge https://github.com/huggingface/datasets/pull/4957, you can subclass `AbstractDatasetInputStream` instead of `AbstractDatasetReader` to not break the contract with the connection object. Also, let's avoid having the default value for the query/table (you can set it to `None` in the builder and raise an error in the builder config's `__post_init__` if it's not provided). Other than that, sounds good!

src/datasets/arrow_dataset.py
```diff
@@ -1091,6 +1092,45 @@ def from_text(
             path_or_paths, split=split, features=features, cache_dir=cache_dir, keep_in_memory=keep_in_memory, **kwargs
         ).read()
 
+    @staticmethod
+    def from_sql(
+        path_or_paths: Union[PathLike, List[PathLike]],
```
Dref360, line 1097: Should this be renamed to `conn` to match pandas or stay as `path_or_paths` to match other `from_*` function?

src/datasets/io/sql.py
```diff
@@ -0,0 +1,136 @@
...
+            download_mode=download_mode,
+            ignore_verifications=ignore_verifications,
+            # try_from_hf_gcs=try_from_hf_gcs,
+            base_path=base_path,
+            use_auth_token=use_auth_token,
+        )
+
+        # Build dataset for splits
+        dataset = self.builder.as_dataset(
+            split="train", ignore_verifications=ignore_verifications, in_memory=self.keep_in_memory
```
Dref360, line 53: Same as `GeneratorDatasetInputStream` where only the train set is possible. let me know if that's alright.

src/datasets/arrow_dataset.py
mariosasko: `con` makes the most sense to be aligned with Pandas/Dask

Dref360:
Thank you so much! I was missing a lot of things sorry about that.
LGTM

src/datasets/packaged_modules/sql/sql.py
```diff
@@ -0,0 +1,97 @@
...
+        if not isinstance(sql, str):
+            if not datasets.config.SQLALCHEMY_AVAILABLE:
+                raise ImportError("Please pip install sqlalchemy.")
+
+            import sqlalchemy
+
+            config_kwargs = config_kwargs.copy()
+            engine = sqlalchemy.create_engine(config_kwargs["con"].split("://")[0] + "://")
+            sql_str = str(sql.compile(dialect=engine.dialect))
+            config_kwargs["sql"] = sql_str
```
lhoestq, line 56: With something like this you can even raise a TypeError if the object is neither a string or a Selectable
```suggestion
            if datasets.config.SQLALCHEMY_AVAILABLE and "sqlalchemy" in sys.modules::
                import sqlalchemy
                
                if isinstance(sql, sqlalchemy.sql.Selectable):
                    config_kwargs = config_kwargs.copy()
                    engine = sqlalchemy.create_engine(config_kwargs["con"].split("://")[0] + "://")
                    sql_str = str(sql.compile(dialect=engine.dialect))
                    config_kwargs["sql"] = sql_str
                else:
                    raise TypeError(f"Supported types for 'sql' are string and sqlalchemy.sql.Selectable but got {type(sql)}: {sql}")
            else:
                raise TypeError(f"Supported types for 'sql' are string and sqlalchemy.sql.Selectable but got {type(sql)}: {sql}")
```

lhoestq (commented):
Wow super impressive, thanks @Dref360 and @mariosasko !

It looks super nice, I just left some minor comments

src/datasets/arrow_dataset.py
```diff
@@ -111,6 +111,10 @@
 
 
 if TYPE_CHECKING:
+    import sqlite3
+
+    import sqlalchemy
```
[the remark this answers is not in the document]
mariosasko: Yea, I don't think this is needed. `TYPE_CHECKING` is equal to `False` anyways at runtime, so adding more complexity to it doesn't seem like a good idea. (also see this [SO thread](https://stackoverflow.com/a/61394121))

mariosasko:
I think we can merge if the tests pass. 

One last thing I would like to get your opinion on - currently, if SQLAlchemy is not installed, the missing dependency error will be thrown inside `pandas.read_sql`. Do you think we should be the ones throwing this error, e.g. after the imports in `packaged_modules/sql/sql.py` if `SQLALCHEMY_AVAILABLE` is `False` (note that this would mean making `sqlalchemy` a required dependency for the docs to be able to add `SqlConfig` to the package reference)?

lhoestq:
> One last thing I would like to get your opinion on - currently, if SQLAlchemy is not installed, the missing dependency error will be thrown inside pandas.read_sql

Is sqlalchemy always required for pd.read_sql ? If so, I think we can raise the error on our side.
But sqlalchemy should still be an optional dependency for `datasets` IMO

mariosasko:
@lhoestq 
> Is sqlalchemy always required for pd.read_sql ? If so, I think we can raise the error on our side.

In our case, it's always required as we only support database URIs.

> But sqlalchemy should still be an optional dependency for datasets IMO

Yes, it will remain optional for datasets but will be required for building the docs (as is`s3fs`, for instance).

lhoestq:
Ok I see ! Sounds good :)

docs/source/loading.mdx
```diff
@@ -196,6 +196,24 @@ To load remote Parquet files via HTTP, pass the URLs instead:
 >>> wiki = load_dataset("parquet", data_files=data_files, split="train")
 ```
 
+### SQL
```
lhoestq, line 199: Maybe mention that users need to install `sqlalchemy` ?

lhoestq (approved):
Thanks !

docs/source/loading.mdx
```diff
@@ -196,6 +196,24 @@ To load remote Parquet files via HTTP, pass the URLs instead:
...
+
+Read database contents with with [`Dataset.from_sql`]. Both table names and queries are supported.
```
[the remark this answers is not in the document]
mariosasko: I decided to add a tip to the `from_sql` docstring instead, but thanks anyways :).

PR #4887, opened 2022-08-24
Add "cc-by-nc-sa-2.0" to list of licenses 

osanseviero:
Datasets side of https://github.com/huggingface/hub-docs/pull/285

PR #4930, opened 2022-09-05
Add cc-by-nc-2.0 to list of licenses

albertvillanova:
This PR adds the `cc-by-nc-2.0` to the list of licenses because it is required by `scifact` dataset: https://github.com/allenai/scifact/blob/master/LICENSE.md

### DIFF
--- a/src/datasets/utils/resources/standard_licenses.tsv
+++ b/src/datasets/utils/resources/standard_licenses.tsv
@@ -16,6 +16,7 @@ Creative Commons Attribution 3.0	cc-by-3.0
 Creative Commons Attribution 4.0	cc-by-4.0
 Creative Commons Attribution Share Alike 3.0	cc-by-sa-3.0
 Creative Commons Attribution Share Alike 4.0	cc-by-sa-4.0
+Creative Commons Attribution Non Commercial 2.0	cc-by-nc-2.0
 Creative Commons Attribution Non Commercial 3.0	cc-by-nc-3.0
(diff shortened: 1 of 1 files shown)


julien-c:
this list needs to be kept in sync with the ones in moon-landing and hub-docs :)

albertvillanova:
@julien-c don't you think it might be better to a have a single file (source of truth) in one of the repos and then use it in every other repo, instead of having 3 copies of the same file that must be kept in sync?

Also note that the licenses we are adding were all already present in our previous `license.json` file: are we regenerating it, step by step? Why don't we use a file with ALL the licenses we previously had in the list?

Licenses added:
- #4887
- #4930 

Previous `license.json` file:
- https://github.com/huggingface/datasets/blob/b7612754928e0fd43b9e3c3becb906ec280ff5d4/src/datasets/utils/resources/licenses.json

Let me know what you think and I can take care of this.

julien-c:
> Let me know what you think and I can take care of this.

What I think is that we shouldn't add licenses that are just used in a couple of datasets, and just use `license_details` for this.

> don't you think it might be better to a have a single file (source of truth) in one of the repos and then use it in every other repo, instead of having 3 copies of the same file that must be kept in sync?

Yes, in my opinion we can just delete this file from `datasets`, the validation is happening hub-side anyways now?

lhoestq:
Feel free to delete the license list in `datasets` @albertvillanova ;)

Also FYI in #4926 I also removed all the validation steps anyway (language, license, types etc.)

PR #4937, opened 2022-09-06
Remove deprecated identical_ok

lhoestq:
`huggingface-hub` says that the `identical_ok` argument of `HfApi.upload_file` is now deprecated, and will be removed soon. It even has no effect at the moment when it's passed:

```python
Args:
...
    identical_ok (`bool`, *optional*, defaults to `True`):
        Deprecated: will be removed in 0.11.0.
        Changing this value has no effect.
...
```

There was only one occurence of `identical_ok=False` but it's maybe not worth adding a check ti verify if the files were the same.

cc @mariosasko


mariosasko (approved):
`identical_ok=False` is a sanity check introduced in https://github.com/huggingface/datasets/pull/4402, so I agree we don't need an extra check. Thanks!

PR #4940, opened 2022-09-06
Fix multilinguality tag and missing sections in xquad_r dataset card

albertvillanova:
This PR fixes issue reported on the Hub:
- Label as multilingual: https://huggingface.co/datasets/xquad_r/discussions/1


lhoestq (approved):
good catch !

Issue #4417, opened 2022-05-29
how to convert a dict generator into a huggingface dataset. 


StephennFernandes:
### Description

Hey there, I have used seqio to get a well distributed mixture of samples from multiple dataset. However the resultant output from seqio is a python generator dict, which I cannot produce back into huggingface dataset.

The generator contains all the samples needed for training the model but I cannot convert it into a huggingface dataset.

The code looks like this:
Hey there, I have used seqio to get a well distributed mixture of samples from multiple dataset. However the resultant output from seqio is a python generator dict, which I cannot produce back into huggingface dataset.

The generator contains all the samples needed for training the model but I cannot convert it into a huggingface dataset.

The code looks like this:

```
for ex in seqio_data:
print(ex[“text”])
```

I need to convert the seqio_data (generator) into huggingface dataset.

the complete seqio code goes here:
```
import functools

import seqio
import tensorflow as tf
import t5.data
from datasets import load_dataset
from t5.data import postprocessors
from t5.data import preprocessors
from t5.evaluation import metrics
from seqio import FunctionDataSource, utils

TaskRegistry = seqio.TaskRegistry



def gen_dataset(split, shuffle=False, seed=None, column="text", dataset_params=None):
    dataset = load_dataset(**dataset_params)
    if shuffle:
        if seed:
            dataset = dataset.shuffle(seed=seed)
        else:
            dataset = dataset.shuffle()
    while True:
        for item in dataset[str(split)]:
            yield item[column]


def dataset_fn(split, shuffle_files, seed=None, dataset_params=None):
    return tf.data.Dataset.from_generator(
        functools.partial(gen_dataset, split, shuffle_files, seed, dataset_params=dataset_params),
        output_signature=tf.TensorSpec(shape=(), dtype=tf.string, name=dataset_name)
    )


@utils.map_over_dataset
def target_to_key(x, key_map, target_key):
    """Assign the value from the dataset to target_key in key_map"""
    return {**key_map, target_key: x}



dataset_name = 'oscar-corpus/OSCAR-2109'
subset= 'mr'
dataset_params = {"path": dataset_name, "language":subset, "use_auth_token":True}
dataset_shapes = None

TaskRegistry.add(
    "oscar_marathi_corpus",
    source=seqio.FunctionDataSource(
        dataset_fn=functools.partial(dataset_fn, dataset_params=dataset_params),
        splits=("train", "validation"),
        caching_permitted=False,
        num_input_examples=dataset_shapes,
    ),
[...]

PR #4957, opened 2022-09-08
Add `Dataset.from_generator`

mariosasko:
Add `Dataset.from_generator` to the API to allow creating datasets from data larger than RAM. The implementation relies on a packaged module not exposed in `load_dataset` to tie this method with `datasets`' caching mechanism.

Closes https://github.com/huggingface/datasets/issues/4417


lhoestq:
I restarted the builder PR job just in case

mariosasko:
CI is now green. https://github.com/huggingface/doc-builder/pull/296 explains why it failed.

docs/source/loading.mdx
```diff
@@ -220,6 +220,22 @@ Load a list of Python dictionaries with [`~Dataset.from_list`]:
...
+### Python generator
+
+Create a dataset from a Python generator with [`~Dataset.from_generator`]:
+
+```py
+>>> from datasets import Dataset
+>>> def my_gen():
+...     yield {"a": 1}
+...     yield {"a": 2}
+...     yield {"a": 3}
```
lhoestq, line 232: (nit) you can use a for loop instead, most users will probably use a for loop IMO

src/datasets/packaged_modules/generator/generator.py
```diff
@@ -0,0 +1,33 @@
+from dataclasses import dataclass
+from typing import Callable, Optional
+
+import datasets
+
+
+@dataclass
+class GeneratorConfig(datasets.BuilderConfig):
+    generator: Optional[Callable] = None
```
lhoestq, line 9: Maybe you don't need to set None by default ?

src/datasets/arrow_dataset.py
```diff
@@ -939,6 +939,45 @@ def from_csv(
...
+        generator: Callable,
+        features: Optional[Features] = None,
+        cache_dir: str = None,
+        keep_in_memory: bool = False,
+        gen_kwargs: Optional[dict] = None,
+    ):
+        """Create a Dataset from a generator.
+
+        Args:
+            generator (:obj:`Callable`): A callable object that returns an object that supports the `iter` protocol.
```
lhoestq, line 953: (nit) suggestion for the docstring
```suggestion
            generator (:obj:`Callable`): A generator function that `yields` examples.
```

lhoestq (approved):
Awesome thank you !

Maybe let's open an issue to also add IterableDataset.from_generator ?

src/datasets/packaged_modules/generator/generator.py
mariosasko: I would expect this to work, but it doesn't due to the class inheritance (see https://bugs.python.org/issue39300 for the explanation)

Issue #4858, opened 2022-08-16
map() function removes columns when input_columns is not None


pramodith:
## Describe the bug
The map function, removes features from the dataset that are not present in the _input_columns_ list of columns, despite the columns being removed not mentioned in the _remove_columns_ argument.
## Steps to reproduce the bug
```python
from datasets import Dataset
ds = Dataset.from_dict({"a" : [1,2,3],"b" : [0,1,0], "c" : [2,4,5]})

def double(x,y):
  x = x*2
  y = y*2
  return {"d" : x, "e" : y}

ds.map(double, input_columns=["a","c"])
```

## Expected results
```
Dataset({
    features: ['a', 'b', 'c', 'd', 'e'],
    num_rows: 3
})
```
## Actual results
```
Dataset({
    features: ['a', 'c', 'd', 'e'],
    num_rows: 3
})
```

In this specific example feature **b** should not be removed.

## Environment info

- `datasets` version: 2.4.0
- Platform: linux (colab)
- Python version: 3.7.13
- PyArrow version: 6.0.1

mariosasko:
Hi! Thanks for reporting! This looks like a bug. I've just opened a PR with the fix.

pramodith:
Awesome! Thank you. I'll close the issue once the PR gets merged. :-)

PR #4971, opened 2022-09-12
Preserve non-`input_colums` in `Dataset.map` if `input_columns` are specified

mariosasko:
Currently, if the `input_columns` list in `Dataset.map` is specified, the columns not in that list are dropped after the `map` transform.

This makes the behavior inconsistent with `IterableDataset.map`.
 
(It seems this issue was introduced by mistake in https://github.com/huggingface/datasets/pull/2246) 

Fix https://github.com/huggingface/datasets/issues/4858

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -2666,7 +2666,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                 validate_function_output(processed_inputs, indices)
             if not update_data:
                 return None  # Nothing to update, let's move on
-            if self._format_type is not None:
+            if self._format_type is not None or input_columns:
                 inputs = self._getitem(
                     key=(indices if isinstance(indices, int) else slice(indices[0], indices[-1] + 1)),
                     format_type=None,
@@ -2731,7 +2731,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                 # Only load the columns we actually need
                 if input_columns:
                     input_dataset = self.with_format(
-                        self._format_type, columns=input_columns, output_all_columns=True, **self._format_kwargs
+                        self._format_type, columns=input_columns, output_all_columns=False, **self._format_kwargs
                     )
                     if remove_columns:
                         remove_columns = list(set(remove_columns) & set(input_columns))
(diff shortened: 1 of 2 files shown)


tests/test_arrow_dataset.py
```diff
@@ -1341,6 +1341,18 @@ def func(example):
                     )
                     self.assertListEqual(dset_test[0]["tensor"], [1, 2, 3])
 
+    def test_map_input_columns(self, in_memory):
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            with self._create_dummy_dataset(in_memory, tmp_dir) as dset:
+                print("dset", dset[0])
```
albertvillanova, line 1347: We could remove this print:
```suggestion
```

albertvillanova (approved):
Good catch. Thanks.

tests/test_arrow_dataset.py
```diff
@@ -1341,6 +1341,18 @@ def func(example):
                     )
                     self.assertListEqual(dset_test[0]["tensor"], [1, 2, 3])
 
+    def test_map_input_columns(self, in_memory):
```
lhoestq, line 1344: I think this test is already passing if `output_all_columns=False`

You may need to add another column than "filename" to have it properly tested

src/datasets/arrow_dataset.py
```diff
@@ -2731,7 +2731,7 @@ def init_buffer_and_writer():
                 # Only load the columns we actually need
                 if input_columns:
                     input_dataset = self.with_format(
-                        self._format_type, columns=input_columns, output_all_columns=False, **self._format_kwargs
+                        self._format_type, columns=input_columns, output_all_columns=True, **self._format_kwargs
```
lhoestq, line 2734: If we do this I think we can remove this part in `apply_function_on_filtered_inputs` no ? Otherwise the data would be loaded twice if a format type is set

```python
            if self._format_type is not None:
                inputs = self._getitem(
                    key=(indices if isinstance(indices, int) else slice(indices[0], indices[-1] + 1)),
                    format_type=None,
                    format_columns=None,
                    format_kwargs=None,
                    decoded=False,
                )
```

lhoestq (commented):
Thanks !

src/datasets/arrow_dataset.py
mariosasko: We still need this for the scenario where `input_columns` is `None` and self._format_type is not `None` (to ignore the formatting)

mariosasko: Okay, now `output_all_columns` is still `False` if `input_columns` is specified, but we also execute the above block to add the non-`input_columns` .

lhoestq (approved):
LGTM !

PR #4972, opened 2022-09-13
Fix map batched with torch output

lhoestq:
Reported in https://discuss.huggingface.co/t/typeerror-when-applying-map-after-set-format-type-torch/23067/2

Currently it fails  if one uses batched `map` and the map function returns a torch tensor.

I fixed it for torch, tf, jax and pandas series.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -22,6 +22,7 @@ import json
 import os
 import re
 import shutil
+import sys
 import tempfile
 import warnings
 import weakref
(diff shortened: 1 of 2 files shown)


src/datasets/arrow_dataset.py
```diff
@@ -2651,7 +2652,19 @@ def validate_function_output(processed_inputs, indices):
                     f"Provided `function` which is applied to all elements of table returns a variable of type {type(processed_inputs)}. Make sure provided `function` returns a variable of type `dict` (or a pyarrow table) to update the dataset or `None` if you are only interested in side effects."
                 )
             elif isinstance(indices, list) and isinstance(processed_inputs, Mapping):
-                allowed_batch_return_types = (list, np.ndarray)
+                allowed_batch_return_types = (list, np.ndarray, pd.Series)
+                if config.TF_AVAILABLE and "tensorflow" in sys.modules:
```
albertvillanova, line 2656: Maybe a silly question: is it needed to check that the module has been previously loaded `in sys.modules`?

albertvillanova (approved):
Thanks for the fix @lhoestq. Seems good to me.

src/datasets/arrow_dataset.py
lhoestq: It's more a quality of life. The check is not needed if the module has not been imported.

And this way the user is not suprised to see TF imported even though they're not using it. Especially since importing TF takes several seconds

PR #4973, opened 2022-09-13
[GH->HF] Load datasets from the Hub

lhoestq:
Currently datasets with no namespace (e.g. squad, glue) are loaded from github.

In this PR I changed this logic to use the Hugging Face Hub instead.

This is the first step in removing all the dataset scripts in this repository

related to discussions in https://github.com/huggingface/datasets/pull/4059 (I should have continued from this PR actually)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,6 @@ on:
       - main
 
 env:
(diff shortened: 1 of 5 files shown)


albertvillanova:
Duplicate of:
- #4059

PR #4984, opened 2022-09-16
docs: ✏️ add links to the Datasets API

severo:
I added some links to the Datasets API in the docs. See https://github.com/huggingface/datasets-server/pull/566 for a companion PR in the datasets-server. The idea is to improve the discovery of the API through the docs.

I'm a bit shy about pasting a lot of links to the API in the docs, so it's minimal for now. I'm interested in ideas to integrate the API better in these docs without being too much. cc @lhoestq @julien-c @albertvillanova @stevhliu.

### DIFF
--- a/docs/source/index.mdx
+++ b/docs/source/index.mdx
@@ -6,7 +6,13 @@
 
 Load a dataset in a single line of code, and use our powerful data processing methods to quickly get your dataset ready for training in a deep learning model. Backed by the Apache Arrow format, process large datasets with zero-copy reads without any memory constraints for optimal speed and efficiency. We also feature a deep integration with the [Hugging Face Hub](https://huggingface.co/datasets), allowing you to easily load and share a dataset with the wider machine learning community.
 
-Find your dataset today on the [Hugging Face Hub](https://huggingface.co/datasets), and take an in-depth look inside of it with the live viewer.
+Find your dataset today on the [Hugging Face Hub](https://huggingface.co/datasets), and take an in-depth look inside of it with the [live viewer](https://huggingface.co/datasets/glue/viewer/cola/train).
+
+<Tip>
+
+Are you looking for the [Datasets REST API](https://huggingface.co/docs/datasets-server)? Integrate into your apps over 10,000 datasets via simple **HTTP requests**, with pre-processed responses and scalability built-in.
+
+</Tip>
 
 <div class="mt-10">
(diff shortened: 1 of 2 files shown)


src/datasets/inspect.py
```diff
@@ -303,6 +303,8 @@ def get_dataset_config_names(
      'rte',
      'wnli',
      'ax']
+
+    Note that you can fetch the list of configs for a dataset on the Hugging Face Hub via an HTTP request with the [Datasets REST API endpoint /splits](https://huggingface.co/docs/datasets-server/splits).
```
lhoestq, line 307: I don't think adding this here would bring a lot of users. Maybe prefer adding some links in the actual docs

docs/source/index.mdx
```diff
@@ -6,7 +6,13 @@
...
 Load a dataset in a single line of code, and use our powerful data processing methods to quickly get your dataset ready for training in a deep learning model. Backed by the Apache Arrow format, process large datasets with zero-copy reads without any memory constraints for optimal speed and efficiency. We also feature a deep integration with the [Hugging Face Hub](https://huggingface.co/datasets), allowing you to easily load and share a dataset with the wider machine learning community.
 
-Find your dataset today on the [Hugging Face Hub](https://huggingface.co/datasets), and take an in-depth look inside of it with the live viewer.
+Find your dataset today on the [Hugging Face Hub](https://huggingface.co/datasets), and take an in-depth look inside of it with the [live viewer](https://huggingface.co/datasets/glue/viewer/cola/train).
+
+<Tip>
+
+Are you looking for the [Datasets REST API](https://huggingface.co/docs/datasets-server)? Integrate into your apps over 10,000 datasets via simple **HTTP requests**, with pre-processed responses and scalability built-in.
+
+</Tip>
```
lhoestq, line 15: Adding this tip will bring the four blocks "tutorials" "how-to guides" etc. at the very bottom of the page and users will have to scroll. As it is the first page, it must also be minimal and straight to the point. Therefore I'm not a big fan of changing this page.

`datasets` users will probably mostly be interested in having the preview working. I think we can have a dedicated page in the "Dataset Repository" section about "Dataset Preview" that explains how it works, and possibly redirect to the REST API

lhoestq (commented):
Good idea to link the two documentations ! I added some suggestions

severo:
OK, thanks @lhoestq. I'll close this PR, and come back to it with @stevhliu once we work on https://github.com/huggingface/datasets-server/issues/568

Issue #4895, opened 2022-08-25
load_dataset method returns Unknown split "validation" even if this dir exists


SamSamhuns:
## Describe the bug
The `datasets.load_dataset` returns a `ValueError: Unknown split "validation". Should be one of ['train', 'test'].` when running `load_dataset(local_data_dir_path, split="validation")` even if the `validation` sub-directory exists in the local data path.

The data directories are as follows and attached to this issue:
```
test_data1
              |_ train
                  |_ 1012.png
                  |_ metadata.jsonl
                  ...
              |_ test
                  ...
              |_ validation
                  |_ 234.png
                  |_ metadata.jsonl
                  ...
test_data2
              |_ train
                  |_ train_1012.png
                  |_ metadata.jsonl
                  ...
              |_ test
                  ...
              |_ validation
                  |_ val_234.png
                  |_ metadata.jsonl
                  ...
```

They contain the same image files and `metadata.jsonl` but the images in `test_data2` have the split names prepended i.e.
`train_1012.png, val_234.png` and the images in  `test_data1` do not have the split names prepended to the image names i.e. `1012.png, 234.png`

I actually saw in another issue `val` was not recognized as a split name but here I would expect the files to take the split from the parent directory name i.e. val should become part of the validation split?

## Steps to reproduce the bug
```python
import datasets
datasets.logging.set_verbosity_error()
from datasets import load_dataset, get_dataset_split_names


# the following only finds train, validation and test splits correctly
path = "./test_data1"
print("######################", get_dataset_split_names(path), "######################")

dataset_list = []
for spt in ["train", "test", "validation"]:
    dataset = load_dataset(path, split=spt)
    dataset_list.append(dataset)


# the following only finds train and test splits
path = "./test_data2"
print("######################", get_dataset_split_names(path), "######################")

dataset_list = []
for spt in ["train", "test", "validation"]:
    dataset = load_dataset(path, split=spt)
    dataset_list.append(dataset)
```

## Expected results
```
###################### ['train', 'test', 'validation'] ######################
###################### ['train', 'test', 'validation'] ######################
```

## Actual results
```
Traceback (most recent call last):
  File "test_data_loader.py", line 11, in <module>

    dataset = load_dataset(path, split=spt)
  File "/home/venv/lib/python3.8/site-packages/datasets/load.py", line 1758, in load_dataset
    ds = builder_instance.as_dataset(split=split, ignore_verifications=ignore_verifications, in_memory=keep_in_memory)
[...]

PR #4985, opened 2022-09-16
Prefer split patterns from directories over split patterns from filenames

polinaeterna:
related to https://github.com/huggingface/datasets/issues/4895

todo:

- [ ] test


src/datasets/data_files.py
```diff
@@ -73,10 +73,10 @@ class Url(str):
     Split.TRAIN: ["**"],
 }
 
-ALL_SPLIT_PATTERNS = [SPLIT_PATTERN_SHARDED]
+ALL_SPLIT_PATTERNS = [SPLIT_PATTERN_SHARDED]  # TODO: should this be changed?
```
lhoestq, line 76: I think it's fine this way. This corresponds to the patterns that contain custom split names (other than train/dev/test)

lhoestq (commented):
Thanks ! adding a test just for this is maybe overkill but as you want

polinaeterna:
@mariosasko sure! could you please approve it?

shaneacton:
Hi there! I have installed 5.2.3.dev0, which should have this fix. Unfortunately, I am still getting the error:
`ValueError: Unknown split "validation". Should be one of ['train'].` When I call `load_dataset("csv", data_files=files, split=split)`

Any help would be greatly appreciated!

PR #4986, opened 2022-09-16
[doc] Fix broken snippet that had too many quotes

tomaarsen:
Hello!

### Pull request overview
* Fix broken snippet in https://huggingface.co/docs/datasets/main/en/process that has too many quotes

### Details
The snippet in question can be found here: https://huggingface.co/docs/datasets/main/en/process#map
This screenshot shows the issue, there is a quote too many, causing the snippet to be colored incorrectly:
image

The change speaks for itself.

Thank you for the detailed documentation, by the way. 

- Tom Aarsen

### DIFF
--- a/docs/source/process.mdx
+++ b/docs/source/process.mdx
@@ -256,7 +256,7 @@ Start by creating a function that adds `'My sentence: '` to the beginning of eac
 
 ```py
 >>> def add_prefix(example):
-...     example["sentence1"] = 'My sentence: '' + example["sentence1"]
+...     example["sentence1"] = 'My sentence: ' + example["sentence1"]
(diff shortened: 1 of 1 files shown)


albertvillanova (approved):
Thank you for the docs fix. :)

tomaarsen:
Spent the day familiarising myself with the huggingface line of products, and happened to run into some small issues here and there. Magically, I've found exactly one small issue in `transformers`, one in `accelerate` and now one in `datasets`, hah!

As for this PR, the issue seems solved according to the [new PR documentation](https://moon-ci-docs.huggingface.co/docs/datasets/pr_4986/en/process#map):
image

PR #4987, opened 2022-09-16
Embed image/audio data in dl_and_prepare parquet

lhoestq:
Embed the bytes of the image or audio files in the Parquet files directly, instead of having a "path" that points to a local file.

Indeed Parquet files are often used to share data or to be used by workers that may not have access to the local files.


mariosasko (approved):
Thanks! LGTM!

PR #5006, opened 2022-09-21
Revert input_columns change

lhoestq:
Revert https://github.com/huggingface/datasets/pull/4971

Fix https://github.com/huggingface/datasets/issues/5005

PR #5008, opened 2022-09-21
Re-apply input columns change

mariosasko:
Fixes the `filter` + `input_columns` combination, which is used in the `transformers` examples for instance.

Revert #5006 (which in turn reverts #4971)

Fix https://github.com/huggingface/datasets/issues/4858

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -2733,7 +2733,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                 validate_function_output(processed_inputs, indices)
             if not update_data:
                 return None  # Nothing to update, let's move on
-            if self._format_type:
+            if self._format_type or input_columns:
                 inputs = self._getitem(
                     key=(indices if isinstance(indices, int) else slice(indices[0], indices[-1] + 1)),
                     format_type=None,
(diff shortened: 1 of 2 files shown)


tests/test_arrow_dataset.py
```diff
@@ -1474,6 +1476,15 @@ def test_filter_batched(self, in_memory):
...
 
+    def test_filter_input_columns(self, in_memory):
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            dset = Dataset.from_dict({"col_1": [0, 1, 2], "col_2": ["a", "b", "c"]})
+            with self._to(in_memory, tmp_dir, dset) as dset:
+                with dset.filter(lambda x: x > 0, input_columns=["col_1"]) as filtered_dset:
+                    self.assertListEqual(filtered_dset.column_names, dset.column_names)
+                    self.assertListEqual(filtered_dset["col_1"], [1, 2])
+                    self.assertListEqual(filtered_dset["col_2"], ["b", "c"])
+
```
albertvillanova, line 1487: Thanks for the regression test.

src/datasets/arrow_dataset.py
```diff
@@ -2800,8 +2800,6 @@ def init_buffer_and_writer():
                     input_dataset = self.with_format(
                         self._format_type, columns=input_columns, output_all_columns=False, **self._format_kwargs
                     )
-                    if remove_columns:
-                        remove_columns = list(set(remove_columns) & set(input_columns))
```
albertvillanova, line 2804: I guess this was the source of the issue with filter...

I'm wondering why this was introduced in:
- #2246

albertvillanova (approved):
Thanks. Now the transformers CI should not be broken.

lhoestq (approved):
Thanks ! LGTM

src/datasets/arrow_dataset.py
mariosasko: I guess the contributor misinterpreted the purpose of `input_columns` and `remove_columns` in that PR.

PR #5020, opened 2022-09-24
Fix URLs of sbu_captions dataset

donglixp:
Forbidden
You don't have permission to access /~vicente/sbucaptions/sbu-captions-all.tar.gz on this server.

Additionally, a 403 Forbidden error was encountered while trying to use an ErrorDocument to handle the request.

Apache/2.4.6 (Red Hat Enterprise Linux) OpenSSL/1.0.2k-fips PHP/5.4.16 mod_fcgid/2.3.9 mod_wsgi/3.4 Python/2.7.5 mod_perl/2.0.11 Perl/v5.16.3 Server at [www.cs.virginia.edu](mailto:csroot@virginia.edu) Port 443


albertvillanova (changes requested):
Thanks for the fix @donglixp.

The homepage URL should also be updated (in the script and in the README.md file) and the "test" CLI command should be runned to update the metadata JSON file:
```shell
datasets-cli test datasets/sbu_captions --save_infos --all_configs
```

Please feel free to ask if you need help with this or if you would prefer us to run it.

albertvillanova (approved):
Thanks for the fix.

PR #5029, opened 2022-09-27
Fix import in `ClassLabel` docstring example

alvarobartt:
This PR addresses a super-simple fix: adding a missing `import` to the `ClassLabel` docstring example, as it was formatted as `from datasets Features`, so it's been fixed to `from datasets import Features`.


albertvillanova (approved):
Thanks, @alvarobartt!! :)

PR #5030, opened 2022-09-27
Fast dataset iter

mariosasko:
Use `pa.Table.to_reader` to make iteration over examples/batches faster.

TODO:
* benchmarking (the only benchmark for now - iterating over (single) examples of `bookcorpus` (75 mil examples) in Colab is approx. 2.3x faster)
* check if iterating over bigger chunks + slicing to fetch individual examples in `_iter` yields better performance

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -1854,17 +1854,61 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         """
         return self.num_rows
 
+    def _iter_batches(self, batch_size, decoded: bool = True):
+        """Iterate through the batches of size `batch_size`.
+
+        If a formatting is set with :meth:`Dataset.set_format` rows will be returned with the
+        selected format.
+        """
+        if self._indices is None and config.PYARROW_VERSION.major >= 8:
+            # Fast iteration
+            format_kwargs = self._format_kwargs if self._format_kwargs is not None else {}
+            formatter = get_formatter(self._format_type, features=self.features, decoded=decoded, **format_kwargs)
+            schema = self.data._schema
+            for i, batch in enumerate(self.data.to_reader(max_chunksize=batch_size)):
+                pa_subtable = pa.Table.from_batches([batch], schema=schema)
+                formatted_output = format_table(
+                    pa_subtable,
+                    slice(i, i + batch_size),
+                    formatter=formatter,
+                    format_columns=self._format_columns,
+                    output_all_columns=self._output_all_columns,
+                )
+                yield formatted_output
+        else:
+            for i in range(0, self.num_rows, batch_size):
+                yield self._getitem(
+                    slice(i, i + batch_size),
+                    decoded=decoded,
+                )
+
     def _iter(self, decoded: bool = True):
         """Iterate through the examples.
 
         If a formatting is set with :meth:`Dataset.set_format` rows will be returned with the
         selected format.
         """
-        for index in range(self.num_rows):
-            yield self._getitem(
-                index,
-                decoded=decoded,
-            )
+        if self._indices is None and config.PYARROW_VERSION.major >= 8:
+            # Fast iteration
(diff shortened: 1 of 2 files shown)


mariosasko:
I ran some benchmarks (focused on the data fetching part of `__iter__`) and it seems like the combination `table.to_reader(batch_size)` + `RecordBatch.slice` performs the best ([script](https://gist.github.com/mariosasko/0248288a2e3a7556873969717c1fe52b) with the results). I think we can choose (implicit) `batch_size=10` in the final implementation to avoid having problems with fetching large examples.

src/datasets/arrow_dataset.py
```diff
@@ -1854,17 +1854,62 @@ def __len__(self):
         """
         return self.num_rows
 
+    def _iter_batches(self, batch_size, decoded: bool = True):
```
lhoestq, line 1857:
```suggestion
    def _iter_batches(self, batch_size: int, decoded: bool = True):
```

src/datasets/arrow_dataset.py
```diff
@@ -2805,14 +2850,16 @@ def init_buffer_and_writer():
...
+                    pbar_iterable = input_dataset._iter(decoded=False)
                 else:
                     num_rows = (
                         len(input_dataset) if not drop_last_batch else len(input_dataset) // batch_size * batch_size
                     )
-                    pbar_iterable = range(0, num_rows, batch_size)
                     pbar_total = (num_rows // batch_size) + 1 if num_rows % batch_size else num_rows // batch_size
+                    pbar_iterable = itertools.islice(
+                        input_dataset._iter_batches(batch_size, decoded=False), pbar_total
+                    )
```
lhoestq, line 2862: Why do you need islice here ?

src/datasets/config.py
```diff
@@ -168,6 +168,9 @@
 # https://github.com/apache/arrow/blob/master/docs/source/cpp/arrays.rst#size-limitations-and-recommendations)
 DEFAULT_MAX_BATCH_SIZE = 10_000
 
+# Number of examples prefetched in `Dataset.__iter__`
+DEFAULT_ITER_BATCH_SIZE = 10
```
lhoestq, line 172: Prefetching usually means bring to memory, which is not 100% the case here no ? We just prefetch the Arrow data, not the python objects. Since this is very specific to the optimized function I think we can use a more specific name, maybe `ARROW_READER_BATCH_SIZE_IN_DATASET_ITER`

src/datasets/arrow_dataset.py
```diff
@@ -1854,17 +1854,62 @@ def __len__(self):
...
         return self.num_rows
 
+    def _iter_batches(self, batch_size, decoded: bool = True):
+        """Iterate through the batches of size `batch_size`.
+
+        If a formatting is set with :meth:`Dataset.set_format` rows will be returned with the
+        selected format.
+        """
+        if self._indices is None and config.PYARROW_VERSION.major >= 8:
+            # Fast iteration
```
lhoestq, line 1864: Maybe link to the benchmarks you did ?

src/datasets/arrow_dataset.py
```diff
@@ -1854,17 +1854,62 @@ def __len__(self):
...
         If a formatting is set with :meth:`Dataset.set_format` rows will be returned with the
         selected format.
         """
-        for index in range(self.num_rows):
-            yield self._getitem(
-                index,
-                decoded=decoded,
-            )
+        if self._indices is None and config.PYARROW_VERSION.major >= 8:
+            # Fast iteration
```
lhoestq, line 1891: same here

lhoestq (commented):
Nice thank you ! The benchmarks are super helpful as well, I think you can link to them in comments in the code

src/datasets/arrow_dataset.py
mariosasko: To account for the `drop_last_batch=True` case.

lhoestq (approved):

PR #5048, opened 2022-09-30
Fix bug with labels of eurlex config of lex_glue dataset

iliaschalkidis:
Fix for a critical bug in the EURLEX dataset label list to make LexGLUE EURLEX results replicable. 

In LexGLUE (Chalkidis et al., 2022), the following is mentioned w.r.t. EUR-LEX: _"It supports four different label granularities, comprising 21, 127, 567, 7390 EuroVoc concepts, respectively. We use the 100 most frequent concepts from level 2 [...]”._ The current label list has all 127 labels, which leads to different (lower) results, as communicated by users.

Thanks!


iliaschalkidis:
@JamesLYC88 here is the fix! Thanks again!

albertvillanova (approved):
Thanks for the fix, @iliaschalkidis.

iliaschalkidis:
Thanks, @albertvillanova. When do you expect that this change will take effect when someone downloads the dataset?

albertvillanova:
The change is immediately available now.

PR #5049, opened 2022-09-30
Add `kwargs` to `Dataset.from_generator`

mariosasko:
Add the `kwargs` param to `from_generator` to align it with the rest of the `from_` methods (this param allows passing custom `writer_batch_size` for instance).


lhoestq (approved):
Cool thanks !

PR #5051, opened 2022-09-30
Revert task removal in folder-based builders

mariosasko:
Reverts the removal of `task_templates` in the folder-based builders. I also added the `AudioClassifaction` task for consistency.

This is needed to fix https://github.com/huggingface/transformers/issues/19177.

I think we should soon deprecate and remove the current task API (and investigate if it's possible to integrate the `train eval index` API), but we need to update the Transformers examples before that so we don't break them.

cc @NielsRogge


lhoestq (approved):
Cool thanks :)

I agree it would be nice to use the train_eval_index metadata in the future

Issue #4988, opened 2022-09-16
Add `IterableDataset.from_generator` to the API


mariosasko:
We've just added `Dataset.from_generator` to the API. It would also be cool to add `IterableDataset.from_generator` to support creating an iterable dataset from a generator.

cc @lhoestq

hamid-vakilzadeh:
#take

PR #5052, opened 2022-09-30
added from_generator method to IterableDataset class.

hamid-vakilzadeh:
Hello, 

This resolves issues #4988.

I added a method `from_generator` to class `IterableDataset`.

I modified the `read` method of input stream generator to also return Iterable_dataset.

### DIFF
--- a/src/datasets/io/abc.py
+++ b/src/datasets/io/abc.py
@@ -1,10 +1,7 @@
 from abc import ABC, abstractmethod
 from typing import Optional, Union
 
-from datasets.iterable_dataset import IterableDataset
-
-from .. import DatasetDict, Features, NamedSplit
-from ..arrow_dataset import Dataset
+from .. import Dataset, DatasetDict, Features, IterableDataset, IterableDatasetDict, NamedSplit
 from ..utils.typing import NestedDataStructureLike, PathLike
 
 
@@ -28,7 +25,7 @@ class AbstractDatasetReader(ABC):
         self.kwargs = kwargs
 
     @abstractmethod
-    def read(self) -> Union[Dataset, DatasetDict, IterableDataset, IterableDataset]:
+    def read(self) -> Union[Dataset, DatasetDict, IterableDataset, IterableDatasetDict]:
         pass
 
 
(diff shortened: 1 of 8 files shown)


src/datasets/io/generator.py
```diff
@@ -24,7 +24,7 @@ def __init__(
             **kwargs,
         )
 
-    def read(self):
+    def read(self, streaming: bool = False):
```
mariosasko, line 27: This is a new param, so let's move to the end.

mariosasko (commented):
Great job, thanks! 

Besides my comment below, we also need a test (let me know if you need help with that), and then we can merge.

src/datasets/io/generator.py
hamid-vakilzadeh: Thank you @mariosasko for your feedback.
I am not sure I understand what I need to do here. Can you please help me on this.
Also a little help on generating the test module would be appreciated.
Thank you!

mariosasko: Oops, I didn't notice that this is the only param in the signature 🙂.

mariosasko:
I added a test and moved the `streaming` param from `read` to `__init_`. Then, I also decided to update the `read` method of the rest of the packaged modules to account for this param. 

@hamid-vakilzadeh Are you OK with these changes?

src/datasets/io/abc.py
```diff
@@ -14,17 +16,19 @@ def __init__(
...
         self.split = split if split or isinstance(path_or_paths, dict) else "train"
         self.features = features
         self.cache_dir = cache_dir
         self.keep_in_memory = keep_in_memory
+        self.streaming = streaming
         self.kwargs = kwargs
 
     @abstractmethod
-    def read(self) -> Union[Dataset, DatasetDict]:
+    def read(self) -> Union[Dataset, DatasetDict, IterableDataset, IterableDataset]:
```
lhoestq, line 31: IterableDatasetDict maybe ?

lhoestq (approved):
Love it ! Thanks :)

PR #5056, opened 2022-10-03
Fix broken URL's (GEM)

manandey:
This PR fixes the broken URL's in GEM. cc. @lhoestq, @albertvillanova


albertvillanova:
Thanks, @manandey. We have removed all dataset scripts from this repo. Subsequent PRs should be opened directly on the Hugging Face Hub.

PR #5063, opened 2022-10-04
Align signature of list_repo_files with latest hfh

albertvillanova:
This PR aligns the signature of `list_repo_files` with the current one in `hfh`, by renaming deprecated `token` to  `use_auth_token`.

This is already the case for `dataset_info`.

CC: @lhoestq

### DIFF
--- a/src/datasets/utils/_hf_hub_fixes.py
+++ b/src/datasets/utils/_hf_hub_fixes.py
@@ -161,7 +161,7 @@ def list_repo_files(
     repo_id: str,
     revision: Optional[str] = None,
     repo_type: Optional[str] = None,
-    use_auth_token: Optional[str] = None,
+    use_auth_token: Optional[Union[bool, str]] = None,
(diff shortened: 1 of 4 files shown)


src/datasets/utils/_hf_hub_fixes.py
```diff
@@ -161,16 +161,18 @@ def list_repo_files(
     repo_id: str,
     revision: Optional[str] = None,
     repo_type: Optional[str] = None,
-    token: Optional[str] = None,
+    use_auth_token: Optional[str] = None,
```
lhoestq, line 164: "token" has type `Optional[str]` while "use_auth_token" has "Optional[Unnion[bool, str]]", so this is a bit inconsistent. Could you add the support for bool as well here ?

albertvillanova: You are right: I forgot to update the type hint.

PR #5064, opened 2022-10-04
Align signature of create/delete_repo with latest hfh

albertvillanova:
This PR aligns the signature of `create_repo`/`delete_repo` with the current one in hfh, by removing deprecated `name` and `organization`, and using `repo_id` instead.

Related to:
- #5063

CC: @lhoestq

### DIFF
--- a/src/datasets/utils/_hf_hub_fixes.py
+++ b/src/datasets/utils/_hf_hub_fixes.py
@@ -39,7 +39,7 @@ def create_repo(
         `str`: URL to the newly created repo.
     """
     if version.parse(huggingface_hub.__version__) < version.parse("0.5.0"):
-        name, organization = repo_id.split("/")
+        organization, name = repo_id.split("/")
         return hf_api.create_repo(
             name=name,
             organization=organization,
@@ -83,7 +83,7 @@ def delete_repo(
         `str`: URL to the newly created repo.
     """
     if version.parse(huggingface_hub.__version__) < version.parse("0.5.0"):
-        name, organization = repo_id.split("/")
+        organization, name = repo_id.split("/")
         return hf_api.delete_repo(
             name=name,
             organization=organization,
--- a/src/datasets/utils/_hf_hub_fixes.py
+++ b/src/datasets/utils/_hf_hub_fixes.py
@@ -8,9 +8,8 @@ from packaging import version
 
 def create_repo(
     hf_api: HfApi,
-    name: str,
+    repo_id: str,
     token: Optional[str] = None,
-    organization: Optional[str] = None,
     private: Optional[bool] = None,
     repo_type: Optional[str] = None,
     exist_ok: Optional[bool] = False,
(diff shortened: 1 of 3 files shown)


src/datasets/utils/_hf_hub_fixes.py
```diff
@@ -42,9 +39,10 @@ def create_repo(
         `str`: URL to the newly created repo.
     """
     if version.parse(huggingface_hub.__version__) < version.parse("0.5.0"):
+        name, *organization = repo_id.split("/")
```
lhoestq, line 42: if you use `*organization`, then `organization` is a list, not a string
```suggestion
        organization, name = repo_id.split("/") if "/" in repo_id else None, repo_id
```

src/datasets/utils/_hf_hub_fixes.py
```diff
@@ -42,9 +39,10 @@ def create_repo(
...
         return hf_api.create_repo(
             name=name,
-            organization=organization,
+            organization=organization or None,
```
lhoestq, line 45:
```suggestion
            organization=organization,
```

src/datasets/utils/_hf_hub_fixes.py
```diff
@@ -88,15 +83,16 @@ def delete_repo(
         `str`: URL to the newly created repo.
     """
     if version.parse(huggingface_hub.__version__) < version.parse("0.5.0"):
+        name, *organization = repo_id.split("/")
```
lhoestq, line 86: same here
```suggestion
        organization, name = repo_id.split("/") if "/" in repo_id else None, repo_id
```

lhoestq (commented):
small bug:

src/datasets/utils/_hf_hub_fixes.py
albertvillanova: That is why I put below:
```python
organization=organization or None
```
because `[] or None` is `None`.

albertvillanova: Anyway, this method is internal and we only call it for Hub datasets (see the rest of the modified files): so organization is never None.

albertvillanova: Oh, I see, it is NOT a str but a list... Oops!!

albertvillanova: Maybe I can just ignore the case where organization is None?
```python
name, organization = repo_id.split("/")
```

albertvillanova: I mean we do not create/delete canonical repos with these methods...

albertvillanova: The only publicly available one is in `_push_parquet_shards_to_hub`, and there if there is no "/", we replace `repo_id` by preceding it with the current username as organization:
```python
        elif len(identifier) == 1:
            dataset_name = identifier[0]
            organization_or_username = api.whoami(token)["name"]
            repo_id = f"{organization_or_username}/{dataset_name}"

PR #5065, opened 2022-10-04
Ci py3.10

lhoestq:
Added a CI job for python 3.10

Some dependencies don't work on 3.10 like apache beam, so I remove them from the extras in this case.
I also removed some s3 fixtures that we don't use anymore (and that don't work on 3.10 anyway)


lhoestq:
Does it sound good to you @albertvillanova ?

tests/test_metric_common.py
```diff
@@ -152,6 +152,7 @@ def wrapper(patcher):
 # --------------------------------
 
 
+@pytest.mark.integration
```
albertvillanova, line 155: If all the tests in this module are going to be marked as "integration", you could do it only once at the module level:
```python
pytestmark = pytest.mark.integration
```
and remove all integration mark decorators.

tests/test_filesystem.py
```diff
@@ -17,26 +15,9 @@
...
-    os.environ["AWS_SECURITY_TOKEN"] = "fake_secrurity_token"
-    os.environ["AWS_SESSION_TOKEN"] = "fake_session_token"
-
-
-@pytest.fixture(scope="function")
-def s3(aws_credentials):
-    with mock_s3():
-        yield boto3.client("s3", region_name="us-east-1")
-
-
```
albertvillanova, line 34: Great!

tests/test_beam.py
```diff
@@ -26,6 +22,8 @@ def _split_generators(self, dl_manager, pipeline):
         return [datasets.SplitGenerator(name=datasets.Split.TRAIN, gen_kwargs={"examples": get_test_dummy_examples()})]
 
     def _build_pcollection(self, pipeline, examples):
+        import apache_beam as beam
```
albertvillanova, line 25: Great as well!

setup.py
```diff
@@ -189,6 +208,7 @@
     "streaming": [],  # for backward compatibility
     "dev": TESTS_REQUIRE + QUALITY_REQUIRE,
     "tests": TESTS_REQUIRE,
+    "metrics_tests": METRICS_TESTS_REQUIRE,
```
albertvillanova, line 211: NIT: maybe using hyphen instead of underscore, as we already have "apache-beam".

setup.py
```diff
@@ -139,6 +140,24 @@
...
+    "boto3",
+    "botocore",
+    "moto",
+    "s3fs"
+]
+
+
+for i, dependency in enumerate(TESTS_REQUIRE):
+    if re.match(r"[\w-]+", dependency).group() in TEST_PY37_ONLY:
+        TESTS_REQUIRE[i] = dependency + ";python_version<'3.8'"
```
albertvillanova, line 157: Why don't adding ";python_version<'3.8'" directly to these elements in `TESTS_REQUIRE` and avoiding a regex here?

albertvillanova (approved):
Nice! Thanks for adding Python 3.10 to our CI.

setup.py
lhoestq: just removed it :)

PR #5071, opened 2022-10-05
Support DEFAULT_CONFIG_NAME when no BUILDER_CONFIGS

### LINKED ISSUE #5070: Support default config name when no builder configs
As discussed with @stas00, we could support defining a default config name, even if no predefined allowed config names are set. That is, support `DEFAULT_CONFIG_NAME`, even when `BUILDER_CONFIGS` is not defined. 

**Additional context**
In order to support creating configs on the fly **by name** (not using kwargs), the list of allowed builder configs `BUILDER_CONFIGS` must not be set.

However, if so, then `DEFAULT_CONFIG_NAME` is not supported.

albertvillanova:
This PR supports defining a default config name, even if no predefined allowed config names are set. 

Fix #5070.

CC: @stas00


lhoestq (approved):
sounds good !

PR #5072, opened 2022-10-05
Image & Audio formatting for numpy/torch/tf/jax

lhoestq:
Added support for image and audio formatting for numpy, torch, tf and jax.

For images, the dtype used is the one of the image (the one returned by PIL.Image), e.g. uint8

I also added support for string, binary and None types. In particular for torch and jax, strings are kept unchanged (previously it was returning an error because you can't create a tensor of strings)


lhoestq:
I just added a consolidation step so that numpy arrays or tensors of images are stacked together if the shapes match, instead of having lists of tensors

docs/source/use_with_pytorch.mdx
```diff
@@ -86,20 +86,36 @@ To get a single tensor, you must explicitly use the [`Array`] feature type and s
 {'data': tensor([0, 0, 1])}
 ```
 
-However, since it's not possible to convert text data to PyTorch tensors, you can't format a `string` column to PyTorch.
-Instead, you can explicitly format certain columns and leave the other columns unformatted:
+String and binary objects are unchanged, since PyTorch only supports numbers.
+
+The [`Image`] and [`Audio`] feature types are also supported:
```
mariosasko, line 91: I would also add an example with a slice to show that batches (of images) can also be formatted.

src/datasets/formatting/np_formatter.py
```diff
@@ -0,0 +1,78 @@
...
+from ..utils.py_utils import map_nested
+from .formatting import Formatter
+
+
+class NumpyFormatter(Formatter[dict, np.ndarray, dict]):
+    def __init__(self, features=None, decoded=True, **np_array_kwargs):
+        super().__init__(features=features, decoded=decoded)
+        self.np_array_kwargs = np_array_kwargs
+
+    def _consolidate(self, column):
```
mariosasko, line 16: We can use `np.object` for ragged tensors, no?

src/datasets/formatting/np_formatter.py
```diff
@@ -0,0 +1,78 @@
...
+            if isinstance(value, PIL.Image.Image):
+                return np.asarray(value, **self.np_array_kwargs)
+
+        return np.array(value, **{**default_dtype, **self.np_array_kwargs})
+
+    def _recursive_tensorize(self, data_struct: dict):
+        # support for nested types like struct of list of struct
+        if isinstance(data_struct, np.ndarray):
+            if data_struct.dtype == object:  # torch tensors cannot be instantied from an array of objects
+                return self._consolidate([self.recursive_tensorize(substruct) for substruct in data_struct])
```
mariosasko, line 51: I don't think we need this for the np formatter :).

mariosasko (commented):
Cool! A few nits, but other than that looks great.

src/datasets/formatting/np_formatter.py
lhoestq: Yes indeed, though I don't know if many users are familiar with np.object numpy arrays, that's shy I switched to using lists.

Not a strong opinion though

src/datasets/formatting/np_formatter.py
lhoestq: I think we do, since arrow outputs np.object numpy arrays for nested lists

src/datasets/formatting/np_formatter.py
mariosasko: This is a breaking change, so this is why I think it's better to return `np.object`.

E.g. `Dataset.from_dict({"a": [[1, 2], [3]]}).with_format("np")[:]` returns a numpy array now, but it would return a list with this change.

lhoestq: I see, I'll revert this

lhoestq:
I added a few lines in the docs and reverted the ragged numpy array change :)

ready for another review !

src/datasets/formatting/np_formatter.py
```diff
@@ -0,0 +1,80 @@
...
+        self.np_array_kwargs = np_array_kwargs
+
+    def _consolidate(self, column):
+        if isinstance(column, list):
+            if column and all(
+                isinstance(x, np.ndarray) and x.shape == column[0].shape and x.dtype == column[0].dtype for x in column
+            ):
+                return np.stack(column)
+            else:
+                return np.array(column, dtype=object)
```
mariosasko, line 23: Casting to `np.object` can sometimes fail (see https://stackoverflow.com/q/51005699), so I think we need to wrap this with try-except and return a list if that happens

mariosasko (commented):
Last nit.

src/datasets/formatting/np_formatter.py
lhoestq: I used
```python
out = np.empty(len(column), dtype=object)
out[:] = column
```
which is suggested in the SO answer, and it works well :)

let me know if it's all good for you

mariosasko (approved):
Thanks, looks all good now!

Issue #5050, opened 2022-09-30
Restore saved format state in `load_from_disk`


mariosasko:
Even though we save the `format` state in `save_to_disk`, we don't restore it in `load_from_disk`. We should fix that.
 
Reported here: https://discuss.huggingface.co/t/save-to-disk-loses-formatting-information/23815

asofiaoliveira:
Hi, can I work on this?

mariosasko:
Hi, sure! Let us know if you need some pointers/help.

PR #5073, opened 2022-10-05
Restore saved format state in `load_from_disk`

asofiaoliveira:
Hello! @mariosasko 

This pull request relates to issue #5050 and intends to add the format to datasets loaded from disk. 
All I did was add a set_format in the Dataset.load_from_disk, as DatasetDict.load_from_disk relies on the first.

I don't know if I should add a test and where, so let me know if I should and I can work on that as well!

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -1346,12 +1346,17 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         split = state["_split"]
         split = Split(split) if split is not None else split
 
-        return Dataset(
+        dataset = Dataset(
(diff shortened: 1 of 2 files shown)


mariosasko (commented):
Thanks! 

Yes, we need to test this. You can add a test as a method of `BaseDatasetTest` in `tests/test_arrow_dataset.py`. In that test, you can save a formatted dataset with `save_to_disk`, then reload it with `load_from_disk` and then check if the format attributes of the original and the loaded dataset are the same.

mariosasko (approved):
I replaced `set_format` with `with_format` because I slightly prefer the latter. Looks all good now, thanks!

lhoestq (approved):
Thanks for the fix and for the test ! :)

Issue #514, opened 2020-08-18
dataset.shuffle(keep_in_memory=True) is never allowed


vegarab:
As of commit ef4aac2, the usage of the parameter `keep_in_memory=True` is never possible: `dataset.select(keep_in_memory=True)`

The commit added the lines
```python
# lines 994-996 in src/nlp/arrow_dataset.py
       assert (
            not keep_in_memory or cache_file_name is None
        ), "Please use either `keep_in_memory` or `cache_file_name` but not both."
```

This affects both `shuffle()` as `select()` is a sub-routine, and `map()` that has the same check. 

I'd love to fix this myself, but unsure what the intention of the assert is given the rest of the logic in the function concerning `ccache_file_name` and `keep_in_memory`.

vegarab:
I'm thinking the intended implementation here is
```python
assert (
            keep_in_memory and  is None
        ), "Please use either `keep_in_memory` or `cache_file_name` but not both."
```

vegarab:
This seems to be fixed in #513 for the filter function, replacing `cache_file_name` with `indices_cache_file_name ` in the assert. Although not for the `map()` function @thomwolf

thomwolf:
Maybe I'm a bit tired but I fail to see the issue here.

Since `cache_file_name` is `None` by default, if you set `keep_in_memory` to `True`, the assert should pass, no?

vegarab:
I failed to realise that this only applies to `shuffle()`. Whenever `keep_in_memory` is set to True, this is passed on to the `select()` function. However, if `cache_file_name` is None, it will be defined in the `shuffle()` function before it is passed on to `select()`. 

Thus, `split()` is called with `keep_in_memory=True` and a not None value for `cache_file_name`. 
This is essentially fixed in #513

thomwolf:
Oh yes ok got it thanks. Should be fixed if we are happy with #513 indeed.

vegarab:
My bad. This is actually not fixed in #513. Sorry about that...
The new `indices_cache_file_name` is set to a non-None value in the new `shuffle()` as well. 

The buffer and caching mechanisms used in the `select()` function are too intricate for me to understand why the check is there at all. I've removed it in my local build and it seems to be working fine for my project, without really considering other implications of the change.

thomwolf:
Ok I'll investigate and add a series of tests on the `keep_in_memory=True` settings which is under-tested atm

epwalsh:
[...]

PR #5082, opened 2022-10-06
adding keep in memory

Mustapha-AJEGHRIR:
Fixing #514 .

Hello @mariosasko 👋, I have implemented what you have recommanded to fix the keep in memory problem for shuffle on the issue #514 .

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -3588,7 +3588,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
             generator = np.random.default_rng(seed)
 
         # Check if we've already cached this computation (indexed by a hash)
-        if self.cache_files and not keep_in_memory:
+        if self.cache_files:
             if indices_cache_file_name is None:
                 # we create a unique hash from the function, current dataset file and the mapping args
                 indices_cache_file_name = self._get_cache_file_path(new_fingerprint)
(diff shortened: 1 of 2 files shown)


src/datasets/arrow_dataset.py
```diff
@@ -3585,7 +3588,7 @@ def shuffle(
             generator = np.random.default_rng(seed)
 
         # Check if we've already cached this computation (indexed by a hash)
-        if self.cache_files:
+        if self.cache_files and not keep_in_memory:
```
mariosasko, line 3591: This line should stay the same:

```suggestion
        if self.cache_files:
```

Instead, [this](https://github.com/huggingface/datasets/blob/7d51531c58127d44565d6d1a6a7aa75e55d2b036/src/datasets/arrow_dataset.py#L3603) line should be replaced with:
```python
indices_cache_file_name=indices_cache_file_name if not keep_in_memory else None
```

mariosasko (commented):
Thanks!

We also need a test with `shuffle(keep_in_memory=True)` for this fix. Let me know if you need help with that.

Mustapha-AJEGHRIR:
Hi @mariosasko , I have added a test for the `keep_in_memory` version. I have also removed the `Compatible with temp_seed` part in the scope of `dset_shuffled`, please verify if that makes sense.

mariosasko (approved):
Looks all good now!

lhoestq (approved):
Nice !

Issue #5083, opened 2022-10-06
Support numpy/torch/tf/jax formatting for IterableDataset


lhoestq:
Right now `IterableDataset` doesn't do any formatting.

In particular this code should return a numpy array:
```python
from datasets import load_dataset

ds = load_dataset("imagenet-1k", split="train", streaming=True).with_format("np")
print(next(iter(ds))["image"])
```

Right now it returns a PIL.Image.

Setting `streaming=False` does return a numpy array after #5072

zutarich:
hii @lhoestq, can you assign this issue to me? Though i am new to open source still I would love to put my best foot forward. I can see there isn't anyone right now assigned to this issue.

lhoestq:
Hi @zutarich ! This issue was fixed by #5852 - sorry I forgot to close it

Feel free to look for other issues and ping me or @mariosasko if you have questions :)
Also let us know if we can help find an issue that can correspond to what you're looking for

PR #5084, opened 2022-10-06
IterableDataset formatting in numpy/torch/tf/jax

lhoestq:
This code now returns a numpy array:
```python

ds = load_dataset("imagenet-1k", split="train", streaming=True).with_format("np")
print(next(iter(ds))["image"])
```

It also works with "arrow", "pandas", "torch", "tf" and "jax"

### Implementation details:

I'm using the existing code to format an Arrow Table to the right output format for simplicity.
Therefore it's probbaly not the most optimized approach.

For example to output PyTorch tensors it does this for every example:

python data -> arrow table -> numpy extracted data -> pytorch formatted data

### Releasing this feature

Even though I consider this as a bug/inconsistency, this change is a breaking change.
And I'm sure some users were relying on the torch iterable dataset to return PIL Image and used data collators to convert to pytorch.

So I guess this is `datasets` 3.0 ?

### TODO

- [x] merge https://github.com/huggingface/datasets/pull/5072
- [ ] docs
- [ ] tests

Close https://github.com/huggingface/datasets/issues/5083


lhoestq:
Actually I'm not happy with this implementation. It always require the iterable dataset to have definite `features`, which removes a lot of flexibility. So I think we need an actual formatting from python objects, not from arrow data.

lhoestq:
Closing this one since it has too many conflicts and still require some work - it will be easier to open a new PR

PR #5092, opened 2022-10-10
Use HTML relative paths for tiles in the docs

lewtun:
This PR replaces the absolute paths in the landing page tiles with relative ones so that one can test navigation both locally in and in future PRs (see [here](https://moon-ci-docs.huggingface.co/docs/datasets/pr_5084/en/index) for an example PR where the links don't work).

I encountered this while working on the `optimum` docs and figured I'd fix it elsewhere too :)

Internal Slack thread: https://huggingface.slack.com/archives/C02GLJ5S0E9/p1665129710176619

### DIFF
--- a/docs/source/index.mdx
+++ b/docs/source/index.mdx
@@ -10,19 +10,19 @@ Find your dataset today on the [Hugging Face Hub](https://huggingface.co/dataset
 
 <div class="mt-10">
   <div class="w-full flex flex-col space-y-4 md:space-y-0 md:grid md:grid-cols-2 md:gap-y-4 md:gap-x-5">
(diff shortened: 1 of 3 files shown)


albertvillanova (approved):
Good catch, @lewtun. Thanks for the fix.

Do you know if there are other absolute paths in the docs that should be fixed as well?

lewtun:
> Good catch, @lewtun. Thanks for the fix.
> 
> Do you know if there are other absolute paths in the docs that should be fixed as well?

I found a few more in [0d4796b](https://github.com/huggingface/datasets/pull/5092/commits/0d4796b747e6620d9fcc17a8f74acc5cf4bba7be).

However, I noticed that none of the cross-references (e.g. to API classes / methods) work locally, but that is probably just a limitation of the local build

albertvillanova:
Thanks.

PR #5101, opened 2022-10-11
Free the "hf" filesystem protocol for `hffs`

lhoestq:


albertvillanova (approved):
Thanks, yes it is necessary to free that protocol name...

PR #5106, opened 2022-10-12
Fix task template reload from dict

lhoestq:
Since #4926 the JSON dumps are simplified and it made task template dicts empty by default.

I fixed this by always including the task name which is needed to reload a task from a dict

### DIFF
--- a/src/datasets/tasks/audio_classificiation.py
+++ b/src/datasets/tasks/audio_classificiation.py
@@ -1,5 +1,5 @@
 import copy
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import ClassVar, Dict
 
 from ..features import Audio, ClassLabel, Features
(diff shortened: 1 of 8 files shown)


albertvillanova (approved):
Thanks for fixing the bug introduced in: 
- #4926
https://github.com/huggingface/datasets/blob/60bbac563dfeb055ea3643f34e7a2104f121b1bb/src/datasets/utils/py_utils.py#L170-L171

Just wondering if there might be other data classes default values missed that could cause an issue... Apart from feature-like classes and tasks, I don't see any other though...

And a question: but this information about the tasks is no longer being saved as YAML tags in the dataset card; won't be a problem with current datasets using task templates (with this information in their metadata JSON) once we replace the JSON by the YAML tags (which do not have this information about the task templates)?

lhoestq:
> Just wondering if there might be other data classes default values missed that could cause an issue... Apart from feature-like classes and tasks, I don't see any others though...

I think we're good ! `asdict` is used on the DatasetInfo attributes like features, tasks etc. and they all support dict conversion properly now

> And a question: but this information about the tasks is no longer being saved as YAML tags in the dataset card; won't be a problem with current datasets using task templates (with this information in their metadata JSON) once we replace the JSON by the YAML tags (which do not have this information about the task templates)?

In the long run we'll use the train_eval_index YAML tags instead, but I agree when removing the JSON files we should try to not break existing code that may rely on this

Issue #2650, opened 2021-07-14
[load_dataset] shard and parallelize the process


stas00:
- Some huge datasets take forever to build the first time. (e.g. oscar/en) as it's done in a single cpu core.
- If the build crashes, everything done up to that point gets lost

Request: Shard the build over multiple arrow files, which would enable:
- much faster build by parallelizing the build process
- if the process crashed, the completed arrow files don't need to be re-built again

Thank you!

@lhoestq

lhoestq:
I think @TevenLeScao is exploring adding multiprocessing in `GeneratorBasedBuilder._prepare_split` - feel free to post updates here :)

TevenLeScao:
Posted a PR to address the building side, still needs something to load sharded arrow files + tests

mariosasko:
Closing as this feature has been implemented in #5107

PR #5107, opened 2022-10-12
Multiprocessed dataset builder [WIP]

TevenLeScao:
This PR adds the multiprocessing part of #2650 (but not the caching of already-computed arrow files). On the other side, loading of sharded arrow files still needs to be implemented (sharded parquet files can already be loaded).


TevenLeScao:
I would also like to add a test, but am not sure whether it should go into `test_builder` (more natural imo) or `test_load` (which already contains a lot of the things I have to import to run my current testing setup). For reference, what I run to test that it works looks like:

```
import os
from pathlib import Path
import shutil

import datasets
from datasets.builder import DatasetBuilder
from datasets.features import Features, Value

DATASET_LOADING_SCRIPT_NAME = "__dummy_dataset1__"

DATASET_LOADING_SCRIPT_CODE = """
import os

import datasets
from datasets import DatasetInfo, Features, Split, SplitGenerator, Value


class __DummyDataset1__(datasets.GeneratorBasedBuilder):

    def _info(self) -> DatasetInfo:
        return DatasetInfo(features=Features({"text": Value("string")}))

        return [
            SplitGenerator(Split.TRAIN, gen_kwargs={"filepaths": [os.path.join(dl_manager.manual_dir, "train1.txt"), os.path.join(dl_manager.manual_dir, "train2.txt")]}),
            SplitGenerator(Split.TEST, gen_kwargs={"filepaths": [os.path.join(dl_manager.manual_dir, "test.txt")]}),
        ]

    def _generate_examples(self, filepaths, **kwargs):
        idx = 0
        for filepath in filepaths:
            with open(filepath, "r", encoding="utf-8") as f:
                for line in f:
                    yield idx, {"text": line.strip()}
                    idx += 1
"""


def dataset_loading_script_dir(tmp_path):
    script_name = DATASET_LOADING_SCRIPT_NAME
    script_dir = tmp_path / script_name
    script_dir.mkdir()
    script_path = script_dir / f"{script_name}.py"
    with open(script_path, "w") as f:
        f.write(DATASET_LOADING_SCRIPT_CODE)
    return str(script_dir)


def data_dir(tmp_path):
    data_dir = tmp_path / "data_dir"
    data_dir.mkdir()
    with open(data_dir / "train1.txt", "w") as f:
        f.write("foo\n" * 10)
    with open(data_dir / "train2.txt", "w") as f:
        f.write("foo\n" * 10)
    with open(data_dir / "test.txt", "w") as f:
        f.write("bar\n" * 10)
    return str(data_dir)


def load_dataset_builder_multiprocessed(tmp_path):
    builder = datasets.load_dataset_builder(
        os.path.join(dataset_loading_script_dir(tmp_path), DATASET_LOADING_SCRIPT_NAME + ".py"),
        data_dir=data_dir(tmp_path),
    )
    assert isinstance(builder, DatasetBuilder)
    assert builder.name == DATASET_LOADING_SCRIPT_NAME
    assert builder.info.features == Features({"text": Value("string")})
    builder.download_and_prepare(tmp_path / "prepare_target", max_shard_size=500, num_proc=2)

if __name__ == "__main__":
    tmp_path = "tmp"
    if os.path.exists(tmp_path):
        raise FileExistsError(f"path {tmp_path} already exists")
    os.makedirs(tmp_path)
    try:
        load_dataset_builder_multiprocessed(Path(tmp_path))
    finally:
        # pass
        shutil.rmtree(tmp_path)
```

lhoestq:
Nice ! I think the test can go in `test_builder.py` :)

src/datasets/iterable_dataset.py
```diff
@@ -133,6 +133,45 @@ def _shard_kwargs(shard_idx: int, kwargs: dict) -> dict:
     return {key: [value[shard_idx]] if isinstance(value, list) else value for key, value in kwargs.items()}
 
 
+def _all_shard_kwargs(kwargs: dict) -> list:
```
lhoestq, line 136: Those functions' sharding logic were previously only used for iterable datasets, but now they're part of dataset builders logic, so feel free to move them in `builder.py`

src/datasets/builder.py
```diff
@@ -1467,43 +1538,125 @@ def _prepare_split(
...
+        features = features_per_rank[0]
+
+        # should rename everything at the end, scheme still TBD
+        def _rename_shard(shard_id_and_rank: Tuple[int]):
+            shard_id, rank = shard_id_and_rank
+            global_shard_id = sum(shards_per_rank[:rank]) + shard_id
+            self._rename(
+                fpath.replace("SSSSS", f"{shard_id:05d}").replace("RRRRR", f"{rank:05d}"),
+                fpath.replace("RRRRR-SSSSS", f"{global_shard_id:05d}").replace("NNNNN", f"{total_shards:05d}"),
+            )
```
lhoestq, line 1624: Do you think we can preserve the order of the dataset with a smart way to rename those shards ?

If every arrow shard contains the index of the args_per_shard that was used, I think we can get the order of the original dataset back.

This is not a super high priority though

lhoestq (commented):
Thanks for doing it for both the GeneratorBasedBuilder and the ArrowBasedBuilder !

I added a comment and started a discussion about file naming:

src/datasets/iterable_dataset.py
TevenLeScao: Wouldn't that cause a circular import since `builder` also imports from `iterable_dataset` and `iterable_dataset` requires `_shard_kwargs` ? Unless you mean move the other two? I'd be in favour of keeping them all in the same place

lhoestq: Oh yes indeed - you can keep it this way then for now

src/datasets/builder.py
TevenLeScao: It does! Or at least, this preserves the order of the shards in `split_generator.gen_kwargs`.

TevenLeScao: Actually it doesn't after testing, but I can't quite figure out why :/

TevenLeScao:
I've added sharded arrow dataset loading. Two WIP items in the PR:
- Order is not conserved (it seems like the sharded files are read in the wrong order)
- the tqdm for preparing the splits is wrong (it compares against the size of the whole split rather than against the size of the multiprocessing shard, but I am not sure how to access the latter)

Also `naming.filenames_for_dataset_split` is not very elegant imo.

@lvwerra if you don't care about order, as I do, it's functional for now but I'd still quite like to get to the bottom of this.

TevenLeScao:
I fixed the tqdm to be less misleading, but it can't tell where to stop. I am a bit hesitant to add a top-level tqdm (on the shard iterator) since for most intents it will do 0 -> N shards straight, but I am not sure what is the best way to present that info here.

lhoestq:
I'm continuing the PR :)

lhoestq:
Did a few changes:
- make shards naming consistent:
  - use `{builder_name}-{split_name}.{file_format}` when there's only 1 shard
  - otherwise use `{builder_name}-{split_name}-{shard_idx:05d}-of-{num_shards:05d}.{file_format}`
- update the reader to support reading several shards
  - added a new `shard_lengths` field in `SplitInfo` (FYI it is saved in `dataset_info.json` next to the shards as usual)
    - it's None when there's only 1 shard
    - otherwise it's a list of integers that correspond to the number of rows per shard
  - implemented partial reading to only memory map the required shards
    - e.g. when someone asks for a partial split like `train[:10%]`
- align the sharding for beam datasets
  - no more combining into 1 big arrow file
- added a tqdm bar
  - only one single bar, handled by the main process
  - gathers progress updates from other processes using `iflatmap_unordered`
  - shows the number of examples (even for datasets prepared by generating arrow tables)
- tests
- docs

lhoestq:
Alright this is ready for review - sorry it ended up so big ^^'

If I can do anything to make it easier for your to review this PR @mariosasko let me know

lhoestq:
Multiprocessing is disabled by default but we may show a warning to encourage users to pass `num_proc` if the dataset is split in many files. Let me know what you think

src/datasets/builder.py
```diff
@@ -1270,6 +1278,83 @@ def _get_examples_iterable_for_split(self, split_generator: SplitGenerator) -> E
         raise NotImplementedError()
 
 
+def _number_of_shards(gen_kwargs: dict) -> int:
```
mariosasko, line 1281: IIRC we use a similar code to shard IterableDataset, so maybe we can put this code in `py_utils.py` and use it there to ensure consistency.

src/datasets/builder.py
```diff
@@ -1331,63 +1416,183 @@ def _prepare_split(
...
             split_info = split_generator.split_info
 
-        suffix = "-SSSSS-of-NNNNN" if file_format == "parquet" else ""
-        fname = f"{self.name}-{split_generator.name}{suffix}.{file_format}"
+        SUFFIX = "-JJJJJ-SSSSS-of-NNNNN"
+        fname = f"{self.name}-{split_generator.name}{SUFFIX}.{file_format}"
         fpath = path_join(self._output_dir, fname)
 
-        generator = self._generate_examples(**split_generator.gen_kwargs)
+        # Default to using 16-way parallelism for preparation if the number of files is higher than 16.
```
mariosasko, line 1436: This is no longer the default behavior

```suggestion
```

And we can align `num_proc` from `load_dataset` with `download_config.num_proc` in a subsequent PR.

src/datasets/builder.py
```diff
@@ -1331,63 +1416,183 @@ def _prepare_split(
...
+                f"Setting num_proc from {num_proc} to {num_input_shards} for the {split_info.name} split as it only contains {num_input_shards} shards."
+            )
+            num_proc = num_input_shards
+
+        pbar = logging.tqdm(
+            unit=" examples",
+            total=split_info.num_examples,
+            leave=False,
+            desc=f"Generating {split_info.name} split",
+        )
```
mariosasko, line 1454: "disable" for consistency:

```suggestion
        pbar = logging.tqdm(
            disable=not logging.is_progress_bar_enabled(),
            unit=" examples",
            total=split_info.num_examples,
            leave=False,
            desc=f"Generating {split_info.name} split",
        )
```

src/datasets/builder.py
```diff
@@ -1467,79 +1665,202 @@ def _prepare_split(
...
```
mariosasko, line 1702:
```suggestion
        pbar = logging.tqdm(
            disable=not logging.is_progress_bar_enabled(),
            unit=" examples",
            total=split_info.num_examples,
            leave=False,
            desc=f"Generating {split_info.name} split",
        )
```

IMO it would be better to use only one unit for the main progress bar and the child ones here

mariosasko (commented):
Looks great!

Do we have some benchmarks to see the speed-up?

Some nits:

parsa-ra:
Hey, is this error seems to you guys natural? 

```bash
>>> data = load_dataset('dataset_loaders/rfw2latentplay', num_proc=14)
  File "/somewhere//mambaforge/envs/datasets/lib/python3.8/site-packages/datasets/load.py", line 1719, in load_dataset
    builder_instance = load_dataset_builder(
  File "/somewhere//mambaforge/envs/datasets/lib/python3.8/site-packages/datasets/load.py", line 1523, in load_dataset_builder
    builder_instance: DatasetBuilder = builder_cls(
  File "/somewhere//mambaforge/envs/datasets/lib/python3.8/site-packages/datasets/builder.py", line 1292, in __init__
    super().__init__(*args, **kwargs)
  File "/somewhere//mambaforge/envs/datasets/lib/python3.8/site-packages/datasets/builder.py", line 303, in __init__
    self.config, self.config_id = self._create_builder_config(
  File "/somewhere//mambaforge/envs/datasets/lib/python3.8/site-packages/datasets/builder.py", line 456, in _create_builder_config
    builder_config = self.BUILDER_CONFIG_CLASS(**config_kwargs)
TypeError: __init__() got an unexpected keyword argument 'num_proc'
```

lhoestq:
> Do we have some benchmarks to see the speed-up?

On my machine running `load_dataset("oscar-corpus/OSCAR-2201", "br")` (which is split in shards) I go from 2-3k examples per sec to 4-5k examples per sec with num_proc=2 😉

lhoestq:
> Hey, is this error seems to you guys natural?
>
> The package built from 0d4e3907 commit tag, and here is the version displayed from the import ...

I don't know where you got the `0d4e3907` commit tag from, it doesn't seem to be in this PR. You should try installing from this PR, or wait for it to be merged on `main`

src/datasets/arrow_writer.py
```diff
@@ -636,19 +636,33 @@ def finalize(self, metrics_query_result: dict):
...
             try:  # stream conversion
-                sources = [beam.io.filesystems.FileSystems.open(shard) for shard in shards]
-                with beam.io.filesystems.FileSystems.create(self._path) as dest:
-                    parquet_to_arrow(sources, dest)
+                disable = not logging.is_progress_bar_enabled()
+                num_bytes = 0
+                for shard in logging.tqdm(shards, unit="shards", disable=disable):
+                    with beam.io.filesystems.FileSystems.open(shard) as source:
+                        with beam.io.filesystems.FileSystems.create(
+                            shard.replace(".parquet", ".arrow")
```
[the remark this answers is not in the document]
lhoestq: I agree ! We can keep that in mind

src/datasets/builder.py
lhoestq: This pbar is not supposed to be there, thanks for catching that.
There's only one pbar handled by the parent process, which counts examples/sec

parsa-ra:
## Splits vs Shards

Maybe it's a good idea to add some documentation on the `sharding` that can be achieved by passing `list` based arguments to the `SplitGenerator`s `gen_kwargs` ...

lhoestq:
> Maybe it's a good idea to add some documentation on the sharding that can be achieved by passing list based arguments to the SplitGenerators gen_kwargs ...

This is part of this PR :) you can check the changes in docs/source/dataset_script.mdx

lhoestq:
I took your comments into account @mariosasko thanks !
Let me know if it's good for you now ;)

src/datasets/builder.py
mariosasko: This code now resides in `utils/sharding.py`. Marking as resolved.

mariosasko (approved):
Looks all good now (besides the failing "PR docs" job, but not sure if we need to address this)!

PR #5108, opened 2022-10-13
Fix a typo in arrow_dataset.py

yangky11:


lhoestq (approved):
good catch, thanks :)

Issue #5111, opened 2022-10-13
map and filter not working properly in multiprocessing with the new release 2.6.0


loubnabnl:
## Describe the bug
When mapping is used on a dataset with more than one process, there is a weird behavior when trying to use `filter` , it's like only the samples from one worker are retrieved, one needs to specify the same `num_proc` in filter for it to work properly. This doesn't happen with `datasets` version 2.5.2

In the code below the data is filtered differently when we increase `num_proc` used in `map` although the datsets before and after mapping have identical elements.

## Steps to reproduce the bug
```python
import datasets
from datasets import load_dataset

def preprocess(example):
    return example

ds = load_dataset("codeparrot/codeparrot-clean-valid", split="train").select([i for i in range(10)])
ds1 = ds.map(preprocess, num_proc=2)
ds2 = ds.map(preprocess)

# the datasets elements are the same
for i in range(len(ds1)):
    assert ds1[i]==ds2[i]

print(f'Target column before filtering {ds1["autogenerated"]}')
print(f'Target column before filtering {ds2["autogenerated"]}')
print(f"datasets version {datasets.__version__}")

ds_filtered_1 = ds1.filter(lambda x: not x["autogenerated"])
ds_filtered_2 = ds2.filter(lambda x: not x["autogenerated"])

# all elements in Target column are false so they should all be kept, but for ds2 only the first 5=num_samples/num_proc are kept
print(ds_filtered_1)
print(ds_filtered_2)
```
```
Target column before filtering [False, False, False, False, False, False, False, False, False, False]
Target column before filtering [False, False, False, False, False, False, False, False, False, False]

Dataset({
    features: ['repo_name', 'path', 'copies', 'size', 'content', 'license', 'hash', 'line_mean', 'line_max', 'alpha_frac', 'autogenerated'],
    num_rows: 5
})
Dataset({
    features: ['repo_name', 'path', 'copies', 'size', 'content', 'license', 'hash', 'line_mean', 'line_max', 'alpha_frac', 'autogenerated'],
    num_rows: 10
})
```
## Expected results
Increasing `num_proc` in mapping shouldn't alter filtering. With the previous version 2.5.2 this doesn't happen

## Actual results
Filtering doesn't work properly when we increase `num_proc` in mapping but not when calling `filter`

## Environment info

- `datasets` version: 2.6.0
- Platform: Linux-4.19.0-22-cloud-amd64-x86_64-with-glibc2.28
- Python version: 3.9.13
[...]

PR #5113, opened 2022-10-14
Fix filter indices when batched

albertvillanova:
This PR fixes a bug introduced by:
- #5030

Fix #5112.

PR #5115, opened 2022-10-14
Fix iter_batches

lhoestq:
The `pa.Table.to_reader()` method available in `pyarrow>=8.0.0` may return chunks of size < `max_chunksize`, therefore `iter_batches` can return batches smaller than the `batch_size` specified by the user

Therefore batched `map` couldn't always use batches of the right size, e.g. this fails because it runs only on one batch of one element:
```python

ds = concatenate_datasets([Dataset.from_dict({"a": [i]}) for i in range(10)])

ds2 = ds.map(lambda _: {}, batched=True)
assert list(ds2) == list(ds)
```

This was introduced in https://github.com/huggingface/datasets/pull/5030

Close https://github.com/huggingface/datasets/issues/5111

This will require a patch release along with https://github.com/huggingface/datasets/pull/5113

TODO:
- [x] fix tests
- [x] add more tests

### DIFF
--- a/src/datasets/table.py
+++ b/src/datasets/table.py
@@ -2154,7 +2154,9 @@ def table_iter_batches(pa_table: pa.Table, batch_size: int, drop_last_batch=Fals
     chunks_buffer = []
     chunks_buffer_size = 0
     for chunk in pa_table.to_reader(max_chunksize=batch_size):
-        if chunks_buffer_size + len(chunk) < batch_size:
+        if len(chunk) == 0:
+            continue
+        elif chunks_buffer_size + len(chunk) < batch_size:
             chunks_buffer.append(chunk)
             chunks_buffer_size += len(chunk)
             continue
@@ -2169,5 +2171,5 @@ def table_iter_batches(pa_table: pa.Table, batch_size: int, drop_last_batch=Fals
             yield pa.Table.from_batches(chunks_buffer)
             chunks_buffer = [chunk.slice(cropped_chunk_length, len(chunk) - cropped_chunk_length)]
             chunks_buffer_size = cropped_chunk_length
-    if not drop_last_batch:
+    if not drop_last_batch and chunks_buffer:
         yield pa.Table.from_batches(chunks_buffer)
--- a/src/datasets/table.py
+++ b/src/datasets/table.py
@@ -2148,3 +2148,26 @@ def table_visitor(table: pa.Table, function: Callable[[pa.Array], None]):
 
     for name, feature in features.items():
         _visit(table[name], feature)
+
+
+def table_iter_batches(pa_table: pa.Table, batch_size: int, drop_last_batch=False):
+    chunks_buffer = []
+    chunks_buffer_size = 0
+    for chunk in pa_table.to_reader(max_chunksize=batch_size):
+        if chunks_buffer_size + len(chunk) < batch_size:
+            chunks_buffer.append(chunk)
+            chunks_buffer_size += len(chunk)
+            continue
+        elif chunks_buffer_size + len(chunk) == batch_size:
+            chunks_buffer.append(chunk)
+            yield pa.Table.from_batches(chunks_buffer)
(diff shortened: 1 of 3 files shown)


lhoestq:
I also ran the code in https://github.com/huggingface/datasets/issues/5111 and it works fine now :)

lhoestq:
This is ready for review :)

src/datasets/table.py
```diff
@@ -2148,3 +2148,35 @@ def _visit(array, feature):
...
+            chunks_buffer.append(chunk)
+            yield pa.Table.from_batches(chunks_buffer)
+            chunks_buffer = []
+            chunks_buffer_size = 0
+        else:
+            cropped_chunk_length = batch_size - chunks_buffer_size
+            chunks_buffer.append(chunk.slice(0, cropped_chunk_length))
+            yield pa.Table.from_batches(chunks_buffer)
+            chunks_buffer = [chunk.slice(cropped_chunk_length, len(chunk) - cropped_chunk_length)]
+            chunks_buffer_size = len(chunk) - cropped_chunk_length
```
albertvillanova, line 2184: Maybe we can remove the variable `chunks_buffer_size`:
```suggestion
    for chunk in pa_table.to_reader(max_chunksize=batch_size):
        if len(chunk) == 0:
            continue
        elif len(chunks_buffer) + len(chunk) < batch_size:
            chunks_buffer.append(chunk)
            continue
        elif len(chunks_buffer) + len(chunk) == batch_size:
            chunks_buffer.append(chunk)
            yield pa.Table.from_batches(chunks_buffer)
            chunks_buffer = []
        else:
            cropped_chunk_length = batch_size - len(chunks_buffer)
            chunks_buffer.append(chunk.slice(0, cropped_chunk_length))
            yield pa.Table.from_batches(chunks_buffer)
            chunks_buffer = [chunk.slice(cropped_chunk_length, len(chunk) - cropped_chunk_length)]
```

src/datasets/table.py
```diff
@@ -2148,3 +2148,35 @@ def _visit(array, feature):
...
+    if not drop_last_batch and chunks_buffer:
+        yield pa.Table.from_batches(chunks_buffer)
```
albertvillanova, line 2186: I'm just wondering if this function may have a performance impact, instead of just calling `for batch in self.data.to_reader(max_chunksize=batch_size)`, as before.

If so, we should check how much impact, so that we do not lose the performance gain introduced by #5030.

albertvillanova (approved):
Thanks for the fix.

Just a few comments below.

src/datasets/table.py
```diff
@@ -2148,3 +2148,39 @@ def _visit(array, feature):
 
     for name, feature in features.items():
         _visit(table[name], feature)
+
+
+def table_iter(pa_table: pa.Table, batch_size: int, drop_last_batch=False):
+    """Iterate ober sub-tables of size `batch_size`.
```
albertvillanova, line 2154: Typo:
```suggestion
    """Iterate over sub-tables of size `batch_size`.
```

src/datasets/table.py
lhoestq: `chunks_buffer_size` is the sum of the length of all the chunks in the buffer - not just the length of the buffer

src/datasets/table.py
lhoestq: The code is roughly the same as in #5030

Also note that the worst case scenario is when the dataset is made of chunks of length 1, but even in this case this is faster than calling `__getitem__` for each item. 
```python
ds = concatenate_datasets([Dataset.from_dict({"a": [i]}) for i in range(100)])
%time list(ds)
# 3ms
%time [ds[i] for i in range(len(ds))]
# 5ms
```

It's even better for big datasets, since `__getitem__` is not O(1) because of interpolation search. Here getting the next item is O(1)

mariosasko (approved):
Thanks, LGTM!

Issue #5117, opened 2022-10-14
Progress bars have color red and never completed to 100%


echatzikyriakidis:
## Describe the bug
Progress bars after transformative operations turn in red and never be completed to 100%

## Steps to reproduce the bug
```python
from datasets import load_dataset

load_dataset('rotten_tomatoes', split='test').filter(lambda o: True)
```

## Expected results
Progress bar should be 100% and green

## Actual results
Progress  bar turn in red and never completed to 100%

## Environment info
- `datasets` version: 2.6.1
- Platform: Linux-5.10.133+-x86_64-with-Ubuntu-18.04-bionic
- Python version: 3.7.14
- PyArrow version: 6.0.1
- Pandas version: 1.3.5

david1542:
Hi @echatzikyriakidis, thanks for submitting the issue.
Which shell are you using exactly? I tried to run the command you sent, but I don't see colors at all 🧐

#self-assign

david1542:
#self-assign

echatzikyriakidis:
Hi @david1542 ,

I use Google Colab.

david1542:
Got it. I created a PR that fixes this issue. Turns out that the wrapping logic for the inner loop was slightly incorrect.

echatzikyriakidis:
Thank you!

11301858:
Hello @mariosasko 

I am still facing this issue. Was this problem fixed?

![Screen Shot 2024-06-19 at 11 35 57 AM](https://github.com/huggingface/datasets/assets/66180831/cd9c6ac4-178c-4297-821c-a56411d34c89)

I cleared the hugging face cache before running, and no error message was given. Let me know if you need a minimal repro of my code.

PR #5120, opened 2022-10-16
Fix `tqdm` zip bug

david1542:
This PR solves #5117, by wrapping the entire `zip` clause in tqdm.

For more information, please checkout this Stack Overflow thread:
https://stackoverflow.com/questions/41171191/tqdm-progressbar-and-zip-built-in-do-not-work-together

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -2935,13 +2935,13 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                         len(input_dataset) if not drop_last_batch else len(input_dataset) // batch_size * batch_size
                     )
                     pbar_total = (num_rows // batch_size) + 1 if num_rows % batch_size else num_rows // batch_size
-                    pbar_iterable = input_dataset._iter_batches(
+                    pbar_iterable = zip(range(0, num_rows, batch_size), input_dataset._iter_batches(
                         batch_size, decoded=False, drop_last_batch=drop_last_batch
-                    )
+                    ))
                 pbar_unit = "ex" if not batched else "ba"
                 pbar_desc = (desc + " " if desc is not None else "") + "#" + str(rank) if rank is not None else desc
                 pbar = logging.tqdm(
-                    zip(range(0, num_rows, batch_size), pbar_iterable),
+                    pbar_iterable,
                     total=pbar_total,
                     disable=disable_tqdm,
                     position=rank,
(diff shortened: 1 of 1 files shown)


albertvillanova (changes requested):
Thanks for trying to fix this bug, @david1542.

However the proposed solution will not work: please note that `pbar` is also used if `not batched`: https://github.com/huggingface/datasets/blob/07b7c38d9e9c72c74b02524c432ca64d0d3738f4/src/datasets/arrow_dataset.py#L2951-L2952
- there it is expected that iterating over `pbar` returns examples, not a tuple like with your implementation.

david1542:
@albertvillanova Thanks for your comment. What do you think about creating 2 `pbar` for each case? I see the `pbar_iterable` is initialized differently. Maybe `pbar` can also be initialized like that.

david1542:
@albertvillanova Another solution I implemented is to change `pbar_iterable` and add the `zip` to it. I updated the PR with this solution. Let me know what you think.

albertvillanova (changes requested):
Thanks for the improvement.

I think it is a good idea to include the indices in `pbar_iterable`.

You need to fix the style: please run:
```shell
make style
```

To be consistent, I would recommend to do the same (include the `enumerate` in `pbar_iterable`) for the non batched case.

dudulasry:
@albertvillanova Done :) Let me know what you think.

david1542:
@albertvillanova Thanks :) I also don't see an easy way to test this. This was just a problem in the way `tqdm` was used. I'm not sure we should cover it in tests.

Cochonaki:
Hi, 

First of all, thanks for this PR. 
It's the first time I join a discussion on GitHUB on problem resolution in libraries such as transformers, so I hope I comply to the best practices for an efficient communication...

I am running `AutoTokenizer.from_pretrained` in a Google Colab notebook for using with BERT base. 
I am experiencing issue [5117](https://github.com/huggingface/datasets/issues/5117).

Each time I run my notebook, I do:

`! pip install transformers 
! pip install datasets 
! pip install huggingface_hub`

As I understand, the issue has been resolved and the solution merged to the released version of the code?
So I expect that the bug is resolved in my notebook, however this is not the case.

Do I get something wrong? 
Do I have to implement some change in the source code myself?

Thanks in advance for your help!

david1542:
@Cochonaki Hi :) The problem was fixed but there wasn't a release since then. I believe a new release should come out in the upcoming weeks. Maybe someone from the core maintainers can answer that :)

cc: @albertvillanova

armw4:
Baby Haiti Coffee SE is born

NH watch

albertvillanova:
Hi, @Cochonaki.

As @david1542 pointed out, we have not made a release since this bug was fixed. We will make one in the following weeks.

In the meantime, if you would like to incorporate the bug fix, you can install `datasets` from this repo main branch:
```shell
pip install git+https://github.com/huggingface/datasets#egg=datasets
```

Cochonaki:
Thanks a lot @albertvillanova and @david1542, it works now!
I am really thankful for your help, that encourages me to participate more in this community.
See you around!

albertvillanova:
Welcome!!! 🤗

PR #5124, opened 2022-10-17
Install tensorflow-macos dependency conditionally

### LINKED ISSUE #5118: Installing `datasets` on M1 computers
## Describe the bug
I wanted to install `datasets` dependencies on my M1 (in order to start contributing to the project). However, I got an error regarding `tensorflow`.

On M1, `tensorflow-macos` needs to be installed instead. Can we add a conditional requirement, so that `tensorflow-macos` would be installed on M1?

## Steps to reproduce the bug
Fresh clone this project (on m1), create a virtualenv and run this:
```python
pip install -e ".[dev]"
```

## Expected results
Installation should be smooth, and all the dependencies should be installed on M1.

## Actual results
You should receive an error, saying pip couldn't find a version that matches this pattern:
```
tensorflow>=2.3,!=2.6.0,!=2.6.1
```

## Environment info

- `datasets` version: 2.6.2.dev0
- Platform: macOS-12.6-arm64-arm-64bit
- Python version: 3.9.6
- PyArrow version: 7.0.0
- Pandas version: 1.5.0

albertvillanova:
Fix #5118.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -132,7 +132,8 @@ TESTS_REQUIRE = [
     "py7zr",
     "rarfile>=4.0",
     "s3fs>=2021.11.1",  # aligned with fsspec[http]>=2021.11.1
-    "tensorflow>=2.3,!=2.6.0,!=2.6.1",
+    "tensorflow>=2.3,!=2.6.0,!=2.6.1; sys_platform != 'darwin' or platform_machine != 'arm64'",
+    "tensorflow-macos; sys_platform == 'darwin' and platform_machine == 'arm64'",
     "torch",
     "torchaudio<0.12.0",
     "soundfile",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -132,7 +132,8 @@
     "py7zr",
     "rarfile>=4.0",
     "s3fs>=2021.11.1",  # aligned with fsspec[http]>=2021.11.1
-    "tensorflow>=2.3,!=2.6.0,!=2.6.1",
+    "tensorflow>=2.3,!=2.6.0,!=2.6.1; sys_platform != 'darwin' or platform_machine != 'arm64'",
+    "tensorflow-macos; sys_platform == 'darwin' and platform_machine == 'arm64'",
```
severo, line 136: BTW I think `and platform_machine == 'arm64'"` is not needed, as I was able to install `tensorflow-macos` on my MacOS with an Intel chip

```
$ sysctl -a | grep Intel
machdep.cpu.vendor: GenuineIntel
machdep.cpu.brand_string: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
```

See https://github.com/huggingface/datasets-server/blob/9b346a79c5e3f51719fb2717683c4e5f86291201/services/worker/pyproject.toml#L34-L35

But maybe it's better to limit the exception cases and only use the fork for arm64.

severo (approved):

setup.py
albertvillanova: @severo I'm not an expert on this subject. But it appears they recommend `tensorflow-macos` also for Intel x86-64: https://developer.apple.com/metal/tensorflow-plugin/

So you think I can safely remove `and platform_machine == 'arm64'`? Or better keeping to reduce the scope?

severo: I really don't know, I just wanted to share my recent experience with the same problem, and use the same solution if you opt for a reduced scope.

PR #5125, opened 2022-10-17
Add `pyproject.toml` for `black`

mariosasko:
Add `pyproject.toml` as a config file for the `black` tool to support VS Code's auto-formatting on save (and to be more consistent with the other HF projects).


albertvillanova (approved):
Thanks for the addition. Better to be consistent!

PR #5126, opened 2022-10-17
Fix class name of symbolic link

### LINKED ISSUE #5098: Classes label error when loading symbolic links using imagefolder
Like this: #4015
When there are **symbolic links** to pictures in the data folder, the parent folder name of the **real file** will be used as the class name instead of the parent folder of the symbolic link itself. Can you give an option to decide whether to enable symbolic link tracking?

This is inconsistent with the `torchvision.datasets.ImageFolder` behavior.

For example:
image
image

It use `others` in green circle as class label but not `abnormal`, I wish `load_dataset` not use the real file parent as label.

**Additional context**
Add any other context about the feature request here.

riccardobucco:
Fix #5098

### DIFF
--- a/src/datasets/data_files.py
+++ b/src/datasets/data_files.py
@@ -275,7 +275,7 @@ def _resolve_single_pattern_locally(
     fs = LocalFileSystem()
     glob_iter = [PurePath(filepath) for filepath in fs.glob(pattern) if fs.isfile(filepath)]
     matched_paths = [
-        Path(filepath).resolve()
+        Path(os.path.abspath(filepath))
         for filepath in glob_iter
         if (filepath.name not in FILES_TO_IGNORE or PurePath(pattern).name == filepath.name)
         and not _is_inside_unrequested_special_dir(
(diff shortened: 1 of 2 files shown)


albertvillanova:
This should be commented in the PR description, so that the PR is appropriately linked by GitHub to its corresponding Issue:

> Fix #5098.

albertvillanova (changes requested):
Thanks for the fix, @riccardobucco.

Could you please implement a regression test as well? This is considered a good practice in software development.

albertvillanova:
The regression test that should be implemented should test what your PR fixes: that is, that `_resolve_single_pattern_locally` function does not resolve any symbolic link when passed a directory that does contain any.

As you are testing a function in `data_files.py`, the corresponding test should be in `tests/test_data_files.py`.

You could take inspiration from other tests there in that file.

albertvillanova (approved):
Thanks @riccardobucco for having tried, and thanks @mariosasko for the proposal of the fix solution and the addition of the regression test...

PR #5128, opened 2022-10-18
Make filename matching more robust

### LINKED ISSUE #5046: Audiofolder creates empty Dataset if files same level as metadata
## Describe the bug
When audio files are at the same level as the metadata (`metadata.csv` or `metadata.jsonl` ), the `load_dataset` returns a `DatasetDict` with no rows but the correct columns.

https://github.com/huggingface/datasets/blob/1ea4d091b7a4b83a85b2eeb8df65115d39af3766/docs/source/audio_dataset.mdx?plain=1#L88

## Steps to reproduce the bug
`metadata.csv`:
```csv
file_name,duration,transcription
./2063_fe9936e7-62b2-4e62-a276-acbd344480ce_1.wav,10.768,hello
```

```python
>>> audio_dataset = load_dataset("audiofolder", data_dir="/audio-data/")
>>> audio_dataset
DatasetDict({
    train: Dataset({
        features: ['audio', 'duration', 'transcription'],
        num_rows: 0
    })
    validation: Dataset({
        features: ['audio', 'duration', 'transcription'],
        num_rows: 0
    })
})
```

I've tried, with no success,:
- setting `split` to something else so I don't get a `DatasetDict`,
- removing the `./`,
- using `.jsonl`.

## Expected results
```
 Dataset({
        features: ['audio', 'duration', 'transcription'],
        num_rows: 1
})
```

## Actual results
```
DatasetDict({
    train: Dataset({
        features: ['audio', 'duration', 'transcription'],
        num_rows: 0
    })
    validation: Dataset({
        features: ['audio', 'duration', 'transcription'],
        num_rows: 0
    })
})
```

## Environment info
- `datasets` version: 2.5.1
- Platform: Linux-5.13.0-1025-aws-x86_64-with-glibc2.29
- Python version: 3.8.10
- PyArrow version: 9.0.0
- Pandas version: 1.5.0

riccardobucco:
Fix  #5046

### DIFF
--- a/src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
+++ b/src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
@@ -316,7 +316,7 @@ class FolderBasedBuilder(datasets.GeneratorBasedBuilder):
                                 pa_metadata_table = pa_metadata_table.drop(["file_name"])
                                 metadata_dir = os.path.dirname(metadata_file)
                                 metadata_dict = {
-                                    file_name: sample_metadata
+                                    os.path.normpath(file_name): sample_metadata
                                     for file_name, sample_metadata in zip(
                                         pa_file_name_array.to_pylist(), pa_table_to_pylist(pa_metadata_table)
                                     )
@@ -385,7 +385,7 @@ class FolderBasedBuilder(datasets.GeneratorBasedBuilder):
                                     pa_metadata_table = pa_metadata_table.drop(["file_name"])
                                     metadata_dir = os.path.dirname(downloaded_metadata_file)
                                     metadata_dict = {
-                                        file_name: sample_metadata
+                                        os.path.normpath(file_name): sample_metadata
                                         for file_name, sample_metadata in zip(
                                             pa_file_name_array.to_pylist(), pa_table_to_pylist(pa_metadata_table)
                                         )
--- a/src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
+++ b/src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
@@ -316,7 +316,7 @@ class FolderBasedBuilder(datasets.GeneratorBasedBuilder):
                                 pa_metadata_table = pa_metadata_table.drop(["file_name"])
                                 metadata_dir = os.path.dirname(metadata_file)
                                 metadata_dict = {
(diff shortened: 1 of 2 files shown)


src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
```diff
@@ -316,7 +316,7 @@ def _generate_examples(self, files, metadata_files, split_name, add_metadata, ad
                                 pa_metadata_table = pa_metadata_table.drop(["file_name"])
                                 metadata_dir = os.path.dirname(metadata_file)
                                 metadata_dict = {
-                                    file_name: sample_metadata
+                                    os.path.normpath(file_name): sample_metadata
```
mariosasko, line 319: `os.path.normpath` inserts the OS-specific file separator, so we need to call `replace("\\", "/")` to make sure it's "/" on all the platforms to avoid the CI error.

```suggestion
                                    os.path.normpath(file_name).replace("\\", "/"): sample_metadata
```

PS: this also means we should remove the `pa_file_name_array = pc.replace_substring(pa_file_name_array, pattern="\\", replacement="/")` line from the script as it's no longer needed.

src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
```diff
@@ -385,7 +385,7 @@ def _generate_examples(self, files, metadata_files, split_name, add_metadata, ad
                                     pa_metadata_table = pa_metadata_table.drop(["file_name"])
                                     metadata_dir = os.path.dirname(downloaded_metadata_file)
                                     metadata_dict = {
-                                        file_name: sample_metadata
+                                        os.path.normpath(file_name): sample_metadata
```
mariosasko, line 388: Same here

mariosasko (commented):
Thanks!

I think we should also modify one of the metadata files in the `folder_based_builder` tests to make sure "./" is ignored now in the `file_name`

riccardobucco:
> I think we should also modify one of the metadata files in the `folder_based_builder` tests to make sure "./" is ignored now in the `file_name`

@mariosasko what do you mean here? I'm not sure which metadata file I should modify here

mariosasko (approved):
Thanks, looks all good now!

lhoestq (approved):
Thanks ! LGTM :)

PR #5130, opened 2022-10-18
Avoid extra cast in `class_encode_column`

mariosasko:
Pass the updated features to `map` to avoid the `cast` in `class_encode_column`.


lhoestq (approved):
nice !

PR #5140, opened 2022-10-19
Make the KeyHasher FIPS compliant

vvalouch:
MD5 is not FIPS compliant thus I am proposing this minimal change to make datasets package FIPS compliant


lhoestq (commented):
I think it's fine if we keep md5 - we don't use it for security stuff.
Thanks for bringing it up though

PR #5141, opened 2022-10-19
Raise ImportError instead of OSError 

### LINKED ISSUE #5134: Raise ImportError instead of OSError if required extraction library is not installed
According to the official Python docs, `OSError` should be thrown in the following situations:
> This exception is raised when a system function returns a system-related error, including I/O failures such as “file not found” or “disk full” (not for illegal argument types or other incidental errors).

Hence, it makes more sense to raise `ImportError` instead of `OSError` when the required extraction/decompression library is not installed.

ayushthe1:
fixes #5134  : Replaced OSError with ImportError if required extraction library is not installed.

### DIFF
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -299,7 +299,7 @@ def _copy_script_and_other_resources_in_importable_dir(
                 if not os.path.exists(full_path_local_import):
                     shutil.copytree(import_path, full_path_local_import)
             else:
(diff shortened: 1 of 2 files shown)


mariosasko (commented):
The extraction is happing in `extract.py`, so please also replace the errors there, and we can merge.

ayushthe1:
Thanks @mariosasko ,i commited the changes as you said.

PR #5136, opened 2022-10-19
Update docs once dataset scripts transferred to the Hub

albertvillanova:
Todo:
- [x] Update docs:
  - [x] Datasets on GitHub (legacy)
  - [x] Load: offline
  - [x] About dataset load:
    - [x] Maintaining integrity
    - [x] Security
- [x] Update docstrings:
  - [x] Inspect:
    - [x] get_dataset_config_info
    - [x] get_dataset_split_names
  - [x] Load:
    - [x] dataset_module_factory
    - [x] load_dataset_builder
    - [x] load_dataset
- [x] Remove `ADD_NEW_DATASET.md`
- [x] Update `.github/ISSUE_TEMPLATE/config.yml`

Fix #5135.

PR #5146, opened 2022-10-21
Delete duplicate issue template file

albertvillanova:
A conflict between two PRs:
- #5116
- #5136

was not properly resolved, resulting in a duplicate issue template.

This PR removes the duplicate template.


lhoestq (approved):
good catch ^^'

PR #5154, opened 2022-10-24
Test latest fsspec in CI

lhoestq:
Following the discussion in https://discuss.huggingface.co/t/attributeerror-module-fsspec-has-no-attribute-asyn/19255 I think we need to test the latest fsspec in the CI


lhoestq:
actually the latest fsspec is already installed

PR #5155, opened 2022-10-24
TextConfig: added "errors"

NightMachinery:


src/datasets/packaged_modules/text/text.py
```diff
@@ -19,6 +19,7 @@ class TextConfig(datasets.BuilderConfig):
 
     features: Optional[datasets.Features] = None
     encoding: str = "utf-8"
+    errors: str = "strict"
```
lhoestq, line 22: I think the default for `open` is `None`, should we have `None` has well for consistency ? It has the same effect than "strict" anyway
```suggestion
    errors: Optional[str] = None
```

lhoestq:
Thanks for adding this ! You can fix the CI by formatting your code using the `make style` command :)

src/datasets/packaged_modules/text/text.py
NightMachinery: Is there a reason for the default being `None`? It's more confusing that way, both in understanding what it does and in the type signature. Is there any advantage to it?

NightMachinery:
[**@lhoestq**](https://github.com/lhoestq) commented on [Oct 27, 2022, 4:08 PM GMT+3:30](https://github.com/huggingface/datasets/pull/5155#issuecomment-1293464680 "2022-10-27T12:38:04Z - Replied by Github Reply Comments"):
> Thanks for adding this ! You can fix the CI by formatting your code using the `make  style` command :)

I ran this and force pushed the changes.

src/datasets/packaged_modules/text/text.py
lhoestq: It doesn't count really into account - it's just a way to make it explicit that the default behavior of `open` is used (default for open is None as well).

However I agree it's not enough to understand what it does. In that case we can maybe add docstrings to the TextConfig class and document all the parameters

NightMachinery: [**@lhoestq**](https://github.com/lhoestq) commented on [Nov 3, 2022, 4:23 PM GMT+3:30](https://github.com/huggingface/datasets/pull/5155#discussion_r1012860965 "2022-11-03T12:53:22Z - Replied by Github Reply Comments"):
> I'm merging this one if you don't mind, I think it's better to have the same default as `open`. We can add docstrings to explain this in a subsequent PR

Yes, it makes sense. Thanks!

ُا

lhoestq (approved):
Thanks for adding this argument :)

PR #5159, opened 2022-10-25
fsspec lock reset in multiprocessing

lhoestq:
`fsspec` added a clean way of resetting its lock - instead of doing it manually

### DIFF
--- a/src/datasets/formatting/dataset_wrappers/torch_iterable_dataset.py
+++ b/src/datasets/formatting/dataset_wrappers/torch_iterable_dataset.py
@@ -1,5 +1,6 @@
-import fsspec.asyn
 import threading
+
+import fsspec.asyn
 import torch
 
 from ...iterable_dataset import IterableDataset, _apply_feature_types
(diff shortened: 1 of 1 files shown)


src/datasets/formatting/dataset_wrappers/torch_iterable_dataset.py
```diff
@@ -15,8 +17,13 @@ def _set_fsspec_for_multiprocess() -> None:
...
     """
-    fsspec.asyn.iothread[0] = None
-    fsspec.asyn.loop[0] = None
+    if hasattr(fsspec.asyn, "reset_lock"):
+        # for future fsspec>2022.05.0
+        fsspec.asyn.reset_lock()
+    else:
+        fsspec.asyn.iothread[0] = None
+        fsspec.asyn.loop[0] = None
+        fsspec.asyn.lock = threading.Lock()
```
lhoestq, line 26: previously, we were not resetting the lock correctly apparently - I added this line anyway

PR #5163, opened 2022-10-25
Reduce default max `writer_batch_size`

mariosasko:
Reduce the default writer_batch_size from 10k to 1k examples. Additionally, align the default values of `batch_size` and `writer_batch_size` in `Dataset.cast` with the values from the corresponding docstring.


lhoestq (approved):
Sounds good ! Hopefully it helps avoiding memory issues with image or audio datasets

PR #5173, opened 2022-10-27
Raise ffmpeg warnings only once

polinaeterna:
Our warnings looks nice now.

`librosa` warning that was raised at each decoding: 
```
/usr/local/lib/python3.7/dist-packages/librosa/core/audio.py:165: UserWarning: PySoundFile failed. Trying audioread instead.
  warnings.warn("PySoundFile failed. Trying audioread instead.")
```
is suppressed with `filterwarnings("ignore")` in a context manager. That means the first warning is also ignored (setting `filterwarnings("once")` didn't work!), so I added info that audioread is used for decoding to our message. Hope it's enough.

Tests failed at first because they used to check if the warning was raised at (each) decoding in `librosa` case but now we throw only one warning (at first decoding). I removed this check for warnings, do you think it's fine?

### DIFF
--- a/src/datasets/features/audio.py
+++ b/src/datasets/features/audio.py
@@ -354,7 +354,7 @@ class Audio:
         global _audioread_warned
 
         with warnings.catch_warnings():
-            if not _audioread_warned:
+            if _audioread_warned:
                 warnings.filterwarnings("ignore", "pysoundfile failed.+?", UserWarning, module=librosa.__name__)
             else:
                 _audioread_warned = True
--- a/src/datasets/features/audio.py
+++ b/src/datasets/features/audio.py
@@ -18,7 +18,7 @@ if TYPE_CHECKING:
     from .features import FeatureType
 
 
-_ffmpeg_warned, _librosa_warned = True, True
+_ffmpeg_warned, _librosa_warned = False, False
 
 
 @dataclass
--- a/src/datasets/features/audio.py
+++ b/src/datasets/features/audio.py
@@ -299,14 +299,14 @@ class Audio:
                 array, sampling_rate = self._decode_mp3_torchaudio(path_or_file)
             except RuntimeError:
                 global _ffmpeg_warned
-                if _ffmpeg_warned == 0:
+                if not _ffmpeg_warned:
                     warnings.warn(
                         "\nTo support 'mp3' decoding with `torchaudio>=0.12.0`, please install `ffmpeg4` system package. On Google Colab you can run:\n\n"
                         "\t!add-apt-repository -y ppa:jonathonf/ffmpeg-4 && apt update && apt install -y ffmpeg\n\n"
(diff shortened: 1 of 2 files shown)


src/datasets/features/audio.py
```diff
@@ -342,5 +353,8 @@ def _decode_mp3_torchaudio(self, path_or_file):
     def _decode_mp3_librosa(self, path_or_file):
         import librosa
 
-        array, sampling_rate = librosa.load(path_or_file, mono=self.mono, sr=self.sampling_rate)
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", "PySoundFile.+?", UserWarning, module=librosa.__name__)
+            array, sampling_rate = librosa.load(path_or_file, mono=self.mono, sr=self.sampling_rate)
```
lhoestq, line 358: maybe you can show the first warning this way:
```suggestion
            if _audioread_warned:
                warnings.filterwarnings("ignore", "PySoundFile.+?", UserWarning, module=librosa.__name__)
            else:
                _audioread_warned = True
            array, sampling_rate = librosa.load(path_or_file, mono=self.mono, sr=self.sampling_rate)
```

src/datasets/features/audio.py
```diff
@@ -309,11 +315,16 @@ def _decode_mp3(self, path_or_file):
...
-                        "\tpip install \"torchaudio<0.12\"`.\n\nOtherwise 'mp3' files will be decoded with `librosa`:\n\n"
+                        "\tpip install \"torchaudio<0.12\".\n\nTo decode 'mp3' files without torchaudio, please install `librosa`:\n\n"
                         "\tpip install librosa\n\nNote that decoding will be extremely slow in that case."
                     ) from err
                 # try to decode with librosa for torchaudio>=0.12.0 as a workaround
-                warnings.warn("Decoding mp3 with `librosa` instead of `torchaudio`, decoding is slow.")
+                global _librosa_warned
+                if not _librosa_warned:
+                    warnings.warn(
+                        "Decoding mp3 with `librosa` and `audioread` instead of `torchaudio`, decoding is slow."
```
lhoestq, line 325: and in that case we don;t need to mention audioread here. If the user has a working `libsndfile` setup then it doesn't use audioread anyway
```suggestion
                        "Decoding mp3 with `librosa` instead of `torchaudio`, decoding is slow."
```

lhoestq (commented):
Awesome thanks :)

added a few comments regarding the librosa+audioread warning

src/datasets/features/audio.py
polinaeterna: hm true! probably mac and windows users do have a working setup if they have the latest version of `pysoundfile`, didn't think about that (being too linux-centric apparently...)

lhoestq (approved):
Looks all good ! thank you :)

PR #5174, opened 2022-10-28
Preserve None in list type cast in PyArrow 10

mariosasko:
The `ListArray` type in PyArrow 10.0.0 supports the `mask` parameter, which allows us to preserve Nones in nested lists in `cast` instead of replacing them with empty lists.

Fix https://github.com/huggingface/datasets/issues/3676


lhoestq (approved):
omg finally !

thanks for updating `array_cast` :)

Issue #5170, opened 2022-10-27
[Caching] Deterministic hashing of torch tensors


lhoestq:
Currently this fails
```python
import torch
from datasets.fingerprint import Hasher

t = torch.tensor([1.])

def func(x):
    return t + x

hash1 = Hasher.hash(func)
t = torch.tensor([1.])
hash2 = Hasher.hash(func)
assert hash1 == hash2
```

Also as noticed in https://discuss.huggingface.co/t/dataset-cant-cache-models-outputs/24945, using a model in a `map` function doesn't work well with caching. Indeed the `bert-base-uncased` model has a different hash every time you reload it. Supporting torch tensors may also help in this case.

This can be fixed by registering a custom pickling functions for torch tensors - as we did for other objects such as CodeType, FunctionType and Regex in `py_utils.py`

PR #5191, opened 2022-11-02
Make torch.Tensor and spacy models cacheable

mariosasko:
Override `Pickler.save` to implement deterministic reduction (lazily registered; inspired by https://github.com/uqfoundation/dill/blob/master/dill/_dill.py#L343) functions for `torch.Tensor` and spaCy models.

Fix https://github.com/huggingface/datasets/issues/5170, fix https://github.com/huggingface/datasets/issues/3178

### DIFF
--- a/src/datasets/utils/py_utils.py
+++ b/src/datasets/utils/py_utils.py
@@ -589,6 +589,65 @@ class Pickler(dill.Pickler):
 
     dispatch = dill._dill.MetaCatchingDict(dill.Pickler.dispatch.copy())
 
+    def save(self, obj, save_persistent_id=True):
+        # lazy registration of reduction functions
+        obj_type = type(obj)
+        if obj_type not in Pickler.dispatch:
+            if (obj_type.__module__, obj_type.__name__) == ("_regex", "Pattern"):
+                try:
+                    import regex
+
+                    @pklregister(obj_type)
+                    def _save_regex(pickler, obj):
+                        dill._dill.log.info(f"Re: {obj}")
+                        args = (
+                            obj.pattern,
+                            obj.flags,
+                        )
+                        pickler.save_reduce(regex.compile, args, obj=obj)
+                        dill._dill.log.info("# Re")
+                        return
+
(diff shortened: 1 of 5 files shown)


src/datasets/utils/py_utils.py
```diff
@@ -599,6 +599,109 @@ class Pickler(dill.Pickler):
...
+                        def _save_lang(pickler, obj):
+                            def _create_lang(config, bytes_data):
+                                lang_cls = spacy.util.get_lang_class(config["nlp"]["lang"])
+                                nlp = lang_cls.from_config(config)
+                                return nlp.from_bytes(bytes_data)
+
+                            dill._dill.logger.trace(pickler, f"Sp: {obj}")
+                            args = (obj.config, obj.to_bytes())
+                            pickler.save_reduce(_create_lang, args, obj=obj)
+                            dill._dill.logger.trace(pickler, "# Sp")
```
lhoestq, line 697: it's maybe simpler to select the dill logger outside this function depending on `dill`'s version ? This way you don't have to duplicate the code.

src/datasets/utils/py_utils.py
```diff
@@ -599,6 +599,109 @@ class Pickler(dill.Pickler):
...
+            elif (obj_type.__module__, obj_type.__name__) == ("torch", "Tensor"):
+                try:
+                    import torch
+
+                    if config.DILL_VERSION < version.parse("0.3.6"):
+
+                        @pklregister(obj_type)
+                        def _save_tensor(pickler, obj):
+                            dill._dill.log.info(f"To: {obj}")
+                            args = (obj.cpu().numpy(),)
```
lhoestq, line 647: (nit) in case the model is training:
```suggestion
                            if obj.requires_grad:
                                args = (obj.detach().cpu().numpy(),)
                            else:
                                args = (obj.cpu().numpy(),)
```

lhoestq (approved):
Love it thanks ! Added some minor comments, but overall LGTM :)

Issue #5153, opened 2022-10-24
default Image/AudioFolder infers labels when there is no metadata files even if there is only one dir


polinaeterna:
### Describe the bug

By default FolderBasedBuilder infers labels if there is not metadata files, even if it's meaningless (for example, they are in a single directory or in the root folder, see this repo as an example: https://huggingface.co/datasets/patrickvonplaten/audios

As this is a corner case for quick exploration of images or audios on the Hub. 

### Steps to reproduce the bug

If you have directory like this:

```
repo
   image1.jpg
   image2.jpg
   image3.jpg
```
or

```
repo
   data
      image1.jpg
      image2.jpg
      image3.jpg
```
doing `ds = load_dataset(repo)` would create `label` feature:
```python
print(ds["train"][0])
>> {'image': <PIL.JpegImagePlugin.JpegImageFile image mode=RGB size=500x375 at 0x7FB5326468E0>, 'label': 0}
```

Also, if you have the following structure:

```
repo
   data
      image1.jpg
      image2.jpg
      image3.jpg
   image4.jpg
   image5.jpg
   image6.jpg
```
it will infer two labels:

```python
print(ds["train"][0])
print(ds["train"][-1])
>> {'image': <PIL.JpegImagePlugin.JpegImageFile image mode=RGB size=500x375 at 0x7FB5326468E0>, 'label': 1}
>> {'image': <PIL.JpegImagePlugin.JpegImageFile image mode=RGB size=500x415 at 0x7FB5326555B0>, 'label': 0}
```

### Expected behavior

We should have only one base feature (Image/Audio) in such cases.

### Environment info

all versions of `datasets`

mariosasko:
Makes sense! For the last structure, we could count the path segments (delimited by "/" for URLs and `os.sep`  for local paths) to ensure all inferred labels are on the same level. Otherwise, I think it's safe to assume they are meaningless and ignore them.

PR #5192, opened 2022-11-02
Drop labels in Image and Audio folders if files are on different levels in directory or if there is only one label

polinaeterna:
Will close https://github.com/huggingface/datasets/issues/5153

Drop labels by default (`drop_labels=None`) when:
* there are files on different levels of directory hierarchy by checking their path depth
* all files are in the same directory (=only one label was inferred)

First one fixes cases like this:
```
repo
   image3.jpg
   image4.jpg
   data
      image1.jpg
      image2.jpg
```
Second one fixes cases like this:
```
repo
   image1.jpg
   image2.jpg
   image3.jpg
```
This is mostly to fix the viewer for people who just drop images in the Hub interface into the root dir.

I added tests for both of the cases on local and remote files. **I also changed data files for old test on drop_labels** (`test_generate_examples_drop_labels`). The files I provide to `test_generate_examples_drop_labels` now has "canonical" classification structure (two dirs) in order not to change the logic of the test (=not to check these two cases addressed in the PR).

### DIFF
--- a/src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
+++ b/src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
@@ -11,7 +11,6 @@ import pyarrow.json as paj
 import datasets
 from datasets.features.features import FeatureType
 from datasets.tasks.base import TaskTemplate
-from datasets.utils.file_utils import is_remote_url
 
 
 logger = datasets.utils.logging.get_logger(__name__)
--- a/src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
+++ b/src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
@@ -86,7 +86,7 @@ class FolderBasedBuilder(datasets.GeneratorBasedBuilder):
         # * `drop_labels` is None (default) or False, to infer the class labels
         # * `drop_metadata` is None (default) or False, to find the metadata files
         do_analyze = not self.config.drop_labels or not self.config.drop_metadata
-        labels = set()
+        labels, path_depths = set(), set()
         metadata_files = collections.defaultdict(set)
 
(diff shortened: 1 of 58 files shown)


lhoestq (commented):
Awesome ! It looks all good to me :)

Feel free to mark your PR as "ready for review"

mariosasko (approved):
Thanks, LGTM!

Nit:  maybe we can use the `count_path_segments` function from this file for counting (updated with your logic to make it faster).

lhoestq (approved):
It looks all good, thanks !

polinaeterna:
> Nit: maybe we can use the count_path_segments function from this file for counting (updated with your logic to make it faster).

@mariosasko just to make sure I understood you correctly - are you okay with this change? (actually `os.path.normpath` is redundant here as paths from `data_files` should be already normalized but just in case)
https://github.com/huggingface/datasets/pull/5192/files#diff-1f09f7a178211f7539b1499b64b69793bd53b30c8b7b34cfcc5835e25d31929fR33
If you are, we can merge.

src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
```diff
@@ -30,17 +31,9 @@ def pa_table_to_pylist(table):
...
-            break
-        elif parts[1] == path:
-            break
-        else:
-            path = parts[0]
-            cnt += 1
-    return cnt
+    if is_remote_url(path):
+        return path.count("/")
+    return os.path.normpath(path).count(os.sep)
```
mariosasko, line 36: Yes, we resolve the paths in `data_files.py`, so `normpath` is redundant here.

This should also work, no:
```suggestion
    return path.replace("\\", "/").count("/")
```

mariosasko (commented):
One nit. Besides that, looks all good.

lhoestq:
awesome ! :D

PR #5201, opened 2022-11-04
Do not sort splits in dataset info

polinaeterna:
I suggest not to sort splits by their names in dataset_info in README so that they are displayed in the order specified in the loading script. Otherwise `test` split is displayed first, see this repo: https://huggingface.co/datasets/paws
What do you think?

But I added sorting in tests to fix CI (for the same dataset).

### DIFF
--- a/src/datasets/splits.py
+++ b/src/datasets/splits.py
@@ -505,7 +505,7 @@ class SplitReadInstruction:
         return split_instruction
 
     def get_list_sliced_split_info(self):
-        return list(sorted(self._splits.values(), key=lambda x: x.split_info.name))
+        return self._splits.values()
 
 
 class SplitDict(dict):
@@ -569,7 +569,7 @@ class SplitDict(dict):
         """Returns a list of SplitInfo protos that we have."""
         # Return the SplitInfo, sorted by name
         out = []
-        for split_name, split_info in sorted(self.items()):
+        for split_name, split_info in self.items():
             split_info = copy.deepcopy(split_info)
             split_info.name = split_name
             out.append(split_info)
--- a/src/datasets/splits.py
+++ b/src/datasets/splits.py
@@ -505,7 +505,7 @@ class SplitReadInstruction:
         return split_instruction
 
     def get_list_sliced_split_info(self):
-        return self._splits.values()
+        return list(self._splits.values())
 
 
 class SplitDict(dict):
--- a/src/datasets/splits.py
+++ b/src/datasets/splits.py
@@ -567,7 +567,6 @@ class SplitDict(dict):
 
     def to_split_dict(self):
         """Returns a list of SplitInfo protos that we have."""
-        # Return the SplitInfo, sorted by name
         out = []
         for split_name, split_info in self.items():
             split_info = copy.deepcopy(split_info)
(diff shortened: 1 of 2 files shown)


severo:
It would be coherent with https://github.com/huggingface/datasets-server/issues/614#issuecomment-1290534153

albertvillanova:
I think we started working on this issue nearly at the same time... :sweat_smile: 

This PR should fix:
- #5202

polinaeterna:
@albertvillanova yeah I noticed it right after the PR :smile: thank you! the fix of the dataset info yaml fixes tests on CI, but in general order of splits in yaml influences the order in which they are displayed in the viewer, if I understand it correctly. So I suggest not to sort splits in yaml initially to avoid this for other datasets in the future. I think [this change](https://github.com/huggingface/datasets/pull/5201/files#diff-198ba4fdf2f94cb3e1aba8a0170a43b08d4ab5636d682374321c5a383a8be24dR571) should work for it. 

Changes to tests here maybe can be reverted considering that order in yaml now corresponds to the one in tests, thanks to your change in the dataset info.

tests/test_inspect.py
```diff
@@ -87,7 +87,7 @@ def test_get_dataset_info(path, expected_configs, expected_splits_in_first_confi
     assert expected_config in infos
     info = infos[expected_config]
     assert info.config_name == expected_config
-    assert list(info.splits.keys()) == expected_splits_in_first_config
+    assert sorted(info.splits.keys()) == sorted(expected_splits_in_first_config)
```
albertvillanova, line 90: I'm not sure we want to avoid testing the order: as already discussed by you @polinaeterna and @severo, splits are not alphabetically sorted. 

Therefore, it makes sense to test that the order returned by `get_dataset_infos` is the expected one.

Anyway, if finally we decide to sort them, we should do it also in `test_get_dataset_config_info`, which was also failing besides `test_get_dataset_info` and `test_get_dataset_split_names`.

albertvillanova (commented):
Thanks for the fix, @polinaeterna.

I agree we should not sort splits alphabetically, but keep them in their original order.

However, I disagree we should add `sorted` to our tests: I think we have to test the returned order (see comment below).

albertvillanova:
Hehe, @polinaeterna, we make comments nearly at the same time as well... :laughing:

tests/test_inspect.py
polinaeterna: yes, agree with you! reverted sorting in tests [a063c6f](https://github.com/huggingface/datasets/pull/5201/commits/a063c6f4b407d5e3bafe789692ca22208da320d1)

src/datasets/splits.py
```diff
@@ -505,7 +505,7 @@ def __getitem__(self, slice_value):
         return split_instruction
 
     def get_list_sliced_split_info(self):
-        return list(sorted(self._splits.values(), key=lambda x: x.split_info.name))
+        return self._splits.values()
```
albertvillanova, line 508: I think this is not used anywhere else... But, just in case, better returning a list here?

albertvillanova (approved):
Thanks for the suggested improvement.

Just a comment below.

src/datasets/splits.py
polinaeterna: didn't find any usages either... applied your suggestion [6e10c33](https://github.com/huggingface/datasets/pull/5201/commits/6e10c33259d213f83ff92fb702b98f6fa3cffb07), thank you!

Issue #5151, opened 2022-10-24
Add ability to create different configs with `push_to_hub` (+ inferring configs from directories with package managers?)


polinaeterna:
Now one can push only different splits within one default config of a dataset.

Would be nice to allow something like:
```
ds.push_to_hub(repo_name, config=config_name)
```

I'm not sure, but this will probably require changes in `data_files.py` patterns. If so, it would also allow to create different configs for packaged modules datasets.

lhoestq:
also asked in https://discuss.huggingface.co/t/create-multiple-dataset-configs-with-push-to-hub-method/25480

Issue #5209, opened 2022-11-07
Implement ability to define splits in metadata section of dataset card


merveenoyan:
### Feature request

If you go here: https://huggingface.co/datasets/inria-soda/tabular-benchmark/tree/main you will see bunch of folders that has various CSV files. I’d like dataset viewer to show these files instead of only one dataset like it currently does. (and also people to be able to load them as splits instead of loading through `data_files`) 
e.g GLUE has various splits on viewer but it’s too overkill to ask people to implement loading script, so it would be better to let them define these in the README file instead.

Also pinging @polinaeterna @lhoestq @adrinjalali

polinaeterna:
@merveenoyan Do you want different files to be splits or configurations?

From [what you specified in `Readme.md`](https://huggingface.co/datasets/inria-soda/tabular-benchmark/commit/fb4575853772c62a20203bdd6cc0202f5db4ce4e) I hypothesize that you want to have 4 **configs** corresponding to directories: `"clf_cat", "clf_num", "reg_cat", "reg_num"`. And inside each config you require to have as many splits as there are `csv` files
so if you run 
```python
load_dataset("inria-soda/tabular-benchmark", "clf_cat", split="compass")
```
you will generate the data only from `compass.csv` file.
In this case, running `load_dataset("inria-soda/tabular-benchmark", "clf_cat"`) without split parameter will return `DatasetDict` object with `"KDDCup09_upselling", "cat_compass", "cat_covertype", ... "road_safety"` keys (which values are splits - `Dataset` objects)

**or**
do you want each file to be a separate config? Like:
```python
load_dataset("inria-soda/tabular-benchmark", "clf_cat_compass")   # returns DatasetDict with a single "train" split
```
**or**
maybe smth completely different? :smile:  

Anyway, now I have an impression that this is probably rather a matter of automatically inferring configs from repository structure rather than providing parameters in metadata yaml.

merveenoyan:
@polinaeterna I want the latter where you can think of every CSV file as a config, like MNLI from GLUE.

polinaeterna:
@merveenoyan @lhoestq I see two solutions to this case.  
1. Parse configurations automatically from directories names. That is, if you have data structure like:
```
tabular-benchmark
  └─clf_cat_compass
      └─compass.csv
  └─clf_cat_cat_covertype
      └─covertype.csv
 ...
  └─reg_cat_house_sales
      └─house_sales.csv
```
[...]

PR #5213, opened 2022-11-08
Add support for different configs with `push_to_hub`

polinaeterna:
will solve  #5151

@lhoestq @albertvillanova @mariosasko 
This is still a super draft so please ignore code issues but I want to discuss some conceptually important things.

I suggest a way to do `.push_to_hub("repo_id", "config_name")` with pushing parquet files to directories named as `config_name` (inside `data/` dir as it is now), for example: 
```
data
   |__config-v1
        train-00000-00002-...-.parquet
        train-00001-00002-...-.parquet
        ...
   |__config-v2
        ....
```

When loading a dataset, I parse these configs from repository data files (only for `"data/{split}-[0-9][0-9][0-9][0-9][0-9]-of-[0-9][0-9][0-9][0-9][0-9]*.*"` pattern that is used for parquet datasets pushed with `.push_to_hub`). 
Therefore, 
- when user tries to load a dataset that has configs parsed from data files dir names without providing a config (like `load_dataset("repo")` instead of `load_dataset("repo", "config-v1")`) - raise error and asks for config - to be aligned with how it works in datasets with scripts.
- for backward compatibility: if user tries to `.push_to_hub(""repo", "config_name")` to an existing parquet repo with no configurations (all parquet files are directly in `data/` dir) - raise error. My initial idea was to raise a warning and move these files to another dir with name (config) like "default" or smth but in a PR and suggest user to merge it on the Hub. But there is no support for renaming (moving) files via `HfApi` yet so it would require deleting and pushing again if I understand it right.

This parsing approach can be extended to other Hub packaged modules, and to local packaged modules and other data files patterns
(except for cases when splits are in dir names `KEYWORDS_IN_DIR_NAME_BASE_PATTERNS` because we allow for arbitrary depth of directory hierarchy).
Do you think it's reasonable? Not sure how to provide flexibility (and backward compatibility) to not parsing configs and load all the data in a single config as it is now.

I also thought about getting information about configs from Readme.md `dataset_info` ([example](https://huggingface.co/datasets/polinaeterna/test_push_two_configs/blob/main/README.md)). But that way we 
are dependent on if it exists. It is created automatically with `.push_to_hub` but what if it is 
accidentally deleted or smth). 

Also, what I don't like is that this parsing is a part of Module/DataFiles logic, not Builder's one, which is not aligned with datasets with custom scripts. But I don't know to implement the second approach in current library's logic. 

 What do you think about this all? Am I missing smth?

TODO:

- [ ] save cache in the same dir for configs of the same datasets
- [ ] fix verification errors
- [ ] correctly update `dataset_infos.json` too
- [ ] ...

### DIFF
--- a/.github/workflows/build_documentation.yml
+++ b/.github/workflows/build_documentation.yml
@@ -2,10 +2,8 @@ name: Build documentation
 
 on:
   push:
-    branches:
-      - main
-      - doc-builder*
-      - v*-release
+    tags:
+      - '[0-9]+.[0-9]+.[0-9]+*'
 
 jobs:
    build:
--- a/.github/workflows/build_pr_documentation.yml
+++ b/.github/workflows/build_pr_documentation.yml
@@ -9,8 +9,11 @@ concurrency:
 
 jobs:
   build:
-    uses: huggingface/doc-builder/.github/workflows/build_pr_documentation.yml@main
+    uses: huggingface/doc-builder/.github/workflows/build_pr_documentation.yml@use_hf_hub
     with:
       commit_sha: ${{ github.event.pull_request.head.sha }}
       pr_number: ${{ github.event.number }}
       package: datasets
+    secrets:
+      token: ${{ secrets.HF_DOC_PUSH }}
+      comment_bot_token: ${{ secrets.HUGGINGFACE_PUSH }}
--- a/.github/workflows/build_pr_documentation.yml
+++ b/.github/workflows/build_pr_documentation.yml
@@ -9,11 +9,8 @@ concurrency:
 
 jobs:
   build:
-    uses: huggingface/doc-builder/.github/workflows/build_pr_documentation.yml@use_hf_hub
+    uses: huggingface/doc-builder/.github/workflows/build_pr_documentation.yml@main
     with:
       commit_sha: ${{ github.event.pull_request.head.sha }}
       pr_number: ${{ github.event.number }}
       package: datasets
-    secrets:
-      token: ${{ secrets.HF_DOC_PUSH }}
-      comment_bot_token: ${{ secrets.HUGGINGFACE_PUSH }}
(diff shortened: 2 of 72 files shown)


lhoestq:
Nice thanks !

Would it be possible to have the new folders at the same level as "data" ? This way they're all separated
```
├─ config-v1/
│  ├── train-00000-00002-...-.parquet
│  └── train-00001-00002-...-.parquet
└ config-v2/
    ├── train-00000-00002-...-.parquet
    └── train-00001-00002-...-.parquet
```
and if you don't provide a config name, it goes in a folder named "default" instead, that would be loaded by default.

We could also write in the YAML something like
```yaml
configs:
- name: config-v1
  data_dir: config-v1
- name: config-v2
  data_dir: config-v2
```
and loading `config-v1` would be equivalent to run `load_dataset(ds_name, "config-v1", data_dir="config-v1")`

Do you think it would make sense ?

For backward compatibility we can just keep the "data/*" pattern. It's ok to expect users to have an updated version of `datasets` to be able to load datasets with configurations.

polinaeterna:
@lhoestq thank you for the feedback! i'll reflect on this on Moday, my mind just melted because of the fever.

@mariosasko  @albertvillanova what do you think?

albertvillanova:
Thanks for addressing this, @polinaeterna. It is good:
- we support configs for datasets without scripts
- we align the behavior to datasets with scripts as much as possible

Maybe adding some tests will help clarify what is the expected behavior...

polinaeterna:
After some discussion with @lhoestq we decided that it's better to rely on metadata file than on data files patterns. 

So we decided to introduce a new field to yaml (like `configs` or smth like that) that would contain arbitrary configs kwargs to be passed to loader, including `data_dir` and `data_files`. 
This is more aligned with datasets with custom scripts where we explicitly write all the supported configs and config parameters in the code and is extendable to all packaged modules.
This would solve https://github.com/huggingface/datasets/issues/5209

(@lhoestq was right 21 days ago, this is a more general solution idk why i ignored this...)

polinaeterna:
closed in favor of https://github.com/huggingface/datasets/pull/5331

PR #5234, opened 2022-11-14
fix: dataset path should be absolute

vigsterkr:
cache_file_name depends on dataset's path.

A simple way where this could cause a problem:
```
import os
import datasets


def add_prefix(example):
    example["text"] = "Review: " + example["text"]
    return example


ds = datasets.load_from_disk("a/relative/path")
os.chdir("/tmp")

ds_1 = ds.map(add_prefix)
```

while it may feel that the `chdir` is quite constructed, there are many scenarios when the current working dir can/will change...

### DIFF
--- a/src/datasets/table.py
+++ b/src/datasets/table.py
@@ -1028,7 +1028,7 @@ class MemoryMappedTable(TableBlock):
 
     def __init__(self, table: pa.Table, path: str, replays: Optional[List[Replay]] = None):
         super().__init__(table)
-        self.path = path
+        self.path = os.path.abspath(path)
         self.replays: List[Replay] = replays if replays is not None else []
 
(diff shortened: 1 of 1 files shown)


lhoestq:
Good catch thanks ! Have you tried to use the absolue path in `MemoryMappedTable.__init__` in `table.py`?

I think it can fix issues with relative paths at more levels than just fixing it `load_from_disk`. If it works I think it would be a more robust fix to this issue

vigsterkr:
@lhoestq right, that actually fixed it indeed. I've pushed the changes (one-liner). lemme know if there's anything else you need for this fix

lhoestq (approved):
Awesome thanks !

PR #5236, opened 2022-11-14
Handle ArrowNotImplementedError caused by try_type being Image or Audio in cast

mariosasko:
Handle the `ArrowNotImplementedError` thrown when `try_type` is `Image` or `Audio` and the input array cannot be converted to their storage formats.

Reproducer:
```python
from datasets import Dataset
from PIL import Image
import requests

ds = Dataset.from_dict({"image": [Image.open(requests.get("https://upload.wikimedia.org/wikipedia/commons/e/e9/Felis_silvestris_silvestris_small_gradual_decrease_of_quality.png", stream=True).raw)]})
ds.map(lambda x: {"image": True}) # ArrowNotImplementedError 
```

PS: This could also be fixed by raising `TypeError` in `{Image, Audio}.cast_storage` for unsupported types instead of passing the array to `array_cast.`

### DIFF
--- a/src/datasets/arrow_writer.py
+++ b/src/datasets/arrow_writer.py
@@ -199,7 +199,15 @@ class TypedSequence:
                 # We only do it if trying_type is False - since this is what the user asks for.
                 out = cast_array_to_feature(out, type, allow_number_to_str=not self.trying_type)
             return out
-        except (TypeError, pa.lib.ArrowInvalid) as e:  # handle type errors and overflows
+        except (
+            TypeError,
+            pa.lib.ArrowInvalid,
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
thanks :)

Not sure how we can have a test that is relevant for this though - feel free to add one if you have ideas. Otherwise I think it's fine anyway

mariosasko:
> Not sure how we can have a test that is relevant for this though - feel free to add one if you have ideas

Yes, this was my reasoning for not adding a test. This change is pretty simple, so I think it's OK not to have a test for it.

PR #5237, opened 2022-11-14
Encode path only for old versions of hfh

lhoestq:
Next version of `huggingface-hub` 0.11 does encode the `path`, and we don't want to encode twice


albertvillanova (approved):
Thanks for the fix.

Issue #5230, opened 2022-11-11
dataclasses error when importing the library in python 3.11


yonikremer:
### Describe the bug

When I import datasets using python 3.11 the dataclasses standard library raises the following error:
`ValueError: mutable default <class 'datasets.utils.version.Version'> for field version is not allowed: use default_factory`

When I tried to import the library using the following jupyter notebook:
```
%%bash
# create python 3.11 conda env
conda create --yes --quiet -n myenv -c conda-forge python=3.11
# activate is
source activate myenv
# install pyarrow
/opt/conda/envs/myenv/bin/python -m pip install --quiet --extra-index-url https://pypi.fury.io/arrow-nightlies/ \
    --prefer-binary --pre pyarrow
# install datasets
/opt/conda/envs/myenv/bin/python -m pip install --quiet datasets
```

```
# create a python file that only imports datasets
with open("import_datasets.py", 'w') as f:
    f.write("import datasets")
# run it with the env
!/opt/conda/envs/myenv/bin/python import_datasets.py
```

I get the following error:
```
Traceback (most recent call last):
  File "/kaggle/working/import_datasets.py", line 1, in <module>
    import datasets
  File "/opt/conda/envs/myenv/lib/python3.11/site-packages/datasets/__init__.py", line 45, in <module>
    from .builder import ArrowBasedBuilder, BeamBasedBuilder, BuilderConfig, DatasetBuilder, GeneratorBasedBuilder
  File "/opt/conda/envs/myenv/lib/python3.11/site-packages/datasets/builder.py", line 91, in <module>
    @dataclass
     ^^^^^^^^^
  File "/opt/conda/envs/myenv/lib/python3.11/dataclasses.py", line 1221, in dataclass
    return wrap(cls)
           ^^^^^^^^^
  File "/opt/conda/envs/myenv/lib/python3.11/dataclasses.py", line 1211, in wrap
    return _process_class(cls, init, repr, eq, order, unsafe_hash,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/envs/myenv/lib/python3.11/dataclasses.py", line 959, in _process_class
    cls_fields.append(_get_field(cls, name, type, kw_only))
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #5238, opened 2022-11-14
Make `Version` hashable

mariosasko:
Add `__hash__` to the `Version` class to make it hashable (and remove the unneeded methods), as  `Version("0.0.0")` is the default value of `BuilderConfig.version` and the default fields of a dataclass need to be hashable in Python 3.11.

Fix https://github.com/huggingface/datasets/issues/5230


lhoestq (approved):
Sounds good to me !

PR #5241, opened 2022-11-14
Support hfh rc version

lhoestq:
otherwise the code doesn't work for hfh 0.11.0rc0

following #5237

### DIFF
--- a/src/datasets/utils/hub.py
+++ b/src/datasets/utils/hub.py
@@ -6,7 +6,7 @@ from packaging import version
 
 
 def hf_hub_url(repo_id: str, path: str, revision: Optional[str] = None) -> str:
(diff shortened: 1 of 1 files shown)


src/datasets/utils/hub.py
```diff
@@ -6,7 +6,7 @@
 
 
 def hf_hub_url(repo_id: str, path: str, revision: Optional[str] = None) -> str:
-    if version.parse(hfh.__version__) < version.parse("0.11.0"):
+    if version.parse(hfh.__version__) < version.parse("0.11.0.rc0"):
```
albertvillanova, line 9: I'm not sure it makes sense to have in our code a reference to a release-candidate version...

PR #5248, opened 2022-11-16
Complete doc migration

mishig25:
Reverts huggingface/datasets#5214

Everything is handled on the doc-builder side now 😊


albertvillanova:
Thanks for the fix @mishig25.

I guess this is the reason why the docs are not generated for the latest release version 0.7.0? https://huggingface.co/docs/datasets/index

PR #5250, opened 2022-11-16
Change release procedure to use only pull requests

### LINKED ISSUE #5251: Docs are not generated after latest release
After the latest `datasets` release version 0.7.0, the docs were not generated.

As we have changed the release procedure (so that now we do not push directly to main branch), maybe we should also change the corresponding GitHub action:

https://github.com/huggingface/datasets/blob/edf1902f954c5568daadebcd8754bdad44b02a85/.github/workflows/build_documentation.yml#L3-L8

Related to:
- #5250

CC: @mishig25

albertvillanova:
This PR changes the release procedure so that:
- it only make changes to main branch via pull requests
- it is no longer necessary to directly commit/push to main branch

Close #5251.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ Steps to make a release:
      ```
      git checkout main
      git pull upstream main
-     git checkout -b release
+     git checkout -b vVERSION-release
      ```
 
 2. Change the version to the release VERSION in:
@@ -38,7 +38,7 @@ Steps to make a release:
      ```
      git add -u
      git commit -m "Release: VERSION"
-     git push upstream release
+     git push upstream vVERSION-release
      ```
    - Go to: https://github.com/huggingface/datasets/pull/new/release
    - Create pull request
--- a/setup.py
+++ b/setup.py
@@ -86,6 +86,7 @@ Steps to make a release:
        ```
        git checkout main
        git pull upstream main
+       git branch -D dev-version
        git checkout -b dev-version
        ```
    - Change the version to X.X.X+1.dev0 (e.g. VERSION=1.18.3 -> 1.18.4.dev0) in:
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ Steps to make a release:
      ```
      git checkout main
      git pull upstream main
-     git checkout -b vVERSION-release
+     git checkout -b release-VERSION
      ```
 
 2. Change the version to the release VERSION in:
@@ -38,7 +38,7 @@ Steps to make a release:
      ```
      git add -u
      git commit -m "Release: VERSION"
-     git push upstream vVERSION-release
+     git push upstream release-VERSION
      ```
    - Go to: https://github.com/huggingface/datasets/pull/new/release
    - Create pull request
--- a/setup.py
+++ b/setup.py
@@ -8,55 +8,98 @@ Note:
 
 Simple check list for release from AllenNLP repo: https://github.com/allenai/allennlp/blob/master/setup.py
 
-To create the package for pypi.
+Steps to make a release:
 
 0. Prerequisites:
    - Dependencies:
-     - twine: "pip install twine"
+     - twine: `pip install twine`
    - Create an account in (and join the 'datasets' project):
      - PyPI: https://pypi.org/
      - Test PyPI: https://test.pypi.org/
-   - Don't break `transformers`
-     - Run the `transformers` CI using the `main` branch and make sure it's green.
-       In `transformers`, use "datasets @ git+https://github.com/huggingface/datasets@main#egg=datasets"
-       in both setup.py and src/transformers/dependency_versions_table.py and then run the CI
+   - Don't break `transformers`: run the `transformers` CI using the `main` branch and make sure it's green.
+     - In `transformers`, use `datasets @ git+https://github.com/huggingface/datasets@main#egg=datasets`
+       in both:
+       - setup.py and
+       - src/transformers/dependency_versions_table.py
+     - and then run the CI
 
-1. Change the version in:
+1. Create the release branch from main branch:
+     ```
+     git checkout main
+     git pull upstream main
+     git checkout -b release
+     ```
+
+2. Change the version to the release VERSION in:
    - __init__.py
    - setup.py
 
-2. Commit these changes: "git commit -m 'Release: VERSION'"
+3. Commit these changes, push and create a Pull Request:
+     ```
+     git add -u
+     git commit -m "Release: VERSION"
+     git push upstream release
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -8,55 +8,98 @@
...
+       - setup.py and
+       - src/transformers/dependency_versions_table.py
+     - and then run the CI
 
-1. Change the version in:
+1. Create the release branch from main branch:
+     ```
+     git checkout main
+     git pull upstream main
+     git checkout -b release
```
lhoestq, line 30: (nit) in case a branch with this name already exists
```suggestion
     git checkout -b release-VERSION
```

setup.py
```diff
@@ -8,55 +8,98 @@
...
+     ```
+     twine upload dist/* -r pypi
+     ```
 
-7. Fill release notes in the tag in github once everything is looking hunky-dory.
+7. Make the release on GitHub once everything is looking hunky-dory:
+   - Merge the release Pull Request
+   - Create a new release: https://github.com/huggingface/datasets/releases/new
+   - Choose a tag: Introduce the new VERSION as tag, that will be created when you publish the release
+     - Create new tag VERSION on publish
```
lhoestq, line 78: I'm wondering if it will trigger the release-conda CI job - it's configured like
```yaml
on:
  push:
    tags:
      - "[0-9]+.[0-9]+.[0-9]+*"
```

setup.py
```diff
@@ -8,55 +8,98 @@
...
 
-2. Commit these changes: "git commit -m 'Release: VERSION'"
-
-3. Add a tag in git to mark the release: "git tag VERSION -m 'Add tag VERSION for pypi'"
-   Push the tag to remote: git push --tags origin main
+3. Commit these changes, push and create a Pull Request:
+     ```
+     git add -u
+     git commit -m "Release: VERSION"
+     git push upstream release
```
lhoestq, line 41:
```suggestion
     git push upstream release-VERSION
```

lhoestq (commented):
cool thanks ! a few comments:

setup.py
albertvillanova: I made it this way to make it simpler: as remote branches are deleted after the PR is merged.

setup.py
albertvillanova: I guess releated to:
- #5251

setup.py
albertvillanova: If a local branch with the same name exists, then previous command generated this error:
```
fatal: A branch named 'release' already exists.
```

lhoestq: oh ok !

albertvillanova: I have changed the release branch name so that it triggers docs generation: 88c9c83

Context:
- #5251

albertvillanova: @lhoestq after my arguments above, do you still think it is important the name of the branches to be unique for each release even if they are deleted after the PR is merged?

I make this question in relation with the other branch: `dev-version`

setup.py
albertvillanova: @lhoestq maybe we could align the triggering of release-conda to be the same as the one for generating the docs? That is: push to branch `v*-release`

setup.py
lhoestq: it's fine to not have it unique then

setup.py
albertvillanova: Indeed, conda-release was not triggered for 2.7.0 release...

setup.py
lhoestq: If we have it unique we won't need to delete the branch on a local clone of the repo though - the branch exists temporarily on upstream, but locally it's not deleted when the PR is merged

albertvillanova: The unique naming for the dev version would be more complicated though: something like `vX.X.X+1.dev0-dev-version`? 

Maybe it is easier to add a step before to delete the local branch `dev-version`?

albertvillanova:
Little recap:
- The release-conda GH action was properly triggered by push-tag event: therefore I guess this event is also created when we publish a release and create a tag within it (as it is the case in the new procedure)
  - However, the package was only uploaded to huggingface channel and not to conda-forge channel
    - [ ] Why? Need to address this.
- The generate-documentation GH action will be triggered by the push-to-branch event if we align the name of the release branch with the expected regex `v*-release`
  - [x] The naming has been aligned in the new procedure
  - [ ] Question: why do we have different triggering events for generate-doc and release-conda? Maybe we could set the same for both: either push-tag (when publishing the release), or push-to-branch
    - I think it will be better to use the push-tag event because in the new release procedure this is made later, when we have tested that everything works using the test-PyPI; on the contrary, the push-to-branch event happens before, even before opening the release PR: we could see afterwards that there is an issue, and cancel the Pull Request, but the docs and conda-package will already be published.
- For the naming of the dev-version branch/PR, instead of having a complicated version naming, I'm proposing:
  - Using always the same branch name `dev-version`
  - Just include a step to delete this branch locally if it exists: `git branch -D dev-version`

setup.py
lhoestq: We maintain the huggingface channel but not the conda forge one, which is maintained by the community

setup.py
lhoestq: Yea let's add a step to delete the local branch then

albertvillanova: OK. That step was already added: [cbd5874](https://github.com/huggingface/datasets/pull/5250/commits/cbd587492d97c6fa5cb1326133f989625292438b)

setup.py
albertvillanova: OK!

albertvillanova:
Just one question to be addressed: why do we have different triggering events for generate-doc and release-conda? Maybe we could set the same for both: either push-tag (when publishing the release), or push-to-branch

I think it will be better to use the push-tag event because in the new release procedure this happens later (when we publish the release), once we have already tested that everything works using the test-PyPI; on the contrary, the push-to-branch event happens before, even before opening the release PR: we could see afterwards that there is an issue, and cancel the Pull Request, but the docs and conda-package will already be published.

setup.py
albertvillanova: I have reverted: 8694901

lhoestq: Push tag for both sounds good :)

lhoestq (approved):
Looks all good to me ! Thanks for improving the release procedure :)

albertvillanova:
@lhoestq now that we have push-tag event for both build_documentation and release-conda, we have no constraint on the naming of the release branch:
- we could name it simpler: maybe as you suggested above: https://github.com/huggingface/datasets/pull/5250#discussion_r1024119018
  `release-VERSION`

Issue #3992, opened 2022-03-23
Image column is not decoded in map when using with with_transform


phihung:
## Describe the bug
Image column is not _decoded_ in **map** when using with `with_transform`

## Steps to reproduce the bug
```python
from datasets import Image, Dataset

def add_C(batch):
    batch["C"] = batch["A"]
    return batch

ds = Dataset.from_dict({"A": ["image.png"]}).cast_column("A",  Image())
ds = ds.with_transform(lambda x: x)  # <= This line causes the problem
ds = ds.map(add_C, batched=True)
print(ds[0])
```

## Expected results
```
{'C': <PIL.PngImagePlugin.PngImageFile>,  ...}
```

## Actual results
```
{'C': {'bytes': None, 'path': 'image.png'}, ...}
```

If we remove the `with_transform` line, we get the expected result.

## Environment info
- `datasets` version: 2.0.0
- Platform: Mac OSX
- Python version: 3.8.12
- PyArrow version: 7.0.0

mariosasko:
Hi! This behavior stems from this line: https://github.com/huggingface/datasets/blob/799b817d97590ddc97cbd38d07469403e030de8c/src/datasets/arrow_dataset.py#L1919
Basically, the `Image`/`Audio` columns are decoded only if the `format_type` attribute is `None` (`set_format`/`with_format` and `set_transform`/`with_transform` assign a non-`None` value to it) and the  `input_columns` param is not specified (see https://github.com/huggingface/datasets/issues/3756). We will remove these limitations soon.

Issue #3756, opened 2022-02-18
Images get decoded when using `map()` with `input_columns` argument on a dataset


kklemon:
## Describe the bug

The `datasets.features.Image` feature class decodes image data by default. Expectedly, when indexing a dataset or using the `map()` method, images are returned as PIL Image instances.

However, when calling `map()` and setting a specific data column with the `input_columns` argument, the image data is passed as raw byte representation to the mapping function.

## Steps to reproduce the bug
```python
from datasets import load_dataset
from torchvision import transforms
from PIL.Image import Image

dataset = load_dataset('mnist', split='train')

def transform_all_columns(example):
    # example['image'] is encoded as PIL Image
    assert isinstance(example['image'], Image)
    return example

def transform_image_column(image):
    # image is decoded here and represented as raw bytes
    assert isinstance(image, Image)
    return image

# single-sample dataset for debugging purposes
dev = dataset.select([0])

dev.map(transform_all_columns)
dev.map(transform_image_column, input_columns='image')
```

## Expected results

Image data should be passed in decoded form, i.e. as PIL Image objects to the mapping function unless the `decode` attribute on the image feature is set to `False`.

## Actual results

The mapping function receives images as raw byte data.

## Environment info
- `datasets` version: 1.18.3
- Platform: Linux-5.11.0-49-generic-x86_64-with-glibc2.32
- Python version: 3.8.0b4
- PyArrow version: 7.0.0

mariosasko:
Hi! If I'm not mistaken, this behavior is intentional, but I agree it could be more intuitive.

@albertvillanova Do you remember why you decided not to decode columns in the `Audio` feature PR when `input_columns` is not `None`? IMO we should decode those columns, and we don't even have to use lazy structures here because the user explicitly requires them in the map transform. 

cc @lhoestq for visibility

albertvillanova:
I think I excluded to decorate the function when `input_columns` were passed as a quick fix for some non-passing tests. That PR was quite complex and I decided to focus on the main feature requests, leaving refinements for subsequent PRs.

Note that when `input_columns` are passed, the signature of the function is effectively changed, while the decorated function expects an item (whether an example or a batch) as first arg (which is not the case when passing `input_columns`.

I agree we should consider supporting the case when `input_columns` are passed.

PR #5252, opened 2022-11-16
Support for decoding Image/Audio types in map when format type is not default one

mariosasko:
Add support for decoding the `Image`/`Audio` types in `map` for the formats (Numpy, TF, Jax, PyTorch) other than the default one (Python).

Additional improvements:
* make `Dataset`'s "iter" API  cleaner by removing `_iter` and replacing `_iter_batches` with `iter(batch_size)` (also implemented for `IterableDataset`)
* iterate over arrow tables in `map` to avoid `_getitem` calls, which are much slower than `__iter__`/`iter(batch_size)`, when the `format_type` is not Python
* fix `_iter_batches` (now named `iter`) when `drop_last_batch=True` and `pyarrow<=8.0.0` is installed
* lazily extract and decode arrow data in the default format 

TODO:
* [x] update the `iter` benchmark in the docs (the `BeamBuilder` cannot load the preprocessed datasets from our bucket, so wait for this to be fixed (cc @lhoestq))

Fix https://github.com/huggingface/datasets/issues/3992, fix https://github.com/huggingface/datasets/issues/3756


src/datasets/formatting/formatting.py
```diff
@@ -306,43 +339,58 @@ def format_batch(self, pa_table: pa.Table) -> pa.Table:
...
+    class LazyBatch(LazyDict):
+        def decode(self, value, feature, key):
+            return (
+                [decode_nested_example(feature, v) if value is not None else None for v in value]
+                if self.formatter.features._column_requires_decoding[key]
+                else value
+            )
+
+    lazy_row_type = LazyExample
+    lazy_batch_type = LazyBatch
```
mariosasko, line 360: I'm all ears if you have any ideas on how to make this part cleaner.

src/datasets/arrow_dataset.py
```diff
@@ -2818,6 +2759,52 @@ def _map_single(
...
+
+            if not batched:
+                input_formatter.lazy_row_type = DictWithDecodableDataToRestore
+            else:
+                input_formatter.lazy_batch_type = DictWithDecodableDataToRestore
+
+        else:
+
+            class DictWithDecodableDataToRestore:
+                pass
```
mariosasko, line 2802: And this one

mariosasko (commented):
Some comments for reviewers:

docs/source/about_arrow.md
```diff
@@ -39,8 +39,8 @@ Iterating over a memory-mapped dataset using Arrow is fast. Iterating over Wikip
 
 ```python
 >>> s = """batch_size = 1000
-... for i in range(0, len(wiki), batch_size):
-...     batch = wiki[i:i + batch_size]
+... for batch in wiki.iter(batch_size):
+...     ...
```
lhoestq, line 43: Love this !

src/datasets/formatting/formatting.py
lhoestq: I think that's fine. Maybe put their definition outside of PythonFormatter for pickling reasons

And maybe rename the method "format" instead of "decode", since it not just decoding but also type formatting (i.e. torch/tf formatting)

src/datasets/formatting/formatting.py
```diff
@@ -260,6 +262,34 @@ def decode_batch(self, batch: pd.DataFrame) -> pd.DataFrame:
         return self.decode_row(batch)
 
 
+class LazyDict(UserDict):
+    def __init__(self, data: dict, formatter: "Formatter"):
+        self.data = data
```
lhoestq, line 267: Here data is the "extracted" data - e.g. a python dictionary from an arrow table.

We could also imagine storing the arrow data instead, and lazily apply the full formatting (extraction + decoding + type formatting)

Maybe it's not the scope of this PR, but it would make all the map/filter functions ultra fast if you only need to read one column.

Right now you have to handle arrow data, extracted data, decoded data and formatted data.
Maybe we only need to handle arrow data and formatted data if it makes things simpler

src/datasets/formatting/__init__.py
```diff
@@ -70,32 +77,32 @@ def _register_unavailable_formatter(
 
 
 # Here we define all the available formatting functions that can be used by `Dataset.set_format`
-_register_formatter(PythonFormatter, None, aliases=["python"])
+_register_formatter(PythonFormatter, None, aliases=["python"], supports_lazy_formatting=True)
```
lhoestq, line 80: Maybe supports_lazy_formatting can be a class attribute instead ? Since this is implementation - specific

src/datasets/arrow_dataset.py
lhoestq: I'm not sure I understand what's going on here tbh :/

lhoestq (commented):
Thanks ! I left a few comments

Sorry I didn't really understand the logic for undecoded data to restore - can you provide more details ?

Overall I feel like we're reaching a point where things are getting too complicated, I added a message about lazy decoding from arrow in case it can help, though I'm not sure what's the best solution here

src/datasets/arrow_dataset.py
mariosasko: I think it's easiest to understand this with an example:
```python
dset = dset.with_format("torch") # torch-formatted dataset

def map_fn(ex):
     return {"shape": tuple(ex["image"].shape)}

dset = dset.map(map_fn)
```
The above example would save the "image" column as `torch.Tensor` (due to being accessed/decoded in `map_fn`) without this "restore" logic, which is not ideal because the image -> tensor -> image conversion is not always lossless. So what we do instead (using this special dict) is we store the "raw" (undecoded) values of decodable columns (only if they are decoded) to be able to restore them later (in the above case, we do this for the "image" column) unless these columns are overwritten with `__setitem__`  (e.g. `ex["image"] = "some val"`) in the transform.

Let me know if it's clear now.

src/datasets/formatting/formatting.py
mariosasko: Had the same idea. I think I'll explore that as part of this PR.

src/datasets/arrow_dataset.py
lhoestq: I see ! That's a super nice optimization trick :)

One nit:
what if the tensor is modified in place ? e.g.
```python
def map_fn(ex):
     ex["image"] += 1
     return ex
```
in that case the image data won't be updated.
IMO it's a case that it rare enough to not have to focus too much on it though - but documenting it would be appreciated.

mariosasko: In this case, the correct behavior could be preserved by disabling lazy decoding for Numpy, TF, Jax and PyTorch. Moreover, the additional rationale is that if one sets the format to one of the above types before the map call, then it's OK to format all the `format_columns` eagerly (even if some of them are decodable) because we can assume these columns will be accessed in the map transform anyways. WDYT?

The options such as subclassing the tensor type or wrapping the tensor with an object that tracks method calls on it seem too error-prone.

src/datasets/arrow_dataset.py
```diff
@@ -2818,6 +2760,40 @@ def _map_single(
...
+
+                def __setitem__(self, key, value):
+                    if key in self.decodable_data_to_restore:
+                        del self.decodable_data_to_restore[key]
+                    super().__setitem__(key, value)
+
+                def __delitem__(self, key):
+                    if key in self.decodable_data_to_restore:
+                        del self.decodable_data_to_restore[key]
+                    super().__delitem__(key)
```
lhoestq, line 2790: (nit) Maybe this can be integrated in LazyDict Directly ? This way we don't need an extra class.

Consider this example

```python
t = pa.table({"image": ...})
batch = LazyBatch(t, formatter=NumpyFormatter())
batch["shape"] = [img.shape for img in batch["image"]]
writer.write_table(batch.to_arrow())
```

Under the hood `batch.to_arrow()` would reuse the arrow data from `t` for "image", and make new arrow data for "shape". This way we avoid one conversion from python to arrow - the same way DictWithDecodableDataToRestore works right now.

Let me know if it makes sense

lhoestq (commented):
I find this amazing ! This will optimize the speed of iterating significantly, since it only loads the data you need.

My first benchmark:
```python
>>> t = pa.table({"text": [["a"*10]*10]*10_000, "id": range(10_000)})                                                                                                                              
>>> %time len(list(t.slice(i, bs).to_pydict()["id"] for i in range(0, len(t), bs)))                                                                                                                
CPU times: user 129 ms, sys: 1.69 ms, total: 131 ms
Wall time: 132 ms
100
>>> %time len(list(PythonLazyBatch(t.slice(i, bs), formatter=formatter)["id"] for i in range(0, len(t), bs)))                                                                                      
CPU times: user 6.22 ms, sys: 459 µs, total: 6.67 ms
Wall time: 6.31 ms
100
```

(and I haven't even used the fast iterator you added recently)

Let me know if I can help :)

src/datasets/arrow_dataset.py
lhoestq: > The options such as subclassing the tensor type or wrapping the tensor with an object that tracks method calls on it seem too error-prone.

I agree - tbh I think we can ignore this for now and focus on this PR. As long as we document it it's fine

src/datasets/arrow_dataset.py
mariosasko: This API has been removed, as we no longer need it (and it's somewhat hard to understand)

src/datasets/formatting/__init__.py
mariosasko: Only the default formatter (Python) can be lazily decoded now, so I removed this `supports_lazy_formatting` API and added `lazy` as a param to the `PythonFormatter`. If we implement lazy decoding for more formatters in the future, we can come up with a more generic approach.

src/datasets/table.py
```diff
@@ -1798,6 +1799,79 @@ def wrapper(array, *args, **kwargs):
...
+        concatenated_offsets = offset
+        for offset in offsets[1:]:
+            offset = pc.subtract(offset, offset[0])
+            offset = pc.add(offset[1:], concatenated_offsets[-1])
+            concatenated_offsets = pa.concat_arrays([concatenated_offsets, offset])
+        return concatenated_offsets
+
+    # breakpoint_ = arrays = [pa.array([{"a": 21, "b": [[1, 2], [3]]}]), pa.array([{"a": 100, "b": [[1], None]}])]
+
+    def _concat_arrays(arrays):
```
lhoestq, line 1838: Isn't it possible to use `pa.concat_arrays` directly here ? It can be used for regular arrays, and also for the storage of extension arrays

src/datasets/formatting/formatting.py
```diff
@@ -304,6 +303,7 @@ def __repr__(self):
         return repr(self.data)
 
     if config.PY_VERSION >= version.parse("3.9"):
+        # merging with the union ("|") operator is supported in Python 3.9+
```
lhoestq, line 306: You don't use it as far as I can tell

src/datasets/table.py
```diff
@@ -1798,6 +1799,79 @@ def wrapper(array, *args, **kwargs):
...
+    def _offsets_concat(offsets):
+        offset = offsets[0]
+        concatenated_offsets = offset
+        for offset in offsets[1:]:
+            offset = pc.subtract(offset, offset[0])
+            offset = pc.add(offset[1:], concatenated_offsets[-1])
+            concatenated_offsets = pa.concat_arrays([concatenated_offsets, offset])
+        return concatenated_offsets
+
+    # breakpoint_ = arrays = [pa.array([{"a": 21, "b": [[1, 2], [3]]}]), pa.array([{"a": 100, "b": [[1], None]}])]
```
lhoestq, line 1836:
```suggestion
```

src/datasets/table.py
```diff
@@ -1798,6 +1799,79 @@ def wrapper(array, *args, **kwargs):
...
+
+    def _concat_arrays(arrays):
+        # if breakpoint_:
+        # breakpoint()
```
lhoestq, line 1840:
```suggestion
```

lhoestq (commented):
Thanks ! Just one question on concat_arrays:

src/datasets/formatting/formatting.py
mariosasko: This class needs to mimic the built-in dict inside the map function, so it needs to support all of its methods.

src/datasets/table.py
mariosasko: I don't think it is. PyArrow extension arrays can be nested inside a list/struct, so we need to call `_concat_arrays` (recursively) to extract their storage arrays before we concatenate them (`pa.concat_arrays` would fail in this scenario).

Another option would be to do the following:
1. run a procedure that extracts the storage arrays
2. concatenate the arrays
3. re-cast to the original type

But this seems more expensive than the current solution.

lhoestq: Ok I see :) sounds good then

lhoestq (commented):
Could you add a test to make sure an image is not decoded when filtering on the labels ?

If I'm not mistaken right now it may decode the image because of this line in `get_indices_from_mask_function`:
```python
num_examples = len(batch[next(iter(batch.keys()))])
```

mariosasko:
Yes, if the image column is the first in the batch, it will decode the images because it reads the actual values. We could avoid this by checking the batch type, and if it's `LazyDict`, `num_examples` is equal to `len(batch.pa_table)`, which doesn't lead to decoding.

lhoestq:
Good idea. This can be done in a subsequent PR btw, since it's out of scope of the original goal of this PR

src/datasets/arrow_writer.py
```diff
@@ -433,7 +433,13 @@ def write_examples_on_file(self):
         batch_examples = {}
         for col in cols:
             # Since current_examples contains (example, key) tuples
-            batch_examples[col] = [row[0][col] for row in self.current_examples]
+            if isinstance(self.current_examples[0][0][col], (pa.Array, pa.ChunkedArray)):
```
lhoestq, line 436:
```suggestion
            if isinstance(self.current_examples[0][0][col], (pa.Array, pa.ChunkedArray)):
                # The examples could be Arrow arrays of 1 element.
                # This can happen in `.map()` when we want to re-write the same Arrow data
```

src/datasets/arrow_dataset.py
```diff
@@ -2863,32 +2822,29 @@ def apply_function_on_filtered_inputs(inputs, indices, check_same_num_examples=F
...
             if check_same_num_examples:
-                input_num_examples = len(inputs[next(iter(inputs.keys()))])
+                input_num_examples = len(pa_inputs)
                 processed_inputs_num_examples = len(processed_inputs[next(iter(processed_inputs.keys()))])
                 if input_num_examples != processed_inputs_num_examples:
                     raise NumExamplesMismatchError()
-            if isinstance(inputs, dict) and isinstance(processed_inputs, Mapping):
-                return {**inputs, **processed_inputs}
+            if isinstance(inputs, Mapping) and isinstance(processed_inputs, Mapping):
+                return {**pa_inputs_dict, **processed_inputs}
```
lhoestq, line 2847:
```suggestion
                # The .map() transform *updates* the dataset:
                # the output dictionary contains both the the input data and the output data.
                # The output dictionary may contain Arrow values from `pa_inputs_dict` so that we can re-write them efficiently.
                return {**pa_inputs_dict, **processed_inputs}
```

src/datasets/formatting/formatting.py
```diff
@@ -260,6 +263,126 @@ def decode_batch(self, batch: pd.DataFrame) -> pd.DataFrame:
         return self.decode_row(batch)
 
 
+class LazyDict(MutableMapping):
```
lhoestq, line 266:
```suggestion
class LazyDict(MutableMapping):
    """A dictionary backed by Arrow data. The values are formatted on-the-fly when accessing the dictionary."""

```

tests/test_table.py
```diff
@@ -1055,6 +1056,43 @@ def test_indexed_table_mixin():
...
 
 
+@pytest.mark.parametrize(
+    "arrays",
+    [
+        [pa.array([[1, 2, 3, 4]]), pa.array([[10, 2]])],
+        [
+            pa.array([[[1, 2], [3]]], pa.list_(pa.list_(pa.int32()), 2)),
+            pa.array([[[10, 2, 3], [2]]], pa.list_(pa.list_(pa.int32()), 2)),
+        ],
```
lhoestq, line 1066: Can you add some tests with sliced arrays ? They may have offsets that don't start at 0 IIRC.
And maybe arrays with None as well.

src/datasets/table.py
```diff
@@ -1744,6 +1799,75 @@ def wrapper(array, *args, **kwargs):
...
+            return pa.StructArray.from_arrays(
+                [_concat_arrays([array.field(field.name) for array in arrays]) for field in array.type],
+                fields=list(array.type),
+                mask=pa.concat_arrays([array.is_null() for array in arrays]),
+            )
+        elif pa.types.is_list(array.type):
+            if config.PYARROW_VERSION.major < 10:
+                warnings.warn(
+                    "None values are converted to empty lists in `pyarrow<10.0.0` when concatenating list arrays with None values. Install `pyarrow>=10.0.0` to avoid this behavior. More info: https://github.com/huggingface/datasets/issues/3676."
+                )
```
lhoestq, line 1850: We may have a regression here: all the users with pyarrow<10 will experience their nulls converted to empty lists in ListArray when they run `.map()`.

Maybe we should use pa.concat_arrays when there's no extension arrays in a ListArray ?

lhoestq (commented):
Thanks ! Added some comments suggestions, and I think I found a regression:
```python
>>> ds = Dataset.from_dict({"a": [[1], None]})
>>> ds.map(lambda x: x).to_dict()
/Users/quentinlhoest/Desktop/hf/datasets/src/datasets/table.py:1849: UserWarning: None values are converted to empty lists in `pyarrow<10.0.0` when concatenating list arrays with None values. Install `pyarrow>=10.0.0` to avoid this behavior. More info: https://github.com/huggingface/datasets/issues/3676.
  "None values are converted to empty lists in `pyarrow<10.0.0` when concatenating list arrays with None values. Install `pyarrow>=10.0.0` to avoid this behavior. More info: https://github.com/huggingface/datasets/issues/3676."
{'a': [[1], []]}
```
while it should return `{'a': [[1], None]}`. I added a comment on how we could fix this regression.

tests/test_table.py
mariosasko: Done! 

> And maybe arrays with None as well.

We already have a test for this case

lhoestq:
Just fixed a small bug where it would show the pyarrow 10 warning about None -> empty lists conversions even with an Array2D with no nulls

lhoestq:
Fixed another bug when your map function returns a mix of LazyDict or regular dict

lhoestq (approved):
It looks all good to me now, feel free to merge if you're ok with my latest changes :)

src/datasets/arrow_writer.py
```diff
@@ -432,8 +432,17 @@ def write_examples_on_file(self):
...
+            # Morever, examples could be Arrow arrays of 1 element.
+            # This can happen in `.map()` when we want to re-write the same Arrow data
+            if all(isinstance(row[0][col], (pa.Array, pa.ChunkedArray)) for row in self.current_examples):
+                arrays = [row[0][col] for row in self.current_examples]
+                batch_examples[col] = array_concat(arrays)
+            else:
+                batch_examples[col] = [
+                    row[0][col].to_pylist()[0] if isinstance(row[0][col], (pa.Array, pa.ChunkedArray)) else row[0][col]
+                    for row in self.current_examples
+                ]
```
mariosasko, line 445: Nit: We could avoid the exhaustive `isinstance` check with `try/except`, but I don't think the peformance gain significant enough: 

```suggestion
            if isinstance(self.current_examples[0][0][col], (pa.Array, pa.ChunkedArray)):
                try:
                    batch_examples[col] = array_concat([row[0][col] for row in self.current_examples])
                except Exception:
                    batch_examples[col] = [
                        row[0][col].to_pylist()[0] if isinstance(row[0][col], (pa.Array, pa.ChunkedArray)) else row[0][col]
                        for row in self.current_examples
                ]
                    
            else:
                batch_examples[col] = [
                    row[0][col].to_pylist()[0] if isinstance(row[0][col], (pa.Array, pa.ChunkedArray)) else row[0][col]
                    for row in self.current_examples
                ]
```

lhoestq: ok !

mariosasko: `except Exception` would make debugging harder in the scenarios when a relevant error is raised inside `array_concat`, so I think we can leave this code as-is

PR #5257, opened 2022-11-17
remove an unused statement

WrRan:
remove the unused statement: `input_pairs = list(zip())`


lhoestq (approved):
indeed that doesn't seem useful :p thanks !

PR #5266, opened 2022-11-18
Specify arguments in librosa.reshape to avoid future errors 

polinaeterna:
Fixes a warning and future deprecation from `librosa.reshape`:
```
FutureWarning: Pass orig_sr=16000, target_sr=48000 as keyword args. From version 0.10 passing these as positional arguments will result in an error

   array = librosa.resample(array, sampling_rate, self.sampling_rate, res_type="kaiser_best")
```


lhoestq (approved):
Thanks !

Issue #4351, opened 2022-05-14
Add optional progress bar for .save_to_disk(..) and .load_from_disk(..) when working with remote filesystems


Rexhaif:
**Is your feature request related to a problem? Please describe.**
When working with large datasets stored on remote filesystems(such as s3), the process of uploading a dataset could take really long time. For instance: I was uploading a re-processed version of wmt17 en-ru to my s3 bucket and it took like 35 minutes(and that's given that I have a fiber optic connection). The only output during that process was a progress bar for flattening indices and then ~35 minutes of complete silence.

**Describe the solution you'd like**
I want to be able to enable a progress bar when calling .save_to_disk(..) and .load_from_disk(..), it would track either amount of bytes sent/received or amount of records written/loaded, and will give some ETA. Basically just tqdm. 

**Describe alternatives you've considered**
- Save dataset to tmp folder at the disk and then upload it using custom wrapper over botocore, which will work with progress bar, like [this](https://alexwlchan.net/2021/04/s3-progress-bars/).

Issue #4196, opened 2022-04-21
Embed image and audio files in `save_to_disk`


lhoestq:
Following https://github.com/huggingface/datasets/pull/4184, currently a dataset saved using `save_to_disk` doesn't actually contain the bytes of the image or audio files. Instead it stores the path to your local files. 

Adding `embed_external_files` and set it to True by default to save_to_disk would be kind of a breaking change since some users will get bigger Arrow files when updating the lib, but the advantages are nice:

- the resulting dataset is self contained, in case you want to delete your cache for example or share it with someone else
- users also upload these Arrow files to cloud storage via the fs parameter, and in this case they would expect to upload a self-contained dataset
- consistency with push_to_hub

This can be implemented at the same time as sharding for `save_to_disk` for efficiency, and reuse the helpers from `push_to_hub` to embed the external files.

cc @mariosasko

PR #5268, opened 2022-11-18
Sharded save_to_disk + multiprocessing

### LINKED ISSUE #5263: Save a dataset in a determined number of shards
This is useful to distribute the shards to training nodes.

This can be implemented in `save_to_disk` and can also leverage multiprocessing to speed up the process

lhoestq:
Added `num_shards=` and `num_proc=` to `save_to_disk()`

EDIT: also added `max_shard_size=` to `save_to_disk()`, and also `num_shards=` to `push_to_hub`

I also:
- deprecated the fs parameter in favor of storage_options (for consistency with the rest of the lib) in save_to_disk and load_from_disk
- always embed the image/audio data in arrow when doing `save_to_disk`
- added a tqdm bar in `save_to_disk`
- Use the MockFileSystem in tests for `save_to_disk` and `load_from_disk`
- removed the unused integration tests with S3, since we can now test with `mockfs` instead of `s3fs`

TODO:
- [x] implem save_to_disk for dataset dict
- [x] save_to_disk for dataset dict tests
- [x] deprecate fs in dataset dict load_from_disk as well
- [x] update docs

Close #5263 
Close https://github.com/huggingface/datasets/issues/4196
Close https://github.com/huggingface/datasets/issues/4351


src/datasets/arrow_dataset.py
```diff
@@ -1246,43 +1258,118 @@ def save_to_disk(self, dataset_path: str, fs=None):
...
+                        num_examples_progress_update = 0
+            else:
+                for i in range(0, shard.num_rows, batch_size):
+                    pa_table = shard.data.slice(i, batch_size)
+                    writer.write_table(pa_table)
+                    num_examples_progress_update += len(pa_table)
+                    if time.time() > _time + refresh_rate:
+                        _time = time.time()
+                        yield job_id, False, num_examples_progress_update
+                        num_examples_progress_update = 0
```
lhoestq, line 1366: I iterate on batches here to update the tqdm bar, but for old versions of pyarrow this may be too slow since `table_iter` only works for pyarrow>=8.

I think we may have to implement `table_iter` even on old versions for performance reasons. It can be based on `pa.Table.to_record_batches` - lmk what you think

lhoestq: Ok I just implemented `pa.Table.to_reader` for pyarrow < 8 for our `datasets.table.Table`. This way we don't have to check the pyarrow version anymore

src/datasets/arrow_dataset.py
```diff
@@ -1179,63 +1181,73 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         # Here `del` is used to del the pyarrow tables. This properly closes the files used for memory mapped tables
         self.__del__()
 
-    def save_to_disk(self, dataset_path: str, fs=None):
+    def save_to_disk(
+        self,
+        dataset_path: PathLike,
+        fs="deprecated",
+        num_shards: Optional[int] = None,
```
mariosasko, line 1188: If we want to be consistent with the other methods (`push_to_hub`, `download_and_prepare`), this param should be renamed to `max_shard_size`.

src/datasets/arrow_dataset.py
```diff
@@ -1179,63 +1181,73 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
-        if is_remote_filesystem(fs):
-            dataset_path = extract_path_from_uri(dataset_path)
-        else:
-            fs = fsspec.filesystem("file")
-            cache_files_paths = [Path(cache_filename["filename"]) for cache_filename in self.cache_files]
+        if is_local:
+            Path(dataset_path).resolve().mkdir(parents=True, exist_ok=True)
+            parent_cache_files_paths = set(
+                Path(cache_filename["filename"]).resolve().parent for cache_filename in self.cache_files
+            )
```
mariosasko, line 1246: I'm okay with checking the parent dir instead of the actual cache filename, but why is this change needed?

mariosasko (commented):
Nice job!

src/datasets/arrow_dataset.py
lhoestq: Nope, this parameter is a number of shard, not the maximum size of every shard :)

For training you often need your dataset to be sharded in a number of files that is a factor of the number of nodes you have - so you need to be able to specify the exact number of shards

src/datasets/arrow_dataset.py
lhoestq: It's just simpler to implelement this verification this way than iterate over the file names x)
And it also prevents the case where you'd have the dataset saved twice in the same folder, but with different number of shards e.g.
```
data-00000-of-00008.parquet
data-00001-of-00008.parquet
...
data-00007-of-00008.parquet
```
and

```
data-00000-of-00004.parquet
data-00001-of-00004.parquet
...
data-00003-of-00004.parquet
```

docs/source/filesystems.mdx
```diff
@@ -19,10 +19,10 @@ Here are examples for S3, Google Cloud Storage and Azure Blob Storage.
 
 ### Amazon S3
 
-1. Install the S3 dependency with 🤗 Datasets:
+1. Install the S3 FileSystem implementation:
```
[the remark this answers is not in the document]
lhoestq: Yea let's do this in another PR

src/datasets/arrow_dataset.py
mariosasko: > Nope, this parameter is a number of shard, not the maximum size of every shard :)

What I meant is renaming + updating the logic.

> For training you often need your dataset to be sharded in a number of files that is a factor of the number of nodes you have - so you need to be able to specify the exact number of shards

For this particular case, one could use `shard` + `save_to_disk`, no? When it comes to the API design, I'm not sure doing this at once in `save_to_disk` is a good idea

lhoestq: > For this particular case, one could use shard + save_to_disk, no?

`save_to_disk` requires one directory per dataset, as a sort of bundle. The idea is to end up with one directory containing all the shards, named automatically with the shard idx. Moreover both saving and reloading take one line: it saves some boilerplate code

> When it comes to the API design, I'm not sure doing this at once in save_to_disk is a good idea

What should the user journey look like in that case ? In Dask you can repartition a dataset and then call to_parquet. In Apache beam you can pass num_shards to the WriteTo* functions.

mariosasko: As I said, my main concern here is consistency with the rest of the API, i.e., I would rather have a bit of boilerplate (like 3-4 lines of code in this particular case) and consistent API than no boilerplate but inconsistent API (or more ways to do the same thing).

> What should the user journey look like in that case ? In Dask you can repartition a dataset and then call to_parquet. In Apache beam you can pass num_shards to the WriteTo* functions.

We can do several things:
* we should clarify how to use Datasets for distributed training in the docs (shard, ...).
* considering `save_to_disk` supports writing to remote filesystems, we should allow the same in the `to_<format>` methods (e.g. to enable writing Parquet files to GCS)
* advanced usage: support `fsspec` paths as `cache_file_name` to allow writing cached (arrow) files directly to remote filesystems (e.g. `ds.shard(..., cache_file_name="gs://")`) 

And in this PR,  instead of `num_shards`, we could add `max_shard_size` to `save_to_disk` to have the ability to reduce the size of the written Arrow files and leverage multiprocessing (IIRC on some platforms loading large files can be problematic)

lhoestq: > As I said, my main concern here is consistency with the rest of the API, i.e., I would rather have a bit of boilerplate (like 3-4 lines of code in this particular case) and consistent API than no boilerplate but inconsistent API (or more ways to do the same thing).

Ok I see, but in this case the boiler plate code is not great IMO. Either it uses `save_to_disk` and we end up with tons of directories, or we use `.map(, cache_file_name=...)` and it feels weird. Maybe we can just add `to_arrow` ?

> we should clarify how to use Datasets for distributed training in the docs (shard, ...).

+ 1 :)

> considering save_to_disk supports writing to remote filesystems, we should allow the same in the to_<format> methods (e.g. to enable writing Parquet files to GCS)

+1 as well !

> advanced usage: support fsspec paths as cache_file_name to allow writing cached (arrow) files directly to remote filesystems (e.g. ds.shard(..., cache_file_name="gs://"))

not 100% sure, since `.shard` creates a new dataset - it doesn't write the dataset on disk. It may write an indices mapping but that's it

mariosasko: > Ok I see, but in this case the boiler plate code is not great IMO. Either it uses save_to_disk and we end up with tons of directories, or we use .map(, cache_file_name=...) and it feels weird. Maybe we can just add to_arrow ?

`map` is needed to embed audio/image bytes, right? I think we should release version 3.0 soon (at the beginning of next year?) and start embedding them as we generate a dataset. And then the boilerplate would become much simpler.

> not 100% sure, since .shard creates a new dataset - it doesn't write the dataset on disk. It may write an indices mapping but that's it

Oh, you are right! We could either raise an error in this case or make sure a new dataset is generated (and only do optimizations if `cache_file_name` is written locally).

lhoestq: > map is needed to embed audio/image bytes, right?

yea it keeps the links to files by default

> and start embedding them as we generate a dataset

that'b be nice, and we can check with audio folks to make sure it doesn't break stuff ^^

So tin the end o save a dataset as sharded Arrow file here are some options:

1. this PR: add `num_shards` to `save_to_disk`
```python
ds.save_to_disk("output_dir", num_shards=128)
# later
ds = load_from_disk("output_dir")
```
2. add `to_arrow`
```python
num_shards = 128
os.makedirs("output_dir")
for shard_idx in tqdm(range(num_shards), unit="shard"):
    ds.shard(
        num_shards=num_shards,
        index=shard_idx,
        contiguous=True
    ).to_arrow(f"output_dir/data_{shard_idx}.arrow")
# later
ds = concatenate_datasets([
    Dataset.from_file(f"output_dir/data_{shard_idx}.arrow"")
    for shard_idx in range(num_shards)
])
```
3. use map
```python
num_shards = 128
os.makedirs("output_dir")
for shard_idx in tqdm(range(num_shards), unit="shard"):
    ds.shard(
        num_shards=num_shards,
        index=shard_idx,
        contiguous=True
    ).map(cache_file_name=f"output_dir/data_{shard_idx}.arrow")
# later
ds = concatenate_datasets([
    Dataset.from_file(f"output_dir/data_{shard_idx}.arrow"")
    for shard_idx in range(num_shards)
])
```

My concerns so far:
1. adds a new parameter `num_shards` that isn't consistent with `push_to_hub` which has `max_shard_sizes`. We could have both `max_shard_sizes` and `num_shards` in `save_to_disk` and `push_to_hub` IMO - they correspond to different use cases. Apache beam also has both options. Dask as well, but named differently.
2. we need to add a new method + maybe the names `to_arrow`/`from_file` are not fully aligned
3. `map` doesn't embed files. And even if you specify a`cache_file_name`, it may end up writing multiple files using a suffix in multiprocessing. In the future I was thinking of sharding `map` by default to avoid large files.
2 & 3. would have been nice to have `contiguous=True` in `.shard()` by default IMO - not sure if we can fix that

Which one do you prefer ? Do you have other ideas ? Personally I think than 1 and 2 are fine (and small pref for 1 ^^)

mariosasko: I like the idea of having both `num_shards` and `max_shard_size` in `push_to_hub`/`save_to_disk` to ensure consistency, so I vote for option 1.

lhoestq:
Added both num_shards and max_shard_size in push_to_hub/save_to_disk. Will take care of updating the tests later

src/datasets/arrow_dataset.py
```diff
@@ -1246,43 +1258,108 @@ def save_to_disk(self, dataset_path: str, fs=None):
...
+        fpath: str = arg["fpath"]
+        storage_options: Optional[dict] = arg["storage_options"]
+        refresh_rate = 0.05  # 20 progress updates per sec
+        batch_size = config.DEFAULT_MAX_BATCH_SIZE
+
+        if shard._indices is not None:
+            raise ValueError(
+                "`_save_to_disk_single` only support shards with flattened indices. "
+                "Please call ds.flatten_indices() before saving to disk."
+            )
```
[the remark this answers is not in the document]
lhoestq: `save_to_disk` calls `flatten_indices()` before calling `_save_to_disk_single`. So in practice this error never happens.

I just added it in case someone plays with the internals of the lib

lhoestq:
It's ready for a final review @mariosasko and @albertvillanova, let me know what you think :)

mariosasko (commented):
Some nits.

lhoestq:
Took your comments into account, and also changed `iflatmap_unordered` to take an iterable of kwargs to make the code more redable :)

mariosasko (approved):
Thanks, LGTM!

Issue #5275, opened 2022-11-22
YAML integer keys are not preserved Hub server-side


albertvillanova:
After an internal discussion (https://github.com/huggingface/moon-landing/issues/4563):
- YAML integer keys are not preserved server-side: they are transformed to strings
  - See for example this Hub PR: https://huggingface.co/datasets/acronym_identification/discussions/1/files
    - Original:
      ```yaml
      class_label:
        names:
          0: B-long
          1: B-short
      ```
    - Returned by the server:
      ```yaml
      class_label:
        names:
          '0': B-long
          '1': B-short
      ```
- They are planning to enforce only string keys
- Other projects already use interger-transformed-to string keys: e.g. `transformers` models `id2label`: https://huggingface.co/roberta-large-mnli/blob/main/config.json
  ```yaml
  "id2label": {
    "0": "CONTRADICTION",
    "1": "NEUTRAL",
    "2": "ENTAILMENT"
  }
  ``` 

On the other hand, at `datasets` we are currently using YAML integer keys for `dataset_info` `class_label`.

Please note (thanks @lhoestq for pointing out) that previous versions (2.6 and 2.7) of `datasets` need being patched:
```python
In [18]: Features._from_yaml_list([{'dtype': {'class_label': {'names': {'0': 'neg', '1': 'pos'}}}, 'name': 'label'}])                                                                                      
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-18-974f07eea526> in <module>
----> 1 Features._from_yaml_list(ry)

~/Desktop/hf/nlp/src/datasets/features/features.py in _from_yaml_list(cls, yaml_data)
   1743                 raise TypeError(f"Expected a dict or a list but got {type(obj)}: {obj}")
   1744 
-> 1745         return cls.from_dict(from_yaml_inner(yaml_data))
   1746 
   1747     def encode_example(self, example):

~/Desktop/hf/nlp/src/datasets/features/features.py in from_yaml_inner(obj)
   1739             elif isinstance(obj, list):
   1740                 names = [_feature.pop("name") for _feature in obj]
-> 1741                 return {name: from_yaml_inner(_feature) for name, _feature in zip(names, obj)}
   1742             else:
   1743                 raise TypeError(f"Expected a dict or a list but got {type(obj)}: {obj}")

~/Desktop/hf/nlp/src/datasets/features/features.py in <dictcomp>(.0)
   1739             elif isinstance(obj, list):
[...]

PR #5277, opened 2022-11-22
Remove YAML integer keys from class_label metadata

albertvillanova:
Fix partially #5275.

### DIFF
--- a/src/datasets/features/features.py
+++ b/src/datasets/features/features.py
@@ -1645,11 +1645,14 @@ class Features(dict):
             #
             # class_label:              ->              class_label:
             #   names:                  ->                names:
-            #   - negative              ->                  0: negative
-            #   - positive              ->                  1: positive
+            #   - negative              ->                  '0': negative
+            #   - positive              ->                  '1': positive
             #
             if isinstance(feature.get("class_label"), dict) and isinstance(feature["class_label"].get("names"), list):
-                feature["class_label"]["names"] = dict(enumerate(feature["class_label"]["names"]))
+                # server-side requirement: keys must be strings
+                feature["class_label"]["names"] = {
+                    str(label_id): label_name for label_id, label_name in enumerate(feature["class_label"]["names"])
+                }
             return feature
 
         def to_yaml_inner(obj: Union[dict, list]) -> dict:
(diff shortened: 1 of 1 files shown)


src/datasets/features/features.py
```diff
@@ -1698,14 +1701,14 @@ def unsimplify(feature: dict) -> dict:
             #
             # class_label:              ->              class_label:
             #   names:                  ->                names:
-            #     0: negative              ->               - negative
-            #     1: positive              ->               - positive
+            #     '0': negative              ->               - negative
+            #     '1': positive              ->               - positive
             #
             if isinstance(feature.get("class_label"), dict) and isinstance(feature["class_label"].get("names"), dict):
                 label_ids = sorted(feature["class_label"]["names"])
```
lhoestq, line 1708: because "2" > "10" in python
```suggestion
                label_ids = sorted([int(label_id) for label_id in feature["class_label"]["names"]])
```

src/datasets/features/features.py
```diff
@@ -1698,14 +1701,14 @@ def unsimplify(feature: dict) -> dict:
...
-                if label_ids and label_ids != list(range(label_ids[-1] + 1)):
+                if label_ids and [int(label_id) for label_id in label_ids] != list(range(int(label_ids[-1]) + 1)):
```
lhoestq, line 1709:
```suggestion
                if label_ids and label_ids != list(range(int(label_ids[-1]) + 1)):
```

src/datasets/features/features.py
```diff
@@ -1698,14 +1701,14 @@ def unsimplify(feature: dict) -> dict:
...
                     raise ValueError(
-                        f"ClassLabel expected a value for all label ids [0:{label_ids[-1] + 1}] but some ids are missing."
+                        f"ClassLabel expected a value for all label ids [0:{int(label_ids[-1]) + 1}] but some ids are missing."
```
lhoestq, line 1711:
```suggestion
                        f"ClassLabel expected a value for all label ids [0:{label_ids[-1] + 1}] but some ids are missing."
```

lhoestq (approved):
Thanks !

src/datasets/features/features.py
albertvillanova: I agree that the sort should be fixed for strings.

However, I wouldn't transform `label_ids` to `int` here: see some lines below:
https://github.com/huggingface/datasets/blob/2205b5f5135d0c36f9a3527beebea0aa9ee89901/src/datasets/features/features.py#L1713

This would raise an error:
- `label_id` is int
- but keys of `feature["class_label"]["names"]` are str

src/datasets/features/features.py
albertvillanova: This is a consequence of the suggested change above

src/datasets/features/features.py
albertvillanova: This is also a consequence of the suggested change above.

src/datasets/features/features.py
albertvillanova: I have used instead:
```python
label_ids = sorted(feature["class_label"]["names"], key=int)
```

lhoestq: Ok ! Feel free to merge when it's all good to you

albertvillanova:
Also note that this approach is valid when metadata keys are str, but also if they are int.

lhoestq:
perfect !

PR #5279, opened 2022-11-22
Warn about checksums

lhoestq:
It takes a lot of time on big datasets to compute the checksums, we should at least add a warning to notify the user about this step. I also mentioned how to disable it, and added a tqdm bar (delay=5 seconds)

cc @ola13

### DIFF
--- a/src/datasets/download/download_manager.py
+++ b/src/datasets/download/download_manager.py
@@ -20,6 +20,7 @@ import io
 import os
 import posixpath
 import tarfile
+import time
 import warnings
 from datetime import datetime
 from functools import partial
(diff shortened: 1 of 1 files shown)


mariosasko (approved):
Thanks, LGTM!

PS: Computing checksums is relatively slow, so it would be nice to eventually stop computing them by default. I think we could even eliminate some more verifications steps, such as the "duplicate keys" check, which is an approximate check and needed only for the outdated HF [wrapper](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/core/community/huggingface_wrapper.py) in TFDS (new [builder](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/core/dataset_builders/huggingface_dataset_builder.py)).

lhoestq:
I'm also in favor of disabling this by default - it's kinda impractical

ola13:
Great, thanks for the quick turnaround on this!

Issue #5276, opened 2022-11-22
Bug in downloading common_voice data and snall chunk of it to one's own hub


capsabogdan:
### Describe the bug

I'm trying to load the common voice dataset. Currently there is no implementation to download just par tof the data, and I need just one part of it, without downloading the entire dataset

Help please?

![image](https://user-images.githubusercontent.com/48530104/203260511-26df766f-6013-4eaf-be26-8aa13794def2.png)

### Steps to reproduce the bug

So here is what I have done:
1. Download common_voice data 
2. Trim part of it and publish it to my own repo.
3. Download data from my own repo, but am getting this error.

### Expected behavior

There shouldn't be an error in downloading part of the data and publishing it to one's own repo

### Environment info

common_voice 11

lhoestq:
Sounds like one of the file is not a valid one, can you make sure you uploaded valid mp3 files ?

capsabogdan:
Well I just sharded the original commonVoice dataset and pushed a small chunk of it in a private rep

What did go wrong?

Holen Sie sich Outlook für iOS<https://aka.ms/o0ukef>
________________________________
Von: Quentin Lhoest ***@***.***>
Gesendet: Tuesday, November 22, 2022 3:03:40 PM
An: huggingface/datasets ***@***.***>
Cc: capsabogdan ***@***.***>; Author ***@***.***>
Betreff: Re: [huggingface/datasets] Bug in downloading common_voice data and snall chunk of it to one's own hub (Issue #5276)


Sounds like one of the file is not a valid one, can you make sure you uploaded valid mp3 files ?

lhoestq:
It should be all good then !
Could you share a link to your repository for me to investigate what went wrong ?

capsabogdan:
https://huggingface.co/datasets/DTU54DL/common-voice-test16k

Am Di., 22. Nov. 2022 um 16:43 Uhr schrieb Quentin Lhoest <
***@***.***>:

lhoestq:
I see ! This is a bug with MP3 files.

When we store audio data in parquet, we store the bytes and the file name. From the file name extension we know if it's a WAV, an MP3 or else. But here it looks like the paths are all None.

It looks like it comes from here:

https://github.com/huggingface/datasets/blob/7feeb5648a63b6135a8259dedc3b1e19185ee4c7/src/datasets/features/audio.py#L212

Cc @polinaeterna maybe we should simply put the file name instead of None values ?

[...]

PR #5285, opened 2022-11-23
Save file name in embed_storage

lhoestq:
Having the file name is useful in case we need to check the extension of the file (e.g. mp3), or in general in case it includes some metadata information (track id, image id etc.)

Related to https://github.com/huggingface/datasets/issues/5276

### DIFF
--- a/src/datasets/features/audio.py
+++ b/src/datasets/features/audio.py
@@ -204,12 +204,11 @@ class Audio:
             storage = pa.StructArray.from_arrays([bytes_array, path_array], ["bytes", "path"], mask=storage.is_null())
         return array_cast(storage, self.pa_type)
 
-    def embed_storage(self, storage: pa.StructArray, drop_paths: bool = True) -> pa.StructArray:
+    def embed_storage(self, storage: pa.StructArray) -> pa.StructArray:
         """Embed audio files into the Arrow array.
 
         Args:
             storage (pa.StructArray): PyArrow array to embed.
-            drop_paths (bool, default ``True``): If True, the paths are set to None.
 
         Returns:
             pa.StructArray: Array in the Audio arrow storage type, that is
@@ -229,7 +228,10 @@ class Audio:
             ],
             type=pa.binary(),
         )
-        path_array = pa.array([None] * len(storage), type=pa.string()) if drop_paths else storage.field("path")
+        path_array = pa.array(
+            [os.path.basename(path) if path is not None else None for path in storage.field("path").to_pylist()],
+            type=pa.string(),
+        )
         storage = pa.StructArray.from_arrays([bytes_array, path_array], ["bytes", "path"], mask=bytes_array.is_null())
         return array_cast(storage, self.pa_type)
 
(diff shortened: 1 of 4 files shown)


src/datasets/features/audio.py
```diff
@@ -229,7 +228,10 @@ def path_to_bytes(path):
             ],
             type=pa.binary(),
         )
-        path_array = pa.array([None] * len(storage), type=pa.string()) if drop_paths else storage.field("path")
+        path_array = pa.array(
+            [os.path.basename(path) if path is not None else None for path in storage.field("path").to_pylist()],
```
polinaeterna, line 232: I  would suggest to store full paths to avoid ambiguities.

polinaeterna, line 232: I would suggest to store full paths to avoid ambiguities.

polinaeterna (commented):
Nice, thank you!

I've left a comment and I think a couple of tests that fail on the CI should be fixed according to his change - not assert that `"path"` is None (like in `test_writer_embed_local_files`)

src/datasets/features/audio.py
lhoestq: It may store personal information about the user though, not a big fan

lhoestq:
I updated the tests, met le know if it sounds good to you now :)

src/datasets/features/audio.py
polinaeterna: okay, agree, users can include full paths in a separate field if needed

tests/test_table.py
```diff
@@ -1107,26 +1107,26 @@ def test_cast_array_to_features_to_null_type():
 def test_embed_array_storage(image_file):
     array = pa.array([{"bytes": None, "path": image_file}], type=Image.pa_type)
     embedded_images_array = embed_array_storage(array, Image())
-    assert embedded_images_array.to_pylist()[0]["path"] is None
+    assert isinstance(embedded_images_array.to_pylist()[0]["path"], str)
```
polinaeterna, line 1110: We can also check the filename correctness explicitly:

```suggestion
    assert isinstance(embedded_images_array.to_pylist()[0]["path"], str)
    assert embedded_images_array.to_pylist()[0]["path"] == "test_image_rgb.jpg"
```
(here and below)

tests/test_upstream_hub.py
```diff
@@ -336,7 +336,7 @@ def test_push_dataset_to_hub_custom_features_audio(self, temporary_repo):
                 hub_ds = hub_ds.cast_column("x", Audio(decode=False))
                 elem = hub_ds[0]["x"]
                 path, bytes_ = elem["path"], elem["bytes"]
-                assert bool(path) == (not embed_external_files)
+                assert isinstance(path, str)
```
polinaeterna, line 339:
```suggestion
                assert isinstance(path, str)
                assert os.path.basename(path) == "test_audio_44100.wav"
```

polinaeterna (approved):
looks good to me! added a small suggestion for tests

PR #5290, opened 2022-11-24
fix error where reading breaks when batch missing an assigned column feature

eunseojo:


src/datasets/packaged_modules/json/json.py
```diff
@@ -128,6 +128,13 @@ def _generate_tables(self, files):
                                     pa_table = paj.read_json(
                                         io.BytesIO(batch), read_options=paj.ReadOptions(block_size=block_size)
                                     )
+                                    feature_columns = set(self.config.features.arrow_schema.names)
```
lhoestq, line 131: `self.config.features` is `None` by default - because before reading the data we don't know which columns it contains.

lhoestq (commented):
Thanks ! Can you share a code example where this was not working ? It will be useful to write tests

src/datasets/packaged_modules/json/json.py
eunseojo: ah this was made to work when features is passed in by the user in the function. maybe i can make this dynamic so that it works even when not explicitly provided.

PR #5294, opened 2022-11-24
Support streaming datasets with pathlib.Path.with_suffix

### LINKED ISSUE #5293: Support streaming datasets with pathlib.Path.with_suffix
Extend support for streaming datasets that use `pathlib.Path.with_suffix`.

This feature will be useful e.g. for datasets containing text files and annotated files with the same name but different extension.

albertvillanova:
This PR extends the support in streaming mode for datasets that use `pathlib.Path.with_suffix`.

Fix #5293.


lhoestq (approved):
Cool thanks !

PR #5297, opened 2022-11-25
Fix xjoin for Windows pathnames

### LINKED ISSUE #5296: Bug in xjoin with Windows pathnames
Currently, `xjoin` function has a bug with local Windows pathnames: instead of returning the OS-dependent join pathname, it always returns it in POSIX format.

```python
from datasets.download.streaming_download_manager import xjoin

path = xjoin("C:\\Users\\USERNAME", "filename.txt")
```

Join path should be:
```python
"C:\\Users\\USERNAME\\filename.txt"
```
However it is:
```python
"C:/Users/USERNAME/filename.txt"
```

albertvillanova:
This PR fixes a bug in `xjoin` function with Windows pathnames.

Fix #5296.

### DIFF
--- a/src/datasets/download/streaming_download_manager.py
+++ b/src/datasets/download/streaming_download_manager.py
@@ -104,10 +104,10 @@ def xjoin(a, *p):
     """
     a, *b = str(a).split("::")
     if is_local_path(a):
-        a = Path(a, *p).as_posix()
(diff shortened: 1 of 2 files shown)


lhoestq (approved):
Nice thanks

PR #5299, opened 2022-11-25
Fix xopen for Windows pathnames

### LINKED ISSUE #5298: Bug in xopen with Windows pathnames
Currently, `xopen` function has a bug with local Windows pathnames:

From its implementation:
```python
def xopen(file: str, mode="r", *args, **kwargs):
    file = _as_posix(PurePath(file))
    main_hop, *rest_hops = file.split("::")
    if is_local_path(main_hop):
        return open(file, mode, *args, **kwargs)
```

On a Windows machine, if we pass the argument:
```python
xopen("C:\\Users\\USERNAME\\filename.txt")
```
it returns
```python
open("C:/Users/USERNAME/filename.txt")
```

albertvillanova:
This PR fixes a bug in `xopen` function for Windows pathnames.

Fix #5298.

### DIFF
--- a/src/datasets/download/streaming_download_manager.py
+++ b/src/datasets/download/streaming_download_manager.py
@@ -9,7 +9,7 @@ import xml.dom.minidom
 from asyncio import TimeoutError
 from io import BytesIO
 from itertools import chain
-from pathlib import Path, PurePath, PurePosixPath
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
thanks !

PR #5300, opened 2022-11-25
Use same `num_proc` for dataset download and generation

mariosasko:
Use the same `num_proc` value for data download and generation. Additionally, do not set `num_proc` to 16 in `DownloadManager` by default (`num_proc` now has to be specified explicitly).


alexjc:
I noticed this bug the other day and was going to look into it! "Where are these processes coming from?" ;-)

lhoestq (approved):
Thank you !

PR #5302, opened 2022-11-25
Improve `use_auth_token` docstring and deprecate `use_auth_token` in `download_and_prepare`

mariosasko:
Clarify in the docstrings what happens when `use_auth_token` is `None` and deprecate the `use_auth_token` param in `download_and_prepare`.


src/datasets/builder.py
```diff
@@ -679,6 +685,12 @@ def download_and_prepare(
         >>> ds = builder.download_and_prepare("s3://my-bucket/my_rotten_tomatoes", storage_options=storage_options, file_format="parquet")
         ```
         """
+        if use_auth_token != "deprecated":
+            warnings.warn(
+                "'use_auth_token' was deprecated in version 2.7.1 and will be removed in 3.0.0. Pass `use_auth_token` to the initializer/`load_dataset_builder` instead.",
+                FutureWarning,
+            )
```
lhoestq, line 692: We can deprecate it, but it should still work if some users were passing it here:

```suggestion
            )
        else:
            use_auth_token = self.use_auth_token
```

src/datasets/builder.py
```diff
@@ -709,7 +721,7 @@ def download_and_prepare(
                     force_download=bool(download_mode == DownloadMode.FORCE_REDOWNLOAD),
                     force_extract=bool(download_mode == DownloadMode.FORCE_REDOWNLOAD),
                     use_etag=False,
-                    use_auth_token=use_auth_token,
+                    use_auth_token=self.use_auth_token,
```
lhoestq, line 724: And following the suggested change of my previous comment, we can keep it as it was:
```suggestion
                    use_auth_token=use_auth_token,
```

lhoestq (approved):
Thanks !

PR #5303, opened 2022-11-25
Skip dataset verifications by default

mariosasko:
Skip the dataset verifications (split and checksum verifications, duplicate keys check) by default unless a dataset is being tested (`datasets-cli test/run_beam`). The main goal is to avoid running the checksum check in the default case due to how expensive it can be for large datasets.

PS: Maybe we should deprecate `ignore_verifications`, which is `True` now by default, and give it a different name?


lhoestq:
100% agree that the checksum verification is overkill and not super useful. But I think this PR would also disable the check on num_examples no ?
 
As a user I would like to know if the dataset I'm loading changed significantly.
And I also think it can be useful to make sure the metadata are up to date.

What do you think ?

We could have a default `ignore_verifications="ignore_checksums"`

mariosasko:
> We could have a default `ignore_verifications="ignore_checksums"`

Accepting multiple types (booleans and strings) at the same time is not the best design. Maybe we could define an enum for this parameter?

polinaeterna:
so we can have three verification levels, - smth like "ignore_all" (to skip both checksums and all other info like num_examples verification), "ignore_checksums" (to skip only checksums verification), and "verify_all" (to perform all verification)?
and deprecate `ignore_verifications` param.

@mariosasko if you're not going to work on this PR in the coming days, I can take over it if you want (this PR will help me with [this issue](https://github.com/huggingface/datasets/issues/5315), not super urgent though).

mariosasko:
Okay, I propose deprecating `ignore_verifications` in favor of `verification_mode` (`load_dataset` already has `download_mode`; some other projects use this name for verification control). `verification_mode` would accept the following enum (or strings in the same manner as `download_mode` does):

```python
class VerificationMode(enum.Enum):
    FULL = "full"           # runs all verification checks 
    BASIC = "basic"     # default, runs only the cheap ones (skips the checksum check)
    NONE = "none"      # skips all the checks
```

WDTY?

lhoestq:
(copy paste from my message on slack)

What do you think of a config variable in config.py to switch from one verification mode to another ? This way we don’t deprecate anything

Many users are familiar with ignore_verifications=True, it might be overkill to deprecate it

polinaeterna:
@lhoestq So we have "basic" verification mode in `config.py` and continue to have `False` as a default 
value for `ignore_verifications`? That way running all verifications including checksums would not be possible without switching the config var, right? 

I like having a `VerificationMode` enum because it's aligned with `DownloadMode` and sounds more natural to me (`ignore_verifications` feels a bit semantically reverted but this is probably just my feeling) and it's flexible (no need to worry about `config.py`, I'm not sure that users even know it exists, wdyt?).

The usage point seems also valid to me, but cases when users are stuck with NonMatchingX errors also happen from time to time (I checked our slack and forum). 

As a note aside - I suggest to add instructions to the NonMatchingX error message (how to use `ignore_verifications` / `verification_mode`), this would save users who don't know about this param a lot of time.

lhoestq:
Ok I see. I'm fine with the new parameter then (even though I had a small pref for the config variable) :)

albertvillanova:
I like the idea of an enum and the `verification_mode` parameter. 

In relation with the config parameter, we could additionally add a `DEFAULT_VERIFICATION_MODE`, maybe only if users require it. Note that until now there wasn't any config parameter for a default `ignore_verifications` value: I guess people are explicitly passing `ignore_verifications=True`...

As a note aside, I like the suggestion by @polinaeterna: we could give actionable messages when verifying checksums. This could be done in other PR.

mariosasko:
This is ready for review. 

If `verification_mode` is None, it defaults to `VerificationMode.BASIC` instead of `VerificationMode.NONE`, so maybe we should find a better name for the latter to avoid confusion.

PS: `ignore_verifications` is still present in the `test`/`run_beam` commands for simplicity. Let me know if you think these commands should support all three modes.

src/datasets/builder.py
```diff
@@ -708,6 +717,13 @@ def download_and_prepare(
         >>> ds = builder.download_and_prepare("s3://my-bucket/my_rotten_tomatoes", storage_options=storage_options, file_format="parquet")
         ```
         """
+        if ignore_verifications != "deprecated":
+            verification_mode = "none" if ignore_verifications else "full"
+            warnings.warn(
+                "'ignore_verifications' was deprecated in favor of 'verification' in version 2.9.1 and will be removed in 3.0.0.\n"
```
lhoestq, line 723:
```suggestion
                "'ignore_verifications' was deprecated in favor of 'verification_mode' in version 2.9.1 and will be removed in 3.0.0.\n"
```

lhoestq (approved):
Looks all good ! thanks 🙌

src/datasets/load.py
```diff
@@ -1541,7 +1541,8 @@ def load_dataset(
     features: Optional[Features] = None,
     download_config: Optional[DownloadConfig] = None,
     download_mode: Optional[DownloadMode] = None,
-    ignore_verifications: bool = False,
+    verification_mode: Optional[VerificationMode] = None,
```
polinaeterna, line 1544: can we set `Optional[Union[VerificationMode,str]]` as a type annotation here (and same for `DownloadMode`)? i'm tired of my IDE shouting at me when I provide string as a `download_mode`.

src/datasets/builder.py
```diff
@@ -724,7 +740,7 @@ def download_and_prepare(
         self._output_dir = fs_token_paths[2][0] if is_local else self._fs.unstrip_protocol(fs_token_paths[2][0])
 
         download_mode = DownloadMode(download_mode or DownloadMode.REUSE_DATASET_IF_EXISTS)
-        verify_infos = not ignore_verifications
+        verification_mode = VerificationMode(verification_mode or VerificationMode.BASIC)
```
polinaeterna, line 743: maybe a stupid question idk but why not `VerificationMode(verification_mode) or VerificationMode.BASIC` (here and everywhere below)?

src/datasets/commands/test.py
```diff
@@ -54,7 +55,9 @@ def register_subcommand(parser: ArgumentParser):
             "--save_info", action="store_true", help="Save the dataset infos in the dataset card (README.md)"
         )
         test_parser.add_argument(
-            "--ignore_verifications", action="store_true", help="Run the test without checksums and splits checks"
+            "--ignore_verifications",
+            action="store_true",
+            help="Run the test without checksums and splits checks. Deprecated: use --verification_mode instead.",
```
polinaeterna, line 60:
```suggestion
            help="Run the test without checksums and splits checks.",
```
I think it's better to drop this sentence as there is no `--verification_mode` argument.

src/datasets/utils/info_utils.py
```diff
@@ -9,6 +10,28 @@
...
+    """`Enum` that specifies which verification checks to run.
+
+    The default mode is `BASIC`, which will perform only rudimentary checks to avoid slowdowns
+    when generating/downloading a dataset for the first time.
+
+    The verification modes:
+
+    |                    | Verification checks                                                           |
+    |--------------------|------------------------------------------------------------------------------ |
+    | `FULL`             | Split checks, uniqueness of the keys yielded in case of the GeneratorBuilder  |
```
polinaeterna, line 23: is uniqueness of the keys check depended on the verification_mode?

polinaeterna (approved):
thank you! I've left a couple of text suggestions and also a couple of questions.

I would also prefer to change the name for the `NONE` verification mode, but don't have really good ideas in mind. maybe smth like `SKIP_ALL` ?

src/datasets/builder.py
mariosasko: Both approaches work, so I guess this is a matter of style

src/datasets/utils/info_utils.py
mariosasko: Yes, it does

mariosasko:
> I would also prefer to change the name for the NONE verification mode, but don't have really good ideas in mind. maybe smth like SKIP_ALL ?

I decided to go with the following names:
* `no_checks` (previously `none`)
* `basic_checks` (previously `basic`)
* `all_checks` (previously `full`)

src/datasets/load.py
```diff
@@ -1715,6 +1729,13 @@ def load_dataset(
     >>> ds = load_dataset('imagefolder', data_dir='/path/to/images', split='train')
     ```
     """
+    if ignore_verifications != "deprecated":
+        verification_mode = "none" if ignore_verifications else "full"
+        warnings.warn(
+            "'ignore_verifications' was deprecated in favor of 'verification' in version 2.9.1 and will be removed in 3.0.0.\n"
```
lhoestq, line 1735:
```suggestion
            "'ignore_verifications' was deprecated in favor of 'verification_mode' in version 2.9.1 and will be removed in 3.0.0.\n"
```

src/datasets/load.py
```diff
@@ -1715,6 +1729,13 @@ def load_dataset(
...
+            f"You can remove this warning by passing 'verification_mode={verification_mode}' instead.",
```
lhoestq, line 1736:
```suggestion
            f"To ignore verifications, you can pass `verification_mode='no_checks'` instead.",
```

src/datasets/utils/info_utils.py
```diff
@@ -36,8 +59,11 @@ def verify_checksums(expected_checksums: Optional[dict], recorded_checksums: dic
     bad_urls = [url for url in expected_checksums if expected_checksums[url] != recorded_checksums[url]]
     for_verification_name = " for " + verification_name if verification_name is not None else ""
     if len(bad_urls) > 0:
-        error_msg = "Checksums didn't match" + for_verification_name + ":\n"
-        raise NonMatchingChecksumError(error_msg + str(bad_urls))
+        raise NonMatchingChecksumError(
+            f"Checksums didn't match{for_verification_name}:\n"
+            f"{bad_urls}\n"
+            'Set `verification_mode="none"` to skip checksums verification and ignore this error'
```
lhoestq, line 65:
```suggestion
            'Set `verification_mode="no_checks"` to skip checksums verification and ignore this error'
```

tests/test_builder.py
```diff
@@ -660,21 +661,25 @@ def test_generator_based_download_and_prepare(self):
                 os.path.exists(os.path.join(tmp_dir, builder.name, "default", "0.0.0", "dataset_info.json"))
             )
 
-        # Test that duplicated keys are ignored if ignore_verifications is True
+        # Test that duplicated keys are ignored if verification_mode is "NONE"
```
lhoestq, line 664:
```suggestion
        # Test that duplicated keys are ignored if verification_mode is "no_checks"
```

lhoestq (approved):
Thank you !

src/datasets/utils/info_utils.py
polinaeterna: can you please point me to the place in the code? I can't find it.... 🙈

mariosasko: This is the commit in which this check was introduced: https://github.com/huggingface/datasets/commit/beca0840a03f45f6089ff26e5776e05916a5be96

PR #5308, opened 2022-11-28
Support `topdown` parameter in `xwalk`

mariosasko:
Add support for the `topdown` parameter in `xwalk` when `fsspec>=2022.11.0` is installed.


albertvillanova (approved):
Nice... Thanks!

PR #5311, opened 2022-11-29
Add `features` param to `IterableDataset.map`

alvarobartt:
## Description

As suggested by @lhoestq in #3888, we should be adding the param `features` to `IterableDataset.map` so that the features can be preserved (not turned into `None` as that's the default behavior) whenever the user passes those as param, so as to be consistent with `Dataset.map`, as it provides the `features` param so that those are not inferred by default, but specified by the user, and later validated by `ArrowWriter`.

This is internally handled already by the functions relying on `IterableDataset.map` such as `rename_column`, `rename_columns`, and `remove_columns` as described in #5287.

## Usage Example

```python
from datasets import load_dataset, Features

ds = load_dataset("rotten_tomatoes", split="validation", streaming=True)
print(ds.info.features)
ds = ds.map(
    lambda x: {"target": x["label"]},
    features=Features(
        {"target": ds.info.features["label"], "label": ds.info.features["label"], "text": ds.info.features["text"]}
    ),
)
print(ds.info.features)
```

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -838,6 +838,7 @@ class IterableDataset(DatasetInfoMixin):
         batch_size: int = 1000,
         drop_last_batch: bool = False,
         remove_columns: Optional[Union[str, List[str]]] = None,
+        features: Optional[Features] = None,
         fn_kwargs: Optional[dict] = None,
     ) -> "IterableDataset":
         """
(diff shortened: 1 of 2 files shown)


src/datasets/iterable_dataset.py
```diff
@@ -877,6 +878,8 @@ def map(
             remove_columns (`Optional[List[str]]`, defaults to `None`): Remove a selection of columns while doing the mapping.
                 Columns will be removed before updating the examples with the output of `function`, i.e. if `function` is adding
                 columns with names in `remove_columns`, these columns will be kept.
+            features (`Optional[datasets.Features]`, defaults to `None`): Use a specific Features to store the cache file
+                instead of the automatically generated one.
```
lhoestq, line 882: there's no notion of "cache" for iterable datasets
```suggestion
            features (`Optional[datasets.Features]`, defaults to `None`): Feature types of the resulting dataset.
```

lhoestq (commented):
Cool thanks :)

I just added a comment on the docstring:

lhoestq (approved):
Thank you :)

PR #5312, opened 2022-11-29
Add DatasetDict.to_pandas

lhoestq:
From discussions in https://github.com/huggingface/datasets/issues/5189, for tabular data it doesn't really make sense to have to do
```python
df = load_dataset(...)["train"].to_pandas()
```
because many datasets are not split.

In this PR I added `to_pandas` to `DatasetDict` which returns the DataFrame:

If there's only one split, you don't need to specify the split name:
```python
df = load_dataset(...).to_pandas()
```
EDIT: and if a dataset has multiple splits:
```python
df = load_dataset(...).to_pandas(splits=["train", "test"])
# or
df = load_dataset(...).to_pandas(splits="all")

# raises an error because you need to select the split(s) to convert
load_dataset(...).to_pandas()
```

I do have one question though @merveenoyan @adrinjalali @mariosasko:

Should we raise an error if there are multiple splits and ask the user to choose one explicitly ?

### DIFF
--- a/src/datasets/dataset_dict.py
+++ b/src/datasets/dataset_dict.py
@@ -37,6 +37,10 @@ from .utils.typing import PathLike
 logger = logging.get_logger(__name__)
 
 
+class SplitsError(ValueError):
+    pass
+
+
 class DatasetDict(Dict[str, Dataset]):
     """A dictionary (dict of str: datasets.Dataset) with dataset transforms methods (map, filter, etc.)"""
 
@@ -1423,6 +1427,9 @@ class DatasetDict(Dict[str, Dataset]):
     ) -> Union[pd.DataFrame, Iterator[pd.DataFrame]]:
         """Returns the dataset as a :class:`pandas.DataFrame`. Can also return a generator for large datasets.
 
+        Raise an error if the dataset is made of several splits.
+        In that case you should first choose which split to use.
+
         Args:
             batched (:obj:`bool`): Set to :obj:`True` to return a generator that yields the dataset as batches
                 of ``batch_size`` rows. Defaults to :obj:`False` (returns the whole datasets once)
@@ -1436,14 +1443,20 @@ class DatasetDict(Dict[str, Dataset]):
 
         ```py
         >>> df = dataset_dict.to_pandas()
+        >>> df_train = dataset_dict["train"].to_pandas()
+        >>> df_test = dataset_dict["test"].to_pandas()
         ```
         """
         self._check_values_type()
         self._check_values_features()
-        if batched:
-            return (df for ds in self.values() for df in ds.to_pandas(batch_size=batch_size, batched=batched))
+        if len(self) > 1:
+            raise SplitsError(
+                "Failed to convert to pandas: please choose which split to convert. "
+                f"Available splits: {list(self)}. For example:\n    "
+                'df = ds["test"].to_pandas()'
+            )
         else:
-            return pd.concat([ds.to_pandas() for ds in self.values()])
+            return next(iter(self.values())).to_pandas(batch_size=batch_size, batched=batched)
 
 
 class IterableDatasetDict(dict):
--- a/src/datasets/dataset_dict.py
+++ b/src/datasets/dataset_dict.py
@@ -1434,7 +1434,7 @@ class DatasetDict(Dict[str, Dataset]):
         You must specify which splits to convert if the dataset is made of multiple splits.
 
         Args:
-            splits (:obj:`List[str]`, optional): List of splits to convert to a DataFrame.
+            splits (:obj:`Union[Literal["all"], List[str]]`, optional): List of splits to convert to a DataFrame.
                 You don't need to specify the splits if there's only one.
                 Use splits="all" to convert all the splits (they will be converted in the order of the dictionary).
             batched (:obj:`bool`): Set to :obj:`True` to return a generator that yields the dataset as batches
@@ -1470,6 +1470,9 @@ class DatasetDict(Dict[str, Dataset]):
                 '\n    df = ds.to_pandas(splits="all")'
             )
         splits = splits if splits is not None and splits != "all" else list(self)
+        bad_splits = list(set(splits) - set(self))
+        if bad_splits:
+            raise ValueError(f"Can't convert those splits to pandas : {bad_splits}. Available splits: {list(self)}.")
         if batched:
             return (df for split in splits for df in self[split].to_pandas(batch_size=batch_size, batched=batched))
         else:
--- a/src/datasets/dataset_dict.py
+++ b/src/datasets/dataset_dict.py
@@ -1423,14 +1423,15 @@ class DatasetDict(Dict[str, Dataset]):
         )
 
     def to_pandas(
-        self, batch_size: Optional[int] = None, batched: bool = False
+        self, splits: Optional[List[str]] = None, batch_size: Optional[int] = None, batched: bool = False
     ) -> Union[pd.DataFrame, Iterator[pd.DataFrame]]:
         """Returns the dataset as a :class:`pandas.DataFrame`. Can also return a generator for large datasets.
 
-        Raise an error if the dataset is made of several splits.
-        In that case you should first choose which split to use.
+        You must specify which splits to convert if the dataset is made of multiple splits.
 
         Args:
+            splits (:obj:`List[str]`, optional): List of splits to convert to a DataFrame.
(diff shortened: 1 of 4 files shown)


polinaeterna:
I think it's better to raise an error in cases when there are multiple splits but no split is specified so that users know for sure with which data they are working. I imagine a case when a user loads a dataset that they don't know much about (like what splits it has), and if they get a concatenation of everything it might lead to incorrect processing or interpretations and it would be hard to notice it.
("explicit is better than implicit")

lhoestq:
I just changed to raise an error if there are multiple splits. The error shows an example of how to choose a split to convert.

src/datasets/dataset_dict.py
```diff
@@ -1417,6 +1422,42 @@ def push_to_hub(
             revision=branch,
         )
 
+    def to_pandas(
+        self, batch_size: Optional[int] = None, batched: bool = False
```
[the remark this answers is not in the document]
lhoestq: Just added support for `splits='all'` and `splits=["a", "b", "c"]` :)

Let me know if it sounds good to you !

adrinjalali (approved):
Thanks, this looks awesome! ❤️

src/datasets/dataset_dict.py
```diff
@@ -1417,6 +1423,57 @@ def push_to_hub(
             revision=branch,
         )
 
+    def to_pandas(
+        self,
+        splits: Optional[Union[Literal["all"], List[str]]] = None,
```
albertvillanova, line 1428: We could also add our `Split` to the type hint.
```python
df_all = dataset_dict.to_pandas(splits=Split.ALL)
```

src/datasets/dataset_dict.py
```diff
@@ -1417,6 +1423,57 @@ def push_to_hub(
...
+        If the dataset has one split:
+        ```py
+        >>> df = dataset_dict.to_pandas()
+        ```
+
+        If the dataset has multiple splits:
+        ```py
+        >>> df_train = dataset_dict["train"].to_pandas()
+        >>> df_test = dataset_dict.to_pandas(splits="all")
+        >>> df_test = dataset_dict.to_pandas(splits=["train", "test"])
```
albertvillanova, line 1459:
```suggestion
        >>> df_all = dataset_dict.to_pandas(splits="all")
        >>> df_train_test = dataset_dict.to_pandas(splits=["train", "test"])
```

src/datasets/dataset_dict.py
albertvillanova: Should we also allow `str`?
```python
df_test = dataset_dict.to_pandas(splits="test")
```

src/datasets/dataset_dict.py
```diff
@@ -1417,6 +1423,57 @@ def push_to_hub(
...
+        splits: Optional[Union[Literal["all"], List[str]]] = None,
+        batch_size: Optional[int] = None,
+        batched: bool = False,
+    ) -> Union[pd.DataFrame, Iterator[pd.DataFrame]]:
+        """Returns the dataset as a :class:`pandas.DataFrame`. Can also return a generator for large datasets.
+
+        You must specify which splits to convert if the dataset is made of multiple splits.
+
+        Args:
+            splits (:obj:`List[str]`, optional): List of splits to convert to a DataFrame.
```
albertvillanova, line 1437: We should update the type hints here accordingly to the ones in the signature of the method.

src/datasets/dataset_dict.py
```diff
@@ -1417,6 +1423,57 @@ def push_to_hub(
...
+        ```
+        """
+        self._check_values_type()
+        self._check_values_features()
+        if splits is None and len(self) > 1:
+            raise SplitsError(
+                "Failed to convert to pandas: please choose which splits to convert. "
+                f"Available splits: {list(self)}. For example:"
+                '\n    df = ds["train"].to_pandas()'
+                '\n    df = ds.to_pandas(splits=["train", "test"])'
```
albertvillanova, line 1469:
```suggestion
                '\n    df = ds.to_pandas(splits=["train", "test"])'
                '\n    df = ds.to_pandas(splits="all")'
```

src/datasets/dataset_dict.py
```diff
@@ -1417,6 +1423,57 @@ def push_to_hub(
...
+            )
+        splits = splits if splits is not None and splits != "all" else list(self)
+        if batched:
+            return (df for split in splits for df in self[split].to_pandas(batch_size=batch_size, batched=batched))
+        else:
+            return pd.concat([self[split].to_pandas() for split in splits])
+
```
albertvillanova, line 1476: Maybe adding a validation to check if any of the passed split names is not in self, before raising an error when:
```python
self[split]
```

albertvillanova (commented):
Thanks for this enhancement that will improve UX for tabular data!!

Below some comments, questions, nits...

src/datasets/dataset_dict.py
lhoestq: If one wants to choose one split they already do `dataset_dict["test"].to_pandas()` - I don't think that introducing `splits="test"` would make it particularly easier.

Although since we don't support the `Split` API fully (e.g. doing `"train+test[:20]"`) I wouldn't necessarily add `Split` in the type hint

lhoestq:
Thanks for the review, I've updated the type hint and added a line to raise an error on bad splits :)

mariosasko:
Merging https://github.com/huggingface/datasets/pull/5301 would eliminate the need for this PR, no?

In the meantime, the current API is cleaner IMO.

lhoestq:
This solution is simpler than https://github.com/huggingface/datasets/pull/5301 and covers most cases for tabular datasets, so I'm in favor of merging this one and put https://github.com/huggingface/datasets/pull/5301 on stand by

lhoestq:
Let me know if it sounds good to you @mariosasko @albertvillanova :)

src/datasets/dataset_dict.py
```diff
@@ -1417,6 +1423,61 @@ def push_to_hub(
...
+        ```py
+        >>> df_train = dataset_dict["train"].to_pandas()
+        >>> df_all = dataset_dict.to_pandas(splits="all")
+        >>> df_train_test = dataset_dict.to_pandas(splits=["train", "test"])
+        ```
+        """
+        self._check_values_type()
+        self._check_values_features()
+        if splits is None and len(self) > 1:
+            raise SplitsError(
```
polinaeterna, line 1465: maybe invent a more specific name for this type of error? smth like `SplitsNotSpecifiedError`/`SplitsNotProvidedError` ? (subclassing `SplitsError`?)

src/datasets/dataset_dict.py
```diff
@@ -1417,6 +1423,61 @@ def push_to_hub(
...
+                "Failed to convert to pandas: please choose which splits to convert. "
+                f"Available splits: {list(self)}. For example:"
+                '\n    df = ds["train"].to_pandas()'
+                '\n    df = ds.to_pandas(splits=["train", "test"])'
+                '\n    df = ds.to_pandas(splits="all")'
+            )
+        splits = splits if splits is not None and splits != "all" else list(self)
+        bad_splits = list(set(splits) - set(self))
+        if bad_splits:
+            raise ValueError(f"Can't convert those splits to pandas : {bad_splits}. Available splits: {list(self)}.")
```
polinaeterna, line 1475: maybe raise a custom error here too? to be aligned with `UnexpectedSplits` exception in `info_utils.py`: https://github.com/huggingface/datasets/blob/cb8dd987df8f5dfa89d818a9b5fec532cf6584e6/src/datasets/utils/info_utils.py#L48 
(subclassing `SplitsError` defined above?)

polinaeterna (approved):
like it! added a small suggestion about errors.

adrinjalali:
@mariosasko the issue we're dealing with is that in tabular scenarios, we often don't have splits in the dataset, and imposing that concept to people dealing with the library hampers adoption.

mariosasko:
@adrinjalali This PR proposes a solution inconsistent with the existing API (in other words, a solution that clutters our API 🙂). Moreover, our library primarily focuses on larger-than-RAM datasets, and tabular datasets don't (directly) fall into this group.

Instead of the temporary "fix" proposed here, it makes much more sense to align `load_dataset` with both tabular and DL workflows "in a consistent way", so I suggest we continue our discussion from https://github.com/huggingface/datasets/issues/5189 to have this resolved by version 3.0.

lhoestq:
closing this one for now

PR #5313, opened 2022-11-29
Fix description of streaming in the docs

polinaeterna:
We say that "the data is being downloaded progressively" which is not true, it's just streamed, so I fixed it. Probably I missed some other places where it is written? 

Also changed docstrings for `StreamingDownloadManager`'s `download` and `extract` to reflect the same, as these docstrings are displayed in the documentation cc @lhoestq

### DIFF
--- a/src/datasets/download/streaming_download_manager.py
+++ b/src/datasets/download/streaming_download_manager.py
@@ -889,7 +889,7 @@ class StreamingDownloadManager:
             url_or_urls (`str` or `list` or `dict`): URL or URLs of files to stream data from. Each url is a `str`.
 
         Returns:
-            `str`: Urls to stream data from corresponding to the given input url_or_urls.
+            `str`: URL(s) to stream data from corresponding to the given input url_or_urls.
 
         Example:
 
@@ -909,13 +909,14 @@ class StreamingDownloadManager:
 
     def extract(self, url_or_urls):
         """Add extraction protocol for given url(s) for streaming.
-        This is the lazy version of DownloadManager.extract for streaming.
+
+        This is the lazy version of `DownloadManager.extract` for streaming.
 
         Args:
             url_or_urls (`str` or `list` or `dict`): URL or URLs of files to stream data from. Each url is a `str`.
 
         Returns:
-            `str`: Urls to stream data from matching the given input path_or_paths.
+            `str`: URL(s) to stream data from matching the given input path_or_paths.
 
         Example:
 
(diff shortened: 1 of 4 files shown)


lhoestq (approved):
Thanks ! Added a few lines in the docstrings to refer to the original methods of the DownloadManager

src/datasets/download/streaming_download_manager.py
```diff
@@ -906,14 +907,15 @@ def _download(self, urlpath: str) -> str:
             urlpath = url_or_path_join(self._base_path, urlpath)
         return urlpath
 
-    def extract(self, path_or_paths):
-        """Extract given path(s).
+    def extract(self, url_or_urls):
+        """Add extraction protocol for given url(s) for streaming.
+        This is the lazy version of DownloadManager.extract for streaming.
```
albertvillanova, line 912: Nit:
```suggestion

        This is the lazy version of `DownloadManager.extract` for streaming.
```

src/datasets/download/streaming_download_manager.py
```diff
@@ -882,13 +882,14 @@ def manual_dir(self):
...
+        """Normalize url(s) of files to stream data from.
+        This is the lazy version of DownloadManager.download for streaming.
 
         Args:
-            url_or_urls (`str` or `list` or `dict`): URL or URLs to download and extract. Each url is a `str`.
+            url_or_urls (`str` or `list` or `dict`): URL or URLs of files to stream data from. Each url is a `str`.
 
         Returns:
-            `str`: Downloaded paths matching the given input url_or_urls.
+            `str`: Urls to stream data from corresponding to the given input url_or_urls.
```
albertvillanova, line 892: Nit, to be consistent with the Args above:
```suggestion
            `str`: URL(s) to stream data from corresponding to the given input url_or_urls.
```

src/datasets/download/streaming_download_manager.py
```diff
@@ -944,20 +946,20 @@ def _extract(self, urlpath: str) -> str:
...
         ```
 
         Args:
-            url_or_urls: url or `list`/`dict` of urls to download and extract. Each
-                url is a `str`.
+            url_or_urls: url or `list`/`dict` of urls to stream from. Each url is a `str`.
 
         Returns:
-            extracted_path(s): `str`, extracted paths of given URL(s).
+            url(s): `str`, extracted paths of given URL(s).
```
albertvillanova, line 962:
```suggestion
            `str`: Extracted paths of given URL(s).
```

albertvillanova (approved):
Thanks!

Some nits below. Feel free to ignore them.

PR #5318, opened 2022-12-01
Origin/fix missing features error

eunseojo:
This fixes the problem of when the dataset_load function reads a function with "features" provided but some read batches don't have columns that later show up. For instance, the provided "features" requires columns A,B,C but only columns B,C show. This fixes this by adding the column A with nulls.

### DIFF
--- a/src/datasets/packaged_modules/json/json.py
+++ b/src/datasets/packaged_modules/json/json.py
@@ -128,6 +128,13 @@ class Json(datasets.ArrowBasedBuilder):
                                     pa_table = paj.read_json(
                                         io.BytesIO(batch), read_options=paj.ReadOptions(block_size=block_size)
                                     )
+                                    feature_columns = set(self.config.features.arrow_schema.names)
+                                    pa_columns = set(pa_table.column_names)
+                                    missing_columns = feature_columns - pa_columns
+                                    if missing_columns:  # some columns are missing
+                                        num_rows = len(pa_table)
+                                        for column_name in missing_columns:
+                                            pa_table = pa_table.append_column(column_name, pa.nulls(num_rows))
                                     break
                                 except (pa.ArrowInvalid, pa.ArrowNotImplementedError) as e:
                                     if (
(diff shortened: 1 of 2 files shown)


src/datasets/packaged_modules/json/json.py
```diff
@@ -128,6 +128,14 @@ def _generate_tables(self, files):
...
                                         io.BytesIO(batch), read_options=paj.ReadOptions(block_size=block_size)
                                     )
+                                    if self.config.features is not None:
+                                        feature_columns = set(self.config.features.arrow_schema.names)
+                                        pa_columns = set(pa_table.column_names)
+                                        missing_columns = feature_columns - pa_columns
+                                        if missing_columns:  # some columns are missing
+                                            num_rows = len(pa_table)
+                                            for column_name in missing_columns:
+                                                pa_table = pa_table.append_column(column_name, pa.nulls(num_rows))
```
lhoestq, line 138: You may have to specify the type of the column to append - otherwise the type would be `null`

```suggestion
                                                type = self.config.features.arrow_schema.field(column_name).type
                                                pa_table = pa_table.append_column(column_name, pa.array([None] * num_rows, type=type))
```

lhoestq (commented):
Awesome thanks !

Could you also add a test in `tests/packaged_modules/test_json.py` ?

You can copy `test_json_generate_tables` and rename it something like `test_json_generate_tables_with_features`, and pass explicit features to `Json(...)`

eunseojo:
please review :) @lhoestq @ola13 thankoo

lhoestq:
Thanks :) I just updated the test to make sure it works even when there's a column missing, and did a minor change to json.py to add the missing columns for the other kinds of JSON files as well

lhoestq (approved):
Thanks for fixing this ! Also cc @mariosasko does it sound good to you as well ?

ola13:
Thanks Unso! If @lhoestq is happy then I'm also happy :D

mariosasko:
When I noticed the ping, this PR had already been merged...

Luckily, PyArrow's `read_json` behaves the same when `explicit_schema` is given via `ParseOptions`, so I'm okay with this change (our JSON loader doesn't use `read_json` for decoding JSON in some scenarios, so this manual approach is the right one).

PR #5319, opened 2022-12-01
Fix Text sample_by paragraph

### LINKED ISSUE #5316: Bug in sample_by="paragraph"
### Describe the bug

I think [this line](https://github.com/huggingface/datasets/blob/main/src/datasets/packaged_modules/text/text.py#L96) is wrong and should be `batch = f.read(self.config.chunksize)`. Otherwise it will never terminate because even when `f` is finished reading, `batch` will still be truthy from the last iteration.

```
> cat test.txt
a b c

d e f
````

```python
>>> import datasets
>>> datasets.load_dataset("text", data_files={"train":"test.txt"}, sample_by="paragraph")
```

This will go on forever.

### Expected behavior

Terminates very quickly. 

### Environment info

`version = "2.6.1"` but I think the bug is still there on main.

albertvillanova:
Fix #5316.

### DIFF
--- a/src/datasets/packaged_modules/text/text.py
+++ b/src/datasets/packaged_modules/text/text.py
@@ -93,9 +93,10 @@ class Text(datasets.ArrowBasedBuilder):
                     batch_idx = 0
                     batch = ""
                     while True:
-                        batch += f.read(self.config.chunksize)
-                        if not batch:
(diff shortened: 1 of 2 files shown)


lhoestq (approved):
Nice, thanks for the fix :)

Issue #5295, opened 2022-11-25
Extractions failed when .zip file located on read-only path (e.g., SageMaker FastFile mode)


verdimrc:
### Describe the bug

Hi,

`load_dataset()` does not work .zip files located on a read-only directory. Looks like it's because Dataset creates a lock file in the [same directory](https://github.com/huggingface/datasets/blob/df4bdd365f2abb695f113cbf8856a925bc70901b/src/datasets/utils/extract.py)  as the .zip file.

Encountered this when attempting `load_dataset()` on a datadir with SageMaker FastFile mode.

### Steps to reproduce the bug

```python
# Showing relevant lines only.

hyperparameters = {
    "dataset_name": "ydshieh/coco_dataset_script",
    "dataset_config_name": 2017,
    "data_dir": "/opt/ml/input/data/coco",
    "cache_dir": "/tmp/huggingface-cache",  # Fix dataset complains out-of-space.
    ...
}

estimator = PyTorch(
    base_job_name="clip",
    source_dir="../src/sm-entrypoint",
    entry_point="run_clip.py",  # Transformers/src/examples/pytorch/contrastive-image-text/run_clip.py
    framework_version="1.12",
    py_version="py38",
    hyperparameters=hyperparameters,
    instance_count=1,
    instance_type="ml.p3.16xlarge",
    volume_size=100,
    distribution={"smdistributed": {"dataparallel": {"enabled": True}}},
)

fast_file = lambda x: TrainingInput(x, input_mode='FastFile') 
estimator.fit(
    {
        "pre-trained": fast_file("s3://vm-sagemakerr-us-east-1/clip/pre-trained-checkpoint/"),
        "coco": fast_file("s3://vm-sagemakerr-us-east-1/clip/coco-zip-files/"),
    }
)
```

Error message:

```text
ErrorMessage "OSError: [Errno 30] Read-only file system: '/opt/ml/input/data/coco/image_info_test2017.zip.lock'
 """
 
 The above exception was the direct cause of the following exception
 Traceback (most recent call last)
 File "/opt/conda/lib/python3.8/runpy.py", line 194, in _run_module_as_main
 return _run_code(code, main_globals, None,
 File "/opt/conda/lib/python3.8/runpy.py", line 87, in _run_code
 exec(code, run_globals)
 File "/opt/conda/lib/python3.8/site-packages/mpi4py/__main__.py", line 7, in <module>
 main()
[...]

PR #5320, opened 2022-12-01
[Extract] Place the lock file next to the destination directory

lhoestq:
Previously it was placed next to the archive to extract, but the archive can be in a read-only directory as noticed in https://github.com/huggingface/datasets/issues/5295

Therefore I moved the lock location to be next to the destination directory, which is required to have write permissions


albertvillanova (changes requested):
The CI is not passing. Apparently, there is an issue with `DownloadConfig.delete_extracted`...

albertvillanova (approved):
Thanks for the fix.

PR #5321, opened 2022-12-01
Fix loading from HF GCP cache

lhoestq:
As reported in https://discuss.huggingface.co/t/error-loading-wikipedia-dataset/26599/4 it's not possible to download a cached version of Wikipedia from the HF GCP cache

I fixed it and added an integration test (runs in 10sec)

### DIFF
--- a/tests/test_hf_gcp.py
+++ b/tests/test_hf_gcp.py
@@ -11,8 +11,6 @@ from datasets.builder import DatasetBuilder
 from datasets.load import dataset_module_factory, import_main_class
 from datasets.utils.file_utils import cached_path
 
-from .utils import require_beam
-
 
 DATASETS_ON_HF_GCP = [
     {"dataset": "wikipedia", "config_name": "20220301.de"},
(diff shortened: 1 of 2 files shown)


albertvillanova (commented):
Thanks for the fix.

Do you know why this stopped working?

tests/test_hf_gcp.py
```diff
@@ -67,3 +68,23 @@ def test_dataset_info_available(self, dataset, config_name):
...
+        cache_dir=tmp_dir,
+        config_name="20220301.frr",
+        hash=dataset_module.hash,
+    )
+
+    # use the HF cloud storage, not the original download_and_prepare that uses apache-beam
+    builder_instance._download_and_prepare = None
+    builder_instance.download_and_prepare()
+    ds = builder_instance.as_dataset()
+    assert ds is not None
```
albertvillanova, line 90: Are you sure this is a regression test?

Before your fix, this test was not raising the reported `MissingBeamOptions` error, but a TypeError:
```
TypeError: 'NoneType' object is not callable
```
in relation with `builder_instance._download_and_prepare = None` in your test.

albertvillanova (approved):
Thanks for the fix.

PR #5333, opened 2022-12-05
fix: 🐛 pass the token to get the list of config names

severo:
Otherwise, get_dataset_infos doesn't work on gated or private datasets, even with the correct token.


albertvillanova (approved):
Thanks for the fix!

Issue #5255, opened 2022-11-17
Add a Depth Estimation dataset - DIODE / NYUDepth / KITTI


sayakpaul:
### Name

NYUDepth

### Paper

http://cs.nyu.edu/~silberman/papers/indoor_seg_support.pdf

### Data

https://cs.nyu.edu/~silberman/datasets/nyu_depth_v2.html

### Motivation

Depth estimation is an important problem in computer vision. We have a couple of Depth Estimation models on Hub as well:

* [GLPN](https://huggingface.co/docs/transformers/model_doc/glpn)
* [DPT](https://huggingface.co/docs/transformers/model_doc/dpt)

Would be nice to have a dataset for depth estimation. These datasets usually have three things: input image, depth map image, and depth mask (validity mask to indicate if a reading for a pixel is valid or not). Since we already have [semantic segmentation datasets on the Hub](https://huggingface.co/datasets?task_categories=task_categories:image-segmentation&sort=downloads), I don't think we need any extended utilities to support this addition. 

Having this dataset would also allow us to author data preprocessing guides for depth estimation, particularly like the ones we have for other tasks ([example](https://huggingface.co/docs/datasets/image_classification)). 

Ccing @osanseviero @nateraw @NielsRogge 

Happy to work on adding it.

osanseviero:
Also cc @mariosasko and @lhoestq

sayakpaul:
> I guess we'll also have to create the NYU CS Department on the Hub ?

Yes, you're right! Let me add it to my profile first, and then we can transfer. Meanwhile, if it's recommended to loop the dataset author in here, let me know. 

Also, the NYU Depth dataset seems big. Any example scripts for creating image datasets that I could refer?

lhoestq:
You can check the imagenet-1k one.

PS: If the licenses allows it, it'b be nice to host the dataset as sharded TAR archives (like imagenet-1k) instead of the ZIP format they use:
- it will make streaming much faster
- ZIP compression is not well suited for images
- it will allow parallel processing of the dataset (you can pass a subset of shards to each worker)

sayakpaul:
> You can check the imagenet-1k one.

Where can I find the script? Are you referring to https://huggingface.co/docs/datasets/image_process ? Or is there anything more specific?

lhoestq:
You can find it here: https://huggingface.co/datasets/imagenet-1k/blob/main/imagenet-1k.py

sayakpaul:
[...]

PR #5335, opened 2022-12-06
Update tasks.json

sayakpaul:
Context:

* https://github.com/huggingface/datasets/issues/5255#issuecomment-1339107195

Cc: @osanseviero

### DIFF
--- a/src/datasets/utils/resources/tasks.json
+++ b/src/datasets/utils/resources/tasks.json
@@ -21,6 +21,9 @@
             "dialogue-generation"
         ]
     },
+    "depth-estimation": {
+        "type": "cv"
+    },
     "feature-extraction": {
         "type": "multimodal"
     },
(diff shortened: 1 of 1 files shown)


sayakpaul:
> I think the only place where we need to add it is here https://github.com/huggingface/hub-docs/blob/main/js/src/lib/interfaces/Types.ts
> 
> And I think we can remove tasks.json completely from this repo

What about the warning I mentioned in https://github.com/huggingface/datasets/issues/5255#issuecomment-1339013527? Also, the depth estimation entry is already present in https://github.com/huggingface/hub-docs/blob/main/js/src/lib/interfaces/Types.ts.

albertvillanova (commented):
Hi @sayakpaul,

Please note that `depth-estimation` is defined as a possible value for `class_categories`, not for `class_ids`. You should set instead:
```yaml
task_categories:
- depth-estimation
```

We no longer use `tasks.json` file.

sayakpaul:
The update is based on what I received in the output of the export job (c.f. https://github.com/huggingface/datasets/issues/5255#issuecomment-1339107195).

albertvillanova:
Yes, my suggestion was for the dataset card: you got the error message because you tried to set `depth-estimation` in `class_ids` instead of `class_categories`.

julien-c:
yes i would suggest rm'ing tasks.json here for clarity

sayakpaul:
Closing it.

lhoestq:
It's not clear if we can remove it btw, since old versions of `evaluate` rely on it (see https://github.com/huggingface/evaluate/pull/309)

cc @lvwerra

lhoestq:
Actually it can be removed without incidence in old versions of evaluate since we kept an hardcoded `known_task_ids` that is marked "DEPRECATED"

PR #5344, opened 2022-12-09
Clean up Dataset and DatasetDict

stevhliu:
This PR cleans up the docstrings for the other half of the methods in `Dataset` and finishes `DatasetDict`.


src/datasets/arrow_dataset.py
```diff
@@ -344,32 +344,40 @@ def to_tf_dataset(
...
+                A function or callable object (such as a `DataCollator`) that will collate
                 lists of samples into a batch.
-            collate_fn_args (:obj:`Dict`, optional): An optional `dict` of keyword arguments to be passed to the
+            collate_fn_args (`Dict`, *optional*):
+                An optional `dict` of keyword arguments to be passed to the
                 `collate_fn`.
-            label_cols (:obj:`List[str]` or :obj:`str`, default ``None``): Dataset column(s) to load as
+            label_cols (`List[str]` or `str`, defaults to `None`):
+                Dataset column(s) to load as
                 labels. Note that many models compute loss internally rather than letting Keras do it, in which case
```
lhoestq, line 372:
```suggestion
                Dataset column(s) to load as labels.
                Note that many models compute loss internally rather than letting Keras do it, in which case
```

src/datasets/arrow_dataset.py
```diff
@@ -344,32 +344,40 @@ def to_tf_dataset(
...
         Args:
-            batch_size (:obj:`int`): Size of batches to load from the dataset.
-            columns (:obj:`List[str]` or :obj:`str`, optional): Dataset column(s) to load in the tf.data.Dataset. Column
-             names that are created by the `collate_fn` and that do not exist in the original dataset can be used.
-            shuffle(:obj:`bool`, default to `False`): Shuffle the dataset order when loading. Recommended True for training, False for
+            batch_size (`int`):
+                Size of batches to load from the dataset.
+            columns (`List[str]` or `str`, *optional*):
+                Dataset column(s) to load in the `tf.data.Dataset`. Column
+                names that are created by the `collate_fn` and that do not exist in the original dataset can be used.
```
lhoestq, line 357:
```suggestion
                Dataset column(s) to load in the `tf.data.Dataset`.
                Column names that are created by the `collate_fn` and that do not exist in the original dataset can be used.
```

lhoestq (approved):
Thanks !

Issue #5345, opened 2022-12-09
Wrong dtype for array in audio features


qmeeus:
### Describe the bug

When concatenating/interleaving different datasets, I stumble into an error because the features can't be aligned. After some investigation, I understood that the audio array had different dtypes, namely `float32` and `float64`. Consequently, the datasets cannot be merged.

### Steps to reproduce the bug

For example, for `facebook/voxpopuli` and `mozilla-foundation/common_voice_11_0`:

```
from datasets import load_dataset, interleave_datasets

covost = load_dataset("mozilla-foundation/common_voice_11_0", "en", split="train", streaming=True)
voxpopuli = datasets.load_dataset("facebook/voxpopuli", "nl", split="train", streaming=True)

sample_cv, = covost.take(1)
sample_vp, = voxpopuli.take(1)

assert sample_cv["audio"]["array"].dtype == sample_vp["audio"]["array"].dtype
# Fails

dataset = interleave_datasets([covost, voxpopuli])

# ValueError: The features can't be aligned because the key audio of features {'audio_id': Value(dtype='string', id=None), 'language': Value(dtype='int64', id=None), 'audio': {'array': Sequence(feature=Value(dtype='float64', id=None), length=-1, id=None), 'path': Value(dtype='string', id=None), 'sampling_rate': Value(dtype='int64', id=None)}, 'normalized_text': Value(dtype='string', id=None), 'gender': Value(dtype='string', id=None), 'speaker_id': Value(dtype='string', id=None), 'is_gold_transcript': Value(dtype='bool', id=None), 'accent': Value(dtype='string', id=None), 'sentence': Value(dtype='string', id=None)} has unexpected type - {'array': Sequence(feature=Value(dtype='float64', id=None), length=-1, id=None), 'path': Value(dtype='string', id=None), 'sampling_rate': Value(dtype='int64', id=None)} (expected either Audio(sampling_rate=16000, mono=True, decode=True, id=None) or Value("null").
```

### Expected behavior

The audio should be loaded to arrays with a unique dtype (I guess `float32`)

### Environment info

```
- `datasets` version: 2.7.1.dev0
- Platform: Linux-4.18.0-425.3.1.el8.x86_64-x86_64-with-glibc2.28
- Python version: 3.9.15
- PyArrow version: 10.0.1
- Pandas version: 1.5.2
```

qmeeus:
[...]

PR #5347, opened 2022-12-09
Force soundfile to return float32 instead of the default float64 (Fix…

qmeeus:
…es issue #5345)

### DIFF
--- a/src/datasets/features/audio.py
+++ b/src/datasets/features/audio.py
@@ -276,7 +276,7 @@ class Audio:
                     "Decoding .opus files requires 'libsndfile'>=1.0.30, "
                     + 'it can be installed via conda: `conda install -c conda-forge "libsndfile>=1.0.30"`'
                 )
-        array, sampling_rate = sf.read(file)
+        array, sampling_rate = sf.read(file, dtype="float32")
         array = array.T
         if self.mono:
             array = librosa.to_mono(array)


lhoestq:
cc @polinaeterna

polinaeterna (commented):
Hi @qmeeus ! Good catch, thank you for the fix! 
Can you please also add a comment for this line so that we don't forget why it's needed (to be aligned with `torchaudio.load` output; we are planning to get rid of `torchaudio` some day so will probably want to get back `float64`)

I wonder if there is a bug in `interleave_datasets` cc @lhoestq , why it tries to infer features from batch when they are specified in the `_info()` and should be parsed directly from it I assume.

lhoestq:
Cool ! Feel free to add a comment in the code to explain that and we can merge :)

qmeeus (commented):
Comment added

lhoestq (dismissed):
Thank you !

mariosasko:
I'm not sure if this is a good change since we plan to get rid of `torchaudio` in the next couple of months...

lhoestq:
What do you think @polinaeterna @patrickvonplaten ? Models are usually using float32 (e.g. Wev2vec2 in `transformers`) IIRC

mariosasko:
@polinaeterna Can you comment on the linked PR to see why it's still a draft? Maybe we can help somehow to get this merged finally.

I think it's weird to align `soundfile` with `torchaudio` when the latter is only used for MP3 (and prob for not much longer).

Issue #3423, opened 2021-12-13
data duplicate when setting num_works > 1 with streaming data


cloudyuyuyu:
## Describe the bug
The data is repeated num_works times when we load_dataset with streaming and set num_works > 1 when construct dataloader

## Steps to reproduce the bug
```python
# Sample code to reproduce the bug
import pandas as pd
import numpy as np
import os

from datasets import load_dataset
from torch.utils.data import DataLoader
from tqdm import tqdm
import shutil

NUM_OF_USER = 1000000
NUM_OF_ACTION = 50000
NUM_OF_SEQUENCE = 10000
NUM_OF_FILES = 32
NUM_OF_WORKERS = 16

if __name__ == "__main__":
    shutil.rmtree("./dataset")
    for i in range(NUM_OF_FILES):
        sequence_data = pd.DataFrame(
            {
                "imei": np.random.randint(1, NUM_OF_USER, size=NUM_OF_SEQUENCE),
                "sequence": np.random.randint(1, NUM_OF_ACTION, size=NUM_OF_SEQUENCE)
            }
        )

        if not os.path.exists("./dataset"):
            os.makedirs("./dataset")

        sequence_data.to_csv(f"./dataset/sequence_data_{i}.csv",
                            
 index=False)

    dataset = load_dataset("csv",
                           data_files=[os.path.join("./dataset",file) for file in os.listdir("./dataset") if file.endswith(".csv")],
                           split="train",
                           streaming=True).with_format("torch")
    data_loader = DataLoader(dataset,
                             batch_size=1024,
                             num_workers=NUM_OF_WORKERS)
    
    result = pd.DataFrame()
    for i, batch in tqdm(enumerate(data_loader)):
        result = pd.concat([result, 
                           pd.DataFrame(batch)],
                           axis=0)
    result.to_csv(f"num_work_{NUM_OF_WORKERS}.csv", index=False)

```

## Expected results
data do not duplicate 

## Actual results
data duplicate NUM_OF_WORKERS = 16 
![image](https://user-images.githubusercontent.com/16486492/145748707-9d2df25b-2f4f-4d7b-a83e-242be4fc8934.png)

## Environment info

- `datasets` version:datasets==1.14.0
- Platform:transformers==4.11.3
- Python version:3.8
- PyArrow version:

lhoestq:
Hi ! Thanks for reporting :)

[...]

PR #5357, opened 2022-12-13
Support torch dataloader without torch formatting

lhoestq:
In https://github.com/huggingface/datasets/pull/5084 we make the torch formatting consistent with the map-style datasets formatting. More specifically, a torch formatted iterable dataset will yield torch tensors.

The previous behavior of the torch formatting for iterable dataset was simply to make the iterable dataset inherit from `torch.utils.data.Dataset` to make it work in a torch DataLoader.

For those who want to keep using a torch DataLoader without using the torch formatting, we need the `datasets.IterableDataset` class to inherit from `torch.utils.data.IterableDataset` by default.

Since we don't want to import torch on startup, I created this PR to dynamically make the `datasets.IterableDataset` class inherit form the torch one. In particular, this will work:

```python
>>> ds = load_dataset("c4", "en", streaming=True, split="train")
>>> from torch.utils.data import DataLoader
>>> dataloader = DataLoader(ds, batch_size=32, num_workers=4)
>>> for example in dataloader:
...:     ...
```

When the DataLoader checks if `ds` inherits from `torch.utils.data.IterableDataset`, I dynamically add it as a parent class. This is done by using `__get__` on `ds.__class__`.


lhoestq:
Need some more time to fix the tests, especially with pickle

src/datasets/iterable_dataset.py
```diff
@@ -739,6 +755,15 @@ def __init__(
...
         self._token_per_repo_id: Dict[str, Union[str, bool, None]] = token_per_repo_id or {}
+        _maybe_add_torch_iterable_dataset_parent_class(self.__class__)
+
+    def __getstate__(self):
+        return self.__dict__
+
+    def __setstate__(self, d):
+        self.__dict__ = d
+        # Re-add torch iterable dataset as parnt class, since dynamically added parent classes are not kept when pickling
+        _maybe_add_torch_iterable_dataset_parent_class(self.__class__)
```
mariosasko, line 766: Typo:

```suggestion
        # Re-add torch iterable dataset as a parent class, since dynamically added parent classes are not kept when pickling
        _maybe_add_torch_iterable_dataset_parent_class(self.__class__)
```

mariosasko (approved):
Looks good!

This is probably the least hacky we can get here :)

docs/source/use_with_pytorch.mdx
```diff
@@ -201,22 +201,22 @@ You must use a `BatchSampler` if you want the transform to be given full batches
 ### Stream data
 
 Loading a dataset in streaming mode is useful to progressively download the data you need while iterating over the dataset.
```
polinaeterna, line 203:
```suggestion
Loading a dataset in streaming mode allows one to iterate over the dataset without downloading it on disk.
```
incorrect streaming description from past again 👻

There is the same in line 152 (in Optimize data loading section) but I can't fix it there.

src/datasets/iterable_dataset.py
```diff
@@ -719,6 +726,15 @@ class ShufflingConfig:
     generator: np.random.Generator
 
 
+def _maybe_add_torch_iterable_dataset_parent_class(cls):
+    """Add torch.utils.data.IterableDataset as a parent class if 'torch' is imported"""
```
polinaeterna, line 730: I think it happens even if `torch` is not imported but is installed in the env. Is it an intended behavior by the way?

polinaeterna (commented):
I like the hack :) 

I just left a fix in the docs (not related to the PR).

And I actually don't quite understand the idea - what's the motivation behind making only `IterableDataset` compatible with torch DataLoader without setting the format explicitly?

lhoestq:
> And I actually don't quite understand the idea - what's the motivation behind making only IterableDataset compatible with torch DataLoader without setting the format explicitly?

Setting the format to pytorch = set the output types of the dataset to be pytorch tensors. However sometimes your dataset is not made of tensors but you still want to be able to use a pytorch DataLoader

mariosasko:
A bit more context. 

The arrow-backed `Dataset` supports `DataLoader(ds)` (even if the format is not "torch"), and we want to be able to do the same with `IterableDataset` for consistency. However, this is when the PyTorch internals come into play - an iterable dataset needs to be an instance of `torch.utils.data.IterableDataset` due to [this](https://github.com/pytorch/pytorch/blob/abc54f93145830b502400faa92bec86e05422fbd/torch/utils/data/dataloader.py#L276) check (notice there is no check for the map-style version). Hence the explicit subclassing in this PR.

lhoestq:
Exactly :) Btw I just took your comments into account @polinaeterna , so feel free to review again

polinaeterna (approved):

corbyrosset:
just checking, does this change still preserve the fix to the "data duplicate when setting num_works > 1 with streaming data" issue from before?

https://github.com/huggingface/datasets/issues/3423

lhoestq:
Yes :)

PR #5359, opened 2022-12-13
Raise error if ClassLabel names is not python list

### LINKED ISSUE #5332: Passing numpy array to ClassLabel names causes ValueError
### Describe the bug

If a numpy array is passed to the names argument of ClassLabel, creating a dataset with those features causes an error.

https://colab.research.google.com/drive/1cV_es1PWZiEuus17n-2C-w0KEoEZ68IX

TLDR:

If I define my classes as:
```
my_classes = np.array(['one', 'two', 'three'])
```

Then this errors:

```py
features = Features({'value': Value('string'), 'label': ClassLabel(names=my_classes)})
dataset = Dataset.from_list(my_data, features=features)
```
```
ValueError                                Traceback (most recent call last)
[<ipython-input-8-a8a9d53ec82f>](https://localhost:8080/#) in <module>
----> 1 dataset = Dataset.from_list(my_data, features=features)

11 frames
[/usr/local/lib/python3.8/dist-packages/datasets/utils/py_utils.py](https://localhost:8080/#) in _asdict_inner(obj)
    183             for f in fields(obj):
    184                 value = _asdict_inner(getattr(obj, f.name))
--> 185                 if not f.init or value != f.default or f.metadata.get("include_in_asdict_even_if_is_default", False):
    186                     result[f.name] = value
    187             return result

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

But this works:
```
features2 = Features({'value': Value('string'), 'label': ClassLabel(names=list(my_classes))})
dataset2 = Dataset.from_list(my_data, features=features2)
```

### Expected behavior

If I provide a numpy array of class names, I would expect either an error that the names list is the wrong type, or for it to be cast internally.

### Environment info

- `datasets` version: 2.7.1
- Platform: Linux-5.15.0-56-generic-x86_64-with-glibc2.10
- Python version: 3.8.15
- PyArrow version: 10.0.1
- Pandas version: 1.5.2

Additionally:
- Numpy version: 1.23.5

freddyheppell:
Checks type of names provided to ClassLabel to avoid easy and hard to debug errors (closes #5332 - see for discussion)

### DIFF
--- a/src/datasets/features/features.py
+++ b/src/datasets/features/features.py
@@ -19,6 +19,7 @@ import json
 import re
 import sys
 from collections.abc import Iterable, Mapping
+from collections.abc import Sequence as SequenceABC
 from dataclasses import InitVar, dataclass, field, fields
 from functools import reduce, wraps
 from operator import mul
@@ -944,7 +945,7 @@ class ClassLabel:
                 self.names = [str(i) for i in range(self.num_classes)]
             else:
                 raise ValueError("Please provide either num_classes, names or names_file.")
-        elif not isinstance(self.names, list):
+        elif not isinstance(self.names, SequenceABC):
             raise ValueError(f"Please provide names as a list, is {type(self.names)}")
         # Set self.num_classes
         if self.num_classes is None:
--- a/src/datasets/features/features.py
+++ b/src/datasets/features/features.py
@@ -946,7 +946,7 @@ class ClassLabel:
             else:
                 raise ValueError("Please provide either num_classes, names or names_file.")
         elif not isinstance(self.names, SequenceABC):
-            raise ValueError(f"Please provide names as a list, is {type(self.names)}")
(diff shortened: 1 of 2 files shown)


albertvillanova (approved):
CI is green. Thanks for the fix and the test!!

src/datasets/features/features.py
```diff
@@ -944,6 +945,8 @@ def __post_init__(self, num_classes, names_file):
                 self.names = [str(i) for i in range(self.num_classes)]
             else:
                 raise ValueError("Please provide either num_classes, names or names_file.")
+        elif not isinstance(self.names, SequenceABC):
+            raise ValueError(f"Please provide names as a list, is {type(self.names)}")
```
lhoestq, line 949: should it be a TypeError instead ?

mariosasko (approved):
Replaced `ValueError` with `TypeError`. Looks good now, thanks!

PR #5364, opened 2022-12-15
Support for writing arrow files directly with BeamWriter

mariosasko:
Make it possible to write Arrow files directly with `BeamWriter` rather than converting from Parquet to Arrow, which is sub-optimal, especially for big datasets for which Beam is primarily used.


src/datasets/utils/beam_utils.py
```diff
@@ -1,52 +1,195 @@
...
+from apache_beam.io.filesystem import CompressionTypes
+from apache_beam.io.iobase import Write
+from apache_beam.transforms import ParDo
+from packaging import version
+
+from ..config import BEAM_VERSION
+
+
+if BEAM_VERSION >= version.parse("2.44.0"):
+    from apache_beam.io.parquetio import _RowDictionariesToArrowTable
```
lhoestq, line 15: Not a big fan of using a private object, maybe we can just keep copied code here ?
The code is under Apache-2 as well so we're fine

src/datasets/utils/beam_utils.py
```diff
@@ -1,52 +1,195 @@
...
+            pcoll
+            | ParDo(_RowDictionariesToArrowTable(self._schema, self._row_group_buffer_size, self._record_batch_size))
+            | Write(self._sink)
+        )
+
+    def display_data(self):
+        return {"sink_dd": self._sink, "row_group_buffer_size": str(self._row_group_buffer_size)}
+
+
+class _IpcStreamSink(filebasedsink.FileBasedSink):
```
lhoestq, line 159: FileBasedSink also supports max_shard_size, which can be useful in our case since _prepare_split also has this argument

lhoestq (commented):
Love it ! I just added two comments:

src/datasets/utils/beam_utils.py
mariosasko: Is [this](https://github.com/apache/beam/blob/62b7247a28e1f9f4fa23cfbe202f939a5634caa8/sdks/python/apache_beam/io/filebasedsink.py#L73) the param you are referring to? Unfortunately, it's not exposed in `_ParquetSink`, so we would also have to subclass `WriteToParquet` and `_ParquetSink` to address this.

lhoestq: indeed, maybe let's ignore for now.

I also opened https://github.com/apache/beam/pull/23808 to add it to parquet sink

src/datasets/utils/beam_utils.py
mariosasko: Yes, we can do this.

Btw, I've noticed that the `window.GlobalWindows.windowed_value_at_end_of_window(table)` call in `_RowDictionariesToArrowTable` requires `apache_beam>=2.43.0` (our min version `2.26.0`). Is it OK to bump the min version to `2.43.0` to avoid these "chains" of copying the beam internals? The beam generation is used in specific scenarios, so I think this is fine. And the cached beam datasets would still be accessible with the older versions.

lhoestq: 2.43 is super recent so it's maybe not ideal

mariosasko: Bumped to 2.35 (released one year ago)

src/datasets/utils/beam_utils.py
mariosasko: I resolved this by adding these parameters to our own implementations of the Write transforms :).

mariosasko:
Deleting `BeamPipeline` and `upload_local_to_remote` would break the existing Beam scripts, so I reverted this change.

From what I understand, we need these components in our scripts for the pattern:
```python
if not pipeline.is_local():
    dl_manager.ship_files_with_pipeline()
```

I plan to address this in a subsequent PR by (implicitly) downloading the files directly to the remote storage of the non-local runners.

src/datasets/builder.py
```diff
@@ -2006,19 +2014,24 @@ def _prepare_split(
...
 
-        if max_shard_size is not None:
-            raise NotImplementedError(
-                "max_shard_size is not supported for Beam datasets."
-                "Please set it to None to use the default Apache Beam sharding and get the best performance."
-            )
+        if max_shard_size is not None and config.BEAM_VERSION < version.parse("2.41.0"):
+            raise NotImplementedError("`apache_beam>=2.41.0` is required to use max_shard_size")
+        else:
+            max_shard_size = convert_file_size_to_int(max_shard_size or config.MAX_SHARD_SIZE)
```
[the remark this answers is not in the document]
lhoestq: Not sure, what makes you think this is not enough ?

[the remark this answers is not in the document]
mariosasko: Never mind, I think it's fine. 500 MB is not small enough that it would degrade the performance due to writing files to disk too often.

tests/test_beam.py
```diff
@@ -85,15 +150,13 @@ def test_download_and_prepare(self):
...
+        original_write_arrow = datasets.utils.beam_utils.WriteToArrow
 
         expected_num_examples = len(get_test_dummy_examples())
         with tempfile.TemporaryDirectory() as tmp_cache_dir:
             builder = DummyBeamDataset(cache_dir=tmp_cache_dir, beam_runner="DirectRunner")
-            with patch("apache_beam.io.parquetio.WriteToParquet") as write_parquet_mock:
-                write_parquet_mock.side_effect = partial(original_write_parquet, num_shards=2)
+            with patch("datasets.utils.beam_utils.WriteToArrow") as write_arrow_mock:
+                write_arrow_mock.side_effect = partial(original_write_arrow, num_shards=2)
                 builder.download_and_prepare()
```
lhoestq, line 160: Maybe you can update this test to use `builder.download_and_prepare(num_shards=2)` instead of mocking ?

lhoestq (commented):
Looks all good ! I think it's be nice to try to build a small subset of wikipedia before merging

lhoestq:
I got `AttributeError: 'Pipeline' object has no attribute 'is_local'` when running
```python
load_dataset("wikipedia", language="af", date="20230101", beam_runner="DirectRunner")
```
```python
~/.cache/huggingface/modules/datasets_modules/datasets/wikipedia/aa542ed919df55cc5d3347f42dd4521d05ca68751f50dbc32bae2a7f1e167559/wikipedia.py in _split_generators(self, dl_manager, pipeline)
    965             # Use dictionary since testing mock always returns the same result.
    966         downloaded_files = dl_manager.download({"xml": xml_urls})
--> 967         if not pipeline.is_local():
    968             downloaded_files = dl_manager.ship_files_with_pipeline(downloaded_files, pipeline)
    969 

AttributeError: 'Pipeline' object has no attribute 'is_local'
```

albertvillanova:
I think we could close this PR.

lhoestq:
indeed, since Beam stuff are deprecated

PR #5365, opened 2022-12-15
fix: image array should support other formats than uint8

vigsterkr:
Currently images that are provided as ndarrays, but not in `uint8` format are going to loose data. Namely, for example in a depth image where the data is in float32 format, the type-casting to uint8 will basically make the whole image blank.
`PIL.Image.fromarray` [does support mode `F`](https://pillow.readthedocs.io/en/stable/handbook/concepts.html#concept-modes).

although maybe some further metadata could be supplied via the [Image](https://huggingface.co/docs/datasets/v2.7.1/en/package_reference/main_classes#datasets.Image) object.


mariosasko:
Hi, thanks for working on this! 

I agree that the current type-casting (always cast to `np.uint8` as Tensorflow Datasets does) is a bit too harsh. However, not all dtypes are supported in `Image.fromarray` (e.g. np.int64), so we need to treat these with special care (e.g. downcast to the closest supported dtype, maybe with warnings to let the user know what's happening).

PS: To avoid the CI failures, we need to handle two more instances of the cast to `np.uint8` (both are in the `image.py` file).

lhoestq (commented):
Looks all good :)

Can you also mention which precisions are supported and which ones are downcasted in the docs ?

Could be in https://huggingface.co/docs/datasets/about_dataset_features for examples (there is a paragraph on audio but none for image yet)

lhoestq (approved):
Just added some docs :) let me know if it sounds good to you @mariosasko and then we can merge IMO

docs/source/about_dataset_features.mdx
```diff
@@ -105,3 +105,45 @@ With `decode=False`, the [`Audio`] type simply gives you the path or the bytes o
  'path': '/root/.cache/huggingface/datasets/downloads/extracted/f14948e0e84be638dd7943ac36518a4cf3324e8b7aa331c5ab11541518e9368c/en-US~JOINT_ACCOUNT/602ba55abb1e6d0fbce92065.wav',
  'transcription': 'I would like to set up a joint account with my partner'}
 ```
+
+## Image feature
+
+Image datasets have a column with type [`Image`], which load `PIL.Image` objects from images stored as bytes:
```
mariosasko, line 111:
```suggestion
Image datasets have a column with type [`Image`], which loads `PIL.Image` objects from images stored as bytes:
```

docs/source/about_dataset_features.mdx
```diff
@@ -105,3 +105,45 @@ With `decode=False`, the [`Audio`] type simply gives you the path or the bytes o
...
+With `decode=False`, the [`Image`] type simply gives you the path or the bytes of the image file, without decoding it into an `PIL.Image`, 
+
+```py
+>>> dataset = load_dataset("beans", split="train").cast_column("image", Image(decode=False))
+>>> dataset[0]["image"]
+{'bytes': None,
+ 'path': '/Users/username/.cache/huggingface/datasets/downloads/extracted/772e7c1fba622cff102b85dd74bcce46e8168634df4eaade7bedd3b8d91d3cd7/train/healthy/healthy_train.265.jpg'}
+```
+
+Depending on the dataset, you may get the path to the local downloaded image, or the content of the image as bytes if the dataset was is not originally made of individual files.
```
mariosasko, line 138:
```suggestion
Depending on the dataset, you may get the path to the local downloaded image, or the content of the image as bytes if the dataset is not made of individual files.
```

mariosasko (commented):
Two nits regarding the docs

PR #5372, opened 2022-12-17
Fix streaming pandas.read_excel

albertvillanova:
This PR fixes `xpandas_read_excel`:
- Support passing a path string, besides a file-like object
- Support passing `use_auth_token`
- First assumes the host server supports HTTP range requests; only if a ValueError is thrown (Cannot seek streaming HTTP file), then it preserves previous behavior (see [#3355](https://github.com/huggingface/datasets/pull/3355)).

Fix https://huggingface.co/datasets/bigbio/meqsum/discussions/1
Fix:
- https://github.com/bigscience-workshop/biomedical/issues/801

Related to:
- #3355


lhoestq (approved):
Cool ! You also need to add `wrap_auth(xpandas_read_excel)` in `streaming.py` to support private datasets before merging

PR #5373, opened 2022-12-17
Simplify skipping

Muennighoff:
Was hoping to find a way to speed up the skipping as I'm running into bottlenecks skipping 100M examples on C4 (it takes 12 hours to skip), but didn't find anything better than this small change :(

Maybe there's a way to directly skip whole shards to speed it up? 🧐


lhoestq (approved):
Thanks !

Yea we don't know the shard sizes so we don't know how many shards to skip

PR #5400, opened 2023-01-03
Support streaming datasets with os.path.exists and Path.exists

albertvillanova:
Support streaming datasets with `os.path.exists` and `pathlib.Path.exists`.


lhoestq (approved):
Nice thanks !

PR #5401, opened 2023-01-03
Support Dataset conversion from/to Spark

albertvillanova:
This PR implements Spark integration by supporting `Dataset` conversion from/to Spark.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -828,6 +828,34 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         return cls(table, info=info, split=split)
 
     @classmethod
+    def from_spark(
+        cls,
+        spark_df: "pyspark.sql.DataFrame",
+        features: Optional[Features] = None,
+        info: Optional[DatasetInfo] = None,
+        split: Optional[NamedSplit] = None,
+    ) -> "Dataset":
+        """
+
+        Args:
+            spark_df (`pyspark.sql.DataFrame`): Spark DataFrame to be converted to dataset.
+            features ([`Features`], *optional*): Dataset features.
+            info ([`DatasetInfo`], *optional*): Dataset information, like description, citation, etc.
+            split ([`NamedSplit`], *optional*): Name of the dataset split.
+
+        Returns:
+            [`Dataset`]
+
+         Example:
+
+        ```py
+        >>> ds = Dataset.from_spark(spark_df)
+        ```
+        """
+        spark_df.sparkSession.conf.set("spark.sql.execution.arrow.pyspark.enabled", "true")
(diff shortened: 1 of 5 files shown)


lhoestq:
Cool thanks !

Spark DataFrame are usually quite big, and I believe here `from_spark` would load everything in the driver node's RAM, which is quite limiting. Same for `to_spark` which would load everything in the driver node's RAM before sending the data to the executor. Maybe we can mention this in the docstring ?

To transfer big datasets from/into the HF ecosystem using Spark maybe we can just make sure that `pyspark` can read/write to the HF Hub, and that `datasets` can read from HDFS/S3/etc.

src/datasets/arrow_dataset.py
```diff
@@ -4553,6 +4589,22 @@ def to_pandas(
                 for offset in range(0, len(self), batch_size)
             )
 
+    def to_spark(self, spark_session: "Optional[pyspark.sql.SparkSession]" = None) -> "pyspark.sql.DataFrame":
+        """Convert to Spark DataFrame.
```
lhoestq, line 4593: Here we can either add the same warning as earlier since the full dataset goes to RAM when it's converted to pandas.

Alternatively we can send the data to the spark session by batch and use spark's `concat` on the resulting dataframes - which would allow datasets bigger than the driver's memory to be sent to the executor (and to have a nice progress bar as well). This can be done later though

lhoestq (commented):
Sounds good ! Could you also add these methods to the documentation ?

I also added one more comment:

src/datasets/arrow_dataset.py
albertvillanova: @lhoestq I didn't add the warning to `to_spark` because in this case the reason to warn is the use of our `Dataset.to_pandas` and we didn't add any warning to that other method.
- If you insist in adding a warning here in `to_spark`, then I would suggest to add the same warning to `to_pandas` for consistency.

In relation with the optimization, my original idea was to implement a simple conversion from/to Spark as a first stage. Then we can get some feedback, benchmark it in different use cases, implement optimizations,...
- See internal conversation here: https://huggingface.slack.com/archives/C012N1J633M/p1672465563591499?thread_ts=1672423801.385339&cid=C012N1J633M

lhoestq: > I didn't add the warning to to_spark because in this case the reason to warn is the use of our Dataset.to_pandas and we didn't add any warning to that other method.

Pandas dataframes are known to live in RAM, while spark DataFrame are known to live in multiple workers' RAM. However here we bring the dataset in the driver node RAM first - which can be confusing and lead to errors since it's usually a small machine. That's why for this case I think it can be helpful to have a warning.

lhoestq:
Thinking more about it I don't really see how those two methods help in practice, since one can already do `datasets` <-> pandas <-> spark and those two methods don't add value over this.

However I think it can be good documentation to explain that it's possible to do it and it's super simple

PR #5403, opened 2023-01-03
Replace one letter import in docs

MKhalusova:
This PR updates a code example for consistency across the docs based on [feedback from this comment](https://github.com/huggingface/transformers/pull/20925/files/9fda31634d203a47d3212e4e8d43d3267faf9808#r1058769500):

"In terms of style we usually stay away from one-letter imports like this (even if the community uses them) as they are not always known by beginners and one letter is very undescriptive. Here it wouldn't change anything to use albumentations instead of A."

### DIFF
--- a/docs/source/image_classification.mdx
+++ b/docs/source/image_classification.mdx
@@ -38,13 +38,13 @@ Now apply some augmentations with `albumentations`. You'll randomly crop the ima
 
 ```py
 >>> import cv2
->>> import albumentations as A
+>>> import albumentations
 >>> import numpy as np
 
->>> transform = A.Compose([
-...     A.RandomCrop(width=256, height=256),
-...     A.HorizontalFlip(p=0.5),
-...     A.RandomBrightnessContrast(p=0.2),
+>>> transform = albumentations.Compose([
+...     albumentations.RandomCrop(width=256, height=256),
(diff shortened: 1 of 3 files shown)


albertvillanova (changes requested):
Thanks for the docs fix for consistency.

Again for consistency, it would be nice to make the same fix across all the docs, e.g.
https://github.com/huggingface/datasets/blob/310cdddd1c43f9658de172b85b6509d07d5e31a1/docs/source/image_classification.mdx?plain=1#L41

MKhalusova:
> Thanks for the docs fix for consistency.
> 
> Again for consistency, it would be nice to make the same fix across all the docs, e.g.
> 
> https://github.com/huggingface/datasets/blob/310cdddd1c43f9658de172b85b6509d07d5e31a1/docs/source/image_classification.mdx?plain=1#L41

Excellent point!

albertvillanova (commented):
Also:
https://github.com/huggingface/datasets/blob/310cdddd1c43f9658de172b85b6509d07d5e31a1/docs/source/semantic_segmentation.mdx?plain=1#L97

MKhalusova:
@albertvillanova Should be all of them now :)

albertvillanova (approved):
Great! Thanks again!

PR #5411, opened 2023-01-06
Update docs of S3 filesystem with async aiobotocore

maheshpec:
[s3fs has migrated to all async calls](https://github.com/fsspec/s3fs/commit/0de2c6fb3d87c08ea694de96dca0d0834034f8bf). 
Updating documentation to use `AioSession` while using s3fs for download manager as well as working with datasets


albertvillanova (approved):
Thanks for the fix, @maheshpec!

PR #5420, opened 2023-01-13
ci: 🎡 remove two obsolete issue templates

severo:
add-dataset is not needed anymore since the "canonical" datasets are on the Hub. And dataset-viewer is managed within the datasets-server project.

See https://github.com/huggingface/datasets/issues/new/choose

Capture d’écran 2023-01-13 à 13 59 58


albertvillanova (approved):
Thanks, @severo, good catch!

PR #5441, opened 2023-01-19
resolving a weird tar extract issue

stas00:
ok, every so often, running tests would fail when it'd fail to install a dataset.

this is what I was getting once I separated the dataset install from the borked dataset install

```
$ python -c 'import sys; from datasets import load_dataset; ds=load_dataset(sys.argv[1])' HuggingFaceM4/general-pmd-synthetic-testing
No config specified, defaulting to: general-pmd-synthetic-testing/100.unique
Downloading and preparing dataset general-pmd-synthetic-testing/100.unique (download: 3.21 KiB, generated: 16.01 MiB, post-processed: Unknown size, total: 16.02 MiB) to /home/stas/.cache/huggingface/datasets/HuggingFaceM4___general-pmd-synthetic-testing/100.unique/1.1.1/86bc445e3e48cb5ef79de109eb4e54ff85b318cd55c3835c4ee8f86eae33d9d2...
Extraction of data is blocked (illegal path)
Extraction of data/1 is blocked (illegal path)
Extraction of data/1/text.null is blocked (illegal path)
[...]
```

I had no idea what to do with that - what in the world does illegal path mean?

I started looking at the code in `TarExtractor` and added a debug print of `base` so that told me that there was a problem with the current directory.

the dataset extracts into a directory `data` and the current dir I was running the tests from already had `data` in it which was a symbolic link to another partition and somehow all that `badpath` code was blowing up there.

I tried hard to come up with a repro, but no matter what I tried it only fails in that particular clone directory and not elsewhere.

In any case, I'm proposing to at least give a user a hint of what seems to be an issue. 

I'm not at all happy with the info I got with this change, but at least it gave me a hint that `TarExtractor` tries to extract into the current directory? Say what?

https://github.com/huggingface/datasets/blob/80eb8db74f49b7ee9c0f73a819c22177fabd61db/src/datasets/utils/extract.py#L110

why won't it use the `datasets` designated directory for that? There would never be a problem if it were to do that.

So perhaps you have a better solution than what I proposed in this PR. I think that code line I quoted is the one that should be fixed.

Thanks.

### DIFF
--- a/src/datasets/utils/extract.py
+++ b/src/datasets/utils/extract.py
@@ -111,7 +111,7 @@ class TarExtractor(BaseExtractor):
 
         for finfo in members:
             if badpath(finfo.name, base):
-                logger.error(f"Extraction of {finfo.name} is blocked (illegal path)")
+                logger.error(f"Extraction of {finfo.name} is blocked (illegal path: {base})")
             elif finfo.issym() and badlink(finfo, base):
                 logger.error(f"Extraction of {finfo.name} is blocked: Hard link to {finfo.linkname}")
             elif finfo.islnk() and badlink(finfo, base):


stas00:
I think I managed to reproduce it:

```
rm -rf ~/.cache/huggingface/datasets/HuggingFaceM4___cm4-synthetic-testing
mkdir -p /tmp/xxx/hf-data
sudo ln -s /tmp/xxx /test
mkdir -p /tmp/yyy
ln -sf /test/hf-data /tmp/yyy/data
cd /tmp/yyy
python -c 'import sys; from datasets import load_dataset; ds=load_dataset(sys.argv[1])' HuggingFaceM4/cm4-synthetic-testing
```

please note it includes a creation of a symlink from the `/` - may be there is a simpler way but I'm just trying to replicate the real setup.

this fails with:

```
No config specified, defaulting to: cm4-synthetic-testing/100.unique
Downloading and preparing dataset cm4-synthetic-testing/100.unique (download: 20.71 KiB, generated: 49.99 MiB, post-processed: Unknown size, total: 50.01 MiB) to /home/stas/.cache/huggingface/datasets/HuggingFaceM4___cm4-synthetic-testing/100.unique/1.1.1/2e33dcc086c7209b8ccff4b19e44f1d41b5be53262e7d793142b96c2e984602b...
Extraction of data is blocked (illegal path: /tmp/yyy)
[...]
Extraction of data/115/texts_03.txt is blocked (illegal path: /tmp/yyy)
Generating 100.unique split:   0%|                                                                         | 0/100 [00:00<?, ? examples/s]Generating 100-long unique records split

  File "/mnt/nvme0/code/huggingface/datasets-master/src/datasets/builder.py", line 1571, in _prepare_split_single
    for key, record in generator:
  File "/home/stas/.cache/huggingface/modules/datasets_modules/datasets/HuggingFaceM4--cm4-synthetic-testing/2e33dcc086c7209b8ccff4b19e44f1d41b5be53262e7d793142b96c2e984602b/cm4-synthetic-testing.py", line 190, in _generate_examples
    raise ValueError(f"can't find any data - check {data_path}")
ValueError: can't find any data - check /home/stas/.cache/huggingface/datasets/downloads/extracted/134227b9b94c4eccf19b205bf3021d4492d0227b9be6c2ddb6bf517d8d55a8cb/data

The above exception was the direct cause of the following exception:

  File "/mnt/nvme0/code/huggingface/datasets-master/src/datasets/load.py", line 1757, in load_dataset
  File "/mnt/nvme0/code/huggingface/datasets-master/src/datasets/builder.py", line 860, in download_and_prepare
    self._download_and_prepare(
  File "/mnt/nvme0/code/huggingface/datasets-master/src/datasets/builder.py", line 1612, in _download_and_prepare
    super()._download_and_prepare(
  File "/mnt/nvme0/code/huggingface/datasets-master/src/datasets/builder.py", line 953, in _download_and_prepare
  File "/mnt/nvme0/code/huggingface/datasets-master/src/datasets/builder.py", line 1450, in _prepare_split
    for job_id, done, content in self._prepare_split_single(
  File "/mnt/nvme0/code/huggingface/datasets-master/src/datasets/builder.py", line 1607, in _prepare_split_single
    raise DatasetGenerationError("An error occurred while generating the dataset") from e
datasets.builder.DatasetGenerationError: An error occurred while generating the dataset
```

note that `illegal path: /tmp/yyy` is now with the mods of this PR.

albertvillanova (approved):
Hi @stas00, thanks a lot for your thorough investigation and detailed analysis.

Precisely, those specific lines in extract code were added by @lhoestq to fix a directory traversal vulnerability (CVE-2007-4559) with this PR:
- #5016

Maybe he can give a better explanation for the chosen solution.

In the meantime, I agree we could improve the log error message with your proposal.

stas00:
Thank you, Albert for the explanation.

To summarize I think what's needed is:

1. add a comment in the code to why this is done for someone being puzzled over the odd code
2. and to use an actionable by the user error message
3. untrapped assert on that tar extract error and not continue, so that the user will not get a later misleading error that the folder is empty and is completely not actionable and it's is far from obvious that one needs to scroll up to find earlier errors, but trapped.

After reading the advisory I'm still not sure why `cwd` is used and not a designated `datasets` `extracted` one, I can't see what difference does it make since I could `chdir` to the designated directory and it would be `cwd`.

thank you!

PR #5453, opened 2023-01-23
Fix base directory while extracting insecure TAR files

albertvillanova:
This PR fixes the extraction of insecure TAR files by changing the base path against which TAR members are compared:
- from: "."
- to: `output_path`

This PR also adds tests for extracting insecure TAR files.

Related to:
- #5441
- #5452

@stas00 please note this PR addresses just one of the issues you pointed out: the use of the cwd by the extractor. The other issues (actionable error messages, raise instead of log error) should be addressed in other PRs.

### DIFF
--- a/src/datasets/utils/extract.py
+++ b/src/datasets/utils/extract.py
@@ -84,7 +84,7 @@ class TarExtractor(BaseExtractor):
         return tarfile.is_tarfile(path)
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


lhoestq (approved):
Thanks !

stas00:
Thanks a lot, @albertvillanova - I validated that your fix solves the original problem!

Issue #5418, opened 2023-01-12
Add ProgressBar for `to_parquet`


zanussbaum:
### Feature request

Add a progress bar for `Dataset.to_parquet`, similar to how `to_json` works.

### Motivation

It's a bit frustrating to not know how long a dataset will take to write to file and if it's stuck or not without a progress bar

### Your contribution

Sure I can help if needed

albertvillanova:
Thanks for your proposal, @zanussbaum. Yes, I agree that would definitely be a nice feature to have!

zanussbaum:
@albertvillanova I’m happy to make a quick PR for the feature! let me know

lhoestq:
That would be awesome ! You can comment `#self-assign` to assign you to this issue and open a PR :) Will be happy to review

zanussbaum:
#self-assign

zanussbaum:
Closing as this has been merged @lhoestq

PR #5456, opened 2023-01-23
feat: tqdm for `to_parquet`

zanussbaum:
As described in #5418

I noticed also that the `to_json` function supports multi-workers whereas `to_parquet`, is that not possible/not needed with Parquet or something that hasn't been implemented yet?


lhoestq (approved):
Thank you ! LGTM :)

The CI fails are unrelated to this PR, so we can merge

PR #5459, opened 2023-01-24
Disable aiohttp requoting of redirection URL

albertvillanova:
The library `aiohttp` performs a requoting of redirection URLs that unquotes the single quotation mark character: `%27` => `'`

This is a problem for our Hugging Face Hub, which requires exact URL from location header.

Specifically, in the query component of the URL (`https://netloc/path?query`), the value for `response-content-disposition` contains `%27`:
```
response-content-disposition=attachment%3B+filename*%3DUTF-8%27%27sample.jsonl.gz%3B+filename%3D%22sample.jsonl.gz%22%3B
```
and after the requoting, the `%27` characters get unquoted to `'`:
```
response-content-disposition=attachment%3B+filename*%3DUTF-8''sample.jsonl.gz%3B+filename%3D%22sample.jsonl.gz%22%3B
```

This PR disables the `aiohttp` requoting of redirection URLs.

### DIFF
--- a/src/datasets/filesystems/compression.py
+++ b/src/datasets/filesystems/compression.py
@@ -39,7 +39,7 @@ class BaseCompressedFileFileSystem(AbstractArchiveFileSystem):
             mode="rb",
             protocol=target_protocol,
             compression=self.compression,
-            client_kwargs={"requote_redirect_url": False},
+            client_kwargs={"requote_redirect_url": False},  # see https://github.com/huggingface/datasets/pull/5459
             **(target_options or {}),
         )
         self.compressed_name = os.path.basename(self.file.path.split("::")[0])
--- a/src/datasets/filesystems/compression.py
+++ b/src/datasets/filesystems/compression.py
@@ -35,7 +35,12 @@ class BaseCompressedFileFileSystem(AbstractArchiveFileSystem):
         super().__init__(self, **kwargs)
         # always open as "rb" since fsspec can then use the TextIOWrapper to make it work for "r" mode
         self.file = fsspec.open(
-            fo, mode="rb", protocol=target_protocol, compression=self.compression, **(target_options or {})
+            fo,
+            mode="rb",
+            protocol=target_protocol,
+            compression=self.compression,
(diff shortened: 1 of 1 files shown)


albertvillanova:
Comment by @lhoestq:
> Do you think we need this in `datasets` if it's fixed on the moon landing side ? In the aiohttp doc they consider those symbols as "non-safe"

albertvillanova:
The lib `requests` does not perform that requote on redirect URLs.

lhoestq (approved):
I'm fine with aligning with `requests` - this way we have less differences between streaming and non-streaming.

Maybe add a comment on why you needed to add this parameter ?

albertvillanova:
Indeed, the `requests` library does perform a requoting, but this does not unquote `%27`:
```python
In [1]: from requests.utils import requote_uri

In [2]: url = "https://netloc/path?param=param%27%27value"

In [3]: url
Out[3]: 'https://netloc/path?param=param%27%27value'

In [4]: requote_uri(url)
Out[4]: 'https://netloc/path?param=param%27%27value'
```

src/datasets/filesystems/compression.py
```diff
@@ -35,7 +35,12 @@ def __init__(
         super().__init__(self, **kwargs)
         # always open as "rb" since fsspec can then use the TextIOWrapper to make it work for "r" mode
         self.file = fsspec.open(
-            fo, mode="rb", protocol=target_protocol, compression=self.compression, **(target_options or {})
+            fo,
+            mode="rb",
+            protocol=target_protocol,
+            compression=self.compression,
+            client_kwargs={"requote_redirect_url": False},
```
lhoestq, line 42:
```suggestion
            client_kwargs={"requote_redirect_url": False},  # see https://github.com/huggingface/datasets/pull/5459
```

lhoestq (commented):
Thanks for the explanations :)

albertvillanova:
See the issues we opened in the respective libraries:
- aiohttp
  - aio-libs/aiohttp#7183
- requests
  - psf/requests#6341

PR #5460, opened 2023-01-24
Document that removing all the columns returns an empty document and the num_row is lost

thomasw21:


src/datasets/arrow_dataset.py
```diff
@@ -1994,6 +1994,11 @@ def remove_columns(self, column_names: Union[str, List[str]], new_fingerprint: O
             features: ['text'],
             num_rows: 1066
         })
+        >>> ds.remove_columns(column_names=ds.column_names) # Removing all the columns returns an empty dataset, and the `num_rows` information is lost
```
mariosasko, line 1997: "is lost" is maybe a bit too harsh

```suggestion
        >>> ds.remove_columns(column_names=ds.column_names) # Removing all the columns returns an empty dataset with the `num_rows` property set to 0
```

mariosasko (commented):
Nit

mariosasko (approved):
Thanks! The CI failures are unrelated to the changes, so feel free to merge

PR #5463, opened 2023-01-25
Imagefolder docs: mention support of CSV and ZIP

lhoestq:


mariosasko (approved):
Thanks for the improvements! LGTM!

PR #5466, opened 2023-01-26
remove pathlib.Path with URIs

jonny-cyberhaven:
Pathlib will convert "//" to "/" which causes retry errors when downloading from cloud storage

### DIFF
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -1841,9 +1841,9 @@ def load_from_disk(
 
     if not fs.exists(dest_dataset_path):
         raise FileNotFoundError(f"Directory {dataset_path} not found")
-    if fs.isfile(Path(dest_dataset_path, config.DATASET_INFO_FILENAME).as_posix()):
+    if fs.isfile(os.path.join(dest_dataset_path, config.DATASET_INFO_FILENAME)):
         return Dataset.load_from_disk(dataset_path, keep_in_memory=keep_in_memory, storage_options=storage_options)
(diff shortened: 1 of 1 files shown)


lhoestq:
`os.path.join` will use a backslash `\` on windows which will also fail. You can use this instead in `load_from_disk`:
```python
from .filesystems import is_remote_filesystem

is_local = not is_remote_filesystem(self._fs)
path_join = os.path.join if is_local else posixpath.join
```

lhoestq:
Thank you ! I did a minor change to not have to define a new function and I ran the CI. If it's green we can merge :)

src/datasets/load.py
```diff
@@ -1835,15 +1836,17 @@ def load_from_disk(
     # gets filesystem from dataset, either s3:// or file:// and adjusted dataset_path
     if is_remote_filesystem(fs):
         dest_dataset_path = extract_path_from_uri(dataset_path)
+        path_join = os.path.join
```
lhoestq, line 1839:
```suggestion
        path_join = posixpath.join
```

src/datasets/load.py
```diff
@@ -1835,15 +1836,17 @@ def load_from_disk(
...
     else:
         fs = fsspec.filesystem("file")
         dest_dataset_path = dataset_path
+        path_join = posixpath.join
```
lhoestq, line 1843:
```suggestion
        path_join = os.path.join
```

lhoestq (commented):
I think it's the other way around ;)

for remote filesystem we need to use a posix path, and for local we use the OS path

lhoestq (approved):
All good now :) thanks !

PR #5467, opened 2023-01-26
Fix conda command in readme

lhoestq:
The [conda forge channel](https://anaconda.org/conda-forge/datasets) is lagging behind (as of right now, only 2.7.1 is available), we should recommend using the [Hugging face channel](https://anaconda.org/HuggingFace/datasets) that we are maintaining
```
conda install -c huggingface datasets
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ pip install datasets
 🤗 Datasets can be installed using conda as follows:
 
 ```bash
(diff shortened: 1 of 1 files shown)


lhoestq:
ah didn't read well - it's all good

lhoestq:
or maybe it isn't ? `-c huggingface -c conda-forge` installs from HF or from conda-forge ?

PR #5469, opened 2023-01-26
Remove deprecated `shard_size` arg from `.push_to_hub()` 

polinaeterna:
The docstrings say that it was supposed to be deprecated since version 2.4.0, can we remove it?


lhoestq (approved):
Yes good catch ! thanks

PR #5471, opened 2023-01-26
Add num_test_batches option

amyeroberts:
`to_tf_dataset` calls can be very costly because of the number of test batches drawn during `_get_output_signature`. The test batches are draw in order to estimate the shapes when creating the tensorflow dataset. This is necessary when the shapes can be irregular, but not in cases when the tensor shapes are the same across all samples. This PR adds an option to change the number of batches drawn, so the user can speed this conversion up. 

Running the following, and modifying `num_test_batches`
```
import time
from transformers import DefaultDataCollator

data_collator = DefaultDataCollator()
dataset = load_dataset("beans")
dataset = dataset["train"].with_format("np")
start = time.time()
dataset = dataset.to_tf_dataset(
    columns=["image"],
    label_cols=["label"],
    batch_size=8,
    collate_fn=data_collator,
    num_test_batches=NUM_TEST_BATCHES,
)
end = time.time()
print(end - start)
```

NUM_TEST_BATCHES=200:  0.8197s
NUM_TEST_BATCHES=50:    0.3070s
NUM_TEST_BATCHES=2:      0.1417s
NUM_TEST_BATCHES=1:      0.1352s

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -313,7 +313,7 @@ class TensorflowDatasetMixin:
         label_cols: Optional[Union[str, List[str]]] = None,
         prefetch: bool = True,
         num_workers: int = 0,
-        num_test_batches: int = 200,
+        num_test_batches: int = 20,
     ):
         """Create a `tf.data.Dataset` from the underlying Dataset. This `tf.data.Dataset` will load and collate batches from
         the Dataset, and is suitable for passing to methods like `model.fit()` or `model.predict()`. The dataset will yield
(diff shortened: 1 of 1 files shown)


src/datasets/arrow_dataset.py
```diff
@@ -347,6 +348,9 @@ def to_tf_dataset(
                 background while the model is training.
             num_workers (`int`, defaults to `0`):
                 Number of workers to use for loading the dataset. Only supported on Python versions >= 3.8.
+            num_test_batches (:obj:`int`, defaults to `200`): Number of batches to use to infer the output signature of the
+                dataset. The higher this number, the more accurate the signature will be, but the longer it will take to
+                create the dataset.
```
lhoestq, line 353:
```suggestion
            num_test_batches (`int`, defaults to `200`):
                Number of batches to use to infer the output signature of the dataset.
                The higher this number, the more accurate the signature will be, but the longer it will take to
                create the dataset.
```

lhoestq (approved):
Sounds good to me ! WDYT @Rocketknight1 ?

Rocketknight1:
I thought this issue was resolved in my parallel `to_tf_dataset` PR! I changed the default `num_test_batches` in `_get_output_signature` to 20 and used a test batch size of 1 to maximize variance to detect shorter samples. I think it's still okay to have this PR, though - but I'd use the new value of 20 as the default!

Rocketknight1 (approved):
Agreed! And LGTM now that that change is in.

Issue #5468, opened 2023-01-26
Allow opposite of remove_columns on Dataset and DatasetDict


hollance:
### Feature request

In this blog post https://huggingface.co/blog/audio-datasets, I noticed the following code:

```python
COLUMNS_TO_KEEP = ["text", "audio"]
all_columns = gigaspeech["train"].column_names
columns_to_remove = set(all_columns) - set(COLUMNS_TO_KEEP)

gigaspeech = gigaspeech.remove_columns(columns_to_remove)
```

This kind of thing happens a lot when you don't need to keep all columns from the dataset. It would be more convenient (and less error prone) if you could just write:

```python
gigaspeech = gigaspeech.keep_columns(["text", "audio"])
```

Internally, `keep_columns` could still call `remove_columns`, but it expresses more clearly what the user's intent is.

### Motivation

Less code to write for the user of the dataset.

### Your contribution

-

mariosasko:
Hi! I agree it would be nice to have a method like this. Instead of `keep_columns`, we can name it `select_columns` to be more aligned with PyArrow's naming convention (`pa.Table.select`).

avisinghal6:
Hi, I am a newbie to open source and would like to contribute. @mariosasko can I take up this issue ?

vijaybirju:
Hey, I also want to work on this issue I am a newbie to open source.

lhoestq:
This sounds related to https://github.com/huggingface/datasets/issues/5474

I'm fine with `select_columns`, or we could also override `select` to also accept a list of columns maybe ?

avisinghal6:
@lhoestq, I am planning to add a member function to the dataset class to perform the selection operation. Do you think its the right way to proceed? or there is a better option ?

lhoestq:
Unless @mariosasko thinks otherwise, I think it can go in `Dataset.select()` :)
Though some parameters like keep_in_memory, indices_cache_file_name or writer_batch_size wouldn't when selecting columns, so we would need to update the docstring as well

lhoestq:
If someone wants to give it a shot, feel free to comment `#self-assign` and it will assign the issue to you.

Feel free to ping us here if you have questions or if we can help :)

mariosasko:
I would rather have this functionality as a separate method. IMO it's always better to be explicit than to have an API where a single method can do different/uncorrelated things (somewhat reminds me of Pandas, and there is probably a good reason why PyArrow is more rigid in this aspect).

PR #5480, opened 2023-01-27
Select columns of Dataset or DatasetDict

### LINKED ISSUE #5474: Column project operation on `datasets.Dataset`
### Feature request

There is no operation to select a subset of columns of original dataset. Expected API follows.
```python
a = Dataset.from_dict({
    'int': [0, 1, 2]
    'char': ['a', 'b', 'c'],
    'none': [None] * 3,
})

b = a.project('int', 'char')  # usually, .select()

print(a.column_names)  # stdout: ['int', 'char', 'none']
print(b.column_names)  # stdout: ['int', 'char']
```
Method project can easily accept not only column names (as a `str)` but univariant function applied to corresponding column as an example. Or keyword arguments can be used in order to rename columns in advance (see `pandas`, `pyspark`, `pyarrow`, and SQL)..

### Motivation

Projection is a typical operation in every data processing library. And it is a basic block of a well-known data manipulation language like SQL. Without this operation `datasets.Dataset` interface is not complete.

### Your contribution

Not sure. Some of my PRs are still open and some do not have any discussions.

daskol:
Close #5474 and #5468.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -2180,9 +2180,11 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
             column_names = [column_names]
         for column_name in column_names:
             if column_name not in self._data.column_names:
-                raise ValueError(f"Column name {column_name} not in the "
-                                 "dataset. Current columns in the dataset: "
-                                 f"{self._data.column_names}.")
+                raise ValueError(
+                    f"Column name {column_name} not in the "
+                    "dataset. Current columns in the dataset: "
+                    f"{self._data.column_names}."
+                )
 
         dataset_columns = frozenset(column_names)
         dataset = copy.deepcopy(self)
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -2169,7 +2169,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         ```py
         >>> from datasets import load_dataset
         >>> ds = load_dataset("rotten_tomatoes", split="validation")
-        >>> ds.select_columns('text')
+        >>> ds.select_columns(['text'])
         Dataset({
             features: ['text'],
             num_rows: 1066
@@ -2188,7 +2188,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                 )
 
         dataset = copy.deepcopy(self)
-        dataset._info.features = {k: v for k, v in dataset._info.features.items() if k in column_names}
+        dataset._info.features = Features({k: v for k, v in dataset._info.features.items() if k in column_names})
         dataset._data = dataset._data.select(column_names)
         dataset._data = update_metadata_with_features(dataset._data, dataset.features)
         dataset._fingerprint = new_fingerprint
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -2178,6 +2178,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         """
         if isinstance(column_names, str):
             column_names = [column_names]
+
         for column_name in column_names:
             if column_name not in self._data.column_names:
                 raise ValueError(
@@ -2186,9 +2187,8 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                     f"{self._data.column_names}."
                 )
 
-        dataset_columns = frozenset(column_names)
         dataset = copy.deepcopy(self)
-        dataset._info.features = {k: v for k, v in dataset._info.features.items() if k in dataset_columns}
+        dataset._info.features = {k: v for k, v in dataset._info.features.items() if k in column_names}
         dataset._data = dataset._data.select(column_names)
         dataset._data = update_metadata_with_features(dataset._data, dataset.features)
         dataset._fingerprint = new_fingerprint
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -2145,6 +2145,53 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         dataset._fingerprint = new_fingerprint
         return dataset
 
+    @transmit_tasks
+    @transmit_format
+    @fingerprint_transform(inplace=False)
+    def select_columns(self, column_names: Union[str, List[str]], new_fingerprint: Optional[str] = None) -> "Dataset":
+        """Select one or several column(s) in the dataset and the features
+        associated to them.
+
+        Args:
+            column_names (`Union[str, List[str]]`):
+                Name of the column(s) to keep.
+            new_fingerprint (`str`, *optional*):
+                The new fingerprint of the dataset after transform. If `None`,
+                the new fingerprint is computed using a hash of the previous
+                fingerprint, and the transform arguments.
+
+        Returns:
+            [`Dataset`]: A copy of the dataset object which only consists of
+            selected columns.
+
+        Example:
+
+        ```py
+        >>> from datasets import load_dataset
+        >>> ds = load_dataset("rotten_tomatoes", split="validation")
+        >>> ds.select_columns('text')
+        Dataset({
+            features: ['text'],
+            num_rows: 1066
+        })
+        ```
+        """
+        if isinstance(column_names, str):
+            column_names = [column_names]
+        for column_name in column_names:
+            if column_name not in self._data.column_names:
+                raise ValueError(f"Column name {column_name} not in the "
+                                 "dataset. Current columns in the dataset: "
+                                 f"{self._data.column_names}.")
+
+        dataset_columns = frozenset(column_names)
+        dataset = copy.deepcopy(self)
+        dataset._info.features = {k: v for k, v in dataset._info.features.items() if k in dataset_columns}
+        dataset._data = dataset._data.select(column_names)
+        dataset._data = update_metadata_with_features(dataset._data, dataset.features)
+        dataset._fingerprint = new_fingerprint
+        return dataset
+
     def __len__(self):
         """Number of rows in the dataset.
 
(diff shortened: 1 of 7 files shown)


src/datasets/arrow_dataset.py
```diff
@@ -2145,6 +2145,55 @@ def rename(columns):
...
+                    f"{self._data.column_names}."
+                )
+
+        dataset_columns = frozenset(column_names)
+        dataset = copy.deepcopy(self)
+        dataset._info.features = {k: v for k, v in dataset._info.features.items() if k in dataset_columns}
+        dataset._data = dataset._data.select(column_names)
+        dataset._data = update_metadata_with_features(dataset._data, dataset.features)
+        dataset._fingerprint = new_fingerprint
+        return dataset
```
mariosasko, line 2195: Some nits to be consistent with `remove_columns`:

```suggestion
        dataset = copy.deepcopy(self)
        if isinstance(column_names, str):
            column_names = [column_names]

        for column_name in column_names:
            if column_name not in self._data.column_names:
                raise ValueError(
                    f"Column name {column_name} not in the "
                    "dataset. Current columns in the dataset: "
                    f"{self._data.column_names}."
                )

        for column_name in column_names:
            del dataset._info.features[column_name]

        dataset._data = dataset._data.select(column_names)
        dataset._data = update_metadata_with_features(dataset._data, dataset.features)
        dataset._fingerprint = new_fingerprint
        return dataset
```

src/datasets/iterable_dataset.py
```diff
@@ -1494,6 +1514,59 @@ def remove_columns(self, column_names: Union[str, List[str]]) -> "IterableDatase
...
+        iterable = SelectColumnsIterable(self._ex_iterable, column_names)
+        return IterableDataset(
+            ex_iterable=iterable,
+            info=info,
+            split=self._split,
+            format_type=self._format_type,
+            shuffling=self._shuffling,
+            distributed=self._distributed,
+            token_per_repo_id=self._token_per_repo_id,
+        )
```
mariosasko, line 1568: Maybe we can use `.map`, as it's done in /`remove_columns` to avoid defining a new iterable class:

```suggestion
        def select_column_fn(example):
            if any(col not in example for col in column_names):
                raise ValueError(
                    f"Error when selecting {list(column_names)}: columns {set(column_names) - set(example)} are not in the dataset."
                )
            return {col: example[col] for col in column_names}

        original_features = self._info.features.copy() if self._info.features else None
        ds_iterable = self.map(select_column_fn, remove_columns=[original_column_name])
        if original_features is not None:
            ds_iterable._info.features = Features(
                {
                    col: original_features[col] for col in column_names
                }
            )
        return ds_iterable
```

mariosasko (commented):
Thanks! I've suggested some nits for consistency with the rest of the codebase, but looks great other than that!

src/datasets/iterable_dataset.py
daskol: I considered applying `.map()` as well. But I am doubt about this particular change since it should fail since `.map()` [updates][1] input example with mapper output (if `batched=False`). This is actually the main reason why I prefer derive from `ExamplesIterable` instead of modifying `.map()`.

Another reason is that `.map()` has a complex and overcomlicated semantic, much more complex than `.remove_columns()` or `.select_columns()` need. At the moment, `remove_columns` kwarg of `.map()` looks like an ad hoc fix and unnecessary complication of `.map()`. Also, it seems that `.map()` has broken semantic from map-reduce paradigm: number of rows in batch may vary but all input columns are preserved if there is not explicit request from my perspective.

In my opinion, column selection or removal is essentially simple iterator in `yield-from` manner and should be implemented as a separate `ProjectionIterable`. Then semantic of `.map()` should be clarified and simplified (get rid of `remove_columns` and return what mapper returns). This approach has implicit benefit in case if one decided to build abstract representation for dataset transformation. In this case `ProjectionIterable` would reflect a separate node in DAG.

[1]: https://github.com/huggingface/datasets/blob/c6e08fcfc3a04e53430c26fa7c07da4cb18d977d/src/datasets/iterable_dataset.py#L469-L470

src/datasets/arrow_dataset.py
mariosasko: > totally unclear how memory is managed

It's always unclear how memory is managed since Python relies on a garbage collector.

I'm OK with having `deepcopy` after argument validation, but let's drop `frozenset` as it seems unnecessary for the simple dict comprehension.

src/datasets/iterable_dataset.py
mariosasko: Indeed, `IterableDataset.map` now also updates the input example with the output dict to be aligned with `Dataset.map`, so let's keep the Select iterable for readability.

You can create an issue and suggest improvements there if you are not satisfied with the current `map` API.

src/datasets/arrow_dataset.py
daskol: >  but let's drop frozenset

Sure. I don't insist on using sets, however we should keep in mind that this will result in poor scalability in theory and on practice. I've made a mini benchmark in case we will return to this in the future
```python
In [1]: from random import sample
   ...: from string import ascii_letters
   ...: 
   ...: # Mean length of english word is about 4-5 chars.
   ...: keys = [''.join(sample(ascii_letters, 5)) for _ in range(1000)]
   ...: 
   ...: subset_lst = keys[:4]
   ...: subset_set = set(subset_lst)
   ...: subset_fro = frozenset(subset_lst)  # The same as set but it shows intent.
   ...: 
   ...: # Small number of columns.
   ...: %timeit {k: None for k in keys[:10] if k in subset_lst}
   ...: %timeit {k: None for k in keys[:10] if k in subset_set}
   ...: %timeit {k: None for k in keys[:10] if k in subset_fro}
   ...: 
   ...: # Large number of columns (wide table).
   ...: %timeit {k: None for k in keys if k in subset_lst}
   ...: %timeit {k: None for k in keys if k in subset_set}
   ...: %timeit {k: None for k in keys if k in subset_fro}
   ...: 
836 ns ± 5.09 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
541 ns ± 2.91 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
561 ns ± 2.72 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
80.2 µs ± 14.4 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
48.1 µs ± 2.83 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
43.6 µs ± 416 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

mariosasko: This benchmark is a bit off since it doesn't include the time needed to instantiate a `set`/`frozenset`.

Also, these dataset functions are not meant to be run in the for loop, so a slight overhead is OK (plus, in Python unless it's an order of magnitude slower, readability >> performance)

daskol: Right. It should demonstrate a typical timing. Construction time of a small `set` is a order of magnitude of list construction. Sets behave like lists for a few columns but scales much better.
```python
In [2]: %timeit list(subset_lst)
   ...: %timeit set(subset_lst)
   ...: %timeit frozenset(subset_lst)
66.4 ns ± 1.04 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
108 ns ± 3.32 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
109 ns ± 2.07 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```
And "readability" argument is always about habits and a matter of taste.

mariosasko (commented):
Simplify if-else

src/datasets/iterable_dataset.py
```diff
@@ -1494,6 +1514,59 @@ def remove_columns(self, column_names: Union[str, List[str]]) -> "IterableDatase
...
+                        "dataset. Columns in the dataset: "
+                        f"{list(self._info.features.keys())}."
+                    )
+            info.features = Features({c: info.features[c] for c in column_names})
+        else:
+            info = None
+
+        iterable = SelectColumnsIterable(self._ex_iterable, column_names)
+        return IterableDataset(
+            ex_iterable=iterable,
```
mariosasko, line 1561:
```suggestion
        if self._info:
            info = copy.deepcopy(self._info)
            if self._info.features is not None:
                for column_name in column_names:
                    if column_name not in self._info.features:
                        raise ValueError(
                            f"Column name {column_name} not in the "
                            "dataset. Columns in the dataset: "
                            f"{list(self._info.features.keys())}."
                        )
                info.features = Features({c: info.features[c] for c in column_names})

        ex_iterable = SelectColumnsIterable(self._ex_iterable, column_names)
        return IterableDataset(
            ex_iterable=ex_iterable,
```

mariosasko (commented):
Last nit

src/datasets/arrow_dataset.py
```diff
@@ -2145,6 +2145,55 @@ def rename(columns):
...
+        for column_name in column_names:
+            if column_name not in self._data.column_names:
+                raise ValueError(
+                    f"Column name {column_name} not in the "
+                    "dataset. Current columns in the dataset: "
+                    f"{self._data.column_names}."
+                )
+
+        dataset = copy.deepcopy(self)
+        dataset._info.features = {k: v for k, v in dataset._info.features.items() if k in column_names}
```
lhoestq, line 2191: type is incorrect here:
```suggestion
        dataset._info.features = Features({k: v for k, v in dataset._info.features.items() if k in column_names})
```

src/datasets/arrow_dataset.py
```diff
@@ -2145,6 +2145,55 @@ def rename(columns):
...
+        Returns:
+            [`Dataset`]: A copy of the dataset object which only consists of
+            selected columns.
+
+        Example:
+
+        ```py
+        >>> from datasets import load_dataset
+        >>> ds = load_dataset("rotten_tomatoes", split="validation")
+        >>> ds.select_columns('text')
```
lhoestq, line 2172: (nit) I'd show the main use case here, i.e. with a list
```suggestion
        >>> ds.select_columns(['text'])
```

lhoestq (commented):
Awesome thank you ! I just added some small suggestions :)

lhoestq (approved):
Thanks :)

Issue #5461, opened 2023-01-24
Discrepency in `nyu_depth_v2` dataset


awsaf49:
### Describe the bug

I think there is a discrepancy between depth map of `nyu_depth_v2` dataset [here](https://huggingface.co/docs/datasets/main/en/depth_estimation) and actual depth map. Depth values somehow got **discretized/clipped** resulting in depth maps that are different from actual ones. Here is a side-by-side comparison,

![image](https://user-images.githubusercontent.com/36858976/214381162-1d9582c2-6750-4114-a01a-61ca1cd5f872.png)

I tried to find the origin of this issue but sadly as I mentioned in tensorflow/datasets/issues/4674, the download link from `fast-depth` doesn't work anymore hence couldn't verify if the error originated there or during porting data from there to HF.

Hi @sayakpaul, as you worked on huggingface/datasets/issues/5255, if you still have access to that data could you please share the data or perhaps checkout this issue?

### Steps to reproduce the bug

This [notebook](https://colab.research.google.com/drive/1K3ZU8XUPRDOYD38MQS9nreQXJYitlKSW?usp=sharing#scrollTo=UEW7QSh0jf0i) from @sayakpaul  could be used to generate depth maps and actual ground truths could be checked from this  [dataset](https://www.kaggle.com/datasets/awsaf49/nyuv2-bts-dataset) from BTS repo.

> Note: BTS dataset has only 36K data compared to the train-test 50K. They sampled the data as adjacent frames look quite the same

### Expected behavior

Expected depth maps should be smooth rather than discrete/clipped.

### Environment info

- `datasets` version: 2.8.1.dev0
- Platform: Linux-5.10.147+-x86_64-with-glibc2.29
- Python version: 3.8.10
- PyArrow version: 9.0.0
- Pandas version: 1.3.5

sayakpaul:
Ccing @dwofk (the author of `fast-depth`). 

Thanks, @awsaf49 for reporting this. I believe this is because the NYU Depth V2 shipped from `fast-depth` is already preprocessed. 

If you think it might be better to have the NYU Depth V2 dataset from BTS [here](https://huggingface.co/datasets/sayakpaul/nyu_depth_v2) feel free to open a PR, I am happy to provide guidance :)

lhoestq:
Good catch ! Ideally it would be nice to have the datasets in the raw form, this way users can choose whatever processing they want to apply

awsaf49:
[...]

PR #5484, opened 2023-01-30
Update docs for `nyu_depth_v2` dataset

awsaf49:
This PR will fix the issue mentioned in #5461.
cc: @sayakpaul @lhoestq

### DIFF
--- a/docs/source/depth_estimation.mdx
+++ b/docs/source/depth_estimation.mdx
@@ -59,7 +59,7 @@ Before we look at the depth map, we need to first convert its data type to `uint
     <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/datasets/depth_est_target.png">
 </div>
 
-It's all black! You'll need to add some color to the depth map to visualize it properly (the utility below is taken from the [FastDepth repository](https://github.com/dwofk/fast-depth/blob/master/utils.py)).
+It's all black! You'll need to add some color to the depth map to visualize it properly. In order to do that, either we can apply color automatically during display using `plt.imshow()` or create color depth map using `plt.cm` then display it. In this example we have used the latter one, as we can save/write the colored depth map later. (the utility below is taken from the [FastDepth repository](https://github.com/dwofk/fast-depth/blob/master/utils.py)).
 
 ```py 
 >>> import numpy as np
--- a/docs/source/depth_estimation.mdx
+++ b/docs/source/depth_estimation.mdx
@@ -37,7 +37,7 @@ The dataset has two fields:
 * `image`: a PIL PNG image object with `uint8` data type.
 * `depth_map`: a PIL Tiff image object with `float32` data type which is the depth map of the image.
 
-It is mention-worthy that, JPEG/PNG format can only store `uint8` or `uint16` data. As depth map is `float32` data, it can't be stored using PNG/JPEG. However, we can save depth map using TIFF format as it supports a wider range of data types, including `float32` data.
+It is mention-worthy that JPEG/PNG format can only store `uint8` or `uint16` data. As the depth map is `float32` data, it can't be stored using PNG/JPEG. However, we can save the depth map using TIFF format as it supports a wider range of data types, including `float32` data.
 
 Next, check out an image with:
 
@@ -59,7 +59,7 @@ Before we look at the depth map, we need to first convert its data type to `uint
     <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/datasets/depth_est_target.png">
 </div>
 
-It's all black! You'll need to add some color to the depth map to visualize it properly. In order to do that, either we can apply color automatically during display using `plt.imshow()` or create color depth map using `plt.cm` then display it. In this example we have used the latter one, as we can save/write the colored depth map later. (the utility below is taken from the [FastDepth repository](https://github.com/dwofk/fast-depth/blob/master/utils.py)).
+It's all black! You'll need to add some color to the depth map to visualize it properly. To do that, either we can apply color automatically during display using `plt.imshow()` or create a colored depth map using `plt.cm` and then display it. In this example, we have used the latter one, as we can save/write the colored depth map later. (the utility below is taken from the [FastDepth repository](https://github.com/dwofk/fast-depth/blob/master/utils.py)).
 
 ```py 
 >>> import numpy as np
--- a/docs/source/depth_estimation.mdx
+++ b/docs/source/depth_estimation.mdx
@@ -91,7 +91,7 @@ It's all black! You'll need to add some color to the depth map to visualize it p
 ```
 
 <div class="flex justify-center">
-    <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/datasets/depth_est_target_viz.png">
+    <img src="https://i.postimg.cc/nLJLRnHM/depth-est-target-viz.png">
 </div>
 
 You can also visualize several different images and their corresponding depth maps.
@@ -121,7 +121,7 @@ You can also visualize several different images and their corresponding depth ma
 ```
 
 <div class="flex justify-center">
-    <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/datasets/depth_est_collage.png">
+    <img src="https://i.postimg.cc/fLWRs5SN/depth-est-collage.png">
 </div>
 
 Now apply some augmentations with `albumentations`. The augmentation transformations include:
@@ -188,7 +188,7 @@ You can verify the transformation worked by indexing into the `pixel_values` and
 ```
 
 <div class="flex justify-center">
-    <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/datasets/depth_est_sample_aug.png">
+    <img src="https://i.postimg.cc/7hfPG6Nt/depth-est-sample-aug.png">
 </div>
 
 Visualize the same transformation on the image's corresponding depth map:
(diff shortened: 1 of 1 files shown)


awsaf49:
I think I need to create another PR on https://huggingface.co/datasets/huggingface/documentation-images/tree/main/datasets for hosting the images there?

docs/source/depth_estimation.mdx
```diff
@@ -50,7 +50,7 @@ Next, check out an image with:
 Now take a look at its corresponding depth map:
 
 ```py
->>> example["depth_map"]
+>>> example["depth_map"].convert("RGB")
```
sayakpaul, line 53: Consider explaining why are we doing this.

docs/source/depth_estimation.mdx
```diff
@@ -29,7 +29,7 @@ Load the `train` split of the dataset and take a look at an example:
 >>> example = train_dataset[index]
 >>> example
 {'image': <PIL.PngImagePlugin.PngImageFile image mode=RGB size=640x480>,
- 'depth_map': <PIL.PngImagePlugin.PngImageFile image mode=L size=640x480>}
+ 'depth_map': <PIL.TiffImagePlugin.TiffImageFile image mode=F size=640x480>}
```
sayakpaul, line 32: This is new type which we should consider explaining.

docs/source/depth_estimation.mdx
```diff
@@ -121,7 +121,7 @@ You can also visualize several different images and their corresponding depth ma
 ```
 
 <div class="flex justify-center">
-    <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/datasets/depth_est_collage.png">
+    <img src="https://i.postimg.cc/fLWRs5SN/depth-est-collage.png">
```
sayakpaul, line 124: Open a PR adding these assets as Git-LFS here: https://huggingface.co/datasets/huggingface/documentation-images/

Don't forget to optimize the images so that the page loading is optimal (you can use a tool like ImageOptim for this).

sayakpaul (changes requested):
Thanks a lot for working on this! Just a few nits.

docs/source/depth_estimation.mdx
awsaf49: Otherwise, it will throw error. As now datatype is `float32`.

docs/source/depth_estimation.mdx
awsaf49: Will do

docs/source/depth_estimation.mdx
awsaf49: Should I just mention that dtype is `float32` and we need `uint8` hence the `.convert('RGB')`?
Btw if we use `plt.imshow(np.array(example["depth_map"]))` it will directly show colorize depth_map but again it will require more code change.

sayakpaul: Yeah just mentioning it sounds good. 

Regarding the second point, could you elaborate?

lhoestq:
Thanks for the update @awsaf49 !

docs/source/depth_estimation.mdx
awsaf49: For context, both code snippets below would give the same colored depth_map.

code1: 
```py
>>> def show_depthmap(depth_map):
...    if not isinstance(depth_map, np.ndarray):
...        depth_map = np.array(depth_map)

...    d_min = np.min(depth_map)
...    d_max = np.max(depth_map)
...    depth_map = colored_depthmap(depth_map, d_min, d_max)

...    plt.imshow(depth_map.astype("uint8"))
...    plt.axis("off")
...    plt.show()
```
code2:
```py
>>> def show_depthmap(depth_map):
...    if not isinstance(depth_map, np.ndarray):
...        depth_map = np.array(depth_map)

...    plt.imshow(depth_map)
...    plt.axis("off")
...    plt.show()
```

Initially depth_map is black as we're displaying it using `example["depth_map"].convert('RGB')` (PIL). If we had used `plt.imshow(np.array(example["depth_map"])` instead, we would've seen color depth_map directly. Because `plt.imshow` applies color/colormap automatically for a 2d array. 

So, even though we initially used PIL for displaying depth_map but in the end, we're using `plt.imshow()` in `show_depth_map`. So, for a reader, it might not be clear that if we want to use `plt.imshow` in the first place then why use manual colorization?

docs/source/depth_estimation.mdx
awsaf49: Simple explanation: "When it comes to storing data, JPEG and PNG are quite picky as they accept 8-bit or 16-bit unsigned integer data. However, However, TIFF is more flexible and supports a wider range of data types, including 32-bit floating point data. So if you need to store float32 data, TIFF is our best bet."

sayakpaul: Sounds good to me! 

@lhoestq WDYT?

docs/source/depth_estimation.mdx
sayakpaul: Thanks for explaining! 

I guess this is because of the type changes introduced in the last release. With an earlier `datasets` version, I don't think it would have been the case. No?

awsaf49: Even with the previous release, we would've gotten the same result. As for the 2d array `plt.imshow()` applies colormap automatically regardless of the `dtype` and also personally I did get the same result from the previous release. 

So, should I reduce the code and get rid of `colored_depthmap`? Or mention in the comment that even though manual colorization was performed but we would've gotten the same result using `plt.imshow()` as it applies colormap automatically?

sayakpaul: I guess commenting might be better as that way users will know how to deal with cases when colorization cannot be perfomed / inferred automatically. WDYT?

awsaf49: Yes I agree. Manual colorization is helpful when we want to save depth colors. By default, depth_map is 2D gray image.

docs/source/depth_estimation.mdx
lhoestq: Sounds good to me as well :)

docs/source/depth_estimation.mdx
awsaf49: Regarding this, I can't seem to clone the repo in the local or remote system, getting this error,
```
Cloning into 'documentation-images'...
error: RPC failed; HTTP 504 curl 22 The requested URL returned error: 504
fatal: the remote end hung up unexpectedly
```

sayakpaul: Maybe because of the platform outage. Could you try now?

docs/source/depth_estimation.mdx
```diff
@@ -29,13 +29,15 @@ Load the `train` split of the dataset and take a look at an example:
...
 ```
 
 The dataset has two fields:
 
-* `image`: a PIL image object.
-* `depth_map`: depth map of the image.
+* `image`: a PIL PNG image object with `uint8` data type.
+* `depth_map`: a PIL Tiff image object with `float32` data type which is the depth map of the image.
+
+It is mention-worthy that, JPEG/PNG format can only store `uint8` or `uint16` data. As depth map is `float32` data, it can't be stored using PNG/JPEG. However, we can save depth map using TIFF format as it supports a wider range of data types, including `float32` data.
```
sayakpaul, line 40:
```suggestion
It is mention-worthy that JPEG/PNG format can only store `uint8` or `uint16` data. As the depth map is `float32` data, it can't be stored using PNG/JPEG. However, we can save the depth map using TIFF format as it supports a wider range of data types, including `float32` data.
```

docs/source/depth_estimation.mdx
```diff
@@ -47,17 +49,17 @@ Next, check out an image with:
...
->>> example["depth_map"]
+>>> example["depth_map"].convert("RGB")
 ```
 
 <div class="flex justify-center">
     <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/datasets/depth_est_target.png">
 </div>
 
-It's all black! You'll need to add some color to the depth map to visualize it properly (the utility below is taken from the [FastDepth repository](https://github.com/dwofk/fast-depth/blob/master/utils.py)).
+It's all black! You'll need to add some color to the depth map to visualize it properly. In order to do that, either we can apply color automatically during display using `plt.imshow()` or create color depth map using `plt.cm` then display it. In this example we have used the latter one, as we can save/write the colored depth map later. (the utility below is taken from the [FastDepth repository](https://github.com/dwofk/fast-depth/blob/master/utils.py)).
```
sayakpaul, line 62:
```suggestion
It's all black! You'll need to add some color to the depth map to visualize it properly. To do that, either we can apply color automatically during display using `plt.imshow()` or create a colored depth map using `plt.cm` and then display it. In this example, we have used the latter one, as we can save/write the colored depth map later. (the utility below is taken from the [FastDepth repository](https://github.com/dwofk/fast-depth/blob/master/utils.py)).
```

sayakpaul (approved):
Thanks a lot for the updates!

Just some minor things remain and the we should be good to ship this 🚀

docs/source/depth_estimation.mdx
awsaf49: Yes it worked now

awsaf49: Created the PR, https://huggingface.co/datasets/huggingface/documentation-images/discussions/40

awsaf49:
> Thanks a lot for the updates!
> 
> Just some minor things remain and the we should be good to ship this 🚀

@sayakpaul I have updated the minor things. Please approve the workflows

docs/source/depth_estimation.mdx
awsaf49: @sayakpaul  I just realized I forgot to use lfs in my PR but it has already been merged. Should I create another PR?

awsaf49: @sayakpaul should I create another PR on https://huggingface.co/datasets/huggingface/documentation-images with lfs?

sayakpaul: Sure.

awsaf49: @sayakpaul  created the PR, https://huggingface.co/datasets/huggingface/documentation-images/discussions/41

awsaf49:
I think this PR is good to go..
@sayakpaul @lhoestq

PR #5410, opened 2023-01-05
Map-style Dataset to IterableDataset

lhoestq:
Added `ds.to_iterable()` to get an iterable dataset from a map-style arrow dataset.

It also has a `num_shards` argument to split the dataset before converting to an iterable dataset. Sharding is important to enable efficient shuffling and parallel loading of iterable datasets.

TODO:
- [ ] tests
- [ ] docs

Fix https://github.com/huggingface/datasets/issues/5265

PR #5485, opened 2023-01-30
Add section in tutorial for IterableDataset

stevhliu:
Introduces an `IterableDataset` and how to access it in the tutorial section. It also adds a brief next step section at the end to provide a path for users who want more explanation and a path for users who want something more practical and learn how to preprocess these dataset types. It'll complement the awesome new doc introduced in:

- #5410

### DIFF
--- a/docs/source/access.mdx
+++ b/docs/source/access.mdx
@@ -102,16 +102,36 @@ An [`IterableDataset`] is loaded when you set the `streaming` parameter to `True
 
 An [`IterableDataset`] progressively iterates over a dataset one example at a time, so you don't have to wait for the whole dataset to download before you can use it. As you can imagine, this is quite useful for large datasets you want to use immediately!
 
-However, this means an [`IterableDataset`]'s behavior is different from a regular [`Dataset`]. You don't get random access to examples in an [`IterableDataset`]. Instead, you can call `next(iter()` to return the next item from the [`IterableDataset`]:
+However, this means an [`IterableDataset`]'s behavior is different from a regular [`Dataset`]. You don't get random access to examples in an [`IterableDataset`]. Instead, you should iterate over its elements, for example, by calling `next(iter())` or with a `for` loop to return the next item from the [`IterableDataset`]:
 
 ```py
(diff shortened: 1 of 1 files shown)


docs/source/access.mdx
```diff
@@ -78,4 +84,34 @@ Slicing returns a slice - or subset - of the dataset, which is useful for viewin
...
+>>> iterable_dataset = load_dataset("food101", split="train", streaming=True)
+>>> for example in iterable_dataset:
+...     print(example)
+...     break
+{'image': <PIL.JpegImagePlugin.JpegImageFile image mode=RGB size=384x512 at 0x7F0681F5C520>, 'label': 6}
+```
+
+An [`IterableDataset`] progressively iterates over a dataset one example at a time, so you don't have to wait for the whole dataset to download before you can use it. As you can imagine, this is quite useful for large datasets you want to use immediately!
+
+However, this means an [`IterableDataset`]'s behavior is different from a regular [`Dataset`]. You don't get random access to examples in an [`IterableDataset`]. Instead, you can call `next(iter()` to return the next item from the [`IterableDataset`]:
```
albertvillanova, line 105: Nit: strictly speaking, elements can be accessed using the iterable protocol; not necessarily only `next(iter())`, also can be iterated using for loops (like in the example above).
```suggestion
However, this means an [`IterableDataset`]'s behavior is different from a regular [`Dataset`]. You don't get random access to examples in an [`IterableDataset`]. Instead, you should iterate over its elements, for example by calling `next(iter())` to return the next item from the [`IterableDataset`]:
```

albertvillanova (approved):
Thanks, @stevhliu. Great to have this introduction to `IterableDataset`s.

docs/source/access.mdx
lhoestq: I'd also suggest to use a for loop instead if mentioning `next()` here.

You can also show how to get the first 5 examples though:
```python
print(list(iterable_dataset.take(5)))
```

lhoestq (commented):
Nice thank you !

PR #5491, opened 2023-02-01
[MINOR] Typo

cakiki:


src/datasets/download/download_config.py
```diff
@@ -16,7 +16,7 @@ class DownloadConfig:
             If `True`, re-dowload the file even if it's already cached in
             the cache dir.
         resume_download (`bool`, defaults to `False`):
-            If `True`, resume the download if incompletly recieved file is
+            If `True`, resume the download if an incompletly received file is
```
stevhliu, line 19:
```suggestion
            If `True`, resume the download if an incompletely received file is
```

stevhliu (approved):
Thanks for fixing, just noticed another typo in there as well! :)

albertvillanova (approved):
Thanks!

PR #5497, opened 2023-02-02
Improved error message for gated/private repos

osanseviero:
Using `use_auth_token=True` is not needed anymore. If a user logged in, the token will be automatically retrieved. Also include a mention for gated repos

See https://github.com/huggingface/huggingface_hub/pull/1064


julien-c (approved):

lhoestq (approved):
thanks !

PR #5501, opened 2023-02-03
Speeding up file downloads

Narsil:
Original fix: https://github.com/huggingface/huggingface_hub/pull/1267
Not sure this function is actually still called though.

### DIFF
--- a/src/datasets/utils/file_utils.py
+++ b/src/datasets/utils/file_utils.py
@@ -377,7 +377,7 @@ def http_get(
         desc=desc or "Downloading",
         disable=not logging.is_progress_bar_enabled(),
     ) as progress:
-        for chunk in response.iter_content(chunk_size=1024):
+        for chunk in response.iter_content(chunk_size=1024 * 1024):
             progress.update(len(chunk))
             temp_file.write(chunk)
 


lhoestq (approved):

lhoestq:
We simply do GET requests to hf.co to download files from the Hub right now. We may switch to hfh when we update how we do cachinh

src/datasets/utils/file_utils.py
```diff
@@ -377,7 +377,7 @@ def http_get(
         desc=desc or "Downloading",
         disable=not logging.is_progress_bar_enabled(),
     ) as progress:
-        for chunk in response.iter_content(chunk_size=1024):
+        for chunk in response.iter_content(chunk_size=1024 * 1024):
```
albertvillanova, line 380: To be aligned with `huggingface_hub`, shouldn't it be 10 MB instead?
```suggestion
        for chunk in response.iter_content(chunk_size=10 * 1024 * 1024):
```

albertvillanova (changes requested):
Thanks for the improvement, @Narsil.

Just a question below.

src/datasets/utils/file_utils.py
Narsil: We could. In my experiements it made only tiny improvements over 1Mb but you're right let's keep consistent! 

I would like to run benchmarks on this too, before claiming victory, however I'm short on time this week :(

Narsil: To do benchmarks, use a machine that has good network (> 50Mo) otherwise you don't see anything.

On DGX or large AWS instances, I was able to get ~500Mo/s of transfer speed where it was stuck at ~50Mo/s before.

For larger networks, it's impossible to do in Python afaik (but there are still ways to achieve it)

Issue #5425, opened 2023-01-16
Sort on multiple keys with datasets.Dataset.sort()


rocco-fortuna:
### Feature request

From discussion on forum: https://discuss.huggingface.co/t/datasets-dataset-sort-does-not-preserve-ordering/29065/1

`sort()` does not preserve ordering, and it does not support sorting on multiple columns, nor a key function.

The suggested solution: 
> ... having something similar to pandas and be able to specify multiple columns for sorting. We’re already using pandas under the hood to do the sorting in datasets.

The suggested workaround:
> convert your dataset to pandas and use `df.sort_values()`

### Motivation

Preserved ordering when sorting is very handy when one needs to sort on multiple columns, A and B, so that e.g. whenever A is equal for two or more rows, B is kept sorted. 
Having a parameter to do this in 🤗datasets would be cleaner than going through pandas and back, and it wouldn't add much complexity to the library.

Alternatives:
- the possibility to specify multiple keys to sort by with decreasing priority (suggested solution),
- the ability to provide a key function for sorting, so that one can manually specify the sorting criteria.

### Your contribution

I'll be happy to contribute by submitting a PR. Will get documented on `CONTRIBUTING.MD`.
Would love to get thoughts on this, if anyone has anything to add.

MichlF:
@mariosasko If I understand the code right, using `pyarrow.compute.sort_indices` would also require changes to the `select` method if it is meant to sort multiple keys. That's because `select` only accepts 1D input for `indices`, not an iterable or similar which would be required for multiple keys unless you want some looping over selects. Doesn't seem that straight-forward but I might be missing something here...

mariosasko:
@MichlF No, it doesn't require modifying select because sorting on multiple keys also returns a 1D array.

It's easier to understand with an example:
```python
>>> import pyarrow as pa
>>> import pyarrow.compute as pc
>>> table = pa.table({
...     "name": ["John", "Eve", "Peter", "John"],
...     "surname": ["Johnson", "Smith", "Smith", "Doe"],
...     "age": [20, 40, 30, 50],
... })
>>> indices = pc.sort_indices(table, sort_keys=[("name", "ascending"), ("surname", "ascending")])
>>> print(indices)
[
  1,
  3,
  0,
  2
]
```

MichlF:
Thanks for clarifying.
I can prepare a PR to address this issue. This would be my first PR here so I have a few maybe silly questions but:
[...]

PR #5502, opened 2023-02-03
Added functionality: sort datasets by multiple keys

MichlF:
Added functionality implementation: sort datasets by multiple keys/columns as discussed in https://github.com/huggingface/datasets/issues/5425.


src/datasets/arrow_dataset.py
```diff
@@ -3738,33 +3727,33 @@ def _select_with_indices_mapping(
...
         keep_in_memory: bool = False,
         load_from_cache_file: bool = True,
         indices_cache_file_name: Optional[str] = None,
         writer_batch_size: Optional[int] = 1000,
         new_fingerprint: Optional[str] = None,
+        kind="deprecated",
     ) -> "Dataset":
-        """Create a new dataset sorted according to a column.
+        """Create a new dataset sorted according to a single or multiple columns.
+        The transformation is applied to all the datasets of the dataset dictionary.
```
mariosasko, line 3741:
```suggestion
```

src/datasets/arrow_dataset.py
```diff
@@ -3738,33 +3727,33 @@ def _select_with_indices_mapping(
...
 
-        Currently sorting according to a column name uses pandas sorting algorithm under the hood.
-        The column should thus be a pandas compatible type (in particular not a nested type).
+        Currently sorting according to a column name(s) uses pyarrow's sort_indices sorting algorithm under the hood.
+        The column(s) should thus be a pyarrow compatible type (in particular not a nested type).
```
mariosasko, line 3744: I don't think this info is particularly useful, so feel free to remove it (`ArrowNotImplementedError` is thrown in that case anyways)

```suggestion
```

src/datasets/arrow_dataset.py
```diff
@@ -3738,33 +3727,33 @@ def _select_with_indices_mapping(
...
-            null_placement (`str`, defaults to `last`):
-                Put `None` values at the beginning if first; last puts `None` values at the end.
+            column_names (`Union[str, Sequence[str]]`):
+                Either a single column name or a list of unqiue column names to sort by.
+            reverse (`Union[bool, Sequence[bool]]`, defaults to `False`):
+                If `True`, sort by descending order rather than ascending. If a single bool is provided,
+                the value is applied to the sorting of all column names. Otherwise a list of bools with the
+                same length and order as column_names must be provided.
+            null_placement (`str`, defaults to `at_end`):
+                Put `None` values at the beginning if `at_start` or `first`; `at_end` or `last` puts `None`
```
mariosasko, line 3755:
```suggestion
                Put `None` values at the beginning if `at_start` or `first` or at the end  if `at_end` or `last` 
```

src/datasets/arrow_dataset.py
```diff
@@ -3738,33 +3727,33 @@ def _select_with_indices_mapping(
...
-            reverse (`bool`, defaults to `False`):
-                If `True`, sort by descending order rather then ascending.
-            kind (`str`, *optional*):
-                Pandas algorithm for sorting selected in `{quicksort, mergesort, heapsort, stable}`,
-                The default is `quicksort`. Note that both `stable` and `mergesort` use `timsort` under the covers and, in general,
-                the actual implementation will vary with data type. The `mergesort` option is retained for backwards compatibility.
-            null_placement (`str`, defaults to `last`):
-                Put `None` values at the beginning if first; last puts `None` values at the end.
+            column_names (`Union[str, Sequence[str]]`):
+                Either a single column name or a list of unqiue column names to sort by.
```
mariosasko, line 3749:
```suggestion
               Column name(s) to sort by.
```

src/datasets/arrow_dataset.py
```diff
@@ -3781,17 +3770,22 @@ def sort(
             new_fingerprint (`str`, *optional*, defaults to `None`):
                 The new fingerprint of the dataset after transform.
                 If `None`, the new fingerprint is computed using a hash of the previous fingerprint, and the transform arguments
+            kind (*deprecated*):
+                Parameter is deprecated and will be removed in future releases.
```
mariosasko, line 3774: Let's preserve the original docstring (and position) of this param and append:
```
<Deprecated version="2.8.0">

`kind` was deprecated in version 2.10.0 and will be removed in 3.0.0.

</Deprecated>
```

src/datasets/arrow_dataset.py
```diff
@@ -3802,12 +3796,48 @@ def sort(
...
             return self
 
-        # Check the column name
-        if not isinstance(column, str) or column not in self._data.column_names:
+        # Deprecation warning
+        if kind != "deprecated":
+            warnings.warn(
+                "The 'kind' parameter is deprecated and will be removed in future releases.",
+                category=DeprecationWarning,
+            )
```
mariosasko, line 3804: We use `FutureWarning` to notify users about deprecations

```suggestion
        if kind != "deprecated":
            warnings.warn(
                "`kind` was deprecated in version 2.10.0 and will be removed in 3.0.0.",
                category=FutureWarning,
            )
```

src/datasets/arrow_dataset.py
```diff
@@ -3802,12 +3796,48 @@ def sort(
...
+
+        # Check proper format of and for duplicates in column_names
+        if not isinstance(column_names, list):
+            column_names = [column_names]
+        if not len(column_names) == len(set(column_names)):
             raise ValueError(
-                f"Column '{column}' not found in the dataset. Please provide a column selected in: {self._data.column_names}"
+                "At least one of the column names in column_names occurs more than once. Ensure each column name is unique."
```
mariosasko, line 3811: I think we can remove this check and delegate it to `sort_indices`

src/datasets/arrow_dataset.py
```diff
@@ -3819,14 +3849,19 @@ def sort(
                     fingerprint=new_fingerprint, indices_cache_file_name=indices_cache_file_name
                 )
 
-        column_data = self._getitem(
-            column, format_type="pandas", format_columns=None, output_all_columns=False, format_kwargs=None
+        queried_dataset = query_table(
+            table=self._data,
+            key=slice(0, len(self)),
+            indices=self._indices if self._indices is not None else None,
```
mariosasko, line 3855: (Nit)

```suggestion
        sort_table = query_table(
            table=self._data,
            key=range(self._data.num_rows),
            indices=self._indices if self._indices is not None else None,
```

src/datasets/arrow_dataset.py
```diff
@@ -3819,14 +3849,19 @@ def sort(
...
         )
 
-        df_sorted = column_data.to_frame().sort_values(
-            column, ascending=not reverse, kind=kind, na_position=null_placement
-        )
-        indices = df_sorted.index.to_numpy()
+        sort_keys = [
+            (column_names[i], "ascending" if not reverse[i] else "descending") for i in range(len(column_names))
+        ]
```
mariosasko, line 3860: It's more pythonic to use `zip` here:
```suggestion
        sort_keys = [
            (col, "ascending" if not col_reverse else "descending") for col, col_reverse in zip(column_names, reverse)
        ]
```

src/datasets/arrow_dataset.py
```diff
@@ -3819,14 +3849,19 @@ def sort(
...
+
+        indices = np.array(
+            pa.compute.sort_indices(queried_dataset, sort_keys=sort_keys, null_placement=null_placement)
+        ).astype(np.int64)
```
mariosasko, line 3864: We can avoid the NumPy conversion here and modify select slightly to support PyArrow arrays (call `to_numpy` on the indices if they are of type `pa.Array`/`pa.ChunkedArray`):

```suggestion
        indices = pc.sort_indices(sort_table, sort_keys=sort_keys, null_placement=null_placement)
```

mariosasko (commented):
Thanks! I've left some comments.

We should also add some tests, mainly to make sure `reverse` behaves as expected. Let me know if you need help with that.

src/datasets/arrow_dataset.py
MichlF: Could you elaborate on what you mean by "_delegate it to `sort_indices`_"?

MichlF:
> Thanks! I've left some comments.
> 
> We should also add some tests, mainly to make sure `reverse` behaves as expected. Let me know if you need help with that.

Thanks for the offer! I couldn't find any guidelines on how huggingface goes about testing, so it would indeed be great to get a few pointers on that. I assume I should expand on the `test_sort` function in `test_arrow_dataset.py` but since I am not very familiar with the `datasets` package, it isn't immediately for which cases I should test (i.e., expand on).

src/datasets/arrow_dataset.py
mariosasko: I don't think it's possible for `pc.sort_indices` to throw this error. Can you paste the full error stack trace?

mariosasko:
@MichlF 

Resolving a comment means that the comment has been addressed with the code change, so since this is not the case here, can you please "unresolve" the comments and address them adequately? 

> I assume I should expand on the `test_sort` function in `test_arrow_dataset.py`

Yes, that's correct. I think one test to check sorting on multiple keys and another one to check if an error is raised when `len(reverse)!=len(column_names)` should be enough.

src/datasets/arrow_dataset.py
MichlF: I didn't write that this error is thrown by `pc.sort_indices`. 
This was just to illustrate that calling sort in this way should be caught/prevented. The error occurs in `table`'s `_interpolation_search`.

Full error stack:
```
  File "c:\Users\miche\Documents\GitHub\test\testing.py", line 24, in <module>
    another_sorted_ds = ds.sort(["label", "label"], reverse=[True, False])
  File "C:\Users\miche\anaconda3\envs\hugging_face\lib\site-packages\datasets\arrow_dataset.py", line 514, in wrapper
  File "C:\Users\miche\anaconda3\envs\hugging_face\lib\site-packages\datasets\fingerprint.py", line 480, in wrapper
  File "C:\Users\miche\anaconda3\envs\hugging_face\lib\site-packages\datasets\arrow_dataset.py", line 3971, in sort
    return self.select(
  File "C:\Users\miche\anaconda3\envs\hugging_face\lib\site-packages\datasets\arrow_dataset.py", line 514, in wrapper
  File "C:\Users\miche\anaconda3\envs\hugging_face\lib\site-packages\datasets\fingerprint.py", line 480, in wrapper
  File "C:\Users\miche\anaconda3\envs\hugging_face\lib\site-packages\datasets\arrow_dataset.py", line 3563, in select
    return self._select_contiguous(start, length, new_fingerprint=new_fingerprint)
  File "C:\Users\miche\anaconda3\envs\hugging_face\lib\site-packages\datasets\arrow_dataset.py", line 514, in wrapper
  File "C:\Users\miche\anaconda3\envs\hugging_face\lib\site-packages\datasets\fingerprint.py", line 480, in wrapper
  File "C:\Users\miche\anaconda3\envs\hugging_face\lib\site-packages\datasets\arrow_dataset.py", line 3616, in _select_contiguous
    self.data.slice(start, length),
  File "C:\Users\miche\anaconda3\envs\hugging_face\lib\site-packages\datasets\table.py", line 1113, in slice
    return MemoryMappedTable(self.fast_slice(offset=offset, length=length), self.path, replays)
  File "C:\Users\miche\anaconda3\envs\hugging_face\lib\site-packages\datasets\table.py", line 139, in fast_slice
    i = _interpolation_search(self._offsets, offset)
  File "C:\Users\miche\anaconda3\envs\hugging_face\lib\site-packages\datasets\table.py", line 93, in _interpolation_search
    if arr[k] <= x < arr[k + 1]:
IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices
```

src/datasets/arrow_dataset.py
```diff
@@ -3738,33 +3727,33 @@ def _select_with_indices_mapping(
...
-        null_placement: str = "last",
+        column_names: Union[str, Sequence_[str]],
+        reverse: Union[bool, Sequence_[bool]] = False,
+        null_placement: str = "at_end",
         keep_in_memory: bool = False,
         load_from_cache_file: bool = True,
         indices_cache_file_name: Optional[str] = None,
         writer_batch_size: Optional[int] = 1000,
         new_fingerprint: Optional[str] = None,
+        kind="deprecated",
```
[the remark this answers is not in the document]
MichlF: With https://github.com/huggingface/datasets/pull/5502/commits/53c0a537b5c5c55f5a2aa72715b39175cc690bf4: reverted back to the original order of params in the signature

src/datasets/arrow_dataset.py
MichlF: Addressed separately in https://github.com/huggingface/datasets/pull/5502/commits/53c0a537b5c5c55f5a2aa72715b39175cc690bf4 because of minor typo

src/datasets/arrow_dataset.py
MichlF: Check removed in https://github.com/huggingface/datasets/pull/5502/commits/53c0a537b5c5c55f5a2aa72715b39175cc690bf4

MichlF:
> Yes, that's correct. I think one test to check sorting on multiple keys and another one to check if an error is raised when `len(reverse)!=len(column_names)` should be enough.

I have added the tests in https://github.com/huggingface/datasets/pull/5502/commits/0efa259732e822e94d67b96a70031a3daccedfc1 by keeping them in the same format of the tests of the old `sort` function. Let me know if they can be improved.

src/datasets/arrow_dataset.py
MichlF: Addressed in https://github.com/huggingface/datasets/pull/5502/commits/aabc95c4c4cfb3020d1356a018a3059a2914b0b1

src/datasets/arrow_dataset.py
```diff
@@ -3806,33 +3811,36 @@ def _select_with_indices_mapping(
...
-        null_placement: str = "last",
+        column_names: Union[str, Sequence_[str]],
+        reverse: Union[bool, Sequence_[bool]] = False,
+        kind="deprecated",
+        null_placement: str = "at_end",
         keep_in_memory: bool = False,
         load_from_cache_file: Optional[bool] = None,
         indices_cache_file_name: Optional[str] = None,
         writer_batch_size: Optional[int] = 1000,
         new_fingerprint: Optional[str] = None,
```
lhoestq, line 3822: Maybe keep `column=` as a deprecated keyword argument as well ?

mariosasko: For `dataset.sort(column="col")` to keep working, we also need to define a default value for `column_names`, so not sure about this.

mariosasko: Yes, but `column_names` is a **positional-only** argument, so it needs a default value, or `sort` would throw the `missing argument error` when one calls it as `dataset.sort(column="col")`. And this makes the code a bit messy IMO.

lhoestq: I checked and couldn't find documentation or script passing `column=` so I think it's fine

lhoestq (approved):
Looks all good ! Thanks :)

mariosasko (approved):
LGTM!

PR #5509, opened 2023-02-07
Add a static `__all__` to `__init__.py` for typecheckers

LoicGrobol:
This adds a static `__all__` field to `__init__.py`, allowing typecheckers to know which symbols are accessible from `datasets` at runtime. In particular [Pyright](https://github.com/microsoft/pylance-release/issues/2328#issuecomment-1029381258) seems to rely on this. At this point I have added all (modulo oversight) the symbols mentioned in the Reference part of [the docs](https://huggingface.co/docs/datasets), but that could be adjusted. As a side effect, only these symbols will be imported by `from datasets import *`, which may or may not be a good thing (and if it isn't, that's easy to fix).

This should fix #[3841](https://github.com/huggingface/datasets/issues/3841). Another option would be to add a pyi stub.


mariosasko:
I've commented on the original issue to provide some context. Feel free to share your opinion there.

PR #5515, opened 2023-02-09
Unify `load_from_cache_file` type and logic

HallerPatrick:
* Updating type annotations for  #`load_from_cache_file`
* Added logic for cache checking if needed
* Updated documentation following the wording of `Dataset.map`


mariosasko (commented):
Good job, thanks! `DatasetDict`'s methods also need to be updated with the new default value, and then we can merge.

HallerPatrick:
The commit also includes the changes to the `DatasetDict` methods or am I missing something?

mariosasko:
Oh, indeed. Feel free to mark the PR as "Ready for review" then.

mariosasko (approved):
Nice, lgtm!

src/datasets/arrow_dataset.py
```diff
@@ -2960,7 +2960,7 @@ def _map_single(
         drop_last_batch: bool = False,
         remove_columns: Optional[List[str]] = None,
         keep_in_memory: bool = False,
-        load_from_cache_file: bool = None,
+        load_from_cache_file: Optional[bool] = None,
```
lhoestq, line 2963: `_map_single` expected a bool, so you can add this line at the beginning of the function:
```python
load_from_cache_file = load_from_cache_file if load_from_cache_file is not None else is_caching_enabled()
```
otherwise by default it would not return the cached data

lhoestq (commented):
Thanks ! Just one comment:

PR #5516, opened 2023-02-09
Reload features from Parquet metadata

### LINKED ISSUE #5482: Reload features from Parquet metadata
The idea would be to allow this :
```python 
ds.to_parquet("my_dataset/ds.parquet")
reloaded = load_dataset("my_dataset")
assert ds.features == reloaded.features 
```
And it should also work with Image and Audio types (right now they're reloaded as a dict type)

This can be implemented by storing and reading the feature types in the parquet metadata, as we do for arrow files.

MFreidank:
Resolves #5482. 

Attaches feature metadata to parquet files serialised using `Dataset.to_parquet`. 

This allows retrieving data with "rich" feature types (e.g., `datasets.features.image.Image` or `datasets.features.audio.Audio`) from parquet files without cumbersome casting (for an example, see #5482).

@lhoestq It seems that it is sufficient to attach metadata to the schema prior to serialising and features are loaded back with correct types afterwards automatically. 

I used the following script to test the implementation:

```python
from pathlib import Path

import datasets

dataset_name = "Maysee/tiny-imagenet"
ds = datasets.load_dataset(dataset_name, split=datasets.Split.TRAIN)
output_directory_path = Path(__file__).parent.joinpath("example_test_outputs", dataset_name.replace("/", "_"))
output_directory_path.mkdir(exist_ok=True, parents=True)
output_filepath = output_directory_path.joinpath("ds.parquet")
ds.to_parquet(str(output_filepath))

reloaded_ds = datasets.load_dataset(str(output_directory_path), split=datasets.Split.TRAIN)

assert ds.features == reloaded_ds.features
```

Prior to the change in this PR this script raises an `AssertionError` and the `Image` features lose their type after serialisation. After the change in this PR, the assertion does not raise an error and manual inspection of the features shows type `Image` for the respective columns of `reloaded_ds `.

Some open questions:
* How/where can I best add new unit tests for this implementation?
* What dataset would I best use in the tests? I chose `Maysee/tiny-imagenet` mainly because it is small and contains an ?Image` feature that can be used to test, but I'd be happy for suggestions on a suitable data source to use.
* Currently I'm calling `datasets.arrow_writer.ArrowWriter._build_metadata` as I need the same logic. However, I'm not happy with the coupling between `datasets.io.parquet` and `datasets.arrow_writer` it leaves me with. Suggest to factor this common logic out into a helper function and reuse it from both of these. Do you agree and if yes, could you please guide me where I would best place this function?

Many thanks in advance and kind regards,
MFreidank

### DIFF
--- a/src/datasets/io/parquet.py
+++ b/src/datasets/io/parquet.py
@@ -1,17 +1,14 @@
-import json
 import os
-from typing import BinaryIO, Dict, Optional, Union
+from typing import BinaryIO, Optional, Union
 
 import pyarrow as pa
 import pyarrow.parquet as pq
 
 from .. import Dataset, Features, NamedSplit, config
 from ..formatting import query_table
-from ..info import DatasetInfo
 from ..packaged_modules import _PACKAGED_DATASETS_MODULES
 from ..packaged_modules.parquet.parquet import Parquet
 from ..utils import logging
-from ..utils.py_utils import asdict
 from ..utils.typing import NestedDataStructureLike, PathLike
 from .abc import AbstractDatasetReader
 
@@ -95,16 +92,6 @@ class ParquetDatasetWriter:
         else:
             written = self._write(file_obj=self.path_or_buf, batch_size=batch_size, **self.parquet_writer_kwargs)
         return written
-
-    @staticmethod
-    def _build_metadata(info: DatasetInfo, fingerprint: Optional[str] = None) -> Dict[str, str]:
-        info_keys = ["features"]  # we can add support for more DatasetInfo keys in the future
-        info_as_dict = asdict(info)
-        metadata = {}
-        metadata["info"] = {key: info_as_dict[key] for key in info_keys}
-        if fingerprint is not None:
-            metadata["fingerprint"] = fingerprint
-        return {"huggingface": json.dumps(metadata)}
 
     def _write(self, file_obj: BinaryIO, batch_size: int, **parquet_writer_kwargs) -> int:
         """Writes the pyarrow table as Parquet to a binary file handle.
--- a/src/datasets/io/parquet.py
+++ b/src/datasets/io/parquet.py
@@ -1,7 +1,6 @@
 import os
 from typing import BinaryIO, Optional, Union
 
-import pyarrow as pa
 import pyarrow.parquet as pq
 
 from .. import Dataset, Features, NamedSplit, config
--- a/src/datasets/io/parquet.py
+++ b/src/datasets/io/parquet.py
@@ -1,14 +1,13 @@
-import os
 import json
+import os
 from typing import BinaryIO, Dict, Optional, Union
 
 import pyarrow as pa
 import pyarrow.parquet as pq
 
 from .. import Dataset, Features, NamedSplit, config
-from ..arrow_writer import ArrowWriter
-from ..info import DatasetInfo
 from ..formatting import query_table
+from ..info import DatasetInfo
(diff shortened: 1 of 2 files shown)


src/datasets/io/parquet.py
```diff
@@ -101,6 +115,13 @@ def _write(self, file_obj: BinaryIO, batch_size: int, **parquet_writer_kwargs) -
         written = 0
         _ = parquet_writer_kwargs.pop("path_or_buf", None)
         schema = pa.schema(self.dataset.features.type)
+
+        schema = schema.with_metadata(
+            ArrowWriter._build_metadata(
+                info=DatasetInfo(features=self.dataset.features)
+            )
+        )
+
```
lhoestq, line 124: this way you already get a schema with metadata :)
```suggestion
        schema = self.dataset.features.arrow_schema
```

lhoestq (commented):
Awesome thank you !

> @lhoestq It seems that it is sufficient to attach metadata to the schema prior to serialising and features are loaded back with correct types afterwards automatically.

Indeed good catch ! Since we keep the schema with its metadata all along to the `ArrowWriter` it works.

> How/where can I best add new unit tests for this implementation?

I think we could have a test in `tests/io/test_parquet.py`

> What dataset would I best use in the tests?

You can get a test dataset using
```python

@require_pil
def test_dataset_to_parquet_keeps_features(shared_datadir):
    import PIL.Image

    image_path = str(shared_datadir / "test_image_rgb.jpg")
    data = {"image": [image_path]}
    features = Features({"image": Image()})
    dset = Dataset.from_dict(data, features=features)
    # TODO: add test here
```

> Do you agree and if yes, could you please guide me where I would best place this function?

There is no need for that if you use `features.arrow_schema`; it returns the schema with metadata ;)

src/datasets/io/parquet.py
MFreidank: ah, wow, that's much simpler :)

lhoestq:
Cool ! I think you just need to remove the unused import in `io/parquet.py`
```
src/datasets/io/parquet.py:4:1: F401 'pyarrow as pa' imported but unused
```
and we're good to merge :)

MFreidank:
> Cool ! I think you just need to remove the unused import in `io/parquet.py`
> 
> ```
> src/datasets/io/parquet.py:4:1: F401 'pyarrow as pa' imported but unused
> ```
> 
> and we're good to merge :)

done!

tests/io/test_parquet.py
```diff
@@ -130,3 +132,20 @@ def test_parquer_write(dataset, tmp_path):
...
     output_table = pf.read()
     assert dataset.data.table == output_table
+
+
+@require_pil
+def test_dataset_to_parquet_keeps_features(shared_datadir, tmp_path):
+    import PIL.Image
+
+    image_path = str(shared_datadir / "test_image_rgb.jpg")
+    PIL.Image.fromarray(np.zeros((5, 5), dtype=np.uint8)).save(image_path, format="png")
```
lhoestq, line 142: Oh actually you can also remove this line, the file already exists and is a dummy image we use for tests

MFreidank: Done! The test actually does not use `PIL` anymore after that so I also dropped the `require_pil`.

lhoestq: `require_pil` is actually required to use image datasets, so you can re-add it ^^

MFreidank: re-added ^^ sorry about that

lhoestq (approved):
Thanks !

PR #5522, opened 2023-02-10
Minor changes in JAX-formatting docstrings & type-hints

alvarobartt:
Hi to whoever is reading this! 🤗

## What's in this PR?

I was exploring the code regarding the `JaxFormatter` implemented in 🤗`datasets`, and found some things that IMO could be changed. Those are mainly regarding the docstrings and the type-hints based on `jax`'s 0.4.1 release where `jax.Array` was introduced as the default type for JAX-arrays (instead of `jnp.DeviceArray`, `jnp.SharedDeviceArray`, and `jnp.GlobalDeviceArray`). Even though `isinstance(..., jax.Array)` also works with lower versions such as e.g. `0.3.25`.

More information about the latter at [`jax` v0.4.1 - Release Notes](https://github.com/google/jax/releases/tag/jax-v0.4.1) and [jax.Array migration - JAX documentation](https://jax.readthedocs.io/en/latest/jax_array_migration.html).

## What's missing?

* Do you want me to write an entry in the documentation on how to use 🤗`datasets` with JAX as https://huggingface.co/docs/datasets/use_with_pytorch with PyTorch?
* Do we need to actually include `pyarrow` under the `TYPE_CHECKING` when needed? I just did it for JAX, but if we are OK with that, I can do that with the rest of the formatters, just LMK.
* Should the License header be included in `datasets.formatting.np_formatter`? If so, do I include the one from 2020 e.g. https://github.com/huggingface/datasets/blob/b065547654efa0ec633cf373ac1512884c68b2e1/src/datasets/formatting/tf_formatter.py#L1-L13
* Is there any reason why `jnp.array` is being used instead of `jnp.asarray`? There's no difference between both, just that `jnp.asarray` has `copy=False` as default, even though `numpy` to `jax.numpy` conversion is not zero-copy, but just asking :)

### DIFF
--- a/src/datasets/formatting/jax_formatter.py
+++ b/src/datasets/formatting/jax_formatter.py
@@ -41,8 +41,7 @@ class JaxFormatter(Formatter[Mapping, "jax.Array", Mapping]):
 
         if isinstance(column, list) and column:
             if all(
-                isinstance(x, jax.Array) and x.shape == column[0].shape and x.dtype == column[0].dtype
-                for x in column
+                isinstance(x, jax.Array) and x.shape == column[0].shape and x.dtype == column[0].dtype for x in column
             ):
                 return jnp.stack(column, axis=0)
         return column
--- a/src/datasets/formatting/jax_formatter.py
+++ b/src/datasets/formatting/jax_formatter.py
@@ -18,6 +18,7 @@ from collections.abc import Mapping
 from typing import TYPE_CHECKING
 
 import numpy as np
+import pyarrow as pa
 
 from .. import config
 from ..utils.py_utils import map_nested
@@ -26,7 +27,6 @@ from .formatting import Formatter
 
 if TYPE_CHECKING:
     import jax
-    import pyarrow as pa
 
 
 class JaxFormatter(Formatter[Mapping, "jax.Array", Mapping]):
@@ -86,19 +86,19 @@ class JaxFormatter(Formatter[Mapping, "jax.Array", Mapping]):
     def recursive_tensorize(self, data_struct: dict):
         return map_nested(self._recursive_tensorize, data_struct)
 
-    def format_row(self, pa_table: "pa.Table") -> Mapping:
+    def format_row(self, pa_table: pa.Table) -> Mapping:
         row = self.numpy_arrow_extractor().extract_row(pa_table)
         row = self.python_features_decoder.decode_row(row)
         return self.recursive_tensorize(row)
 
-    def format_column(self, pa_table: "pa.Table") -> "jax.Array":
+    def format_column(self, pa_table: pa.Table) -> "jax.Array":
         column = self.numpy_arrow_extractor().extract_column(pa_table)
         column = self.python_features_decoder.decode_column(column, pa_table.column_names[0])
         column = self.recursive_tensorize(column)
         column = self._consolidate(column)
         return column
 
-    def format_batch(self, pa_table: "pa.Table") -> Mapping:
+    def format_batch(self, pa_table: pa.Table) -> Mapping:
         batch = self.numpy_arrow_extractor().extract_batch(pa_table)
         batch = self.python_features_decoder.decode_batch(batch)
         batch = self.recursive_tensorize(batch)
--- a/src/datasets/formatting/jax_formatter.py
+++ b/src/datasets/formatting/jax_formatter.py
@@ -43,7 +43,7 @@ class JaxFormatter(Formatter[Mapping, "jnp.ndarray", Mapping]):
                 isinstance(x, jnp.ndarray) and x.shape == column[0].shape and x.dtype == column[0].dtype
                 for x in column
             ):
-                return jnp.stack(column)
+                return jnp.stack(column, axis=0)
         return column
 
     def _tensorize(self, value):
--- a/src/datasets/formatting/jax_formatter.py
+++ b/src/datasets/formatting/jax_formatter.py
@@ -18,7 +18,6 @@ from collections.abc import Mapping
 from typing import TYPE_CHECKING
 
 import numpy as np
-import pyarrow as pa
 
 from .. import config
 from ..utils.py_utils import map_nested
@@ -27,6 +26,7 @@ from .formatting import Formatter
 
 if TYPE_CHECKING:
     import jax.numpy as jnp
+    import pyarrow as pa
 
 
 class JaxFormatter(Formatter[Mapping, "jnp.ndarray", Mapping]):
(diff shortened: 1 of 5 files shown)


alvarobartt:
P.S. For more context, I'm currently exploring the integration of 🤗`datasets` with JAX, so in case you need any help or want me to try something specific just let me know! (`jnp.asarray`/`jnp.array(..., copy=False)` still no zero-copy 😭)

src/datasets/formatting/jax_formatter.py
```diff
@@ -18,32 +18,32 @@
...
 
     def _consolidate(self, column):
+        import jax
         import jax.numpy as jnp
 
         if isinstance(column, list) and column:
             if all(
-                isinstance(x, jnp.ndarray) and x.shape == column[0].shape and x.dtype == column[0].dtype
-                for x in column
+                isinstance(x, jax.Array) and x.shape == column[0].shape and x.dtype == column[0].dtype for x in column
```
lhoestq, line 44: Does this work with older version of jax ?
If not, we may need to specify `jax>=0.4.1` in `setup.py`

lhoestq (commented):
Hi ! Thanks for improving this :)

src/datasets/formatting/jax_formatter.py
alvarobartt: > Hi ! Thanks for improving this :)

Glad to help, @lhoestq! Also, what's your input in the `## What's missing in this PR` section in the PR description above?

alvarobartt: We should specify the `jax` and `jaxlib` versions! Nice catch!

alvarobartt:
> Hi ! Thanks for improving this :)

Glad to help, @lhoestq! Also, regarding the questions in the `## What's missing in this PR?` can I have your input? Thanks 🤗

lhoestq:
Whoops forgot to reply to these matters - sorry x)

Yea a JAX guide would be welcome in the documentation ! This can be done in a separate PR if you want :)

Pyarrow is always imported with `datasets`, so it doesn't really matter if it's under TYPE_CHECKING or not.

Regarding the license : yes indeed it should be in every file, thanks for reporting.

No big preference between jnp.array and jnp.asarray, unless one offers better performance

alvarobartt:
> Whoops forgot to reply to these matters - sorry x)
> 
> Yea a JAX guide would be welcome in the documentation ! This can be done in a separate PR if you want :)
> 
> Pyarrow is always imported with `datasets`, so it doesn't really matter if it's under TYPE_CHECKING or not.
> 
> Regarding the license : yes indeed it should be in every file, thanks for reporting.
> 
> No big preference between jnp.array and jnp.asarray, unless one offers better performance

Cool @lhoestq thanks for the input there!

1. I can create a separate PR for JAX-format usage
2. Regarding that, makes sense, we can just not put it there, unless it's more clear that in that file `pyarrow` is just required for typing?
3. Do you want me to add the License? In this PR? In a separate one?
4. Ideally `jnp.asarray` is similar to `np.asarray` which in the case of `numpy` tends to be more efficient as it does zero-copy when possible, while `np.array` has `copy=True` by default, anyway as I mentioned before (and as you already know) the copy from `numpy` to `jax` is not zero-copy, while the other way around (`jax` to `numpy`) it is

lhoestq:
Thanks, feel free to create separate PRs for the docs and the license.

I guess you can move the `pyarrow` import back to where it was for consistency with the other files and we can merge this one ;)

alvarobartt:
> Thanks, feel free to create separate PRs for the docs and the license.
> 
> I guess you can move the `pyarrow` import back to where it was for consistency with the other files and we can merge this one ;)

Cool thanks I'll do that! 👍🏻

lhoestq:
Actually I just checked and there are still tens of thousands of users with jax 0.3.25 - so we need to support older versions as well. I guess it comes from `transformers` which doesn't support jax 0.4 (and doesn't want to until the jax team stops breaking the lib all the time).

Could you make sure your changes work with older versions as well ? Sorry for not spotting this earlier.
If we have `"jax>=0.2.8,!=0.3.2,<=0.4.3"` that'b be nice, and we can update the latest supported release from time to time.

In the CI you can add `jax==0.2.8` for the `deps-minimum` job, and use `jax~=0.4.1` for the `deps-latest`.

alvarobartt:
> Actually I just checked and there are still tens of thousands of users with jax 0.3.25 - so we need to support older versions as well. I guess it comes from `transformers` which doesn't support jax 0.4 (and doesn't want to until the jax team stops breaking the lib all the time).
> 
> Could you make sure your changes work with older versions as well ? Sorry for not spotting this earlier. If we have `"jax>=0.2.8,!=0.3.2,<=0.4.3"` that'b be nice, and we can update the latest supported release from time to time.
> 
> In the CI you can add `jax==0.2.8` for the `deps-minimum` job, and use `jax~=0.4.1` for the `deps-latest`.

Ok, didn't know that @lhoestq thanks for the detailed context! Sure, I'll update it and make sure it's also compatible with older versions.

alvarobartt:
Oops forgot to add you as co-author of the last commit @lhoestq my bad 😞

alvarobartt:
So it should be fixed right now @lhoestq! The thing is that `jax` doesn't provide support for Python 3.7 due to its EOL next June (more information at https://endoflife.date/python)...

Anyway, I can confirm that `jax.Array` type works with 0.3.25 and that the following code works fine:

```python
import jax
import jax.numpy as jnp

x = jnp.ones((1, 10), dtype=jnp.float32)  # Is a `jnp.DeviceArray`
assert isinstance(x, jax.Array)  # Is `True`
```

So we can still use 0.3.25 as the maximum supported version, as well as 0.3.6 for `jaxlib` so as to be consistent with 🤗`transformers`.

Thanks for your comments @lhoestq those were really useful!

alvarobartt:
Sorry for the spam, pinning versions leads to failure runs (not related to the type-hinting); I'll check that locally instead of here to avoid spam... Not pinning the dependencies work but I'll check the minimum required versions for both `jax` and `jaxlib` in Python 3.7

lhoestq (approved):
Cool ! Thanks for trying to make the CI support it, but it's maybe not worth spending more time on this for now ^^

merging :)

alvarobartt:
> Cool ! Thanks for trying to make the CI support it, but it's maybe not worth spending more time on this for now ^^
> 
> merging :)

Do you want me to work on the CI in a separate branch? Thanks for merging and for your help as always :)

lhoestq:
> Do you want me to work on the CI in a separate branch? Thanks for merging and for your help as always :)

In the end I think we can keep it as is since we didn't modify the core code for jax. Maybe later if we do further changes and need to make sure we don't break anything ;) For example when we decide to add support for more recent versions

alvarobartt:
> > Do you want me to work on the CI in a separate branch? Thanks for merging and for your help as always :)
> 
> In the end I think we can keep it as is since we didn't modify the core code for jax. Maybe later if we do further changes and need to make sure we don't break anything ;) For example when we decide to add support for more recent versions

Makes sense, thank you @lhoestq!

PR #5526, opened 2023-02-10
Allow loading/saving of FAISS index using fsspec

### LINKED ISSUE #5428: Load FAISS index from stream
### Feature request

From what I understand `faiss` already support this [link](https://github.com/facebookresearch/faiss/wiki/Index-IO,-cloning-and-hyper-parameter-tuning#generic-io-support)

I would like if we could use a stream as input to `Dataset.load_faiss_index` and `Dataset.save_faiss_index`.

### Motivation

In my case, I'm saving faiss index in cloud storage and use `fsspec` to load them. It would be ideal if I could send the stream directly instead of copying the file locally (or mounting the bucket) and then load the index.

### Your contribution

I can submit the PR no problem.

Dref360:
Fixes #5428

Allow loading/saving of FAISS index using fsspec:
1. Simply use BufferedIOWriter/Reader to Read/Write indices on fsspec stream.
2. Needed `mockfs` in the test, so I took it out of the `TestCase`. Let me know if that makes sense.

I can work on the documentation once the code changes are approved.


src/datasets/search.py
```diff
@@ -384,20 +385,23 @@ def save(self, file: Union[str, PurePath]):
         else:
             index = self.faiss_index
 
-        faiss.write_index(index, str(file))
+        with fsspec.open(str(file), "wb", **(storage_options or {})) as f:
+            index = faiss.write_index(index, faiss.BufferedIOWriter(faiss.PyCallbackIOWriter(f.write)))
```
alvarobartt, line 389: There's no need to capture `index` here, right?

```suggestion
            faiss.write_index(index, faiss.BufferedIOWriter(faiss.PyCallbackIOWriter(f.write)))
```

tests/test_search.py
```diff
@@ -145,26 +145,24 @@ def test_custom(self):
...
-            index = FaissIndex.load(tmp_file.name)
-        os.unlink(tmp_file.name)
 
-        query = np.zeros(5, dtype=np.float32)
-        query[1] = 1
-        scores, indices = index.search(query)
-        self.assertGreater(scores[0], 0)
-        self.assertEqual(indices[0], 1)
+@require_faiss
+def test_serialization(mockfs):
```
alvarobartt, line 150: IMO the function above should be kept to still run the test to make sure that local storage works, while you can rename this to something like:

```suggestion
def test_serialization_fs(mockfs):
```

To be compliant with other functions like e.g. https://github.com/huggingface/datasets/blob/cc637d107ef3e3b9948691379312a8099b6476aa/tests/test_arrow_dataset.py#L3655

alvarobartt (commented):
Just checked this and left some comments, hope you don't mind... Good work 💪🏻

Dref360:
Thanks for the quick review! I updated the code with your suggestion

src/datasets/search.py
```diff
@@ -520,24 +524,27 @@ def add_faiss_index_from_external_arrays(
...
         self._indexes[index_name] = faiss_index
 
-    def save_faiss_index(self, index_name: str, file: Union[str, PurePath]):
+    def save_faiss_index(self, index_name: str, file: Union[str, PurePath], storage_options: Optional[Dict] = None):
         """Save a FaissIndex on disk.
 
         Args:
             index_name (`str`): The index_name/identifier of the index. This is the index_name that is used to call `.get_nearest` or `.search`.
-            file (`str`): The path to the serialized faiss index on disk.
+            file (`str`): The path to the serialized faiss index on disk or remote URI (e.g. `"s3//my-bucket/index.faiss"`).
```
albertvillanova, line 532: Typo:
```suggestion
            file (`str`): The path to the serialized faiss index on disk or remote URI (e.g. `"s3://my-bucket/index.faiss"`).
```

src/datasets/search.py
```diff
@@ -547,11 +554,14 @@ def load_faiss_index(
         Args:
             index_name (`str`): The index_name/identifier of the index. This is the index_name that is used to
                 call `.get_nearest` or `.search`.
-            file (`str`): The path to the serialized faiss index on disk.
+            file (`str`): The path to the serialized faiss index on disk or remote URI (e.g. `"s3//my-bucket/index.faiss"`).
```
albertvillanova, line 557: As above:
```suggestion
            file (`str`): The path to the serialized faiss index on disk or remote URI (e.g. `"s3://my-bucket/index.faiss"`).
```

src/datasets/search.py
```diff
@@ -520,24 +524,27 @@ def add_faiss_index_from_external_arrays(
...
-    def save_faiss_index(self, index_name: str, file: Union[str, PurePath]):
+    def save_faiss_index(self, index_name: str, file: Union[str, PurePath], storage_options: Optional[Dict] = None):
         """Save a FaissIndex on disk.
 
         Args:
             index_name (`str`): The index_name/identifier of the index. This is the index_name that is used to call `.get_nearest` or `.search`.
-            file (`str`): The path to the serialized faiss index on disk.
+            file (`str`): The path to the serialized faiss index on disk or remote URI (e.g. `"s3//my-bucket/index.faiss"`).
+            storage_options (`dict`, *optional*):
+                Key/value pairs to be passed on to the file-system backend, if any.
```
albertvillanova, line 534: Version added...
```suggestion
                Key/value pairs to be passed on to the file-system backend, if any.
                
                <Added version="2.10.0"/>
                
```

src/datasets/search.py
```diff
@@ -547,11 +554,14 @@ def load_faiss_index(
         Args:
             index_name (`str`): The index_name/identifier of the index. This is the index_name that is used to
                 call `.get_nearest` or `.search`.
-            file (`str`): The path to the serialized faiss index on disk.
+            file (`str`): The path to the serialized faiss index on disk or remote URI (e.g. `"s3//my-bucket/index.faiss"`).
             device (Optional `Union[int, List[int]]`): If positive integer, this is the index of the GPU to use. If negative integer, use all GPUs.
                 If a list of positive integers is passed in, run only on those GPUs. By default it uses the CPU.
+            storage_options (`dict`, *optional*):
+                Key/value pairs to be passed on to the file-system backend, if any.
```
albertvillanova, line 561:
```suggestion
                Key/value pairs to be passed on to the file-system backend, if any.
                
                <Added version="2.10.0"/>
```

albertvillanova (changes requested):
Thanks for the enhancement, @Dref360.

Dref360:
Thanks for the quick review @albertvillanova! I updated the code with your suggestions

src/datasets/search.py
```diff
@@ -547,11 +557,16 @@ def load_faiss_index(
...
             index_name (`str`): The index_name/identifier of the index. This is the index_name that is used to
                 call `.get_nearest` or `.search`.
-            file (`str`): The path to the serialized faiss index on disk.
+            file (`str`): The path to the serialized faiss index on disk or remote URI (e.g. `"s3://my-bucket/index.faiss"`).
             device (Optional `Union[int, List[int]]`): If positive integer, this is the index of the GPU to use. If negative integer, use all GPUs.
                 If a list of positive integers is passed in, run only on those GPUs. By default it uses the CPU.
+            storage_options (`dict`, *optional*):
+                Key/value pairs to be passed on to the file-system backend, if any.
+                
+                <Added version="2.10.0"/>
```
mariosasko, line 566:
```suggestion
                <Added version="2.11.0"/>
```

mariosasko (commented):
Thanks, one nit from me. Other than that, looks all good!

mariosasko (approved):
Thanks! @albertvillanova Your requested changes are addressed now, so I think we can merge. Do you agree?

albertvillanova (approved):
Thanks!!

PR #5527, opened 2023-02-12
Fix benchmarks CI - pin protobuf

lhoestq:
fix https://github.com/huggingface/datasets/actions/runs/4156059127/jobs/7189576331

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -154,6 +154,7 @@ BENCHMARKS_REQUIRE = [
     "tensorflow==2.3.0",
     "torch==1.7.1",
     "transformers==3.0.2",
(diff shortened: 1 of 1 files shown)


albertvillanova (approved):
Thanks for the fix, @lhoestq: this pin indeed fixes our CI benchmarks.

I would merge it as it is, so that our CI becomes green.

There is one thing that does not make sense to me: protobuf-4.21.12 was released on 14/12/2022 and we are getting our CI error just since 10/02/2023...

lhoestq:
Yea there must have been an update in another package that unconstrained the protobuf dependency - idk which one though

albertvillanova:
It is `tensorboard`. I have reported the issue to `tensorflow`:
- https://github.com/tensorflow/tensorflow/issues/59665

PR #5529, opened 2023-02-13
Fix `datasets.load_from_disk`, `DatasetDict.load_from_disk` and `Dataset.load_from_disk`

alvarobartt:
## What's in this PR?

After playing around a little bit with 🤗`datasets` in Google Cloud Storage (GCS), I found out some things that should be fixed IMO in the code:

* `datasets.load_from_disk` is not checking whether `state.json` is there too when trying to load a `Dataset`, just `dataset_info.json` is checked
* `DatasetDict.load_from_disk` is not checking whether `state.json` is there too when redirecting the user to load it as `datasets.load_from_disk`, just `dataset_info.json` is checked, which is misleading, as it won't be loadable that way either
* `Dataset.load_from_disk` is missing the `extract_path_from_uri` call before checking in the `fs` whether `dataset_info.json` and `dataset_dict.json` exist, which when using `gcsfs` leads to 400 error code (not blocking) due to `gcsfs.retry.HttpError: Invalid bucket name: 'gs:', 400`
* And, finally, the exception messages are a little bit misleading / incomplete IMO so I've tried to include all the relevant information in the messages to avoid issues when interpreting the exceptions


alvarobartt:
Hmm, should this also be updated in `Dataset.load_from_disk` and `DatasetDict.load_from_disk`? https://github.com/huggingface/datasets/pull/5466 As there the paths are joined using `Path(..., ...)` and it won't work on Windows OS according to that PR, right?

alvarobartt:
Hi, @lhoestq could you review this PR? Thank you in advance and sorry for the ping 🤗

src/datasets/arrow_dataset.py
```diff
@@ -1530,31 +1530,52 @@ def load_from_disk(
...
+            path_join = posixpath.join
+        else:
+            fs = fsspec.filesystem("file")
+            dest_dataset_path = dataset_path
+            path_join = os.path.join
+
+        dataset_dict_json_path = path_join(dest_dataset_path, config.DATASETDICT_JSON_FILENAME)
+        dataset_state_json_path = path_join(dest_dataset_path, config.DATASET_STATE_JSON_FILENAME)
+        dataset_info_path = path_join(dest_dataset_path, config.DATASET_INFO_FILENAME)
+        if fs.isfile(dataset_dict_json_path):
```
lhoestq, line 1545: `is_file` may do some http calls - maybe you can call it once for dataset_info_path and for dataset_state_json_path to save some time ?

lhoestq (commented):
Nice thanks ! Just one nit:

src/datasets/arrow_dataset.py
alvarobartt: True, I didn't realize! I'll update that now 👍🏻

alvarobartt:
Besides that, I was also thinking of adding a `skip_validation` boolean arg in both `Dataset.load_from_disk` and `DatasetDict.load_from_disk` to avoid duplicating those calls too when those functions are called from `datasets.load_from_disk`.

So that `skip_validation` is set to `False` by default, but passed as `True` if called from `datasets.load_from_disk`, and that just affects the file checking part of the code on both functions, do you agree?

src/datasets/arrow_dataset.py
```diff
@@ -1533,31 +1533,54 @@ def load_from_disk(
...
-            dataset_path = Dataset._build_local_temp_path(src_dataset_path)
-            fs.download(src_dataset_path, dataset_path.as_posix(), recursive=True)
+            dest_dataset_path = extract_path_from_uri(dataset_path)
+            path_join = posixpath.join
+        else:
+            fs = fsspec.filesystem("file")
+            dest_dataset_path = dataset_path
+            path_join = os.path.join
+
+        dataset_dict_json_path = path_join(dest_dataset_path, config.DATASETDICT_JSON_FILENAME)
```
alvarobartt, line 1545: Something like:

```suggestion
        if skip_validation:
            dataset_dict_json_path = path_join(dest_dataset_path, config.DATASETDICT_JSON_FILENAME)
            ...
```

src/datasets/dataset_dict.py
```diff
@@ -1285,26 +1285,35 @@ def load_from_disk(
...
+            path_join = posixpath.join
         else:
             fs = fsspec.filesystem("file")
             dest_dataset_dict_path = dataset_dict_path
-        dataset_dict_json_path = Path(dest_dataset_dict_path, config.DATASETDICT_JSON_FILENAME).as_posix()
-        dataset_info_path = Path(dest_dataset_dict_path, config.DATASET_INFO_FILENAME).as_posix()
-        if fs.isfile(dataset_info_path) and not fs.isfile(dataset_dict_json_path):
+            path_join = os.path.join
+
+        dataset_dict_json_path = path_join(dest_dataset_dict_path, config.DATASETDICT_JSON_FILENAME)
```
alvarobartt, line 1296: Same thing here:

```suggestion
        if skip_validation:
            dataset_dict_json_path = path_join(dest_dataset_dict_path, config.DATASETDICT_JSON_FILENAME)
            ...
```

src/datasets/arrow_dataset.py
alvarobartt, line 1545: Something like:

```suggestion
        if skip_validation:
            dataset_dict_json_path = path_join(dest_dataset_path, config.DATASETDICT_JSON_FILENAME)
            ...
```

src/datasets/load.py
```diff
@@ -1868,11 +1868,13 @@ def load_from_disk(
 
     if not fs.exists(dest_dataset_path):
         raise FileNotFoundError(f"Directory {dataset_path} not found")
-    if fs.isfile(path_join(dest_dataset_path, config.DATASET_INFO_FILENAME)):
+    if fs.isfile(path_join(dest_dataset_path, config.DATASET_INFO_FILENAME)) and fs.isfile(
+        path_join(dest_dataset_path, config.DATASET_STATE_JSON_FILENAME)
+    ):
         return Dataset.load_from_disk(dataset_path, keep_in_memory=keep_in_memory, storage_options=storage_options)
```
alvarobartt, line 1874:
```suggestion
        return Dataset.load_from_disk(dataset_path, keep_in_memory=keep_in_memory, storage_options=storage_options, skip_validation=True)
```

src/datasets/load.py
```diff
@@ -1868,11 +1868,13 @@ def load_from_disk(
...
     elif fs.isfile(path_join(dest_dataset_path, config.DATASETDICT_JSON_FILENAME)):
         return DatasetDict.load_from_disk(dataset_path, keep_in_memory=keep_in_memory, storage_options=storage_options)
```
alvarobartt, line 1876:
```suggestion
        return DatasetDict.load_from_disk(dataset_path, keep_in_memory=keep_in_memory, storage_options=storage_options, skip_validation=True)
```

alvarobartt (commented):
Something similar to what I've mocked here? @lhoestq Not sure about the naming anyway, `ignore_verifications` or `ignore_validation` may be more consistent?

lhoestq:
I think we should always verify

alvarobartt:
> I think we should always verify

But with the current way we're also verifying twice right? First on `datasets.load_from_disk` then on `Dataset.load_from_disk`, right?

lhoestq:
I don't think it's worth adding a new argument just for that. Usually we keep the set of arguments to the strict minimum

alvarobartt:
> I don't think it's worth adding a new argument just for that. Usually we keep the set of arguments to the strict minimum

What about the warning?

Anyway, if you don't think that's worth it feel free to merge 👍🏻

lhoestq:
> What about the warning?

We may show warnings for suggestions, but only if the user does a very unoptimized thing. Here we're not at that level ^^'

alvarobartt:
Thanks for the explanation and feedback @lhoestq 🤗

lhoestq (commented):
Thank you :) Added my last suggestions:

src/datasets/arrow_dataset.py
```diff
@@ -1533,31 +1533,54 @@ def load_from_disk(
...
+        dataset_state_json_path = path_join(dest_dataset_path, config.DATASET_STATE_JSON_FILENAME)
+        dataset_info_path = path_join(dest_dataset_path, config.DATASET_INFO_FILENAME)
+        if fs.isfile(dataset_dict_json_path):
+            dataset_info_is_file = fs.isfile(dataset_info_path)
+            dataset_state_is_file = fs.isfile(dataset_state_json_path)
+            if not dataset_info_is_file and not dataset_state_is_file:
+                raise FileNotFoundError(
+                    f"Neither file or directory: '{dataset_info_path}', nor '{dataset_state_json_path}' found. Expected to load a `Dataset` object, but got a `DatasetDict`. Please use either `datasets.load_from_disk` or `DatasetDict.load_from_disk` instead."
```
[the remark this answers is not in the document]
alvarobartt: I agree, I just kept it because it was like that before the PR :)

alvarobartt:
> Thank you :) Added my last suggestions:

Thanks for the feedback, I agree with everything besides one nit! 👍🏻

src/datasets/dataset_dict.py
```diff
@@ -1285,26 +1285,35 @@ def load_from_disk(
...
+        dataset_state_json_path = path_join(dest_dataset_dict_path, config.DATASET_STATE_JSON_FILENAME)
+        dataset_info_path = path_join(dest_dataset_dict_path, config.DATASET_INFO_FILENAME)
+        if not fs.isfile(dataset_dict_json_path):
+            if fs.isfile(dataset_info_path) and fs.isfile(dataset_state_json_path):
+                raise FileNotFoundError(
+                    f"No such file or directory: '{dataset_dict_json_path}'. Expected to load a `DatasetDict` object, but got a `Dataset`. Please use either `datasets.load_from_disk` or `Dataset.load_from_disk` instead."
+                )
             raise FileNotFoundError(
-                f"No such file or directory: '{dataset_dict_json_path}'. Expected to load a DatasetDict object, but got a Dataset. Please use datasets.load_from_disk instead."
+                f"No such file or directory: '{dataset_dict_json_path}'. Expected to load a `DatasetDict` object, but provided path is neither a `DatasetDict` nor a `Dataset`"
```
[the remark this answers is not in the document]
alvarobartt: I included that extra information basically to let the user know that it's not because it's a `Dataset` instead of ` DatasetDict`, as with that if-statement we can also guide the user, I thought that specifying that it's neither of those was useful!

P.S. Same thing applies to the `FileNotFound` error you mentioned in the other suggestion below in `src/datasets/arrow_dataset.py`

src/datasets/arrow_dataset.py
```diff
@@ -1533,31 +1533,54 @@ def load_from_disk(
...
+            fs = fsspec.filesystem("file")
+            dest_dataset_path = dataset_path
+            path_join = os.path.join
+
+        dataset_dict_json_path = path_join(dest_dataset_path, config.DATASETDICT_JSON_FILENAME)
+        dataset_state_json_path = path_join(dest_dataset_path, config.DATASET_STATE_JSON_FILENAME)
+        dataset_info_path = path_join(dest_dataset_path, config.DATASET_INFO_FILENAME)
+        if fs.isfile(dataset_dict_json_path):
+            dataset_info_is_file = fs.isfile(dataset_info_path)
+            dataset_state_is_file = fs.isfile(dataset_state_json_path)
```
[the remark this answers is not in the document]
alvarobartt: Same thing as mentioned in the other comment, if we can indeed let the user know that it's neither `Dataset` nor `DatasetDict` shouldn't we tell them? It's just a nit, but I think both options are correct :)

src/datasets/dataset_dict.py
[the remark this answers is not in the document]
lhoestq: Here we're talking about `DatasetDict.load_from_disk` - I don't think we need to tell the user that or they may think it's also fine to provide a `Dataset` directory, which is not correct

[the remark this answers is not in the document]
alvarobartt: True @lhoestq, ok then I'll commit this separately to join this suggestion and the one below in a single commit :)

lhoestq (approved):
Thank you !

PR #5530, opened 2023-02-13
Add missing license in `NumpyFormatter`

alvarobartt:
## What's in this PR?

As discussed with @lhoestq in https://github.com/huggingface/datasets/pull/5522, the license for `NumpyFormatter` at `datasets/formatting/np_formatter.py` was missing, but present on the rest of the `formatting/*.py` files. So this PR is basically to include it there.


albertvillanova (approved):
Thanks @alvarobartt.

PR #5533, opened 2023-02-15
Add reduce function

### LINKED ISSUE #5496: Add a `reduce` method
### Feature request

Right now the `Dataset` class implements `map()` and `filter()`, but leaves out the third functional idiom popular among Python users: `reduce`. 

### Motivation

A `reduce` method is often useful when calculating dataset statistics, for example, the occurrence of a particular n-gram or the average line length of a code dataset.  

### Your contribution

I haven't contributed to `datasets` before, but I don't expect this will be too difficult, since the implementation will closely follow that of `map` and `filter`. I could have a crack over the weekend.

AJDERS:
This PR closes #5496 .

I tried to imitate the `reduce`-method from `functools`, i.e. the function input must be a binary operation. I assume that the input type has an empty element, i.e. `input_type()` is defined, as the acumulant is instantiated as this object - im not sure that is this a reasonable assumption?

If `batched= True`  the reduction of each shard is _not_ returned, but the reduction of the entire dataset. I was unsure wether this was an intuitive API, or it would make more sense to return the reduction of each shard?


lhoestq (commented):
Cool thanks ! Would be cool to also be able to also be able to specify a custom `initializer` to be the first value of `accumulator`. This can be added in a subsequent PR, since we may also need to add a new argument `combiner` to combine results from each shard into one in multiprocessing.

Maybe it's possible to rely on `map` to simplify the implementation ?

Finally regarding batched `reduce`, I think it's not obvious what should be the initializer (especially for formatted datasets). So I'd be ok with forcing the user to specify an `initializer` in that case. What do you think ?

AJDERS:
I agree that it would be a good idea to introduce a `combiner` argument in another PR.

I did take quite a lot of inspiration from the implementation of  `map`, but it did not seem obvious how to resuse `map` for the implementation. Do you have any suggestions, i could give a try?

Those were exactly my thoughts, regarding the non-obvious initializer for batched and formatted datasets, so i agree! I'll  introduce a `initializer` argument, and have it mandatory when `batched=True`.

AJDERS:
I added `initializer`. It is optional for `batched=False` and mandatory for `batched=True`. It has to be of the same length as `input_columns`, if `input_columns=None` it has to have the same length as `_data.column_names`. 

If the initializer is not set for `batched=False` the first example is set as the `initializer`. 

The initializer is used to initiliaze for each shard, so that means if that:
```python
dset = Dataset.from_dict({"x": [1, 2, 3]})
sum_reduce = lambda x, y: x + y
reduction = dset.reduce(sum_reduce, batched=True, initializer=1, input_columns='x', num_proc=2)
# reduction is 8, i.e. reduction + num_proc * initializer
```

src/datasets/arrow_dataset.py
```diff
@@ -3353,6 +3354,421 @@ def init_buffer_and_writer():
...
+        num_proc: Optional[int] = None,
+        new_fingerprint: Optional[str] = None,
+        desc: Optional[str] = None,
+    ) -> Dict[str, Any]:
+        """
+        Reduce the examples in the table (individually or in batches) using a binary operation and return the result.
+
+        The function should be a binary operation, i.e. it takes in 2 objects of the same type in and returns 1 object of the same type as the inputs.
+        Examples include addition, multiplication, maximum, minimum, etc. for `int`, `float`, `bool` and `str` types, and concatenation for `list` types.
+        The first input is the accumulant, which is the result of the previous application of the function on the previous examples, and the second input is the current example.
```
lhoestq, line 3378: (nit) I'd call this "accumulator" here and in the rest of the code

```suggestion
        The first input is the accumulator, which is the result of the previous application of the function on the previous examples, and the second input is the current example.
```

AJDERS: Good point, ill get on it asap.

AJDERS:
> > I added initializer. It is optional for batched=False and mandatory for batched=True. It has to be of the same length as input_columns, if input_columns=None it has to have the same length as _data.column_names.
> > If the initializer is not set for batched=False the first example is set as the initializer.
> 
> Sounds good to me !
> 
> > The initializer is used to initiliaze for each shard, so that means if that:
> > ```python
> > dset = Dataset.from_dict({"x": [1, 2, 3]})
> > sum_reduce = lambda x, y: x + y
> > reduction = dset.reduce(sum_reduce, batched=True, initializer=1, input_columns='x', num_proc=2)
> > # reduction is 8, i.e. reduction + num_proc * initializer
> > ```
> 
> Hmm this can be confusing for some users. Maybe we should consider making `combiner` mandatory for multiprocessing.
> 
> If we agree on this, maybe for this PR you can either:
> 
> * remove multiprocessing (and we add combiner + multiprocessing in a subsequent PR)
> * OR add `combiner` directly
> 
> Maybe we can get more feedback from @huggingface/datasets as well

I think i prefer adding `combiner` in this PR. I think ill make `combiner` mandatory for `batched=True`, to avoid the above problem. Ill look at this one of the coming days.

AJDERS:
I added the `combiner` parameter as described. I added some examples in the docstring, as i felt it might still be a bit confusing what happens during multiprocessing / batching.

Still need to look at `DatasetDict`.

src/datasets/arrow_dataset.py
```diff
@@ -3371,71 +3371,100 @@ def reduce(
...
+         ```py
+         >>> from datasets import load_dataset
+         >>> string_ds = load_dataset("rotten_tomatoes", split="validation")
+         >>> str_concat = lambda accumulator, example: accumulator + ' ' + example
+         >>> result = string_ds[0:3].reduce(str_concat, input_columns="text")
+         >>> result
+         {'text': Review: compassionately explores the seemingly irreconcilable situation between conservative christian parents and their estranged gay and lesbian children . Review: the soundtrack alone is worth the price of admission . Review: rodriguez does a splendid job of racial profiling hollywood style--casting excellent latin actors of all ages--a trend long overdue .'}
+
+         # process a batch of examples
+         >>> result = ds.reduce(str_concat, input_columns="text", batched=True)
```
lhoestq, line 3435: this requires an initializer no ?

src/datasets/arrow_dataset.py
```diff
@@ -3371,71 +3371,100 @@ def reduce(
...
+             desc (`str`, *optional*, defaults to `None`):
+                 Meaningful description to be displayed alongside with the progress bar while mapping examples.
+
+         Example:
+
+         ```py
+         >>> from datasets import load_dataset
+         >>> string_ds = load_dataset("rotten_tomatoes", split="validation")
+         >>> str_concat = lambda accumulator, example: accumulator + ' ' + example
+         >>> result = string_ds[0:3].reduce(str_concat, input_columns="text")
```
lhoestq, line 3430: Usually reduce operations are used to compute aggregates, not merging examples together (you could just do `" ".join(ds["text"])` here)

Maybe we can find a use case that only `reduce` can achieve ? Maybe computing the `sum` of an integer column.
An actual use case would be to sum all the review of the `imdb` dataset, which can be used to derive the average review score.

src/datasets/arrow_dataset.py
```diff
@@ -3371,71 +3371,100 @@ def reduce(
...
+         >>> result
+         {'text': Review: compassionately explores the seemingly irreconcilable situation between conservative christian parents and their estranged gay and lesbian children . Review: the soundtrack alone is worth the price of admission . Review: rodriguez does a splendid job of racial profiling hollywood style--casting excellent latin actors of all ages--a trend long overdue .'}
+
+         # process a batch of examples
+         >>> result = ds.reduce(str_concat, input_columns="text", batched=True)
+
+         # set number of processors
+         >>> result = ds.reduce(str_concat, combiner=str_concat, input_columns="text", num_proc=4)
+
+         # set number of processors, with non-empty initializer, for input type `int`
```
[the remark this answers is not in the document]
AJDERS: Great example!

lhoestq:
Feel free to merge `main` into your branch - we fixed some CI failures today

src/datasets/arrow_dataset.py
```diff
@@ -3358,6 +3358,468 @@ def init_buffer_and_writer():
...
+         >>> count = lambda counter, text: counter + Counter(text.lower().split())
+         >>> sum_counts = lambda counter1, counter2: counter1 + counter2
+         >>> result = result = rotten_ds.select(range(1)).reduce(count, initializer=Counter(), input_columns="text")
+         >>> result
+        {'text': Counter({'and': 2, 'compassionately': 1, 'explores': 1, 'the': 1, 'seemingly': 1, 'irreconcilable': 1, 'situation': 1, 'between': 1, 'conservative': 1, 'christian': 1, 'parents': 1, 'their': 1, 'estranged': 1, 'gay': 1, 'lesbian': 1, 'children': 1, '.': 1})}
+        
+         # Calculate the average number of stars in the "Video_Games_v1_00" subset of the Amazon US reviews dataset
+         >>> review_ds = load_dataset("amazon_us_reviews", "Video_Games_v1_00", split="train")
+         >>> sum = lambda accumulator, review: accumulator + review
+         >>> sum_sums = lambda sum1, sum2: sum1 + sum2
```
lhoestq, line 3446: maybe introduce the combiner only for the multiprocessed case ?

src/datasets/arrow_dataset.py
```diff
@@ -3358,6 +3358,468 @@ def init_buffer_and_writer():
...
+
+         ```py
+         >>> from datasets import load_dataset
+         >>> from collections import Counter
+
+         # Count the number of words in the validation set of the Rotten Tomatoes dataset
+         >>> rotten_ds = load_dataset("rotten_tomatoes", split="validation")
+         >>> count = lambda counter, text: counter + Counter(text.lower().split())
+         >>> sum_counts = lambda counter1, counter2: counter1 + counter2
+         >>> result = result = rotten_ds.select(range(1)).reduce(count, initializer=Counter(), input_columns="text")
```
lhoestq, line 3439:
```suggestion
         >>> result = rotten_ds.select(range(1)).reduce(count, initializer=Counter(), input_columns="text")
```

src/datasets/arrow_dataset.py
```diff
@@ -3358,6 +3358,468 @@ def init_buffer_and_writer():
...
+         >>> result
+        {'text': Counter({'and': 2, 'compassionately': 1, 'explores': 1, 'the': 1, 'seemingly': 1, 'irreconcilable': 1, 'situation': 1, 'between': 1, 'conservative': 1, 'christian': 1, 'parents': 1, 'their': 1, 'estranged': 1, 'gay': 1, 'lesbian': 1, 'children': 1, '.': 1})}
```
lhoestq, line 3441: Since the initializer is a `Counter()`, I think we can expect the output to be a `Counter()` as well no ?

src/datasets/arrow_dataset.py
```diff
@@ -3358,6 +3358,468 @@ def init_buffer_and_writer():
...
+         >>> result['star_rating'] / length(review_ds)
+        4.059892597803915
+
+         # process a batch of examples
+         >>> review_ds = review_ds.reduce(sum, combiner=sum_sums, initializer=0, input_columns="star_rating", batched=True)
+
+         # set number of processors
+         >>> result = review_ds.reduce(sum, combiner=sum_sums, initializer=0, input_columns="star_rating", num_proc=4)
+
+         # set number of processors, with non-empty initializer, for input type `int`
```
lhoestq, line 3457:
```suggestion
         # in multiprocessing, the combiner is used to combine intermediate results which can cause surprising behaviors when using non-zero initializers
```

src/datasets/arrow_dataset.py
```diff
@@ -3358,6 +3358,468 @@ def init_buffer_and_writer():
...
+         >>> reduction = int_ds.reduce(sum_reduce, combiner=sum_reduce, initializer=1, input_columns='x', num_proc=2)
+         >>> reduction       # reduction is (1+(1+2)+1+(3)) = 8, i.e. initializer + reduction_1 + initializer + reduction_2
+        {'x': 8}
+
+         # set number of processors, with non-empty initializer, for input type `str`
+         >>> string_ds = Dataset.from_dict({"x": ["1", "2", "3"]})
+         >>> sum_reduce_reverse = lambda x, y: y + x
+         >>> reduction = string_ds.reduce(sum_reduce, combiner=sum_reduce_reverse, initializer="X", input_columns='x', num_proc=2)
+         >>> reduction       # reduction is ('X'+('3')+'X'+('2'+'1')) = 'X3X12', i.e. `initializer + reduction_2 + initializer + reduction_1`, note that the order of the reduction is reversed
+         {'x': 'X3X12'}
```
lhoestq, line 3469: I think we can remove this one - we're explaining the same thing as in the previous example

lhoestq (commented):
The documentation is starting to have a great shape ! Thanks a lot :)

I added some suggestions, and also one question regarding the output which IMO should be of the same type as the initializer

src/datasets/arrow_dataset.py
AJDERS: I see your point! It would also align more with the behaviour of `functools.reduce`. On the other hand, its a bit annoying that the output of different parameter variations is not the same type, as an example, `r1` and `r2` below would be `int` and `dict` respectively.
```python
int_ds = Dataset.from_dict({"x": [1, 2, 3], "y": [1, 2, 3]})
r1 = reduce(function, initializer=0)
r2 = reduce(function)
```
I don't really have a strong opinion one way or the other, either would be confusion/annoying in some way. Which do you prefer?

mariosasko:
The proposed API doesn't seem intuitive to me - one can already use `functools.reduce` or `Dataset.map` for this purpose ([Colab](https://colab.research.google.com/drive/1jCLv31Y4cDfqD0lhO0AnqEv3Or-LLvWe?usp=sharing) with examples), so perhaps we could have a section in the docs that uses these methods to perform reductions rather than introducing a new method (which needs to be maintained later)

mariosasko:
> native support for lambdas (using dill)

Replacing `multiprocessing` with `multiprocess` in the example would allow that.

> no need to use map just to iterate on the dataset (not its main purpose)

Not the main purpose, but this was mentioned as a "feature" in the previous docs if I remember.

And all this is related to the multi-processing case, which we can document.

Besides the linked issue, I can't find requests for `Dataset.reduce`, which makes me think `functools.reduce` does the job for most users.

lhoestq:
> Besides the linked issue, I can't find requests for Dataset.reduce, which makes me think functools.reduce does the job for most users.

I think @srush was looking for a way to do a word count but ended up using a single processed `map`. I also saw some users on the forum wanting to compute `max`

> Not the main purpose, but this was mentioned as a "feature" in the previous docs if I remember.
> 
> And all this is related to the multi-processing case, which we can document.

Yup indeed

srush:
While counting is one example, I often find I want to compute different statistics over a dataset. This seems like a natural way to do it in a stateless manner.


I guess you could use functools reduce, but that wouldn't allow batching, right?

mariosasko:
I've updated the [Colab](https://colab.research.google.com/drive/1jCLv31Y4cDfqD0lhO0AnqEv3Or-LLvWe?usp=sharing) with an example that reduces batches with `map` and then computes the final result. It would be nice to have a similar example (explained in detail) in the docs to show the full power of `map`.

> I guess you could use functools reduce, but that wouldn't allow batching, right?

You can use `.iter(batch_size)` to get batches

srush:
That `functools` tools example is clean. I didn't know about `iter`. That would handle my use case.

The stateful `map` with a global variable is pretty hairy. I don't think we should recommend people do that.

AJDERS:
Whenever I in the past wanted to calculate statistics for datasets I used `functools` similarly to how it's described in the colab, but I always felt it was a bit of a hassle to do multiprocessing, which is why I picked up the issue, to do it "once and for all".

AJDERS:
Should i close this and open another PR, with descriptions of how to use `map` for reduction, or?

lhoestq:
Yes I think good documentation is the way to go here. @mariosasko 's examples are clear and efficient.

Maybe we could have an `Aggregations` section in the `Process` page with some guides on how to:
- use `.map()` to compute aggregates
- use `.with_format("arrow")` for max, min, etc. to save RAM and get max speed
- use a multiprocessed `.map()` to get partial results in parallel and combine them (max text length example)
- (advanced) use multiprocessing with an arbitrary accumulator (word count example)

And also a new conceptual guide on `Multiprocessed mapping` to say that it helps speed up CPU intensive processing but why it may lead to incorrect results when computing aggregates.

cc @stevhliu for visibility and if you have some comments

stevhliu:
I would create a `Reduce` - to be more exact - subsection under `Map` to demonstrate these examples since we're showing how they can be done with the `Dataset.map` function. It'd also be good to add a link to the new concept guide from this section to solidify user understanding :)

AJDERS:
Coolio. Ill close this PR and get going on another one adding what we've discussed during the next couple of days!

taha-yassine:
Is adding a section to the docs still planned? Couldn't find any related PR.

lhoestq:
There is a new integration with polars which is convenient btw. Here is an example for computing the length of the longest dialogue in a dataset using polars:

```python
>>> ds = load_dataset("HuggingFaceTB/smoltalk", "all", split="train")
>>> df = ds.to_polars()
>>> df.head()
shape: (5, 2)
┌─────────────────────────────────┬───────────────────┐
│ messages                        ┆ source            │
│ ---                             ┆ ---               │
│ list[struct[2]]                 ┆ str               │
╞═════════════════════════════════╪═══════════════════╡
│ [{"The function \( g(x) \) sat… ┆ numina-cot-100k   │
│ [{"Ben twice chooses a random … ┆ numina-cot-100k   │
│ [{"Find all values of $x$ that… ┆ numina-cot-100k   │
│ [{"How can you help me? I'm wr… ┆ smol-magpie-ultra │
│ [{"Extract and present the mai… ┆ smol-summarize    │
└─────────────────────────────────┴───────────────────┘
>>> df["messages"].list.len().max()
58
```

For very large scale dataset it can be worth using `map()` on batches on data to compute intermediate results, save some memory, and cache the result:

```python
>>> f = lambda df: pl.DataFrame({"messages_max_length": [df["messages"].list.len().max()]})
>>> intermediate_ds = ds.with_format("polars").map(f, batched=True)
>>> intermediate_ds.to_polars()["messages_max_length"].max()
58
```
This last method can be used to implement a map + intermediate reduce + final reduce approach

PR #5535, opened 2023-02-15
Add JAX-formatting documentation

alvarobartt:
## What's in this PR?

As a follow-up of #5522, I've created this entry in the documentation to explain how to use `.with_format("jax")` and why is it useful.

@lhoestq Feel free to drop any feedback and/or suggestion, as probably more useful features can be included there!

### DIFF
--- a/docs/source/use_with_jax.mdx
+++ b/docs/source/use_with_jax.mdx
@@ -102,7 +102,7 @@ To use the [`Image`] feature type, you'll need to install the `vision` extra as
 >>> ds = Dataset.from_dict({"image": images}, features=features)
 >>> ds = ds.with_format("jax")
 >>> ds[0]["image"].shape
-torch.Size([512, 512, 3])
+(512, 512, 3)
 >>> ds[0]
 {'image': DeviceArray([[[ 255, 255, 255],
               [ 255, 255, 255],
@@ -110,7 +110,7 @@ torch.Size([512, 512, 3])
               [ 255, 255, 255],
               [ 255, 255, 255]]], dtype=uint8)}
 >>> ds[:2]["image"].shape
-torch.Size([2, 512, 512, 3])
+(2, 512, 512, 3)
 >>> ds[:2]
 {'image': DeviceArray([[[[ 255, 255, 255],
               [ 255, 255, 255],
--- a/docs/source/use_with_jax.mdx
+++ b/docs/source/use_with_jax.mdx
@@ -1,0 +1,98 @@
+# Use with JAX
+
+This document is a quick introduction to using `datasets` with JAX, with a particular focus on how to get
+`jax.Array` objects out of our datasets, and how to use them to train JAX models.
+
+<Tip>
+
+`jax` and `jaxlib` are required to reproduce to code above, so please make sure you
+install them as `pip install datasets[jax]`.
+
+</Tip>
+
+## Dataset format
+
+By default, datasets return regular Python objects: integers, floats, strings, lists, etc.
+
+To get JAX arrays (numpy-like) instead, you can set the format of the dataset to `jax`:
+
+```py
+>>> from datasets import Dataset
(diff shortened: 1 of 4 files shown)


lhoestq (commented):
Awesome thank you !

Could you also explain how to use certain types like ClassLabel, Image or Audio with jax ? You can get a lot of inspiration from the "Other feature types" section in the [PyTorch page](https://huggingface.co/docs/datasets/use_with_pytorch)

I also think it's be nice if this page had the same structure as the pytorch or tf ones, with sections named
- Dataset format
- N-dimensional arrays
- Other feature types
- Data loading

alvarobartt:
> Awesome thank you !
> 
> Could you also explain how to use certain types like ClassLabel, Image or Audio with jax ? You can get a lot of inspiration from the "Other feature types" section in the [PyTorch page](https://huggingface.co/docs/datasets/use_with_pytorch)
> 
> I also think it's be nice if this page had the same structure as the pytorch or tf ones, with sections named
> 
>     * Dataset format
> 
>     * N-dimensional arrays
> 
>     * Other feature types
> 
>     * Data loading

Sure @lhoestq I'll do that later this afternoon whenever I'm done working! Thanks for the feedback as always 🤗

alvarobartt:
Also, @lhoestq do you want me to elaborate more on the `## Data loading` section on how to use `datasets` to train a JAX model offering alternatives e.g. `Flax`, or do I keep it pure JAX? Thanks!

lhoestq:
If you have a good example with `flax` it can also be helpful for users

alvarobartt:
For now I think that probably it's not worth to add a `Flax` example, as train loops need to be done manually as in pure JAX, so probably the JAX example is enough. Anyway, let me know if you see something missing/incomplete/misleading/etc. and I'll update that ASAP 👍🏻

alvarobartt:
P.S. I see that the `benchmark` action is being triggered on every PR, is it worth it? e.g. now I'm just editing the docs, so does it make any sense to trigger still the whole CI pipeline (including `benchmark`)? Just asking because in this PR for example it could be skipped.

lhoestq:
> P.S. I see that the benchmark action is being triggered on every PR, is it worth it? e.g. now I'm just editing the docs, so does it make any sense to trigger still the whole CI pipeline (including benchmark)? Just asking because in this PR for example it could be skipped.

We could restrict it to PRs modifying files in src/ indeed ^^'

docs/source/use_with_jax.mdx
```diff
@@ -0,0 +1,183 @@
...
+>>> ds[0]["image"].shape
+torch.Size([512, 512, 3])
+>>> ds[0]
+{'image': DeviceArray([[[ 255, 255, 255],
+              [ 255, 255, 255],
+              ...,
+              [ 255, 255, 255],
+              [ 255, 255, 255]]], dtype=uint8)}
+>>> ds[:2]["image"].shape
+torch.Size([2, 512, 512, 3])
```
lhoestq, line 113:
```suggestion
(2, 512, 512, 3)
```

docs/source/use_with_jax.mdx
```diff
@@ -0,0 +1,183 @@
...
+</Tip>
+
+```py
+>>> from datasets import Dataset, Features, Image
+>>> images = ["path/to/image.png"] * 10
+>>> features = Features({"image": Image()})
+>>> ds = Dataset.from_dict({"image": images}, features=features)
+>>> ds = ds.with_format("jax")
+>>> ds[0]["image"].shape
+torch.Size([512, 512, 3])
```
lhoestq, line 105:
```suggestion
(512, 512, 3)
```

lhoestq (approved):
LGTM :)

alvarobartt:
> LGTM :)

Cool thanks! My bad I didn't update those code blocks 🙃 Thanks for doing so before merge!

PR #5540, opened 2023-02-16
Tutorial for creating a dataset

stevhliu:
A tutorial for creating datasets based on the folder-based builders and `from_dict` and `from_generator` methods. I've also mentioned loading scripts as a next step, but I think we should keep the tutorial focused on the low-code methods. Let me know what you think! 🙂

### DIFF
--- a/docs/source/create_dataset.mdx
+++ b/docs/source/create_dataset.mdx
@@ -1,0 +1,108 @@
+# Create a dataset
+
+Sometimes, you may need to create a dataset if you"re working with your own data. Creating a dataset with 🤗 Datasets confers all the advantages of the library to your dataset: fast loading and processing, [stream enormous datasets](stream), [memory-mapping](https://huggingface.co/course/chapter5/4?fw=pt#the-magic-of-memory-mapping), and more. You can easily and rapidly create a dataset with 🤗 Datasets low-code approaches, reducing the time it takes to start training a model. In many cases, it is as easy as [dragging and dropping](upload_dataset#upload-with-the-hub-ui) your data files into a dataset repository on the Hub.
+
+In this tutorial, you"ll learn how to use 🤗 Datasets low-code methods for creating all types of datasets:
+
+* Folder-based builders for quickly creating an image or audio dataset
+* `from_` methods for creating datasets from local files
+
+## Folder-based builders
+
+There are two folder-based builders, [`ImageFolder`] and [`AudioFolder`]. These are low-code methods for quickly creating an image or speech and audio dataset with several thousand examples. They are great for rapidly prototyping computer vision and speech models before scaling to a larger dataset. Folder-based builders takes your data and automatically generates the dataset"s features, splits, and labels. Under the hood:
+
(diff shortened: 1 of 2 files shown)


docs/source/create_dataset.mdx
```diff
@@ -0,0 +1,108 @@
...
+    >>> ds = Dataset.from_generator(gen)
+    ```
+
+    Generator-based datasets need to be iterated over with either a `for` loop or by calling `next(iter())`:
+
+    ```py
+    >>> for example in ds:
+    ...     print(example)
+    {"pokemon": "bulbasaur", "type": "grass"}
+    {"pokemon": "squirtle", "type": "water"}
```
lhoestq, line 84: A Dataset created with a generator can be queried using normal indexing:
```suggestion
    >>> ds = Dataset.from_generator(gen)
    >>> ds[0]
    {"pokemon": "bulbasaur", "type": "grass"}
    >>> ds[1]
    {"pokemon": "squirtle", "type": "water"}
```

Not sure if it fits the tutorial scope here, but we can also mention `IterableDataset` that indeed requires to be iterated with a `for` loop:

```python
    >>> ds = IterableDataset.from_generator(gen)
    >>> for example in ds:
    ...     print(example)
    {"pokemon": "bulbasaur", "type": "grass"}
    {"pokemon": "squirtle", "type": "water"}
```

docs/source/create_dataset.mdx
```diff
@@ -0,0 +1,108 @@
...
+squirtle.png, When it retracts its long neck into its shell, it squirts out water with vigorous force.
+```
+
+To learn more about each of these folder-based builders, check out the and <a href="https://huggingface.co/docs/datasets/image_dataset#imagefolder"><span class="underline decoration-yellow-400 decoration-2 font-semibold">ImageFolder</span></a> or <a href="https://huggingface.co/docs/datasets/audio_dataset#audiofolder"><span class="underline decoration-pink-400 decoration-2 font-semibold">AudioFolder</span></a> guides.
+
+## From local files
+
+You can also create a dataset from local files by specifying the path to the data files. There are two ways you can create a dataset using the `from_` methods:
+
+    * The [`~Dataset.from_generator`] method is the most memory-efficient way to create a dataset from a [generator](https://wiki.python.org/moin/Generators) due to a generators lazy behavior. This is especially useful when you"re working with a really large dataset that may not fit in memory.
```
lhoestq, line 68: I would keep the work "lazy" for iterable datasets

```suggestion
    * The [`~Dataset.from_generator`] method is the most memory-efficient way to create a dataset from a [generator](https://wiki.python.org/moin/Generators) due to a generators iterative behavior. This is especially useful when you"re working with a really large dataset that may not fit in memory, since the dataset is generated on disk progressively and then memory-mapped.
```

lhoestq (commented):
Thanks ! A few comments:

docs/source/create_dataset.mdx
```diff
@@ -0,0 +1,111 @@
...
+    ...     yield {"pokemon": "squirtle", "type": "water"}
+    >>> ds = Dataset.from_generator(gen)
+    >>> ds[0]
+    {"pokemon": "bulbasaur", "type": "grass"}
+    ```
+
+    A generator-based [`IterableDataset`] needs to be iterated over with a `for` loop for example:
+
+    ```py
+    >>> ds = IterableDataset.from_generator(gen)
```
lhoestq, line 83:
```suggestion
    >>> from datasets import IterableDataset
    >>> ds = IterableDataset.from_generator(gen)
```

lhoestq (approved):
Thanks !

mariosasko (approved):
LGTM!

Issue #5541, opened 2023-02-17
Flattening indices in selected datasets is extremely inefficient


marioga:
### Describe the bug

If we perform a `select`  (or `shuffle`, `train_test_split`, etc.) operation on a dataset , we end up with a dataset with an `indices_table`. Currently, flattening such dataset consumes a lot of memory and the resulting flat dataset contains ChunkedArrays with as many chunks as there are rows. This is extremely inefficient and slows down the operations on the flat dataset, e.g., saving/loading the dataset to disk becomes really slow. 

Perhaps more importantly, loading the dataset back from disk basically loads the whole table into RAM, as it cannot take advantage of memory mapping.

### Steps to reproduce the bug

The following script reproduces the issue:

```python
import gc
import os
import psutil
import tempfile
import time

from datasets import Dataset

DATASET_SIZE = 5000000


def profile(func):
    def wrapper(*args, **kwargs):
        mem_before = psutil.Process(os.getpid()).memory_info().rss / (1024 * 1024)
        start = time.time()

        # Run function here
        out = func(*args, **kwargs)

        end = time.time()
        mem_after = psutil.Process(os.getpid()).memory_info().rss / (1024 * 1024)
        print(f"{func.__name__} -- RAM memory used: {mem_after - mem_before} MB -- Total time: {end - start:.6f} s")

        return out

    return wrapper


def main():
    ds = Dataset.from_list([{'col': i} for i in range(DATASET_SIZE)])
    print(f"Num chunks for original ds: {ds.data['col'].num_chunks}")

    with tempfile.TemporaryDirectory() as tmpdir:
        path1 = os.path.join(tmpdir, 'ds1')
        print("Original ds save/load")
        profile(ds.save_to_disk)(path1)

        ds_loaded = profile(Dataset.load_from_disk)(path1)
        print(f"Num chunks for original ds after reloading: {ds_loaded.data['col'].num_chunks}")
        print("")

        ds_select = ds.select(reversed(range(len(ds))))
        print(f"Num chunks for selected ds: {ds_select.data['col'].num_chunks}")
        del ds
        del ds_loaded
        gc.collect()

        # This would happen anyway when we call save_to_disk
        ds_select = profile(ds_select.flatten_indices)()
        print(f"Num chunks for selected ds after flattening: {ds_select.data['col'].num_chunks}")
        print("")

        path2 = os.path.join(tmpdir, 'ds2')
        print("Selected ds save/load")
[...]

PR #5542, opened 2023-02-17
Avoid saving sparse ChunkedArrays in pyarrow tables

marioga:
Fixes https://github.com/huggingface/datasets/issues/5541


lhoestq (approved):
Good catch ! Thanks a lot for the fix :)

This fix is pretty important so we'll do a new release soon

PR #5547, opened 2023-02-18
Add JAX device selection when formatting

alvarobartt:
## What's in this PR?

After exploring for a while the JAX integration in 🤗`datasets`, I found out that, even though JAX prioritizes the TPU and GPU as the default device when available, the `JaxFormatter` doesn't let you specify the device where you want to place the `jax.Array`s in case you don't want to rely on JAX's default array placement.

So on, I've included the `device` param in `JaxFormatter` but there are some things to take into consideration:
* A formatted `Dataset` is copied with `copy.deepcopy` which means that if one adds the param `device` in `JaxFormatter` as a `jaxlib.xla_extension.Device`, it "fails" because that object cannot be serialized (instead of serializing the param adds a random hash instead). That's the reason why I added a function `_map_devices_to_str` to basically create a mapping of strings to `jaxlib.xla_extension.Device`s so that `self.device` is a string and not a `jaxlib.xla_extension.Device`.
* To create a `jax.Array` in a device you need to either create it in the default device and then move it to the desired device with `jax.device_put` or directly create it in the device you want with `jax.default_device()` context manager.
* JAX will create an array by default in `jax.devices()[0]`

More information on JAX device management is available at https://jax.readthedocs.io/en/latest/faq.html#controlling-data-and-computation-placement-on-devices

## What's missing in this PR?

I've tested it both locally in CPU (Mac M2 and Mac M1, as no GPU support for Mac yet), and in GPU and TPU in Google Colab, let me know if you want me to provide you the Notebook for the latter.

But I did not implement any integration test as I wanted to get your feedback first.

### DIFF
--- a/src/datasets/formatting/jax_formatter.py
+++ b/src/datasets/formatting/jax_formatter.py
@@ -42,7 +42,8 @@ class JaxFormatter(Formatter[Mapping, "jax.Array", Mapping]):
         )
         self.jnp_array_kwargs = jnp_array_kwargs
 
-    def _map_devices_to_str(self) -> Dict[str, "jaxlib.xla_extension.Device"]:
+    @staticmethod
+    def _map_devices_to_str() -> Dict[str, "jaxlib.xla_extension.Device"]:
         import jax
 
         return {str(device): device for device in jax.devices()}
--- a/src/datasets/formatting/jax_formatter.py
+++ b/src/datasets/formatting/jax_formatter.py
@@ -26,9 +26,10 @@ from .formatting import Formatter
 
 
 if TYPE_CHECKING:
+    from typing import Dict
+
     import jax
     import jaxlib
-    from typing import Dict
 
 
 class JaxFormatter(Formatter[Mapping, "jax.Array", Mapping]):
--- a/src/datasets/formatting/jax_formatter.py
+++ b/src/datasets/formatting/jax_formatter.py
@@ -27,13 +27,25 @@ from .formatting import Formatter
 
 if TYPE_CHECKING:
     import jax
+    import jaxlib
 
 
 class JaxFormatter(Formatter[Mapping, "jax.Array", Mapping]):
-    def __init__(self, features=None, **jnp_array_kwargs):
+    def __init__(self, features=None, device=None, **jnp_array_kwargs):
         super().__init__(features=features)
+        import jax
+        from jaxlib.xla_client import Device
+
+        self.device_mapping = self._map_devices_to_str()
+        self.device = (
+            device if isinstance(device, str) else str(device) if isinstance(device, Device) else str(jax.devices()[0])
+        )
         self.jnp_array_kwargs = jnp_array_kwargs
-        import jax  # noqa import jax at initialization
+
+    def _map_devices_to_str(self) -> Mapping[str, "jaxlib.xla_extension.Device"]:
+        import jax
+
+        return {str(device): device for device in jax.devices()}
 
     def _consolidate(self, column):
         import jax
@@ -72,9 +84,10 @@ class JaxFormatter(Formatter[Mapping, "jax.Array", Mapping]):
             if isinstance(value, PIL.Image.Image):
                 value = np.asarray(value)
 
-        # calling jnp.array on a np.ndarray does copy the data
-        # see https://github.com/google/jax/issues/4486
-        return jnp.array(value, **{**default_dtype, **self.jnp_array_kwargs})
+        with jax.default_device(self.device_mapping[self.device]):
+            # calling jnp.array on a np.ndarray does copy the data
+            # see https://github.com/google/jax/issues/4486
+            return jnp.array(value, **{**default_dtype, **self.jnp_array_kwargs})
 
     def _recursive_tensorize(self, data_struct: dict):
         # support for nested types like struct of list of struct
--- a/src/datasets/formatting/jax_formatter.py
+++ b/src/datasets/formatting/jax_formatter.py
@@ -32,6 +32,8 @@ if TYPE_CHECKING:
 
 logger = get_logger()
 
+DEVICE_MAPPING = {}
+
 
 class JaxFormatter(Formatter[Mapping, "jax.Array", Mapping]):
     def __init__(self, features=None, device=None, **jnp_array_kwargs):
@@ -47,11 +49,15 @@ class JaxFormatter(Formatter[Mapping, "jax.Array", Mapping]):
                 "to the actual `jaxlib.xla_extension.Device`."
             )
         self.device = device if isinstance(device, str) else str(jax.devices()[0])
-        self.device_mapping = self._map_devices_to_str()
-        if self.device not in list(self.device_mapping.keys()):
+        # using global variable since `jaxlib.xla_extension.Device` is not serializable neither
+        # with `pickle` nor with `dill`, so we need to use a global variable instead
+        global DEVICE_MAPPING
+        if not DEVICE_MAPPING:
+            DEVICE_MAPPING = self._map_devices_to_str()
+        if self.device not in list(DEVICE_MAPPING.keys()):
             logger.warning(
                 f"Device with string identifier {self.device} not listed among the available "
-                "devices: {list(self.device_mapping.keys())}, so falling back to the default "
+                f"devices: {list(DEVICE_MAPPING.keys())}, so falling back to the default "
                 f"device: {str(jax.devices()[0])}."
             )
             self.device = str(jax.devices()[0])
@@ -100,7 +106,13 @@ class JaxFormatter(Formatter[Mapping, "jax.Array", Mapping]):
             if isinstance(value, PIL.Image.Image):
                 value = np.asarray(value)
 
-        with jax.default_device(self.device_mapping[self.device]):
+        # using global variable since `jaxlib.xla_extension.Device` is not serializable neither
+        # with `pickle` nor with `dill`, so we need to use a global variable instead
+        global DEVICE_MAPPING
+        if not DEVICE_MAPPING:
+            DEVICE_MAPPING = self._map_devices_to_str()
+
+        with jax.default_device(DEVICE_MAPPING[self.device]):
             # calling jnp.array on a np.ndarray does copy the data
             # see https://github.com/google/jax/issues/4486
             return jnp.array(value, **{**default_dtype, **self.jnp_array_kwargs})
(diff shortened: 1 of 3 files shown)


alvarobartt:
The code below was throwing a warning:

```python
class JaxFormatter(Formatter[Mapping, "jax.Array", Mapping]):
    def __init__(self, features=None, device=None, **jnp_array_kwargs):
        super().__init__(features=features)
        import jax
        from jaxlib.xla_extension import Device
        
        self.device = (
            device if isinstance(device, Device) else jax.devices()[0]
        )
        self.jnp_array_kwargs = jnp_array_kwargs

    ...

    def _tensorize(self, value):
        ...

        with jax.default_device(self.device):
            # calling jnp.array on a np.ndarray does copy the data
            # see https://github.com/google/jax/issues/4486
            return jnp.array(value, **{**default_dtype, **self.jnp_array_kwargs})
```

When providing `device` via param:

```python
from datasets import Dataset
import jax

ds = Dataset.from_dict({"a": [1, 2, 3], "b": [4, 5, 6]})
ds = ds.with_format("jax", device=jax.devices()[0])
print(ds[0])
```

Producing the following warning:

```
WARNING:datasets.fingerprint:Parameter 'device'=TFRT_CPU_0 of the transform datasets.arrow_dataset.Dataset.set_format couldn't be hashed properly, a random hash was used instead. Make sure your transforms and parameters are serializable with pickle or dill for the dataset fingerprinting and caching to work. If you reuse this transform, the caching mechanism will consider it to be different from the previous calls and recompute everything. This warning is only showed once. Subsequent hashing failures won't be showed.
```

That's why I decided to map all the available devices, and assign their string representation e.g. `TFRT_CPU_0` to `self.device` instead of `jaxlib.xla_extension.Device`, so that the value of the param `device` is washable. So on, the code that remains at the end is:

```python
class JaxFormatter(Formatter[Mapping, "jax.Array", Mapping]):
    def __init__(self, features=None, device=None, **jnp_array_kwargs):
        super().__init__(features=features)
        import jax
        from jaxlib.xla_client import Device

        self.device_mapping = self._map_devices_to_str()
        self.device = (
            device if isinstance(device, str) else str(device) if isinstance(device, Device) else str(jax.devices()[0])
        )
        self.jnp_array_kwargs = jnp_array_kwargs

    def _map_devices_to_str(self) -> Mapping[str, "jaxlib.xla_extension.Device"]:
        import jax

        return {str(device): device for device in jax.devices()}

    ...

    def _tensorize(self, value):
        ...

        with jax.default_device(self.device_mapping[self.device]):
            # calling jnp.array on a np.ndarray does copy the data
            # see https://github.com/google/jax/issues/4486
            return jnp.array(value, **{**default_dtype, **self.jnp_array_kwargs})
```

alvarobartt:
After some investigation, it seems that when using `device=jaxlib.xla_extension.Device` instead of `device=string` it shows the warning so that later formats fail as that cannot be unpickled.

So I think we can either add that specifically in `use_with_jax.mdx` documentation entry I'm creating at #5535 so that the users know that they need to surroung the `jaxlib.xla_extension.Device` with `str()`, or find a workaround to override default `deepcopy` behavior with `def __deepcopy__(self)` so that the device param is converted to string if provided as a `jaxlib.xla_extension.Device`, but not sure if the latter works 😕 

Do you think there's any other possible solution to this issue? Thanks, @lhoestq

lhoestq:
Cool ! Specifying the device is indeed super important.


I think we can just require `device` to always be a string for now, and add an example in the doc on how to get the string that corresponds to a `jaxlib.xla_extension.Device` ? This way we never deal with objects that are not picklable

alvarobartt:
> Cool ! Specifying the device is indeed super important.
> 
> I think we can just require `device` to always be a string for now, and add an example in the doc on how to get the string that corresponds to a `jaxlib.xla_extension.Device` ? This way we never deal with objects that are not picklable

Sure, then I'll restrict it to string for now! Also regarding the documentation update, should we wait until #5535 is merged so that I add this on top of that?

alvarobartt:
CI is failing due to missing `resampy` in `librosa` already being fixed by @lhoestq in https://github.com/huggingface/datasets/pull/5554

src/datasets/formatting/jax_formatter.py
```diff
@@ -15,25 +15,53 @@
...
+
+        if isinstance(device, Device):
+            raise ValueError(
+                f"Expected {device} to be a `str` not {type(device)}, as `jaxlib.xla_extension.Device` "
+                "is not serializable neither with `pickle` nor with `dill`. Instead you can surround "
+                "the device with `str()` to get its string identifier that will be internally mapped "
+                "to the actual `jaxlib.xla_extension.Device`."
+            )
+        self.device = device if isinstance(device, str) else str(jax.devices()[0])
+        self.device_mapping = self._map_devices_to_str()
```
lhoestq, line 50: wouldn't this end up make the formatter unpicklable ?

alvarobartt: So as far as I could test it was just unpickable when the `device` param is `jaxlib.xla_extension.Device`, otherwise it works fine. But let me check again maybe I skipped something?

alvarobartt: Indeed this works

```python
import jax
from datasets import Dataset
data = [[1, 2], [3, 4]]
ds = Dataset.from_dict({"data": data})
device = str(jax.devices()[0])
ds = ds.with_format("jax", device=device)
print(ds[0])
print(ds[0]["data"].device())
ds = ds.with_format("torch")
print(ds[0])
print(ds[0]["data"])
```

But if I remove the `ValueError` exception and instead replace that with the following:

```python
...
if isinstance(device, Device):
    device = str(device)
...
```

And then instead of the code above, I run the code below without the `str` casting:

```python
import jax
from datasets import Dataset
data = [[1, 2], [3, 4]]
ds = Dataset.from_dict({"data": data})
device = jax.devices()[0]
ds = ds.with_format("jax", device=device)
print(ds[0])
print(ds[0]["data"].device())
ds = ds.with_format("torch")
print(ds[0])
print(ds[0]["data"])
```

Then a warning will be raised on `with_format("jax", device=device)` and then `TypeError: cannot pickle 'jaxlib.xla_extension.Device' object` when calling `with_format("torch")`.

lhoestq: Yea but I thought the dictionary values were `jaxlib.xla_extension.Device` objects ? Because of that the formatter itself is not picklable
```python
pickle.dumps(formatter)
```

I guess it's not possible to pass a string device to `jnp.array` ?

alvarobartt: So what I understand is that it should fail when `copy.deepcopy(self)` but it's not, so maybe the `copy.deepcopy` is done as part of the `super().__init__()` so that just the params are copied?

alvarobartt: And that's right, device management is done either with the context manager or with `device_put`, and it expects a `jaxlib.xla_extension.Device` not a string :/

lhoestq: > So what I understand is that it should fail when copy.deepcopy(self) but it's not, so maybe the copy.deepcopy is done as part of the super().__init__() so that just the params are copied?

Actually we should have copy.deepcopy AND pickle.dumps working. That's because a formatter can also be used with other objects than a `Dataset`.

We could have the mapping as a global variable that is initialized when `_tensorize` is called for the first time ? This way we don't need to reload it every time we instantiate a formatter, and the formatter can become picklable

alvarobartt: Oh OK didn't know that, then I'll do what you're suggesting :)

alvarobartt:
@lhoestq already moved to a global variable, I can confirm that the following now works:

```python
import copy
import pickle

import jax
import pyarrow as pa

from datasets.formatting import JaxFormatter


_COL_A = [0, 1, 2]
_COL_B = ["foo", "bar", "foobar"]
_COL_C = [[[1.0, 0.0, 0.0]] * 2, [[0.0, 1.0, 0.0]] * 2, [[0.0, 0.0, 1.0]] * 2]
pa_table = pa.Table.from_pydict({"a": _COL_A, "b": _COL_B, "c": _COL_C})

device = jax.devices()[0]
formatter = JaxFormatter(device=str(device))

pickle.dumps(formatter)
copy.deepcopy(formatter)
```

src/datasets/formatting/jax_formatter.py
```diff
@@ -47,11 +49,15 @@ def __init__(self, features=None, device=None, **jnp_array_kwargs):
                 "to the actual `jaxlib.xla_extension.Device`."
             )
         self.device = device if isinstance(device, str) else str(jax.devices()[0])
-        self.device_mapping = self._map_devices_to_str()
-        if self.device not in list(self.device_mapping.keys()):
+        # using global variable since `jaxlib.xla_extension.Device` is not serializable neither
+        # with `pickle` nor with `dill`, so we need to use a global variable instead
+        global DEVICE_MAPPING
+        if not DEVICE_MAPPING:
```
lhoestq, line 55:
```suggestion
        if DEVICE_MAPPING is None:
```

src/datasets/formatting/jax_formatter.py
```diff
@@ -100,7 +106,13 @@ def _tensorize(self, value):
             if isinstance(value, PIL.Image.Image):
                 value = np.asarray(value)
 
-        with jax.default_device(self.device_mapping[self.device]):
+        # using global variable since `jaxlib.xla_extension.Device` is not serializable neither
+        # with `pickle` nor with `dill`, so we need to use a global variable instead
+        global DEVICE_MAPPING
+        if not DEVICE_MAPPING:
```
lhoestq, line 112:
```suggestion
        if DEVICE_MAPPING is None:
```

src/datasets/formatting/jax_formatter.py
```diff
@@ -15,25 +15,59 @@
 # Lint as: python3
 import sys
 from collections.abc import Mapping
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Dict
```
lhoestq, line 18:
```suggestion
from typing import TYPE_CHECKING, Dict, Optional
```

alvarobartt:
> Looks all good now thank you !
> 
> Is there anything else you wanted to add ? Otherwise I think it's ready for merge

Nothing else to add, I've already applied your suggestions. Thanks for your input/feedback @lhoestq :hugs:

lhoestq (approved):
let's merge then :)

PR #5551, opened 2023-02-20
Suggest scikit-learn instead of sklearn

osbm:
This is kinda unimportant fix but, the suggested `pip install sklearn` does not work.

The current error message if sklearn is not installed:

```
ImportError: To be able to use [dataset name], you need to install the following dependency: sklearn.
Please install it using 'pip install sklearn' for instance.
```


julien-c:
good catch!

lhoestq (approved):
indeed thanks !

PR #5554, opened 2023-02-20
Add resampy dep

lhoestq:
In librosa 0.10 they removed the `resmpy` dependency and set it to optional.

However it is necessary for resampling. I added it to the "audio" extra dependencies.


mariosasko:
I think we should also suggest installing `resampy` in the error message thrown by the Audio feature when `librosa` is not installed.

lhoestq:
exploring a better solution at https://github.com/huggingface/datasets/pull/5556

PR #5556, opened 2023-02-21
Use default audio resampling type

lhoestq:
...instead of relying on the optional librosa dependency `resampy`.

It was only used for `_decode_non_mp3_file_like` anyway and not for the other ones - removing it fixes consistency between decoding methods (except torchaudio decoding)

Therefore I think it is a better solution than adding `resampy` as a dependency in https://github.com/huggingface/datasets/pull/5554

cc @polinaeterna


lhoestq:
If it's good for you @polinaeterna I'd like to merge it and then run the `transformers` CI to see if it changes anything

polinaeterna (approved):
@lhoestq lgtm thank you!

PR #5560, opened 2023-02-21
Ensure last tqdm update in `map`

mariosasko:
This PR modifies `map` to:
* ensure the TQDM bar gets the last progress update
* when a map function fails, avoid throwing a chained exception in the single-proc mode


lhoestq (approved):
thanks !

lhoestq:
I just tried on colab and it didn't finish the progress bar for some reason.

Maybe we need to call `pbar.close()` before `return`

lhoestq:
(just added .close() - let me try quickly if it works now)

lhoestq:
it worked ! :)

image

lhoestq (approved):

mariosasko:
(the BadZipFile error is unrelated to the changes)

PR #5562, opened 2023-02-22
Update csv.py

XDoubleU:
Removed mangle_dup_cols=True from BuilderConfig.

It triggered following deprecation warning:
/usr/local/lib/python3.8/dist-packages/datasets/download/streaming_download_manager.py:776: FutureWarning: the 'mangle_dupe_cols' keyword is deprecated and will be removed in a future version. Please take steps to stop the use of 'mangle_dupe_cols'
  return pd.read_csv(xopen(filepath_or_buffer, "rb", use_auth_token=use_auth_token), **kwargs)

Further documentation of pandas: https://pandas.pydata.org/docs/whatsnew/v1.4.0.html#mangle-dupe-cols-in-read-csv-no-longer-renames-unique-columns-conflicting-with-target-names

At first sight it seems like this flag is resolved internally, it might need some more research.

### DIFF
--- a/src/datasets/packaged_modules/csv/csv.py
+++ b/src/datasets/packaged_modules/csv/csv.py
@@ -15,7 +15,7 @@ from datasets.table import table_cast
 logger = datasets.utils.logging.get_logger(__name__)
 
 _PANDAS_READ_CSV_NO_DEFAULT_PARAMETERS = ["names", "prefix"]
-_PANDAS_READ_CSV_DEPRECATED_PARAMETERS = ["warn_bad_lines", "error_bad_lines"]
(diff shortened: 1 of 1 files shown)


lhoestq (commented):
Thanks for noticing and removing it.

Can you also remove the use of `mangle_dupe_cols ` in `pd_read_csv_kwargs` later in that code ?

XDoubleU:
Removed it :)

lhoestq (approved):
thanks !

lhoestq (changes requested):
Oh actually I just found a better solution: 

we can keep the parameter to keep compatibility with older `pandas` versions, and simply add `"mangle_dupe_cols"` to `_PANDAS_READ_CSV_DEPRECATED_PARAMETERS ` earlier in the code - sorry I didn't see that earlier ;)

XDoubleU:
Changed it :)

PR #5565, opened 2023-02-22
Add writer_batch_size for ArrowBasedBuilder

lhoestq:
This way we can control the size of the record_batches/row_groups of arrow/parquet files.

This can be useful for `datasets-server` to keep control of the row groups size which can affect random access performance for audio/image/video datasets

Right now having 1,000 examples per row group cause some image datasets to be pretty slow for random access (e.g. 4seconds for `beans` to get 20 rows)

### DIFF
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -251,6 +251,11 @@ class DatasetBuilder:
             `os.path.join(data_dir, "**")` as `data_files`.
             For builders that require manual download, it must be the path to the local directory containing the
             manually downloaded data.
+        writer_batch_size (`int`, *optional*):
+            Batch size used by the ArrowWriter.
+            It defines the number of samples that are kept in memory before writing them
+            and also the length of the arrow chunks.
+            None means that the ArrowWriter will use its default value.
         name (`str`): Configuration name for the dataset.
 
             <Deprecated version="2.3.0">
@@ -276,6 +281,13 @@ class DatasetBuilder:
     # Optional default config name to be used when name is None
     DEFAULT_CONFIG_NAME = None
 
+    # Default batch size used by the ArrowWriter
+    # It defines the number of samples that are kept in memory before writing them
+    # and also the length of the arrow chunks
+    # None means that the ArrowWriter will use its default value
+    DEFAULT_WRITER_BATCH_SIZE = None
+
+
     def __init__(
         self,
         cache_dir: Optional[str] = None,
@@ -288,6 +300,7 @@ class DatasetBuilder:
         repo_id: Optional[str] = None,
         data_files: Optional[Union[str, list, dict, DataFilesDict]] = None,
         data_dir: Optional[str] = None,
+        writer_batch_size: Optional[int] = None,
         name="deprecated",
         **config_kwargs,
     ):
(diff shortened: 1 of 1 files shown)


src/datasets/builder.py
```diff
@@ -1987,6 +1979,10 @@ def _download_and_prepare(self, dl_manager, verification_mode, **prepare_splits_
                 "`DirectRunner` (you may run out of memory). \nExample of usage: "
                 f"\n\t`{usage_example}`"
             )
+        if self._writer_batch_size is not None:
+            logger.warning(
+                "Writer batch size is not supported for beam pipelines yet. Using the default chunk size for writing."
+            )
```
lhoestq, line 1985: leaving a warning for the BeamBasedBuilder until we merge https://github.com/huggingface/datasets/pull/5364

lhoestq:
Can you review this @mariosasko ? since Albert is off

src/datasets/builder.py
```diff
@@ -276,6 +281,12 @@ class DatasetBuilder:
     # Optional default config name to be used when name is None
     DEFAULT_CONFIG_NAME = None
 
+    # Default batch size used by the ArrowWriter
+    # It defines the number of samples that are kept in memory before writing them
+    # and also the length of the arrow chunks
```
mariosasko, line 286: From looking at the source, only the `"the length of the arrow chunks"` part seems to be true for the `ArrowWriter` due to using `write_table` there, which always writes to disk (with `max_chunksize=writer_batch_size`)

src/datasets/builder.py
```diff
@@ -1987,6 +1979,10 @@ def _download_and_prepare(self, dl_manager, verification_mode, **prepare_splits_
                 "`DirectRunner` (you may run out of memory). \nExample of usage: "
                 f"\n\t`{usage_example}`"
             )
+        if self._writer_batch_size is not None:
+            logger.warning(
+                "Writer batch size is not supported for beam pipelines yet. Using the default chunk size for writing."
```
mariosasko, line 1984:
```suggestion
                "`writer_batch_size` is not supported for beam pipelines yet. Using the default chunk size for writing."
```

src/datasets/builder.py
lhoestq: I naively copy pasted the comment from GeneratorBasedBuidler.

What do you think of 
```python
    # Default batch size used by the ArrowWriter.
    # It defines the maximum size of the arrow record batches (or parquet row groups).
    # Setting it to a low value may also help reducing RAM usage in certain cases
    # by flushing the data on disk more frequently.
```

mariosasko: Yes, either this or we could check the table's `num_rows` in `write_table` before passing it to the `pa_writer`.

lhoestq: I'll think I'd keep it as it is right now to avoid changing the logic (out of scope of this PR).

Lmk if it's fine for you

mariosasko: Yes, that's fine

mariosasko (commented):
LGTM!

mariosasko (approved):

Issue #5568, opened 2023-02-23
dataset.to_iterable_dataset() loses useful info like dataset features


Hubert-Bonisseur:
### Describe the bug

Hello,
I like the new `to_iterable_dataset` feature but I noticed something that seems to be missing.

When using `to_iterable_dataset` to transform your map style dataset into iterable dataset, you lose valuable metadata like the features. 

These metadata are useful if you want to interleave iterable datasets, cast columns etc.

### Steps to reproduce the bug

```python
dataset = load_dataset("lhoestq/demo1")["train"]
print(dataset.features)
# {'id': Value(dtype='string', id=None), 'package_name': Value(dtype='string', id=None), 'review': Value(dtype='string', id=None), 'date': Value(dtype='string', id=None), 'star': Value(dtype='int64', id=None), 'version_id': Value(dtype='int64', id=None)}
dataset = dataset.to_iterable_dataset()
print(dataset.features)
# None
```

### Expected behavior

Keep the relevant information

### Environment info

datasets==2.10.0

lhoestq:
Hi ! Oh good catch. I think the features should be passed to `IterableDataset.from_generator()` in `to_iterable_dataset()` indeed.

Setting this as a good first issue if someone would like to contribute, otherwise we can take care of it :)

Hubert-Bonisseur:
#self-assign

YellowFoxH4XOR:
seems like the feature parameter is missing from `return IterableDataset.from_generator(Dataset._iter_shards, gen_kwargs={"shards": shards})`  hence it defaults to None.

PR #5569, opened 2023-02-23
pass the dataset features to the IterableDataset.from_generator function

Hubert-Bonisseur:
[5558](https://github.com/huggingface/datasets/issues/5568)


mariosasko (approved):
Thanks for fixing this!

lhoestq:
Hmm I think we need to add more tests. Not sure what would happen with :
- decodable features that may end up decoded twice 
- formatted datasets 

I'd be in favor of reverting this until we checked those

Issue #5488, opened 2023-01-31
Error loading MP3 files from CommonVoice


kradonneoh:
### Describe the bug

When loading a CommonVoice dataset with `datasets==2.9.0` and `torchaudio>=0.12.0`, I get an error reading the audio arrays: 

```python
---------------------------------------------------------------------------
LibsndfileError                           Traceback (most recent call last)
~/.local/lib/python3.8/site-packages/datasets/features/audio.py in _decode_mp3(self, path_or_file)
    310             try:  # try torchaudio anyway because sometimes it works (depending on the os and os packages installed)
--> 311                 array, sampling_rate = self._decode_mp3_torchaudio(path_or_file)
    312             except RuntimeError:

~/.local/lib/python3.8/site-packages/datasets/features/audio.py in _decode_mp3_torchaudio(self, path_or_file)
    351 
--> 352         array, sampling_rate = torchaudio.load(path_or_file, format="mp3")
    353         if self.sampling_rate and self.sampling_rate != sampling_rate:

~/.local/lib/python3.8/site-packages/torchaudio/backend/soundfile_backend.py in load(filepath, frame_offset, num_frames, normalize, channels_first, format)
    204     """
--> 205     with soundfile.SoundFile(filepath, "r") as file_:
    206         if file_.format != "WAV" or normalize:

~/.local/lib/python3.8/site-packages/soundfile.py in __init__(self, file, mode, samplerate, channels, subtype, endian, format, closefd)
    654                                          format, subtype, endian)
--> 655         self._file = self._open(file, mode_int, closefd)
    656         if set(mode).issuperset('r+') and self.seekable():

~/.local/lib/python3.8/site-packages/soundfile.py in _open(self, file, mode_int, closefd)
   1212             err = _snd.sf_error(file_ptr)
-> 1213             raise LibsndfileError(err, prefix="Error opening {0!r}: ".format(self.name))
   1214         if mode_int == _snd.SFM_WRITE:

LibsndfileError: Error opening <_io.BytesIO object at 0x7fa539462090>: File contains data in an unknown format.

```

I assume this is because there's some issue with the mp3 decoding process. I've verified that I have `ffmpeg>=4` (on a Linux distro), which appears to be the fallback backend for `torchaudio,` (at least according to #4889).

### Steps to reproduce the bug

```python
dataset = load_dataset("mozilla-foundation/common_voice_11_0", "be", split="train")
dataset[0]
```

### Expected behavior

[...]

PR #5573, opened 2023-02-23
Use soundfile for mp3 decoding instead of torchaudio

polinaeterna:
I've removed `torchaudio` completely and switched to use `soundfile` for everything. With the new version of `soundfile` package this should work smoothly because the `libsndfile` C library is bundled, in Linux wheels too. 

Let me know if you think it's too harsh and we should continue to support `torchaudio` decoding.

I decided that we can drop it completely because:
1. it's always something wrong with `torchaudio` (for example recently https://github.com/huggingface/datasets/issues/5488 )
2. the results of mp3 decoding are different depending on `torchaudio` version 
3. `soundfile` is slightly faster then the latest `torchaudio`
4. anyway users can pass any custom decoding function with any library they want if needed (worth putting a snippet in the docs).

cc @sanchit-gandhi @vaibhavad

### DIFF
--- a/docs/source/installation.md
+++ b/docs/source/installation.md
@@ -68,7 +68,7 @@ pip install datasets[audio]
 <Tip warning={true}>
 
 To decode mp3 files you should have the `libsndfile` system library of at least version 1.1.0. Normally, it's bundled into the python [`soundfile`](https://github.com/bastibe/python-soundfile) package which is installed as an extra audio dependency of 🤗 Datasets.
-On Linux, the required version of `libsndfile` is bundled into the `soundfile` wheel since version 0.12.0 of `soundfile`. You can check which version of `libsndfile` is used by `soundfile` running the following code:
+On Linux, the required version of `libsndfile` is bundled into the `soundfile` wheel since version 0.12.0 of `soundfile`. You can run the following command to check which version of `libsndfile` is used by `soundfile`:
 
 ```bash
 python -c "import soundfile; print(soundfile.__libsndfile_version__)"
--- a/docs/source/installation.md
+++ b/docs/source/installation.md
@@ -67,8 +67,8 @@ pip install datasets[audio]
 
 <Tip warning={true}>
 
-To decode mp3 files you should have the `libsndfile` system library of at least version 1.1.0. Normally, it's bundled into the python [`soundfile`](https://github.com/bastibe/python-soundfile) package which is installed as an extra audio dependency of 🤗 Datasets.
-On Linux, the required version of `libsndfile` is bundled into the `soundfile` wheel since version 0.12.0 of `soundfile`. You can run the following command to check which version of `libsndfile` is used by `soundfile`:
+To decode mp3 files, you need to have at least version 1.1.0 of the `libsndfile` system library. Usually, it's bundled with the python [`soundfile`](https://github.com/bastibe/python-soundfile) package, which is installed as an extra audio dependency for 🤗 Datasets.
+For Linux, the required version of `libsndfile` is bundled with `soundfile` starting from version 0.12.0. You can run the following command to determine which version of `libsndfile` is being used by `soundfile`:
 
 ```bash
 python -c "import soundfile; print(soundfile.__libsndfile_version__)"
--- a/docs/source/installation.md
+++ b/docs/source/installation.md
@@ -67,31 +67,15 @@ pip install datasets[audio]
 
 <Tip warning={true}>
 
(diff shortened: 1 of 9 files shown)


docs/source/installation.md
```diff
@@ -67,31 +67,15 @@ pip install datasets[audio]
 
 <Tip warning={true}>
 
-On Linux, non-Python dependency on `libsndfile` package must be installed manually, using your distribution package manager, for example:
+To decode mp3 files you should have the `libsndfile` system library of at least version 1.1.0. Normally, it's bundled into the python [`soundfile`](https://github.com/bastibe/python-soundfile) package which is installed as an extra audio dependency of 🤗 Datasets.
+On Linux, the required version of `libsndfile` is bundled into the `soundfile` wheel since version 0.12.0 of `soundfile`. You can check which version of `libsndfile` is used by `soundfile` running the following code:
```
mariosasko, line 71:
```suggestion
On Linux, the required version of `libsndfile` is bundled into the `soundfile` wheel since version 0.12.0 of `soundfile`. You can run the following command to check which version of `libsndfile` is used by `soundfile`:
```

mariosasko (commented):
Good stuff!

polinaeterna:
@mariosasko thank you for the review! do you have any idea why `test_hash_torch_tensor` fails on "ubuntu-latest deps-minimum"? I removed the `torchaudio<0.12.0` test dependency so it uses the latest `torch` now, might it be connected?

docs/source/installation.md
stevhliu, line 71:
```suggestion
To decode mp3 files, you need to have at least version 1.1.0 of the `libsndfile` system library. Usually, it's bundled with the python [`soundfile`](https://github.com/bastibe/python-soundfile) package, which is installed as an extra audio dependency for 🤗 Datasets.
For Linux, version 0.12.0 or later of the `soundfile` wheel includes the required version of `libsndfile`. You can determine which version of libsndfile is being used by soundfile by running the following code:
```

stevhliu (approved):
Nice, thanks for also updating the docs!

mariosasko:
@polinaeterna The failure is due to `torch.from_numpy` not being picklable in newer versions of PyTorch. You can replace the current definition of `_save_tensor` in `utils/py_utils.py` with the following one to fix it: 

```python
@pklregister(obj_type)
def _save_tensor(pickler, obj):
    def _create_tensor(np_array):
        return torch.from_numpy(np_array)

    dill_log(pickler, f"To: {obj}")
    args = (obj.detach().cpu().numpy(),)
    pickler.save_reduce(torch.from_numpy, args, obj=obj)
    dill_log(pickler, "# To")
    return
```

lhoestq (approved):
Thanks a lot ! you can merge when the CI is green (either with mario's fix or by skipping the torch test for recent torch versions)

src/datasets/features/audio.py
```diff
@@ -150,20 +148,49 @@ def decode_example(
...
-                array, sampling_rate = self._decode_non_mp3_path_like(
-                    path, "opus", token_per_repo_id=token_per_repo_id
+
+        try:
+            import librosa
+            import soundfile as sf
+        except ImportError as err:
+            raise ImportError("To support decoding audio files, please install 'librosa' and 'soundfile'.") from err
+
+        audio_format = os.path.splitext(path)[1].lstrip(".").lower() if path is not None else None
```
mariosasko, line 158: I think we need to use `xsplitext` (from `download/streaming_download_manager.py`) here to account for URLs

```suggestion
        audio_format = xsplitext(path)[1][1:].lower() if path is not None else None
```

mariosasko (commented):
Some additional comments

src/datasets/features/audio.py
```diff
@@ -150,20 +148,49 @@ def decode_example(
...
+            if not config.IS_OPUS_SUPPORTED:
+                raise RuntimeError(
+                    "Decoding 'opus' files requires system library 'libsndfile'>=1.0.31, "
+                    'You can try to update `soundfile` python library: `pip install "soundfile>=0.12.1"`. '
+                )
+        if audio_format == "mp3":
+            if not config.IS_MP3_SUPPORTED:
+                raise RuntimeError(
+                    "Decoding 'mp3' files requires system library 'libsndfile'>=1.1.0, "
+                    'You can try to update `soundfile` python library: `pip install "soundfile>=0.12.1"`. '
```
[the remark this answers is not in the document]
polinaeterna: lol sorry i'm blind

mariosasko (approved):
Looks all good now!

PR #5582, opened 2023-02-27
Add column_names to IterableDataset

### LINKED ISSUE #5383: IterableDataset missing column_names, differs from Dataset interface
### Describe the bug

The documentation on [Stream](https://huggingface.co/docs/datasets/v1.18.2/stream.html) seems to imply that IterableDataset behaves just like a Dataset. However, examples like
```
dataset.map(augment_data, batched=True, remove_columns=dataset.column_names, ...)
```
will not work because `.column_names` does not exist on IterableDataset. I cannot find any clear explanation on why this is not available, is it an oversight? We do have `iterable_ds.features` available.

See above

### Expected behavior

Dataset and IterableDataset would be expected to have the same interface, with any differences noted in the documentation.

### Environment info

n/a

patrickloeber:
This PR closes #5383 

* Add column_names property to IterableDataset
* Add multiple tests for this new property

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -1387,9 +1387,7 @@ class IterableDataset(DatasetInfoMixin):
         ['text', 'label']
         ```
         """
-        if self._info.features is not None:
-            return list(self._info.features.keys())
-        return None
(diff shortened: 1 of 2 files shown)


lhoestq (approved):
thanks !

PR #5583, opened 2023-02-27
Do no write index by default when exporting a dataset

mariosasko:
Ensures all the writers that use Pandas for conversion (JSON, CSV, SQL) do not export `index` by default (https://github.com/huggingface/datasets/pull/5490 only did this for CSV)


lhoestq (approved):
nice !

Issue #5507, opened 2023-02-06
Optimise behaviour in respect to indices mapping


mariosasko:
_Originally [posted](https://huggingface.slack.com/archives/C02V51Q3800/p1675443873878489?thread_ts=1675418893.373479&cid=C02V51Q3800) on Slack_

Considering all this, perhaps for Datasets 3.0, we can do the following:
* have `continuous=True` by default in `.shard`  (requested in the survey and makes more sense for us since it doesn't create an indices mapping)
* allow calling `save_to_disk` on "unflattened" datasets
* remove "hidden" expensive calls in `save_to_disk`, `unique`, `concatenate_datasets`, etc. For instance, instead of silently calling `flatten_indices`  where it's needed, it's probably better to be explicit (considering how expensive these ops can be) and raise an error instead

PR #5588, opened 2023-02-28
Flatten dataset on the fly in `save_to_disk`

mariosasko:
Flatten a dataset on the fly in `save_to_disk` instead of doing it with `flatten_indices`  to avoid creating an additional cache file.

(this is one of the sub-tasks in https://github.com/huggingface/datasets/issues/5507)


lhoestq (approved):
Very nice ! LGTM

PR #5598, opened 2023-03-01
Fix push_to_hub with no dataset_infos

lhoestq:
As reported in https://github.com/vijaydwivedi75/lrgb/issues/10, `push_to_hub` fails if the remote repository already exists and has a README.md without `dataset_info` in the YAML tags

cc @clefourrier


mariosasko (approved):
LGTM!

PR #5605, opened 2023-03-03
Update README logo

gary149:


mariosasko (commented):
Nice! 

Can you delete the previous logo's file before merging as we no longer need it?

gary149:
Are you sure it's safe to remove? https://github.com/huggingface/datasets/pull/3866

mariosasko (approved):
I guess we can keep it then to avoid breaking the existing forks 🙂. 

PS: the failing test is flaky, so feel free to ignore it

PR #5607, opened 2023-03-03
Fix outdated `verification_mode` values

polinaeterna:
~I think it makes sense not to save `dataset_info.json` file to a dataset cache directory when loading dataset with `verification_mode="no_checks"` because otherwise when next time the dataset is loaded **without** `verification_mode="no_checks"`, it will be loaded successfully, despite some values in info might not correspond to the ones in the repo which was the reason for using `verification_mode="no_checks"` first.~

Updated values of `verification_mode` to the current ones in some places ("none" -> "no_checks", "all" -> "all_checks")

### DIFF
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -719,7 +719,7 @@ class DatasetBuilder:
         ```
         """
         if ignore_verifications != "deprecated":
-            verification_mode = "none" if ignore_verifications else "full"
+            verification_mode = "no_checks" if ignore_verifications else "all_checks"
             warnings.warn(
                 "'ignore_verifications' was deprecated in favor of 'verification_mode' in version 2.9.1 and will be removed in 3.0.0.\n"
                 f"You can remove this warning by passing 'verification_mode={verification_mode}' instead.",
@@ -1080,7 +1080,7 @@ class DatasetBuilder:
         ```
         """
         if ignore_verifications != "deprecated":
-            verification_mode = "none" if ignore_verifications else "full"
+            verification_mode = "no_checks" if ignore_verifications else "all_checks"
             warnings.warn(
                 "'ignore_verifications' was deprecated in favor of 'verification' in version 2.9.1 and will be removed in 3.0.0.\n"
                 f"You can remove this warning by passing 'verification_mode={verification_mode}' instead.",
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -358,14 +358,13 @@ class DatasetBuilder:
             lock_path = os.path.join(self._cache_dir_root, self._cache_dir.replace(os.sep, "_") + ".lock")
             with FileLock(lock_path):
                 if os.path.exists(self._cache_dir):  # check if data exist
-                    if len(os.listdir(self._cache_dir)) > 0 and os.path.exists(
-                        path_join(self._cache_dir, config.DATASET_INFO_FILENAME)
-                    ):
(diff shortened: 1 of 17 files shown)


src/datasets/builder.py
```diff
@@ -357,10 +357,11 @@ def __init__(
             os.makedirs(self._cache_dir_root, exist_ok=True)
             lock_path = os.path.join(self._cache_dir_root, self._cache_dir.replace(os.sep, "_") + ".lock")
             with FileLock(lock_path):
-                if os.path.exists(self._cache_dir):  # check if data exist
-                    if len(os.listdir(self._cache_dir)) > 0:
-                        logger.info("Overwrite dataset info from restored data version.")
-                        self.info = DatasetInfo.from_directory(self._cache_dir)
+                if os.path.exists(self._cache_dir) > 0:  # check if data exist
+                    if len(os.listdir(self._cache_dir)):
```
lhoestq, line 361:
```suggestion
                if os.path.exists(self._cache_dir):  # check if data exist
                    if len(os.listdir(self._cache_dir)) > 0:
```

src/datasets/builder.py
```diff
@@ -880,7 +881,8 @@ def incomplete_dir(dirname):
                     self.info.download_checksums = dl_manager.get_recorded_sizes_checksums()
                     self.info.size_in_bytes = self.info.dataset_size + self.info.download_size
                     # Save info
-                    self._save_info()
+                    if verification_mode is not VerificationMode.NO_CHECKS:
+                        self._save_info()
```
lhoestq, line 885: We should always save the info. For example it stores the feature types and the arrow shards lengths.

lhoestq (commented):
actually we need the dataset info

src/datasets/builder.py
polinaeterna: hm true...
what should we do then? I don't think it's fine that we get different results running the same command, like:
```python
load_dataset("ds_name")  # error - failed verifications
# then
load_dataset("ds_name", verification_mode="no_checks")  # successfully loaded
# then
load_dataset("ds_name")  # successfully loaded despite it was an error before
```
am I missing something?

lhoestq: I don't think we can do anything tbh - e.g. reloading a cached dataset works even if you don't have the downloaded raw files anymore so we can't check for their checksums

lhoestq (commented):
Yea the fix for the values is definitely welcome, thank you :)

src/datasets/builder.py
```diff
@@ -358,8 +358,10 @@ def __init__(
...
             with FileLock(lock_path):
                 if os.path.exists(self._cache_dir):  # check if data exist
-                    if len(os.listdir(self._cache_dir)) > 0:
-                        logger.info("Overwrite dataset info from restored data version.")
+                    if len(os.listdir(self._cache_dir)) > 0 and os.path.exists(
+                        path_join(self._cache_dir, config.DATASET_INFO_FILENAME)
+                    ):
+                        logger.info("Overwrite dataset info from restored data version if exists.")
                         self.info = DatasetInfo.from_directory(self._cache_dir)
                     else:  # dir exists but no data, remove the empty dir as data aren't available anymore
```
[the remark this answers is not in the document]
polinaeterna: thanks for noticing, fixed it, can I merge?

lhoestq (approved):
Thanks :)

mariosasko (approved):
Good catch!

Issue #5606, opened 2023-03-03
Add `Dataset.to_list` to the API


mariosasko:
Since there is `Dataset.from_list` in the API, we should also add `Dataset.to_list` to be consistent.
 
Regarding the implementation, we can re-use `Dataset.to_dict`'s code and replace the `to_pydict` calls with `to_pylist`.

kyoto7250:
Hello, I have an interest in this issue.
Is the `Dataset.to_dict` you are describing correct in the code here?

https://github.com/huggingface/datasets/blob/35b789e8f6826b6b5a6b48fcc2416c890a1f326a/src/datasets/arrow_dataset.py#L4633-L4667

mariosasko:
Yes, this is where `Dataset.to_dict` is defined.

kyoto7250:
#self-assign

PR #5142, opened 2022-10-20
Deprecate num_proc parameter in  DownloadManager.extract

ayushthe1:
fixes #5132 : Deprecated the `num_proc` parameter in `DownloadManager.extract` by passing `num_proc` parameter to `map_nested` .

PR #5611, opened 2023-03-06
add Dataset.to_list

kyoto7250:
close https://github.com/huggingface/datasets/issues/5606

This PR is for adding the `Dataset.to_list` method.
Thank you in advance.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -4666,6 +4666,42 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                 for offset in range(0, len(self), batch_size)
             )
 
+    def to_list(self, batch_size: Optional[int] = None, batched: bool = False) -> Union[list, Iterator[list]]:
+        """Returns the dataset as a Python list. Can also return a generator for large datasets.
+
+        Args:
+            batched (`bool`):
+                Set to `True` to return a generator that yields the dataset as batches
+                of `batch_size` rows. Defaults to `False` (returns the whole datasets once).
+            batch_size (`int`, *optional*): The size (number of rows) of the batches if `batched` is `True`.
+                Defaults to `datasets.config.DEFAULT_MAX_BATCH_SIZE`.
+
+        Returns:
+            `list` or `Iterator[list]`
+
+        Example:
+
+        ```py
+        >>> ds.to_list()
+        ```
+        """
+        if not batched:
+            return query_table(
+                table=self._data,
+                key=slice(0, len(self)),
+                indices=self._indices if self._indices is not None else None,
+            ).to_pylist()
+        else:
+            batch_size = batch_size if batch_size else config.DEFAULT_MAX_BATCH_SIZE
+            return (
+                query_table(
+                    table=self._data,
+                    key=slice(offset, offset + batch_size),
+                    indices=self._indices if self._indices is not None else None,
+                ).to_pylist()
+                for offset in range(0, len(self), batch_size)
+            )
+
     def to_json(
         self,
         path_or_buf: Union[PathLike, BinaryIO],
--- a/tests/test_arrow_dataset.py
+++ b/tests/test_arrow_dataset.py
@@ -2171,6 +2171,36 @@ class BaseDatasetTest(TestCase):
                         self.assertIsInstance(dset_to_dict[col_name], list)
                         self.assertEqual(len(dset_to_dict[col_name]), len(dset))
 
+    def test_to_list(self, in_memory):
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            # Batched
+            with self._create_dummy_dataset(in_memory, tmp_dir, multiple_columns=True) as dset:
+                batch_size = dset.num_rows - 1
+                to_list_generator = dset.to_list(batched=True, batch_size=batch_size)
+
+                for batch in to_list_generator:
+                    self.assertIsInstance(batch, list)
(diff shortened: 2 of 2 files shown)


mariosasko:
Hi, thanks for working on this! `Table.to_pylist` requires PyArrow 7.0+, and our minimal version requirement is 6.0, so we need to bump the version requirement to avoid CI failure. I'll do this in a separate PR.

mariosasko (approved):
CI is green now. Thanks!

src/datasets/arrow_dataset.py
```diff
@@ -4666,6 +4666,42 @@ def to_dict(self, batch_size: Optional[int] = None, batched: bool = False) -> Un
                 for offset in range(0, len(self), batch_size)
             )
 
+    def to_list(self, batch_size: Optional[int] = None, batched: bool = False) -> Union[list, Iterator[list]]:
```
lhoestq, line 4669: The order of the arguments in the function signature don't match the order in the docstring
```suggestion
    def to_list(self, batched: bool = False, batch_size: Optional[int] = None) -> Union[list, Iterator[list]]:
```

src/datasets/arrow_dataset.py
```diff
@@ -4666,6 +4666,42 @@ def to_dict(self, batch_size: Optional[int] = None, batched: bool = False) -> Un
...
+        else:
+            batch_size = batch_size if batch_size else config.DEFAULT_MAX_BATCH_SIZE
+            return (
+                query_table(
+                    table=self._data,
+                    key=slice(offset, offset + batch_size),
+                    indices=self._indices if self._indices is not None else None,
+                ).to_pylist()
+                for offset in range(0, len(self), batch_size)
+            )
```
lhoestq, line 4703: You can already use `dataset.iter(batch_size=batch_size)` to get an iterator.

A function named `to_list()` should return a list IMO

lhoestq (commented):
thanks, a few comments:

src/datasets/arrow_dataset.py
mariosasko: This is for consistency with `Dataset.to_dict`

src/datasets/arrow_dataset.py
mariosasko: The same thing here. We could remove these params here and then deprecate them in `to_dict` (to be aligned), but I don't think this is worth the hassle, considering the default behavior is the "expected" one (returning an entire list/dict). Hence, I think we can merge the PR in the current state.

mariosasko: Oops, I noticed your 2nd message after replying. I agree this is not the best design (e.g., I wasn't even aware of these parameters before this PR 😄), so let's deprecate these parameters.

@kyoto7250 Let me know if you'd like to address this, or I can do it later this week.

kyoto7250: @mariosasko 

Of course, I want to try it!

So it means that we should return a batch of 'loaded objects' (not an iterator), correct?
And, should we change both `Dataset.to_dict` and `Dataset.to_list` in this PR?

mariosasko: No, it means `to_dict`/`to_list` should always return a dictionary/list, not an iterator. So we need to remove `batch_size` and `batched` in `to_list` and deprecate these parameters in `to_dict` (check https://github.com/huggingface/datasets/pull/5142 to see how we deprecate params in `datasets`).

kyoto7250: I added two commits.

Could you please tell me in which version this feature will be deprecated and in which version this feature will be removed, for the sake of displaying the message?

src/datasets/arrow_dataset.py
```diff
@@ -4637,6 +4637,11 @@ def to_dict(self, batch_size: Optional[int] = None, batched: bool = False) -> Un
             batched (`bool`):
                 Set to `True` to return a generator that yields the dataset as batches
                 of `batch_size` rows. Defaults to `False` (returns the whole datasets once).
+
+            <Deprecated>
+            Use `Dataset.iter(batch_size=batch_size)` instead.
+            </Deprecated>
```
mariosasko, line 4643:
```suggestion
                <Deprecated version="2.11.0">
                
                Use `.iter(batch_size=batch_size)` followed by `.to_dict()` on the individual batches instead.
            
                </Deprecated>
```

src/datasets/arrow_dataset.py
```diff
@@ -4656,6 +4661,10 @@ def to_dict(self, batch_size: Optional[int] = None, batched: bool = False) -> Un
                 indices=self._indices if self._indices is not None else None,
             ).to_pydict()
         else:
+            warnings.warn(
+                "'batched' was deprecated. Use `Dataset.iter(batch_size=batch_size)` instead.",
+                FutureWarning,
+            )
```
mariosasko, line 4667:
```suggestion
            if batched != "deprecated":
                warnings.warn(
                    "'batched' was deprecated in version 2.11.0 and will be removed in version 3.0.0. Use `.iter(batch_size=batch_size)` followed by `.to_dict()` on the individual batches instead.",
                    FutureWarning,
                )
            else:
                batched = False 
```
We should have this warning at the beginning of the function. Plus, the default value of `batched` (in the function's signature) should be `"deprecated"`

src/datasets/arrow_dataset.py
mariosasko: You can find the versions in my review.

mariosasko (approved):
Thanks! Will merge if it looks good to you @lhoestq.

PR #5614, opened 2023-03-06
Fix archive fs test

lhoestq:


tests/test_filesystem.py
```diff
@@ -45,7 +45,7 @@ def test_compression_filesystems(compression_fs_class, gz_file, bz2_file, lz4_fi
     assert isinstance(fs, compression_fs_class)
     expected_filename = os.path.basename(input_path)
     expected_filename = expected_filename[: expected_filename.rindex(".")]
-    assert fs.ls("/") == [expected_filename]
+    assert fs.ls("/", detail=False) == [expected_filename]
```
albertvillanova, line 48: Since we never use `fs.ls` in the datasets library (but `fs.glob` for ZIP archive), what about aligning the test with what we really use from `fsspec`? This way we will avoid unnecessarily breaking our CI with any other future breaking change on `fs.ls`.
```suggestion
    assert fs.glob("*") == [expected_filename]
```

docs/source/filesystems.mdx
```diff
@@ -191,7 +191,7 @@ Remember to define your credentials in your [FileSystem instance](#set-up-your-c
 List files from a cloud storage with your FileSystem instance `fs`, using `fs.ls`:
 
 ```py
->>> fs.ls("my-private-datasets/imdb/train")
+>>> fs.ls("my-private-datasets/imdb/train", detail=False)
```
albertvillanova, line 194: This is indeed an example about `fsspec` we provide in our docs, but we really do not use `fs.ls` in the library. I agree that the example is informative/useful for the end user though...

albertvillanova (approved):
Thanks for the CI fix. Just a comment/suggestion below.

PR #5620, opened 2023-03-07
Bump pyarrow to 8.0.0

lhoestq:
Fix those for Pandas 2.0 (tested [here](https://github.com/huggingface/datasets/actions/runs/4346221280/jobs/7592010397) with pandas==2.0.0.rc0):

```python
FAILED tests/test_arrow_dataset.py::BaseDatasetTest::test_to_parquet_in_memory - ImportError: Unable to find a usable engine; tried using: 'pyarrow', 'fastparquet'.
A suitable version of pyarrow or fastparquet is required for parquet support.
Trying to import the above resulted in these errors:
 - Pandas requires version '7.0.0' or newer of 'pyarrow' (version '6.0.1' currently installed).
 - Missing optional dependency 'fastparquet'. fastparquet is required for parquet support. Use pip or conda to install fastparquet.
FAILED tests/test_arrow_dataset.py::BaseDatasetTest::test_to_parquet_on_disk - ImportError: Unable to find a usable engine; tried using: 'pyarrow', 'fastparquet'.
A suitable version of pyarrow or fastparquet is required for parquet support.
Trying to import the above resulted in these errors:
 - Pandas requires version '7.0.0' or newer of 'pyarrow' (version '6.0.1' currently installed).
 - Missing optional dependency 'fastparquet'. fastparquet is required for parquet support. Use pip or conda to install fastparquet.
===== 2 failed, 2137 passed, 18 skipped, 32 warnings in 212.76s (0:03:32) ======
```

EDIT: also for performance - with 8.0 we can use `.to_reader()`

### DIFF
--- a/.github/conda/meta.yaml
+++ b/.github/conda/meta.yaml
@@ -15,7 +15,7 @@ requirements:
     - python
     - pip
     - numpy >=1.17
-    - pyarrow >=6.0.0
+    - pyarrow >=8.0.0
     - python-xxhash
     - dill
     - pandas
@@ -32,7 +32,7 @@ requirements:
     - python
     - pip
     - numpy >=1.17
-    - pyarrow >=6.0.0
+    - pyarrow >=8.0.0
     - python-xxhash
     - dill
     - pandas
--- a/.github/workflows/benchmarks.yaml
+++ b/.github/workflows/benchmarks.yaml
@@ -19,8 +19,8 @@ jobs:
           pip install setuptools wheel
           pip install -e .[benchmarks]
 
-          # pyarrow==6.0.0
-          pip install pyarrow==6.0.0
+          # pyarrow==8.0.0
+          pip install pyarrow==8.0.0
 
           dvc repro --force
 
--- a/.github/workflows/benchmarks.yaml
+++ b/.github/workflows/benchmarks.yaml
@@ -29,7 +29,7 @@ jobs:
 
           python ./benchmarks/format.py report.json report.md
 
-          echo "<details>\n<summary>Show benchmarks</summary>\n\nPyArrow==6.0.0\n" > final_report.md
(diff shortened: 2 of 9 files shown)


mariosasko:
We haven't updated the minimal version requirement for PyArrow in a while, so it's ok to make a bigger leap, e.g., PyArrow 8.0 (Colab installs 9.0). With this change, we should also remove the PyArrow version check in `folder_based_builder.py`, and the ones in `table.py`/`arrow_dataset.py` regarding the `to_reader` API if we decide to bump PyArrow to version 8.0.

albertvillanova (commented):
Thanks for addressing the compatibility issue.

Looking at the failing tests, I see there is no issue with the `datasets` library itself. Instead, the exceptions are raised by the way we perform the tests using this code line:
```python
pd.read_parquet(file_path)
```
Do you think the increase of the minimum version is really necessary or we could just test it differently?

lhoestq:
I think it's a good opportunity to bump the version to 8.0 which offers higher performance anyway, I wouldn't bother trying to support 6.0.1 anymore. Only 1% of users based on 6.0.1 use the latest `datasets` version 2.10.1

Bumping to 8.0 if it sounds good to you

albertvillanova:
Sure, it is OK for that other reasons. I would just not stress that the bump is to support pandas 2.0 though...

albertvillanova:
If requiring min 8.0, do you know the percentage of people using 7.0 and latest datasets version?

lhoestq:
Around 10% of users have 7.0.0, and 25% among them use the latest datasets version

lhoestq:
ready for re-review :)

albertvillanova (commented):
Thank you!! Just a suggestion below.

I think these lines could be removed: https://github.com/huggingface/datasets/blob/a0a35c5fa84a8a7df656c1f5b0a7266126fa9b75/src/datasets/table.py#L821-L823

albertvillanova (approved):
Thanks!

PR #5622, opened 2023-03-08
Update README template to better template

emiltj:


mariosasko:
IMO this template should stay generic.

Also, we now use [the card template](https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/templates/datasetcard_template.md) from `hugginface_hub` as the source of truth on the Hub (you now have the option to import it into the dataset card/README.md), so I think the next step would be deleting this template rather than updating it.

emiltj:
Agreed, the PR was a mistake and meant for my own repo. My bad

mariosasko:
Feel free to close the PR then.

PR #5626, opened 2023-03-10
Support streaming datasets with numpy.load

albertvillanova:
Support streaming datasets with `numpy.load`.

See: https://huggingface.co/datasets/qgallouedec/gia_dataset/discussions/1


lhoestq (approved):
thanks :)

PR #5636, opened 2023-03-14
Fix CI: ignore C901 ("some_func" is to complex) in `ruff`

polinaeterna:
idk if I should have added this ignore to `ruff` too, but I added :)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ target_version = ['py37']
 # Ignored rules:
 #   "E501" -> line length violation
 #   "F821" -> undefined named in type annotation (e.g. Literal["something"])
-#   "C901" -> `function_name` is too complex (introduced in version 0.0.255)
+#   "C901" -> `function_name` is too complex
 ignore = ["E501", "F821", "C901"]
 select = ["C", "E", "F", "I", "W"]
 line-length = 119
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -1,12 +1,13 @@
 [tool.black]
 line-length = 119
 target_version = ['py37']
+ignore = ["C901"]
 
 [tool.ruff]
 # Ignored rules:
 #   "E501" -> line length violation
 #   "F821" -> undefined named in type annotation (e.g. Literal["something"])
```
lhoestq, line 9: maybe add why we ignore it ?

lhoestq (approved):
thanks for the fix !

pyproject.toml
```diff
@@ -1,12 +1,13 @@
 [tool.black]
 line-length = 119
 target_version = ['py37']
+ignore = ["C901"]
```
albertvillanova, line 4: Are you sure we need it here as well?

albertvillanova (approved):
Thanks for the fix.

pyproject.toml
polinaeterna: ah true, thanks! I first thought it was `black` that triggers this error.

Issue #5639, opened 2023-03-15
Parquet file wrongly recognized as zip prevents loading a dataset


clefourrier:
### Describe the bug

When trying to `load_dataset_builder` for `HuggingFaceGECLM/StackExchange_Mar2023`, extraction fails, because parquet file [devops-00000-of-00001-22fe902fd8702892.parquet](https://huggingface.co/datasets/HuggingFaceGECLM/StackExchange_Mar2023/resolve/1f8c9a2ab6f7d0f9ae904b8b922e4384592ae1a5/data/devops-00000-of-00001-22fe902fd8702892.parquet) is wrongly identified by python as being a zip not a parquet.
(Full thread on [Slack](https://huggingface.slack.com/archives/C02V51Q3800/p1678890880803599))

### Steps to reproduce the bug

```python
from datasets import load_dataset_builder

ds = load_dataset_builder("HuggingFaceGECLM/StackExchange_Mar2023")
```

### Expected behavior

Loading the file normally.

### Environment info

- `datasets` version: 2.3.2
- Platform: Linux-5.14.0-1058-oem-x86_64-with-glibc2.29
- Python version: 3.8.10
- PyArrow version: 8.0.0
- Pandas version: 1.4.3

PR #5640, opened 2023-03-15
Less zip false positives

lhoestq:
`zipfile.is_zipfile` return false positives for some Parquet files. It causes errors when loading certain parquet datasets, where some files are considered ZIP files by `zipfile.is_zipfile`

This is a known issue: https://github.com/python/cpython/issues/72680

At first I wanted to rely only on magic numbers, but then I found that someone contributed a [fix to is_zipfile](https://github.com/python/cpython/pull/5053) - do you think we should use it @albertvillanova or not ?

IMO it's ok to rely on magic numbers only for now, since in streaming mode we've had no issue checking only the magic number so far.

Close https://github.com/huggingface/datasets/issues/5639

### DIFF
--- a/src/datasets/utils/extract.py
+++ b/src/datasets/utils/extract.py
@@ -3,12 +3,13 @@ import gzip
 import lzma
 import os
 import shutil
+import struct
 import tarfile
 import warnings
 import zipfile
 from abc import ABC, abstractmethod
 from pathlib import Path
-from typing import List, Optional, Union
+from typing import Dict, List, Optional, Type, Union
 
 from .. import config
 from .filelock import FileLock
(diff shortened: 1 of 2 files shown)


lhoestq:
CI is failing due to unrelated issues, hopefilly https://github.com/huggingface/datasets/pull/5642 fixes it

src/datasets/utils/extract.py
```diff
@@ -137,10 +139,48 @@ def extract(input_path: Union[Path, str], output_path: Union[Path, str]) -> None
...
+                    elif endrec[_ECD_DISK_NUMBER] == endrec[_ECD_DISK_START]:
+                        fp.seek(endrec[_ECD_OFFSET])  # Central directory is on the same disk
+                        if fp.tell() == endrec[_ECD_OFFSET] and endrec[_ECD_SIZE] >= sizeCentralDir:
+                            data = fp.read(sizeCentralDir)  # CD is where we expect it to be
+                            if len(data) == sizeCentralDir:
+                                centdir = struct.unpack(structCentralDir, data)  # CD is the right size
+                                if centdir[_CD_SIGNATURE] == stringCentralDir:
+                                    return True  # First central directory entry  has correct magic number
+        except Exception:  # catch all errors in case future python versions change the zipfile internals
+            return False
```
albertvillanova, line 183: Please note that this function as it is could return `None`. To fix this:
```suggestion
            return False
        return False
```

tests/test_extract.py
```diff
@@ -183,3 +184,20 @@ def test_tar_extract_insecure_files(
...
+    data = (
+        b"\x89PNG\r\n\x1a\n\x00\x00\x00\rIHDR\x00\x00\x00\x01\x00\x00"
+        b"\x00\x02\x08\x06\x00\x00\x00\x99\x81\xb6'\x00\x00\x00\x15I"
+        b"DATx\x01\x01\n\x00\xf5\xff\x00PK\x05\x06\x00PK\x06\x06\x07"
+        b"\xac\x01N\xc6|a\r\x00\x00\x00\x00IEND\xaeB`\x82"
+    )
+    with not_a_zip_file.open("wb") as f:
+        f.write(data)
+    assert zipfile.is_zipfile(str(not_a_zip_file))  # is a false positive for `zipfile`
+    assert not ZipExtractor.is_extractable(not_a_zip_file)  # but we're right
```
albertvillanova, line 203: The test passes because `not None` is `True`.

albertvillanova (approved):
Thanks for the improvement.

I agree we should tweak the `zipfile.is_zipfile` default implementation if that is flaky.

src/datasets/utils/extract.py
```diff
@@ -137,10 +139,48 @@ def extract(input_path: Union[Path, str], output_path: Union[Path, str]) -> None
...
+                    if endrec[_ECD_ENTRIES_TOTAL] == 0 and endrec[_ECD_SIZE] == 0 and endrec[_ECD_OFFSET] == 0:
+                        return True  # Empty zipfiles are still zipfiles
+                    elif endrec[_ECD_DISK_NUMBER] == endrec[_ECD_DISK_START]:
+                        fp.seek(endrec[_ECD_OFFSET])  # Central directory is on the same disk
+                        if fp.tell() == endrec[_ECD_OFFSET] and endrec[_ECD_SIZE] >= sizeCentralDir:
+                            data = fp.read(sizeCentralDir)  # CD is where we expect it to be
+                            if len(data) == sizeCentralDir:
+                                centdir = struct.unpack(structCentralDir, data)  # CD is the right size
+                                if centdir[_CD_SIGNATURE] == stringCentralDir:
+                                    return True  # First central directory entry  has correct magic number
```
albertvillanova, line 181: The fix could also be added here:
```suggestion
                                    return True  # First central directory entry  has correct magic number
            return False
```

albertvillanova (approved):

PR #5642, opened 2023-03-15
Bump hfh to 0.11.0

lhoestq:
to fix errors like

```
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://hub-ci.huggingface.co/api/datasets/__DUMMY_TRANSFORMERS_USER__/...
```

(e.g. from this [failing CI](https://github.com/huggingface/datasets/actions/runs/4428956210/jobs/7769160997))

0.11.0 is the current minimum version in `transformers`

around 5% of users are currently using versions `<0.11.0`


albertvillanova (commented):
It is quite a big jump for a minimum requirement from 0.2.0 to 0.11.0.

Do you think this is really necessary and convenient? I would naively say that 5% of the users is not a negligible number...

Alternatively, we could try to find a different fix for the client error issue and add a deprecation warning to tell users to update their `huggingface_hub` version because it will not be supported in future releases.

Otherwise, if you think it is OK as it is, then go on!

lhoestq:
merging, but lmk if you have other concerns

PR #5643, opened 2023-03-15
Support PyArrow arrays as column values in `from_dict`

mariosasko:
For consistency with `pa.Table.from_pydict`, which supports both Python lists and PyArrow arrays as column values.

"Fixes" https://discuss.huggingface.co/t/pyarrow-lib-floatarray-did-not-recognize-python-value-type-when-inferring-an-arrow-data-type/33417


lhoestq (approved):
Thanks !

PR #5644, opened 2023-03-15
Allow direct cast from binary to Audio/Image

mariosasko:
To address https://github.com/huggingface/datasets/discussions/5593.


lhoestq (approved):
Nice ! LGTM :)

PR #5652, opened 2023-03-20
Copy features

lhoestq:
Some users (even internally at HF) are doing
```python
dset_features = dset.features
dset_features.pop(col_to_remove)
dset = dset.map(..., features=dset_features)
```

Right now this causes issues because it modifies the features dict in place before the map.

In this PR I modified `dset.features` to return a copy of the features, so that users can modify it if they want.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -183,7 +183,7 @@ class DatasetInfoMixin:
 
     @property
     def features(self) -> Features:
-        return self._info.features
+        return self._info.features.copy()
 
     @property
     def homepage(self) -> Optional[str]:
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -568,7 +568,10 @@ def transmit_tasks(func):
                 dataset.info.task_templates = [
                     template
                     for template in self.info.task_templates
-                    if all(dataset.features.get(k) == self.features.get(k) for k in template.column_mapping.keys())
+                    if all(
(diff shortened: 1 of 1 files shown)


lhoestq:
Arf I need to fix some tests first - sorry

albertvillanova (commented):
I'm not sure of understanding what the issue is: the code snippet you provide behaves as expected. In Python, mutable objects are passed by reference.

If the user does not want to pass a mutable object by reference, but a copy of it, they should call its `.copy()` method:
```python
dset_features = dset.features.copy()
```

An example of the expected behavior with a dict (mutable):
```python
In [11]: class Dataset:
    ...:     def __init__(self, features):
    ...:         self.features = features

In [12]: dset = Dataset({"a": 1, "b": 2})

In [13]: dset.features
Out[13]: {'a': 1, 'b': 2}

In [14]: dset_features = dset.features

In [15]: _ = dset_features.pop("a")

In [16]: dset.features
Out[16]: {'b': 2}
```

Another example with a class instance (mutable as well):
```python
In [20]: class Features:
    ...:     def __init__(self, value):
    ...:         self.value = value

In [21]: dset = Dataset(Features(1))

In [22]: dset.features.value
Out[22]: 1

In [23]: dset_features = dset.features

In [24]: dset_features.value = 2

In [25]: dset.features.value
Out[25]: 2
```

albertvillanova (approved):

mariosasko (approved):
Indeed, requiring to manually `copy` the `features` can be annoying, so thanks for "fixing" this!

Issue #5666, opened 2023-03-23
Support tensorflow 2.12.0 in CI


albertvillanova:
Once we find out the root cause of:
- #5663

we should revert the temporary pin on tensorflow introduced by:
- #5664

PR #5667, opened 2023-03-23
Jax requires jaxlib

lhoestq:
close https://github.com/huggingface/datasets/issues/5666


albertvillanova:
I self-assigned #5666 and I was working on it... without success: https://github.com/huggingface/datasets/tree/fix-5666

I think your approach is the right one because installation of jax is not trivial...

Next time it would be better that you self-assign an issue before working on it, so that we avoid duplicate work... :sweat_smile:

albertvillanova (approved):
Thank you for the quick fix!

lhoestq:
Oh sorry I forgot to self assign this time

PR #5668, opened 2023-03-23
Support for downloading only provided split

polinaeterna:
We can pass split to `_split_generators()`.
But I'm not sure if it's possible to solve cache issues, mostly with `dataset_info.json`


albertvillanova (commented):
Thank you.

As discussed in our weekly meeting:
- I already started to work on this long ago: you can see the corresponding PR for context:
  - https://github.com/huggingface/datasets/pull/2249
- The caching issue should be addressed first:
  - Currently, the cache only checks for the existence of the corresponding dataset directory in the cache
  - It should be enhanced and check the content of that directory, because it might be partially filled with only some of the splits

albertvillanova:
My previous comment didn't create the retro-link in the PR. I write it here again.

You can check the context and the discussions we had about this feature enhancement in this PR:
- #2249

PR #5673, opened 2023-03-26
Pass down storage options

dwyatte:
Remove implementation-specific kwargs from `file_utils.fsspec_get` and `file_utils.fsspec_head`, instead allowing them to be passed down via `storage_options`. This fixes an issue where s3fs did not recognize a timeout arg as well as fixes an issue mentioned in https://github.com/huggingface/datasets/issues/5281 by allowing users to pass down `storage_options` all the way from `datasets.load_dataset` to support implementation-specific credentials

Supports something like the following to provide credentials explicitly instead of relying on boto's methods of locating them

```
load_dataset(..., data_files=["s3://..."], storage_options={"profile": "..."})
```

### DIFF
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -1471,7 +1471,9 @@ def load_dataset_builder(
             Optional string or boolean to use as Bearer token for remote files on the Datasets Hub.
             If `True`, or not specified, will get token from `"~/.huggingface"`.
         storage_options (`dict`, *optional*, defaults to `None`):
-            Key/value pairs to be passed on to the dataset file-system backend, if any.
+            **Experimental**. Key/value pairs to be passed on to the dataset file-system backend, if any.
+
+            <Added version="2.11.0"/>
         **config_kwargs (additional keyword arguments):
             Keyword arguments to be passed to the [`BuilderConfig`]
             and used in the [`DatasetBuilder`].
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -1550,6 +1550,7 @@ def load_dataset(
     task: Optional[Union[str, TaskTemplate]] = None,
     streaming: bool = False,
     num_proc: Optional[int] = None,
+    storage_options: Optional[Dict] = None,
     **config_kwargs,
 ) -> Union[DatasetDict, Dataset, IterableDatasetDict, IterableDataset]:
     """Load a dataset from the Hugging Face Hub, or a local dataset.
@@ -1671,6 +1672,8 @@ def load_dataset(
             Multiprocessing is disabled by default.
 
             <Added version="2.7.0"/>
+        storage_options (`dict`, *optional*, defaults to `None`):
+            Key/value pairs to be passed on to the file-system backend, if any.
         **config_kwargs (additional keyword arguments):
             Keyword arguments to be passed to the `BuilderConfig`
             and used in the [`DatasetBuilder`].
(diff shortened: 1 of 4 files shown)


src/datasets/utils/file_utils.py
```diff
@@ -499,7 +501,8 @@ def get_from_cache(
         if scheme == "ftp":
             connected = ftp_head(url)
         elif scheme not in ("http", "https"):
-            response = fsspec_head(url)
+            response = fsspec_head(url, storage_options=storage_options)
+            # use the hash of the response as a pseudo ETag to detect changes
```
lhoestq, line 505:
```suggestion
```

lhoestq (commented):
Awesome ! Just one comment:

`download_and_prepare` is not called when streaming a dataset, so we may need to have `storage_options` in the `DatasetBuilder.__init__` ? This way it could also be passed later to `as_streaming_dataset` and the `StreamingDownloadManager`

Currently the `storage_options` parameter in `download_and_prepare` are for the target filesystem where the dataset must be downloaded and prepared as arrow files

dwyatte:
> download_and_prepare is not called when streaming a dataset, so we may need to have storage_options in the DatasetBuilder.__init__ ? This way it could also be passed later to as_streaming_dataset and the StreamingDownloadManager

> Currently the storage_options parameter in download_and_prepare are for the target filesystem where the dataset must be downloaded and prepared as arrow files

Ah, I noted this when looking for ways to plumb down storage_options although I think I was looking at adding to `BuilderConfig`. The `DatasetBuilder` constructor looks more appropriate for this, will get that added in a future commit

src/datasets/utils/file_utils.py
dwyatte: Thanks, this fell through the cracks of a rebase!

src/datasets/load.py
```diff
@@ -1671,6 +1676,8 @@ def load_dataset(
             Multiprocessing is disabled by default.
 
             <Added version="2.7.0"/>
+        storage_options (`dict`, *optional*, defaults to `None`):
+            Key/value pairs to be passed on to the dataset file-system backend, if any.
```
lhoestq, line 1680: I'd keep this feature experimental until we have a proper data files resolution based on fsspec (e.g. request_etag and is_remote_url don't really support fsspec) as well as some tests:
```suggestion
            **Experimental**. Key/value pairs to be passed on to the dataset file-system backend, if any.
```

src/datasets/load.py
```diff
@@ -1469,6 +1470,8 @@ def load_dataset_builder(
         use_auth_token (`str` or `bool`, *optional*):
             Optional string or boolean to use as Bearer token for remote files on the Datasets Hub.
             If `True`, or not specified, will get token from `"~/.huggingface"`.
+        storage_options (`dict`, *optional*, defaults to `None`):
+            Key/value pairs to be passed on to the dataset file-system backend, if any.
```
lhoestq, line 1474:
```suggestion
            **Experimental**. Key/value pairs to be passed on to the dataset file-system backend, if any.
```

lhoestq (approved):
Amazing ! It works like a charm :)

Just before we merge I wanted to mention this in the docstring, let me know what you think:

dwyatte:
Noting as experimental SGTM. The only tests I can think of to add at the moment would be mocks that assert the storage options get passed all the way down using `mock.assert_called_with` but if Hugging Face has some S3/GCS buckets for testing, maybe those would be better in a future PR. Let me know what you think

PR #5689, opened 2023-03-31
Support streaming Beam datasets from HF GCS preprocessed data

albertvillanova:
This PR implements streaming Apache Beam datasets that are already preprocessed by us and stored in the HF Google Cloud Storage:
- natural_questions
- wiki40b
- wikipedia

This is done by streaming from the prepared Arrow files in HF Google Cloud Storage.

This will fix their corresponding dataset viewers. Related to:
- https://github.com/huggingface/datasets-server/pull/988#discussion_r1150767138

Related to:
- https://huggingface.co/datasets/natural_questions/discussions/4
- https://huggingface.co/datasets/wiki40b/discussions/2
- https://huggingface.co/datasets/wikipedia/discussions/9

CC: @severo

### DIFF
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -1245,9 +1245,6 @@ class DatasetBuilder:
         split: Optional[str] = None,
         base_path: Optional[str] = None,
     ) -> Union[Dict[str, IterableDataset], IterableDataset]:
-        if not isinstance(self, (GeneratorBasedBuilder, ArrowBasedBuilder)):
-            raise ValueError(f"Builder {self.name} is not streamable.")
-
         is_local = not is_remote_filesystem(self._fs)
         if not is_local:
             raise NotImplementedError(
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -2100,7 +2100,7 @@ class BeamBasedBuilder(DatasetBuilder):
     def _get_examples_iterable_for_split(self, split: str) -> ExamplesIterable:
         return ExamplesIterable(self._generate_examples_from_hf_gcs, {"split": split})
 
-    def _generate_examples_from_hf_gcs(self, split):  # , split_name):
+    def _generate_examples_from_hf_gcs(self, split):
         import pyarrow as pa
 
         from .download.streaming_download_manager import xopen
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -2078,3 +2078,56 @@ class BeamBasedBuilder(DatasetBuilder):
 
         # Add the PCollection to the pipeline
         _ = pipeline | split_name >> _build_pcollection()  # pylint: disable=no-value-for-parameter max_bytes_per_shard
+
+    def as_streaming_dataset(
+        self,
+        split: Optional[str] = None,
+    ) -> Union[Dict[str, IterableDataset], IterableDataset]:
+        self._request_info_from_hf_gcs()
+        datasets = {
+            split: IterableDataset(self._get_examples_iterable_for_split(split), info=self.info, split=split)
+            for split in self.info.splits
+        }
+        if split:
+            try:
+                datasets = datasets[split]
+            except KeyError:
+                raise ValueError(f"Bad split: {split}. Available splits: {list(datasets)}")
+        if isinstance(datasets, dict):
+            datasets = IterableDatasetDict(datasets)
+        return datasets
+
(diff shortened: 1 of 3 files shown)


albertvillanova:
```python
In [1]: from datasets import load_dataset; 

In [2]: ds = load_dataset("wikipedia", "20220301.en", split="train", streaming=True); item = next(iter(ds)); item
Out[2]: 
{'id': '12',
 'url': 'https://en.wikipedia.org/wiki/Anarchism',
 'title': 'Anarchism',
 'text': 'Anarchism is a political philosophy and movement that is sceptical of authority and rejects all involuntary, coercive forms of hierarchy. Anarchism calls for the abolition of the state, which it holds to be unnecessary, undesirable, and harmful. As a historically left-wing movement, placed on the farthest left of the political spectrum, it is usually described alongside communalism and libertarian Marxism as the libertarian wing (libertarian socialism) of the socialist movement,...}
```

severo:
I love your example 🏴‍🅰️

src/datasets/builder.py
```diff
@@ -2081,3 +2078,56 @@ def _build_pcollection(pipeline):
 
         # Add the PCollection to the pipeline
         _ = pipeline | split_name >> _build_pcollection()  # pylint: disable=no-value-for-parameter max_bytes_per_shard
+
+    def as_streaming_dataset(
+        self,
+        split: Optional[str] = None,
+    ) -> Union[Dict[str, IterableDataset], IterableDataset]:
+        self._request_info_from_hf_gcs()
```
lhoestq, line 2086: Maybe raise NotImplementedError if the dataset is not on the HF GCS ?

src/datasets/builder.py
```diff
@@ -2081,3 +2078,56 @@ def _build_pcollection(pipeline):
...
+            datasets = IterableDatasetDict(datasets)
+        return datasets
+
+    def _get_examples_iterable_for_split(self, split: str) -> ExamplesIterable:
+        return ExamplesIterable(self._generate_examples_from_hf_gcs, {"split": split})
+
+    def _generate_examples_from_hf_gcs(self, split):
+        import pyarrow as pa
+
+        from .download.streaming_download_manager import xopen
```
lhoestq, line 2106: I think this can be moved those imports at the top of the file ?

lhoestq (commented):
Amazing thanks !!

You could also have a simple integration test in test_hf_gcp.py for wikipedia to make sure it keeps working in the long run

src/datasets/builder.py
albertvillanova: This method already may raise `DatasetNotOnHfGcsError`. You don't like this error and you prefer `NotImplementedError` instead?

lhoestq: Ah yes indeed, it's fine then I think

src/datasets/builder.py
```diff
@@ -2081,3 +2079,52 @@ def _build_pcollection(pipeline):
...
+                raise ValueError(f"Bad split: {split}. Available splits: {list(datasets)}")
+        if isinstance(datasets, dict):
+            datasets = IterableDatasetDict(datasets)
+        return datasets
+
+    def _get_examples_iterable_for_split(self, split: str) -> ExamplesIterable:
+        return ExamplesIterable(self._generate_examples_from_hf_gcs, {"split": split})
+
+    def _generate_examples_from_hf_gcs(self, split):
+        remote_prepared_filename = f"{self._remote_cache_dir_from_hf_gcs}/{self.name}-{split}.arrow"
```
lhoestq, line 2105: (nit) Beam builders may create sharded arrow files now.

It's the case for none of the datasets on GCP but if we regenerate a new one it may en up sharded.
You can check the dataset `info.shard_lengths` to know if it's sharded or not and how many shards there are.


```python
    if self.info.shard_lengths:
        num_shards = len(self.info.shard_lengths)
        urls = [
            f"f"{self._remote_cache_dir_from_hf_gcs}/{self.name}-{split}-{shard_id:05d}-of-{num_shards:05d}.arrow"
            for shard_id in range(num_shards)
        ]
    else:
        urls = [f"{self._remote_cache_dir_from_hf_gcs}/{self.name}-{split}.arrow"]
```

albertvillanova: As the attribute `shard_lengths` belongs to `SplitInfo` (and not `DatasetInfo`), I have refactored the method so that split is now of type `SplitInfo` instead of `str`.

lhoestq (approved):
LGTM ! :)

Issue #5681, opened 2023-03-29
Add information about patterns search order to the doc about structuring repo


polinaeterna:
Following [this](https://github.com/huggingface/datasets/issues/5650) issue I think we should add a note about the order of patterns that is used to find splits, see [my comment](https://github.com/huggingface/datasets/issues/5650#issuecomment-1488412527). Also we should reference this page in pages about packaged loaders. 

I have a déjà vu that it had already been discussed as some point but I don't remember....

stevhliu:
Good idea, I think I've seen this a couple of times before too on the forums. I can work on this :)

stevhliu:
Closed in #5693

PR #5693, opened 2023-03-31
[docs] Split pattern search order

stevhliu:
This PR addresses #5681 about the order of split patterns 🤗 Datasets searches for when generating dataset splits.

### DIFF
--- a/docs/source/repository_structure.mdx
+++ b/docs/source/repository_structure.mdx
@@ -26,7 +26,7 @@ my_dataset_repository/
 
 ### Custom filename split
 
-The most flexible pattern to accomodate other split types in addition to the traditional train, validation, and test sets, is to name your data files like `data/<split_name>-xxxxx-of-xxxxx.csv`.
+If your dataset splits have custom names that aren't `train`, `test`, or `validation`, then you should name your data files like `data/<split_name>-xxxxx-of-xxxxx.csv`.
 
 Here is an example with three splits, `train`, `test`, and `random`:
 
@@ -82,11 +82,11 @@ my_dataset_repository/
 
 ### Single split
 
-When 🤗 Datasets can't find any of the above patterns, then it'll treat all the files as a single train split. If your dataset splits aren't loading as expected, this may be why!
+When 🤗 Datasets can't find any of the above patterns, then it'll treat all the files as a single train split. If your dataset splits aren't loading as expected, it may be due to an incorrect pattern.
 
 ## Split name keywords
 
-There are several acceptable ways to name splits. Validation splits are sometimes called "dev", and test splits may be referred to as "eval".
+There are several ways to name splits. Validation splits are sometimes called "dev", and test splits may be referred to as "eval".
 These other split names are also supported, and the following keywords are equivalent:
 
 - train, training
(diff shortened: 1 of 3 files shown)


docs/source/repository_structure.mdx
```diff
@@ -41,9 +80,29 @@ my_dataset_repository/
...
-Note that if a file contains *test* but is embedded in another word (e.g. `testfile.csv`), it's not counted as a test file.
-It must be delimited by non-word characters, e.g. `test_file.csv`.
-Supported delimiters are underscores, dashes, spaces, dots and numbers.
+### Single split
+
+When 🤗 Datasets can't find any of the above patterns, then it'll treat all the files as a single train split. If your dataset splits aren't loading as expected, this may be why!
+
+## Split name keywords
+
+There are several acceptable ways to name splits. Validation splits are sometimes called "dev", and test splits may be referred to as "eval".
```
polinaeterna, line 89:
```suggestion
There are several ways to name splits. Validation splits are sometimes called "dev", and test splits may be referred to as "eval".
```

docs/source/repository_structure.mdx
```diff
@@ -20,11 +20,50 @@ my_dataset_repository/
...
-## Splits and file names
+## Split pattern hierarchy
 
-🤗 Datasets automatically infer a dataset's train, validation, and test splits from the file names.
+🤗 Datasets searches for certain patterns in the dataset repository to automatically infer the dataset splits. There is an order to the patterns, beginning with the custom filename split format to treating all files as a single split if no pattern is found.
 
-All the files that contain a split name in their names (delimited by non-word characters, see below) are considered part of that split:
+### Custom filename split
+
+The most flexible pattern to accomodate other split types in addition to the traditional train, validation, and test sets, is to name your data files like `data/<split_name>-xxxxx-of-xxxxx.csv`.
```
polinaeterna, line 29:
```suggestion
If your dataset has split that have non traditional names, you should name your data files like `data/<split_name>-xxxxx-of-xxxxx.csv`.
```
Not the best wording in my suggestion, feel free to change, I just wanted to point out that I wouldn't encourage for this pattern, I mean no need to name it the most flexible. it's quite a specific one. Putting all the data into a single directory named as a split (the next pattern) is easier and for most cases enough.

docs/source/repository_structure.mdx
```diff
@@ -41,9 +80,29 @@ my_dataset_repository/
     └── validation.csv
 ```
 
-Note that if a file contains *test* but is embedded in another word (e.g. `testfile.csv`), it's not counted as a test file.
-It must be delimited by non-word characters, e.g. `test_file.csv`.
-Supported delimiters are underscores, dashes, spaces, dots and numbers.
+### Single split
+
+When 🤗 Datasets can't find any of the above patterns, then it'll treat all the files as a single train split. If your dataset splits aren't loading as expected, this may be why!
```
polinaeterna, line 85:
```suggestion
When 🤗 Datasets can't find any of the above patterns, then it'll treat all the files as a single train split. If your dataset splits aren't loading as expected, it may be due to incorrect pattern.
```

polinaeterna (approved):
thank you! left a few comments, feel free to reword my suggestions :)

Issue #5699, opened 2023-04-03
Issue when splitting in memory a cached dataset


FrancoisNoyez:
### Describe the bug

In the 'train_test_split' method of the Dataset class (defined datasets/arrow_dataset.py), if 'self.cache_files' is not empty, then, regarding the input parameters 'train_indices_cache_file_name' and 'test_indices_cache_file_name', if they are None, we modify them to make them not None, to see if we can just provide back / work from cached data. But if we can't provide cached data, we move on with the call to the method, except those two values are not None anymore, which will conflict with the use of the 'keep_in_memory' parameter down the line.
Indeed, at some point we end up calling the 'select' method, and if 'keep_in_memory' is True, since the value of this method's parameter 'indices_cache_file_name' is now not None anymore, an exception is raised, whose message is "Please use either 'keep_in_memory' or 'indices_cache_file_name' but not both.".
Because of that, it's impossible to perform a train / test split of a cached dataset while requesting that the result not be cached. Which is inconvenient when one is just performing experiments, with no intention of caching the result.

Aside from this being inconvenient, the code which lead up to that situation seems simply wrong to me: the input variable should not be modified so as to change the user's intention just to perform a test, if that test can fail and respecting the user's intention is necessary to proceed in that case.
To fix this, I suggest to use other variables / other variable names, in order to host the value(s) needed to perform the test, so as not to change the originally input values needed by the rest of the method's code.

Also, I don't see why an exception should be raised when the 'select' method is called with both 'keep_in_memory'=True and 'indices_cache_file_name'!=None: should the use of 'keep_in_memory' not prevail anyway, specifying that the user does not want to perform caching, and so making irrelevant the value of 'indices_cache_file_name'? This is indeed what happens when we look further in the code, in the '\_select_with_indices_mapping' method: when 'keep_in_memory' is True, then the value of indices_cache_file_name does not matter, the data will be written to a stream buffer anyway.
Hence I suggest to remove the raising of exception in those circumstances. Notably, to remove the raising of it in the 'select', '\_select_with_indices_mapping' and 'shuffle' methods.

### Steps to reproduce the bug

```python
import datasets

def generate_examples():
    for i in range(10):
        yield {"id": i}

dataset_ = datasets.Dataset.from_generator(
    generate_examples,
    keep_in_memory=False,
)
dataset_.train_test_split(
    test_size=3,
    shuffle=False,
    keep_in_memory=True,
    train_indices_cache_file_name=None,
[...]

PR #5700, opened 2023-04-03
fix: fix wrong modification of the 'cache_file_name' -related paramet…

FrancoisNoyez:
…ers values in 'train_test_split' + fix bad interaction between 'keep_in_memory' and 'cache_file_name' -related parameters (#5699)

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -2882,9 +2882,6 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         >>> ds = ds.map(add_prefix, num_proc=4)
         ```
         """
-        if keep_in_memory and cache_file_name is not None:
-            raise ValueError("Please use either `keep_in_memory` or `cache_file_name` but not both.")
-
         if num_proc is not None and num_proc <= 0:
             raise ValueError("num_proc must be an integer > 0.")
 
@@ -3675,9 +3672,6 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         })
         ```
         """
-        if keep_in_memory and indices_cache_file_name is not None:
-            raise ValueError("Please use either `keep_in_memory` or `indices_cache_file_name` but not both.")
-
         if len(self.list_indexes()) > 0:
             raise DatasetTransformationNotAllowedError(
                 "Using `.select` on a dataset with attached indexes is not allowed. You can first run `.drop_index() to remove your index and then re-add it."
@@ -3813,9 +3807,6 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         })
         ```
         """
-        if keep_in_memory and indices_cache_file_name is not None:
-            raise ValueError("Please use either `keep_in_memory` or `indices_cache_file_name` but not both.")
-
         if len(self.list_indexes()) > 0:
             raise DatasetTransformationNotAllowedError(
                 "Using `.select` on a dataset with attached indexes is not allowed. You can first run `.drop_index() to remove your index and then re-add it."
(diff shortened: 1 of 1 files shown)


lhoestq:
Have you tried to set the cache file names if `keep_in_memory`is True ?

```diff
- if self.cache_files:
+ if self.cache_files and not keep_in_memory
```

This way it doesn't change the indice cache arguments and leave them as `None`

FrancoisNoyez:
Regarding what you suggest:
The thing is, if cached files already exist and do correspond to the split that we are currently trying to perform, then it would be a shame not to use them, would it not? So I don't think that we should necessarily bypass this step in the method (corresponding to the reading of already existing data), if 'keep_in_memory' = True. For me, 'keep_in_memory' = True is supposed to mean "don't cache the output of this method", but it should say nothing regarding what to do with potentially already existing cached data, should it?
Besides, even if we do what you suggest, and do only that (so, not the modifs that I suggested), then, assuming that 'keep_in_memory' = False and that there exist cached files, if the following check on the existence of cached files with specific name fails, we will still have ended up modifying an input value which will be then used in the remaining of the method, potentially altering the behavior that the user intended the method's call to have. Basically, the issue with what you suggest is that we can't guaranty that we won't continue with the remaining of the method even if this condition is met. Because of that, in my opinion, the best way to not have to worry about potential, unwanted side effects in the rest of the code is to not modify those variables in place, and so, here, to use other variables.
So, I'm sorry, but for those two reasons, I don't think that what you are suggesting addresses the problems which are described in the opened issue.

lhoestq:
Makes sense ! Therefore removing the ValueError messages sounds good to me, thanks for detailing.

Then I think it's fine to keep using the same variables for the cache file names is enough instead of defining new ones - it doesn't alter the behavior of the function. Otherwise it would feel a bit confusing to have similar variables with slightly modified names just for that

FrancoisNoyez:
Ok for the removing the ValueError exceptions, thanks.

That said, it seems to me like we should still find a way not to modify the values input by the user, insofar as they can be used elsewhere down the line in the program. Sure, here, by removing the raising of those ValueError exceptions, we have fixed one use cases were allowing this modification actually caused an issue, but maybe there are other use cases where this would also caused an issue? Also, maybe in the future we will add other functionalities which will depend on the values of those input parameters, with then new risks of such an issue occurring?
That's why, in order not to have to worry about that, and in order to make the code a bit more future -proof, I suggest that make sure those input values are not modified.

One way that I did this is to create different but similar looking variable names. If you find this confusing, we can always add a comment.
Another way would be to not store the result of the conditional definition of the values (the '\_cache_file_name = (... if condition else ...)' in my proposition of code), and to use it every time we need. But since we use those new variables at least twice, that creates code redundancy, which is not great either.
Finally, a third way that I can imagine would be to put all this logic into its own method, which would then encapsulate it, and protect the remaining of the 'train_test_split' code from all unintended side effect that this logic can currently cause. This one is probably best. Also, maybe it could be used to remove some code redundancy elsewhere in the definition of the Dataset class? I have not checked if such a code redundancy exists.

FrancoisNoyez:
> We're already replacing the user's input by default values automatically in other methods, it's fine to do it here as well and actually fits the library's style.

I think the fact that it's a style of the library is not really an argument in itself; however, after thinking through it several times, I think I know see why your solution is acceptable: as soon as the user specifies that 'keep_in_memory=True', they should not care anymore about the value of the '\_indices_cache_file_name' variables, since from their point of view those are now irrelevant. So it's "fine" if we allow ourselves to modify the value of those variables, if it helps the internal code being more concise.
Still, I find that it's a bit unintuitive, and a risk as far as future evolution of the method / of the code is concerned; someone tasked with doing that would need to have the knowledge of a lot of, if not all, the other methods of the class, in order to understand the potentially far-reaching impact of some modifications made to this portion of the code. But I guess that's a choice which is the library's owners to make.

> Note that the case where it would reload the cache even if `keep_in_memory=True` is not implemented though

I'm not sure what you mean here:
Within the current code trying to load up the potentially already existing split data, there is no trace of the 'keep_in_memory' variable. So why do you say that 'the case where it would reload the cache even if keep_in_memory=True is not implemented' (I assume that you mean 'currently implemented')? Surely, currently, this bit of code works regardless of the value of the 'keep_in_memory' variable', does it not?

PR #5718, opened 2023-04-07
Reorder default data splits to have validation before test

albertvillanova:
This PR reorders data splits, so that by default validation appears before test.

The default order becomes: [train, validation, test] instead of [train, test, validation].


albertvillanova:
After this CI error: https://github.com/huggingface/datasets/actions/runs/4639528358/jobs/8210492953?pr=5718
```
FAILED tests/test_data_files.py::test_get_data_files_patterns[data_file_per_split4] - AssertionError: assert ['random', 'train'] == ['train', 'random']
  At index 0 diff: 'random' != 'train'
  Full diff:
  - ['train', 'random']
  + ['random', 'train']
```
I have checked locally and found out that the data split order is nondeterministic. I am addressing this in a separate issue.

lhoestq (approved):
it's better this way indeed !

PR #5729, opened 2023-04-11
Fix nondeterministic sharded data split order

### LINKED ISSUE #5728: The order of data split names is nondeterministic
After this CI error: https://github.com/huggingface/datasets/actions/runs/4639528358/jobs/8210492953?pr=5718
```
FAILED tests/test_data_files.py::test_get_data_files_patterns[data_file_per_split4] - AssertionError: assert ['random', 'train'] == ['train', 'random']
  At index 0 diff: 'random' != 'train'
  Full diff:
  - ['train', 'random']
  + ['random', 'train']
```
I have checked locally and found out that the data split order is nondeterministic.

This is caused by the use of `set` for sharded splits.

albertvillanova:
This PR makes the order of the split names deterministic. Before it was nondeterministic because we were iterating over `set` elements.

Fix #5728.


albertvillanova:
The error in the CI was unrelated to this PR. I have merged main branch once that has been fixed.

PR #5733, opened 2023-04-11
Unpin fsspec

albertvillanova:
In `fsspec--2023.4.0` default value for clobber when registering an implementation was changed from True to False. See:
- https://github.com/fsspec/filesystem_spec/pull/1237

This PR recovers previous behavior by passing clobber True when registering mock implementations.


lhoestq (approved):
Thanks !

PR #5735, opened 2023-04-11
Implement sharding on merged iterable datasets

Hubert-Bonisseur:
This PR allows sharding of merged iterable datasets.

Merged iterable datasets with for instance the `interleave_datasets` command are comprised of multiple sub-iterable, one for each dataset that has been merged.

With this PR, sharding a merged iterable will result in multiple merged datasets each comprised of sharded sub-iterable, ensuring that there is no duplication of data.

As a result it is now possible to set any amount of workers in the dataloader as long as it is lower or equal to the lowest amount of shards amongst the datasets. Before it had to be set to 0.

I previously talked about this issue on the forum [here](https://discuss.huggingface.co/t/interleaving-iterable-dataset-with-num-workers-0/35801)

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -251,7 +251,7 @@ class CyclingMultiSourcesExamplesIterable(_BaseExamplesIterable):
 
     @property
     def n_shards(self) -> int:
-        return sum(ex_iterable.n_shards for ex_iterable in self.ex_iterables)
+        return min(ex_iterable.n_shards for ex_iterable in self.ex_iterables)
 
     def shard_data_sources(self, worker_id: int, num_workers: int) -> "CyclingMultiSourcesExamplesIterable":
         """Either keep only the requested shard, or propagate the request to the underlying iterable."""
@@ -293,7 +293,7 @@ class VerticallyConcatenatedMultiSourcesExamplesIterable(_BaseExamplesIterable):
 
     @property
     def n_shards(self) -> int:
-        return sum(ex_iterable.n_shards for ex_iterable in self.ex_iterables)
+        return min(ex_iterable.n_shards for ex_iterable in self.ex_iterables)
 
     def shard_data_sources(
         self, worker_id: int, num_workers: int
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -22,7 +22,6 @@ from .table import table_cast
 from .utils.logging import get_logger
 from .utils.sharding import _merge_gen_kwargs, _number_of_shards_in_gen_kwargs, _shuffle_gen_kwargs, _split_gen_kwargs
 
-
 logger = get_logger(__name__)
 
 
@@ -95,9 +94,12 @@ class _BaseExamplesIterable:
         """
         raise NotImplementedError(f"{type(self)} doesn't implement shuffle_data_sources yet")
 
-    def shard_data_sources(self, shard_indices: List[int]) -> "_BaseExamplesIterable":
+    def shard_data_sources(self, worker_id: int, num_workers: int) -> "_BaseExamplesIterable":
         """Either keep only the requested shard, or propagate the request to the underlying iterable."""
         raise NotImplementedError(f"{type(self)} doesn't implement shard_data_sources yet")
+
+    def split_shard_indices_by_worker(self, num_workers: int) -> List[List[int]]:
+        return [list(range(worker_id, self.n_shards, num_workers)) for worker_id in range(num_workers)]
 
     @property
     def n_shards(self) -> int:
@@ -115,9 +117,10 @@ class ExamplesIterable(_BaseExamplesIterable):
     def shuffle_data_sources(self, generator: np.random.Generator) -> "ExamplesIterable":
         return ShuffledDataSourcesExamplesIterable(self.generate_examples_fn, self.kwargs, generator)
 
-    def shard_data_sources(self, shard_indices: List[int]) -> "ExamplesIterable":
+    def shard_data_sources(self, worker_id: int, num_workers: int) -> "ExamplesIterable":
         """Keep only the requested shard."""
         gen_kwargs_list = _split_gen_kwargs(self.kwargs, max_num_jobs=self.n_shards)
+        shard_indices = self.split_shard_indices_by_worker(num_workers)[worker_id]
         requested_gen_kwargs = _merge_gen_kwargs([gen_kwargs_list[i] for i in shard_indices])
         return ExamplesIterable(self.generate_examples_fn, requested_gen_kwargs)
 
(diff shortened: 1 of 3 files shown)


Hubert-Bonisseur:
Hi ! 
I just tested this out with the code below and it seems to be ok. Both datasets are alternating and we get all the examples with no duplicates.

On thing to keep in mind is that the max amount of workers is equal to the lowest amount of shard amongst the datasets to be merged (1 in this example).

 ```python

from datasets import load_dataset, interleave_datasets


def process_dataset_train(batch):
    return {"input": f'train: {batch["review"][:20]}'}


def process_dataset_test(batch):
    return {"input": f'test: {batch["review"][:20]}'}


def identity_collator(x):
    return x


if __name__ == "__main__":
    ds = load_dataset("lhoestq/demo1")
    ds["train"] = ds["train"].map(process_dataset_train, remove_columns=ds["train"].column_names)
    ds["test"] = ds["test"].map(process_dataset_test, remove_columns=ds["test"].column_names)

    ds1 = ds["train"].to_iterable_dataset(num_shards=5)
    ds2 = ds["test"].to_iterable_dataset(num_shards=1)

    ds_merged = interleave_datasets([ds1, ds2], stopping_strategy="all_exhausted")

    dataloader = DataLoader(ds_merged, collate_fn=identity_collator, num_workers=1, batch_size=1)

    for i, element in enumerate(dataloader):
        print(i, element)

```

```
0 [{'input': 'train: Great app! The new v'}]
1 [{'input': 'test: Works with RTL and N'}]
2 [{'input': "train: Great It's not fully"}]
3 [{'input': 'test: Works with RTL SDR W'}]
4 [{'input': 'train: Works on a Nexus 6p '}]
5 [{'input': 'test: Awsome App! Easy to '}]
6 [{'input': 'train: The bandwidth seemed'}]
7 [{'input': "test: I'll forgo the refun"}]
8 [{'input': 'train: Works well with my H'}]
9 [{'input': 'test: looks like a great p'}]
```

**Comment by lhoestq** [comment]
Great ! It's ok to have the max amount of workers is equal to the lowest amount of shard :)

So in the case of `num_workers>min(n_shards_per_dataset)` maybe some workers should turn off, and a warning can probably be shown. This is already the case if you use a single dataset with a single shard and `num_workers>1`.


Right now it seems to raise an error:

```python
  File "/Users/quentinlhoest/hf/datasets/src/datasets/iterable_dataset.py", line 979, in __iter__
    yield from self._iter_pytorch(ex_iterable)
  File "/Users/quentinlhoest/hf/datasets/src/datasets/iterable_dataset.py", line 912, in _iter_pytorch
    for key, example in ex_iterable.shard_data_sources(worker_info.id, worker_info.num_workers):
  File "/Users/quentinlhoest/hf/datasets/src/datasets/iterable_dataset.py", line 259, in shard_data_sources
    [iterable.shard_data_sources(worker_id, num_workers) for iterable in self.ex_iterables],
  File "/Users/quentinlhoest/hf/datasets/src/datasets/iterable_dataset.py", line 259, in <listcomp>
    [iterable.shard_data_sources(worker_id, num_workers) for iterable in self.ex_iterables],
  File "/Users/quentinlhoest/hf/datasets/src/datasets/iterable_dataset.py", line 125, in shard_data_sources
    requested_gen_kwargs = _merge_gen_kwargs([gen_kwargs_list[i] for i in shard_indices])
  File "/Users/quentinlhoest/hf/datasets/src/datasets/utils/sharding.py", line 76, in _merge_gen_kwargs
    for key in gen_kwargs_list[0]
IndexError: list index out of range
```

Hubert-Bonisseur:
Good point. I have fixed the n_shards property of merged iterable datasets so that this warning is raised properly

Hubert-Bonisseur:
Hey @lhoestq, what do you think of the last modifications ?

src/datasets/iterable_dataset.py
```diff
@@ -355,9 +366,13 @@ def shuffle_data_sources(
...
 
-    def shard_data_sources(self, shard_idx: int) -> "HorizontallyConcatenatedMultiSourcesExamplesIterable":
+    def shard_data_sources(
+        self, worker_id: int, num_workers: int
+    ) -> "HorizontallyConcatenatedMultiSourcesExamplesIterable":
         """Either keep only the requested shard, or propagate the request to the underlying iterable."""
-        raise NotImplementedError("Sharding a HorizontallyConcatenatedMultiSourcesExamplesIterable is not implemented")
+        return HorizontallyConcatenatedMultiSourcesExamplesIterable(
+            [iterable.shard_data_sources(worker_id, num_workers) for iterable in self.ex_iterables]
+        )
```
lhoestq, line 375: I think it should still be NotImplemented for HorizontallyConcatenatedMultiSourcesExamplesIterable, otherwise you could end up with subiterables that are not aligned and don't have the same length

```suggestion
        raise NotImplementedError("Sharding a HorizontallyConcatenatedMultiSourcesExamplesIterable is not implemented")
```

lhoestq (commented):
It looks good thanks !

I added a comment for HorizontallyConcatenatedMultiSourcesExamplesIterable, as I think we can't really apply the same strategy for this one.

Could you also add a test that uses interleave_datasets and a DataLoader on sharded datasets before we merge ? You can get some inspiration from the test called `test_sharded_iterable_dataset_torch_dataloader_parallel`

lhoestq:
> But now my issue is that the "first_exhausted" method breaks the loop when one of the datasets of one of the shards is empty which means that all shards stop yielding and we could be missing up to n_workers samples. I don't know if this is the behaviour expected, but I had to modify the test to accomodate this.

This looks reasonable, maybe this can be documented in the `interleave_datasets` docstring ?
```
In a distributed setup or in PyTorch DataLoader workers, the stopping strategy is applied per process.
Therefore the "first_exhausted" strategy on a sharded dataset can generate less samples in total (up to 1 missing sample per subdataset per worker).
```

tests/test_iterable_dataset.py
```diff
@@ -1266,3 +1266,24 @@ def test_interleave_datasets_with_oversampling():
...
+    dataset2 = IterableDataset(ex_iterable2).with_format("torch")
+
+    dataset_merged = interleave_datasets([dataset1, dataset2], stopping_strategy="first_exhausted")
+    assert dataset_merged.n_shards == min(n_shards1, nshards2)
+    dataloader = DataLoader(dataset_merged, batch_size=None, num_workers=num_workers)
+    result = list(dataloader)
+    expected_length = 2 * min(
+        len([example for _, example in ex_iterable1]), len([example for _, example in ex_iterable2])
+    )
+    assert expected_length - num_workers <= len(result) <= expected_length
```
lhoestq, line 1288:
```suggestion
    # some samples may be missing because the stopping strategy is applied per process
    assert expected_length - num_workers <= len(result) <= expected_length
```

lhoestq (approved):
Awesome ! It looks all good :D

PR #5742, opened 2023-04-13
Warning specifying future change in to_tf_dataset behaviour

amyeroberts:
Warning specifying future changes happening to `to_tf_dataset` behaviour when #5602 is merged in

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -381,6 +381,16 @@ class TensorflowDatasetMixin:
         else:
             raise ImportError("Called a Tensorflow-specific function but Tensorflow is not installed.")
 
+        warnings.warn(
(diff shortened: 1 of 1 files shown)


src/datasets/arrow_dataset.py
```diff
@@ -381,6 +381,16 @@ def to_tf_dataset(
         else:
             raise ImportError("Called a Tensorflow-specific function but Tensorflow is not installed.")
 
+        warnings.warn(
```
lhoestq, line 384: Cool thanks !

Maybe we can warn only if the user passes a list with one element ?

amyeroberts: Certainly :) I originally had it like that, and wasn't sure if it was better to let everyone know. Your vote tips the balance - I'll update 👍

lhoestq (approved):
LGTM !

Issue #5744, opened 2023-04-13
[BUG] With Pandas 2.0.0, `load_dataset` raises `TypeError: read_csv() got an unexpected keyword argument 'mangle_dupe_cols'`


keyboardAnt:
The `load_dataset` function with Pandas `1.5.3` has no issue (just a FutureWarning) but crashes with Pandas `2.0.0`.
For your convenience, I opened a draft Pull Request to fix it quickly: https://github.com/huggingface/datasets/pull/5745

---

* The FutureWarning mentioned above: 
```
FutureWarning: the 'mangle_dupe_cols' keyword is deprecated and will be removed in a future version. Please take steps to stop the use of 'mangle_dupe_cols'
```

albertvillanova:
Thanks for reporting, @keyboardAnt.

We haven't noticed any crash in our CI tests. Could you please indicate specifically the `load_dataset` command that crashes in your side, so that we can reproduce it?

lhoestq:
This has been fixed in `datasets` 2.11

pratt3000:
I am still getting this bug with the latest pandas and datasets lib installed. Anyone else?

lhoestq:
Feel free to update `datasets` to fix this issue

```
pip install -U datasets
```

aadarshsingh191198:
I get this with pip when doing `pip install -U datasets`

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
cudf 23.8.0 requires cubinlinker, which is not installed.
cudf 23.8.0 requires cupy-cuda11x>=12.0.0, which is not installed.
cudf 23.8.0 requires ptxcompiler, which is not installed.
cuml 23.8.0 requires cupy-cuda11x>=12.0.0, which is not installed.
dask-cudf 23.8.0 requires cupy-cuda11x>=12.0.0, which is not installed.
apache-beam 2.46.0 requires dill<0.3.2,>=0.3.1.1, but you have dill 0.3.8 which is incompatible.
apache-beam 2.46.0 requires numpy<1.25.0,>=1.14.3, but you have numpy 1.26.4 which is incompatible.
apache-beam 2.46.0 requires pyarrow<10.0.0,>=3.0.0, but you have pyarrow 15.0.1 which is incompatible.
beatrix-jupyterlab 2023.128.151533 requires jupyterlab~=3.6.0, but you have jupyterlab 4.1.2 which is incompatible.
[...]

PR #5745, opened 2023-04-13
[FIX] Issue 5744

keyboardAnt:
A temporal fix for https://github.com/huggingface/datasets/issues/5744.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -115,7 +115,7 @@ REQUIRED_PKGS = [
     # For smart caching dataset processing
     "dill>=0.3.0,<0.3.7",  # tmp pin until next 0.3.7 release: see https://github.com/huggingface/datasets/pull/5166
     # For performance gains with apache arrow
-    "pandas",
+    "pandas==1.5.3",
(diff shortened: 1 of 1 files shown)


albertvillanova (commented):
Thanks for your proposal, @keyboardAnt.

However, in principle, I don't see why we should impose a specific version of pandas from our side.

If end users need a specific pandas version, I think it should be them to pin pandas on their side.

albertvillanova:
Pandas 2.0.0 has removed support to passing `mangle_dupe_cols`.

However, our `datasets` library does not use this parameter: it only passes it to pandas if the user passes it to `load_dataset`.

You should better:
- Either "take steps to stop the use of 'mangle_dupe_cols'" (as it was suggested in the deprecation warning in pandas-1.5.3)
- Or pin pandas (< 2.0.0) in your local requirements file

lhoestq:
`datasets` 2.11 doesn't pass `mangle_dupe_cols` unless the user specifies it indeed, so I think we're fine

PR #5747, opened 2023-04-13
[WIP] Add Dataset.to_spark

maddiedawson:


src/datasets/arrow_dataset.py
```diff
@@ -4906,6 +4906,18 @@ def to_parquet(
...
         return ParquetDatasetWriter(self, path_or_buf, batch_size=batch_size, **parquet_writer_kwargs).write()
 
+    def to_spark(self) -> pyspark.sql.DataFrame:
+        """Creates a Spark DataFrame from the dataset.
+        The DataFrame is backed by the dataset's underlying cache files, so if those files change or are cleaned up, it will become invalid.
+        """
+        files = [entry["filename"] for entry in self.cache_files]
+        for entry in self.cache_files:
+            file = entry["filename"]
+            if not file.endswith(".parquet"):
```
mariosasko, line 4916: `cache_files` are in the Arrow format (and end with `.arrow`), so this will always be `True` 

(Unlike Arrow files, Parquet files cannot be directly mapped from disk)

mariosasko (commented):
Thanks for the effort!

However, converting a Dataset to Parquet and creating a Spark DataFrame with `spark.read.parquet` from them is already simple enough, so I don't think we need a dedicated function.

PR #5784, opened 2023-04-24
Raise subprocesses traceback when interrupting

lhoestq:
When a subprocess hangs in `filter` or `map`, one should be able to get the subprocess' traceback when interrupting the main process. Right now it shows nothing.

To do so I `.get()` the subprocesses async results even the main process is stopped with e.g. `KeyboardInterrupt`. I added a timeout in case the subprocess is hanging or crashed.


lhoestq:
Whoops mario you're off this week sorry. I'm taking the liberty to merge this one

PR #5787, opened 2023-04-24
Fix inferring module for unsupported data files

### LINKED ISSUE #5785: Unsupported data files raise TypeError: 'NoneType' object is not iterable
Currently, we raise a TypeError for unsupported data files:
```
TypeError: 'NoneType' object is not iterable
```
See:
- https://github.com/huggingface/datasets-server/issues/1073

We should give a more informative error message.

albertvillanova:
This PR raises a FileNotFoundError instead:
```
FileNotFoundError: No (supported) data files or dataset script found in <dataset_name>
```

Fix #5785.

### DIFF
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -368,6 +368,7 @@ def infer_module_for_data_files(
                 return _EXTENSION_TO_MODULE[ext]
             elif ext == "zip":
                 return infer_module_for_data_files_in_archives(data_files_list, use_auth_token=use_auth_token)
+    return None, {}
 
 
 def infer_module_for_data_files_in_archives(
@@ -404,6 +405,7 @@ def infer_module_for_data_files_in_archives(
         most_common = extensions_counter.most_common(1)[0][0]
         if most_common in _EXTENSION_TO_MODULE:
             return _EXTENSION_TO_MODULE[most_common]
+    return None, {}
 
 
 @dataclass
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -626,7 +626,9 @@ class LocalDatasetModuleFactoryWithoutScript(_DatasetModuleFactory):
 
     def get_module(self) -> DatasetModule:
         base_path = os.path.join(self.path, self.data_dir) if self.data_dir else self.path
-        patterns = sanitize_patterns(self.data_files) if self.data_files else get_data_patterns_locally(base_path)
+        patterns = (
+            sanitize_patterns(self.data_files) if self.data_files is not None else get_data_patterns_locally(base_path)
+        )
         data_files = DataFilesDict.from_local_or_remote(
             patterns,
             base_path=base_path,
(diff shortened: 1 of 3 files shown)


src/datasets/load.py
```diff
@@ -632,14 +634,14 @@ def get_module(self) -> DatasetModule:
...
         )
-        module_names = {
-            key: infer_module_for_data_files(data_files_list) for key, data_files_list in data_files.items()
+        split_modules = {
+            split: infer_module_for_data_files(data_files_list) for split, data_files_list in data_files.items()
         }
-        if len(set(list(zip(*module_names.values()))[0])) > 1:
-            raise ValueError(f"Couldn't infer the same data file format for all splits. Got {module_names}")
-        module_name, builder_kwargs = next(iter(module_names.values()))
+        module_name, builder_kwargs = next(iter(split_modules.values()))
```
lhoestq, line 640: Maybe check that i split_modules is not empty ?

src/datasets/load.py
```diff
@@ -772,15 +774,15 @@ def get_module(self) -> DatasetModule:
...
-            key: infer_module_for_data_files(data_files_list, use_auth_token=self.download_config.use_auth_token)
-            for key, data_files_list in data_files.items()
+        split_modules = {
+            split: infer_module_for_data_files(data_files_list, use_auth_token=self.download_config.use_auth_token)
+            for split, data_files_list in data_files.items()
         }
-        if len(set(list(zip(*module_names.values()))[0])) > 1:
-            raise ValueError(f"Couldn't infer the same data file format for all splits. Got {module_names}")
-        module_name, builder_kwargs = next(iter(module_names.values()))
+        module_name, builder_kwargs = next(iter(split_modules.values()))
```
lhoestq, line 781: same here

lhoestq (approved):
Thanks !

src/datasets/load.py
albertvillanova: Thanks for your review, @lhoestq.

I think it can only be empty if the user passes `data_files={}`, otherwise there are 2 options: either it is not empty or an exception is raised.
- `split_modules` is derived from `data_files`, which is instance of `DataFilesDict.from_local_or_remote` with `patterns`
- `patterns` is derived either from `sanitize_patterns` or `get_data_patterns_locally`
  -  `sanitize_patterns` can only return an empty dict if the user passes `data_files={}`
  - `get_data_patterns_locally` can only return a non-empty dict or raise a `EmptyDatasetError`

I think the validation of  `data_files={}` should be elsewhere though. What do you think?

albertvillanova: Maybe changing?
```python
sanitize_patterns(self.data_files) if self.data_files is not None else get_data_patterns_locally(base_path)
```
to
```python
sanitize_patterns(self.data_files) if self.data_files else get_data_patterns_locally(base_path)
```
This way, we are sure `split_modules` is never empty.

lhoestq: > I think the validation of data_files={} should be elsewhere though. What do you think?

Yea indeed, probably in load_dataset_builder ?

> Maybe changing?

I think it's better if it raises an error rather than trying to make it run with data files that were not requested

lhoestq: Feel free to merge then :)

lhoestq:
I think you can revert the last commit - it should fail if data_files={} IMO

Issue #5798, opened 2023-04-27
Support parallelized downloading and processing in load_dataset with Spark


es94129:
### Feature request

When calling `load_dataset` for datasets that have multiple files, support using Spark to distribute the downloading and processing job to worker nodes when `cache_dir` is a cloud file system shared among nodes.

```python
load_dataset(..., use_spark=True)
```

### Motivation

Further speed up `dl_manager.download` and `_prepare_split` by distributing the workloads to worker nodes. 

### Your contribution

I can submit a PR to support this.

lhoestq:
Hi ! We're using process pools for parallelism right now. I was wondering if there's a package that implements the same API as a process pool but runs with Spark under the hood ? That or something similar would be cool because users could use whatever distributed framework they want this way.

Feel free to ping us when you'd like to open PRs for this kind of things, so that we can discuss this before you start working on it ^^

es94129:
Hi, thanks for taking a look and providing your input! I don't know of such packages, and even it exists, I don't think with the process pool API it's possible to run Spark as backend properly; otherwise I understand a unified API would be preferable.

The process pool API requires splitting the workload to a fixed number parts for multiprocessing; meanwhile distributed framework such as Spark has sophisticated scheduler to distribute the workload to the processes on multiple machines in a cluster, so the way of splitting things for `multiprocessing.pool` would not suit / be as flexible as directly calling the `sparkContext.parallelize` API.

I think this could be a good addition to scale the `datasets` implementation to distributed workers, and from my benchmark results so far it looks promising compared with multiprocessing.

es94129:
For Spark, `rdd.map` (where `rdd` can be created by `sparkContext.parallelize`) is the most similar as `pool.map`, but it requires creating a Spark RDD first that is used for distributing the `iterable` and the actual parallelization is managed by the Spark framework; `pool.map` takes the splits of `iterable` that are split into `num_proc` parts by the Python code. You can also check my PR #5807 in the `src/datasets/utils/py_utils.py` file to compare the differences of the APIs, it might make more sense than the the above description.

Given the different inputs and mechanisms of calling the `map` functions, this is why I think it's not that feasible to reuse most of the `multiprocessing` code.

Progress bar updating might be challenging with Spark, I'll consider it as a followup work.

lhoestq:
Indeed I think the current use of multiprocessing.Pool in `map_nested` can be rewritten to work like `sparkContext.parallelize` - without splitting the iterable.

[...]

PR #5807, opened 2023-04-28
Support parallelized downloading in load_dataset with Spark

es94129:
As proposed in https://github.com/huggingface/datasets/issues/5798, this adds support to parallelized downloading in `load_dataset` with Spark, which can speed up the process by distributing the workload to worker nodes.

Parallelizing dataset processing is not supported in this PR.

### DIFF
--- a/src/datasets/utils/py_utils.py
+++ b/src/datasets/utils/py_utils.py
@@ -381,6 +381,7 @@ def map_nested(
     types: Optional[tuple] = None,
     disable_tqdm: bool = True,
     desc: Optional[str] = None,
+    use_spark: bool = False,
 ) -> Any:
     """Apply a function recursively to each element of a nested data struct.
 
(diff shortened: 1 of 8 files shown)


src/datasets/utils/spark_utils.py
```diff
@@ -0,0 +1,35 @@
+import os
+import uuid
+
+import pyspark
+
+
+def validate_cache_dir(cache_dir, spark_function_name):
```
maddiedawson, line 7: Let's add an option to pass in spark so that Spark's self._spark can be reused?

maddiedawson: Also let's add a comment on this function that also describes what `spark_function_name` is being used for (error message)

maddiedawson (commented):
Looks good overall!

src/datasets/utils/py_utils.py
```diff
@@ -439,11 +440,29 @@ def map_nested(
...
+            import pyspark
+
+            spark = pyspark.sql.SparkSession.builder.getOrCreate()
+            mapped = (
+                spark.sparkContext.parallelize(iterable)
+                .map(lambda obj: _single_map_nested((function, obj, types, None, True, None)))
+                .collect()
+            )
+
+        except ImportError:
```
lu-wang-dl, line 460: We may need to provide some more details for the exceptions. If the import failed, we need to give a warning saying something like "pyspark module is not found. Downloading in sequentially". If spark session creation fails, we need to give corresponding messages.

src/datasets/utils/spark_utils.py
```diff
@@ -0,0 +1,48 @@
+import os
+import uuid
+from typing import Optional
+
+import pyspark
```
lu-wang-dl, line 5: We need to put this import into try catch. pyspark is not a hard dependencies for datasets.

es94129:
Hi @lhoestq or other maintainers, this is ready for review, could you please take a look?

PR #5826, opened 2023-05-05
Support working_dir in from_spark

maddiedawson:
Accept `working_dir` as an argument to `Dataset.from_spark`. Setting a non-NFS working directory for Spark workers to materialize to will improve write performance.

### DIFF
--- a/src/datasets/packaged_modules/spark/spark.py
+++ b/src/datasets/packaged_modules/spark/spark.py
@@ -91,7 +91,6 @@ class Spark(datasets.DatasetBuilder):
 
         self._spark = pyspark.sql.SparkSession.builder.getOrCreate()
         self.df = df
-        self._validate_cache_dir(cache_dir)
         self._working_dir = working_dir
 
         super().__init__(
@@ -169,7 +168,8 @@ class Spark(datasets.DatasetBuilder):
         import pyspark
 
         writer_class = ParquetWriter if file_format == "parquet" else ArrowWriter
-        working_fpath = os.path.join(self._working_dir, os.path.basename(fpath)) if self._working_dir else fpath
+        working_dir = self._working_dir or os.environ.get("HF_WORKING_DIR")
+        working_fpath = os.path.join(working_dir, os.path.basename(fpath)) if working_dir else fpath
         embed_local_files = file_format == "parquet"
 
         # Define these so that we don't reference self in write_arrow, which will result in a pickling error due to
--- a/src/datasets/packaged_modules/spark/spark.py
+++ b/src/datasets/packaged_modules/spark/spark.py
@@ -168,8 +168,7 @@ class Spark(datasets.DatasetBuilder):
         import pyspark
 
         writer_class = ParquetWriter if file_format == "parquet" else ArrowWriter
-        working_dir = self._working_dir or os.environ.get("HF_WORKING_DIR")
-        working_fpath = os.path.join(working_dir, os.path.basename(fpath)) if working_dir else fpath
+        working_fpath = os.path.join(self._working_dir, os.path.basename(fpath)) if self._working_dir else fpath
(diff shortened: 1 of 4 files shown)


src/datasets/packaged_modules/spark/spark.py
```diff
@@ -85,6 +87,28 @@ def _info(self):
...
+        import pyspark
+
+        def get_arrow_batch_size(it):
+            for batch in it:
+                yield pa.RecordBatch.from_pydict({"batch_bytes": [batch.nbytes]})
+
+        df_num_rows = self.df.count()
+        # Approximate the size of each row (in Arrow format) by averaging over a 100-row sample.
+        approx_bytes_per_row = (
+            self.df.limit(100)
```
lu-wang-dl, line 100: What happened if the number of rows are less than 100?

src/datasets/packaged_modules/spark/spark.py
```diff
@@ -85,6 +87,28 @@ def _info(self):
     def _split_generators(self, dl_manager: datasets.download.download_manager.DownloadManager):
         return [datasets.SplitGenerator(name=datasets.Split.TRAIN)]
 
+    def _repartition_df_if_needed(self, max_shard_size):
```
lu-wang-dl, line 90: Could we add a test for this function?

src/datasets/packaged_modules/spark/spark.py
maddiedawson: Fixed

src/datasets/packaged_modules/spark/spark.py
maddiedawson: Done

src/datasets/packaged_modules/spark/spark.py
```diff
@@ -85,6 +87,31 @@ def _info(self):
...
+        import pyspark
```
lu-wang-dl, line 91: Do we need to import pyspark inside the function?

lu-wang-dl (commented):
Could we also add the support by setting an environment variable for the working directory?

src/datasets/packaged_modules/spark/spark.py
maddiedawson: I don't think we need it since we imported in the ctor. Removed

maddiedawson:
Added env var

src/datasets/packaged_modules/spark/spark.py
maddiedawson: Never mind, tests failed with NameError so added it back

maddiedawson:
@lhoestq would you be able to review please? :)

lu-wang-dl (approved):

src/datasets/packaged_modules/spark/spark.py
```diff
@@ -196,6 +225,11 @@ def write_arrow(it):
                     names=["task_id", "num_examples", "num_bytes"],
                 )
 
+            if working_fpath != fpath:
+                for file in os.listdir(os.path.dirname(working_fpath)):
+                    dest = os.path.join(os.path.dirname(fpath), os.path.basename(file))
+                    os.rename(file, dest)
```
lhoestq, line 231: os.rename doesn't always work depending on the source/target disks, you can use shutil.move instead

lhoestq (commented):
Nice ! I think `HF_WORKING_DIR` is not a good env variable name for this case, I guess you can remove it.

src/datasets/packaged_modules/spark/spark.py
maddiedawson: Thanks, done!

maddiedawson:
I removed the env var

lhoestq (approved):
LGTM :)

maddiedawson:
Thank you!

PR #5836, opened 2023-05-09
[docs] Custom decoding transforms

### LINKED ISSUE #5782: Support for various audio-loading backends instead of always relying on SoundFile
### Feature request

Introduce an option to select from a variety of audio-loading backends rather than solely relying on the SoundFile library. For instance, if the ffmpeg library is installed, it can serve as a fallback loading option.

### Motivation

- The SoundFile library, used in [features/audio.py](https://github.com/huggingface/datasets/blob/649d5a3315f9e7666713b6affe318ee00c7163a0/src/datasets/features/audio.py#L185), supports only a [limited number of audio formats](https://pysoundfile.readthedocs.io/en/latest/index.html?highlight=supported#soundfile.available_formats). 
- However, current methods for creating audio datasets permit the inclusion of audio files in formats not supported by SoundFile. 
- As a result, developers may potentially create a dataset they cannot read back.

In my most recent project, I dealt with phone call recordings in `.amr` or `.gsm` formats and was genuinely surprised when I couldn't read the dataset I had just packaged a minute prior. Nonetheless, I can still accurately read these files using the librosa library, which employs the audioread library that internally leverages ffmpeg to read such files.

Example:
```python
audio_dataset_amr = Dataset.from_dict({"audio": ["audio_samples/audio.amr"]}).cast_column("audio", Audio())
audio_dataset_amr.save_to_disk("audio_dataset_amr")
audio_dataset_amr = Dataset.load_from_disk("audio_dataset_amr")
print(audio_dataset_amr[0])
```
Results in:
```
...
    raise LibsndfileError(err, prefix="Error opening {0!r}: ".format(self.name))
soundfile.LibsndfileError: Error opening <_io.BytesIO object at 0x7f316323e4d0>: Format not recognised.
```

While I acknowledge that support for these rare file types may not be a priority, I believe it's quite unfortunate that it's possible to create an unreadable dataset in this manner.

### Your contribution

I've created a [simple demo repository](https://github.com/BoringDonut/hf-datasets-ffmpeg-audio) that highlights the mentioned issue. It demonstrates how to create an .amr dataset that results in an error when attempting to read it just a few lines later.

Additionally, I've made a [fork with a rudimentary solution](https://github.com/BoringDonut/datasets/blob/fea73a8fbbc8876467c7e6422c9360546c6372d8/src/datasets/features/audio.py#L189) that utilizes ffmpeg to load files not supported by SoundFile.

Here you may see github actions fails to read `.amr` dataset using the version of the current dataset, but will work with the patched version:
- https://github.com/BoringDonut/hf-datasets-ffmpeg-audio/actions/runs/4773780420/jobs/8487063785
- https://github.com/BoringDonut/hf-datasets-ffmpeg-audio/actions/runs/4773780420/jobs/8487063829

As evident from the GitHub action above, this solution resolves the previously mentioned problem.

I'd be happy to create a proper pull request, provide runtime benchmarks and tests if you could offer some guidance on the following:
- Where should I incorporate the ffmpeg (or other backends) code? For example, should I create a new file or simply add a function within the Audio class?
- Is it feasible to pass the audio-loading function as an argument within the current architecture? This would be useful if I know in advance that I'll be reading files not supported by SoundFile.

A few more notes:
- In theory, it's possible to load audio using librosa/audioread since librosa is already expected to be installed. However, librosa [will soon discontinue audioread support](https://github.com/librosa/librosa/blob/aacb4c134002903ae56bbd4b4a330519a5abacc0/librosa/core/audio.py#L227). Moreover, using audioread on its own seems inconvenient because it requires a file [path as input](https://github.com/beetbox/audioread/blob/ff9535df934c48038af7be9617fdebb12078cc07/audioread/__init__.py#L108) and cannot work with bytes already loaded into memory or an open file descriptor (as mentioned in [librosa docs](https://librosa.org/doc/main/generated/librosa.load.html#librosa.load), only SoundFile backend supports an open file descriptor as an input).

stevhliu:
Adds custom decoding transform solution to the docs to fix #5782.

### DIFF
--- a/docs/source/process.mdx
+++ b/docs/source/process.mdx
@@ -594,6 +594,18 @@ The [`~Dataset.set_transform`] function applies a custom formatting transform on
 {'type': 'custom', 'format_kwargs': {'transform': <function __main__.encode(batch)>}, 'columns': ['idx', 'label', 'sentence1', 'sentence2'], 'output_all_columns': False}
 ```
 
+You can also use the [`~Dataset.set_transform`] function to decode audio formats not supported by the `soundfile` library. You'll need to create a custom decoding function to decode the audio files not supported by `soundfile`. For example, you can use the `read_ffmpeg` function from the `ffmpeg` library to read the audio files and decode them into an acceptable format. Then use [`~Dataset.set_transform`] to apply the custom decoding transform on the dataset:
+
+```py
+audio_dataset_amr = Dataset.from_dict({"audio": ["audio_samples/audio.amr"]})
+
+def decode_audio(batch):
+    batch["audio"] = [read_ffmpeg(audio_path) for audio_path in batch["audio"]]
+    return batch
+
+audio_dataset_amr.set_transform(decode_amr)
+```
+
 ## Save
 
 Once you are done processing your dataset, you can save and reuse it later with [`~Dataset.save_to_disk`].
--- a/docs/source/process.mdx
+++ b/docs/source/process.mdx
@@ -594,16 +594,30 @@ The [`~Dataset.set_transform`] function applies a custom formatting transform on
 {'type': 'custom', 'format_kwargs': {'transform': <function __main__.encode(batch)>}, 'columns': ['idx', 'label', 'sentence1', 'sentence2'], 'output_all_columns': False}
 ```
 
-You can also use the [`~Dataset.set_transform`] function to decode audio formats not supported by the `soundfile` library. You'll need to create a custom decoding function to decode the audio files not supported by `soundfile`. For example, you can use the `read_ffmpeg` function from the `ffmpeg` library to read the audio files and decode them into an acceptable format. Then use [`~Dataset.set_transform`] to apply the custom decoding transform on the dataset:
+You can also use the [`~Dataset.set_transform`] function to decode formats not supported by [`Features`]. For example, the [`Audio`] feature uses [`soundfile`](https://python-soundfile.readthedocs.io/en/0.11.0/) - a fast and simple library to install - but it does not provide support for less common audio formats. Here is where you can use [`~Dataset.set_transform`] to apply a custom decoding transform on the fly. You're free to use any library you like to decode the audio files.
+
+The example below uses the [`pydub`](http://pydub.com/) package to open an audio format not supported by `soundfile`:
 
 ```py
-audio_dataset_amr = Dataset.from_dict({"audio": ["audio_samples/audio.amr"]})
+import numpy as np
+from pydub import AudioSegment
 
-def decode_audio(batch):
-    batch["audio"] = [read_ffmpeg(audio_path) for audio_path in batch["audio"]]
+audio_dataset_amr = Dataset.from_dict({"audio": [""audio_samples/audio.amr""]})
+
+def decode_audio_with_pydub(batch, sampling_rate=16_000):
+    def pydub_decode_file(audio_path):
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
Nice addition :) thanks

docs/source/process.mdx
```diff
@@ -594,6 +594,18 @@ The [`~Dataset.set_transform`] function applies a custom formatting transform on
...
+You can also use the [`~Dataset.set_transform`] function to decode audio formats not supported by the `soundfile` library. You'll need to create a custom decoding function to decode the audio files not supported by `soundfile`. For example, you can use the `read_ffmpeg` function from the `ffmpeg` library to read the audio files and decode them into an acceptable format. Then use [`~Dataset.set_transform`] to apply the custom decoding transform on the dataset:
+
+```py
+audio_dataset_amr = Dataset.from_dict({"audio": ["audio_samples/audio.amr"]})
+
+def decode_audio(batch):
+    batch["audio"] = [read_ffmpeg(audio_path) for audio_path in batch["audio"]]
+    return batch
+
+audio_dataset_amr.set_transform(decode_amr)
```
mariosasko, line 606: This is a working example (it uses `pydub`, which calls `subprocess` + `ffmpeg` under the hood)
```suggestion
import numpy as np
from pydub import AudioSegment

audio_dataset_amr = Dataset.from_dict({"audio": [""audio_samples/audio.amr""]})

def decode_audio_with_pydub(batch, sampling_rate=16_000):
    def pydub_decode_file(audio_path):
      sound = AudioSegment.from_file(audio_path)
      if sound.frame_rate != sampling_rate:
        sound = sound.set_frame_rate(sampling_rate)
      channel_sounds = sound.split_to_mono()
      samples = [s.get_array_of_samples() for s in channel_sounds]

      fp_arr = np.array(samples).T.astype(np.float32)
      fp_arr /= np.iinfo(samples[0].typecode).max
      return fp_arr

    batch["audio"] = [pydub_decode_file(audio_path) for audio_path in batch["audio"]]
    return batch

audio_dataset_amr.set_transform(decode_audio_with_pydub)
```

mariosasko (commented):
Thanks!

docs/source/process.mdx
```diff
@@ -594,6 +594,32 @@ The [`~Dataset.set_transform`] function applies a custom formatting transform on
...
 
+You can also use the [`~Dataset.set_transform`] function to decode formats not supported by [`Features`]. For example, the [`Audio`] feature uses [`soundfile`](https://python-soundfile.readthedocs.io/en/0.11.0/) - a fast and simple library to install - but it does not provide support for less common audio formats. Here is where you can use [`~Dataset.set_transform`] to apply a custom decoding transform on the fly. You're free to use any library you like to decode the audio files.
+
+The example below uses the [`pydub`](http://pydub.com/) package to open an audio format not supported by `soundfile`:
+
+```py
+import numpy as np
+from pydub import AudioSegment
+
+audio_dataset_amr = Dataset.from_dict({"audio": [""audio_samples/audio.amr""]})
```
mariosasko, line 605:
```suggestion
audio_dataset_amr = Dataset.from_dict({"audio": ["audio_samples/audio.amr"]})
```

mariosasko (approved):
LGTM :)

mariosasko:
The error seems unrelated to the changes, so feel free to merge.

PR #5850, opened 2023-05-12
Make packaged builders skip non-supported file formats

### LINKED ISSUE #5849: CSV datasets should only read the CSV data files in the repo
When a no-script dataset has many CSV files and a JPG file, the library infers to use the Csv builder, but tries to read as CSV all files in the repo, also the JPG file.

I think the Csv builder should filter out non-CSV files when reading.

An analogue solution should be implemented for other packaged builders.

albertvillanova:
Fix #5849.

### DIFF
--- a/src/datasets/packaged_modules/csv/csv.py
+++ b/src/datasets/packaged_modules/csv/csv.py
@@ -1,4 +1,5 @@
 import itertools
+import os.path
 from dataclasses import dataclass
 from typing import Any, Callable, Dict, List, Optional, Union
 
@@ -128,6 +129,7 @@ class CsvConfig(datasets.BuilderConfig):
 
 class Csv(datasets.ArrowBasedBuilder):
     BUILDER_CONFIG_CLASS = CsvConfig
+    _EXTENSIONS = [".csv", ".tsv"]
 
     def _info(self):
         return datasets.DatasetInfo(features=self.config.features)
@@ -174,6 +176,8 @@ class Csv(datasets.ArrowBasedBuilder):
             else None
         )
         for file_idx, file in enumerate(itertools.chain.from_iterable(files)):
+            if os.path.splitext(file)[1] not in self._EXTENSIONS:
+                continue
             csv_file_reader = pd.read_csv(file, iterator=True, dtype=dtype, **self.config.pd_read_csv_kwargs)
             try:
                 for batch_idx, df in enumerate(csv_file_reader):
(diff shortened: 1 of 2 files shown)


mariosasko (commented):
Nice! 

I think it's better to perform the filtering in `_split_generators` than in `_generate_examples`, to ensure each process gets the same number of supported data files in the multi-proc mode.

albertvillanova:
Good idea. @mariosasko!!!

lhoestq:
This change could create silent problems when loading files with extensions that are not listed here. For example

```python
load_dataset("text", data_files=["20230515.log"])
```

wouldn't even log anything to say that the file was ignored.

Maybe it's possible to do this at data files patterns resolution ?

e.g. in get_data_patterns_in_dataset_repository / get_data_patterns_locally we could return patterns that include the most common extension

albertvillanova:
@lhoestq the issue you evoke (.log files skipped by text builder if .log is not added to .txt as supported extension) persists whether you perform the skip at the pattern resolution or in the builder itself.

The solution is to add the .log extension (besides the .txt) as supported by text, independently of where we perform the skip (at pattern resolution or in the builder itself).

Additionally, at the time we call for pattern resolution, we do not know the builder class yet, so that we cannot pass specific file extensions. Once we call data files pattern resolution, afterwards we call `infer_module_for_data_files` and then know the builder class.

mariosasko:
I agree with @lhoestq - it should still be possible to request parsing a file with a specific builder even if the file's extension is "invalid" for the builder, and only ignore non-supported file formats when inferring the patterns.

albertvillanova:
Therefore, if I understand correctly, what you suggest is:
- if the user passes a packaged builder to `load_dataset` (e.g. `load_dataset("csv",...`), then the passed `data_files` should not be filtered to remove unsupported extensions. No breaking change in this case
- if the user passes a no-script repo/folder to `load_dataset` (e.g. load_dataset("my_dataset_repository",...`), then the inferred data files should be filtered to remove the extensions that are not supported by the inferred module name builder

lhoestq:
Yes that would be ideal imo !

albertvillanova:
I think this now fulfills all the requirements.

lhoestq:
I find it a bit confusing to still be able to pass data_files that are going to be silently ignored based on the value of `only_supported_extensions`. My suggestion was to have the right data files pattern, not to filter a posteriori (sorry if my last message was confusing).

Having the right data files pattern would also allow users to inspect what's actually being loaded with
```
load_dataset_builder(...).config.data_files
```
and it would list exactly what data files are used.

mariosasko (commented):
I also think the data files should be filtered before passing them to the builders to avoid exposing an internal parameter (`only_supported_extensions`) via the configs.

@lhoestq 
> My suggestion was to have the right data files pattern, not to filter a posteriori (sorry if my last message was confusing).

I think the data files patterns approach could be slow when the number of supported extensions > 1, so maybe we can just use the `in` operator after the initial resolution

Issue #5793, opened 2023-04-26
IterableDataset.with_format("torch") not working


jiangwy99:
### Describe the bug

After calling the with_format("torch") method on an IterableDataset instance, the data format is unchanged.

### Steps to reproduce the bug

```python
from datasets import IterableDataset
def gen():
    for i in range(4):
        yield {"a": [i] * 4}
dataset = IterableDataset.from_generator(gen).with_format("torch")
next(iter(dataset))
```

### Expected behavior

`{"a": torch.tensor([0, 0, 0, 0])}` is expected, but `{"a": [0, 0, 0, 0]}` is observed.

### Environment info

```bash
platform==ubuntu 22.04.01
python==3.10.9
datasets==2.11.0
```

lhoestq:
Hi ! Thanks for reporting, I'm working on it ;)

PR #5852, opened 2023-05-12
Iterable torch formatting

lhoestq:
Used the TorchFormatter to get torch tensors in iterable dataset with format set to "torch".

It uses the data from Arrow if possible, otherwise applies recursive_tensorize.

When set back to format_type=None, cast_to_python_objects is used.

requires https://github.com/huggingface/datasets/pull/5821

close https://github.com/huggingface/datasets/issues/5793

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -671,7 +671,7 @@ class MappedExamplesIterable(_BaseExamplesIterable):
                 ):  # ignore last batch
                     return
                 batch = _examples_to_batch(examples)
-                batch = format_dict(batch if format_dict else batch)
+                batch = format_dict(batch) if format_dict else batch
                 # then apply the transform
                 inputs = batch
                 function_args = [inputs] if self.input_columns is None else [inputs[col] for col in self.input_columns]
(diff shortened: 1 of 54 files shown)


lhoestq:
@albertvillanova could you take a look at this one ? It directly follows the arrow formatting PR

src/datasets/iterable_dataset.py
```diff
@@ -1118,22 +1148,25 @@ def __init__(
         ex_iterable: _BaseExamplesIterable,
         info: Optional[DatasetInfo] = None,
         split: Optional[NamedSplit] = None,
-        format_type: Optional[str] = None,
+        formatting: Optional[FormattingConfig] = None,
         shuffling: Optional[ShufflingConfig] = None,
         distributed: Optional[DistributedConfig] = None,
         token_per_repo_id: Optional[Dict[str, Union[str, bool, None]]] = None,
+        format_type="deprecated",
```
albertvillanova, line 1155: Nice you deprecate it. What about in the classes above?
- MappedExamplesIterable
- FilteredExamplesIterable

src/datasets/iterable_dataset.py
```diff
@@ -1118,22 +1148,25 @@ def __init__(
...
     ):
         if distributed and distributed.world_size > 1 and shuffling and shuffling._original_seed is None:
             raise RuntimeError(
                 "The dataset doesn't have a fixed random seed across nodes to shuffle and split the list of dataset shards by node. "
                 "Please pass e.g. `seed=42` in `.shuffle()` to make all the nodes use the same seed. "
             )
+        if format_type != "deprecated":
+            formatting = FormattingConfig(format_type=format_type)
```
albertvillanova, line 1163: Maybe worth raising a warning?

albertvillanova (commented):
Thanks. Some comments below.

lhoestq:
I added tests for the `__array__` case which lets you go from any tensor format to any other tensor format.

I also properly deprecated format_type and added a warning message.

albertvillanova (approved):
Thank you!!!

PR #5853, opened 2023-05-12
[docs] Redirects, migrated from nginx

julien-c:


mishig25 (approved):
lgtm !

the format is looking correct (working example [here](https://github.com/huggingface/optimum/blob/main/docs/source/_redirects.yml))
```
old_page: new_page
```

lhoestq (approved):

julien-c:
@mishig25 note that it's not exactly the same behavior as in nginx as here it interacts a bit with the `version` and the `language`

Should be close enough, though.

PR #5860, opened 2023-05-15
Minor tqdm optim

lhoestq:
Don't create a tqdm progress bar when `disable_tqdm` is passed to `map_nested`.

On my side it sped up some iterable datasets by ~30% when `map_nested` is used extensively to recursively tensorize python dicts.


mariosasko (approved):
Nice :)!

PR #5861, opened 2023-05-15
Better error message when combining dataset dicts instead of datasets

lhoestq:
close https://github.com/huggingface/datasets/issues/5851

### DIFF
--- a/src/datasets/combine.py
+++ b/src/datasets/combine.py
@@ -1,6 +1,7 @@
 from typing import List, Optional, TypeVar
 
 from .arrow_dataset import Dataset, _concatenate_map_style_datasets, _interleave_map_style_datasets
+from .dataset_dict import DatasetDict, IterableDatasetDict
 from .info import DatasetInfo
 from .iterable_dataset import IterableDataset, _concatenate_iterable_datasets, _interleave_iterable_datasets
 from .splits import NamedSplit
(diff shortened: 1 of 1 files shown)


mariosasko (approved):
LGTM!

We could use this for-loop to check if all input datasets share the same type (`Dataset`/`Iterable`), but maybe the current solution is easier to read.

src/datasets/combine.py
```diff
@@ -119,6 +120,17 @@ def interleave_datasets(
 
     if not datasets:
         raise ValueError("Unable to interleave an empty list of datasets.")
+    for i, dataset in enumerate(datasets):
+        if isinstance(dataset, (DatasetDict, IterableDatasetDict)):
+            if not datasets[i]:
```
albertvillanova, line 125: Nit: better use `dataset` than `datasets[i]`. The same in all cases below.

albertvillanova (changes requested):
Thanks for the fix.

I find the validations in both methods quite convoluted:
- first (in the loop you have implemented) we iterate over all `datasets` to check if an element is instance of `DatasetDict` or `IterableDatasetDict`
- second we take the first element of `datasets` (`datasets[0]`) to check if it is not an instance of `Dataset` or `IterableDataset`
- third we iterate over the rest of element of `datasets` (`datasets[[1:]]`) to check if they are an instance of the same type as the first element

Maybe we could really simplify the validation by just checking if the elements in the list of `datasets` are:
- either all an instance of `Dataset` 
- or all an instance of `IterableDataset`

and raise a ValueError otherwise

lhoestq:
Having those different checks helps providing an appropriate error message.

If the input is a dict, we suggest to select a split. If the input lists is a mix of iterable and non-iterable, we mention that it must be one or the other.

albertvillanova (commented):
Do we really need to iterate over all datasets 2 times to validate them?

Feel free to merge as it is...

PR #5863, opened 2023-05-15
Use a new low-memory approach for tf dataset index shuffling

### LINKED ISSUE #5855: `to_tf_dataset` consumes too much memory
### Describe the bug

Hi, I'm using `to_tf_dataset` to convert a _large_ dataset to `tf.data.Dataset`. I observed that the data loading before training took a lot of time and memory, even with `batch_size=1`.

After some digging, i believe the reason lies in the shuffle behavior. The [source code](https://github.com/huggingface/datasets/blob/main/src/datasets/utils/tf_utils.py#L185) uses `len(dataset)` as the `buffer_size`, which may load all the data into the memory, and the [tf.data doc](https://www.tensorflow.org/guide/data#randomly_shuffling_input_data) also states that "While large buffer_sizes shuffle more thoroughly, they can take a lot of memory, and significant time to fill".

```python
from datasets import Dataset

def gen():  # some large data
    for i in range(50000000):
        yield {"data": i}

ds = Dataset.from_generator(gen, cache_dir="./huggingface")

tf_ds = ds.to_tf_dataset(
    batch_size=64,
    shuffle=False,  # no shuffle
    drop_remainder=False,
    prefetch=True,
)

# fast and memory friendly 🤗
for batch in tf_ds: 
    ...

tf_ds_shuffle = ds.to_tf_dataset(
    batch_size=64,
    shuffle=True,
    drop_remainder=False,
    prefetch=True,
)

# slow and memory hungry 😱
for batch in tf_ds_shuffle: 
    ...

```

### Expected behavior

Shuffling should not load all the data into the memory. Would adding a `buffer_size` parameter in the `to_tf_dataset` API alleviate the problem?

### Environment info

- `datasets` version: 2.11.0
- Platform: Linux-5.17.1-051701-generic-x86_64-with-glibc2.17
- Python version: 3.8.13
- Huggingface_hub version: 0.13.4
- PyArrow version: 11.0.0
- Pandas version: 1.4.3

Rocketknight1:
This PR tries out a new approach to generating the index tensor in `to_tf_dataset`, which should reduce memory usage for very large datasets. I'll need to do some testing before merging it!

Fixes #5855

### DIFF
--- a/src/datasets/utils/tf_utils.py
+++ b/src/datasets/utils/tf_utils.py
@@ -15,6 +15,8 @@
 """TF-specific utils import."""
 
 import os
+import warnings
+from functools import partial
 from math import ceil
 from uuid import uuid4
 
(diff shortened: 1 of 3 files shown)


Rocketknight1:
The approach we take here is to no longer materialize the entire index array or shuffle buffer. Instead, we do the following:

1) Generate a dataset with `tf.data.Dataset.range`. This dataset is not materialized - it's basically a range iterator.
2) When we begin iterating over a dataset, generate a random seed. This value is constant for each pass over the dataset, and is regenerated if we start a new iteration or epoch over the dataset.
3) Use the function `tf.random.index_shuffle`. This converts indices into the equivalent values in a permuted array. In other words `tf.random.index_shuffle(indices, maxval=10)` is equivalent to `np.random.permutation(10)[indices]`, but without ever materializing the `np.random.permutation(10)` array.

Using this approach gives us a complete iteration over the dataset that does not skip any samples, compiles in TF and also never materializes the complete index array, which should avoid the memory usage issues. I'm testing that now!

Rocketknight1:
Looks good in testing - this should be ready for review! cc @lhoestq @massquantity

massquantity:
Looks good to me, but i doubt that very few people will update to TF >= 2.9:)

lhoestq:
Is it more efficient than using numpy to shuffle as in multiprocessing ? Why not use the same strategy ?

Rocketknight1:
Good question, honestly! The NumPy strategy works fine, but requires us to handle multiple processes instead of doing everything in `tf.data`. We could just scrap this entire code path and always use the multiprocessing NumPy approach, but I think single-threaded throughput would be lower if we did that. If you prefer it for code simplicity, though, I can do that.

In the longer term, I'm hoping that `tf.data` gets native support for our data structures and we can transition the whole pipeline to pure `tf.data`, but that still hasn't happened 🫠

Rocketknight1:
And @massquantity TF 2.13 is going to release in a couple of days, so I hope most users are at least on TF 2.9 by now!

lhoestq:
Unless there is a big gap in performance I think code simplicity would be appreciated ^^

Rocketknight1:
This is probably ready, but likely conflicts with #5883. I'll wait for that PR to be merged and then rebase and merge this one.

Rocketknight1:
@alvarobartt @lhoestq This should be ready for re-review. I've rebased it on the recent PR to allow `batch_size=None`, and it should also support unbatched loading now.

Having a variety of different methods like this is annoying, but once our minimum Python version is 3.8 I can go back and clear a lot of this out!

src/datasets/utils/tf_utils.py
```diff
@@ -173,6 +174,21 @@ def dataset_to_tf(
     else:
         raise ImportError("Called a Tensorflow-specific function but Tensorflow is not installed.")
 
+    # TODO Matt: When our minimum Python version is 3.8 or higher, we can delete all of this and move everything
```
alvarobartt, line 177: Hi Matt, is `datasets` going to drop Python 3.7 support due to its upcoming EOL? Because it will happen by the end of the month in case we want to wait and set the minimum version to 3.8, even though I assume some users may still be using 3.7?

lhoestq (approved):
lgtm :) feel free to run some tests before merging though

Rocketknight1:
@alvarobartt I'll probably merge now, just to avoid the major memory usage issues we currently have! Feel free to run the comparisons before/after the commit.

alvarobartt:
> @alvarobartt I'll probably merge now, just to avoid the major memory usage issues we currently have! Feel free to run the comparisons before/after the commit.

I'll ping you back with the comparison this weekend! 🤗

PR #5865, opened 2023-05-15
Deprecate task api

mariosasko:
The task API is not well adopted in the ecosystem, so this PR deprecates it. The `train_eval_index` is a newer, more flexible solution that should be used instead (I think?).

These are the projects that still use the task API :
* the image classification example in Transformers: [here](https://github.com/huggingface/transformers/blob/8f76dc8e5aaad58f2df7748b6d6970376f315a9a/examples/pytorch/image-classification/run_image_classification_no_trainer.py#L262) and [here](https://github.com/huggingface/transformers/blob/8f76dc8e5aaad58f2df7748b6d6970376f315a9a/examples/tensorflow/image-classification/run_image_classification.py#L277)
* autotrain: [here](https://github.com/huggingface/autotrain-backend/blob/455e274004b56f9377d64db4ab03671508fcc4cd/zeus/zeus/run/utils.py#L666)
* api-inference-community: [here](https://github.com/huggingface/api-inference-community/blob/fb8fb29d577a5bf01c82944db745489a6d6ed3d4/manage.py#L64) (but the rest of the code does not call the `resolve_dataset` function)

So we need to update these files after the merge.

cc @lewtun


lhoestq:
If it's easy to keep supporting it we can keep it no ? There are many datasets on the hub that implement the tasks templates in dataset scripts and it's maybe easier to keep task templates than opening PRs to those datasets.

polinaeterna:
do we know if people use the tasks api?

mariosasko:
@lhoestq 

Less than 80 public datasets (all canonical) implement `task_templates`, so updating them should be easy.

PS: I skipped gated datasets when checking for the presence of task templates, but it's safe to assume their contribution to the total count is insignificant.

lhoestq:
Ok ! I also know https://huggingface.co/datasets/hf-internal-testing/cats_vs_dogs_sample/blob/main/cats_vs_dogs_sample.py that needs to be updated as well

docs/source/package_reference/task_templates.mdx
```diff
@@ -1,5 +1,12 @@
 # Task templates
 
+<Tip warning={true}>
+
+The Task API is deprecated due to its inflexibility, and consenquently, the lack of adoption in the ecosystem.
```
lhoestq, line 5: maybe no need to explain why, and rather redirect users to train_eval_index ?

```suggestion
The Task API is deprecated due to its inflexibility, and consenquently, the lack of adoption in the ecosystem.
```

lhoestq (approved):

Issue #5839, opened 2023-05-10
Make models/functions optimized with `torch.compile` hashable


mariosasko:
As reported in https://github.com/huggingface/datasets/issues/5819, hashing functions/transforms that reference a model, or a function, optimized with `torch.compile` currently fails due to them not being picklable (the concrete error can be found in the linked issue).

The solutions to consider:
1. hashing/pickling the original, uncompiled version of a compiled model/function (attributes `_orig_mod`/`_torchdynamo_orig_callable`) (less precise than the 2nd option as it ignores the other params of `torch.compute`)
2. wait for https://github.com/pytorch/pytorch/issues/101107 to be resolved

PR #5867, opened 2023-05-15
Add logic for hashing modules/functions optimized with `torch.compile`

mariosasko:
Fix https://github.com/huggingface/datasets/issues/5839

PS: The `Pickler.save` method is becoming a bit messy, so I plan to refactor the pickler a bit at some point.

### DIFF
--- a/tests/test_fingerprint.py
+++ b/tests/test_fingerprint.py
@@ -328,6 +328,42 @@ class HashingTest(TestCase):
         self.assertEqual(hash1, hash3)
         self.assertNotEqual(hash1, hash2)
 
+    @require_torch
+    def test_hash_torch_compiled_function(self):
+        import torch
+
+        def f(x):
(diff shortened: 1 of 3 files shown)


tests/test_fingerprint.py
```diff
@@ -328,6 +328,42 @@ def test_hash_torch_tensor(self):
...
+            def forward(self, x):
+                x = F.relu(self.conv1(x))
+                return F.relu(self.conv2(x))
+
+        m = Model()
+        m1 = m
+        hash1 = md5(datasets.utils.py_utils.dumps(m1)).hexdigest()
+        m2 = torch.compile(m)
+        hash2 = md5(datasets.utils.py_utils.dumps(m2)).hexdigest()
+        self.assertEqual(hash1, hash2)
```
lhoestq, line 365: Cool ! Maybe add a test that it returns different hashes is the weights are different ?

PR #5883, opened 2023-05-22
Fix string-encoding, make `batch_size` optional, and minor improvements in `Dataset.to_tf_dataset`

alvarobartt:
## What's in this PR?

This PR addresses some minor fixes and general improvements in the `to_tf_dataset` method of `datasets.Dataset`, to convert a 🤗HuggingFace Dataset as a TensorFlow Dataset.

The main bug solved in this PR comes with the string-encoding, since for safety purposes the internal conversion of `numpy.arrays` when `dtype` is unicode/string, is to convert it into `numpy.bytes`, more information in the docstring of https://github.com/tensorflow/tensorflow/blob/388d952114e59a1aeda440ed4737b29f8b7c6e8a/tensorflow/python/ops/script_ops.py#L210. That's triggered when using `tensorflow.numpy_function` as it's applying another type cast besides the one that `datasets` does, so the casting is applied at least twice per entry/batch. So this means that the definition of the `numpy.unicode_` dtype when the data in the batch is a string, is ignored, and replaced by `numpy.bytes_`.

Besides that, some other minor things have been fixed:

* Made `batch_size` an optional parameter in `to_tf_dataset`
* Map the `tensorflow` output dtypes just once, and not in every `tf.function` call during `map`
* Keep `numpy` formatting in the `datasets.Dataset` if already formatted like it, no need to format it again as `numpy`
* Docstring indentation in `dataset_to_tf` and `multiprocess_dataset_to_tf`

## What's missing in this PR?

I can include some integration tests if needed, to validate that `batch_size` is optional, and that the tensors in the TF-Dataset can be looped over with no issues as before.

### DIFF
--- a/src/datasets/utils/tf_utils.py
+++ b/src/datasets/utils/tf_utils.py
@@ -262,7 +262,8 @@ class NumpyMultiprocessingGenerator:
         }
         self.output_signature = output_signature
         self.shuffle = shuffle
-        self.batch_size = batch_size
+        # Make sure that we set the `batch_size=1` if `batch_size=None`
+        self.batch_size = batch_size or 1
         self.drop_remainder = drop_remainder
         self.num_workers = num_workers
         # Because strings are converted to characters, we need to add one extra dimension to the shape
--- a/src/datasets/utils/tf_utils.py
+++ b/src/datasets/utils/tf_utils.py
@@ -136,27 +136,27 @@ def dataset_to_tf(
     """Create a tf.data.Dataset from the underlying Dataset. This is a single-process method - the multiprocess
     equivalent is multiprocess_dataset_to_tf.
 
-            Args:
-                dataset (`Dataset`): Dataset to wrap with tf.data.Dataset.
-                cols_to_retain (`List[str]`): Dataset column(s) to load in the
-                    tf.data.Dataset. It is acceptable to include column names that are created by the `collate_fn` and
-                    that do not exist in the original dataset.
-                collate_fn(`Callable`): A function or callable object (such as a `DataCollator`) that will collate
-                    lists of samples into a batch.
-                collate_fn_args (`Dict`): A  `dict` of keyword arguments to be passed to the
-                    `collate_fn`. Can be empty.
-                columns_to_np_types (`Dict[str, np.dtype]`): A `dict` mapping column names to numpy dtypes.
-                output_signature (`Dict[str, tf.TensorSpec]`): A `dict` mapping column names to
-                    `tf.TensorSpec` objects.
-                shuffle(`bool`): Shuffle the dataset order when loading. Recommended True for training, False for
-                    validation/evaluation.
-                batch_size (`int`, default `None`): Size of batches to load from the dataset. Defaults to `None`, which implies that 
-                    the dataset won't be batched, but the returned dataset can be batched later with `tf_dataset.batch(batch_size)`.
-                drop_remainder(`bool`, default `None`): Drop the last incomplete batch when loading. If not provided,
-                    defaults to the same setting as shuffle.
+    Args:
+        dataset (`Dataset`): Dataset to wrap with tf.data.Dataset.
+        cols_to_retain (`List[str]`): Dataset column(s) to load in the
+            tf.data.Dataset. It is acceptable to include column names that are created by the `collate_fn` and
+            that do not exist in the original dataset.
+        collate_fn(`Callable`): A function or callable object (such as a `DataCollator`) that will collate
+            lists of samples into a batch.
+        collate_fn_args (`Dict`): A  `dict` of keyword arguments to be passed to the
+            `collate_fn`. Can be empty.
+        columns_to_np_types (`Dict[str, np.dtype]`): A `dict` mapping column names to numpy dtypes.
+        output_signature (`Dict[str, tf.TensorSpec]`): A `dict` mapping column names to
+            `tf.TensorSpec` objects.
+        shuffle(`bool`): Shuffle the dataset order when loading. Recommended True for training, False for
+            validation/evaluation.
+        batch_size (`int`, default `None`): Size of batches to load from the dataset. Defaults to `None`, which implies that 
+            the dataset won't be batched, but the returned dataset can be batched later with `tf_dataset.batch(batch_size)`.
+        drop_remainder(`bool`, default `None`): Drop the last incomplete batch when loading. If not provided,
+            defaults to the same setting as shuffle.
 
-            Returns:
-                `tf.data.Dataset`
+    Returns:
+        `tf.data.Dataset`
     """
     if config.TF_AVAILABLE:
         import tensorflow as tf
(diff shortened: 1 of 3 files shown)


alvarobartt:
To showcase the current issue, here's a Colab Gist, that shows that the `imdb` dataset cannot be read/iterated, since one or more samples contain a non-ascii character that is being converted to `numpy.bytes`, and so on fails.

Colab Gist at 

Also, here's a quick sample of what's happening:

```python

ds = load_dataset("imdb", split="train")
tfds = ds.to_tf_dataset(batch_size=16)
for batch in tfds:
  print(batch)
>>> UnicodeEncodeError: 'ascii' codec can't encode character '\xe9' in position 0: ordinal not in range(128)
```

A more detailed version of it:

```python
from datasets import Dataset

ds = Dataset.from_dict(
    {
        "a": [1],
        "b": ["é"],
    }
)
tfds = ds.to_tf_dataset(batch_size=1)
for batch in tfds:
    print(batch)
>>> UnicodeEncodeError: 'ascii' codec can't encode character '\xe9' in position 0: ordinal not in range(128)
```

The original issue comes from https://github.com/tensorflow/tensorflow/blob/388d952114e59a1aeda440ed4737b29f8b7c6e8a/tensorflow/python/ops/script_ops.py#LL234C4-L234C4, which could easily be solved by replacing that line with `return result.astype(np.unicode_)` but they are mentioning that it may lead to issues.

Even the following fails in `numpy`:

```python
import numpy as np

x = np.array(["é"]).astype(np.bytes_)
```
```

**Comment by alvarobartt** [comment]
cc. @lhoestq :hugs:

**Comment by lhoestq** [comment]
cc @Rocketknight1

**Review by lhoestq — COMMENTED** [review]
Nice ! Could you add some tests to make sure that batch_size=None works as expected ?

**Comment by alvarobartt** [comment]
> Nice ! Could you add some tests to make sure that batch_size=None works as expected ?

Sure, I'll add the tests for everything, including the string-encoding issue to make sure it's solved!

File: src/datasets/utils/tf_utils.py [hunk 1]
```diff
@@ -166,16 +173,18 @@ def dataset_to_tf(
...
+        return_dict=False,
     )
 
+    # This works because dictionaries always output in the same order
+    tout = [tf.dtypes.as_dtype(dtype) for dtype in columns_to_np_types.values()]
+
     @tf.function(input_signature=[tf.TensorSpec(None, tf.int64)])
     def fetch_function(indices):
-        output = tf.numpy_function(
+        output = tf.py_function(
```
Rocketknight1, line 184: The change from `numpy_function` to `py_function` here should be okay, but is it necessary? I understand you want to be able to pass a single index instead of a batch of indices, but I feel like you could just pass a Numpy scalar in that case (with shape == ()) and leave the rest of the code the same.

src/datasets/utils/tf_utils.py
```diff
@@ -285,7 +297,7 @@ def __iter__(self):
         with SharedMemoryContext() as shm_ctx:
             for i in range(num_workers):
                 worker_random_id = str(uuid4())
-                worker_name = f"datasets_tf_worker_{i}_{worker_random_id}"
+                worker_name = f"dw_{i}_{worker_random_id}"[:10]
```
Rocketknight1, line 300: This is probably cleaner than the whole UUID, I agree

Rocketknight1 (approved):
Overall it looks good, with one comment!

Also, bear in mind that performance may be negatively impacted if we don't load samples in a batch - it's generally more efficient to grab multiple indices from the underlying arrow dataset at once than to do them sequentially, especially if the indices are contiguous (which they are during eval). Still, I think it's good to offer the option when users want it!

src/datasets/utils/tf_utils.py
alvarobartt: I was eventually running into `OSError: [Errno 36] File name too long: ...` this is why I just stripped it to keep the first 10 characters

src/datasets/utils/tf_utils.py
alvarobartt: More information at https://github.com/tensorflow/tensorflow/blob/388d952114e59a1aeda440ed4737b29f8b7c6e8a/tensorflow/python/ops/script_ops.py#L210

Rocketknight1: Ah, that makes sense! Okay, I'm fine with this change.

alvarobartt:
@Rocketknight1 then I'll add the integration tests for the optional `batch_size` as well as for the encoding of non-ASCII compatible characters 😄 Do we set the default `batch_size` to 16 instead of `None`?

lhoestq:
I think default should be None, not 16.
Users won't expect to have it batched by default.

alvarobartt:
Then I'll leave it as is, and add the unit/integration tests, thanks @Rocketknight1 and @lhoestq

alvarobartt:
Hi @Rocketknight1 @lhoestq! So the string-encoding issue is already solved, but I've got one doubt about the `batch_size` being optional in the multiprocessing approach, since in that case I assume the `batch_size` should be mandatory, for the moment I'm assuming it is/should be mandatory, but let me know if you want me to add a check to disallow `batch_size=None` when `num_workers>1`. Thanks!

alvarobartt:
> To showcase the current issue, here's a Colab Gist, that shows that the `imdb` dataset cannot be read/iterated, since one or more samples contain a non-ascii character that is being converted to `numpy.bytes_`, and so on fails.
> 
> Colab Gist at https://gist.github.com/alvarobartt/1746959d1abb9a33e0c593f3bd82a2fb

I've used the Colab shared above for testing purposes, and it works fine, plus the unit/integration tests are passing. I've also trained a `KerasNLP` model with incoming data from 🤗`datasets` with no issue at all!

lhoestq:
> in the multiprocessing approach, since in that case I assume the batch_size should be mandatory,

No I think they're quite orthogonal, no need to have it mandatory

alvarobartt:
> No I think they're quite orthogonal, no need to have it mandatory

But it will break if `batch_size=None` as the multiprocessing approach will aim to prepare batches and distribute those to every worker, and assuming `batch_size=1` when `batch_size=None` I guess is not a good assumption, right?

lhoestq:
Ah I see. Multiprocessing should support batch_size=None indeed. If you have ideas you can do it in this PR, or raise a NotImplementedError and we can see later

alvarobartt:
Sure @lhoestq , I can add a `NotImplementedError` for the moment, and prepare the next PR straight-away to tackle the multiprocessing approach with `batch_size=None`, but not sure if that may eventually collide with @Rocketknight1 PR at https://github.com/huggingface/datasets/pull/5863

Rocketknight1:
Yes, let me merge the PR at #5863 after this one, and then we can open another to improve the behaviour with multiprocessing and `batch_size=None`!

Rocketknight1:
`raise NotImplementedError` for now with an error telling the user that multiprocessing needs them to specify a batch size, I think!

alvarobartt:
Ready to merge @Rocketknight1! 🤗

alvarobartt:
Since you recently approved @Rocketknight1, are we ready to merge? Thanks 🤗

src/datasets/utils/tf_utils.py
```diff
@@ -98,10 +107,12 @@ def np_get_batch(
...
 
-    actual_size = len(list(batch.values())[0])  # Get the length of one of the arrays, assume all same
-    # Our collators expect a list of dicts, not a dict of lists/arrays, so we invert
-    batch = [{key: value[i] for key, value in batch.items()} for i in range(actual_size)]
-    batch = collate_fn(batch, **collate_fn_args)
+    if is_batched:
+        actual_size = len(list(batch.values())[0])  # Get the length of one of the arrays, assume all same
+        # Our collators expect a list of dicts, not a dict of lists/arrays, so we invert
+        batch = [{key: value[i] for key, value in batch.items()} for i in range(actual_size)]
+        batch = collate_fn(batch, **collate_fn_args)
```
lhoestq, line 114: We still need to call collate_fn if not `batched`
```suggestion
    batch = collate_fn(batch, **collate_fn_args)
```

lhoestq (approved):
Looks all good ! I just added a minor bug fix

alvarobartt:
Hi @lhoestq so I didn't include the call to `collate_fn`, as we won't need to collate the incoming data e.g. "str" should remain a "str" not a ["str"], and the `minimal_collate_fn` was indeed putting everything into a list, so the output was not un-batched, but batched with size 1

lhoestq:
What if the user passes a collate_fn ? The torch DataLoader still applies it if batch_size=None for example.

Does my last change look of to you ? If so I think we can merge

alvarobartt:
> What if the user passes a collate_fn ? The torch DataLoader still applies it if batch_size=None for example.
> 
> Does my last change look of to you ? If so I think we can merge

I think we're good, since it won't batch it under the scenario of `str` being provided instead of `List[str]`, and the unit/integration tests are passing, so I'm OK to merge. Maybe we can double check with Matt? cc @Rocketknight1

alvarobartt:
Will you eventually need help with your PR @Rocketknight1? I'll be happy to help if needed 😄

Rocketknight1:
@alvarobartt Yes, I'll ping you for a review once it's ready!

PR #5885, opened 2023-05-23
Modify `is_remote_filesystem` to return True for FUSE-mounted paths

maddiedawson:


src/datasets/filesystems/__init__.py
```diff
@@ -49,10 +71,11 @@ def is_remote_filesystem(fs: fsspec.AbstractFileSystem) -> bool:
         fs (`fsspec.spec.AbstractFileSystem`):
             An abstract super-class for pythonic file-systems, e.g. `fsspec.filesystem(\'file\')` or [`datasets.filesystems.S3FileSystem`].
     """
-    if fs is not None and fs.protocol != "file":
-        return True
-    else:
+    if fs is None:
         return False
+    if isinstance(fs, FuseFileSystem) or fs.protocol != "file":
```
es94129, line 76: Is it possible for a FUSE-mounted file system to not be a remote file system?

maddiedawson:
@lhoestq would you or another maintainer be able to review please? :)

lhoestq:
Why you do need to support FUSE mounted paths ?

`datasets` uses data that live on disk for fast lookups - FUSE mounted disks would lead to poor performance and I wouldn't recomment using it.

maddiedawson:
Fuse is commonly used to mount remote file systems (e.g. S3, DBFS) as a local directory. Since it's slower than using an actual local device, it's better to treat it as remote to reduce latency.

es94129 (approved):

lhoestq:
I think people would be confused if they don't have the same dataset behavior depending on the disk type.

If they want to use a remote bucket they should use the remote URI instead, e.g. `s3://...`. Advancements on this are tracked at #5281

Issue #5481, opened 2023-01-27
Load a cached dataset as iterable


lhoestq:
The idea would be to allow something like 

```python 
ds = load_dataset("c4", "en", as_iterable=True)
```

To be used to train models. It would load an IterableDataset from the cached Arrow files.

Cc @stas00

jalajk24:
Can I work on this issue? I am pretty new to this.

lhoestq:
Hi ! Sure :) you can comment `#self-assign` to assign yourself to this issue.

I can give you some pointers to get started:

`load_dataset` works roughly this way:
1. it instantiate a dataset builder using `load_dataset_builder()`
2. the builder download and prepare the dataset as Arrow files in the cache using `download_and_prepare()`
3. the builder returns a Dataset object with `as_dataset()`

One way to approach this would be to implement `as_iterable_dataset()` in `builder.py`.

And similarly to `as_dataset()`, you can use the `ArrowReader`. It has a `get_file_instructions()` method that can be helpful. It gives you the files to read as list of dictionaries with those keys: `filename`, `skip` and `take`.

The `skip` and `take` arguments are used in case the user wants to load a subset of the dataset, e.g.
```python
load_dataset(..., split="train[:10]")
```

Let me know if you have questions or if I can help :)

mariosasko:
This use-case is a bit specific, and `load_dataset` already has enough parameters (plus, `streaming=True` also returns an iterable dataset, so we would have to explain the difference), so I think it would be better to add `IterableDataset.from_file` to the API (more flexible and aligned with the goal from https://github.com/huggingface/datasets/issues/3444) instead.

lhoestq:
> This use-case is a bit specific

This allows to use `datasets` for large scale training where map-style datasets are too slow and use too much memory in PyTorch. So I would still consider adding it.

Alternatively we could add this feature one level bellow:
```python
builder = load_dataset_builder(...)
builder.download_and_prepare()
ids = builder.as_iterable_dataset()
```

mariosasko:
Yes, I see how this can be useful. Still, I think `Dataset.to_iterable` + `IterableDataset.from_file` would be much cleaner in terms of the API design (and more flexible since `load_dataset` can only access the "initial" (unprocessed) version of a dataset).

And since it can be tricky to manually find the "initial" version of a dataset in the cache, maybe `load_dataset` could return an iterable dataset streamed from the cache if `streaming=True` and the cache is up-to-date.

[...]

PR #5893, opened 2023-05-23
Load cached dataset as iterable 

mariusz-jachimowicz-83:
To be used to train models it allows to load an IterableDataset from the cached Arrow file.  
See https://github.com/huggingface/datasets/issues/5481

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -12,14 +12,13 @@ import pyarrow as pa
 
 from . import config
 from .arrow_dataset import Dataset, DatasetInfoMixin
-from .arrow_reader import ArrowReader
 from .features import Features
 from .features.features import FeatureType, _align_features, _check_if_features_can_be_aligned
 from .filesystems import _reset_fsspec_lock
 from .formatting import PythonFormatter, get_format_type_from_alias
 from .info import DatasetInfo
 from .splits import NamedSplit
-from .table import cast_table_to_features, table_cast
+from .table import arrow_table_schema_from_file, cast_table_to_features, table_cast
 from .utils.logging import get_logger
 from .utils.py_utils import Literal
 from .utils.sharding import _merge_gen_kwargs, _number_of_shards_in_gen_kwargs, _shuffle_gen_kwargs, _split_gen_kwargs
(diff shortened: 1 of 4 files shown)


mariusz-jachimowicz-83:
@lhoestq Could you please look into that and review?

src/datasets/iterable_dataset.py
```diff
@@ -1115,6 +1116,24 @@ def from_spark(
...
+        Args:
+            filename (`str`):
+                File name of the dataset.
+            in_memory (`bool`, defaults to `False`):
+                Whether to copy the data in-memory.
+
+        Returns:
+            [`IterableDataset`]
+        """
+        pa_table = ArrowReader.read_table(filename, in_memory=in_memory)
```
lhoestq, line 1132: This would read the entire file metadata, which can take some time, in particular if you have thousands of files (which is going to be the case for people interested in https://github.com/huggingface/datasets/issues/5481)

I think it should be possible to have an arrow table generator that yields tables from the file without reading all its metadata.

Maybe using `pyarrow.ipc.RecordBatchStreamReader` and then `read_next_batch` ?

lhoestq (commented):
Awesome ! I just added a comment to make it even faster for super big datasets:

mariusz-jachimowicz-83:
@lhoestq I refactored the code. Could you please check is it what you requested?

src/datasets/table.py
```diff
@@ -46,13 +46,37 @@ def _in_memory_arrow_table_from_buffer(buffer: pa.Buffer) -> pa.Table:
...
 def _memory_mapped_arrow_table_from_file(filename: str) -> pa.Table:
-    memory_mapped_stream = pa.memory_map(filename)
-    opened_stream = pa.ipc.open_stream(memory_mapped_stream)
-    pa_table = opened_stream.read_all()
+    with _memory_mapped_arrow_stream_from_file(filename) as memory_mapped_stream:
+        pa_table = pa.ipc.open_stream(memory_mapped_stream).read_all()
     return pa_table
 
 
+def arrow_table_batches_from_file(filename: str, max_chunksize: Optional[int] = None):
```
lhoestq, line 75: for consistency with the other loading functions, you can name it _memory_mapped_record_batch_reader_from_file

src/datasets/table.py
```diff
@@ -46,13 +46,37 @@ def _in_memory_arrow_table_from_buffer(buffer: pa.Buffer) -> pa.Table:
     return table
 
 
+def _memory_mapped_arrow_stream_from_file(filename: str):
```
lhoestq, line 49: I think this function is not necessary, we can call pa.memory map directly

src/datasets/table.py
```diff
@@ -46,13 +46,37 @@ def _in_memory_arrow_table_from_buffer(buffer: pa.Buffer) -> pa.Table:
...
+        schema = pa.ipc.open_stream(memory_mapped_stream).schema
+    return schema
+
+
 def _memory_mapped_arrow_table_from_file(filename: str) -> pa.Table:
-    memory_mapped_stream = pa.memory_map(filename)
-    opened_stream = pa.ipc.open_stream(memory_mapped_stream)
-    pa_table = opened_stream.read_all()
+    with _memory_mapped_arrow_stream_from_file(filename) as memory_mapped_stream:
+        pa_table = pa.ipc.open_stream(memory_mapped_stream).read_all()
```
lhoestq, line 71: I think you can remove the `with` statement here, since the memory map needs to exist for the returned pa_table to be used

lhoestq (commented):
Thanks ! I added my final comments :)

src/datasets/table.py
mariusz-jachimowicz-83: Tip applied.

src/datasets/table.py
mariusz-jachimowicz-83: Tip applied

mariusz-jachimowicz-83:
@lhoestq  Thanks for a review. Excellent tips. All tips applied.

lhoestq:
I think there is just PythonFormatter that needs to be imported in the test file and we should be good to merge

mariusz-jachimowicz-83:
@lhoestq that is weird. I have linter error when I do it.

mariusz-jachimowicz-83:
@lhoestq Now it should work properly.

lhoestq (approved):
It looks all good ! merging :)

PR #5894, opened 2023-05-24
Force overwrite existing filesystem protocol

### LINKED ISSUE #5876: Incompatibility with DataLab
### Describe the bug

Hello,
I am currently working on a project where both [DataLab](https://github.com/ExpressAI/DataLab) and [datasets](https://github.com/huggingface/datasets) are subdependencies.
I noticed that I cannot import both libraries, as they both register FileSystems in `fsspec`, expecting the FileSystems not being registered before.

When running the code below, I get the following error:
```
  File "C:\Users\Bened\anaconda3\envs\ner-eval-dashboard2\lib\site-packages\datalabs\__init__.py", line 28, in <module>
    from datalabs.arrow_dataset import concatenate_datasets, Dataset
  File "C:\Users\Bened\anaconda3\envs\ner-eval-dashboard2\lib\site-packages\datalabs\arrow_dataset.py", line 60, in <module>
    from datalabs.arrow_writer import ArrowWriter, OptimizedTypedSequence
  File "C:\Users\Bened\anaconda3\envs\ner-eval-dashboard2\lib\site-packages\datalabs\arrow_writer.py", line 28, in <module>
    from datalabs.features import (
  File "C:\Users\Bened\anaconda3\envs\ner-eval-dashboard2\lib\site-packages\datalabs\features\__init__.py", line 2, in <module>
    from datalabs.features.audio import Audio
  File "C:\Users\Bened\anaconda3\envs\ner-eval-dashboard2\lib\site-packages\datalabs\features\audio.py", line 21, in <module>
    from datalabs.utils.streaming_download_manager import xopen
  File "C:\Users\Bened\anaconda3\envs\ner-eval-dashboard2\lib\site-packages\datalabs\utils\streaming_download_manager.py", line 16, in <module>
    from datalabs.filesystems import COMPRESSION_FILESYSTEMS
  File "C:\Users\Bened\anaconda3\envs\ner-eval-dashboard2\lib\site-packages\datalabs\filesystems\__init__.py", line 37, in <module>
    fsspec.register_implementation(fs_class.protocol, fs_class)
  File "C:\Users\Bened\anaconda3\envs\ner-eval-dashboard2\lib\site-packages\fsspec\registry.py", line 51, in register_implementation
    raise ValueError(
ValueError: Name (bz2) already in the registry and clobber is False
```

I think as simple solution would be to just set `clobber=True` in https://github.com/huggingface/datasets/blob/main/src/datasets/filesystems/__init__.py#L28. This allows the register to discard previous registrations. This should work, as the datalabs FileSystems are copies of the datasets FileSystems. However, I don't know if it is guaranteed to be compatible with other libraries that might use the same protocols.

I am linking the symmetric issue on [DataLab](https://github.com/ExpressAI/DataLab/issues/425) as ideally the issue is solved in both libraries the same way. Otherwise, it could lead to different behaviors depending on which library gets imported first.

1. Run `pip install datalabs==0.4.15 datasets==2.12.0`
2. Run the following python code:
   ```
      import datalabs
      import datasets
   ```

### Expected behavior

It should be possible to import both libraries without getting a Value Error

### Environment info

datalabs==0.4.15
datasets==2.12.0

baskrahmer:
Fix #5876

### DIFF
--- a/src/datasets/filesystems/__init__.py
+++ b/src/datasets/filesystems/__init__.py
@@ -1,8 +1,8 @@
 import importlib
 import shutil
 import threading
-from typing import List
 import warnings
+from typing import List
 
 import fsspec
 import fsspec.asyn
--- a/src/datasets/filesystems/__init__.py
+++ b/src/datasets/filesystems/__init__.py
@@ -2,6 +2,7 @@ import importlib
 import shutil
 import threading
 from typing import List
+import warnings
 
 import fsspec
 import fsspec.asyn
@@ -25,7 +26,9 @@ COMPRESSION_FILESYSTEMS: List[compression.BaseCompressedFileFileSystem] = [
 
 # Register custom filesystems
 for fs_class in COMPRESSION_FILESYSTEMS + [HfFileSystem]:
-    fsspec.register_implementation(fs_class.protocol, fs_class)
+    if fs_class.protocol in fsspec.registry and fsspec.registry[fs_class.protocol] is not fs_class:
+        warnings.warn(f"A filesystem protocol was already set for {fs_class.protocol} and will be overwritten.")
+    fsspec.register_implementation(fs_class.protocol, fs_class, clobber=True)
 
 
 def extract_path_from_uri(dataset_path: str) -> str:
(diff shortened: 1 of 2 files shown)


albertvillanova (changes requested):
Thanks for the fix, @baskrahmer.

In order to fix the quality code issue, could you please run
```
make style
```

albertvillanova (approved):
The tests are OK now. Thank you!

PR #5897, opened 2023-05-25
Fix `FixedSizeListArray` casting

### LINKED ISSUE #5866: Issue with Sequence features
### Describe the bug

Sequences features  sometimes causes errors when the specified length is not -1

```python
import numpy as np
from datasets import Features, ClassLabel, Sequence, Value, Dataset
feats = Features(**{'target': ClassLabel(names=[0, 1]),'x': Sequence(feature=Value(dtype='float64',id=None), length=2, id=None)})
Dataset.from_dict({"target": np.ones(2000).astype(int), "x": np.random.rand(2000,2)},features = feats).flatten_indices()
```
Throws:
```
  fixed_size_list<item: double>[2]
  to
  Sequence(feature=Value(dtype='float64', id=None), length=2, id=None)
```
The same code works without any issues when `length = -1`

EDIT: The error seems to happen only when the length of the dataset is bigger than 1000 for some reason
### Expected behavior

No exception

### Environment info

- `datasets` version: 2.10.1
- Python version: 3.9.5
- PyArrow version: 11.0.0
- Pandas version: 1.4.1

mariosasko:
Fix cast on sliced `FixedSizeListArray`s.

Fix #5866

### DIFF
--- a/src/datasets/table.py
+++ b/src/datasets/table.py
@@ -1903,6 +1903,7 @@ def array_concat(arrays: List[pa.Array]):
             )
         elif pa.types.is_fixed_size_list(array_type):
             if config.PYARROW_VERSION.major < 13:
+                # PyArrow bug: https://github.com/apache/arrow/issues/35360
                 return pa.FixedSizeListArray.from_arrays(
                     _concat_arrays([array.values[array.offset * array.type.list_size :] for array in arrays]),
                     array_type.list_size,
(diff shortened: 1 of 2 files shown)


src/datasets/table.py
```diff
@@ -1902,10 +1902,18 @@ def _concat_arrays(arrays):
                 _concat_arrays([array.values for array in arrays]),
             )
         elif pa.types.is_fixed_size_list(array_type):
-            return pa.FixedSizeListArray.from_arrays(
-                _concat_arrays([array.values for array in arrays]),
-                array_type.list_size,
-            )
+            if config.PYARROW_VERSION.major < 13:
+                # PyArrow bug: https://github.com/apache/arrow/issues/35360
```
lhoestq, line 1906: how does it relate to this issue about hashing ?

mariosasko: [Here](https://github.com/apache/arrow/issues/35573#issuecomment-1547224289) it's mentioned that fixing this issue should also solve the issue with sliced fixed-size list arrays

lhoestq: ah I see !

lhoestq (approved):
This one was not obvious to fix Oo good catch !

PR #5902, opened 2023-05-26
Align `bert-base-cased` usage, install missing `seqeval`, and re-run `Overview.ipynb`

alvarobartt:
## What's in this PR?

This PR solves #5887 since there was a mismatch between the tokenizer and the model used, since the tokenizer was `bert-base-cased` while the model was `distilbert-base-case` both for the PyTorch and TensorFlow alternatives. Since DistilBERT doesn't use/need the `token_type_ids`, the `**batch` was failing, as the batch contained `input_ids`, `attention_mask`, `token_type_ids`, `start_positions` and `end_positions`, and `token_type_ids` was not required.

Besides that, at the end `seqeval` was being used to evaluate the model predictions, and just `evaluate` was being installed, so I've also included the `seqeval` installation.

Finally, I've re-run everything in Google Colab, and every cell was successfully executed!

PR #5903, opened 2023-05-26
Relax `ci.yml` trigger for `pull_request` based on modified paths

alvarobartt:
## What's in this PR?

As of a previous PR at #5902, I've seen that the CI was automatically trigger on any file, in that case when modifying a Jupyter Notebook (.ipynb), which IMO could be skipped, as the modification on the Jupyter Notebook has no effect/impact on the `ci.yml` outcome. So this PR controls the paths that trigger the `ci.yml` to avoid wasting resources when not needed.

## What's pending in this PR?

I would like to confirm whether this should affect both `push` and `pull_request`, since just modifications in those files won't change the `ci.yml` outcome, so maybe it's worth skipping it too in the `push` trigger.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,11 @@ on:
   pull_request:
     branches:
       - main
+    paths-ignore:
+      - "**.md"
+      - "**.mdx"
+      - "**.cff"
+      - "**.ipynb"
   push:
     branches:
       - main


alvarobartt:
Also this could be extended to the rest of the GitHub Action `yml` files, so let me know whether you want me to have a look into it! 🤗

mariosasko:
Maybe we can add
```python
paths-ignore:
  - "docs/**"
```
to `ci.yml` and `benchmarks.yml`. The other supporting files are not modified often, so leaving them out is fine.

PR #5909, opened 2023-05-27
Use more efficient and idiomatic way to construct list.

ttsugriy:
Using `*` is ~2X faster according to [benchmark](https://colab.research.google.com/gist/ttsugriy/c964a2604edf70c41911b10335729b6a/for-vs-mult.ipynb) with just 4 patterns. This doesn't matter much since this tiny difference is not going to be noticeable, but why not?


lhoestq (approved):
Thanks, as you said: why not :p

mariosasko:
It's faster because all the items are the same object, but this also means modifying one of them will alter each unless these items are immutable, and they are in this case (tuples). So we should be careful when using this idiom.

PR #5921, opened 2023-06-01
Fix streaming parquet with image feature in schema

lhoestq:
It was not reading the feature type from the parquet arrow schema


albertvillanova (approved):
Good catch! Thanks.

PR #5924, opened 2023-06-02
Add parallel module using joblib for Spark

es94129:
Discussion in https://github.com/huggingface/datasets/issues/5798

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -160,6 +160,7 @@ BENCHMARKS_REQUIRE = [
 TESTS_REQUIRE = [
     # test dependencies
     "absl-py",
+    "joblibspark",
     "pytest",
     "pytest-datadir",
     "pytest-xdist",
--- a/src/datasets/parallel/parallel.py
+++ b/src/datasets/parallel/parallel.py
@@ -1,6 +1,6 @@
 from multiprocessing import Pool, RLock
 from tqdm.auto import tqdm
-from typing import List, Optional
+from typing import List
 import contextlib
 import joblib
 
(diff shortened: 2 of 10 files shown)


es94129:
Hi @lhoestq, I added the `parallel` part according to the discussion we had. Could you take a look to see if this is aligned with your proposal?

Meanwhile I'm working on adding a `parallel_backend` parameter to `load_datasets` so that it can be used like:
```python
with parallel_backend('spark', steps=['downloading']) as backend:
  ds = load_dataset(..., parallel_backend=backend)
```
where `parallel_backend` is a `ParallelBackend` class.

src/datasets/parallel/parallel.py
```diff
@@ -0,0 +1,86 @@
...
+        with joblib.parallel_backend(self.backend_name, n_jobs=-1):
+            return joblib.Parallel()(
+                joblib.delayed(single_map_nested_func)((function, obj, types, None, True, None)) for obj in iterable
+            )
+
+
+@contextlib.contextmanager
+def parallel_backend(backend_name: str, steps: List[str]):
+    if backend_name not in ["spark", "multiprocessing"]:
+        raise ValueError(f"Unsupported parallel backend: {backend_name}")
```
lhoestq, line 69: I wouldn't exclude other backend names - depending on what joblib backend the user may have

src/datasets/parallel/parallel.py
```diff
@@ -0,0 +1,86 @@
...
+    def _map_with_joblib(self, function, iterable, types, single_map_nested_func):
+        # TODO: take num_proc, tqdm args
+        with joblib.parallel_backend(self.backend_name, n_jobs=-1):
+            return joblib.Parallel()(
+                joblib.delayed(single_map_nested_func)((function, obj, types, None, True, None)) for obj in iterable
+            )
+
+
+@contextlib.contextmanager
+def parallel_backend(backend_name: str, steps: List[str]):
```
lhoestq, line 67: In order to not have to pass parallel_backend to every function in `datasets` and for consistency with scikit-learn which uses a similar API, I think this can sets the backend globally.

And this way we don't need to change anything in the core of `datasets`, except specifying the steps names in the correct locations (eg a decorator on DownloadManager.download, and later on Dataset.map)

src/datasets/parallel/parallel.py
es94129: Got it, I've edited it now.
Now if the user passes an unsupported backend, `_map_with_joblib` will show
```text
ValueError: Invalid backend: somebackend, expected one of ['loky', 'multiprocessing', 'sequential', 'spark', 'threading']
```

es94129:
@lhoestq Thanks for the comments!
With your suggestion, no changes made to `load_dataset` and I validated that downloading with spark is working now this:
```py
with parallel_backend('spark', steps=["download"]):
      dataset = load_dataset(..., num_proc=2)
```

src/datasets/parallel/parallel.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+class ParallelBackendConfig:
+    backend_name = None
+    steps = []
+
+
+def parallel_map(function, iterable, num_proc, types, disable_tqdm, desc, single_map_nested_func):
+    """
+    Apply a function to iterable elements in parallel, where the implementation uses either multiprocessing.Pool or
```
lhoestq, line 19: Could you mark this function as experimental in the docstring ? This way we can still modify it if we want
```suggestion
    **Experimental.** Apply a function to iterable elements in parallel, where the implementation uses either multiprocessing.Pool or
```

src/datasets/parallel/parallel.py
```diff
@@ -0,0 +1,103 @@
...
+    with joblib.parallel_backend(ParallelBackendConfig.backend_name, n_jobs=num_proc):
+        return joblib.Parallel()(
+            joblib.delayed(single_map_nested_func)((function, obj, types, None, True, None)) for obj in iterable
+        )
+
+
+@contextlib.contextmanager
+def parallel_backend(backend_name: str, steps: List[str]):
+    """
+    Configures the parallel backend for parallelized dataset loading, steps including download and prepare.
```
lhoestq, line 75: same here. And maybe mention that it uses joblib under the hood.
```suggestion
    **Experimental.** Configures the joblib parallel backend for parallelized dataset loading, steps including download and prepare.
```

tests/test_parallel.py
```diff
@@ -0,0 +1,51 @@
...
+from .utils import require_joblibspark
+
+
+def add_one(i):  # picklable for multiprocessing
+    return i + 1
+
+
+@require_joblibspark
+def test_parallel_backend_input():
+    with parallel_backend("spark", steps=["downloading"]):
```
lhoestq, line 15: I think you can replace "downloading" by "download". Adding -ing at the end of every step name is not necessary imo :)

lhoestq (commented):
Love it ! :)

Note that right now it would use Spark for all the map_nested calls when the `parallel_backend` context is used. I think we can use Spark for the map_nested calls related to downloads

One idea is to decorate the `download` method to set the current global step to "download", and then only use joblib if the current step is one of the steps provided in `parallel_backend `. Though this can be done in a subsequent PR - lmk what you think

and some minor comments:

src/datasets/parallel/parallel.py
mariosasko: Maybe we could decorate these functions with a decorator similar to this one: https://github.com/huggingface/huggingface_hub/blob/0410930a5b5e24fa33da3a0e9d14c96c4f6e2b44/src/huggingface_hub/utils/_experimental.py#L15

src/datasets/parallel/parallel.py
```diff
@@ -0,0 +1,103 @@
...
+    # and it requires monkey-patching joblib internal classes which is subject to change
+
+    with joblib.parallel_backend(ParallelBackendConfig.backend_name, n_jobs=num_proc):
+        return joblib.Parallel()(
+            joblib.delayed(single_map_nested_func)((function, obj, types, None, True, None)) for obj in iterable
+        )
+
+
+@contextlib.contextmanager
+def parallel_backend(backend_name: str, steps: List[str]):
```
mariosasko, line 73: Not a fan of the `steps` parameter and "limiting" the context manager's usage to dataset loading, e.g., changing the backend for `Dataset.map`, which can be implemented later, also seems useful. So it's best to remove this parameter to keep things simple IMO.

lhoestq: The idea is to add support for `map` later. I thought of the "steps" argument to force users to be explicit for now until it's implemented for `map`.

But maybe we can simply remove this argument and raise NotImplementedError if the backend is set to spark and a parallel `map` is run ?

src/datasets/parallel/parallel.py
lhoestq: ah good idea !

src/datasets/parallel/parallel.py
es94129: Yes I think removing it and raising `NotImplementedError` makes sense.

tests/test_parallel.py
maddiedawson: +1

tests/test_parallel.py
```diff
@@ -0,0 +1,51 @@
...
+from datasets.utils.py_utils import map_nested
+
+from .utils import require_joblibspark
+
+
+def add_one(i):  # picklable for multiprocessing
+    return i + 1
+
+
+@require_joblibspark
```
maddiedawson, line 13: You may also need to require_not_windows etc

src/datasets/parallel/parallel.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class ParallelBackendConfig:
+    backend_name = None
+    steps = []
+
+
+def parallel_map(function, iterable, num_proc, types, disable_tqdm, desc, single_map_nested_func):
+    """
+    Apply a function to iterable elements in parallel, where the implementation uses either multiprocessing.Pool or
```
maddiedawson, line 21: Could we include in this comment the expected args for single_map_nested_func?

setup.py
```diff
@@ -160,6 +160,7 @@
 TESTS_REQUIRE = [
     # test dependencies
     "absl-py",
+    "joblibspark",
```
lu-wang-dl, line 163: joblib or joblibspark

es94129: `joblib` is a requirement of `joblibspark`, while both would be used, specifying `joblibspark` should be fine.

es94129:
@lhoestq Can a maintainer help trigger the tests again?
> One idea is to decorate the download method to set the current global step to "download", and then only use joblib if the current step is one of the steps provided in parallel_backend.

Yes I think this is doable in a subsequent PR.
For throwing `NotImplementedError` I also think it can be done in a subsequent PR, because I'm not sure if `Dataset.map` is the only function that a user would expect to run using `with parallel_backend`.

lhoestq:
Just triggered the tests :)

> Yes I think this is doable in a subsequent PR.
For throwing NotImplementedError I also think it can be done in a subsequent PR, because I'm not sure if Dataset.map is the only function that a user would expect to run using with parallel_backend.

I think any Dataset method that has a `num_proc` argument: Dataset.map (the other methods like filter or cast or based on map), and later we can see for the to_xxx methods (to_csv, to_parquet, etc.)

es94129:
Hi maintainers, I've just addressed most of the comments, please take another look, thank you.

src/datasets/parallel/parallel.py
```diff
@@ -0,0 +1,113 @@
...
+        start = div * index + min(index, mod)
+        end = start + div + (1 if index < mod else 0)
+        split_kwds.append((function, iterable[start:end], types, index, disable_tqdm, desc))
+
+    if len(iterable) != sum(len(i[1]) for i in split_kwds):
+        raise ValueError(
+            f"Error dividing inputs iterable among processes. "
+            f"Total number of objects {len(iterable)}, "
+            f"length: {sum(len(i[1]) for i in split_kwds)}"
+        )
```
lhoestq, line 57: Maybe this can still be in map_nested, so that the signature of parallel_map could be 
```python
def parallel_map(function, iterable, num_proc):
```

and `map_nested` would call
```python
parallel_map(_single_map_nested, split_kwds)
```

This way it will be easier to start using `parallel_map` in other places in the code no ?

lhoestq: `n_jobs` is specified to joblib anyway no ? not a big deal anyway

src/datasets/parallel/parallel.py
```diff
@@ -0,0 +1,113 @@
+import contextlib
+from multiprocessing import Pool, RLock
+
+import joblib
```
lhoestq, line 4: The import should be moved to `_map_with_joblib`, since joblib is an optional dependency

src/datasets/parallel/parallel.py
es94129: Might leave it like this so that `n_jobs=-1` could be used when the user wants to let `joblib` decide the number of workers / processes etc.

lhoestq: ah good idea !

lhoestq (approved):
LGTM for this one :) awesome !

PR #5935, opened 2023-06-08
Better row group size in push_to_hub

lhoestq:
This is a very simple change that improves `to_parquet` to use a more reasonable row group size for image and audio datasets.

This is especially useful for `push_to_hub` and will provide a better experience with the dataset viewer on HF

### DIFF
--- a/src/datasets/io/parquet.py
+++ b/src/datasets/io/parquet.py
@@ -1,9 +1,11 @@
 import os
 from typing import BinaryIO, Optional, Union
 
+import numpy as np
 import pyarrow.parquet as pq
 
-from .. import Dataset, DatasetInfo, Features, NamedSplit, config
+from .. import Audio, Dataset, Features, Image, NamedSplit, Value, config
+from ..features.features import FeatureType, _visit
 from ..formatting import query_table
 from ..packaged_modules import _PACKAGED_DATASETS_MODULES
 from ..packaged_modules.parquet.parquet import Parquet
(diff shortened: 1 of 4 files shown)


severo:
Good idea!

I was wondering: if we want to optimize the balance between the size of downloading a row group, and the number of rows in the group, would it make sense to compute the row group size by checking the average size of the rows?

eg. 32x32 images could have a larger row group size than full HD images, no? Relying on the size would even remove the need to check the column types.

(in this proposal, we could use the computed row group size, eg 837, or use the nearest row group size in a list of values: 10, 100, 1000, 10000)

lhoestq:
Probably, but I would go for a simpler solution first :p

src/datasets/io/parquet.py
```diff
@@ -12,6 +12,35 @@
...
+            If `None`, then it will use the `datasets` default.
+    """
+    if "Audio(" in str(ds_config_info.features):
+        return PARQUET_ROW_GROUP_SIZE_FOR_AUDIO_DATASETS
+    elif "Image(" in str(ds_config_info.features):
+        return PARQUET_ROW_GROUP_SIZE_FOR_IMAGE_DATASETS
+    elif "'binary'" in str(ds_config_info.features):
+        return PARQUET_ROW_GROUP_SIZE_FOR_BINARY_DATASETS
+    else:
+        return None
```
mariosasko, line 41: I think we should traverse the features here instead of checking the strings' presence.

mariosasko (commented):
512MB-1GB is the [suggested](https://parquet.apache.org/docs/file-format/configurations/#row-group-size) row group size per the official parquet docs, so 100 rows would only be optimal for, e.g.,  Full HD images, which is still not as common as lower resolutions such as 640x480 images, for instance.

And writing optimal-sized row groups is important for sequential reads, and this is our primary use case, no?

Also, coupling the default behavior in `datasets` with `datasets-server` doesn't seem a good idea. (`datasets-server` should optimize parquet files for its use case, but here you want to copy the `push_to_hub` files to the `refs/convert/parquet` branch without modifying them to save resources?)

mariosasko:
DuckDB [paragraph](https://duckdb.org/docs/data/parquet/tips.html#selecting-a-row_group_size) that explains how to choose the `row_group_size`. Our default shard size is 500 MB in `push_to_hub`, so, ideally, we should aim for 64 MB row groups (and make this part configurable for power users 🙂).

So, before merging this PR, let's add a TODO or open an issue as a reminder that this can be improved.

src/datasets/io/parquet.py
```diff
@@ -12,6 +12,35 @@
 from .abc import AbstractDatasetReader
 
 
+PARQUET_ROW_GROUP_SIZE_FOR_AUDIO_DATASETS = 100
+PARQUET_ROW_GROUP_SIZE_FOR_IMAGE_DATASETS = 100
+PARQUET_ROW_GROUP_SIZE_FOR_BINARY_DATASETS = 100
```
mariosasko, line 17: Additional nit: Let's move these variables to `datasets.config`

lhoestq:
I moved the config values, improved the features check and mentioned the improvements we could do in the docstring :)

mariosasko (approved):
LGTM!

PR #5942, opened 2023-06-12
Pass datasets-cli additional args as kwargs to DatasetBuilder in `run_beam.py`

graelo:
Hi,

Following this <https://discuss.huggingface.co/t/how-to-preprocess-a-wikipedia-dataset-using-dataflowrunner/41991/3>, here is a simple PR to pass any additional args to datasets-cli as kwargs in the DatasetBuilder in `run_beam.py`.

I also took the liberty to add missing setup steps to the `beam.mdx` docs in order to help everyone.

### DIFF
--- a/docs/source/beam.mdx
+++ b/docs/source/beam.mdx
@@ -19,12 +19,25 @@ BUCKET=your_bucket
 REGION=your_region
 ```
 
+If you have never done so, don't forget to create a `default_applications_credentials.json` file (full docs [here](https://cloud.google.com/sdk/gcloud/reference/auth/application-default/login))
+
+```
+gcloud auth application-default login
+```
+
+This will create a file in `~/.config/gcloud/application_default_credentials.json` which the gcfs library will use to authenticate to Google Cloud Platform.
+
 3. Specify your Python requirements:
+
+Here are the dependpencies for the wikipedia dataset Beam pipeline:
 
 ```
 echo "datasets" > /tmp/beam_requirements.txt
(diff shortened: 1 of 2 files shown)


docs/source/beam.mdx
```diff
@@ -44,3 +57,20 @@ datasets-cli run_beam datasets/$DATASET_NAME \
...
+--date 20230601 \
+--save_info \
+--cache_dir gs://$BUCKET/cache/datasets \
+--beam_pipeline_options=\
+"runner=DataflowRunner,project=${PROJECT},job_name=${DATASET_NAME}-gen,"\
+"staging_location=gs://${BUCKET}/binaries,temp_location=gs://${BUCKET}/temp,"\
+"region=${REGION},requirements_file=/tmp/beam_requirements.txt"
+```
+
+Maybe a bit counter-intuitive is that you have to specify the current config name (`20220301.fr`) and then override it with the `language` and `date` arguments. This allows the dataset script [wikipedia](https://huggingface.co/datasets/wikipedia/blob/main/wikipedia.py) to have a valid config. That script allows a specific date and language to be specified, but only on top of a valid configuration.
```
lhoestq, line 76: Would it be possible to not ask for the config name when some extra config_kwargs are passed ?

Because right now it would use the `20220301.fr` config and modify it to match the extra config_kwargs, instead of creating a new config on-the-fly

graelo:
Sorry, I was way too long.

Issue #3035, opened 2021-10-05
`load_dataset` does not work with uploaded arrow file


patrickvonplaten:
## Describe the bug

I've preprocessed and uploaded a dataset here: https://huggingface.co/datasets/ami-wav2vec2/ami_headset_single_preprocessed

The dataset can correctly be loaded when doing:

```bash
git lfs install
git clone https://huggingface.co/datasets/ami-wav2vec2/ami_headset_single_preprocessed
```

followed by 

```python
from datasets import load_from_disk

ds = load_from_disk("./ami_headset_single_preprocessed")
```

However when I try to directly download the dataset as follows:

```python
from datasets import load_dataset

ds = load_dataset("ami-wav2vec2/ami_headset_single_preprocessed")
```

the following error occurs:

```bash
/usr/local/lib/python3.7/dist-packages/datasets/load.py in load_dataset(path, name, data_dir, data_files, split, cache_dir, features, download_config, download_mode, ignore_verifications, keep_in_memory, save_infos, script_version, use_auth_token, task, streaming, **config_kwargs)
   1115         ignore_verifications=ignore_verifications,
   1116         try_from_hf_gcs=try_from_hf_gcs,
-> 1117         use_auth_token=use_auth_token,
   1118     )
   1119 

/usr/local/lib/python3.7/dist-packages/datasets/builder.py in download_and_prepare(self, download_config, download_mode, ignore_verifications, try_from_hf_gcs, dl_manager, base_path, use_auth_token, **download_and_prepare_kwargs)
    635                     if not downloaded_from_gcs:
    636                         self._download_and_prepare(
--> 637                             dl_manager=dl_manager, verify_infos=verify_infos, **download_and_prepare_kwargs
    638                         )
    639                     # Sync info

/usr/local/lib/python3.7/dist-packages/datasets/builder.py in _download_and_prepare(self, dl_manager, verify_infos, **prepare_split_kwargs)
    724             try:
    725                 # Prepare split will record examples associated to the split
--> 726                 self._prepare_split(split_generator, **prepare_split_kwargs)
    727             except OSError as e:
    728                 raise OSError(

/usr/local/lib/python3.7/dist-packages/datasets/builder.py in _prepare_split(self, split_generator)
   1186                 generator, unit=" tables", leave=False, disable=bool(logging.get_verbosity() == logging.NOTSET)
   1187             ):
-> 1188                 writer.write_table(table)
   1189             num_examples, num_bytes = writer.finalize()
   1190 

[...]

PR #5944, opened 2023-06-12
Arrow dataset builder to be able to load and stream Arrow datasets

mariusz-jachimowicz-83:
This adds a Arrow dataset builder to be able to load and stream from already preprocessed Arrow files.
It's related to https://github.com/huggingface/datasets/issues/3035

### DIFF
--- a/src/datasets/packaged_modules/arrow/arrow.py
+++ b/src/datasets/packaged_modules/arrow/arrow.py
@@ -1,0 +1,73 @@
+import itertools
+from dataclasses import dataclass
+from typing import List, Optional
+
+import pyarrow as pa
+
+import datasets
+from datasets.table import _memory_mapped_record_batch_reader_from_file, read_schema_from_file, table_cast
+
+
+logger = datasets.utils.logging.get_logger(__name__)
+
+
+@dataclass
+class ArrowConfig(datasets.BuilderConfig):
+    """BuilderConfig for Arrow."""
+
+    batch_size: int = 10_000
+    columns: Optional[List[str]] = None
+    features: Optional[datasets.Features] = None
+
+
+class Arrow(datasets.ArrowBasedBuilder):
+    BUILDER_CONFIG_CLASS = ArrowConfig
+
+    def _info(self):
+        return datasets.DatasetInfo(features=self.config.features)
+
+    def _split_generators(self, dl_manager):
+        """We handle string, list and dicts in datafiles"""
+        if not self.config.data_files:
+            raise ValueError(f"At least one data file must be specified, but got data_files={self.config.data_files}")
+        data_files = dl_manager.download_and_extract(self.config.data_files)
+        if isinstance(data_files, (str, list, tuple)):
+            files = data_files
+            if isinstance(files, str):
+                files = [files]
+            # Use `dl_manager.iter_files` to skip hidden files in an extracted archive
+            files = [dl_manager.iter_files(file) for file in files]
+            return [datasets.SplitGenerator(name=datasets.Split.TRAIN, gen_kwargs={"files": files})]
+        splits = []
+        for split_name, files in data_files.items():
+            if isinstance(files, str):
+                files = [files]
+            # Use `dl_manager.iter_files` to skip hidden files in an extracted archive
+            files = [dl_manager.iter_files(file) for file in files]
+            # Infer features is they are stoed in the arrow schema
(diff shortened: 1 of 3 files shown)


src/datasets/packaged_modules/arrow/arrow.py
```diff
@@ -0,0 +1,73 @@
...
+        schema = self.info.features.arrow_schema if self.info.features is not None else None
+        if self.info.features is not None and self.config.columns is not None:
+            if sorted(field.name for field in schema) != sorted(self.config.columns):
+                raise ValueError(
+                    f"Tried to load parquet data with columns '{self.config.columns}' with mismatching features '{self.info.features}'"
+                )
+        for file_idx, file in enumerate(itertools.chain.from_iterable(files)):
+            for batch_idx, record_batch in enumerate(_memory_mapped_record_batch_reader_from_file(file)):
+                pa_table = pa.Table.from_batches([record_batch])
+                yield f"{file_idx}_{batch_idx}", self._cast_table(pa_table)
```
lhoestq, line 73:
```suggestion
            with open(file, "rb") as f:
                for batch_idx, record_batch in enumerate(pa.ipc.open_stream(f)):
                    pa_table = pa.Table.from_batches([record_batch])
                    yield f"{file_idx}_{batch_idx}", self._cast_table(pa_table)
```

src/datasets/packaged_modules/arrow/arrow.py
```diff
@@ -0,0 +1,73 @@
...
+        for split_name, files in data_files.items():
+            if isinstance(files, str):
+                files = [files]
+            # Use `dl_manager.iter_files` to skip hidden files in an extracted archive
+            files = [dl_manager.iter_files(file) for file in files]
+            # Infer features is they are stoed in the arrow schema
+            if self.info.features is None:
+                for file in itertools.chain.from_iterable(files):
+                    self.info.features = datasets.Features.from_arrow_schema(read_schema_from_file(file))
+                    # with open(file, "rb") as f:
```
lhoestq, line 51: Same here, we can't use read_schema_from_file. But using `open()` and pyarrow:
```suggestion
                    with open(file, "rb") as f:
                        self.info.features = datasets.Features.from_arrow_schema(pa.ipc.open_stream(f).schema)
```

lhoestq (commented):
Nice good job !

Note: when a dataset is loaded in streaming mode, the `files` used in `_generate_tables` may point to remote files (e.g. URLs). Therefore you can't really use `_memory_mapped_record_batch_reader_from_file` which only works on local files.

Instead you can use the `open` built-in, that we actually extend to support remote files, see my suggestion. I haven't tested though, let me know if it works !

src/datasets/packaged_modules/arrow/arrow.py
mariusz-jachimowicz-83: Indeed. Tips applied.

src/datasets/packaged_modules/arrow/arrow.py
mariusz-jachimowicz-83: Indeed. Tips applied.

src/datasets/packaged_modules/arrow/arrow.py
```diff
@@ -0,0 +1,83 @@
...
+            pa_table = table_cast(pa_table, self.info.features.arrow_schema)
+        return pa_table
+
+    def _generate_tables(self, files):
+        schema = self.info.features.arrow_schema if self.info.features is not None else None
+        if self.info.features is not None and self.config.columns is not None:
+            if sorted(field.name for field in schema) != sorted(self.config.columns):
+                raise ValueError(
+                    f"Tried to load parquet data with columns '{self.config.columns}' with mismatching features '{self.info.features}'"
+                )
```
lhoestq, line 69:
```suggestion
```

lhoestq (approved):
Looks like it's working like a charm ! Congrats :)

Let me just remove the "columns" config parameter since it's not used anywhere and we can merge.

EDIT: and actually `batch_size` as well, since it requires calling `read_all()` which is not super efficient in streaming mode and can take some time

lhoestq:
Let's add some documentation in a subsequent PR :)

In particular I think it's important to note to users that local arrow data are copied to cache according to the way load_dataset works, but if they want they can use Dataset.from_file instead

PR #5964, opened 2023-06-19
Always return list in `list_datasets`

### LINKED ISSUE #5925: Breaking API change in datasets.list_datasets caused by change in HfApi.list_datasets
### Describe the bug

Hi all,

after an update of the `datasets` library, we observer crashes in our code. We relied on `datasets.list_datasets` returning a `list`. Now, after the API of the HfApi.list_datasets was changed and it returns a `list` instead of an `Iterable`, the `datasets.list_datasets` now sometimes returns a `list` and somesimes an `Iterable`. 

It would be helpful to indicate that by the return type of the `datasets.list_datasets` function.

Thanks,
Martin 

Here, the code crashed after we updated the `datasets` library:

```python
# list_datasets no longer returns a list, which leads to an error when one tries to slice it
for datasets.list_datasets(with_details=True)[:limit]:
    ...
```

### Expected behavior

It would be helpful to indicate that by the return type of the `datasets.list_datasets` function.

### Environment info

Ubuntu 22.04
datasets 2.12.0

mariosasko:
Fix #5925 

Plus, deprecate `list_datasets`/`inspect_dataset` in favor of `huggingface_hub.list_datasets`/"git clone workflow" (downloads data files)

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,20 +86,18 @@ For more details on using the library with NumPy, pandas, PyTorch or TensorFlow,
 
 # Usage
 
-🤗 Datasets is made to be very simple to use. The main methods are:
-
-- `datasets.list_datasets()` to list the available datasets
-- `datasets.load_dataset(dataset_name, **kwargs)` to instantiate a dataset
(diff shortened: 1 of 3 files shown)


lhoestq (approved):
LGTM :)

PR #5969, opened 2023-06-20
Add `encoding` and `errors` params to JSON loader

mariosasko:
"Requested" in https://discuss.huggingface.co/t/utf-16-for-datasets/43828/3.

`pd.read_json` also has these parameters, so it makes sense to be consistent.

### DIFF
--- a/src/datasets/packaged_modules/json/json.py
+++ b/src/datasets/packaged_modules/json/json.py
@@ -20,6 +20,8 @@ class JsonConfig(datasets.BuilderConfig):
     """BuilderConfig for JSON."""
 
     features: Optional[datasets.Features] = None
+    encoding: str = "utf-8"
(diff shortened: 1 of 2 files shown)


src/datasets/packaged_modules/json/json.py
```diff
@@ -20,6 +20,8 @@ class JsonConfig(datasets.BuilderConfig):
     """BuilderConfig for JSON."""
 
     features: Optional[datasets.Features] = None
+    encoding: str = "utf-8"
+    errors: Optional[str] = None
```
lhoestq, line 24: Maybe rename "encoding_errors" for clarity and consistency with pd.read_json

lhoestq (approved):
Nice ! just one comment:

PR #5972, opened 2023-06-21
Filter unsupported extensions

lhoestq:
I used a regex to filter the data files based on their extension for packaged builders.

I tried and a regex is 10x faster that using `in` to check if the extension is in the list of supported extensions.

Supersedes https://github.com/huggingface/datasets/pull/5850

Close https://github.com/huggingface/datasets/issues/5849

I also did a small change to favor the parquet module in case of a draw in the extension counter.

### DIFF
--- a/src/datasets/data_files.py
+++ b/src/datasets/data_files.py
@@ -751,7 +751,7 @@ class DataFilesList(List[Union[Path, Url]]):
 
     def filter_extensions(self, extensions: List[str]) -> "DataFilesList":
         pattern = "|".join("\\" + ext for ext in extensions)
-        pattern = re.compile(f".*({pattern})(\\..+)?")
+        pattern = re.compile(f".*({pattern})(\\..+)?$")
         return DataFilesList(
             [
                 data_file
--- a/src/datasets/data_files.py
+++ b/src/datasets/data_files.py
@@ -1,4 +1,5 @@
 import os
+import re
 from functools import partial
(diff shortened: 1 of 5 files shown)


src/datasets/data_files.py
```diff
@@ -748,6 +749,18 @@ def from_local_or_remote(
...
+        pattern = "|".join("\\" + ext for ext in extensions)
+        pattern = re.compile(f".*({pattern})(\\..+)?$")
+        return DataFilesList(
+            [
+                data_file
+                for data_file in self
+                if pattern.match(data_file.name if isinstance(data_file, Path) else data_file)
+            ],
+            origin_metadata=self.origin_metadata,
+        )
```
mariosasko, line 762: Here we should also drop the origin metadata of the removed data files, no?

mariosasko (commented):
Nice! One nit:

src/datasets/data_files.py
lhoestq: `origin_metadata` is the list of origin per **pattern**, not per data file. We don't know which pattern generated which data file, and a pattern may have generated multiple data files. So I don't think we can easily drop origin metadata during filtering.

Note that for packaged builders patterns often look like "**" or "train/**" so there's only a few origin_metadata. Adding or removing new unsupported files in the dataset repo won't change the origin metadata, and they'll be filtered out.

PR #5976, opened 2023-06-21
Avoid stuck map operation when subprocesses crashes

pappacena:
I've been using Dataset.map() with `num_proc=os.cpu_count()` to leverage multicore processing for my datasets, but from time to time I get stuck processes waiting forever. Apparently, when one of the subprocesses is abruptly killed (OOM killer, segfault, SIGKILL, etc), the main process keeps waiting for the async task sent to that child process to finish.

It seems to be easy to reproduce the issue with the following script:

```
import os


def do_stuck(item):
    os.kill(os.getpid(), 9)

data = {
    "col1": list(range(5)),
    "col2": list(range(5)),
}

ds = Dataset.from_dict(
    data,
    features=Features({
        "col1": Value("int64"),
        "col2": Value("int64"),
    }),
)

print(ds.map(do_stuck, num_proc=4))
```

This is an old behavior in Python, which apparently was fixed a few years ago in `concurrent.futures.ProcessPoolExecutor` ([ref](https://bugs.python.org/issue9205)), but not in `multiprocessing.pool.Pool` / `multiprocess.pool.Pool`, which is used by `Dataset.map` ([ref](https://bugs.python.org/issue22393)).

This PR is an idea to try to detect when a child process gets killed, and raises a `RuntimeError` warning the dataset.map() caller.

EDIT: Related proposal for future improvement: https://github.com/huggingface/datasets/discussions/5977

### DIFF
--- a/src/datasets/utils/py_utils.py
+++ b/src/datasets/utils/py_utils.py
@@ -32,7 +32,7 @@ from multiprocessing import Manager
 from queue import Empty
 from shutil import disk_usage
 from types import CodeType, FunctionType
-from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, TypeVar, Union, Set
+from typing import Any, Callable, Dict, Iterable, List, Optional, Set, Tuple, TypeVar, Union
 from urllib.parse import urlparse
 
 import dill
@@ -1331,7 +1331,7 @@ def _write_generator_to_queue(queue: queue.Queue, func: Callable[..., Iterable[Y
 
 
 def _get_pool_pid(pool: Union[multiprocessing.pool.Pool, multiprocess.pool.Pool]) -> Set[int]:
-    return set([f.pid for f in pool._pool])
+    return {f.pid for f in pool._pool}
 
 
 def iflatmap_unordered(
--- a/src/datasets/utils/py_utils.py
+++ b/src/datasets/utils/py_utils.py
@@ -32,7 +32,7 @@ from multiprocessing import Manager
 from queue import Empty
 from shutil import disk_usage
 from types import CodeType, FunctionType
-from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, TypeVar, Union
+from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, TypeVar, Union, Set
 from urllib.parse import urlparse
 
 import dill
@@ -1330,18 +1330,24 @@ def _write_generator_to_queue(queue: queue.Queue, func: Callable[..., Iterable[Y
     return i
 
 
+def _get_pool_pid(pool: Union[multiprocessing.pool.Pool, multiprocess.pool.Pool]) -> Set[int]:
+    return set([f.pid for f in pool._pool])
+
+
 def iflatmap_unordered(
     pool: Union[multiprocessing.pool.Pool, multiprocess.pool.Pool],
     func: Callable[..., Iterable[Y]],
     *,
     kwargs_iterable: Iterable[dict],
 ) -> Iterable[Y]:
+    initial_pool_pid = _get_pool_pid(pool)
     manager_cls = Manager if isinstance(pool, multiprocessing.pool.Pool) else multiprocess.Manager
     with manager_cls() as manager:
         queue = manager.Queue()
         async_results = [
             pool.apply_async(_write_generator_to_queue, (queue, func, kwargs)) for kwargs in kwargs_iterable
         ]
+        subproc_killed = False
         try:
             while True:
                 try:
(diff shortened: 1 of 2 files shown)


lhoestq:
Hi ! Do you think this can be fixed at the Pool level ? Ideally it should be the Pool responsibility to handle this, not the `map` code. We could even subclass Pool if needed (at least the one from `multiprocess`)

pappacena:
@lhoestq it makes sense to me. Just pushed a refactoring creating a `class ProcessPool(multiprocess.pool.Pool)` to keep track of the PID changes.

src/datasets/utils/py_utils.py
```diff
@@ -1330,8 +1330,33 @@ def _write_generator_to_queue(queue: queue.Queue, func: Callable[..., Iterable[Y
...
+            return True
+        current_pids = self._get_current_pool_pids()
+        self._has_restarted_subprocess = current_pids != self._last_pool_pids
+        self._last_pool_pids = current_pids
+        return self._has_restarted_subprocess
+
+
 def iflatmap_unordered(
-    pool: Union[multiprocessing.pool.Pool, multiprocess.pool.Pool],
+    pool: ProcessPool,
```
lhoestq, line 1359: I think we still need to keep `Union[multiprocessing.pool.Pool, multiprocess.pool.Pool]` support in `iflatmap_unordered`

src/datasets/utils/py_utils.py
```diff
@@ -1349,6 +1374,13 @@ def iflatmap_unordered(
                 except Empty:
                     if all(async_result.ready() for async_result in async_results) and queue.empty():
                         break
+                if pool.has_restarted_subprocess():
+                    # One of the subprocesses has died. We should not wait forever.
+                    raise RuntimeError(
+                        "One of the subprocesses has abruptly died during map operation."
+                        "To debug the error, disable multiprocessing."
+                    )
```
lhoestq, line 1382: Could we do this automatically outside of `iflatmap_unordered` ? There is `pool.append()` that is automatically called in a thread to repopulate processes that crashed. Maybe you can just override `Pool.append` to raise an error if `pool_state` is set and `pool._state == multiprocessinf.pool.RUN`.

lhoestq (commented):
Nice ! Some additional suggestions:

src/datasets/utils/py_utils.py
lhoestq: hmm maybe the error wouldn't be handled by the thread correctly and the pool wouldn't close gracefully

lhoestq:
I managed to raise an error without subclassing Pool with two addition to `iflatmap_unordered`:

1. at the beggining
```python
original_pool = list(pool._pool)
```

2. in the loop
```python
if any(async_result._pool != original_pool for async_result in async_results) and queue.empty():
    raise RuntimeError(
        "One of the subprocesses has abruptly died during map operation."
        "To debug the error, disable multiprocessing."
    )
```

It's still a fix that only works for `iflatmap_unordered` (so not for map, imap etc) but is maybe simpler that subclassing. It also works for both multiprocessing.Pool and multiprocess.Pool

pappacena:
@lhoestq sorry for the delay. Busy weeks here. 

I just pushed the change you requested. It looks closer to the original proposal, actually.

It seems that `map` actually uses `iflatmap_unordered` ([here](https://github.com/huggingface/datasets/blob/819bb4346434912eb405ce3f3e9f21dc25a2fe85/src/datasets/arrow_dataset.py#L1509)). I think this solution works fine for the `map` method (which is the one being tested by the new `tests/test_arrow_dataset.py::BaseDatasetTest::test_map_crash_subprocess`, right?).

lhoestq:
Yes fixing iflatmap_unordered does fix Dataset.map, but it won't fix any Pool.map that we may use elsewhere so we'll have to keep this in mind.

lhoestq:
It looks all good to me, feel free to fix code formatting by running `make style` and we can merge :)

lhoestq:
It looks like checking the async_result._pool doesn't always work - sorry about that. We might just go back to your original solution then. Would also be cool to open an issue in `multiprocess` to ask if they have a solution or if they plan to fix this.

pappacena:
@lhoestq no problem! Reverted to the previous version.

TBH, given the discussions [in this python issue](https://github.com/python/cpython/issues/66587), I don't think the error in `multiprocess` will be merged upstream any time soon...

tests/test_arrow_dataset.py
```diff
@@ -1632,6 +1632,25 @@ def __call__(self, example):
                 dset.map(ex_cnt)
                 self.assertEqual(ex_cnt.cnt, len(dset))
 
+    def test_map_crash_subprocess(self, in_memory):
```
lhoestq, line 1635: maybe skip the test on windows ? it seems to raise a PermissionError
```suggestion
    @require_not_windows
    def test_map_crash_subprocess(self, in_memory):
```

lhoestq (approved):
Thanks !

PR #5995, opened 2023-06-27
Support returning dataframe in map transform

mariosasko:
Allow returning Pandas DataFrames in `map` transforms.

(Plus, raise an error in the non-batched mode if a returned PyArrow table/Pandas DataFrame has more than one row)


lhoestq (approved):
Awesome ! :)

PR #5996, opened 2023-06-28
Deprecate `use_auth_token` in favor of `token`

mariosasko:
... to be consistent with `transformers` and `huggingface_hub`.

### DIFF
--- a/docs/source/upload_dataset.mdx
+++ b/docs/source/upload_dataset.mdx
@@ -113,16 +113,16 @@ To set your dataset as private, set the `private` parameter to `True`. This para
 
 A private dataset is only accessible by you. Similarly, if you share a dataset within your organization, then members of the organization can also access the dataset.
 
-Load a private dataset by providing your authentication token to the `use_auth_token` parameter:
+Load a private dataset by providing your authentication token to the `token` parameter:
 
 ```py
 >>> from datasets import load_dataset
 
 # Load a private individual dataset
->>> dataset = load_dataset("stevhliu/demo", use_auth_token=True)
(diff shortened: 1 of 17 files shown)


lhoestq (commented):
Cool ! AFAIK transformers and use_auth_token still uses `use_auth_token` for models no ? cc @LysandreJik

There is also the option of not having a deprecation message before we update all the scripts in `transformers`

mariosasko:
They use `token` and emit a deprecation warning if use_auth_token is passed instead (see https://github.com/huggingface/transformers/blob/78a2b19fc84ed55c65f4bf20a901edb7ceb73c5f/src/transformers/modeling_utils.py#L1933).

lhoestq:
> I think we can update the examples scripts after merging this PR.

We should do a release before updated in the examples scripts no ? That's why it's an option to not have a deprecation warning until transformers and co are updated with the `token` arg

mariosasko:
> We should do a release before updated in the examples scripts no ? That's why it's an option to not have a deprecation warning until transformers and co are updated with the token arg

This would avoid the warning only for the latest `datasets` release. TBH, I don't think this is worth the hassle, considering how simple it is to remove it.

lhoestq (approved):
Looks all good then, let us know @LysandreJik if it sounds good to you

LysandreJik:
Yes, sounds great! Thanks

julien-c (commented):
note that anyways like transformers datasets already picks up the user's token automaticaly no?

lhoestq:
yup

PR #6005, opened 2023-07-04
Drop Python 3.7 support

mariosasko:
`hfh` and `transformers` have dropped Python 3.7 support, so we should do the same :).

(Based on the stats, it seems less than 10% of the users use `datasets` with Python 3.7)


lhoestq (approved):
Nice ! Always a pleasure to see old code removed :)

PR #6021, opened 2023-07-11
[docs] Update return statement of index search

stevhliu:
Clarifies in the return statement of the docstring that the retrieval score is `IndexFlatL2` by default (see [PR](https://github.com/huggingface/transformers/issues/24739) and internal Slack [convo](https://huggingface.slack.com/archives/C01229B19EX/p1689105179711689)), and fixes the formatting because multiple return values are not supported.

### DIFF
--- a/src/datasets/search.py
+++ b/src/datasets/search.py
@@ -697,7 +697,7 @@ class IndexableMixin:
         Returns:
             `(scores, indices)`:
                 A tuple of `(scores, indices)` where:
-                - **scores** (`List[List[float]`): the retrieval scores (`IndexFlatL2` score by default) of the retrieved examples
+                - **scores** (`List[List[float]`): the retrieval scores from either FAISS (`IndexFlatL2` by default) or ElasticSearch of the retrieved examples
                 - **indices** (`List[List[int]]`): the indices of the retrieved examples
         """
         self._check_index_is_initialized(index_name)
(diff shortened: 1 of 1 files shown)


src/datasets/search.py
```diff
@@ -757,8 +763,10 @@ def get_nearest_examples_batch(
                 The number of examples to retrieve per query.
 
         Returns:
-            - total_scores (`List[List[float]`): The retrieval scores of the retrieved examples per query.
-            - total_examples (`List[dict]`): The retrieved examples per query.
+            `(total_scores, total_examples)`:
+                A tuple of `(total_scores, total_examples)` where:
+                - **total_scores** (`List[List[float]`): the retrieval scores (`IndexFlatL2` score by default) of the retrieved examples per query
+                - **total_samples** (`List[dict]`): the retrieved examples per query
```
stevhliu, line 769: Is this supposed to be `total_samples` (see [line 770](https://github.com/huggingface/datasets/blob/b8067c0262073891180869f700ebef5ac3dc5cce/src/datasets/search.py#L770)) instead of `total_examples`?

lhoestq: ideally we should be consistent and always use "examples" instead of "samples"

src/datasets/search.py
```diff
@@ -695,8 +695,10 @@ def search(self, index_name: str, query: Union[str, np.array], k: int = 10, **kw
                 The number of examples to retrieve.
 
         Returns:
-            - scores (`List[List[float]`): The retrieval scores of the retrieved examples.
-            - indices (`List[List[int]]`): The indices of the retrieved examples.
+            `(scores, indices)`:
+                A tuple of `(scores, indices)` where:
+                - **scores** (`List[List[float]`): the retrieval scores (`IndexFlatL2` score by default) of the retrieved examples
```
lhoestq, line 700: it's `IndexFlatL2` by default for a faiss index, but not for an ElasticSearch index
```suggestion
                - **scores** (`List[List[float]`): the retrieval scores from either FAISS (`IndexFlatL2` by default) or ElasticSearch of the retrieved examples
```

lhoestq (approved):
Thanks ! Just one suggestion:

Issue #6017, opened 2023-07-11
Switch to huggingface_hub's HfFileSystem


lhoestq:
instead of the current datasets.filesystems.hffilesystem.HfFileSystem which can be slow in some cases

related to https://github.com/huggingface/datasets/issues/5846 and https://github.com/huggingface/datasets/pull/5919

PR #5919, opened 2023-06-01
add support for storage_options for load_dataset API

janineguo:
to solve the issue in #5880 

1.  add s3 support in the link check step, previous we only check `http` and `https`,

2. change the parameter of `use_auth_token` to `download_config` to support both `storage_options` and `use_auth_token` parameter when trying to handle(list, open, read, etc,.) the remote files.

3.  integrate the check part's duplicate code to make adding or deleting other sources easier.

PR #6028, opened 2023-07-13
Use new hffs

lhoestq:
Thanks to @janineguo 's work in https://github.com/huggingface/datasets/pull/5919 which was needed to support HfFileSystem.

Switching to `HfFileSystem` will help implementing optimization in data files resolution

## Implementation details

I replaced all the from_hf_repo and from_local_or_remote in data_files.py to only use a new `from_patterns` which works for any fsspec path, including hf:// paths, https:// URLs and local paths. This simplifies the codebase since there is no logic duplication anymore when it comes to data files resolution.

I added `_prepare_path_and_storage_options` which returns the right storage_options to use given a path and a `DownloadConfig`. This is the only place where the logic depends on the filesystem type that must be used.

I also removed the `get_metadata_data_files_list ` and `get_patterns_and_data_files` functions added recently, since data files resolution is now handled using a common interface.

## New features

hf:// paths are now supported in data_files

## Breaking changes

DataFilesList and DataFilesDict:
- use `str` paths instead of `Union[Path, Url]`
- require posix paths for windows paths

close https://github.com/huggingface/datasets/issues/6017

### DIFF
--- a/src/datasets/data_files.py
+++ b/src/datasets/data_files.py
@@ -287,7 +287,7 @@ def resolve_pattern(
     fs_base_path = base_path.split("::")[0].split("://")[-1] or fs.root_marker
     fs_pattern = pattern.split("::")[0].split("://")[-1]
     protocol_prefix = fs.protocol + "://" if fs.protocol != "file" else ""
-    files_to_ignore = set(FILES_TO_IGNORE) - set(xbasename(pattern))
+    files_to_ignore = set(FILES_TO_IGNORE) - {xbasename(pattern)}
     matched_paths = [
         filepath if filepath.startswith(protocol_prefix) else protocol_prefix + filepath
         for filepath in fs.glob(pattern)
--- a/src/datasets/data_files.py
+++ b/src/datasets/data_files.py
@@ -614,7 +614,7 @@ class DataFilesDict(Dict[str, DataFilesList]):
     @classmethod
     def from_patterns(
         cls,
-        patterns: List[str],
+        patterns: Dict[str, Union[List[str], DataFilesList]],
         base_path: Optional[str] = None,
         allowed_extensions: Optional[List[str]] = None,
         download_config: Optional[DownloadConfig] = None,
--- a/src/datasets/data_files.py
+++ b/src/datasets/data_files.py
@@ -405,7 +405,7 @@ def get_data_patterns(base_path: str, download_config: Optional[DownloadConfig] 
 
     Output:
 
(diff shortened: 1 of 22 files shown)


src/datasets/data_files.py
```diff
@@ -235,20 +240,20 @@ def _get_data_files_patterns(pattern_resolver: Callable[[str], List[PurePath]])
...
-                    if len(data_files) > 0:
-                        non_empty_splits.append(split)
-                        break
-            except FileNotFoundError:
-                pass
+                except FileNotFoundError:
+                    continue
+                if len(data_files) > 0:
+                    non_empty_splits.append(split)
+                    break
```
lhoestq, line 250: this is needed because now `resolve_pattern` returns FileNotFoundError if it can't resolve at least one file

src/datasets/load.py
```diff
@@ -415,27 +415,26 @@ def _create_importable_file(
...
         tuple[str, dict[str, Any]]: Tuple with
             - inferred module name
             - dict of builder kwargs
     """
     extensions_counter = Counter(
-        suffix.lower()
+        "." + suffix.lower()
         for filepath in data_files_list[: config.DATA_FILES_MAX_NUMBER_FOR_MODULE_INFERENCE]
-        for suffix in Path(filepath).suffixes
+        for suffix in xbasename(filepath).split(".")[1:]
```
lhoestq, line 437: this kind of changes are needed to support chained fsspec URLs

mariosasko (approved):
Great stuff 🙂!

We should also remove the legacy version of `HfFileSystem`, but this can be done in a subsequent PR.

src/datasets/download/streaming_download_manager.py
```diff
@@ -442,34 +399,62 @@ def _get_extraction_protocol(urlpath: str, token: Optional[Union[str, bool]] = N
...
+    prepared_urlpath = []
+    prepared_storage_options = {}
+    for hop in urlpath.split("::"):
+        hop, storage_options = _prepare_single_hop_path_and_storage_options(hop, download_config=download_config)
+        prepared_urlpath.append(hop)
+        prepared_storage_options.update(storage_options)
+    return "::".join(prepared_urlpath), storage_options
+
+
+def _prepare_single_hop_path_and_storage_options(
```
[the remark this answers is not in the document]
lhoestq: I opened https://github.com/huggingface/datasets/issues/6046

PR #6030, opened 2023-07-13
fixed typo in comment

NightMachinery:
This mistake was a bit confusing, so I thought it was worth sending a PR over.


mariosasko (approved):
Good catch!

PR #6036, opened 2023-07-14
Deprecate search API

mariosasko:
The Search API only supports Faiss and ElasticSearch as vector stores, is somewhat difficult to maintain (e.g., it still doesn't support ElasticSeach 8.0, difficult testing, ...), does not have the best design (adds a bunch of methods to the `Dataset` class that are only useful after creating an index), the usage doesn't seem to be significant and is not integrated with the Hub. Since we have no plans/bandwidth to improve it and better alternatives such as `langchain` and `docarray` exist, I think it should be deprecated (and eventually removed).

If we decide to deprecate/remove it, the following usage instances need to be addressed:
* [Course](https://github.com/huggingface/course/blob/0018bb434204d9750a03592cb0d4e846093218d8/chapters/en/chapter5/6.mdx#L342 ) and [Blog](https://github.com/huggingface/blog/blob/4897c6f73d4492a0955ade503281711d01840e09/image-search-datasets.md?plain=1#L252) - calling the FAISS API directly should be OK in these instances as it's pretty simple to use for basic scenarios. Alternatively, we can use `langchain`, but this adds an extra dependency
*  [Transformers](https://github.com/huggingface/transformers/blob/50726f9ea7afc6113da617f8f4ca1ab264a5e28a/src/transformers/models/rag/retrieval_rag.py#L183) - we can use the FAISS API directly and store the index as a separate attribute (and instead of building the `wiki_dpr` index each time the dataset is generated, we can generate it once and push it to the Hub repo, and then read it from there

cc @huggingface/datasets @LysandreJik for the opinion

### DIFF
--- a/docs/source/faiss_es.mdx
+++ b/docs/source/faiss_es.mdx
@@ -1,5 +1,11 @@
 # Search index
 
+<Tip warning={true}>
+
+Vector search is deprecated in 🤗 Datasets.
+
+</Tip>
+
 [FAISS](https://github.com/facebookresearch/faiss) and [ElasticSearch](https://www.elastic.co/elasticsearch/) enables searching for examples in a dataset. This can be useful when you want to retrieve specific examples from a dataset that are relevant to your NLP task. For example, if you are working on a Open Domain Question Answering task, you may want to only return examples that are relevant to answering your question.
 
 This guide will show you how to build an index for your dataset that will allow you to search it.
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -4386,7 +4386,8 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         """
         from .dataset_dict import DatasetDict  # import here because of circular dependency
 
-        with warnings.filterwarnings("ignore", category=FutureWarning):
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", FutureWarning)
             if len(self.list_indexes()) > 0:
                 raise DatasetTransformationNotAllowedError(
                     "Using `.train_test_split` on a dataset with attached indexes is not allowed. You can first run `.drop_index() to remove your index and then re-add it."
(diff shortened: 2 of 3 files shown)


lhoestq:
I don't think `transformers` should have any dataset indexing code. So before deprecating I'd be in favor of finding a suitable replacement. Not sure about the stats of the RAG model that uses `datasets` indexing though

mariosasko:
The RAG downloads stats are decent (over 20k downloads last month).

I think it's suboptimal to maintain an API that only a single model uses. One option is to put this code into a separate lib. However, `langchain` and `docarray` already provide a unified interface to vector stores, so I don't see this as an impactful project. Considering how specific this model is, I think we should go with the simplest solution and combine an index with a dataset in Transformers (this wouldn't require too much code).

julien-c:
What about migrating to the [datasets-server](https://github.com/huggingface/datasets-server) search feature instead? Would make more sense from a product perspective

lhoestq:
I don't think it's a good idea:
- using datasets-server would require to upload the data and to not control the indexing, whereas the current feature is about using a local index that you control
- faiss indexes are vector indexes that are not supported by datasets-server, and they are also very customised. For instance RAG uses DPR embeddings and cosine similarity
- FTS is only done for the first 5GB of data for now in datasets-server

LysandreJik:
After discussing it a bit with @lhoestq, do we need to deprecate the search API? While I understand it's imperfect, it looks like this will result in significant work to update it everywhere, so I'd favor keeping it until there's an obviously better alternative; this way we can focus on different things in the meantime.

mariosasko:
FAISS/ES are simple to use (probably the main reason why they are so popular), so creating "better alternatives" is not easy - they usually add more complexity (as is the case here, `langchain`, etc.)

So, instead of waiting for better alternatives, IMO it makes more sense to wait for the RAG model to be deprecated in Transformers (less than 1,000 cumulated downloads over all checkpoints in the past 30 days) before deprecating this API here.

In the meantime, we should make it clear that the vector search API is in maintenance mode (no new features, etc.).

How does that sound?

PR #6042, opened 2023-07-17
Fix unused DatasetInfosDict code in push_to_hub

lhoestq:


src/datasets/dataset_dict.py
```diff
@@ -1722,8 +1711,15 @@ def push_to_hub(
                 MetadataConfigs({"default": default_metadata_configs_to_dump}).to_dataset_card_data(dataset_card_data)
         # push to the deprecated dataset_infos.json
         if config.DATASETDICT_INFOS_FILENAME in repo_files:
+            download_config = DownloadConfig()
```
polinaeterna, line 1714: why did you add this here?

src/datasets/dataset_dict.py
```diff
@@ -1722,8 +1711,15 @@ def push_to_hub(
...
+            download_config.download_desc = "Downloading metadata"
+            download_config.token = token
+            dataset_infos_path = cached_path(
+                hf_hub_url(repo_id, config.DATASETDICT_INFOS_FILENAME),
+                download_config=download_config,
+            )
```
lhoestq, line 1720: this code needs to be here, because in the previous location it would be run only if there is no README.md and if there is a json file. But it needs to run if there is a json file, no matter if there is a README.md or not

polinaeterna: ah lol true good catch!

polinaeterna (commented):
thank you :heart:

polinaeterna (approved):

PR #6061, opened 2023-07-24
Dill 3.7 support

mariosasko:
Adds support for dill 3.7.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -113,7 +113,7 @@ REQUIRED_PKGS = [
     # Minimum 8.0.0 to be able to use .to_reader()
     "pyarrow>=8.0.0",
     # For smart caching dataset processing
-    "dill>=0.3.0",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -113,7 +113,7 @@
     # Minimum 8.0.0 to be able to use .to_reader()
     "pyarrow>=8.0.0",
     # For smart caching dataset processing
-    "dill>=0.3.0,<0.3.7",  # tmp pin until next 0.3.7 release: see https://github.com/huggingface/datasets/pull/5166
+    "dill>=0.3.0",
```
lhoestq, line 116: I think we should pin <0.3.8 no ? In case a future version of dill breaks the deterministic behaviors the cache is relying on

lhoestq: especially since determinism is not officially supported yet: https://github.com/uqfoundation/dill/pull/501

lhoestq (approved):
LGTM :)

Issue #5965, opened 2023-06-19
"Couldn't cast array of type" in complex datasets


piercefreeman:
### Describe the bug

When doing a map of a dataset with complex types, sometimes `datasets` is unable to interpret the valid schema of a returned datasets.map() function. This often comes from conflicting types, like when both empty lists and filled lists are competing for the same field value.

This is prone to happen in batch mapping, when the mapper returns a sequence of null/empty values and other batches are non-null. A workaround is to manually cast the new batch to a pyarrow table (like implemented in this [workaround](https://github.com/piercefreeman/lassen/pull/3)) but it feels like this ideally should be solved at the core library level.

Note that the reproduction case only throws this error if the first datapoint has the empty list. If it is processed later, datasets already detects its representation as list-type and therefore allows the empty list to be provided.

### Steps to reproduce the bug

A trivial reproduction case:

```python
from typing import Iterator, Any
import pandas as pd
from datasets import Dataset

def batch_to_examples(batch: dict[str, list[Any]]) -> Iterator[dict[str, Any]]:
    for i in range(next(iter(lengths))):
        yield {feature: values[i] for feature, values in batch.items()}

def examples_to_batch(examples) -> dict[str, list[Any]]:
    batch = {}

    for example in examples:
        for feature, value in example.items():
            if feature not in batch:
                batch[feature] = []
            batch[feature].append(value)

    return batch

def batch_process(examples, explicit_schema: bool):
    new_examples = []
    for example in batch_to_examples(examples):
        new_examples.append(dict(texts=example["raw_text"].split()))
    return examples_to_batch(new_examples)

df = pd.DataFrame(
    [
        {"raw_text": ""},
        {"raw_text": "This is a test"},
        {"raw_text": "This is another test"},
    ]
)

dataset = Dataset.from_pandas(df)

# datasets won't be able to typehint a dataset that starts with an empty example.
with pytest.raises(TypeError, match="Couldn't cast array of type"):
    dataset = dataset.map(
        batch_process,
        batched=True,
        batch_size=1,
        num_proc=1,
        remove_columns=dataset.column_names,
    )
```

This results in crashes like:

```bash
  File "/Users/piercefreeman/Library/Caches/pypoetry/virtualenvs/example-9kBqeSPy-py3.11/lib/python3.11/site-packages/datasets/table.py", line 1819, in wrapper
    return func(array, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #6065, opened 2023-07-25
Add column type guessing from map return function

piercefreeman:
As discussed [here](https://github.com/huggingface/datasets/issues/5965), there are some cases where datasets is unable to automatically promote columns during mapping. The fix is to explicitly provide a `features` definition so pyarrow can configure itself with the right column types from the outset.

This PR provides an alternative approach, which is functionally equivalent to specifying features but a bit cleaner within a larger mapping pipeline. It allows clients to typehint the return variable coming from the mapper function - if we find one of these type annotations specified, and no explicit features have been passed in, we'll try to convert it into a Features map. If the map function runs and casting is unable to succeed, it will raise a DatasetTransformationNotAllowedError that indicates the typehint may be to blame. It works for batched and non-batched mapping functions.

Currently supported column types:
- builtins primitives: string, int, float, bool
- dictionaries, lists (nested and one-deep)
- Optional types and None-Unions (synonymous with optional types)

It's used like:

```python
class DatasetTyped(TypedDict):
  texts: list[str]

def dataset_typed_map(batch) -> DatasetTyped:
  return {"texts": [text.split() for text in batch["raw_text"]]}

dataset = {"raw_text": ["", "This is a test", "This is another test"]}

with Dataset.from_dict(dataset) as dset:
  new_dataset = dset.map(
    dataset_typed_map,
    batched=True,
    batch_size=1,
    num_proc=1,
)
```

Open questions:
- Should logging indicate we have automatically guessed these types? Or proceed quietly until we hit an error (as is the current implementation).


mariosasko:
Thanks for working on this. However, having thought about this issue a bit more, supporting this doesn't seem like a good idea - it's better to be explicit than implicit, according to the Zen of Python 🙂. Also, I don't think many users would use this, so this raises the question of whether this is something we want to maintain.

cc @lhoestq for the 2nd opinion

piercefreeman:
@mariosasko I was going to quote the Zen of Python in the other direction :) To me, this actually is much more explicit than the current behavior of guessing pyarrow types based on the raw dictionary return values. Explicit typehinting is increasingly the de facto way to deal with this dynamic type serialization - plus it feels like a clearer fit to me than separating out the mapper function from the feature column definition in the call to the actual `.map()`. Another benefit is providing typehinting support for clients that use mypy or other static typecheckers to detect return mismatches.

But will leave it to you and @lhoestq to see if it's something you'd like in core versus a support package.

mariosasko:
I meant that explicitly specifying the target features  (the `features` param) is cleaner (easier to track) than relying on type hints.

lhoestq:
Passing features= to `map()` is richer and more explicit. Also I don't think users would guess that such API exist.

Other libraries like dask also infer the type from the output or requires the typing to be specified using the `meta` argument

piercefreeman:
Point about discoverability is a fair one, would certainly need some docs around it. All good! Will close this out and keep in our extension utilities.

Issue #6066, opened 2023-07-25
AttributeError: '_tqdm_cls' object has no attribute '_lock'


codingl2k1:
### Describe the bug

```python
  File "/Users/codingl2k1/.pyenv/versions/3.11.4/lib/python3.11/site-packages/datasets/load.py", line 1034, in get_module
    data_files = DataFilesDict.from_patterns(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/codingl2k1/.pyenv/versions/3.11.4/lib/python3.11/site-packages/datasets/data_files.py", line 671, in from_patterns
    DataFilesList.from_patterns(
  File "/Users/codingl2k1/.pyenv/versions/3.11.4/lib/python3.11/site-packages/datasets/data_files.py", line 586, in from_patterns
    origin_metadata = _get_origin_metadata(data_files, download_config=download_config)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/codingl2k1/.pyenv/versions/3.11.4/lib/python3.11/site-packages/datasets/data_files.py", line 502, in _get_origin_metadata
    return thread_map(
           ^^^^^^^^^^^
  File "/Users/codingl2k1/.pyenv/versions/3.11.4/lib/python3.11/site-packages/tqdm/contrib/concurrent.py", line 70, in thread_map
    return _executor_map(ThreadPoolExecutor, fn, *iterables, **tqdm_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/codingl2k1/.pyenv/versions/3.11.4/lib/python3.11/site-packages/tqdm/contrib/concurrent.py", line 48, in _executor_map
    with ensure_lock(tqdm_class, lock_name=lock_name) as lk:
  File "/Users/codingl2k1/.pyenv/versions/3.11.4/lib/python3.11/contextlib.py", line 144, in __exit__
    next(self.gen)
  File "/Users/codingl2k1/.pyenv/versions/3.11.4/lib/python3.11/site-packages/tqdm/contrib/concurrent.py", line 25, in ensure_lock
    del tqdm_class._lock
        ^^^^^^^^^^^^^^^^
AttributeError: '_tqdm_cls' object has no attribute '_lock'
```

[...]

PR #6068, opened 2023-07-25
fix tqdm lock deletion

lhoestq:
related to https://github.com/huggingface/datasets/issues/6066

### DIFF
--- a/src/datasets/utils/logging.py
+++ b/src/datasets/utils/logging.py
@@ -202,7 +202,6 @@ _tqdm_active = True
 
 
 class _tqdm_cls:
-
     def __call__(self, *args, disable=False, **kwargs):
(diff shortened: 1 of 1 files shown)


src/datasets/utils/logging.py
```diff
@@ -202,8 +202,6 @@ def __exit__(self, type_, value, traceback):
 
 
 class _tqdm_cls:
-    _lock = None
```
lhoestq, line 205: revert https://github.com/huggingface/datasets/pull/6067

mariosasko (approved):
A bit surprised this hasn't been reported earlier, considering this part of `tqdm.concurrent` is pretty old (3 years, to be precise)

PR #6074, opened 2023-07-26
Misc doc improvements

mariosasko:
Removes the warning about requiring to write a dataset loading script to define multiple configurations, as the README YAML can be used instead. Also, deletes the section about using the `BatchSampler` in `torch<=1.12.1` to speed up loading, as `torch 1.12.1` is over a year old (and `torch 2.0` has been out for a while).


lhoestq (approved):
LGTM :)

lhoestq:
merging now if you don't mind - this way I can make a patch release

Issue #2918, opened 2021-09-15
`Can not decode content-encoding: gzip` when loading `scitldr` dataset with streaming


SBrandeis:
## Describe the bug

Trying to load the `"FullText"` config of the `"scitldr"` dataset with `streaming=True` raises an error from `aiohttp`:
```python
ClientPayloadError: 400, message='Can not decode content-encoding: gzip'
```

cc @lhoestq 

## Steps to reproduce the bug
```python
from datasets import load_dataset

iter_dset = iter(
    load_dataset("scitldr", name="FullText", split="test", streaming=True)
)

next(iter_dset)
```

## Expected results
Returns the first sample of the dataset

## Actual results
Calling `__next__` crashes with the following Traceback:

```python
----> 1 next(dset_iter)

~\miniconda3\envs\datasets\lib\site-packages\datasets\iterable_dataset.py in __iter__(self)
    339
    340     def __iter__(self):
--> 341         for key, example in self._iter():
    342             if self.features:
    343                 # we encode the example for ClassLabel feature types for example

~\miniconda3\envs\datasets\lib\site-packages\datasets\iterable_dataset.py in _iter(self)
    336         else:
    337             ex_iterable = self._ex_iterable
--> 338         yield from ex_iterable
    339
    340     def __iter__(self):

~\miniconda3\envs\datasets\lib\site-packages\datasets\iterable_dataset.py in __iter__(self)
     76
     77     def __iter__(self):
---> 78         for key, example in self.generate_examples_fn(**self.kwargs):
     79             yield key, example
     80

~\.cache\huggingface\modules\datasets_modules\datasets\scitldr\72d6e2195786c57e1d343066fb2cc4f93ea39c5e381e53e6ae7c44bbfd1f05ef\scitldr.py in _generate_examples(self, filepath, split)
    162
    163         with open(filepath, encoding="utf-8") as f:
--> 164             for id_, row in enumerate(f):
    165                 data = json.loads(row)
    166                 if self.config.name == "AIC":

~\miniconda3\envs\datasets\lib\site-packages\fsspec\implementations\http.py in read(self, length)
    496         else:
    497             length = min(self.size - self.loc, length)
--> 498         return super().read(length)
    499
    500     async def async_fetch_all(self):

~\miniconda3\envs\datasets\lib\site-packages\fsspec\spec.py in read(self, length)
   1481             # don't even bother calling fetch
   1482             return b""
[...]

PR #6076, opened 2023-07-26
No gzip encoding from github

lhoestq:
Don't accept gzip encoding from github, otherwise some files are not streamable + seekable.

fix https://huggingface.co/datasets/code_x_glue_cc_code_to_code_trans/discussions/2#64c0e0c1a04a514ba6303e84

and making sure https://github.com/huggingface/datasets/issues/2918 works as well


albertvillanova (commented):
Thanks for the fix.

Some questions: won't this have an impact on downloading time, once we do not longer compress the payload? What is the advantage of this approach over the one with `block_size: 0`?

See: https://huggingface.co/datasets/code_x_glue_cc_code_to_code_trans/discussions/3

lhoestq:
> Some questions: won't this have an impact on downloading time, once we do not longer compress the payload? What is the advantage of this approach over the one with block_size: 0?

Surely, but this prevents random access which is needed at multiple places in the code (eg to check the compression type).
Github isn't a good place for big files anyway so we should be fine

PR #6093, opened 2023-07-28
Deprecate `download_custom`

mariosasko:
Deprecate `DownloadManager.download_custom`. Users should use `fsspec` URLs (cacheable) or make direct requests with `fsspec`/`requests` (not cacheable) instead.

We should deprecate this method as it's not compatible with streaming, and implementing the streaming version of it is hard/impossible. There have been requests to implement the streaming version of this method on the forum, but the reason for this seems to be a tip in the docs that "promotes" this method (this PR removes it).


lhoestq (approved):
good idea !

mariosasko:
I forgot to mention this in the initial comment, but only one public dataset (excluding gated) uses this method - `pg19`, which I just fixed.

ProgramComputer:
@mariosasko How would you stream a split zip file with just [download_and_extract or download](https://github.com/huggingface/datasets/blob/main/src/datasets/download/download_manager.py#L353)? With download_custom, it is possible to combine a split zip file. Perhaps add an option in [download ](https://huggingface.co/docs/datasets/v2.2.1/en/package_reference/builder_classes#datasets.DownloadManager.download)to combine split zips.

PR #6098, opened 2023-07-29
Expanduser in save_to_disk()

### LINKED ISSUE #5651: expanduser in save_to_disk
### Describe the bug

save_to_disk() does not expand `~`

1. `dataset = load_datasets("any dataset")`
2. `dataset.save_to_disk("~/data")`
3. a folder named "~" created in current folder
4. FileNotFoundError is raised, because the expanded path does not exist (`/home/<user>/data`)

related issue https://github.com/huggingface/transformers/issues/10628

As described above.

### Expected behavior

expanduser correctly

### Environment info

- datasets 2.10.1
- python 3.10

Unknown3141592:
Fixes #5651. The same problem occurs when loading from disk so I fixed it there too.

I am not sure why the case distinction between local and remote filesystems is even necessary for `DatasetDict` when saving to disk. Imo this could be removed (leaving only `fs.makedirs(dataset_dict_path, exist_ok=True)`).

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -1459,15 +1459,14 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
             raise ValueError("please remove all the indexes using `dataset.drop_index` before saving a dataset")
 
         if is_local:
-            Path(dataset_path).expanduser().resolve().mkdir(parents=True, exist_ok=True)
+            resolved_path = Path(dataset_path).expanduser().resolve()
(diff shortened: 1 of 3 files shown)


mariosasko (approved):
I renamed `resolved_path` to `save_path`. Looks all good now. Thanks!

mariosasko:
> I am not sure why the case distinction between local and remote filesystems is even necessary for DatasetDict when saving to disk. Imo this could be removed (leaving only fs.makedirs(dataset_dict_path, exist_ok=True)).

Indeed. But it's better to address this in a separate PR.

PR #6145, opened 2023-08-12
Export to_iterable_dataset to document

npuichigo:
Fix the export of a missing method of `Dataset`


stevhliu (commented):
Thanks for adding this to the API reference!

I think it'd be nice to also mention this in the [Stream](https://huggingface.co/docs/datasets/stream) guide. WDYT @lhoestq @mariosasko?

lhoestq (approved):
Thanks ! 

> I think it'd be nice to also mention this in the [Stream](https://huggingface.co/docs/datasets/stream) guide. WDYT @lhoestq @mariosasko?

Yes indeed that would be nice ! This can be done in another PR if anyone wants to complete the documentation :)

PR #6148, opened 2023-08-14
Ignore parallel warning in map_nested

lhoestq:
This warning message was shown every time you pass num_proc to `load_dataset` because of `map_nested`

```
parallel_map is experimental and might be subject to breaking changes in the future
```

This PR removes it for `map_nested`. If someone uses another parallel backend they're already warned when `parallel_backend` is called anyway


mariosasko (approved):
LGTM!

PR #6155, opened 2023-08-17
Raise FileNotFoundError when passing data_files that don't exist

lhoestq:
e.g. when running `load_dataset("parquet", data_files="doesnt_exist.parquet")`

### DIFF
--- a/src/datasets/data_files.py
+++ b/src/datasets/data_files.py
@@ -573,14 +573,18 @@ class DataFilesList(List[str]):
         base_path = base_path if base_path is not None else Path().resolve().as_posix()
         data_files = []
         for pattern in patterns:
-            data_files.extend(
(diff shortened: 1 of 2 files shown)


src/datasets/data_files.py
```diff
@@ -583,7 +583,8 @@ def from_patterns(
                     )
                 )
             except FileNotFoundError:
-                pass
+                if "*" not in pattern and "[" not in pattern and "?" not in pattern:
```
mariosasko, line 586: The `glob` module `has_magic` (public, but not documented IIRC; e.g., `fsspec` uses it) to check this:
```suggestion
                if has_magic(pattern):
```

But feel free to decide which one you prefer.

mariosasko (approved):
Nice!

PR #6161, opened 2023-08-17
Fix protocol prefix for Beam

### LINKED ISSUE #6147: ValueError when running BeamBasedBuilder with GCS path in cache_dir
### Describe the bug

When running the BeamBasedBuilder with a GCS path specified in the cache_dir, the following ValueError occurs:

```
ValueError: Unable to get filesystem from specified path, please use the correct path or ensure the required dependency is installed, e.g., pip install apache-beam[gcp]. Path specified: gcs://my-bucket/huggingface_datasets/my_beam_dataset/default/0.0.0/my_beam_dataset-train [while running 'train/Save to parquet/Write/WriteImpl/InitializeWrite']
```

Same error occurs after running `pip install apache-beam[gcp]` as instructed.

Put `my_beam_dataset.py`:

```python
import datasets


class MyBeamDataset(datasets.BeamBasedBuilder):
    def _info(self):
        features = datasets.Features({"value": datasets.Value("int64")})
        return datasets.DatasetInfo(features=features)

    def _split_generators(self, dl_manager, pipeline):
        return [datasets.SplitGenerator(name=datasets.Split.TRAIN, gen_kwargs={})]

    def _build_pcollection(self, pipeline):
        import apache_beam as beam

        return pipeline | beam.Create([{"value": i} for i in range(10)])
```

Run:
```bash
datasets-cli run_beam my_beam_dataset.py --cache_dir=gs://my-bucket/huggingface_datasets/ --beam_pipeline_options="runner=DirectRunner"
```

### Expected behavior

Running the BeamBasedBuilder with a GCS cache path without any errors.

### Environment info

- `datasets` version: 2.14.4
- Platform: macOS-13.4-arm64-arm-64bit
- Python version: 3.9.17
- Huggingface_hub version: 0.16.4
- PyArrow version: 9.0.0
- Pandas version: 2.0.3

mariosasko:
Fix #6147

### DIFF
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -690,6 +690,9 @@ class DatasetBuilder:
     def _rename(self, src: str, dst: str):
         rename(self._fs, src, dst)
 
+    def _format_path(self, path: str) -> str:
+        return path  # no-op
+
     def download_and_prepare(
         self,
         output_dir: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


mariosasko:
The CI errors are unrelated to the changes

lhoestq (commented):
Maybe we can simply raise an error if gs:// or gcs:// are used ?

PR #6165, opened 2023-08-22
Fix multiprocessing with spawn in iterable datasets

Hubert-Bonisseur:
The "Spawn" method is preferred when multiprocessing on macOS or Windows systems, instead of the "Fork" method on linux systems.

This causes some methods of Iterable Datasets to break when using a dataloader with more than 0 workers.

I fixed the issue by replacing lambda and local methods which are not pickle-able.

See the example below:

```python


if __name__ == "__main__":
    dataset = load_dataset("lhoestq/demo1", split="train")
    dataset = dataset.to_iterable_dataset(num_shards=3)

    dataset = dataset.remove_columns(["package_name"])
    dataset = dataset.rename_columns({
        "review": "review1"
    })
    dataset = dataset.rename_column("date", "date1")
    for sample in DataLoader(dataset, batch_size=None, num_workers=3):
        print(sample)
```

To notice the fix on a linux system, adding these lines should do the trick:

```python
import multiprocessing
multiprocessing.set_start_method('spawn')
```

I also removed what looks like code duplication between rename_colums and rename_column

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -5,6 +5,7 @@ import warnings
 from collections import Counter
 from copy import deepcopy
 from dataclasses import dataclass
+from functools import partial
 from itertools import cycle, islice
 from typing import Any, Callable, Dict, Iterable, Iterator, List, Optional, Tuple, Union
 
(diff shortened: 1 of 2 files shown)


Hubert-Bonisseur:
@lhoestq 
A test is failing, but I don't think it is due to my changes

Hubert-Bonisseur:
@lhoestq 
I added the test and fixed one last method

lhoestq (approved):
Awesome thanks !

PR #6167, opened 2023-08-22
Allow hyphen in split name

mariosasko:
To fix https://discuss.huggingface.co/t/error-when-setting-up-the-dataset-viewer-streamingrowserror/51276.


severo:
Note that the https://github.com/huggingface/datasets-server/ explicitly relies on the fact that a split cannot contain an hyphen. cc @lhoestq

lhoestq:
We can't enable this that easily unfortunately because it could make arrow file names ambiguous in the cache.

e.g. dataset_name-train-0000-of-0008.arrow

mariosasko:
Oh, this would indeed make the caching for the multi-proc case ambiguous. Implementing this is only worth it if we get more requests, so I'm closing this PR for now.

PR #6191, opened 2023-08-29
Add missing `revision` argument

qgallouedec:
I've noticed that when you're not working on the main branch, there are sometimes errors in the files returned. After some investigation, I realized that the revision was not properly passed  everywhere. This PR proposes a fix.


liyucheng09:
I have found the same issue. Good fix. Should be merged as soon as possible.

lhoestq (approved):
Good catch ! thanks for fixing it

PR #6211, opened 2023-09-04
Fix empty splitinfo json

lhoestq:
If a split is empty, then the JSON split info should mention num_bytes = 0 and num_examples = 0.

Until now they were omited because the JSON dumps ignore the fields that are equal to the default values.

This is needed in datasets-server since we parse this information to the viewer


albertvillanova (approved):
Thanks for the fix.

mariosasko (approved):

PR #6213, opened 2023-09-04
Better list array values handling in cast/embed storage

mariosasko:
Use [`array.flatten`](https://arrow.apache.org/docs/python/generated/pyarrow.ListArray.html#pyarrow.ListArray.flatten) that takes `.offset` into account instead of `array.values` in array cast/embed.


mariosasko:
A bug in `FixedSizeArray.flatten` in `PyArrow<10.0.0` makes CI fail. Colab installs 9.0.0 by default, so we should be able to set the minimal version to `10.0.0` soon. Keeping this PR as a draft in the meantime.

PR #6215, opened 2023-09-05
Fix checking patterns to infer packaged builder 

polinaeterna:
Don't ignore results of pattern resolving if `self.data_files` is not None. Otherwise lines 854 and 1037 make no sense.


lhoestq:
oh wow good catch

lhoestq (approved):

PR #6218, opened 2023-09-06
Rename old push_to_hub configs to "default" in dataset_infos

lhoestq:
Fix

```python

b = load_dataset_builder("lambdalabs/pokemon-blip-captions", "default")
print(b.info)
```

which should return

```
DatasetInfo(
  features={'image': Image(decode=True, id=None), 'text': Value(dtype='string', id=None)}, 
  dataset_name='pokemon-blip-captions',
  config_name='default', 
  version=0.0.0,
  splits={'train': SplitInfo(name='train', num_bytes=119417410.0, num_examples=833, shard_lengths=None, dataset_name='pokemon-blip-captions')},
  download_checksums=None,
  download_size=99672355,
  dataset_size=119417410.0,
  size_in_bytes=219089765.0,
  ...
)
```

instead of and empty dataset info.

The dataset has a dataset_infos.json file with a deprecated config name "lambdalabs--pokemon-blip-captions". We switched those config names to "default" in 2.14, so the builder.info should take this into account.

### DIFF
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -913,9 +913,9 @@ class LocalDatasetModuleFactoryWithoutScript(_DatasetModuleFactory):
                     }
                 )
                 if len(legacy_dataset_infos) == 1:
+                    # old config e.g. named "username--dataset_name"
                     legacy_config_name = next(iter(legacy_dataset_infos))
-                    if "--" in legacy_config_name:  # old config named "username--dataset_name"
-                        legacy_dataset_infos["default"] = legacy_dataset_infos.pop(legacy_config_name)
+                    legacy_dataset_infos["default"] = legacy_dataset_infos.pop(legacy_config_name)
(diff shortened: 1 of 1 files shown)


albertvillanova (approved):
Thanks.

polinaeterna:
thanks! i wonder if we should also fix (change config name) all the old `dataset_infos.json` on the Hub?

albertvillanova:
I would say we should delete all `dataset_infos.json` on the Hub...

polinaeterna:
@albertvillanova @lhoestq @mariosasko should we really delete it?
(bc if not, I've found a bug in updating `dataset_infos.json` with `.push_to_hub` and I'd open a PR to fix it)

lhoestq:
We can only delete them for the datasets without namespace and open PRs for the others, so we need to keep supporting them for now

PR #6222, opened 2023-09-06
fix typo in Audio dataset documentation

prassanna-ravishankar:
There is a typo in the section of the documentation dedicated to creating an audio dataset. The Dataset is incorrectly suffixed with a `Config`

https://huggingface.co/datasets/indonesian-nlp/librivox-indonesia/blob/main/librivox-indonesia.py#L59


mariosasko (approved):
Thanks, good catch!

(CI failure is unrelated)

PR #6223, opened 2023-09-07
Update README.md

NinoRisteski:
fixed a few typos

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -33,13 +33,13 @@ pip install git+https://github.com/huggingface/doc-builder
 **NOTE**
 
 You only need to generate the documentation to inspect it locally (if you're planning changes and want to
-check how they look before committing for instance). You don't have to commit to the built documentation.
+check how they look before committing for instance). You don't have to `git commit` the built documentation.
 
 ---
 
 ## Building the documentation
 
-Once you have set the `doc-builder` and additional packages, you can generate the documentation by typing th
+Once you have setup the `doc-builder` and additional packages, you can generate the documentation by typing
 the following command:
 
 ```bash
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -33,14 +33,14 @@ pip install git+https://github.com/huggingface/doc-builder
 **NOTE**
 
 You only need to generate the documentation to inspect it locally (if you're planning changes and want to
-check how they look like before committing for instance). You don't have to commit the built documentation.
+check how they look before committing for instance). You don't have to commit to the built documentation.
```
stevhliu, line 36: Might be better to revert this here to avoid changing the meaning of it (`git commit` the generated docs versus commit to it).

```suggestion
check how they look before committing for instance). You don't have to `git commit` the built documentation.
```

docs/README.md
```diff
@@ -33,14 +33,14 @@ pip install git+https://github.com/huggingface/doc-builder
...
 
 ---
 
 ## Building the documentation
 
-Once you have setup the `doc-builder` and additional packages, you can generate the documentation by typing th
-following command:
+Once you have set the `doc-builder` and additional packages, you can generate the documentation by typing th
+the following command:
```
stevhliu, line 43: I think setup was correct the way before

```suggestion
Once you have setup the `doc-builder` and additional packages, you can generate the documentation by typing
the following command:
```

docs/README.md
```diff
@@ -151,8 +151,8 @@ not to be displayed in the documentation, you can do so by specifying which meth
     - save_vocabulary
 ```
 
-If you just want to add a method that is not documented (for instance magic method like `__call__` are not documented
-byt default) you can put the list of methods to add in a list that contains `all`:
+If you just want to add a method that is not documented (for instance magic method like `__call__` is not documented
+by default) you can add the list of methods to add in a list that contains `all`:
```
stevhliu, line 155: I think it sounded better before to avoid using two `add`'s in the same sentence

```suggestion
by default) you can put the list of methods to add in a list that contains `all`:
```

stevhliu (approved):
Thanks for fixing the typos and clarification!

mariosasko (approved):
Nice! 

I took the liberty to commit the suggestions myself 🙂.

PR #6231, opened 2023-09-11
Overwrite legacy default config name in `dataset_infos.json` in packaged datasets

polinaeterna:
Currently if we push data as default config with `.push_to_hub` to a repo that has a legacy `dataset_infos.json` file containing a legacy default config name like `{username}--{dataset_name}`, new key `"default"` is added to `dataset_infos.json` along with the legacy one. I think the legacy one should be dropped in this case.

### DIFF
--- a/src/datasets/dataset_dict.py
+++ b/src/datasets/dataset_dict.py
@@ -1728,6 +1728,9 @@ class DatasetDict(dict):
             )
             with open(dataset_infos_path, encoding="utf-8") as f:
                 dataset_infos: dict = json.load(f)
+            legacy_config_name = "--".join(repo_id.split("/"))
+            if config_name == "default" and legacy_config_name in dataset_infos:
+                dataset_infos.pop(legacy_config_name)
             dataset_infos[config_name] = asdict(info_to_dump)
             buffer = BytesIO()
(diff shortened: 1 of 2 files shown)


polinaeterna:
realized that this pr is still not merged, @lhoestq maybe you can take a look at it?

lhoestq:
I think https://github.com/huggingface/datasets/pull/6218 fixed the issue (a bit differently though)

lhoestq:
ah actually nope, let me check

polinaeterna:
@lhoestq yeah the pr you're referencing doesn't fix the problem when two semantically analogous configs occur in datasets_info.json, i suggest to rewrite the legacy one if it exists during .push_to_hub

lhoestq:
(also minor: not checking the name makes it more robust to dataset renaming)

polinaeterna:
@lhoestq okay makes sense... so you think it's not a problem that in some cases we might end up with `dataset_infos.json` having two keys in it?

lhoestq:
> @lhoestq okay makes sense... so you think it's not a problem that in some cases we might end up with dataset_infos.json having two keys in it?

Ideally they should have only one config no ? Since old versions of `datasets` simply load the first config in the JSON.
We can overwrite it with the new default one (and no matter the name of the outdated config in the JSON)

PR #6232, opened 2023-09-11
Improve error message for missing function parameters

suavemint:
The error message in the fingerprint module was missing the f-string 'f' symbol, so the error message returned by fingerprint.py, line 469 was literally "function {func} is missing parameters {fingerprint_names} in signature."

This has been fixed.


mariosasko (approved):
Good catch! I replaced the `func.__name__` with `func.__qualname__` to be aligned with the function `__repr__`  (equivalent to `def __repr__(self):  return f"<function {self.__qualname__} {hex(id(self))}>"`)

mariosasko:
CI errors are unrelated

PR #6244, opened 2023-09-15
Add support for `fsspec>=2023.9.0`

### LINKED ISSUE #6214: Unpin fsspec < 2023.9.0
Once root issue is fixed, remove temporary pin of fsspec < 2023.9.0 introduced by:
- #6210

Related to issue:
- #6209

After investigation, I think the root issue is related to the new glob behavior with double asterisk `**` they have introduced in:
- https://github.com/fsspec/filesystem_spec/pull/1329

mariosasko:
Fix #6214

### DIFF
--- a/src/datasets/data_files.py
+++ b/src/datasets/data_files.py
@@ -9,6 +9,7 @@ import huggingface_hub
 from fsspec import get_fs_token_paths
 from fsspec.implementations.http import HTTPFileSystem
 from huggingface_hub import HfFileSystem
+from packaging import version
 from tqdm.contrib.concurrent import thread_map
 
 from . import config
@@ -42,9 +43,13 @@ SPLIT_KEYWORDS = {
     Split.TEST: ["test", "testing", "eval", "evaluation"],
 }
 NON_WORDS_CHARS = "-._ 0-9"
-KEYWORDS_IN_FILENAME_BASE_PATTERNS = ["**[{sep}/]{keyword}[{sep}]*", "{keyword}[{sep}]*"]
-KEYWORDS_IN_DIR_NAME_BASE_PATTERNS = ["{keyword}[{sep}/]**", "**[{sep}/]{keyword}[{sep}/]**"]
-
+KEYWORDS_IN_FILENAME_BASE_PATTERNS = ["**[{sep}/]{keyword}[{sep}]*"]
+KEYWORDS_IN_DIR_NAME_BASE_PATTERNS = ["**[{sep}/]{keyword}[{sep}/]**"]
+# KEYWORDS_IN_FILENAME_BASE_PATTERNS = ["**[{sep}/]{keyword}[{sep}]*"]
+# KEYWORDS_IN_DIR_NAME_BASE_PATTERNS = ["**[{sep}/]{keyword}[{sep}/]**"]
+# if config.FSSPEC_VERSION < version.parse("2023.9.0"):
+# KEYWORDS_IN_FILENAME_BASE_PATTERNS += ["{keyword}[{sep}]*"]
+# KEYWORDS_IN_DIR_NAME_BASE_PATTERNS += ["{keyword}[{sep}/]**"]
 DEFAULT_SPLITS = [Split.TRAIN, Split.VALIDATION, Split.TEST]
 DEFAULT_PATTERNS_SPLIT_IN_FILENAME = {
     split: [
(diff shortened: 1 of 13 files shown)


src/datasets/data_files.py
```diff
@@ -42,8 +43,12 @@ class EmptyDatasetError(FileNotFoundError):
...
 }
 NON_WORDS_CHARS = "-._ 0-9"
-KEYWORDS_IN_FILENAME_BASE_PATTERNS = ["**[{sep}/]{keyword}[{sep}]*", "{keyword}[{sep}]*"]
-KEYWORDS_IN_DIR_NAME_BASE_PATTERNS = ["{keyword}[{sep}/]**", "**[{sep}/]{keyword}[{sep}/]**"]
+if config.FSSPEC_VERSION < version.parse("2023.9.0"):
+    KEYWORDS_IN_FILENAME_BASE_PATTERNS = ["**[{sep}/]{keyword}[{sep}]*", "{keyword}[{sep}]*"]
+    KEYWORDS_IN_DIR_NAME_BASE_PATTERNS = ["{keyword}[{sep}/]**", "**[{sep}/]{keyword}[{sep}/]**"]
+else:
+    KEYWORDS_IN_FILENAME_BASE_PATTERNS = ["**/*[{sep}/]{keyword}[{sep}]*", "{keyword}[{sep}]*"]
+    KEYWORDS_IN_DIR_NAME_BASE_PATTERNS = ["{keyword}[{sep}/]**", "**/*[{sep}/]{keyword}[{sep}/]**"]
```
albertvillanova, line 51: With the current behavior (double asterisk only between slashes meaning zero or more path segments) I find confusing several patterns here:
- why do we add the slash to the sep characters here: `[{sep}/]`? I think it would be clearer to remove it as we already prefix the pattern with `**/`
- as they are:
  - the first pattern matches `dir/train.csv`, `my-train.csv` and `dir/my-train.csv`, 
  - whereas the second pattern only matches `train.csv`
- I think it would be clearer to use these patterns: `["**/*[{sep}]{keyword}[{sep}]*", "**/{keyword}[{sep}]*"]`, so that
  - the first pattern matches: `my-train.csv` and `dir/my-train.csv`
  - and the second: `train.csv` and `dir/train.csv`

albertvillanova (commented):
Thanks. Some comments below.

Maybe we should update the docstring of `get_data_patterns` accordingly? Currently it only gives examples of outputs with `**` not in a single path segment (i.e. not with a `/` as prefix or suffix)

tests/test_load.py
```diff
@@ -1168,6 +1168,9 @@ def test_load_dataset_zip_text(data_file, streaming, zip_text_path, zip_text_wit
 
 @pytest.mark.parametrize("streaming", [False, True])
 def test_load_dataset_arrow(streaming, data_dir_with_arrow):
+    print("#################")
+    print(data_dir_with_arrow)
+    print("#################")
```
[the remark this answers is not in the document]
mariosasko: Yup, a leftover print from debugging 😄.

src/datasets/data_files.py
albertvillanova: I see... they introduced new bugs... :confused:

albertvillanova (commented):
Thanks. Any comment on my comment below?

> Maybe we should update the docstring of `get_data_patterns` accordingly? Currently it only gives examples of outputs with `**` not in a single path segment (i.e. not with a `/` as prefix or suffix).

lhoestq (approved):
LGTM ! I think the fix is future proof so we should be fine

tests/test_data_files.py
```diff
@@ -493,15 +493,15 @@ def mock_fs(file_paths: List[str]):
     Example:
 
     ```py
-    >>> fs = mock_fs(["data/train.txt", "data.test.txt"])
+    >>> DummyTestFS = mock_fs(["data/train.txt", "data.test.txt"])
```
lhoestq, line 496:
```suggestion
    >>> DummyTestFS = mock_fs(["data/train.txt", "data.test.txt"])
    >>> fs = DummyTestFS()
```

lhoestq:
> Thanks. Any comment on my comment below?
> 
> >Maybe we should update the docstring of get_data_patterns accordingly? Currently it only gives examples of outputs with ** not in a single path segment (i.e. not with a / as prefix or suffix).

Yea maybe let's replace all mentions of "**" with "**/*" ? This way we encourage the use of "**/*" which is the one that is consistent with non-fsspec globs

lhoestq:
Also just noticed that `KEYWORDS_IN_DIR_NAME_BASE_PATTERNS` seems to include `KEYWORDS_IN_FILENAME_BASE_PATTERNS`. I guess we can try to remove the filename one in another PR to remove this redundancy

mariosasko:
I've addressed the comments. Let me know if it looks all good now :)

lhoestq:
Actually just found out that the current `**/*[-._ 0-9/]train[-._ 0-9/]**` doesn't match `data/train.csv` in bash.

So there might be a risk that this pattern breaks in the future no ?

mariosasko:
@lhoestq `fsspec` has tests to check their specific (non-posix) behavior, so I think merging in the current state is fine. And if  they make a breaking change in the future, we can align the patterns once again :)

PR #6247, opened 2023-09-18
Update create_dataset.mdx

EswarDivi:
modified , as AudioFolder and ImageFolder not in Dataset Library. 

``` from datasets import AudioFolder ```  and ```from datasets import ImageFolder```  to ```from datasets import load_dataset```

``` 
cannot import name 'AudioFolder' from 'datasets' (/home/eswardivi/miniconda3/envs/Hugformers/lib/python3.10/site-packages/datasets/__init__.py)

```


mariosasko (approved):
`ImageFolder` and `AudioFolder` are the builder classes' names not meant to be instantiated (directly), so thanks for fixing this!

PR #6251, opened 2023-09-20
Support streaming datasets with pyarrow.parquet.read_table

albertvillanova:
Support streaming datasets with `pyarrow.parquet.read_table`.

See: https://huggingface.co/datasets/uonlp/CulturaX/discussions/2

CC: @AndreaFrancis

### DIFF
--- a/src/datasets/download/streaming_download_manager.py
+++ b/src/datasets/download/streaming_download_manager.py
@@ -783,6 +783,16 @@ def xpandas_read_excel(filepath_or_buffer, download_config: Optional[DownloadCon
             )
 
 
+def xpyarrow_parquet_read_table(filepath_or_buffer, download_config: Optional[DownloadConfig] = None, **kwargs):
+    import pyarrow.parquet as pq
+
+    if hasattr(filepath_or_buffer, "read"):
+        return pq.read_table(filepath_or_buffer, **kwargs)
+    else:
+        filepath_or_buffer = str(filepath_or_buffer)
+        return pq.read_table(xopen(filepath_or_buffer, mode="rb", download_config=download_config), **kwargs)
+
+
 def xsio_loadmat(filepath_or_buffer, download_config: Optional[DownloadConfig] = None, **kwargs):
     import scipy.io as sio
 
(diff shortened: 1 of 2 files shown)


mariosasko:
This function reads an entire Arrow table in one go, which is not ideal memory-wise, so I don't think we should encourage using this function, considering we want to keep RAM usage as low as possible in the streaming mode. 

(Note that Parquet files are compressed, meaning the loaded table can be significantly larger than the size in Parquet.)

Instead, we should suggest the authors to use:
```python
with open(doc_path, "rb") as f:
    parquet_file = pq.ParquetFile(f)
    for batch in parquet_file.iter_batches():
        pa_table = pa.Table.from_batches([batch])
        yield idx, pa_table
        idx += 1
```

albertvillanova:
@mariosasko I think the potential problem you evoke is independent of whether or not we support streaming mode:
- if the user's script with `read_table` works in non-streaming mode, it will also work in streaming mode after this PR

In fact, what we should suggest instead is to follow the scriptless approach, so that our `parquet` packaged module is used, with all the optimizations implemented. But this approach is not possible in all cases, and some use cases need to implement a script. And if they have small Parquet files and use `read_table`, I think we should support streaming.

In summary, let me clarify the goal and the scope of this PR:
- a user needs using a loading script
- their files are small enough so that they use `read_table`
- their loading script works in non-streaming mode
- therefore, this PR allows loading their dataset in streaming mode as well

mariosasko:
Yes, the no-script approach with metadata configs makes the most sense.

> their files are small enough so that they use read_table

Some of the Parquet files in that repo are larger than 1 GB ...

Also, I'd wait for more instances of people using the `read_table` function on the Hub before merging this PR.

albertvillanova:
@mariosasko, yes, this solution is not specifically for the "uonlp/CulturaX" dataset, but for other use cases as I explained above: indeed, they finally removed the use of `read_table` because their data files are too large.

> Also, I'd wait for more instances of people using the `read_table` function on the Hub before merging this PR.

Do you know how many datasets are currently using `read_table`?

mariosasko:
> Do you know how many datasets are currently using read_table?

Zero (based on the script that checks the script contents of the public Hub datasets).

albertvillanova:
I see... Thanks! :hugs:

albertvillanova:
I'm merging this PR as discussed in private.

PR #6255, opened 2023-09-23
Parallelize builder configs creation

lhoestq:
For datasets with lots of configs defined in YAML

E.g. `load_dataset("uonlp/CulturaX", "fr", revision="refs/pr/6")` from >1min to 15sec


mariosasko (commented):
It looks good, but I think a cleaner (and faster) solution would be to resolve data files lazily (maybe rename `DataFilesList` to `DataFilesIterable` and resolve them when needed).

lhoestq:
Yea lazy data files resolution seems a better approach actually

Issue #4129, opened 2022-04-08
dataset metadata for reproducibility


nbroad1881:
When pulling a dataset from the hub, it would be useful to have some metadata about the specific dataset and version that is used. The metadata could then be passed to the `Trainer` which could then be saved to a model card. This is useful for people who run many experiments on different versions (commits/branches) of the same dataset. 

The dataset could have a list of “source datasets” metadata and ignore what happens to them before arriving in the Trainer (i.e. ignore mapping, filtering, etc.).

Here is a basic representation (made by @lhoestq )
```python
>>> from datasets import load_dataset
>>> 
>>> my_dataset = load_dataset(...)["train"]
>>> my_dataset = my_dataset.map(...)
>>> 
>>> my_dataset.sources
[HFHubDataset(repo_id=..., revision=..., arguments={...})]
```

davanstrien:
+1 on this idea. This could be powerful for helping better track datasets used for model training and help with automatic model card creation. 

One possible way of doing this would be to store some/most/all the arguments passed to `load_dataset` if a hub id is passed. i.e. store the Hub ID, configuration, etc. 

cc @tomaarsen

PR #6268, opened 2023-09-29
Add repo_id to DatasetInfo

lhoestq:
```python

ds = load_dataset("squad", split="validation[:3]")
ds = ds.map(lambda x: {}, num_proc=2).filter(lambda x: True).remove_columns(["question"])
print(ds.repo_id)
# squad
```

### DIFF
--- a/src/datasets/info.py
+++ b/src/datasets/info.py
@@ -153,6 +153,7 @@ class DatasetInfo:
     dataset_name: Optional[str] = None  # for packaged builders, to be different from builder_name
     config_name: Optional[str] = None
     version: Optional[Union[str, Version]] = None
+    repo_id: Optional[str] = None
     # Set later by `download_and_prepare`
     splits: Optional[dict] = None
     download_checksums: Optional[dict] = None
@@ -282,6 +283,12 @@ class DatasetInfo:
         supervised_keys = None
         task_templates = None
 
+        repo_ids = {dset_info.repo_id for dset_info in dataset_infos}
+        if len(repo_ids) == 1 and next(iter(repo_ids)) is not None:
+            repo_id = next(iter(repo_ids))
+        else:
+            repo_id = None
+
         # Find common task templates across all dataset infos
         all_task_templates = [info.task_templates for info in dataset_infos if info.task_templates is not None]
         if len(all_task_templates) > 1:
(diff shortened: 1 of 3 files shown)


lhoestq:
In https://github.com/huggingface/datasets/issues/4129 we want to track the origin of a dataset, e.g. if it comes from multiple datasets.

I think it's out of scope of DatasetInfo alone, which has info for one dataset only.
Therefore it makes sense to add repo_id, which is for one dataset only.

IMO if we want to track multiple origins we will need a new DatasetInfo that would have fields relevant to a mix of datasets (out of scope of this PR)

src/datasets/info.py
```diff
@@ -282,6 +283,12 @@ def from_merge(cls, dataset_infos: List["DatasetInfo"]):
         supervised_keys = None
         task_templates = None
 
+        repo_ids = {dset_info.repo_id for dset_info in dataset_infos}
+        if len(repo_ids) == 1:
+            repo_id = next(iter(repo_ids))
```
tomaarsen, line 288:
```suggestion
            repo_id = repo_ids.pop()
```
This is usually my goto here

davanstrien:
Really happy to see this! It could also be helpful to track some other metadata about how the dataset was built in the future. i.e. for the Stack loaded like this:

```
ds = load_dataset("bigcode/the-stack", data_dir="data/dockerfile", split="train")
```
It could be helpful to have easy access to the `data_dir` argument used during loading since that changes the training data quite a bit vs. loading the full dataset. You can also recover this from `download_checksums`, which seems a bit hacky. That is not necessary for this PR, though.

tomaarsen:
Perhaps it is also interesting to track the revision?

davanstrien:
One other thought. Is it worth tracking if a `token` was passed during loading? 

The Hub ID for private datasets could in some cases contain information someone wouldn't want to make public i.e. `davanstrien/super_secret_dataset_using_GPT_created_data`. 

Adding a bool like `is_private` could then be used by another library to determine if the dataset ID should be shared or not (or default to not sharing the ID for private datasets). i.e. in SpanMarker @tomaarsen might do a check like 

```python
if ds.is_private and not push_hub_id_for_private_ds:
	ds_name  = None
```
Potentially this is overkill but could be useful for downstream libraries who might use this information for creating automatic model cards.

mariosasko:
We should probably find a way to remove `DatasetInfo`, as (most of) its attributes are outdated (homepage, description, etc.), not introduce new ones :). But I guess storing `repo_id` there is the simplest solution for now, so I'm OK with it.

Issue #5430, opened 2023-01-17
Support Apache Beam >= 2.44.0


albertvillanova:
Once we find out the root cause of:
- #5426 

we should revert the temporary pin on apache-beam introduced by:
- #5429

albertvillanova:
Some of the shard files now have 0 number of rows.

We have opened an issue in the Apache Beam repo:
- https://github.com/apache/beam/issues/25041

PR #6283, opened 2023-10-05
Fix array cast/embed with null values

### LINKED ISSUE #6280: Couldn't cast array of type fixed_size_list to Sequence(Value(float64))
### Describe the bug

I have a dataset with an embedding column, when I try to map that dataset I get the following exception:

```
  File "/Users/jmif/.virtualenvs/llm-training/lib/python3.10/site-packages/datasets/arrow_dataset.py", line 3189, in map
    for rank, done, content in iflatmap_unordered(
  File "/Users/jmif/.virtualenvs/llm-training/lib/python3.10/site-packages/datasets/utils/py_utils.py", line 1387, in iflatmap_unordered
    [async_result.get(timeout=0.05) for async_result in async_results]
  File "/Users/jmif/.virtualenvs/llm-training/lib/python3.10/site-packages/datasets/utils/py_utils.py", line 1387, in <listcomp>
    [async_result.get(timeout=0.05) for async_result in async_results]
  File "/Users/jmif/.virtualenvs/llm-training/lib/python3.10/site-packages/multiprocess/pool.py", line 774, in get
    raise self._value
fixed_size_list<item: float>[2]
to
Sequence(feature=Value(dtype='float32', id=None), length=2, id=None)
```

Here's a simple repro script:

```
from datasets import Features, Value, Sequence, ClassLabel, Dataset

dataset_features = Features({
    'text': Value('string'),
    'embedding': Sequence(Value('double'), length=2),
    'categories': Sequence(ClassLabel(names=sorted([
        'one',
        'two',
        'three'
    ]))),
})

dataset = Dataset.from_dict(
    {
        'text': ['A'] * 10000,
        'embedding': [[0.0, 0.1]] * 10000,
        'categories': [[0]] * 10000,
    },
    features=dataset_features
)

def test_mapper(r):
    r['text'] = list(map(lambda t: t + ' b', r['text']))
    return r


dataset = dataset.map(test_mapper, batched=True, batch_size=10, features=dataset_features, num_proc=2)
```

Removing the embedding column fixes the issue!

### Expected behavior

The mapping completes successfully.

### Environment info

- `datasets` version: 2.14.4
- Platform: macOS-14.0-arm64-arm-64bit
- Python version: 3.10.12
- Huggingface_hub version: 0.17.1
- PyArrow version: 13.0.0
- Pandas version: 2.0.3

### LINKED ISSUE #6311: cast_column to Sequence with length=4 occur exception raise in datasets/table.py:2146
### Describe the bug

i load a dataset from local csv file which has 187383612 examples, then use `map` to generate new columns for test.
here is my code : 
`
import os

from datasets.features import Sequence, Value

def add_new_path(example):

    example["ais_bbox"] = [100,100,200,200]
    example["ais_image_path"] = os.path.join("images", example["image_path"]) if example["image_path"] else ""
    return example

ais_dataset = load_dataset("/data/ryan.gao/ais_dataset_cache/raw/1749/")
hf_ds = ais_dataset.map(add_new_path, batched=False, num_proc=32)
ds = hf_ds.cast_column("ais_bbox", Sequence(Value("int32"), length=4))
`
and the `cast_column` raise an exception
`
Casting the dataset:   3%|███▉
...
  File "/home/protoss.gao/.local/lib/python3.9/site-packages/datasets/arrow_dataset.py", line 2110, in cast_column
    return self.cast(features)
  File "/home/protoss.gao/.local/lib/python3.9/site-packages/datasets/arrow_dataset.py", line 2055, in cast
    dataset = dataset.map(
  File "/home/protoss.gao/.local/lib/python3.9/site-packages/datasets/arrow_dataset.py", line 592, in wrapper
  File "/home/protoss.gao/.local/lib/python3.9/site-packages/datasets/arrow_dataset.py", line 557, in wrapper
  File "/home/protoss.gao/.local/lib/python3.9/site-packages/datasets/arrow_dataset.py", line 3097, in map
    for rank, done, content in Dataset._map_single(**dataset_kwargs):
  File "/home/protoss.gao/.local/lib/python3.9/site-packages/datasets/arrow_dataset.py", line 3474, in _map_single
    batch = apply_function_on_filtered_inputs(
  File "/home/protoss.gao/.local/lib/python3.9/site-packages/datasets/arrow_dataset.py", line 3353, in apply_function_on_filtered_inputs
    processed_inputs = function(*fn_args, *additional_args, **fn_kwargs)
  File "/home/protoss.gao/.local/lib/python3.9/site-packages/datasets/table.py", line 2329, in table_cast
    return cast_table_to_schema(table, schema)
  File "/home/protoss.gao/.local/lib/python3.9/site-packages/datasets/table.py", line 2288, in cast_table_to_schema
    arrays = [cast_array_to_feature(table[name], feature) for name, feature in features.items()]
  File "/home/protoss.gao/.local/lib/python3.9/site-packages/datasets/table.py", line 2288, in <listcomp>
    arrays = [cast_array_to_feature(table[name], feature) for name, feature in features.items()]
  File "/home/protoss.gao/.local/lib/python3.9/site-packages/datasets/table.py", line 1831, in wrapper
    return pa.chunked_array([func(chunk, *args, **kwargs) for chunk in array.chunks])
  File "/home/protoss.gao/.local/lib/python3.9/site-packages/datasets/table.py", line 1831, in <listcomp>
    return pa.chunked_array([func(chunk, *args, **kwargs) for chunk in array.chunks])
  File "/home/protoss.gao/.local/lib/python3.9/site-packages/datasets/table.py", line 2145, in cast_array_to_feature
    raise TypeError(f"Couldn't cast array of type\n{array.type}\nto\n{feature}")
list<item: int64>
to
Sequence(feature=Value(dtype='int32', id=None), length=4, id=None)
`
i check the source code and make debug info:
in datasets/table.py:2092
`
2091             if feature.length > -1:
2092                 if feature.length * len(array) == len(array.values):
2093                     return pa.FixedSizeListArray.from_arrays(_c(array.values, feature.feature), feature.length)
2094                 print(len(array))
2095                 print(len(array.values))
`
my feature.length is 4.  but feature.length * len(array) == len(array.values) is false.
print(len(array)) is 262
print(len(array.values)) is 4000

then I use "for item in array" to print each item then see 262 of [100,100,200,200]
and use "for item in array.values" to print each item and get 4000 int32. which seems 1000 * [100,100,200,200]

i'm wondering the chunk in each array.chunk, the "chunk.values" may get all the chunks's value rather than single chunk? but i check the pyarrow's doc seems chunk.values is chunk's value not all.

code provided above.

### Expected behavior

feature.length * len(array) == len(array.values) should be true. and there should not has Exception.

### Environment info

python3.9
x86_64
datasets: 2.14.4
pyarrow: 13.0.0 or 10.0.0

### LINKED ISSUE #6360: Add support for `Sequence(Audio/Image)` feature in `push_to_hub`
### Feature request

Allow for `Sequence` of `Image` (or `Audio`) to be embedded inside the shards.

### Motivation

Currently, thanks to #3685, when `embed_external_files` is set to True (which is the default) in `push_to_hub`, features of type `Image` and `Audio` are embedded inside the arrow/parquet shards, instead of only storing paths to the files.

I've noticed that this behavior does not extend to `Sequence` of `Image`, when working with a [dataset of timelapse images](https://huggingface.co/datasets/1aurent/Human-Embryo-Timelapse).

### Your contribution

I'll submit a PR if I find a way to add this feature

mariosasko:
Fixes issues with casting/embedding PyArrow list arrays with null values. It also bumps the required PyArrow version to 12.0.0 (over 9 months old) to simplify the implementation.

Fix #6280, fix #6311, fix #6360 

(Also fixes https://github.com/huggingface/datasets/issues/5430 to make Beam compatible with PyArrow>=12.0.0)


mariosasko:
CI failures are unrelated

lhoestq (approved):
Thanks for the fix !

mariosasko:
I also plan to address https://github.com/huggingface/datasets/issues/6280#issuecomment-1749310065 in this PR :).

lhoestq:
Oh ok, ping me again whenever you want another review :)

mariosasko:
@lhoestq Feel free to review this again. I've bumped PyArrow to 12.0.0 to simplify the implementation (no need for custom `array_concat` and less `pa.Array.from_buffers`). However, this makes `apache-beam` complain as it only supports `<12.0.0`. The next `apache-beam` release will set this boundary to `<15.0.0.`, so I think the only solution is to wait for it to be published.

src/datasets/table.py
```diff
@@ -1941,44 +1870,59 @@ def array_cast(array: pa.Array, pa_type: pa.DataType, allow_number_to_str=True):
...
-                    )
+            if _are_list_values_of_length(array, pa_type.list_size):
+                if array.null_count > 0:
+                    # Ensure each null value in the array translates to [null] * pa_type.list_size in the array's values array
+                    array_type = array.type
+                    storage_type = _storage_type(array_type)
+                    if array_type != storage_type:
+                        # Temporarily convert to the storage type to support extension types in the slice operation
+                        array = _c(array, storage_type)
+                        array = pc.list_slice(array, 0, pa_type.list_size, return_fixed_size_list=True)
```
lhoestq, line 1881: this may bring the data in memory no ?

maybe it's fine for now though

src/datasets/table.py
```diff
@@ -2021,6 +1965,9 @@ def cast_array_to_feature(array: pa.Array, feature: "FeatureType", allow_number_
     """
     from .features.features import Sequence, get_nested_type
 
+    # if isinstance(feature, Sequence):
+    #     print("array off", array.offset)
+
```
lhoestq, line 1970:
```suggestion
```

lhoestq (approved):
Nice ! Handling extension types can be quite complicated

Btw if you have some pyarrow issues we can link to this PR feel free to add them, this way we can follow the advancements and maybe later simplify this code

src/datasets/table.py
mariosasko: Yes, `pc.list_slice` brings the array in memory. Unfortunately, I don't think it's possible to avoid this as we need to "expand" the `null` values in the `.values` array to prepare them for the `ListArray` -> `FixedSizeListArray` cast, which requires a memory allocation.

We run these casts on subtables before writing them to disk, so this solution should be fine for now.

mariosasko: Casting `ListArray` -> `FixedSizeListArray` using PyArrow's `ListArray.cast` also allocates memory if the array contains `null` lists, so indeed there isn't much we can do about this due to the difference in the null values storage layout.

20141888:
Still the problem is occured.
Huggingface is sucks 🤮🤮🤮🤮

PR #6286, opened 2023-10-09
Create DefunctDatasetError

albertvillanova:
Create `DefunctDatasetError` as a specific error to be raised when a dataset is defunct and no longer accessible.


mariosasko (approved):
LGTM, thanks!

Issue #6293, opened 2023-10-11
Choose columns to stream parquet data in streaming mode


lhoestq:
Currently passing columns= to load_dataset in streaming mode fails

```
Tried to load parquet data with columns '['link']' with mismatching features '{'caption': Value(dtype='string', id=None), 'image': {'bytes': Value(dtype='binary', id=None), 'path': Value(dtype='null', id=None)}, 'link': Value(dtype='string', id=None), 'message_id': Value(dtype='string', id=None), 'timestamp': Value(dtype='string', id=None)}'
```

similar to https://github.com/huggingface/datasets/issues/6039

reported at https://huggingface.co/datasets/laion/dalle-3-dataset/discussions/3#65259a09617407d4520f4ad9

PR #6295, opened 2023-10-11
Fix parquet columns argument in streaming mode

lhoestq:
It was failing when there's a DatasetInfo with non-None info.features from the YAML (therefore containing columns that should be ignored)

Fix https://github.com/huggingface/datasets/issues/6293


lhoestq:
CI errors are unrelated or due to flaky tests

mariosasko (approved):

PR #6300, opened 2023-10-12
Unpin `jax` maximum version

### LINKED ISSUE #6202: avoid downgrading jax version
### Feature request

Whenever I `pip install datasets[jax]` it downgrades jax to version 0.3.25. I seem to be able to install this library first then upgrade jax back to version 0.4.13.

### Motivation

It would be nice to not overwrite currently installed version of jax if possible.

### Your contribution

I would be willing to beta test. Or maybe write some code if I could get pointed in the right direction, I'm not super familiar with this codebase.

### LINKED ISSUE #6299: Support for newer versions of JAX
### Feature request

Hi,

I like your idea of adapting the datasets library to be usable with JAX. Thank you for that.

However, in your [setup.py](https://github.com/huggingface/datasets/blob/main/setup.py), you enforce old versions of JAX <= 0.3... It is very cumbersome !

What is the rationale for such a limitation ? Can you remove it please ?

Thanks,

### Motivation

This library is unusable with new versions of JAX ?

### Your contribution

Yes.

mariosasko:
fix #6299
fix #6202

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -165,8 +165,8 @@ TESTS_REQUIRE = [
     "apache-beam>=2.26.0,<2.44.0;python_version<'3.10'",  # doesn't support recent dill versions for recent python versions
     "elasticsearch<8.0.0",  # 8.0 asks users to provide hosts or cloud_id when instantiating ElasticSearch()
     "faiss-cpu>=1.6.4",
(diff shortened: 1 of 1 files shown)


mariosasko:
CI failures are unrelated

PR #6312, opened 2023-10-18
docs: resolving namespace conflict, refactored variable 

smty2018:
In docs of about_arrow.md, in the below example code
image
The variable name 'time' was being used in a way that could potentially lead to a namespace conflict with Python's built-in 'time' module. It is not a good convention and can lead to unintended variable shadowing for any user re-using the example code.
To ensure code clarity, and prevent potential naming conflicts renamed the variable 'time' to 'elapsed_time' in the example  code.


stevhliu (approved):
Thanks for improving the clarity! Pinging @lhoestq for a final look :)

lhoestq (approved):

PR #6316, opened 2023-10-19
Fix loading Hub datasets with CSV metadata file

### LINKED ISSUE #6315: Hub datasets with CSV metadata raise ArrowInvalid: JSON parse error: Invalid value. in row 0
When trying to load a Hub dataset that contains a CSV metadata file, it raises an `ArrowInvalid` error:
```
E   pyarrow.lib.ArrowInvalid: JSON parse error: Invalid value. in row 0

pyarrow/error.pxi:100: ArrowInvalid
```

See: https://huggingface.co/datasets/lukarape/public_small_papers/discussions/1

albertvillanova:
Currently, the reading of the metadata file infers the file extension (.jsonl or .csv) from the passed filename. However, downloaded files from the Hub don't have file extension. For example:
- the original file: `hf://datasets/__DUMMY_TRANSFORMERS_USER__/test-dataset-5916a4-16977085077831/metadata.jsonl`
- corresponds to the downloaded path: `/tmp/pytest-of-username/pytest-46/cache/datasets/downloads/9f5374dbb470f711f6b89d66a5eec1f19cc96324b26bcbebe29138bda6cb20e6`, which does not have extension

In the case where the metadata file does not have an extension, the reader assumes it is a JSONL file, thus the reported error when trying to read a CSV file as a JSONL one: `ArrowInvalid: JSON parse error: Invalid value. in row 0`

This behavior was introduced by:
- #4837

This PR extracts the metadata file extension from the original filename (instead of the downloaded one) and passes it as a parameter to the read_metadata function.

Fix #6315.

### DIFF
--- a/src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
+++ b/src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
@@ -164,15 +164,15 @@ class FolderBasedBuilder(datasets.GeneratorBasedBuilder):
 
             # Check that all metadata files share the same format
             metadata_ext = {
-                os.path.splitext(downloaded_metadata_file)[1][1:]
-                for _, downloaded_metadata_file in itertools.chain.from_iterable(metadata_files.values())
+                os.path.splitext(original_metadata_file)[-1]
+                for original_metadata_file, _ in itertools.chain.from_iterable(metadata_files.values())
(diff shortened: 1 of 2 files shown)


mariosasko (approved):
Thanks!

Surprised this bug was not reported earlier.

PR #6318, opened 2023-10-19
Deterministic set hash

lhoestq:
Sort the items in a set according to their `datasets.fingerprint.Hasher.hash` hash to get a deterministic hash of sets.

This is useful to get deterministic hashes of tokenizers that use a trie based on python sets.

reported in https://github.com/huggingface/datasets/issues/3847


mariosasko (approved):
Thanks, LGTM!

PR #6334, opened 2023-10-23
datasets.filesystems: fix is_remote_filesystems

ap--:
Close #6330

`fsspec.implementations.LocalFilesystem.protocol`
was changed from `str` "file" to `tuple[str,...]` ("file", "local") in `fsspec>=2023.10.0`

This commit supports both styles.

PR #6335, opened 2023-10-23
Support fsspec 2023.10.0

### LINKED ISSUE #6333: Support fsspec 2023.10.0
Once root issue is fixed, remove temporary pin of fsspec < 2023.10.0 introduced by:
- #6331

Related to issue:
- #6330

As @ZachNagengast suggested, the issue might be related to:
- https://github.com/fsspec/filesystem_spec/pull/1381

albertvillanova:
Fix #6333.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -126,7 +126,7 @@ REQUIRED_PKGS = [
     "multiprocess",
     # to save datasets locally or on any filesystem
     # minimum 2023.1.0 to support protocol=kwargs in fsspec's `open`, `get_fs_token_paths`, etc.: see https://github.com/fsspec/filesystem_spec/pull/1143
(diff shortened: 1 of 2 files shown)


lhoestq:
I think https://github.com/huggingface/datasets/pull/6334 fixes it already no ?

albertvillanova:
Yes, @lhoestq, you are right. I think we cross-send fixing PRs in a 15 minute interval... :sweat_smile: 

I would say the code in this PR is simpler and easier to understand, but feel free to ignore it.

lhoestq:
I think the correct way it to check if "file" in in the tuple if it's a tuple (in case someone adds another protocol name for the local filesystem)

PR #6337, opened 2023-10-23
Pin supported upper version of fsspec

albertvillanova:
Pin upper version of `fsspec` to avoid disruptions introduced by breaking changes (and the need of urgent patch releases with hotfixes) on each release on their side. See:
- #6331
- #6210 
- #5731
- #5617
- #5447

I propose that we explicitly test, introduce fixes and support each new `fsspec` version release.

CC: @LysandreJik


lhoestq:
In particular I expect fsspec to do another breaking change in the next release (switch to glob.glob)

albertvillanova:
Yes, unfortunately breaking changes are quite usual from their part.

lhoestq (approved):

PR #6351, opened 2023-10-25
Fix use_dataset.mdx

angel-luis:
The current example isn't working because it can't find `labels` inside the Dataset object. So I've added an extra step to the process. Tested and working in Colab.

### DIFF
--- a/docs/source/use_dataset.mdx
+++ b/docs/source/use_dataset.mdx
@@ -64,19 +64,14 @@ These values are actually the model inputs.
 >>> dataset = dataset.map(tokenization, batched=True)
 ```
 
-***4**. Add the `labels` key with the current label info to match the columns of the current model:
-```py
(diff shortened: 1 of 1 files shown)


docs/source/use_dataset.mdx
```diff
@@ -64,7 +64,12 @@ These values are actually the model inputs.
 >>> dataset = dataset.map(tokenization, batched=True)
 ```
 
-**4**. Set the format of your dataset to be compatible with your machine learning framework:
+***4**. Add the `labels` key with the current label info to match the columns of the current model:
+```py
+>>> dataset = dataset.map(lambda example: {"labels": example["label"]}, batched=True)
+```
+
```
mariosasko, line 71: We can avoid introducing a new step by replacing `labels` mentions with `label` in the existing ones.

mariosasko (commented):
Good catch!

mariosasko (approved):
I made the changes myself. Looks good now. Thanks!

PR #6355, opened 2023-10-26
More hub centric docs

lhoestq:
Let's have more hub-centric documentation in the datasets docs

Tutorials
- Add “Configure the dataset viewer” page
- Change order:
  - Overview
    - and more focused on the Hub rather than the library
  - Then all the hub related things
    - and mention how to read/write with other tools like pandas
  - Then all the datasets lib related things in a subsection

Also:
- Rename “know your dataset” page to “Explore your dataset”
- Remove “Evaluate Predictions” page since it's 'evaluate' stuff (or move to legacy section ?)

TODO:
- [ ] write the “Configure the dataset viewer” page


docs/source/upload_dataset.mdx
```diff
@@ -87,29 +87,51 @@ pip install huggingface_hub
...
+    >>> df = pd.read_csv("hf://datasets/stevhliu/demo/train.csv")
+    # df = dataset.apply(...)  # do all your processing here
+    >>> df.write_csv("hf://datasets/stevhliu/processed_demo/train.csv")
+    ```
+
+    3. Use the `huggingface-cli`
 
-To add a new configuration (or subset) to a dataset or to add a new split (train/validation/test), please refer to the [`Dataset.push_to_hub`] documentation.
+    ```bash
+    huggingface-cli upload --repo_ype dataset stevhliu/processed_demo local/path/to/train.csv remote/path/to/train.csv
```
severo, line 127:
```suggestion
    huggingface-cli upload --repo_type dataset stevhliu/processed_demo local/path/to/train.csv remote/path/to/train.csv
```

julien-c (commented):
love this!

maybe tagging @Wauplin for a quick review too

docs/source/upload_dataset.mdx
mariosasko, line 127: Maybe https://huggingface.co/docs/datasets/upload_dataset#upload-dataset is a better place for this (can be mentioned as a (programmatic) alternative to the UI workflow)

docs/source/upload_dataset.mdx
```diff
@@ -87,29 +87,51 @@ pip install huggingface_hub
...
+    ```
 
 ### Privacy
 
 A private dataset is only accessible by you. Similarly, if you share a dataset within your organization, then members of the organization can also access the dataset.
 
-Load a private dataset by providing your authentication token to the `token` parameter:
+Load a private dataset by providing your authentication token to the `token` parameter or using `huggingface-cli login`:
```
mariosasko, line 134: We should probably suggest using `huggingface-cli` as a preferred way of logging in (so maybe swap the order). It's easier to leak the `token` when using other methods (as mentioned in the [huggingface_hub](https://huggingface.co/docs/huggingface_hub/quick-start#login) docs).

docs/source/metrics.mdx
mariosasko, line 1: I think we can remove this once we remove the metrics API

mariosasko (commented):
I think our docs should primarily focus on the `datasets` lib (we can rename them to make this more obvious 🙂), so I'm not a fan of polluting them with "Hub-centric" stuff. 

Instead, it feels more natural to put all this info in the [Hub docs](https://huggingface.co/docs/hub/datasets) (some of it is already present there, so this would avoid **duplicating** it), and have links to it in the `datasets` docs.

(I forgot that the Hub docs exist when I suggested using our docs for showcasing the Datasets Server/Viewer features.)

lhoestq:
Yea I think some of it should be in the Hub docs indeed, let me open a new PR there.

Then I'll update the `datasets` docs anyway to avoid redundant stuff and add redirects instead

PR #6362, opened 2023-10-27
Simplify filesystem logic

mariosasko:
Simplifies the existing filesystem logic (e.g., to avoid unnecessary if-else as mentioned in https://github.com/huggingface/datasets/pull/6098#issue-1827655071)

### DIFF
--- a/src/datasets/arrow_writer.py
+++ b/src/datasets/arrow_writer.py
@@ -16,6 +16,7 @@ import errno
 import json
 import os
 import sys
+from pathlib import Path
 from typing import Any, Dict, Iterable, List, Optional, Tuple, Union
 
 import fsspec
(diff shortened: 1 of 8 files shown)


lhoestq:
Cool ! Nice to simplify this

src/datasets/arrow_writer.py
```diff
@@ -671,8 +672,9 @@ def finalize(self, metrics_query_result: dict):
 
         from .utils import beam_utils
 
+        # Beam FileSystems require the system's path separator in the older versions
         shards_metadata = list(
-            beam.io.filesystems.FileSystems.match([self._parquet_path + "*.parquet"])[0].metadata_list
+            beam.io.filesystems.FileSystems.match([str(Path(self._parquet_path)) + "*.parquet"])[0].metadata_list
```
lhoestq, line 677: `Path` replaces `://` with `:/` no ? This can be an issue for those writing datasets in a cloud storage

```python
>>> from pathlib import Path
>>> Path("gs://foo")
PosixPath('gs:/foo')
```

Maybe do the change only if it's a local path ?

lhoestq (approved):
Nice ! Just one comment regarding Beam stuff:

src/datasets/arrow_writer.py
mariosasko: Fixed :)

PR #6368, opened 2023-10-31
Fix python formatting for complex types in `format_table`

### LINKED ISSUE #6366: with_format() function returns bytes instead of PIL images even when image column is not part of "columns"
### Describe the bug

When using the with_format() function on a dataset containing images, even if the image column is not part of the columns provided in the function, its type will be changed to bytes.
Here is a minimal reproduction of the bug:
https://colab.research.google.com/drive/1hyaOspgyhB41oiR1-tXE3k_gJCdJUQCf?usp=sharing

1. Load the image dataset
2. apply with_format(columns=["text"])
3. Check the type of images in the "image" column before and after applying with_format

### Expected behavior

The type should stay the same, but it does not

### Environment info

datasets==2.14.6

mariosasko:
Fix #6366

### DIFF
--- a/src/datasets/formatting/formatting.py
+++ b/src/datasets/formatting/formatting.py
@@ -624,7 +624,7 @@ def format_table(
     else:
         pa_table = table
     query_type = key_to_query_type(key)
-    python_formatter = PythonFormatter(features=None)
+    python_formatter = PythonFormatter(features=formatter.features)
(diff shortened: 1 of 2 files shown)


lhoestq:
Thanks for the fix ! It was probably my mistake (forgot to re-apply the features)

PR #6391, opened 2023-11-08
Webdataset dataset builder

lhoestq:
Allow `load_dataset` to support the Webdataset format.

It allows users to download/stream data from local files or from the Hugging Face Hub.

Moreover it will enable the Dataset Viewer for Webdataset datasets on HF.

## Implementation details

- I added a new Webdataset builder
- dataset with TAR files are now read using the Webdataset builder
- Basic decoding from `webdataset` is used by default, except unsafe ones like pickle
- HF authentication support is done by registering a `webdataset.gopen` reader
- `webdataset` uses buffering when reading files, so I had to add buffering support in `xopen`

## TODOS

- [x] tests
- [x] docs


src/datasets/packaged_modules/webdataset/webdataset.py
```diff
@@ -0,0 +1,195 @@
...
+logger = datasets.utils.logging.get_logger(__name__)
+
+
+class Webdataset(datasets.GeneratorBasedBuilder):
+    DEFAULT_WRITER_BATCH_SIZE = 100
+    IMAGE_EXTENSIONS: List[str]  # definition at the bottom of the script
+    ENABLED_BASIC_HANDLERS: List[str]  # definition at the bottom of the script
+
+    def _basic_handlers(self, key, data):
+        from webdataset.autodecode import decoders
```
mariosasko, line 20: The format is simple enough to avoid this dependency, no?

src/datasets/packaged_modules/__init__.py
```diff
@@ -48,6 +50,7 @@ def _hash_python_lines(lines: List[str]) -> str:
     ".parquet": ("parquet", {}),
     ".arrow": ("arrow", {}),
     ".txt": ("text", {}),
+    ".tar": ("webdataset", {}),
```
mariosasko, line 53: Maybe we can make the module inference more robust by inspecting the contents of TAR archives (e.g., consecutive files with the same name (stem) but different extensions)

mariosasko (commented):
Thanks for working on this!

A high-level review :)

src/datasets/packaged_modules/__init__.py
lhoestq: Some webdatasets may contain only one field (e.g. only images) so I'm not sure it would make sense.

Also I like the idea of keeping the TAR loading simple and only support webdataset for TAR

src/datasets/packaged_modules/webdataset/webdataset.py
lhoestq: The dependency on webdataset or on webdataset/autoencode.decoders in particular ?

mariosasko: On `webdataset`

src/datasets/packaged_modules/__init__.py
mariosasko: But this would break some existing repos on the Hub, no?

lhoestq: TAR has never been supported on the Hub, what would it break ?

lhoestq: Ah you mean that some TAR datasets not in webdataset format won't work properly.

Maybe I can add an error message if the first webdataset examples don't have the same type

src/datasets/packaged_modules/webdataset/webdataset.py
lhoestq: As a first step I think it's fine and will be easy to maintain. It's also a way to make sure we support webdataset completely (even in the future, as long as they don't move their decoders)

lhoestq:
I added an error message if the first examples don't appear to be in webdataset format
```
"The TAR archives of the dataset should be in Webdataset format, "
"but the files in the archive don't share the same prefix or the same types."
```

src/datasets/packaged_modules/webdataset/webdataset.py
mariosasko: I think it's nice to avoid introducing new dependencies if possible.

lhoestq: It's optional and pretty lightweight, and as a user I think it's ok to require `webdataset` to read Webdataset data. Do you have concerns regarding breaking changes they may have ? or for user experience ?

docs/source/image_dataset.mdx
```diff
@@ -148,6 +148,46 @@ Upload your dataset with [`~datasets.DatasetDict.push_to_hub`]:
...
+ede6e66b2fb59aab.jpg
+ede6e66b2fb59aab.json
+ed600d57fcee4f94.jpg
+ed600d57fcee4f94.json
+...
+```
+
+You can put your images labels/captions/bounding boxes using JSON or text files for example.
+
+Fore more details on the Webdataset format and the python library, please check the [Webdataset documentation](https://webdataset.github.io/webdataset).
```
severo, line 179:
```suggestion
For more details on the Webdataset format and the python library, please check the [Webdataset documentation](https://webdataset.github.io/webdataset).
```

lhoestq:
@mariosasko could you review this ? I think it's fine to have webdataset as an optional dependency for now, then depending on usage and user feedbacks see if it makes sense to have our own implementation or not

src/datasets/packaged_modules/webdataset/webdataset.py
mariosasko: IMO it makes sense to put some effort into keeping dependencies at minimum if possible. 

An optional dependency still requires a `pip install` call (and restarting a notebook's env), so it's not the best user experience.

src/datasets/packaged_modules/webdataset/webdataset.py
```diff
@@ -0,0 +1,211 @@
...
+        first_examples = list(islice(pipeline, 5))
+        if any(example.keys() != first_examples[0].keys() for example in first_examples):
+            raise ValueError(
+                "The TAR archives of the dataset should be in Webdataset format, "
+                "but the files in the archive don't share the same prefix or the same types."
+            )
+        inferred_arrow_schema = pa.Table.from_pylist(first_examples[:1]).schema
+        features = datasets.Features.from_arrow_schema(inferred_arrow_schema)
+
+        # Set Image types
```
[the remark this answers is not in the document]
lhoestq: Yes correct, that's for another PR :p

src/datasets/packaged_modules/webdataset/webdataset.py
lhoestq: Ah yes and it's not even installed in colab by default.
Let me see if I can re-use their decoders code (webdataset license is BSD-3-Clause license so it's fine)

docs/source/loading.mdx
```diff
@@ -224,6 +224,31 @@ For more details, check out the [how to load tabular datasets from SQL databases
...
+
+To load remote Webdatasets via HTTP, pass the URLs instead:
+
+```python
+>>> from datasets import load_dataset
+>>>
+>>> base_url = "http://storage.googleapis.com/nvdata-openimages/openimages-train-{i:06d}.tar"
+>>> urls = [base_url.format(i=i) for i in range(10)]
+>>> dataset = load_dataset("webdataset", data_files={"train": urls}, split="train", streaming=True)
+```
```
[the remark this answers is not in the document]
lhoestq: I changed to a Hub HTTP URL

src/datasets/packaged_modules/webdataset/webdataset.py
```diff
@@ -0,0 +1,211 @@
...
+        for split_name, files in data_files.items():
+            if isinstance(files, str):
+                files = [files]
+            splits.append(datasets.SplitGenerator(name=split_name, gen_kwargs={"files": files}))
+
+        # Get one example to get the feature types
+        pipeline = wds.DataPipeline(
+            wds.SimpleShardList(files[:1]), wds.tarfile_to_samples(), wds.decode(post=[self._basic_handlers])
+        )
+        first_examples = list(islice(pipeline, 5))
```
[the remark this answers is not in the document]
lhoestq: done !

lhoestq:
I just removed the dependency on `webdataset` @mariosasko :)

src/datasets/packaged_modules/__init__.py
mariosasko: Nevermind, I thought this could fail on https://github.com/huggingface/datasets/blob/27d1fe52857c6a25a29cac63a296405136b2797c/src/datasets/load.py#L449-L450 

But this logic only inspects ZIP (and ignores TAR) archives :)

src/datasets/download/streaming_download_manager.py
```diff
@@ -492,8 +492,10 @@ def xopen(file: str, mode="r", *args, download_config: Optional[DownloadConfig]
     # add headers and cookies for authentication on the HF Hub and for Google Drive
     file, storage_options = _prepare_path_and_storage_options(file_str, download_config=download_config)
     kwargs = {**kwargs, **(storage_options or {})}
+    if buffering > 1:
+        kwargs["blocksize"] = buffering
```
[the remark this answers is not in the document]
lhoestq: I first added this to be able to register `xopen` in the `webdataset` `gopen` registry which expects a function with the same signature as `open` (open's third positional argument is `buffering` which can be an integer that corresponds to fsspec blocksiz), but this is unecessary now so I'll remove it

mariosasko (approved):
Nice work, LGTM!

yuvalkirstain:
Can we use it to read tars from disk as well? if so, how can this be done?

PR #6398, opened 2023-11-10
Remove redundant condition in builders

albertvillanova:
Minor refactoring to remove redundant condition.


mariosasko (approved):
Good catch!

PR #6416, opened 2023-11-14
Rename audio_classificiation.py to audio_classification.py

carlthome:


lhoestq:
Oh good catch. Can you also rename it in `src/datasets/tasks/__init__.py` ?

carlthome:
Fixed! 

(I think, tough word to spell right TBH)

lhoestq (approved):
Thanks :)

Issue #6400, opened 2023-11-10
Safely load datasets by disabling execution of dataset loading script


irenedea:
### Feature request

Is there a way to disable execution of dataset loading script using `load_dataset`? This is a security vulnerability that could lead to arbitrary code execution. 

Any suggested workarounds are welcome as well. 

### Motivation

This is a security vulnerability that could lead to arbitrary code execution. 

### Your contribution

n/a

julien-c:
great idea IMO

dakinggg:
@julien-c that would be great!

lhoestq:
We added the `trust_remote_code` argument to `load_dataset()` in `datasets` 2.16:
- in the future users will have to pass trust_remote_code=True to use datasets with a script
- for now we just show a warning when a dataset script is used
- we fallback on the Hugging Face Parquet exports when possible (to keep compatibility with old datasets with scripts)

So feel free to use `trust_remote_code=False` in the meantime to disable loading from dataset loading scripts :)

lhoestq:
Passing `trust_remote_code=True` explicitly is now mandatory to load a dataset with a script since https://github.com/huggingface/datasets/pull/6954

PR #6429, opened 2023-11-16
Add trust_remote_code argument

lhoestq:
Draft about adding `trust_remote_code` to `load_dataset`.

```python
ds = load_dataset(..., trust_remote_code=True)  # run remote code (current default)
```

It would default to `True` (current behavior) and in the next major release it will prompt the user to check the code before running it (we'll communicate on this before doing it of course).

```python
# in the future
ds = load_dataset(...)  # prompt the user to check the code before running it (future default)
ds = load_dataset(..., trust_remote_code=True)  # run remote code
ds = load_dataset(..., trust_remote_code=False)  # disallow remote code
```

Related to https://github.com/huggingface/datasets/issues/6400

Will do a separate PR to use the parquet export when possible


src/datasets/load.py
```diff
@@ -88,6 +89,45 @@
...
 
 
+def _raise_timeout_error(signum, frame):
+    raise ValueError(
+        "Loading this dataset requires you to execute custom code contained in the dataset repository on your local "
+        "machine. Please set the option `trust_remote_code=True` to permit loading of this dataset."
+    )
+
+
+def resolve_trust_remote_code(trust_remote_code: Optional[bool], repo_id: str) -> bool:
```
mariosasko, line 99: It complicates the logic, but Transformers does the same, so it's okay. 

A comment to the `transformers` version would be nice (to make it clear why it's implemented as-is)

mariosasko (commented):
Nice! Some nits:

src/datasets/load.py
```diff
@@ -1887,6 +2120,7 @@ def load_dataset(
     streaming: bool = False,
     num_proc: Optional[int] = None,
     storage_options: Optional[Dict] = None,
+    trust_remote_code: bool = None,
```
albertvillanova, line 2123: I think we should raise a `FutureWarning` to warn users that the default behavior will change in the future if they do not explicitly pass `trust_remote_code`.

albertvillanova (commented):
Thanks. Just a suggestion below.

src/datasets/load.py
lhoestq: good idea !

lhoestq:
I added tests and docs @mariosasko @albertvillanova let le know what you think !

lhoestq:
I switched to using `deepmind/code_contests` in examples in the docs to avoid having to pass trust_remote_code, and remove the DEFAULT naming stuff :)

mariosasko (approved):
LGTM, thanks!

PR #6431, opened 2023-11-16
Create DatasetNotFoundError and DataFilesNotFoundError

### LINKED ISSUE #6397: Raise a different exception for inexisting dataset vs files without known extension
See https://github.com/huggingface/datasets-server/issues/2082#issuecomment-1805716557

We have the same error for:
- https://huggingface.co/datasets/severo/a_dataset_that_does_not_exist: a dataset that does not exist
- https://huggingface.co/datasets/severo/test_files_without_extension: a dataset with files without a known extension

```
>>> import datasets
>>> datasets.get_dataset_config_names('severo/a_dataset_that_does_not_exist')
  File "/home/slesage/hf/datasets-server/services/worker/.venv/lib/python3.9/site-packages/datasets/inspect.py", line 351, in get_dataset_config_names
    dataset_module = dataset_module_factory(
  File "/home/slesage/hf/datasets-server/services/worker/.venv/lib/python3.9/site-packages/datasets/load.py", line 1508, in dataset_module_factory
    raise FileNotFoundError(
FileNotFoundError: Couldn't find a dataset script at /home/slesage/hf/datasets-server/services/worker/severo/a_dataset_that_does_not_exist/a_dataset_that_does_not_exist.py or any data file in the same directory. Couldn't find 'severo/a_dataset_that_does_not_exist' on the Hugging Face Hub either: FileNotFoundError: Dataset 'severo/a_dataset_that_does_not_exist' doesn't exist on the Hub. If the repo is private or gated, make sure to log in with `huggingface-cli login`.
>>> datasets.get_dataset_config_names('severo/test_files_without_extension')
  File "/home/slesage/hf/datasets-server/services/worker/.venv/lib/python3.9/site-packages/datasets/inspect.py", line 351, in get_dataset_config_names
    dataset_module = dataset_module_factory(
  File "/home/slesage/hf/datasets-server/services/worker/.venv/lib/python3.9/site-packages/datasets/load.py", line 1508, in dataset_module_factory
    raise FileNotFoundError(
FileNotFoundError: Couldn't find a dataset script at /home/slesage/hf/datasets-server/services/worker/severo/test_files_without_extension/test_files_without_extension.py or any data file in the same directory. Couldn't find 'severo/test_files_without_extension' on the Hugging Face Hub either: FileNotFoundError: No (supported) data files or dataset script found in severo/test_files_without_extension.
```

To differentiate, we must parse the error message (only the end is different). We should have a different exception for these two errors.

albertvillanova:
Create `DatasetNotFoundError` and `DataFilesNotFoundError`.

Fix #6397.

CC: @severo

### DIFF
--- a/src/datasets/exceptions.py
+++ b/src/datasets/exceptions.py
@@ -5,10 +5,23 @@
 class DatasetsError(Exception):
     """Base class for exceptions in this library."""
 
-    pass
-
 
 class DefunctDatasetError(DatasetsError):
     """The dataset has been defunct."""
 
-    pass
+
+class FileNotFoundDatasetsError(DatasetsError, FileNotFoundError):
+    """FileNotFoundError raised by this library."""
+
+
+class DataFilesNotFoundError(FileNotFoundDatasetsError):
+    """No (supported) data files found."""
+
+
+class DatasetNotFoundError(FileNotFoundDatasetsError):
+    """Dataset not found.
+
+    Raised when trying to access:
+    - a missing dataset, or
+    - a private/gated dataset and the user is not authenticated.
+    """
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -1504,6 +1504,16 @@ def dataset_module_factory(
                     raise ConnectionError(f"Couldn't reach the Hugging Face Hub for dataset '{path}': {e1}") from None
                 if isinstance(e1, EmptyDatasetError):
                     raise e1 from None
+                if isinstance(e1, DataFilesNotFoundError):
+                    raise DataFilesNotFoundError(
+                        f"Couldn't find a dataset script at {relative_to_absolute_path(combined_path)} or any data file in the same directory. "
+                        f"Couldn't find '{path}' on the Hugging Face Hub either: {type(e1).__name__}: {e1}"
+                    ) from None
+                if isinstance(e1, DatasetNotFoundError):
+                    raise DatasetNotFoundError(
+                        f"Couldn't find a dataset script at {relative_to_absolute_path(combined_path)} or any data file in the same directory. "
+                        f"Couldn't find '{path}' on the Hugging Face Hub either: {type(e1).__name__}: {e1}"
+                    ) from None
                 if isinstance(e1, FileNotFoundError):
                     raise FileNotFoundError(
                         f"Couldn't find a dataset script at {relative_to_absolute_path(combined_path)} or any data file in the same directory. "
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -1495,12 +1495,12 @@ def dataset_module_factory(
                     download_config=download_config,
                     download_mode=download_mode,
                 ).get_module()
-        except (
-            Exception
-        ) as e1:  # noqa all the attempts failed, before raising the error we should check if the module is already cached.
+        except Exception as e1:
+            # All the attempts failed, before raising the error we should check if the module is already cached
             try:
                 return CachedDatasetModuleFactory(path, dynamic_modules_path=dynamic_modules_path).get_module()
-            except Exception:  # noqa if it's not in the cache, then it doesn't exist.
+            except Exception:
+                # If it's not in the cache, then it doesn't exist.
                 if isinstance(e1, OfflineModeIsEnabled):
                     raise ConnectionError(f"Couldn't reach the Hugging Face Hub for dataset '{path}': {e1}") from None
                 if isinstance(e1, EmptyDatasetError):
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -495,9 +495,10 @@ def infer_module_for_data_files(
     """Infer module (and builder kwargs) from data files. Raise if module names for different splits don't match.
 
     Args:
-        data_files (DataFilesDict): List of data files.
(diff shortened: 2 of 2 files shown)


src/datasets/load.py
```diff
@@ -1493,16 +1495,26 @@ def dataset_module_factory(
...
+                if isinstance(e1, DataFilesNotFoundError):
+                    raise DataFilesNotFoundError(
+                        f"Couldn't find a dataset script at {relative_to_absolute_path(combined_path)} or any data file in the same directory. "
+                        f"Couldn't find '{path}' on the Hugging Face Hub either: {type(e1).__name__}: {e1}"
+                    ) from None
+                if isinstance(e1, DatasetNotFoundError):
+                    raise DatasetNotFoundError(
+                        f"Couldn't find a dataset script at {relative_to_absolute_path(combined_path)} or any data file in the same directory. "
+                        f"Couldn't find '{path}' on the Hugging Face Hub either: {type(e1).__name__}: {e1}"
+                    ) from None
```
albertvillanova, line 1517: Should we change and make more specific the raised messages?

severo: Indeed, it's weird to show the same error in both cases, no? The e1 messages seem clearer to me, maybe just re-raise them?

src/datasets/exceptions.py
```diff
@@ -5,10 +5,23 @@
...
 
-    pass
-
 
 class DefunctDatasetError(DatasetsError):
     """The dataset has been defunct."""
 
-    pass
+
+class FileNotFoundDatasetsError(DatasetsError, FileNotFoundError):
```
severo, line 13: why do you create this intermediate exception? it's not used as far as I can understand

albertvillanova: Both `DataFilesNotFoundError` and `DatasetNotFoundError` derive from `FileNotFoundDatasetsError`. Therefore both are subclasses of `DatasetsError` (the base error class for all errors raised by `datasets`) and also subclasses of `FileNotFoundError` (so that they are backward compatible).

severo: > so that they are backward compatible

OK! I didn't get this part

src/datasets/load.py
albertvillanova: Done.

severo (approved):

albertvillanova:
I think this PR can be merged.

severo:
you already have an approval, feel free to merge!

src/datasets/exceptions.py
```diff
@@ -5,10 +5,23 @@
...
+    """No (supported) data files found."""
+
+
+class DatasetNotFoundError(FileNotFoundDatasetsError):
+    """Dataset not found.
+
+    Raised when trying to access:
+    - a missing dataset, or
+    - a private/gated dataset and the user is not authenticated.
+    """
```
mariosasko, line 27: Maybe we could re-use `huggingface_hub`'s `RepositoryNotFound` and `GatedRepo` errors instead of introducing our own (this exception should at least subclass them)

For example, `transformers` throws an `EnvironmentError` (with the error description)  and chains the caught (`huggingface_hub`) exception, so consistency with them would be nice.

src/datasets/load.py
```diff
@@ -512,7 +514,7 @@ def infer_module_for_data_files(
         raise ValueError(f"Couldn't infer the same data file format for all splits. Got {split_modules}")
     if not module_name:
         path = f" in {path}. " if path else ". "
-        raise FileNotFoundError(f"No (supported) data files or dataset script found{path}")
+        raise DataFilesNotFoundError(f"No (supported) data files or dataset script found{path}")
```
mariosasko, line 517: Maybe `NotADatasetError` would be a better name for this error (the "dataset script" is not a data file)

albertvillanova: I had kept the original message errors for backward compatibility: before we were raising the same `FileNotFoundError` for all cases and end users could only differentiate among them by parsing the error messages
- This is indeed the case for datasets-server; see https://github.com/huggingface/datasets/issues/6397#issue-1987622152
  > To differentiate, we must parse the error message (only the end is different). We should have a different exception for these two errors.

However, I would propose changing the error message in this specific case (and keep the error name) because:
- this error is raised by a function that just checks the data files and tries to infer the module from them
- for the cases where it is a script-dataset, this is handled in other part of the code inside `dataset_module_factory`
- anyway, we are pushing no-script datasets, so their usage should be exceptional

src/datasets/exceptions.py
albertvillanova: I agree with subclassing `huggingface_hub` errors as well, but at the same time our `datasets` `DatasetsError` base clase, so that a user of `datasets` can catch all `datasets` errors by using this `DatasetsError` class.

albertvillanova: I would propose to address in a subsequent PR to catch specific `huggingface_hub` errors (instead of parsing the HTTP error status code: this is done by `huggingface_hub`) and raise our specific errors accordingly.

src/datasets/load.py
albertvillanova: Done.

albertvillanova:
I am merging this PR because we need it by `datasets-server`.

PR #6433, opened 2023-11-17
Better `tqdm` wrapper

### LINKED ISSUE #6409: using DownloadManager to download from local filesystem and disable_progress_bar, there will be an exception
### Describe the bug

i'm using datasets.download.download_manager.DownloadManager to download files like "file:///a/b/c.txt", and i disable_progress_bar() to disable bar. there will be an exception as follows:

`AttributeError: 'function' object has no attribute 'close'
Exception ignored in: <function TqdmCallback.__del__ at 0x7fa8683d84c0>
  File "/home/protoss.gao/.local/lib/python3.9/site-packages/fsspec/callbacks.py", line 233, in __del__
    self.tqdm.close()`

i check your source code in datasets/utils/file_utils.py:348 you define TqdmCallback derive from fsspec.callbacks.TqdmCallback

but in the newest fsspec code [https://github.com/fsspec/filesystem_spec/blob/master/fsspec/callbacks.py](url) , line 146, in this case, _DEFAULT_CALLBACK will take effect, but in line 234, it calls "close()" function which _DEFAULT_CALLBACK don't have such thing.

so i think the class "TqdmCallback" in datasets/utils/file_utils.py may override "__del__" function or report this bug to fsspec.

as i said

### Expected behavior

no exception

### Environment info

datasets: 2.14.4
python: 3.9
platform: x86_64

mariosasko:
This PR aligns the `tqdm` logic with `huggingface_hub` (without introducing breaking changes), as the current one is error-prone.

Additionally, it improves the doc page about the `datasets`' utilities, and the handling of local `fsspec` paths in `cached_path`.
 
Fix #6409


src/datasets/config.py
```diff
@@ -192,6 +192,18 @@
 # Offline mode
 HF_DATASETS_OFFLINE = os.environ.get("HF_DATASETS_OFFLINE", "AUTO").upper() in ENV_VARS_TRUE_VALUES
 
+# Here, `True` will disable progress bars globally without possibility of enabling it
+# programmatically. `False` will enable them without possibility of disabling them.
+# If environment variable is not set (None), then the user is free to enable/disable
+# them programmatically.
+# TL;DR: env variable has priority over code
+__HF_DATASETS_DISABLE_PROGRESS_BARS = os.environ.get("HF_HUB_DISABLE_PROGRESS_BARS")
```
lhoestq, line 200:
```suggestion
__HF_DATASETS_DISABLE_PROGRESS_BARS = os.environ.get("HF_DATASETS_DISABLE_PROGRESS_BARS")
```

src/datasets/utils/file_utils.py
```diff
@@ -177,6 +178,17 @@ def cached_path(
...
+    # Convert fsspec URL in the format "file://local/path" to "local/path"
+    # Copied from: https://github.com/fsspec/filesystem_spec/blob/14a778879b6a68b65031ef86e232b9d88c005bab/fsspec/implementations/local.py#L214-L221
+    if url_or_filename.startswith("file://"):
+        url_or_filename = url_or_filename[7:]
+    elif url_or_filename.startswith("file:"):
+        url_or_filename = url_or_filename[5:]
+    elif url_or_filename.startswith("local://"):
+        url_or_filename = url_or_filename[8:]
+    elif url_or_filename.startswith("local:"):
+        url_or_filename = url_or_filename[6:]
```
lhoestq, line 190: Maybe use LocalFileSystem._strip_protocol instead ?

src/datasets/utils/tqdm.py
```diff
@@ -0,0 +1,130 @@
+"""Utility helpers to handle progress bars in `huggingface_hub`.
```
lhoestq, line 1:
```suggestion
"""Utility helpers to handle progress bars in `datasets`.
```

lhoestq (approved):
Cool ! just one comment:

src/datasets/utils/file_utils.py
mariosasko: `LocalFileSystem._strip_protocol` always returns an absolute path, but I think this is fine for `cached_path` (there was a test checking relative paths are preserved, so I had to update it)

PR #6444, opened 2023-11-22
Remove `Table.__getstate__` and `Table.__setstate__`

LZHgrla:
When using distributed training, the code of `os.remove(filename)` may be executed separately by each rank, leading to `FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmprxxxxxxx.arrow'`

```python
from torch import distributed as dist

if dist.get_rank() == 0:
    dataset = process_dataset(*args, **kwargs)
    objects = [dataset]
else:
    objects = [None]
dist.broadcast_object_list(objects, src=0)
dataset = objects[0]
```


mariosasko:
Thanks for working on this! The [issue](https://bugs.python.org/issue24658) with pickling objects larger than 4GB seems to be patched in Python 3.8 (the minimal supported version was 3.6 at the time of implementing this), so a simple solution would be removing the `Table.__setstate__` and `Table.__getstate__` overrides.

LZHgrla:
@mariosasko 
Cool!
I removed these overrides, and it worked.

All modifications are committed. Ready for review!

mariosasko (approved):
LGTM, thanks!

PR #6456, opened 2023-11-28
Don't require trust_remote_code in inspect_dataset

lhoestq:
don't require `trust_remote_code` in (deprecated) `inspect_dataset` (it defeats its purpose)

(not super important but we might as well keep it until the next major release)

this is needed to fix the tests in https://github.com/huggingface/datasets/pull/6448


mariosasko (approved):
LGTM!

PR #6458, opened 2023-11-29
Lazy data files resolution

lhoestq:
Related to discussion at https://github.com/huggingface/datasets/pull/6255

this makes this code run in 2sec instead of >10sec

```python

ds = load_dataset("glue", "sst2", streaming=True, trust_remote_code=False)
```

For some datasets with many configs and files it can be up to 100x faster.
This is particularly important now that some datasets will be loaded from the Parquet export instead of the scripts.

The data files are only resolved in the builder `__init__`. To do so I added DataFilesPatternsList and DataFilesPatternsDict that have `.resolve()` to return resolved DataFilesList and DataFilesDict


lhoestq:
Getting the same windows error as in my other PR. I couldn't reproduce on my windows machine though 🧐

mariosasko (commented):
Maybe we can avoid adding this much complexity for the YAML case (not used much?) by turning `DataFilesList` into a lazy iterable that caches its elements as it's being iterated over (we don't need random access, so no need for the list).

lhoestq:
`DataFilesDict` is used in some repositories when people want to list files from a repo using glob patterns

lhoestq:
Also making DataFilesList lazy would require to make the pickling more complex, since we don't want to resolve the data files when pickling. At the same time we want to get different hashes if the data files and origin metadata are different so revolving the patterns is needed in that case (we hash the data files when creating the config_id, used in the cache)

lhoestq:
> The BuilderConfig.data_files property suggested above should address this, no?

That works indeed ! let me try something

lhoestq:
Implementing lazy DataFilesList and .data_files brings more complexity (less readable, more bad side effects) so I think the current solution is the best one

lhoestq:
I opened https://github.com/huggingface/datasets/pull/6493 to continue this and fix conflicts with https://github.com/huggingface/datasets/pull/6459

PR #6463, opened 2023-12-01
Disable benchmarks in PRs

lhoestq:
In order to keep PR pages less spammy / more readable.
Having the benchmarks on commits on `main` is enough imo


albertvillanova (approved):
Thanks.

Just a naive question: what is the purpose of these benchmarks? Are we following their results and taking action accordingly?

lhoestq:
It's a way to detect regressions in performance sensitive methods like map

PR #6482, opened 2023-12-08
Fix max lock length on unix

lhoestq:
reported in https://github.com/huggingface/datasets/pull/6482

### DIFF
--- a/src/datasets/utils/_filelock.py
+++ b/src/datasets/utils/_filelock.py
@@ -17,7 +17,8 @@
 
 import os
 
-from filelock import FileLock as FileLock_, UnixFileLock
+from filelock import FileLock as FileLock_
+from filelock import UnixFileLock
 
 
 class FileLock(FileLock_):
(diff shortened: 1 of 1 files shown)


src/datasets/utils/_filelock.py
```diff
@@ -34,11 +35,14 @@ def __init__(self, lock_file, *args, **kwargs):
     @classmethod
     def hash_filename_if_too_long(cls, path: str) -> str:
         filename = os.path.basename(path)
-        if len(filename) > cls.MAX_FILENAME_LENGTH:
+        max_filename_length = cls.MAX_FILENAME_LENGTH
+        if issubclass(cls, UnixFileLock):
+            max_filename_length = min(max_filename_length, os.statvfs(os.path.dirname(path)).f_namemax)
```
mariosasko, line 40: A bit simpler:

```suggestion
        max_filename_length = min(cls. MAX_FILENAME_LENGTH, os.statvfs(os.path.dirname(path)).f_namemax)
```

lhoestq:
I'm getting `AttributeError: module 'os' has no attribute 'statvfs'` on windows - reverting

PR #6487, opened 2023-12-11
Update builder hash with info

lhoestq:
Currently if you change the `dataset_info` of a dataset (e.g. in the YAML part of the README.md), the cache ignores this change.

This is problematic because you want to regenerate a dataset if you change the features or the split sizes for example (e.g. after push_to_hub)

Ideally we should take the resolved files into account as well but this will be for another PR

### DIFF
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -604,6 +604,17 @@ def update_hash_with_config_parameters(hash: str, config_parameters: dict) -> st
     return m.hexdigest()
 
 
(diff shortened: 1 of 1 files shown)


lhoestq:
Closing this one in favor of https://github.com/huggingface/datasets/pull/6458/commits/565c294fc12bc547730a023a610ed4f92313d8fb in https://github.com/huggingface/datasets/pull/6458

PR #6492, opened 2023-12-12
Make push_to_hub return CommitInfo

albertvillanova:
Make `push_to_hub` return `CommitInfo`.

This is useful, for example, if we pass `create_pr=True` and we want to know the created PR ID.

CC: @severo for the use case in https://huggingface.co/datasets/jmhessel/newyorker_caption_contest/discussions/4


albertvillanova:
This PR is ready to review @huggingface/datasets.

lhoestq (approved):
Nice !

PR #6498, opened 2023-12-14
Fallback on dataset script if user wants to load default config

lhoestq:
Right now this code is failing on `main`:

```python
load_dataset("openbookqa")
```

This is because it tries to load the dataset from the Parquet export but the dataset has multiple configurations and the Parquet export doesn't know which one is the default one.

I fixed this by simply falling back on using the dataset script (which tells the user to pass `trust_remote_code=True`):

```python
load_dataset("openbookqa", trust_remote_code=True)
```

Note that if the user happened to specify a config name I don't fall back on the script since we can use the Parquet export in this case (no need to know which config is the default)

```python
load_dataset("openbookqa", "main")
```

### DIFF
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -31,7 +31,7 @@ from typing import Any, Dict, List, Mapping, Optional, Sequence, Tuple, Type, Un
 
 import fsspec
 import requests
-from huggingface_hub import DatasetCard, DatasetCardData, HfApi
+from huggingface_hub import DatasetCard, DatasetCardData, HfApi, HfFileSystem
 
 from . import config
 from .arrow_dataset import Dataset
@@ -1621,6 +1621,7 @@ def dataset_module_factory(
     data_dir: Optional[str] = None,
     data_files: Optional[Union[Dict, List, str, DataFilesDict]] = None,
     trust_remote_code: Optional[bool] = None,
+    _ignore_default_config=False,
     **download_kwargs,
 ) -> DatasetModule:
     """
@@ -1772,8 +1773,15 @@ def dataset_module_factory(
                 else:
                     raise e
             if filename in [sibling.rfilename for sibling in dataset_info.siblings]:  # contains a dataset script
-                if config.USE_PARQUET_EXPORT:
+                fs = HfFileSystem(endpoint=config.HF_ENDPOINT, token=download_config.token)
+                with fs.open(f"datasets/{path}/{filename}", "r", revision=revision) as f:
+                    can_load_config_from_parquet_export = (
+                        "DEFAULT_CONFIG_NAME" not in f.read() or _ignore_default_config
+                    )
+                if config.USE_PARQUET_EXPORT and can_load_config_from_parquet_export:
                     # If the parquet export is ready (parquet files + info available for the current sha), we can use it instead
+                    # This fails when the dataset has multiple configs and a default config and
+                    # the user didn't specify a configuration name (_ignore_default_config=False).
                     try:
                         return HubDatasetModuleFactoryWithParquetExport(
                             path, download_config=download_config, revision=dataset_info.sha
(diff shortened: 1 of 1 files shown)


src/datasets/load.py
```diff
@@ -1621,6 +1621,7 @@ def dataset_module_factory(
     data_dir: Optional[str] = None,
     data_files: Optional[Union[Dict, List, str, DataFilesDict]] = None,
     trust_remote_code: Optional[bool] = None,
+    _ignore_default_config=False,
```
albertvillanova, line 1624: Why the leading underscore?

albertvillanova: I would suggest naming it differently, to make clear that config_name was passed or not: if config_name is passed, we don't even care whether DEFAULT_CONFIG_NAME is set or not.

src/datasets/load.py
```diff
@@ -1772,31 +1773,30 @@ def dataset_module_factory(
                 else:
                     raise e
             if filename in [sibling.rfilename for sibling in dataset_info.siblings]:  # contains a dataset script
-                if config.USE_PARQUET_EXPORT:
+                fs = HfFileSystem(endpoint=config.HF_ENDPOINT, token=download_config.token)
+                with fs.open(f"datasets/{path}/{filename}", "r", revision=revision, encoding="utf-8") as f:
+                    can_load_config_from_parquet_export = (
+                        "DEFAULT_CONFIG_NAME" not in f.read() or _ignore_default_config
+                    )
```
albertvillanova, line 1780: I think it would be better to refactor this code and avoid the network call and file read if the variable "_ignore_default_config" is True.

src/datasets/load.py
```diff
@@ -1772,31 +1773,30 @@ def dataset_module_factory(
...
+                    # This fails when the dataset has multiple configs and a default config and
+                    # the user didn't specify a configuration name (_ignore_default_config=False).
                     try:
                         return HubDatasetModuleFactoryWithParquetExport(
                             path, download_config=download_config, revision=dataset_info.sha
                         ).get_module()
                     except _datasets_server.DatasetsServerError:
                         pass
                 # Otherwise we must use the dataset script if the user trusts it
-                trust_remote_code = resolve_trust_remote_code(trust_remote_code, path)
```
albertvillanova, line 1784: I guess this is no longer necessary?

albertvillanova (commented):
I was just thinking: what if the user does not pass a config name and the dataset has only a config with a name different from "default"?

src/datasets/load.py
lhoestq: The underscore is to signify this is an internal parameter (not supposed to be used by users)

lhoestq:
> I was just thinking: what if the user does not pass a config name and the dataset has only a config with a name different from "default"?

You mean if there is a DEFAULT_CONFIG_NAME defined in the script but the dataset only has one configuration ? We can't easily get the number of configs without running the python code so I don't think we can support detect this case

src/datasets/load.py
lhoestq: It's already called inside the HubDatasetModuleFactoryWithScript - no need to have it here

src/datasets/load.py
lhoestq: I think I like this name: if `_require_default_config_name=True` we require the builder class to have a default config name. If the user doesn't specify a config name we set it to `True`

src/datasets/load.py
lhoestq: done !

lhoestq:
Most datasets with a script don't define DEFAULT_CONFIG_NAME if there is only one configuration anyway.

So there is no issue e.g. for `squad`

albertvillanova (commented):
> You mean if there is a DEFAULT_CONFIG_NAME defined in the script but the dataset only has one configuration ? We can't easily get the number of configs without running the python code so I don't think we can support detect this case

I was trying to mean the case where DEFAULT_CONFIG_NAME is None but there is only a single config in BUILDER_CONFIGS, with a name different from "default".

src/datasets/load.py
albertvillanova: > The underscore is to signify this is an internal parameter (not supposed to be used by users)

I mean, it is the first time I see a "private" function parameter... Wondering if that makes sense...

lhoestq:
> I was trying to mean the case where DEFAULT_CONFIG_NAME is None but there is only a single config in BUILDER_CONFIGS, with a name different from "default".

In this case we can detect if "DEFAULT_CONFIG_NAME" is not mentioned and use the Parquet export. If it is mentioned (and maybe it is set to None or to the single config) I consider that it may have multiple configs and fall back on using the script

src/datasets/load.py
lhoestq: It's not a good practice :/

I hesitated to not have the feature to use the parquet export if the user does specify a config name explicitly just because I don't like doing "private" arguments

albertvillanova:
... but the user does not pass the config name.

lhoestq:
see 

https://github.com/huggingface/datasets/blob/2feaa589de86dd85941301fc8c3fa091731a67c0/src/datasets/builder.py#L532-L532

src/datasets/load.py
albertvillanova: I see...

albertvillanova (approved):
Thanks for the explanations!! :sweat_smile:

PR #6500, opened 2023-12-15
Enable passing default config name to push_to_hub

### LINKED ISSUE #6497: Support setting a default config name in push_to_hub
In order to convert script-datasets to no-script datasets, we need to support setting a default config name for those scripts that set one.

albertvillanova:
Fix #6497.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -5225,6 +5225,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         self,
         repo_id: str,
         config_name: str = "default",
+        as_default_config_name: Optional[bool] = None,
         split: Optional[str] = None,
         commit_message: Optional[str] = None,
         private: Optional[bool] = False,
@@ -5250,6 +5251,8 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                 of the logged-in user.
             config_name (`str`, defaults to "default"):
                 The configuration name (or subset) of a dataset. Defaults to "default".
+            as_default_config_name (`bool`, *optional*):
+                Whether to set the config name as default (not necessary if the config_name is "default").
             split (`str`, *optional*):
                 The name of the split that will be given to that dataset. Defaults to `self.split`.
             commit_message (`str`, *optional*):
@@ -5485,6 +5488,8 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
             }
         else:
             metadata_config_to_dump = {"data_files": [{"split": split, "path": f"{data_dir}/{split}-*"}]}
+        if as_default_config_name:
+            metadata_config_to_dump["default"] = True
         # push to the deprecated dataset_infos.json
         if repo_with_dataset_infos:
             dataset_infos_path = api.hf_hub_download(
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -5278,13 +5278,14 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
 
                 </Deprecated>
             create_pr (`bool`, *optional*, defaults to `False`):
-                Whether or not to create a PR with the uploaded files or directly commit.
+                Whether to create a PR with the uploaded files or directly commit.
 
                 <Added version="2.15.0"/>
             max_shard_size (`int` or `str`, *optional*, defaults to `"500MB"`):
                 The maximum size of the dataset shards to be uploaded to the hub. If expressed as a string, needs to be digits followed by
                 a unit (like `"5MB"`).
-            num_shards (`int`, *optional*): Number of shards to write. By default the number of shards depends on `max_shard_size`.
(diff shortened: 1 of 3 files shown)


albertvillanova:
This is ready for review @huggingface/datasets.

src/datasets/arrow_dataset.py
```diff
@@ -5225,6 +5225,7 @@ def push_to_hub(
         self,
         repo_id: str,
         config_name: str = "default",
+        as_default_config_name: Optional[bool] = None,
```
lhoestq, line 5228: Maybe name it `set_default_config` (boolean) ?

src/datasets/arrow_dataset.py
```diff
@@ -5250,6 +5251,8 @@ def push_to_hub(
                 of the logged-in user.
             config_name (`str`, defaults to "default"):
                 The configuration name (or subset) of a dataset. Defaults to "default".
+            as_default_config_name (`bool`, *optional*):
+                Whether to set the config name as default (not necessary if the config_name is "default").
```
lhoestq, line 5255: maybe simpler ?
```suggestion
                Whether to set the config name as default (not necessary if you push only one dataset configuration).
```

lhoestq (commented):
Nice ! Just 2 comments:

src/datasets/arrow_dataset.py
albertvillanova: I know what you mean, but strictly speaking:
- when you call push_to_hub you are pushing just only one configuration; so it is always the case that you push only one
- even if you push multiple configurations (by calling push_to_hub multiple times) and one of them is named "default", it is not necessary either

src/datasets/arrow_dataset.py
albertvillanova: OK for the rename.

On the other hand, for the type, I decided to make it default None, because I find it weird to set the default config_name "default" but set the default set_default_config False...

src/datasets/arrow_dataset.py
lhoestq: Ah yes correct ! 

https://github.com/huggingface/datasets/blob/2feaa589de86dd85941301fc8c3fa091731a67c0/src/datasets/utils/metadata.py#L235-L235

Maybe this then ?

```suggestion
                Whether to set this configuration as the default one. Otherwise the default configuration is the one named "default"
```

src/datasets/arrow_dataset.py
lhoestq: Right, makes sense !

albertvillanova: What about `is_default_config`? That way we make clear is boolean...

lhoestq: I like `set_xxx` because it's explicit that it would unset the previous default config

src/datasets/arrow_dataset.py
```diff
@@ -5485,6 +5490,8 @@ def push_to_hub(
             }
         else:
             metadata_config_to_dump = {"data_files": [{"split": split, "path": f"{data_dir}/{split}-*"}]}
+        if set_default_config:
+            metadata_config_to_dump["default"] = True
```
lhoestq, line 5494: You might also have to remove the default field for all the already existing configurations (we should end up with only one configuration with default: True)

albertvillanova: I was thinking of that...

Indeed this is already the case: currently, multiple configuration can be defined in the YAML as default, and we are returning the first one.

albertvillanova: It is already the case that we cannot have a configuration named "default" as non-default...

src/datasets/arrow_dataset.py
albertvillanova: Normally `set_xxx` would be used to set a value for the variable `xxx`.

Anyway, I set it as you said.

lhoestq: Well technically we do set `default: True` ^^'

lhoestq: But maybe it should be named `set_default` then instead of `set_default_config`

lhoestq:
Also what if the config is being overwritten and it was the default config and the user doesn't pass `set_default` ?
I'd expect the config to keep being the default one but lmk what you think

albertvillanova:
How can you unset a config as default one? In the case you mentioned, I would expect the config not being the default one.

lhoestq:
Maybe by passing `set_default=False` ?

albertvillanova:
I think that way we are unnecessarily complicating the logic of `push_to_hub` and as I told you, I would expect the contrary: the result of calling `push_to_hub` with a determined set of arguments should always be the same, independently of previous calls and the current state of the config on the Hub. Push to hub should be somehow stateless in that sense, and IMO the user expects that the push overwrites previous config if already present on the Hub. I find very confusing making it to partially update the config on the Hub.

lhoestq:
That makes sense, having it stateless is simpler and no need to do something too fancy indeed

src/datasets/arrow_dataset.py
albertvillanova: Done.

lhoestq (approved):
LGTM !

PR #6520, opened 2023-12-21
Support commit_description parameter in push_to_hub

albertvillanova:
Support `commit_description` parameter in `push_to_hub`.

CC: @Wauplin


Wauplin (approved):
Thanks for adding it!

Issue #6533, opened 2023-12-26
ted_talks_iwslt | Error: Config name is missing


rayliuca:
### Describe the bug

Running load_dataset using the newest `datasets` library like below on the ted_talks_iwslt using year pair data will throw an error "Config name is missing"

see also:
https://huggingface.co/datasets/ted_talks_iwslt/discussions/3

likely caused by #6493, where the `and not config_kwargs` part in the if logic was removed 
 https://github.com/huggingface/datasets/blob/ef3b5dd3633995c95d77f35fb17f89ff44990bc4/src/datasets/builder.py#L512

### Steps to reproduce the bug

run: 
```python
load_dataset("ted_talks_iwslt", language_pair=("ja", "en"), year="2015")
```

### Expected behavior

Load the data without error

### Environment info

datasets 2.16.0

lhoestq:
Hi ! Thanks for reporting. I opened https://github.com/huggingface/datasets/pull/6544 to fix this

lhoestq:
We just released 2.16.1 with a fix:

```
pip install -U datasets
```

PR #6544, opened 2023-12-30
Fix custom configs from script

lhoestq:
We should not use the parquet export when the user is passing config_kwargs

I also fixed a regression that would disallow creating a custom config when a dataset has multiple predefined configs

fix https://github.com/huggingface/datasets/issues/6533


albertvillanova:
Thanks for the fix and the patch release. This confirms that, as I suggested in the Summer, maybe we should avoid making a release right before leaving on holidays.

PR #6551, opened 2024-01-02
Fix parallel downloads for datasets without scripts

lhoestq:
Enable parallel downloads using multiprocessing when `num_proc` is passed to `load_dataset`.

It was enabled for datasets with scripts already (if they passed lists to `dl_manager.download`) but not for no-script datasets (we pass dicts {split: [list of files]} to `dl_manager.download` for those ones).

I fixed this by parallelising on the lists contained in the data files dicts when possible.

I also added a context manager `stack_multiprocessing_download_progress_bars` in `DownloadManager` to stack the progress bard of the downloads (from `cached_path(...)` calls). Otherwise the progress bars overlap each other with an annoying flickering effect.

### DIFF
--- a/src/datasets/download/download_manager.py
+++ b/src/datasets/download/download_manager.py
@@ -30,7 +30,14 @@ from typing import Callable, Dict, Generator, List, Optional, Tuple, Union
 from .. import config
 from ..utils import tqdm as hf_tqdm
 from ..utils.deprecation_utils import DeprecatedEnum, deprecated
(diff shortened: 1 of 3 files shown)


kopyl:
@lhoestq 
image
it's still not fixed =(

kopyl:
@lhoestq i was thinking uninstalling `datasets` and then `pip install git+https://github.com/huggingface/datasets.git` has to fix it. Buuuuut. I'm not sure what's going on actually...

Now instead of showing progress bars one after another it seems to be downloading the dataset way way way faster (like 4 mins instead of 58, thank you very much) but does not show any progress bars related to downloading at all.

image

image

PR #6555, opened 2024-01-03
Do not use Parquet exports if revision is passed

### LINKED ISSUE #6554: Parquet exports are used even if revision is passed
We should not used Parquet exports if `revision` is passed.

I think this is a regression.

albertvillanova:
Fix #6554.

### DIFF
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -1794,7 +1794,7 @@ def dataset_module_factory(
                 if _require_custom_configs or (revision and revision != "main"):
                     can_load_config_from_parquet_export = False
                 elif _require_default_config_name:
-                    with fs.open(f"datasets/{path}/{filename}", "r", revision=revision, encoding="utf-8") as f:
+                    with fs.open(f"datasets/{path}/{filename}", "r", encoding="utf-8") as f:
                         can_load_config_from_parquet_export = "DEFAULT_CONFIG_NAME" not in f.read()
                 else:
                     can_load_config_from_parquet_export = True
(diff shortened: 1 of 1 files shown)


src/datasets/load.py
```diff
@@ -1791,10 +1791,10 @@ def dataset_module_factory(
                     raise e
             if filename in [sibling.rfilename for sibling in dataset_info.siblings]:  # contains a dataset script
                 fs = HfFileSystem(endpoint=config.HF_ENDPOINT, token=download_config.token)
-                if _require_custom_configs:
+                if _require_custom_configs or (revision and revision != "main"):
                     can_load_config_from_parquet_export = False
                 elif _require_default_config_name:
-                    with fs.open(f"datasets/{path}/{filename}", "r", revision=revision, encoding="utf-8") as f:
+                    with fs.open(f"datasets/{path}/{filename}", "r", encoding="utf-8") as f:
```
albertvillanova, line 1797: I have also removed passing `revision` to `fs.open`: 3acefe542dd4b1f06a879b6c1081089e95588f9d
- Previous condition makes this unnecessary
- Additionally, note that this has never worked (the revision is ignored). I have opened a PR to fix it: 
  - huggingface/huggingface_hub#1948

lhoestq:
As shared on slack, `HubDatasetModuleFactoryWithParquetExport` raises a `DatasetsServerError` already if the user tries to load another revision that the one from the parquet export. And therefore it fall backs on using `HubDatasetModuleFactoryWithScript`

src/datasets/load.py
lhoestq: good catch !

albertvillanova:
@lhoestq I would say that although current implementation finally returns `HubDatasetModuleFactoryWithScript` as expected, with this PR we avoid the useless call to `HubDatasetModuleFactoryWithParquetExport.get_module`, so this is more optimal.

lhoestq (approved):
Indeed, lgtm then :)

PR #6557, opened 2024-01-03
Support standalone yaml

lhoestq:
see (internal) https://huggingface.slack.com/archives/C02V51Q3800/p1703885853581679


src/datasets/load.py
```diff
@@ -1004,7 +1005,15 @@ def __init__(
 
     def get_module(self) -> DatasetModule:
         readme_path = os.path.join(self.path, "README.md")
+        standalone_yaml_path = os.path.join(self.path, ".huggingface.yml")
```
julien-c, line 1008: i think it's .yaml on the hub...

lhoestq: Good catch, I got confused by an internal PR name

julien-c: never trust an internal PR 🤣

d710055071:
@lhoestq 
hello
I think it should be defined in config.py
DATASET_ README_ FILENAME="README. md"
This can replace all "README. md"

lhoestq:
Thanks for the feedback :) merging now

PR #6578, opened 2024-01-10
Faster webdataset streaming

lhoestq:
requests.get(..., streaming=True) is faster than using HTTP range requests when streaming large TAR files

it can be enabled using block_size=0 in fsspec

cc @rwightman


lhoestq:
I added faster streaming support using streaming Requests instances in `huggingface_hub`.

This PR can be used with https://github.com/huggingface/huggingface_hub/pull/1967 to get fast WebDataset streaming

src/datasets/download/streaming_download_manager.py
```diff
@@ -911,7 +917,9 @@ def _iter_from_urlpath(
         cls, urlpath: str, download_config: Optional[DownloadConfig] = None
     ) -> Generator[Tuple, None, None]:
         compression = _get_extraction_protocol(urlpath, download_config=download_config)
-        with xopen(urlpath, "rb", download_config=download_config) as f:
+        # Set chunk_size=0 to get faster streaming
```
mariosasko, line 920:
```suggestion
        # Set block_size=0 to get faster streaming
```

mariosasko (approved):

PR #6601, opened 2024-01-18
add safety checks when using only part of dataset

benseddikismail:
Added some checks to prevent errors that arrise when using evaluate.py on only a portion of the squad 2.0 dataset.


lhoestq:
Hi ! The metrics in `datasets` are deprecated in favor of https://github.com/huggingface/evaluate

You can open a PR here instead: https://huggingface.co/spaces/evaluate-metric/squad_v2/tree/main

Issue #6566, opened 2024-01-08
I train controlnet_sdxl in bf16 datatype, got unsupported ERROR in datasets


HelloWorldBeginner:
### Describe the bug

```
Traceback (most recent call last):
  File "train_controlnet_sdxl.py", line 1252, in <module>
    main(args)
  File "train_controlnet_sdxl.py", line 1013, in main
    train_dataset = train_dataset.map(compute_embeddings_fn, batched=True, new_fingerprint=new_fingerprint)
  File "/home/miniconda3/envs/mhh_df/lib/python3.8/site-packages/datasets/arrow_dataset.py", line 592, in wrapper
    out: Union["Dataset", "DatasetDict"] = func(self, *args, **kwargs)
  File "/home/miniconda3/envs/mhh_df/lib/python3.8/site-packages/datasets/arrow_dataset.py", line 557, in wrapper
    out: Union["Dataset", "DatasetDict"] = func(self, *args, **kwargs)
  File "/home/miniconda3/envs/mhh_df/lib/python3.8/site-packages/datasets/arrow_dataset.py", line 3093, in map
    for rank, done, content in Dataset._map_single(**dataset_kwargs):
  File "/home/miniconda3/envs/mhh_df/lib/python3.8/site-packages/datasets/arrow_dataset.py", line 3489, in _map_single
    writer.write_batch(batch)
  File "/home/miniconda3/envs/mhh_df/lib/python3.8/site-packages/datasets/arrow_writer.py", line 557, in write_batch
    arrays.append(pa.array(typed_sequence))
  File "pyarrow/array.pxi", line 248, in pyarrow.lib.array
  File "pyarrow/array.pxi", line 113, in pyarrow.lib._handle_arrow_array_protocol
  File "/home/miniconda3/envs/mhh_df/lib/python3.8/site-packages/datasets/arrow_writer.py", line 191, in __arrow_array__
    out = pa.array(cast_to_python_objects(data, only_1d_for_numpy=True))
  File "/home/miniconda3/envs/mhh_df/lib/python3.8/site-packages/datasets/features/features.py", line 447, in cast_to_python_objects
    return _cast_to_python_objects(
  File "/home/miniconda3/envs/mhh_df/lib/python3.8/site-packages/datasets/features/features.py", line 324, in _cast_to_python_objects
    for x in obj.detach().cpu().numpy()
TypeError: Got unsupported ScalarType BFloat16
```

### Steps to reproduce the bug

[...]

PR #6607, opened 2024-01-20
Update features.py to avoid bfloat16 unsupported error

skaulintel:
Fixes https://github.com/huggingface/datasets/issues/6566

Let me know if there's any tests I need to clear.


lhoestq:
I think not all torch tensors should be converted to float, what if it's a tensor of integers for example ?
Maybe you can check for the tensor dtype before converting

stoical07:
@lhoestq Please could this be merged? 🙏

lhoestq (approved):
Yes ! sorry for the delay :)

PR #6617, opened 2024-01-25
Fix CI: pyarrow 15, pandas 2.2 and sqlachemy

lhoestq:
this should fix the CI failures on `main`

close https://github.com/huggingface/datasets/issues/5477

### DIFF
--- a/src/datasets/table.py
+++ b/src/datasets/table.py
@@ -2088,7 +2088,7 @@ def cast_array_to_feature(array: pa.Array, feature: "FeatureType", allow_number_
         elif isinstance(feature, Sequence):
             if feature.length > -1:
                 if array.offset and feature.length * len(array) != len(array_values):
-                    array_values = array.values[array.offset * array.type.list_size :]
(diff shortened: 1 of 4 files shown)


src/datasets/table.py
```diff
@@ -2090,6 +2087,8 @@ def cast_array_to_feature(array: pa.Array, feature: "FeatureType", allow_number_
             return pa.ListArray.from_arrays(array.offsets, _c(array_values, feature[0]))
         elif isinstance(feature, Sequence):
             if feature.length > -1:
+                if array.offset and feature.length * len(array) != len(array_values):
+                    array_values = array.values[array.offset * array.type.list_size :]
```
mariosasko, line 2091: We also need to specify where the slice ends, no?

```suggestion
                    array_values = array.values[array.offset * array.type.list_size : (array.offset + len(array)) * array.type.list_size]
```

mariosasko (commented):
Thanks! One nit.

PS: feel free to link https://github.com/huggingface/datasets/issues/5477 to the PR

PR #6635, opened 2024-01-31
Fix missing info when loading some datasets from Parquet export

lhoestq:
Fix getting the info for script-based datasets with Parquet export with a single config not named "default".

E.g.

```python

b = load_dataset_builder("bookcorpus")
print(b.info.features)
# should print {'text': Value(dtype='string', id=None)}
```

I fixed this by setting the default config name when there is only one config.

### DIFF
--- a/src/datasets/utils/metadata.py
+++ b/src/datasets/utils/metadata.py
@@ -252,10 +252,8 @@ class MetadataConfigs(Dict[str, Dict[str, Any]]):
 
     def get_default_config_name(self) -> Optional[str]:
         default_config_name = None
-        if len(self) == 1:
-            return next(iter(self))
         for config_name, metadata_config in self.items():
-            if config_name == "default" or metadata_config.get("default"):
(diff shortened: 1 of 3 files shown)


src/datasets/utils/metadata.py
```diff
@@ -252,6 +252,8 @@ def to_dataset_card_data(self, dataset_card_data: DatasetCardData) -> None:
 
     def get_default_config_name(self) -> Optional[str]:
         default_config_name = None
+        if len(self) == 1:
+            return next(iter(self))
         for config_name, metadata_config in self.items():
             if config_name == "default" or metadata_config.get("default"):
```
albertvillanova, line 258: What about this refactoring to avoid having 2 returns?
```suggestion
        for config_name, metadata_config in self.items():
            if len(self) == 1 or config_name == "default" or metadata_config.get("default"):
```

albertvillanova (approved):
Thanks for the fix! Cool.

Just a nit formal comment below, but feel free to ignore it.

PR #6636, opened 2024-01-31
Faster column validation and reordering

psmyth94:
I work with bioinformatics data and often these tables have thousands and even tens of thousands of features. These tables are also accompanied by metadata that I do not want to pass in the model. When I perform `set_format('pt', columns=large_column_list)` , it can take several minutes before it finishes. The culprit is when the following check is performed: `any(col not in self._data.column_names for col in columns)`. Replacing this by  `set(columns) - (self._data.column_names)` is more efficient.


mariosasko (commented):
Thanks for working on this! 

Besides `Dataset.set_format`, we also use this inefficient check in the following methods: 
* `Dataset.select_columns`
* `Dataset.remove_columns`
* `Dataset.rename_columns`
* `Dataset.map` (`input_columns` and `remove_columns` checks)
* `IterableDataset.select_columns`

So, let's also update them in this PR to be consistent across the codebase.

psmyth94:
Thanks @mariosasko, I made the changes. However, I did some tests with `map` and I still saw that it took ~3.5 minutes per batch on 6000 features when using `dataset.map(lambda x: x, batched=True). From the profile, the culprits were mainly with `ArrowWriter.write_batch` and `ArrowWriter._build_writer`. The slow down from `_build_writer` is due to updating existing features with the inferred ones. I don't think this can be optimized any further, but fortunately, I can avoid this by setting the `features` in `map`. On the other hand, `write_batch` selects cols based on intersection and difference between schema names and example keys using two for loops. The same exists in `ArrowWriter.write_examples_on_file`. Optimizing the column selection using set operations effectively brings it from 3.5 minutes per batch down to 6 seconds per batch. Can we add these changes along with this PR?

src/datasets/arrow_writer.py
```diff
@@ -428,12 +428,17 @@ def write_examples_on_file(self):
...
+        if self.schema:
+            schema_names_set = set(self.schema.names)
+            current_example_keys_set = set(self.current_examples[0][0].keys())
+
+            common_cols = schema_names_set & current_example_keys_set
+            extra_cols = current_example_keys_set - schema_names_set
+
+            cols = list(common_cols) + list(extra_cols)
+        else:
+            cols = self.current_examples[0][0].keys()
```
mariosasko, line 440: Python sets (unlike dict keys) don't preserve the order, so we need to use lists to construct `common_cols` and `extra_cols`.

```suggestion
        if self.schema:
            schema_cols = set(self.schema.names)
            common_cols, extra_cols = [], []
            for col in self.current_examples[0][0]:
                if col in schema_cols:
                    common_cols.append(col)
                else:
                    extra_cols.append(col) 
            cols = common_cols + extra_cols
        else:
            cols = list(self.current_examples[0][0])
```

src/datasets/arrow_writer.py
```diff
@@ -538,12 +543,16 @@ def write_batch(
...
+        if self.schema:
+            schema_names_set = set(self.schema.names)
+            batch_examples_keys_set = set(batch_examples.keys())
+
+            common_cols = schema_names_set & batch_examples_keys_set
+            extra_cols = batch_examples_keys_set - schema_names_set
+
+            cols = list(common_cols) + list(extra_cols)
+        else:
+            cols = batch_examples.keys()
```
mariosasko, line 555: Same here:

```suggestion
        if self.schema:
            schema_cols = set(self.schema.names)
            common_cols, extra_cols = [], []
            for col in self.batch_examples:
                if col in schema_cols:
                    common_cols.append(col)
                else:
                    extra_cols.append(col) 
            cols = common_cols + extra_cols
        else:
            cols = list(self.batch_examples)
```

mariosasko (commented):
Indeed, let's also improve this column reordering logic to avoid quadratic time complexity. 


Two comments to fix the CI failure.

mariosasko (approved):
I took the liberty to apply the code suggestions so that we can include this PR in the next release.

PR #6646, opened 2024-02-07
Better multi-gpu example

lhoestq:
Use Qwen1.5-0.5B-Chat as an easy example for multi-GPU

the previous example was using a model for translation and the way it was setup was not really the right way to use the model.


albertvillanova (approved):
Thanks, better now.

mfidabel:
Thanks, I was needing this example today <3

PR #6647, opened 2024-02-07
Update loading.mdx to include "jsonl" file loading.

mosheber:
* A small update to the documentation, noting the ability to load jsonl files.

### DIFF
--- a/docs/source/loading.mdx
+++ b/docs/source/loading.mdx
@@ -130,6 +130,13 @@ JSON files are loaded directly with [`load_dataset`] as shown below:
 >>> dataset = load_dataset("json", data_files="my_file.json")
 ```
 
+In addition, you can also load "jsonl" files as well. It will load up every line as an entry in the dataset:
+
+```py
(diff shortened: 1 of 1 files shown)


albertvillanova (changes requested):
Thanks for adding the explicit loading command.

However, I would move it just below, where we present the JSON-Lines example.
- Maybe adding that this format is called JSON-Lines
- Add the example after the JSON-Lines data example

mosheber:
> Thanks for adding the explicit loading command.
> 
> However, I would move it just below, where we present the JSON-Lines example.
> 
> * Maybe adding that this format is called JSON-Lines
> * Add the example after the JSON-Lines data example
> 
> https://github.com/huggingface/datasets/blob/14d9afbb7ae1b787c450261ca0ff374551993031/docs/source/loading.mdx#L135-L138

Thank you @albertvillanova for the feedback! I moved the jsonl file loading example to a more appropriate location.

Issue #6663, opened 2024-02-15
`write_examples_on_file` and `write_batch` are broken in `ArrowWriter`


bryant1410:
### Describe the bug

`write_examples_on_file` and `write_batch` are broken in `ArrowWriter` since #6636. The order between the columns and the schema is not preserved anymore. So these functions don't work anymore unless the order happens to align well.

### Steps to reproduce the bug

Try to do `write_batch` with anything that has many columns, and it's likely to break.

### Expected behavior

I expect these functions to work, instead of it trying to cast a column to its incorrect type.

### Environment info

- `datasets` version: 2.17.0
- Platform: Linux-5.15.0-1040-aws-x86_64-with-glibc2.35
- Python version: 3.10.13
- `huggingface_hub` version: 0.19.4
- PyArrow version: 15.0.0
- Pandas version: 2.2.0
- `fsspec` version: 2023.10.0

mariosasko:
Thanks for reporting! I've left some comments on the PR on how to fix this recent change rather than reverting it.

bryant1410:
> Thanks for reporting! I've left some comments on the PR on how to fix this recent change rather than reverting it.

I feel that'd be good, but it'd be great to release a hotfix ASAP (a revert is a fast thing to do) so people can continue using this library and then focus on still applying the improvement.

albertvillanova:
Fixed by #6664

PR #6664, opened 2024-02-15
Revert the changes in `arrow_writer.py` from #6636

bryant1410:
#6636 broke `write_examples_on_file` and `write_batch` from the class `ArrowWriter`. I'm undoing these changes. See #6663.

Note the current implementation doesn't keep the order of the columns and the schema, thus setting a wrong schema for each column.

### DIFF
--- a/src/datasets/arrow_writer.py
+++ b/src/datasets/arrow_writer.py
@@ -426,8 +426,7 @@ class ArrowWriter:
         """Write stored examples from the write-pool of examples. It makes a table out of the examples and write it."""
         if not self.current_examples:
             return
-
-        # order the columns properly
+        # preserve the order the columns
         if self.schema:
             schema_cols = set(self.schema.names)
             examples_cols = self.current_examples[0][0].keys()  # .keys() preserves the order (unlike set)
@@ -545,7 +544,8 @@ class ArrowWriter:
         try_features = self._features if self.pa_writer is None and self.update_features else None
         arrays = []
         inferred_features = Features()
-         if self.schema:
+        # preserve the order the columns
+        if self.schema:
             schema_cols = set(self.schema.names)
             batch_cols = batch_examples.keys()  # .keys() preserves the order (unlike set)
             common_cols = [col for col in self.schema.names if col in batch_cols]
--- a/src/datasets/arrow_writer.py
+++ b/src/datasets/arrow_writer.py
@@ -428,18 +428,12 @@ class ArrowWriter:
             return
 
         # order the columns properly
-        if self.schema:
-            schema_cols = set(self.schema.names)
-            common_cols, extra_cols = [], []
-            for col in self.current_examples[0][0]:
-                if col in schema_cols:
-                    common_cols.append(col)
-                else:
-                    extra_cols.append(col)
-            cols = common_cols + extra_cols
-        else:
-            cols = list(self.current_examples[0][0])
-
+        cols = (
+            [col for col in self.schema.names if col in self.current_examples[0][0]]
+            + [col for col in self.current_examples[0][0].keys() if col not in self.schema.names]
+            if self.schema
+            else self.current_examples[0][0].keys()
+        )
         batch_examples = {}
         for col in cols:
             # We use row[0][col] since current_examples contains (example, key) tuples.
@@ -549,17 +543,12 @@ class ArrowWriter:
         try_features = self._features if self.pa_writer is None and self.update_features else None
         arrays = []
         inferred_features = Features()
-        if self.schema:
-            schema_cols = set(self.schema.names)
-            common_cols, extra_cols = [], []
-            for col in batch_examples:
-                if col in schema_cols:
-                    common_cols.append(col)
-                else:
-                    extra_cols.append(col)
-            cols = common_cols + extra_cols
-        else:
-            cols = list(batch_examples)
+        cols = (
+            [col for col in self.schema.names if col in batch_examples]
+            + [col for col in batch_examples.keys() if col not in self.schema.names]
+            if self.schema
+            else batch_examples.keys()
+        )
         for col in cols:
             col_values = batch_examples[col]
             col_type = features[col] if features else None
--- a/tests/test_arrow_writer.py
+++ b/tests/test_arrow_writer.py
@@ -87,7 +87,13 @@ def _check_output(output, expected_num_chunks: int):
 
 @pytest.mark.parametrize("writer_batch_size", [None, 1, 10])
 @pytest.mark.parametrize(
-    "fields", [None, {"col_1": pa.string(), "col_2": pa.int64()}, {"col_1": pa.string(), "col_2": pa.int32()}]
+    "fields",
+    [
+        None,
+        {"col_1": pa.string(), "col_2": pa.int64()},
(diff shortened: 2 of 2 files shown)


src/datasets/arrow_writer.py
```diff
@@ -428,18 +428,12 @@ def write_examples_on_file(self):
...
-            cols = common_cols + extra_cols
-        else:
-            cols = list(self.current_examples[0][0])
-
+        cols = (
+            [col for col in self.schema.names if col in self.current_examples[0][0]]
+            + [col for col in self.current_examples[0][0].keys() if col not in self.schema.names]
+            if self.schema
+            else self.current_examples[0][0].keys()
+        )
```
mariosasko, line 436:
```suggestion
        if self.schema:
            schema_cols = set(self.schema.names)
            examples_cols = self.current_examples[0][0].keys()
            common_cols = [col for col in self.schema.names if col in examples_cols]
            extra_cols = [col for col in examples_cols if col not in schema_cols]
            cols = common_cols + extra_cols
        else:
            cols = list(self.current_examples[0][0])
```

src/datasets/arrow_writer.py
```diff
@@ -549,17 +543,12 @@ def write_batch(
...
-                    extra_cols.append(col)
-            cols = common_cols + extra_cols
-        else:
-            cols = list(batch_examples)
+        cols = (
+            [col for col in self.schema.names if col in batch_examples]
+            + [col for col in batch_examples.keys() if col not in self.schema.names]
+            if self.schema
+            else batch_examples.keys()
+        )
```
mariosasko, line 551:
```suggestion
         if self.schema:
            schema_cols = set(self.schema.names)
            batch_cols = batch_examples.keys()
            common_cols = [col for col in self.schema.names if col in batch_cols]
            extra_cols = [col for col in batch_cols if col not in schema_cols]
            cols = common_cols + extra_cols
        else:
            cols = list(batch_examples)
```

mariosasko (commented):
Hi! We can't revert this as the "reverted" implementation has quadratic time complexity. Instead, let's fix it:

bryant1410:
> Hi! We can't revert this as the "reverted" implementation has quadratic time complexity. Instead, let's fix it:

I agree, but it's the implementation we have had so far. Why don't we:
1. Release a hotfix ASAP so people can continue using this library fine since AFAIU right now mostly writing examples for people is broken.
2. Then, focus on still applying the performance improvement.

src/datasets/arrow_writer.py
bryant1410: What about using a set also for `examples_cols`?

Why cast `cols` to a list? And wouldn't it be clearer if we used `.keys()` there as well?

(if you agree with any change, please send a new code snippet suggestion so when done I can commit it)

mariosasko: Unlike `set()`, `.keys` guarantees the same order of the elements as in the input dictionary.

mariosasko:
The fix is straightforward, so one patch release (after this PR is merged) is enough.

Btw, let's also add a test to `tests/test_arrow_writer.py` to avoid this issue in the future.

src/datasets/arrow_writer.py
mariosasko: > Why cast cols to a list? And wouldn't it be clearer if we used .keys() there as well?

To ensure `cols` is a list in both scenarios (this step is not required, but is usually a good practice to avoid type-checking errors, for instance)

bryant1410: > Unlike set(), .keys guarantees the same order of the elements as in the input dictionary.

What if you had 2 copies of it? One for checking (`col in examples_cols`), and another for iterating it.

bryant1410: > To ensure `cols` is a list in both scenarios (this step is not required, but is usually a good practice to avoid type-checking errors, for instance)

I feel like it's unnecessary because it's gonna do an extra memory allocation (which also takes time, depending on the size) and extra iteration, but in the end, `cols` will be iterated only once later.

bryant1410:
> Btw, let's also add a test to tests/test_arrow_writer.py to avoid this issue in the future.

Would you mind adding such test, as you're more familiar with the codebase?

src/datasets/arrow_writer.py
```diff
@@ -426,20 +426,15 @@ def write_examples_on_file(self):
...
             schema_cols = set(self.schema.names)
-            common_cols, extra_cols = [], []
-            for col in self.current_examples[0][0]:
-                if col in schema_cols:
-                    common_cols.append(col)
-                else:
-                    extra_cols.append(col)
+            examples_cols = self.current_examples[0][0].keys()  # .keys() preserves the order (unlike set)
+            common_cols = [col for col in self.schema.names if col in examples_cols]
+            extra_cols = [col for col in examples_cols if col not in schema_cols]
```
bryant1410, line 434: What I was trying to say is the following, to have a speedup:

```suggestion
            examples_cols = self.current_examples[0][0].keys()  # .keys() preserves the order (unlike set)
            examples_cols_set = set(examples_cols)
            common_cols = [col for col in self.schema.names if col in examples_cols_set]
            extra_cols = [col for col in examples_cols if col not in schema_cols]
```

src/datasets/arrow_writer.py
```diff
@@ -549,14 +544,12 @@ def write_batch(
...
-                if col in schema_cols:
-                    common_cols.append(col)
-                else:
-                    extra_cols.append(col)
+            batch_cols = batch_examples.keys()  # .keys() preserves the order (unlike set)
+            common_cols = [col for col in self.schema.names if col in batch_cols]
+            extra_cols = [col for col in batch_cols if col not in schema_cols]
             cols = common_cols + extra_cols
         else:
             cols = list(batch_examples)
```
bryant1410, line 555: I feel like we should really avoid this extra copy, especially if the inner iterable is large.

src/datasets/arrow_writer.py
mariosasko: There shouldn't be any difference in speed (when checking membership) between the built-in  `set` and the `keys` object.

src/datasets/arrow_writer.py
mariosasko: This negligible optimization (`write_examples_on_file` and `write_batch` are called every `writer_batch_size`) goes against good code practices.

We wouldn't use Python for this project if we wanted to optimize every aspect of the API.

mariosasko (approved):
Thanks! 

(The CI failure is not related to the changes)

src/datasets/arrow_writer.py
bryant1410: Oh, TIL!

src/datasets/arrow_writer.py
bryant1410: I'm not sure it's negligible. #6636's OP stated:

> I work with bioinformatics data and often these tables have thousands and even tens of thousands of features. 

We'd make a copy of tens of thousands of strings for every batch, for every processing step (e.g., a `map`).

And it's easy to remove (just `cols = batch_samples`, instead of copying it into a list).

Among other things, this library is about large data processing efficiency, so I think it'd be nice to consider it.

PR #6474, opened 2023-12-05
Deprecate Beam API and download from HF GCS bucket

mariosasko:
Deprecate the Beam API and download from the HF GCS bucked. 

TODO:
- [x] Convert the Beam-based [`wikipedia`](https://huggingface.co/datasets/wikipedia) to an Arrow-based dataset ([Hub PR](https://huggingface.co/datasets/wikipedia/discussions/19))
- [x] Make [`natural_questions`](https://huggingface.co/datasets/natural_questions) a no-code dataset ([Hub PR](https://huggingface.co/datasets/natural_questions/discussions/7))
- [x] Make [`wiki40b`](https://huggingface.co/datasets/wiki40b) a no-code dataset ([Hub PR](https://huggingface.co/datasets/wiki40b/discussions/5))
- [x] Make [`wiki_dpr`](https://huggingface.co/datasets/wiki_dpr) an Arrow-based dataset ([Hub PR](https://huggingface.co/datasets/wiki_dpr/discussions/14))

PR #6677, opened 2024-02-18
Pass through information about location of cache directory.

stridge-cruxml:
If cache directory is set, information is not passed through.
Pass download config in as an arg too.


mariosasko (approved):
Thanks! We will deprecate the GCS download in https://github.com/huggingface/datasets/pull/6474, but this is still a good catch nonetheless.

PR #6684, opened 2024-02-20
Improve error message for gated datasets on load

lewtun:
Internal Slack discussion: https://huggingface.slack.com/archives/C02V51Q3800/p1708424971135029

### DIFF
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -1807,10 +1807,10 @@ def dataset_module_factory(
                     msg = f"Dataset '{path}' doesn't exist on the Hub"
                     raise DatasetNotFoundError(msg + f" at revision '{revision}'" if revision else msg)
                 elif "401" in str(e):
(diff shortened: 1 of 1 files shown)


lhoestq:
Thank you ! Should we also add the link to the dataset page ?

lhoestq:
Looks like a test is failing: `test_load_dataset_cached_local_script `.

Apparently your new message is also shown for datasets that don't exist, which is maybe not ideal

lewtun:
Ah let me take a look!

lewtun:
> Looks like a test is failing: `test_load_dataset_cached_local_script `.
> 
> Apparently your new message is also shown for datasets that don't exist, which is maybe not ideal

Fixed by reverting the error message root + added a small clarifying part

lhoestq (approved):
Cool thanks !

PR #6687, opened 2024-02-22
fsspec: support fsspec>=2023.12.0 glob changes

### LINKED ISSUE #6644: Support fsspec 2023.12
Support fsspec 2023.12 by handling previous and new glob behavior.

### LINKED ISSUE #6645: Support fsspec 2024.2
Support fsspec 2024.2.

pmrowla:
- adds support for the `fs.glob` changes introduced in `fsspec==2023.12.0` and unpins the current upper bound

Should close #6644
Should close #6645

The `test_data_files` glob/pattern tests pass for me in:
- `fsspec==2023.10.0` (the pinned max version in datasets `main`)
- `fsspec==2023.12.0` (#6644)
- `fsspec==2024.2.0` (#6645)


setup.py
```diff
@@ -131,7 +131,7 @@
     "multiprocess",
     # to save datasets locally or on any filesystem
     # minimum 2023.1.0 to support protocol=kwargs in fsspec's `open`, `get_fs_token_paths`, etc.: see https://github.com/fsspec/filesystem_spec/pull/1143
-    "fsspec[http]>=2023.1.0,<=2023.10.0",
+    "fsspec[http]>=2023.1.0",
```
pmrowla, line 134: Was unsure if the preference here is to remove the upper bound or to increase the range to the current fsspec release (`<=2024.2.0`)

src/datasets/data_files.py
```diff
@@ -134,6 +144,23 @@ def sanitize_patterns(patterns: Union[Dict, List, str]) -> Dict[str, Union[List[
         return sanitize_patterns(list(patterns))
 
 
+def _expand_globstar(pattern: str) -> str:
```
lhoestq, line 147: I think we should fully align with the new fsspec logic, so ideally it would be great to remove the extra `_expand_globstar` logic.

Therefore it's fine to simply remove the test examples using unsupported patterns like `data/**`, `**train*` or `**test*`, and and update the remaining patterns that still use `**` like `DEFAULT_PATTERNS_ALL` and `METADATA_PATTERNS` (and some patterns from the new `KEYWORDS_IN_PATH_NAME_BASE_PATTERNS ` you added)

What do you think ?

pmrowla: This is fine with me, I was just going off of the original comment from https://github.com/huggingface/datasets/issues/6644 which said the intention was to continue supporting the old glob behavior

lhoestq: cc @albertvillanova IMO we can drop the `<prefix>**<suffix>` support, since it's been a behavior that only existed in `fsspec` and no other well known globbing tool afaik

I double checked and actually we just need to remove the test involving `**train*` and `**test*`.

No need to modify `data/**`, `DEFAULT_PATTERNS_ALL`, `METADATA_PATTERNS` or `KEYWORDS_IN_PATH_NAME_BASE_PATTERNS` as I said in my previous message.

pmrowla: @lhoestq I've pushed an update that removes the unsupported `**train*`/`**test*` tests, removes the `_expand_globstar` behavior, and simplifies the updated keyword patterns to
```
    KEYWORDS_IN_PATH_NAME_BASE_PATTERNS = [
        "**/{keyword}[{sep}]*",
        "**/{keyword}/**",
        "**/*[{sep}]{keyword}[{sep}]*",
        "**/*[{sep}]{keyword}[{sep}]*/**",
    ]
```

pmrowla:
Looking into the CI failure, this PR is incompatible with `huggingface-hub>=0.20.0`. It looks like there were several changes made to HfFileSystem in 0.20.0, @lhoestq any ideas on what the issue might be in particular?

pmrowla:
It looks like huggingface-hub's `HfFileSystem.glob` is broken for exact string matches (that don't contain glob wildcards) when combining `huggingface-hub>=0.20.0` and `fsspec>=2023.12.0`.

I did a quick test with huggingface-hub `main`, and adding this test case to `tests/test_hf_filesystem::HfFileSystemTests::test_glob` (https://github.com/huggingface/huggingface_hub/blob/main/tests/test_hf_file_system.py) passes with `fsspec==2023.10.0` and fails with `fsspec==2023.12.0`
```python
        self.assertEqual(
            sorted(self.hffs.glob(self.hf_path + "/.gitattributes")),
            sorted([self.hf_path + "/.gitattributes"]),
        )

```

the `hffs.glob()` call with a pattern that does not contain any wildcards returns an empty list:
```
E       AssertionError: Lists differ: [] != ['datasets/__DUMMY_TRANSFORMERS_USER__/rep[35 chars]tes']
E
E       Second list contains 1 additional elements.
E       First extra element 0:
E       'datasets/__DUMMY_TRANSFORMERS_USER__/repo-7d0ae9-17091013467064/.gitattributes'
E
E       - []
E       + ['datasets/__DUMMY_TRANSFORMERS_USER__/repo-7d0ae9-17091013467064/.gitattributes']
```

So it looks like the CI failure here isn't directly related to this PR. The failing patterns that don't contain any `*` wildcards are generated by `datasets` with or without this PR, but now that this PR installs the incompatible fsspec version, the underlying `HfFileSystem.glob()` call ends up failing.

lhoestq:
I just opened https://github.com/huggingface/huggingface_hub/pull/2056 to fix this.

Do you mind if I continue this PR to run the CI against this branch of `huggingface_hub` until the fix is released ?

lhoestq:
I just added two additional patterns to cover cases like `test-data/xxx.csv` and `data-test/xxx.csv`

lhoestq (approved):
Alright it looks all good now, thanks for the fix ! also cc @mariosasko

albertvillanova:
> EDIT: the fix has been released in `huggingface_hub` 0.21.2 - I removed my commits that were using `huggingface_hub@main`

Please note that people using `huggingface_hub` < 0.21.2 and latest `fsspec` will have issues when using `datasets`:
- https://github.com/huggingface/lighteval/actions/runs/8139147047/job/22241658122?pr=86
- https://github.com/huggingface/lighteval/pull/84

CC: @clefourrier

PR #6693, opened 2024-02-25
Update the print message for chunked_dataset in process.mdx

gzbfgjf2:
Update documentation to align with `Dataset.__repr__` change after #423


mariosasko (approved):
Good catch, thanks!

PR #6698, opened 2024-02-27
Faster `xlistdir`

mariosasko:
Pass `detail=False` to the `fsspec` `listdir`  to avoid unnecessarily fetching expensive metadata about the paths.


mariosasko:
CI failure is unrelated to the changes.

PR #6704, opened 2024-03-01
Improve default patterns resolution

mariosasko:
Separate the default patterns that match directories from the ones matching files and ensure directories are checked first (reverts the change from https://github.com/huggingface/datasets/pull/6244, which merged these patterns). Also, ensure that the glob patterns do not overlap to avoid duplicates in the result.

fix https://github.com/huggingface/datasets/issues/6259
fix https://github.com/huggingface/datasets/issues/6272

TODO:
- [ ] optimize repeated `LocalFileSystem.ls` calls when resolving the default patterns

### DIFF
--- a/src/datasets/data_files.py
+++ b/src/datasets/data_files.py
@@ -46,28 +46,48 @@ SPLIT_KEYWORDS = {
 }
 NON_WORDS_CHARS = "-._ 0-9"
 if config.FSSPEC_VERSION < version.parse("2023.9.0"):
-    KEYWORDS_IN_PATH_NAME_BASE_PATTERNS = ["{keyword}[{sep}/]**", "**[{sep}/]{keyword}[{sep}/]**"]
+    KEYWORDS_IN_FILENAME_BASE_PATTERNS = ["**[{sep}/]{keyword}[{sep}]*", "{keyword}[{sep}]*"]
+    KEYWORDS_IN_DIR_NAME_BASE_PATTERNS = [
+        "{keyword}/**",
+        "{keyword}[{sep}]*/**",
+        "**[{sep}/]{keyword}/**",
+        "**[{sep}/]{keyword}[{sep}]*/**",
+    ]
 elif config.FSSPEC_VERSION < version.parse("2023.12.0"):
-    KEYWORDS_IN_PATH_NAME_BASE_PATTERNS = ["{keyword}[{sep}/]**", "**/*[{sep}/]{keyword}[{sep}/]**"]
+    KEYWORDS_IN_FILENAME_BASE_PATTERNS = ["**/*[{sep}/]{keyword}[{sep}]*", "{keyword}[{sep}]*"]
+    KEYWORDS_IN_DIR_NAME_BASE_PATTERNS = [
+        "{keyword}/**/*",
+        "{keyword}[{sep}]*/**/*",
+        "**/*[{sep}/]{keyword}/**/*",
+        "**/*[{sep}/]{keyword}[{sep}]*/**/*",
+    ]
 else:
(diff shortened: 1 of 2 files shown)


lhoestq:
Awesome !

Note that it can still create duplicates if a path matches several dir patterns, e.g.

```
data/train-train/data/txt
```
matches two dir patterns:
```
**/{keyword}[{sep}]*/**
**/*[{sep}]{keyword}/**
```

mariosasko:
Yes, I didn't mention that case on purpose 🙂. One solution would be deprecating the `**/*[{sep}]{keyword}/**` pattern (and eventually removing). This way, the directory patterns would align more with the filename ones. Or do you think this is too big of a breaking change?

lhoestq:
I think it's too big of a breaking change yes :/ Though I think your approach is already a nice step in the right direction

mariosasko:
These changes to the `resolve_pattern` function lead to 20-30x faster local file resolution in my benchmarks.

lhoestq:
Nice ! Though since `fsspec` caches the filesystem, is there a risk when adding new files and reloading a dataset ?


```python
with open("my/local/dir/0000.txt", "w") as f:
    f.write("Hello there")
d1 = load_dataset("my/local/dir")
with open("my/local/dir/0001.txt", "w") as f:
    f.write("General Kenobi")
d2 = load_dataset("my/local/dir")
assert list(d1) != list(d2)
```

mariosasko:
Yes. But I think I have a solution for this.

mariosasko:
I'm not satisfied with the context manager approach...

A clean solution would require a bigger rewrite of the resolution logic (e.g., merging `get_data_patterns` and `DataFilesDict.from_patterns` into a `get_data_files` function that would match the paths using `fs.find` and `fsspec.utils.glob_translate` (available in `fsspec>=2023.12.0`))

The current changes make the local resolution 2-3x faster, which is good enough for now, I think.

lhoestq (approved):
Sounds good

src/datasets/data_files.py
```diff
@@ -46,36 +46,57 @@ class EmptyDatasetError(FileNotFoundError):
 }
 NON_WORDS_CHARS = "-._ 0-9"
 if config.FSSPEC_VERSION < version.parse("2023.9.0"):
-    KEYWORDS_IN_PATH_NAME_BASE_PATTERNS = ["{keyword}[{sep}/]**", "**[{sep}/]{keyword}[{sep}/]**"]
+    KEYWORDS_IN_FILENAME_BASE_PATTERNS = ["**[{sep}/]{keyword}[{sep}]*", "{keyword}[{sep}]*"]
```
albertvillanova, line 49: I think this is a breaking change impacting all uses of the variable `KEYWORDS_IN_PATH_NAME_BASE_PATTERNS`. See: https://github.com/huggingface/dataset-viewer/actions/runs/8753796799/job/24024224560?pr=2740

mariosasko: `dataset-viewer` seems to be the only repo on GH using `KEYWORDS_IN_PATH_NAME_BASE_PATTERNS`...

But I think it's okay to add it back if it's hard to fix `datasets-viewer`'s CI otherwise.

lhoestq: no need to add it back imo, we can just use the new variable in dataset-viewer

albertvillanova: I am fixing this in dataset-viewer: https://github.com/huggingface/dataset-viewer/pull/2740#issuecomment-2071741972

lhoestq:
I opened https://github.com/huggingface/datasets/pull/6828 to add proper Path support to save_to_disk / load_from_disk

PR #6714, opened 2024-03-04
Expand no-code dataset info with datasets-server info

mariosasko:
E.g., to have info about a dataset's number of examples for more informative TQDM bars.


lhoestq (approved):
Nice ! Pretty useful for big datasets

PR #6715, opened 2024-03-04
Fix sliced ConcatenationTable pickling with mixed schemas vertically

lhoestq:
A sliced + pickled ConcatenationTable could end up with a different schema than the original schema, if the slice only contains blocks with only a subset of the columns.

This can lead to issues when saving datasets from a concatenation of datasets with mixed schemas

Reported in https://discuss.huggingface.co/t/datasetdict-save-to-disk-with-num-proc-1-seems-to-hang-with-error/75595

### DIFF
--- a/src/datasets/table.py
+++ b/src/datasets/table.py
@@ -1311,11 +1311,20 @@ class ConcatenationTable(Table):
                     )
 
     def __getstate__(self):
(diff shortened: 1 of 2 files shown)


albertvillanova (approved):
Thank you.

PR #6727, opened 2024-03-10
Using a registry instead of calling globals for fetching feature types

psmyth94:
Hello,

When working with bio-data, each feature often has metadata associated with it (e.g. species, lineage, snp position, etc). To store this, I like to use the feature classes with the added `metadata` attribute. However, when saving or loading with custom features, you get an error since that class doesn't exist in the global namespace in `datasets.features.features`. Take for example,

```python
from dataclasses import dataclass, field
from datasets import Dataset
from datasets.features.features import Value, Features

@dataclass
class FeatureA(Value):
    metadata: dict = field(default=dict)
    _type: str = field(default="FeatureA", init=False, repr=False)

@dataclass
class FeatureB(Value):
    metadata: dict = field(default_factory=dict)
    _type: str = field(default="FeatureB", init=False, repr=False)

test_data = {
    "a": [1, 2, 3],
    "b": [4, 5, 6],
}
test_data = Dataset.from_dict(
    test_data, 
    features=Features({
        "a": FeatureA("int32", metadata={"species": "lactobacillus acetotolerans"}),
        "b": FeatureB("int32", metadata={"species": "lactobacillus iners"}),
    })
)

# returns an error since FeatureA and FeatureB are not in the global namespace
test_data.save_to_disk('./test_data')
```

    Saving the dataset (0/1 shards):   0%|          | 0/3 [00:00<?, ? examples/s]

    KeyError                                  Traceback (most recent call last)

    Cell In[2], line 28
         19 test_data = Dataset.from_dict(
         20     test_data, 
         21     features=Features({
       (...)
         24     })
         25 )
         27 # returns an error since FeatureA and FeatureB are not in the global namespace
    ---> 28 test_data.save_to_disk('./test_data')
    ...
    File ~\Documents\datasets\src\datasets\features\features.py:1361, in generate_from_dict(obj)
       1359     return {key: generate_from_dict(value) for key, value in obj.items()}
       1360 obj = dict(obj)
    -> 1361 class_type = globals()[obj.pop("_type")]
       1363 if class_type == Sequence:
       1364     return Sequence(feature=generate_from_dict(obj["feature"]), length=obj.get("length", -1))
    

    KeyError: 'FeatureA'

We can avoid this by having a registry (like formatters) and doing

```python
from datasets.features.features import register_feature
register_feature(FeatureA, "FeatureA")
register_feature(FeatureB, "FeatureB")
test_data.save_to_disk('./test_data')
```

    Saving the dataset (1/1 shards): 100%|------| 3/3 [00:00<00:00, 211.13 examples/s]

and loading from disk returns with all metadata information

```python
from datasets import load_from_disk
test_data = load_from_disk('./test_data')
test_data.features
```

    {'a': FeatureA(dtype='int32', id=None, metadata={'species': 'lactobacillus acetotolerans'}),
     'b': FeatureB(dtype='int32', id=None, metadata={'species': 'lactobacillus iners'})}


src/datasets/features/features.py
```diff
@@ -1342,6 +1342,24 @@ def decode_nested_example(schema, obj, token_per_repo_id: Optional[Dict[str, Uni
...
 
+_FEATURE_TYPES: Dict[Optional[str], FeatureType] = {}
+
+
+def register_feature(
+    feature_cls: type,
+    feature_type: Optional[str],
+):
+    """
+    Register a Feature object using a name and class.
```
lhoestq, line 1353: Maybe mention that this feature is experimental in the docstring ? In case we want to improve/change the API at one point

src/datasets/features/features.py
```diff
@@ -1360,7 +1378,11 @@ def generate_from_dict(obj: Any):
     if "_type" not in obj or isinstance(obj["_type"], dict):
         return {key: generate_from_dict(value) for key, value in obj.items()}
     obj = dict(obj)
-    class_type = globals()[obj.pop("_type")]
+    _type = obj.pop("_type")
+    class_type = _FEATURE_TYPES.get(_type, None) or globals().get(_type, None)
+
+    if class_type is None:
+        raise ValueError(f"Feature type '{_type}' not found. Available feature types: {list(_FEATURE_TYPES.keys())}")
```
lhoestq, line 1385: Maybe populate _FEATURE_TYPES with the existing features so that the error message doesn't show an empty list by default ?

src/datasets/features/features.py
```diff
@@ -1342,6 +1342,24 @@ def decode_nested_example(schema, obj, token_per_repo_id: Optional[Dict[str, Uni
     return obj
 
 
+_FEATURE_TYPES: Dict[Optional[str], FeatureType] = {}
```
lhoestq, line 1345: str is enough no ?

```suggestion
_FEATURE_TYPES: Dict[str, FeatureType] = {}
```

src/datasets/features/features.py
```diff
@@ -1342,6 +1342,24 @@ def decode_nested_example(schema, obj, token_per_repo_id: Optional[Dict[str, Uni
...
+
+
+def register_feature(
+    feature_cls: type,
+    feature_type: Optional[str],
```
lhoestq, line 1350:
```suggestion
    feature_type: str,
```

lhoestq (commented):
Cool ! Though for now I'd keep this feature experimental if it's fine for you.

I just added a few comments:

src/datasets/features/features.py
psmyth94: should I use the experimental decorator also?

lhoestq: Yes ideally :)

src/datasets/features/features.py
psmyth94: Yeah, I had `{k: v for k, v in globals().items() if isinstance(v, FeatureType)}` to avoid hard coding it. Subscripted don't work in 3.8 unfortunately. I'll just hard code it since it's experimental anyways.

psmyth94:
looks like some files are missing in your google storage

lhoestq:
cc @mariosasko is it related to https://github.com/huggingface/datasets/pull/6474 ? The files should ideally not move for backward compatibility anyway

mariosasko:
@lhoestq All the files are still there.

The problem is that the `natural_questions` is now a no-code dataset, so the test's paths are no longer correct (unless the revision is pinned to the previous version). 

@psmyth94 This has been fixed on `main`, so you can make the CI tests green with the following:
```python
git remote add upstream https://github.com/huggingface/datasets.git
git pull upstream main
git push
```

lhoestq:
Thank you @mariosasko ! I'm updating this branch if you don't mind @psmyth94

lhoestq (approved):
Looks all good now, thanks !

PR #6730, opened 2024-03-12
Deprecate Pandas builder

mariosasko:
The Pandas packaged builder is undocumented and relies on `pickle` to read the data, making it **unsafe**. Moreover, I haven't seen a single instance of this builder being used (not even using the GH/Hub search), so we should deprecate it.


lhoestq (approved):
Sounds good to me

Issue #6252, opened 2023-09-21
exif_transpose not done to Image (PIL problem)


rhajou:
### Feature request

I noticed that some of my images loaded using PIL have some metadata related to exif that can rotate them when loading.
Since the dataset.features.Image uses PIL for loading, the loaded image may be rotated (width and height will be inverted) thus for tasks as object detection and layoutLM this can create some inconsistencies (between input bboxes and input images). 

For now there is no option in datasets.features.Image to specify that. We need to do the following when preparing examples (when preparing images for training, test or inference): 
```
from PIL import Image, ImageOps 
pil = ImageOps.exif_transpose(pil)
```

reference: https://stackoverflow.com/a/63950647/5720150 

Is it possible to add this by default to the datasets.feature.Image ? or to add the option to do the ImageOps.exif_transpose?

Thank you

### Motivation

Prevent having inverted data related to exif metadata that may affect object detection tasks

### Your contribution

Changing in datasets.featrues.Image I can help with that.

mariosasko:
Indeed, it makes sense to do this by default. 

In the meantime, you can use `.with_transform` to transpose the images when accessing them:

```python
import PIL

def exif_transpose_transform(batch):
  batch["image"] = [ImageOps.exif_transpose(image) for image in batch["image"]]
  return batch

dataset.with_transform(exif_transpose_transform)
```

mariosasko:
This operation sets some `Image` attributes to `None` (`.format`, `.filename`, etc.), causing our tests to fail, so I think we should wait for Datasets 3.0 to make this change. In version 3.0, storing image paths will be replaced by embedding image bytes, so there will be fewer instances where we use the `.filename` attribute.

PR #6739, opened 2024-03-18
Transpose images with EXIF Orientation tag

mariosasko:
Closes https://github.com/huggingface/datasets/issues/6252


lhoestq (approved):

Seas0:
This commit seems to crash the whole program if the EXIF tag is contaminated like that in #7668 (which is common, due to str codec misconfiguration) and thus I'd like to suggest making the parsing feature optional, or just implement some error handling.

PR #6743, opened 2024-03-19
Allow null values in dict columns

### LINKED ISSUE #6738: Dict feature is non-nullable while nested dict feature is
When i try to create a `Dataset` object with None values inside a dict column, like this: 
```python

Dataset.from_dict(
    {
        "dict": [{"a": 0, "b": 0}, None],
    }, features=Features(
        {"dict": {"a": Value("int16"), "b": Value("int16")}}
    )
)
```
i get `ValueError: Got None but expected a dictionary instead`.

At the same time, having None in _nested_ dict feature works, for example, this doesn't throw any errors:

```python
from datasets import Dataset, Features, Value, Sequence

dataset = Dataset.from_dict(
    {
        "list_dict": [[{"a": 0, "b": 0}], None],
        "sequence_dict": [[{"a": 0, "b": 0}], None],
    }, features=Features({
        "list_dict": [{"a": Value("int16"), "b": Value("int16")}],
        "sequence_dict": Sequence({"a": Value("int16"), "b": Value("int16")}),
    })
)
```
Other types of features also seem to be nullable (but I haven't checked all of them).

Version of `datasets` is the latest atm (2.18.0)

Is this an expected behavior or a bug?

mariosasko:
Fix #6738

### DIFF
--- a/src/datasets/features/features.py
+++ b/src/datasets/features/features.py
@@ -1937,7 +1937,7 @@ class Features(dict):
             `list[Any]`
         """
         column = cast_to_python_objects(column)
-        return [encode_nested_example(self[column_name], obj) for obj in column]
+        return [encode_nested_example(self[column_name], obj, level=1) for obj in column]
 
     def encode_batch(self, batch):
         """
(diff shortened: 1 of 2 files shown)


lhoestq (approved):
Thanks for the quick fix !

severo:
notify https://huggingface.co/datasets/chaoyi-wu/PMC-Inline/discussions/1 once it's merged in dataset-viewer

PR #6751, opened 2024-03-23
Use 'with' operator for some download functions

Moisan:
Some functions in `streaming_download_manager.py` are not closing the file they open which lead to `Unclosed file` warnings in our code. This fixes a few of them.


Moisan:
I was mistaken on the intent of those functions, closing the PR.

PR #6762, opened 2024-03-28
Allow polars as valid output type

psmyth94:
I was trying out polars as an output for a map function and found that it wasn't a valid return type in `validate_function_output`. Thought that we should accommodate this by creating and adding it to the `allowed_processed_input_types` variable.

### DIFF
--- a/tests/test_arrow_dataset.py
+++ b/tests/test_arrow_dataset.py
@@ -1535,6 +1535,7 @@ class BaseDatasetTest(TestCase):
             with self._create_dummy_dataset(in_memory, tmp_dir) as dset:
                 self.assertRaises(ValueError, dset.map, func_return_multi_row_pd_dataframe)
 
+    @require_polars
     def test_map_return_pl_dataframe(self, in_memory):
         import polars as pl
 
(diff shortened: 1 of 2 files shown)


lhoestq (commented):
Cool ! Sorry for the late review. Can you add a test to make sure map() with a polars-output function works ?
You can place the test in test_arrow_dataset.py

psmyth94:
Hello @lhoestq, I added the test and modified [update_data](https://github.com/huggingface/datasets/blob/bececdac927160b5c7e883736d7cc79d5699ad0a/src/datasets/arrow_dataset.py#L3437) to include `polars` as an updatable type. Although, it seems pretty redundant to do the type checks both before `validate_function_output` and then immediately afterward within the call stack. Could consider adding `allowable_types` in `validation_function_output`.

tests/test_arrow_dataset.py
```diff
@@ -1528,6 +1535,47 @@ def func_return_multi_row_pd_dataframe(x):
             with self._create_dummy_dataset(in_memory, tmp_dir) as dset:
                 self.assertRaises(ValueError, dset.map, func_return_multi_row_pd_dataframe)
 
+    def test_map_return_pl_dataframe(self, in_memory):
```
lhoestq, line 1538:
```suggestion
    @require_polars
    def test_map_return_pl_dataframe(self, in_memory):
```

lhoestq (approved):
Cool ! thanks for the additions :)

It looks all good to me as is. Not a big deal to have it at two places, since it's not for the same objective (check whether the type is a data update type, and whether the type is a valid type)

PR #6781, opened 2024-04-05
Remove get_inferred_type from ArrowWriter write_batch

Modexus:
Inferring the type seems to be unnecessary given that the pyarrow array has already been created.
Because pyarrow array creation is sometimes extremely slow this doubles the time write_batch takes.


Modexus:
Close in favor of #6786.

Issue #6782, opened 2024-04-05
Map/Saving Image from external filepath extremely slow


Modexus:
### Describe the bug

Operations that save an image from a path into parquet are very slow.
I believe the reason for this is that the image data (`numpy`) is converted into `pyarrow` format but then back to python using `.pylist()` before being converted to a numpy array again.

`pylist` is already slow but used on a multi-dimensional numpy array such as an image it takes a very long time.

From the trace below we can see that `__arrow_array__` takes a long time.
It is currently also called in `get_inferred_type`, this should be removable #6781 but doesn't change the underyling issue.

The conversion to `pyarrow` and back also leads to the `numpy` array having type `int64` which causes a warning message because the image type excepts `uint8`.
However, originally the `numpy` image array was in `uint8`.

### Steps to reproduce the bug

```python
from PIL import Image
import numpy as np
import datasets
import cProfile

image = Image.fromarray(np.random.randint(0, 255, (2048, 2048, 3), dtype=np.uint8))
image.save("test_image.jpg")

ds = datasets.Dataset.from_dict(
    {"image": ["test_image.jpg"]},
    features=datasets.Features({"image": datasets.Image(decode=True)}),
)

# load as numpy array, e.g. for further processing with map
# same result as map returning numpy arrays
ds.set_format("numpy")

cProfile.run("ds.map(writer_batch_size=1, load_from_cache_file=False)", "restats")
```
```bash
Fri Apr  5 14:56:17 2024    restats

         66817 function calls (64992 primitive calls) in 33.382 seconds

   Ordered by: cumulative time
   List reduced from 1073 to 20 due to restriction <20>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     46/1    0.000    0.000   33.382   33.382 {built-in method builtins.exec}
        1    0.000    0.000   33.382   33.382 <string>:1(<module>)
        1    0.000    0.000   33.382   33.382 arrow_dataset.py:594(wrapper)
        1    0.000    0.000   33.382   33.382 arrow_dataset.py:551(wrapper)
        1    0.000    0.000   33.379   33.379 arrow_dataset.py:2916(map)
        4    0.000    0.000   33.327    8.332 arrow_dataset.py:3277(_map_single)
        1    0.000    0.000   33.311   33.311 arrow_writer.py:465(write)
        2    0.000    0.000   33.311   16.656 arrow_writer.py:423(write_examples_on_file)
[...]

PR #6786, opened 2024-04-05
Make Image cast storage faster

Modexus:
PR for issue #6782.
Makes `cast_storage` of the `Image` class faster by removing the slow call to `.pylist`.
Instead directly convert each `ListArray` item to either `Array2DExtensionType` or `Array3DExtensionType`.

This also preserves the `dtype` removing the warning if the array is already `uint8`.

### DIFF
--- a/src/datasets/arrow_writer.py
+++ b/src/datasets/arrow_writer.py
@@ -561,8 +561,9 @@ class ArrowWriter:
             else:
                 col_try_type = try_features[col] if try_features is not None and col in try_features else None
                 typed_sequence = OptimizedTypedSequence(col_values, type=col_type, try_type=col_try_type, col=col)
-                arrays.append(pa.array(typed_sequence))
-                inferred_features[col] = typed_sequence.get_inferred_type()
+                array = pa.array(typed_sequence)
+                arrays.append(array)
+                inferred_features[col] = generate_from_arrow_type(array.type)
         schema = inferred_features.arrow_schema if self.pa_writer is None else self.schema
         pa_table = pa.Table.from_arrays(arrays, schema=schema)
         self.write_table(pa_table, writer_batch_size)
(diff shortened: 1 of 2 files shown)


lhoestq:
Hi ! Thanks for diving into this, this conversion to python lists is indeed quite slow.

Array2DExtensionType and Array3DExtensionType currently rely on pyarrow lists, but we will soon modify them to use FixedShapeTensorArray instead which is more efficient (e.g. doesn't need to store an offset for each value). So ideally it would be cool to speed this code up without using those extension types or it will be blocking to improve Array2DExtensionType and Array3DExtensionType.

If I understand correctly you just need the logic from ArrayExtensionArray.to_numpy ? If so feel free to make a separate function and ArrayExtensionArray.to_numpy can call it

Modexus:
Hey! I didn't have time to look into this but I just stumbled upon another problem. 
While my fix kind of made it usable I now pre-embedded the images and even as Array3D they are really slow to load. 
Don't think this can be resolved with using ArrayExtensionArray.to_numpy.

I think actually making the Array3DExtensionType faster would probably resolve both issues as you mentioned.
Is there an update on using FixedShapeTensorArray?
I'd gladly help implementing/testing it if there is some outline how to do it.

Modexus:
Thanks, I have looked into this and have a working solution at least for my specific case.
But I had quite a few issues along the way that are not solved nicely.
It follows your suggestion though internally it is then just a fixed_shape_tensor as there is no ExtensionType anymore.

Hopefully, I can create a separate PR with these changes soon.

Modexus:
I have run into some issues, notably I don't think `FixedShapeTensorArray` is completely supported by `pandas `and `polars`.
Well it seems to work for `pandas `but one loses the actual shape of the extension.
`Polars `just throws an error and this cannot be changed with `schema_overrides` as they are applied after.

I have tried to somehow cast the `FixedShapeTensorArray` to something else like a nested FixedSizeLists, however I have not found a clean solution to do that.
If somebody has a clean solution to cast it to something such that the shape survives the roundtrip to `pandas`/`polars `and back, it may be possible.

lhoestq:
Can we start using FixedShapeTensor or FixedSizeList even if pandas/polars don't support them fully yet ?

We would still get the benefit of optimized conversion to numpy

PR #6794, opened 2024-04-09
Multithreaded downloads

lhoestq:
...for faster dataset download when there are many many small files (e.g. imagefolder, audiofolder)

### Behcnmark

for example on [lhoestq/tmp-images-writer_batch_size](https://hf.co/datasets/lhoestq/tmp-images-writer_batch_size) (128 images)

|    | duration of the download step in `load_dataset()` |
|--| ----------------------------------------------------------------------|
| Before | 58s |
| Now | 3s |

This should fix issues with the Dataset Viewer taking too much time to show up for imagefolder/audiofolder datasets.

### Implementation details

The main change is in the `DownloadManager`:

```diff
- download_func = partial(self._download, download_config=download_config)
+ download_func = partial(self._download_batched, download_config=download_config)
downloaded_path_or_paths = map_nested(
    download_func,
    url_or_urls,
    map_tuple=True,
    num_proc=download_config.num_proc,
    desc="Downloading data files",
+   batched=True,
+   batch_size=-1,
)
```

and `_download_batched` is a multithreaded function.

I only enable multithreading if there are more than 16 files and files are small though, otherwise the progress bar that counts the number of downloaded files is not fluid (updating when a big batch of big files are done downloading). To do so I simply check if the first file is smaller than 20MB.

I also had to tweak `map_nested` to support batching. In particular it slices the data correctly if the user also enables multiprocessing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
         if: ${{ matrix.os == 'ubuntu-latest' }}
         run: echo "installing pinned version of setuptools-scm to fix seqeval installation on 3.7" && pip install "setuptools-scm==6.4.2"
       - name: Install uv
-        run: pip install uv==0.1.29
+        run: pip install --upgrade uv
       - name: Install dependencies
         run: |
           uv pip install --system "datasets[tests,metrics-tests] @ ."
@@ -89,7 +89,7 @@ jobs:
       - name: Upgrade pip
         run: python -m pip install --upgrade pip
       - name: Install uv
-        run: pip install uv==0.1.29
+        run: pip install --upgrade uv
       - name: Install dependencies
         run: uv pip install --system "datasets[tests] @ ."
       - name: Test with pytest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
         if: ${{ matrix.os == 'ubuntu-latest' }}
         run: echo "installing pinned version of setuptools-scm to fix seqeval installation on 3.7" && pip install "setuptools-scm==6.4.2"
       - name: Install uv
-        run: pip install --upgrade uv
+        run: pip install uv==0.1.29
       - name: Install dependencies
         run: |
           uv pip install --system "datasets[tests,metrics-tests] @ ."
@@ -89,7 +89,7 @@ jobs:
       - name: Upgrade pip
         run: python -m pip install --upgrade pip
       - name: Install uv
-        run: pip install --upgrade uv
+        run: pip install uv==0.1.29
       - name: Install dependencies
         run: uv pip install --system "datasets[tests] @ ."
       - name: Test with pytest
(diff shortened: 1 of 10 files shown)


lhoestq:
CI is failing because of the missing parquet export of one test dataset, PR to fix this at https://github.com/huggingface/dataset-viewer/pull/2689

.github/workflows/ci.yml
```diff
@@ -54,7 +54,7 @@ jobs:
         if: ${{ matrix.os == 'ubuntu-latest' }}
         run: echo "installing pinned version of setuptools-scm to fix seqeval installation on 3.7" && pip install "setuptools-scm==6.4.2"
       - name: Install uv
-        run: pip install --upgrade uv
+        run: pip install uv==0.1.29
```
mariosasko, line 57: Would remove the pin to be consistent with `huggingface_hub` and `diffusers`:
```suggestion
```

(we don't use `uv`'s advanced/experimental features, so a breaking change here is unlikely)

.github/workflows/ci.yml
```diff
@@ -89,7 +89,7 @@ jobs:
       - name: Upgrade pip
         run: python -m pip install --upgrade pip
       - name: Install uv
-        run: pip install --upgrade uv
+        run: pip install uv==0.1.29
```
mariosasko, line 92: Same here:

```suggestion
```

src/datasets/data_files.py
```diff
@@ -544,8 +544,8 @@ def _get_single_origin_metadata(
 
 def _get_origin_metadata(
     data_files: List[str],
-    max_workers=64,
     download_config: Optional[DownloadConfig] = None,
+    max_workers: int = 16,
```
mariosasko, line 548: Maybe this should be a config variable (which we would also use in `DownloadManager`)

src/datasets/download/download_manager.py
```diff
@@ -451,6 +458,43 @@ def download(self, url_or_urls):
...
+                position=multiprocessing.current_process()._identity[-1]  # contains the ranks of subprocesses
+                if os.environ.get("HF_DATASETS_STACK_MULTIPROCESSING_DOWNLOAD_PROGRESS_BARS") == "1"
+                and multiprocessing.current_process()._identity
+                else None,
+                max_workers=max_workers,
+                tqdm_class=tqdm,
+            )
+        else:
+            return [
+                self._download(url_or_filename, download_config=download_config)
```
mariosasko, line 494: Would rename this method to `_download_single`

.github/workflows/ci.yml
lhoestq: I had pinned it because 0.1.30 had bugs - I'll see if 0.1.31 has fixed them

mariosasko: It's been fixed in https://github.com/astral-sh/uv/issues/2941 :)

lhoestq:
I took your comments into account :) lmk what you think

mariosasko (approved):
LGTM!

PR #6795, opened 2024-04-09
Add CLI function to convert script-dataset to Parquet

### LINKED ISSUE #6690: Add function to convert a script-dataset to Parquet
Add function to convert a script-dataset to Parquet and push it to the Hub, analogously to the Space: "Convert a Hugging Face dataset to Parquet"

albertvillanova:
Close #6690.

### DIFF
--- a/src/datasets/commands/convert_to_parquet.py
+++ b/src/datasets/commands/convert_to_parquet.py
@@ -91,46 +91,39 @@ def infer_pr(dataset_id, token=None):
 def delete_files(dataset_id, revision=None, token=None):
     dataset_name = dataset_id.split("/")[-1]
     hf_api = HfApi(token=token)
-    try:
-        hf_api.delete_file(
-            f"{dataset_name}.py",
-            dataset_id,
-            repo_type="dataset",
-            revision=revision,
-            commit_message="Delete loading script",
-        )
-    except EntryNotFoundError:
-        pass
-    try:
-        hf_api.delete_file(
-            "dataset_infos.json",
-            dataset_id,
-            repo_type="dataset",
-            revision=revision,
-            commit_message="Delete legacy dataset_infos.json",
-        )
-    except EntryNotFoundError:
-        pass
     repo_files = hf_api.list_repo_files(
         dataset_id,
         repo_type="dataset",
     )
-    if ".gitattributes" in repo_files:
-        repo_files.remove(".gitattributes")
-    if "README.md" in repo_files:
-        repo_files.remove("README.md")
-    if f"{dataset_name}.py" in repo_files:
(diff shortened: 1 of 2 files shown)


albertvillanova:
@huggingface/datasets once this PR is merged, I would suggest making a release. Do you agree?

src/datasets/commands/convert_to_parquet.py
```diff
@@ -0,0 +1,143 @@
...
+        print(f"{configs = }")
+        default_config = get_dataset_default_config_name(dataset_id, token=token, revision=revision)
+        print(f"{default_config = }")
+        if default_config:
+            config = default_config
+            configs.remove(default_config)
+        else:
+            config = configs.pop(0)
+        print(f"{config = }")
+        dataset = load_dataset(dataset_id, config, revision=revision)
```
lhoestq, line 53: Maybe add `trust_remote_code` ?

(also in get_dataset_config_names and get_dataset_default_config_name)

lhoestq (approved):
LGTM :)

src/datasets/commands/convert_to_parquet.py
albertvillanova: Yes, otherwise we are raising warning now and eventually raise an error.

albertvillanova: Done.

src/datasets/commands/convert_to_parquet.py
```diff
@@ -0,0 +1,154 @@
...
+            dataset = load_dataset(dataset_id, config, revision=revision, trust_remote_code=trust_remote_code)
+            dataset.push_to_hub(
+                dataset_id,
+                config_name=config,
+                commit_message=f"Add {config} data files",
+                revision=pr_revision,
+                token=token,
+            )
+            time.sleep(5)
+        delete_files(dataset_id, revision=pr_revision, token=token)
```
severo, line 90: When providing support, we sometimes suggest that users store their script in a `script` branch. What do you think of this alternative to deleting the files?

TBH, I think it's OK to delete the files and to let the users move their script to another branch manually if they want to. Just wanted to check it's not an oversight

albertvillanova: If we agree this is interesting, I could have a look at it, using the huggingface_hub client `create_branch` function, and see if the logic is easily implementable.

albertvillanova: I agree many users will like to keep the script somewhere for future use/updates.

albertvillanova: Done. Thanks @severo for the proposal.

Issue #6805, opened 2024-04-12
Batched mapping of existing string column casts boolean to string


starmpcc:
### Describe the bug

Let the dataset contain a column named 'a', which is of the string type.
If 'a' is converted to a boolean using batched mapping, the mapper automatically casts the boolean to a string (e.g., True -> 'true').
It only happens when the original column and the mapped column name are identical.

Thank you!

### Steps to reproduce the bug

```python
from datasets import Dataset

dset = Dataset.from_dict({'a': ['11', '22']})
dset = dset.map(lambda x: {'a': [True for _ in x['a']]}, batched=True)
print(dset['a'])
```
```
> ['true', 'true']
```

### Expected behavior

[True, True]

### Environment info

- `datasets` version: 2.18.0
- Platform: Linux-5.4.0-148-generic-x86_64-with-glibc2.31
- Python version: 3.10.13
- `huggingface_hub` version: 0.21.4
- PyArrow version: 15.0.2
- Pandas version: 2.2.1
- `fsspec` version: 2023.12.2

Modexus:
This seems to be hardcoded behavior in table.py `array_cast`.
```python
if (
            not allow_number_to_str
            and pa.types.is_string(pa_type)
            and (pa.types.is_floating(array.type) or pa.types.is_integer(array.type))
        ):
            raise TypeError(
                f"Couldn't cast array of type {array.type} to {pa_type} since allow_number_to_str is set to {allow_number_to_str}"
            )
        if pa.types.is_null(pa_type) and not pa.types.is_null(array.type):
            raise TypeError(f"Couldn't cast array of type {array.type} to {pa_type}")
        return array.cast(pa_type)
```
where floats and integers are not cast to string but booleans are.
Maybe this should be extended to booleans?

mariosasko:
Thanks for reporting! @Modexus Do you want to open a PR with the suggested fix?

Modexus:
I'll gladly create a PR but not sure what the behavior should be.

Should a value returned from map be cast to the current feature?
At the moment this seems very inconsistent since `datetime `is also cast (this would only fix `boolean`) but nested structures are not.

```python
dset = Dataset.from_dict({"a": ["Hello world!"]})
dset = dset.map(lambda x: {"a": date(2021, 1, 1)})
# dset[0]["a"] == '2021-01-01'
```
```python
dset = Dataset.from_dict({"a": ["Hello world!"]})
dset = dset.map(lambda x: {"a": [True]})
[...]

PR #6811, opened 2024-04-15
add allow_primitive_to_str and allow_decimal_to_str instead of allow_number_to_str

Modexus:
PR for #6805

### DIFF
--- a/src/datasets/arrow_writer.py
+++ b/src/datasets/arrow_writer.py
@@ -243,7 +243,9 @@ class TypedSequence:
                             cast_to_python_objects(data, only_1d_for_numpy=True, optimize_list_casting=False)
                         )
                         if type is not None:
-                            out = cast_array_to_feature(out, type, allow_primitive_to_str=True, allow_decimal_to_str=True)
+                            out = cast_array_to_feature(
+                                out, type, allow_primitive_to_str=True, allow_decimal_to_str=True
+                            )
                         return out
                     else:
                         raise
--- a/src/datasets/arrow_writer.py
+++ b/src/datasets/arrow_writer.py
@@ -205,7 +205,9 @@ class TypedSequence:
                 # We use cast_array_to_feature to support casting to custom types like Audio and Image
                 # Also, when trying type "string", we don't want to convert integers or floats to "string".
                 # We only do it if trying_type is False - since this is what the user asks for.
-                out = cast_array_to_feature(out, type, allow_number_to_str=not self.trying_type)
+                out = cast_array_to_feature(
+                    out, type, allow_primitive_to_str=not self.trying_type, allow_decimal_to_str=not self.trying_type
+                )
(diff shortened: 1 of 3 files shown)


Modexus:
@mariosasko pytest seems to be missing on windows?

mariosasko:
CI is not behaving well today 🙂

src/datasets/table.py
```diff
@@ -1933,13 +1940,14 @@ def array_cast(array: pa.Array, pa_type: pa.DataType, allow_number_to_str=True):
...
-            and (pa.types.is_floating(array.type) or pa.types.is_integer(array.type))
+        if pa.types.is_string(pa_type) and (
+            (not allow_primitive_to_str and pa.types.is_primitive(array.type))
+            or (not allow_decimal_to_str and pa.types.is_decimal(array.type))
         ):
             raise TypeError(
-                f"Couldn't cast array of type {array.type} to {pa_type} since allow_number_to_str is set to {allow_number_to_str}"
+                f"Couldn't cast array of type {array.type} to {pa_type} "
+                f"since allow_primitive_to_str is set to {allow_primitive_to_str} "
+                f"and allow_decimal_to_str is set to {allow_decimal_to_str}."
```
mariosasko, line 1950: Maybe let's be more precise here depending on the type - if it's primitive, then we should mention `allow_primitive_to_str `, and if it's decimal  `allow_decimal_to_str `

mariosasko (commented):
Thanks! A few comments:

src/datasets/arrow_writer.py
```diff
@@ -241,7 +243,7 @@ def __arrow_array__(self, type: Optional[pa.DataType] = None):
                             cast_to_python_objects(data, only_1d_for_numpy=True, optimize_list_casting=False)
                         )
                         if type is not None:
-                            out = cast_array_to_feature(out, type, allow_number_to_str=True)
+                            out = cast_array_to_feature(out, type, allow_primitive_to_str=True)
```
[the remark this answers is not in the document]
Modexus: Didn't think about the other calls..., I added this now.

src/datasets/table.py
Modexus: Split this into two separate errors.

mariosasko (approved):
LGTM! Also, thanks for adding tests!

mariosasko:
I could find an instance of the `allow_number_to_str` parameter (or `array_cast`/`cast_array_to_feature` more generally) being used in the wild. So, I think simply removing `allow_number_to_str` instead of deprecating it should be fine, considering `array_cast`/`cast_array_to_feature` are somewhat hidden. Do you agree @lhoestq?

lhoestq:
Yup we can remove without any deprecation cycle

lhoestq (approved):
Nice fix !

PR #6815, opened 2024-04-16
Remove `os.path.relpath` in `resolve_patterns`

mariosasko:
... to save a few seconds when resolving repos with many data files.


lhoestq (approved):
good catch !

PR #6816, opened 2024-04-16
Improve typing of Dataset.search, matching definition

Dref360:
Previously, the output of `score, indices = Dataset.search(...)` would be numpy arrays.

The definition in `SearchResult` is a `List[int]` so this PR now matched the expected type.

The previous behavior is a bit annoying as `Dataset.__getitem__` doesn't support `numpy.int64` which forced me to convert `indices` to int eg:

```python
score, indices = ds.search(...)
item = ds[int(indices[0])]
```

### DIFF
--- a/src/datasets/search.py
+++ b/src/datasets/search.py
@@ -364,7 +364,7 @@ class FaissIndex(BaseIndex):
         if not queries.flags.c_contiguous:
             queries = np.asarray(queries, order="C")
         scores, indices = self.faiss_index.search(queries, k, **kwargs)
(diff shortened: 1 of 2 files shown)


mariosasko:
Hi! This is a breaking change. A better solution is to check for "indexable" types in `__getitem__` to support keys such as `np.int64`:
```python
import operator

def _query_table_with_indices_mapping(...):  # or _query_table
    ...
    try:
        operator.index(key)
    except TypeError:
        pass
 
    _raise_bad_key_type(key)
```

PR #6822, opened 2024-04-18
Fix parquet export infos

lhoestq:
Don't use the parquet export infos when USE_PARQUET_EXPORT is False.

Otherwise the `datasets-server` might reuse erroneous data when re-running a job

this follows https://github.com/huggingface/datasets/pull/6714


albertvillanova (approved):
Thanks for the quick fix. I was not able to reproduce the issue locally...

PR #6828, opened 2024-04-23
Support PathLike input in save_to_disk / load_from_disk

lhoestq:


src/datasets/arrow_dataset.py
```diff
@@ -1449,7 +1450,7 @@ def save_to_disk(
         If you want to store paths or urls, please use the Value("string") type.
 
         Args:
-            dataset_path (`str`):
+            dataset_path (`PathLike`):
```
albertvillanova, line 1453: Great, so now this is aligned with the type hint: `dataset_path: PathLike`

albertvillanova (approved):
Thanks you.

Do you think it would be worth adding a test?

src/datasets/arrow_dataset.py
lhoestq: yes we should have one imo before merging

PR #6784, opened 2024-04-05
Extract data on the fly in packaged builders

mariosasko:
Instead of waiting for data files to be extracted in the packaged builders, we can prepend the compression prefix and extract them as they are being read (using `fsspec`). This saves disk space (deleting extracted archives is not set by default) and slightly speeds up dataset generation (less disk reads)

PR #6862, opened 2024-05-03
Fix load_dataset for data_files with protocols other than HF

matstrand:
Fixes huggingface/datasets/issues/6598

I've added a new test case and a solution. Before applying the solution the test case was failing with the same error described in the linked issue.

MRE:
```
pip install "datasets[s3]"
python -c "from datasets import load_dataset; load_dataset('csv', data_files={'train': 's3://noaa-gsod-pds/2024/A5125600451.csv'})"
```


src/datasets/utils/file_utils.py
```diff
@@ -568,6 +568,10 @@ def get_from_cache(
         if scheme == "ftp":
             connected = ftp_head(url)
         elif scheme not in ("http", "https"):
+            if scheme in ("s3", "s3a") and storage_options is not None and "hf" in storage_options:
+                # Issue 6598: **storage_options is passed to botocore.session.Session() 
+                # and must not contain keys that become invalid kwargs.
+                del storage_options["hf"]
```
lhoestq, line 574: it would be great to have a more general solution for any filesystem than hardcoding something for S3

maybe in `_prepare_single_hop_path_and_storage_options()` in file_utils.py ?

stoicio: Thanks ! Can you please explain a bit more what a better  general solution could be?  Is this an option ? Move [this logic](https://github.com/stoicio/datasets/blob/a188022dc43a76a119d90c03832d51d6e4a94d91/src/datasets/download/download_config.py#L96-L97)  to set the hf token to  this protocol check [Link](https://github.com/stoicio/datasets/blob/5e72fb13b4824dcb27aedb807e4e28c420dec244/src/datasets/utils/file_utils.py#L1188)

matstrand: Thanks both, seems the root issue is setting`DownloadConfig.storage_options['hf']` in all situations. We may be able to remove that as `test_load.py` passes without it. It was added at https://github.com/huggingface/datasets/pull/6028 but seems possibly obsoleted by https://github.com/huggingface/datasets/pull/6784.

I've pushed a change. @lhoestq can you please trigger the CI so we can see if this introduces any regressions?

lhoestq: sure, I triggered the ci

stoicio (commented):
Looks good to me, Thanks for fixing !

tests/test_load.py
```diff
@@ -1648,6 +1651,39 @@ def test_load_from_disk_with_default_in_memory(
         _ = load_from_disk(dataset_path)
 
 
+@contextmanager
+def moto_server():
+    with patch.dict(os.environ, {"AWS_ENDPOINT_URL": "http://localhost:5000"}):
```
stoicio, line 1656: You might have to add some dummy creds here to get past the NoCredentialsError in the test http://docs.getmoto.org/en/latest/docs/getting_started.html#recommended-usage

matstrand: thanks, yes this is correct. I setup a docker container to reproduce and fix. I also decided to minimize the change to hopefully avoid unintended consequences. 🤞 

Pushed, please have a look when you can.

src/datasets/download/download_config.py
```diff
@@ -93,7 +93,7 @@ def __post_init__(self, use_auth_token):
                 FutureWarning,
             )
             self.token = use_auth_token
-        if "hf" not in self.storage_options:
+        if self.token is not None and "hf" not in self.storage_options:
```
matstrand, line 96: I was worried that removing this altogether might break functionality for someone. This still might break CI but let's see before spending more time on it.

lhoestq: you might have to revert this, since we need the endpoint to be in the storage options even if token is None

matstrand: I'm trying to account for that [here](https://github.com/huggingface/datasets/pull/6862/files#diff-b25bef23b0bcf8e3bb59f625d727e585236772d449eae9dee5d32def98ffd321R201). This way `_prepare_path_and_storage_options` ensures that the endpoint is in the storage options without needing to populate it for all protocols in `__post_init__`.

albertvillanova: I think we can completely remove the default "hf" storage option from `DownloadConfig`, now that it is properly handled by the call to `_prepare_path_and_storage_options` made in `cached_path`.

albertvillanova (approved):
Thank you, @matstrand!! :hugs: 

I have merged the main branch to fix the CI, and added a new test for `cached_path` with different protocols, so we are sure that the passed `storage_options` (to `fsspec_head` and `fsspec_get`) is only the one that matches the protocol in the URL.

I would need this PR merged for a subsequent PR I am working on.

src/datasets/download/download_config.py
albertvillanova: Do you agree, @lhoestq?

lhoestq: yes sounds good :)

PR #6874, opened 2024-05-06
Use pandas ujson in JSON loader to improve performance

### LINKED ISSUE #6867: Improve performance of JSON loader
As reported by @natolambert, loading regular JSON files with `datasets` shows poor performance.

The cause is that we use the `json` Python standard library instead of other faster libraries. See my old comment: https://github.com/huggingface/datasets/pull/2638#pullrequestreview-706983714
> There are benchmarks that compare different JSON packages, with the Standard Library one among the worst performant: 
> - https://github.com/ultrajson/ultrajson#benchmarks
> - https://github.com/ijl/orjson#performance

I remember having a discussion about this and it was decided that it was better not to include an additional dependency on a 3rd-party library.

However:
- We already depend on `pandas` and `pandas` depends on `ujson`: so we have an indirect dependency on `ujson`
- Even if the above were not the case, we always could include `ujson` as an optional extra dependency, and check at runtime if it is installed to decide which library to use, either json or ujson

albertvillanova:
Use pandas ujson in JSON loader to improve performance.

Note that `datasets` has `pandas` as required dependency. And `pandas` includes `ujson` in `pd.io.json.ujson_loads`.

Fix #6867.

CC: @natolambert

### DIFF
--- a/src/datasets/packaged_modules/json/json.py
+++ b/src/datasets/packaged_modules/json/json.py
@@ -15,6 +15,14 @@ from datasets.utils.file_utils import readline
 logger = datasets.utils.logging.get_logger(__name__)
 
 
+def ujson_loads(*args, **kwargs):
+    try:
+        return pd.io.json.ujson_loads(*args, **kwargs)
+    except AttributeError:
+        # Before pandas-2.1.0, ujson_loads was renamed to loads: import ujson_loads as loads
+        return pd.io.json.loads(*args, **kwargs)
+
+
 @dataclass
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
Great ! do you know what the speed-up is (roughly) ?

albertvillanova:
Thanks for your review, @lhoestq.

The performance gain depends on many factors, such as underlying data structures, file size...

In my benchmark, the performance gain was around 8.1%.

PR #6904, opened 2024-05-17
Fix decoding multi part extension

lhoestq:
e.g. a field named `url.txt` should be a treated as text

I also included a small fix to support .npz correctly


lhoestq:
takign the liberty to merge this for the viewer and a new dataset being released

PR #6915, opened 2024-05-22
Validate config name and data_files in packaged modules

albertvillanova:
Validate the config attributes `name` and `data_files` in packaged modules by making the derived classes call their parent `__post_init__` method.

Note that their parent `BuilderConfig` validates its attributes `name` and `data_files` in its `__post_init__` method: https://github.com/huggingface/datasets/blob/60d21efbc01e15d0b596ac1072750cbecd91548a/src/datasets/builder.py#L128-L137

This PR makes the derived config classes call their parent `__post_init__` method to validate their `name` and `data_files` attributes.

### DIFF
--- a/src/datasets/packaged_modules/arrow/arrow.py
+++ b/src/datasets/packaged_modules/arrow/arrow.py
@@ -17,6 +17,9 @@ class ArrowConfig(datasets.BuilderConfig):
 
     features: Optional[datasets.Features] = None
 
+    def __post_init__(self):
+        super().__post_init__()
+
 
 class Arrow(datasets.ArrowBasedBuilder):
     BUILDER_CONFIG_CLASS = ArrowConfig
--- a/src/datasets/packaged_modules/audiofolder/audiofolder.py
+++ b/src/datasets/packaged_modules/audiofolder/audiofolder.py
@@ -15,6 +15,9 @@ class AudioFolderConfig(folder_based_builder.FolderBasedBuilderConfig):
     drop_labels: bool = None
     drop_metadata: bool = None
 
+    def __post_init__(self):
+        super().__post_init__()
+
 
 class AudioFolder(folder_based_builder.FolderBasedBuilder):
     BASE_FEATURE = datasets.Audio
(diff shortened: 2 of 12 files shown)


lhoestq (approved):

tests/test_load.py
```diff
@@ -1749,7 +1749,7 @@ def test_resolve_trust_remote_code_future(trust_remote_code, expected):
 def test_reload_old_cache_from_2_15(tmp_path: Path):
     cache_dir = tmp_path / "test_reload_old_cache_from_2_15"
     builder_cache_dir = (
-        cache_dir / "polinaeterna___audiofolder_two_configs_in_metadata/v2-374bfde4f55442bc/0.0.0/7896925d64deea5d"
+        cache_dir / "polinaeterna___audiofolder_two_configs_in_metadata/v2-374bfde4f55442bc/0.0.0/cf191ad706de653e"
```
albertvillanova, line 1752: @lhoestq note that the hash now will be different from before this PR. I hope this does not have a negative impact...

lhoestq: ah it's not supposed to change if we want to support reloading old cache directories

lhoestq:
I pushed a change that fixes 2.15 cache reloading (I fixed the packaged module hash), feel free to merge if this change is fine for you

tests/test_load.py
albertvillanova: @lhoestq then we cannot change the builder code anymore?

albertvillanova (commented):
> @lhoestq then we cannot change the packaged builders code anymore?

Oh, I saw your fix! 
That will make the code more stable against future changes in the package builder files. Thanks.

albertvillanova:
Something weird happened in GitHub: I just merged this PR to main, See:
- https://github.com/huggingface/datasets/commit/5bbbf1b19766e31a6905f3e82bf3aa3f9f84a982

However this PR still appears as Open...

albertvillanova:
Indeed the merge commit is: https://github.com/huggingface/datasets/commit/5bbbf1b19766e31a6905f3e82bf3aa3f9f84a982

The following commit is just empty: https://github.com/huggingface/datasets/commit/09ebf5190afbd017f3ca24ef444be2d933411eed

PR #6920, opened 2024-05-26
[WebDataset] Add `.pth` support for torch tensors

lhoestq:
In this PR I add support for `.pth` but with `weights_only=True` to disallow the use of pickle


albertvillanova (approved):
Thanks.

PR #6934, opened 2024-05-30
Revert ci user

lhoestq:


albertvillanova (commented):
Could you please explain why you updated and then reverted the CI user? Thanks.

lhoestq:
There was an incident in hub-ci that invalidated our token. It's been fixed so I reverted this change

PR #6952, opened 2024-06-04
Move info_utils errors to exceptions module

albertvillanova:
Move `info_utils` errors to `exceptions` module.

Additionally rename some of them, deprecate the former ones, and make the deprecation backward compatible (by making the new errors inherit from the former ones).

### DIFF
--- a/src/datasets/exceptions.py
+++ b/src/datasets/exceptions.py
@@ -101,6 +101,9 @@ class ChecksumVerificationException(Exception):
 class ChecksumVerificationError(DatasetsError, ChecksumVerificationException):
     """Error raised during checksums verifications of downloaded files."""
 
+    def __init__(self, *args, **kwargs):
+        DatasetsError.__init__(self, *args, **kwargs)
+
 
 @deprecated("Use 'UnexpectedDownloadedFileError' instead.")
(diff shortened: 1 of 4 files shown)


src/datasets/exceptions.py
```diff
@@ -83,3 +84,107 @@ def from_cast_error(
...
+
+    <Deprecated version="2.20.0">
+
+    Use `SplitsVerificationError` instead.
+
+    </Deprecated>
+    """
+
+
+class SplitsVerificationError(DatasetsError, SplitsVerificationException):
```
lhoestq, line 153: is it also marked as deprecated since it inherits from SplitsVerificationException ?

albertvillanova: I have renamed `SplitsVerificationException` to `SplitsVerificationError` and deprecated the former. The new `SplitsVerificationError` is not deprecated. 

It inherits from `SplitsVerificationException` for backward compatibility: if a user is catching `SplitsVerificationException`, now it will also catch `SplitsVerificationError`.

albertvillanova: @lhoestq I fixed it.

lhoestq (approved):
nice !

PR #6954, opened 2024-06-04
Remove default `trust_remote_code=True`

lhoestq:
TODO:
- [x] fix tests


julien-c:
yay! 🎉

henilp105:
@lhoestq Thanks for the PR, Is there a way to detect if `trust_remote_code=True` will be required for loading the dataset, without loading it? It would be great if you could please point me to the relevant documentation.

lhoestq:
You can check the presence of a python loading script in the repository.

If there is a .py file named after the repository name, then it requires trust_remote_code.

henilp105:
Thanks @lhoestq for the reference.

PR #6966, opened 2024-06-12
Remove underlines between badges

novialriptide:
## Before:
image

## After:
image


lhoestq (approved):
oh I've never noticed them ^^'

PR #6976, opened 2024-06-17
Ensure compatibility with numpy 2.0.0

KennethEnevoldsen:
Following the conversion guide, copy=False is no longer required and will result in an error: https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword.

The following fix should resolve the issue.

### DIFF
--- a/src/datasets/formatting/formatting.py
+++ b/src/datasets/formatting/formatting.py
@@ -187,14 +187,20 @@ class NumpyArrowExtractor(BaseArrowExtractor[dict, np.ndarray, dict]):
             else:
                 zero_copy_only = _is_zero_copy_only(pa_array.type) and not _is_array_with_nulls(pa_array)
                 array: List = pa_array.to_numpy(zero_copy_only=zero_copy_only).tolist()
+
         if len(array) > 0:
             if any(
                 (isinstance(x, np.ndarray) and (x.dtype == object or x.shape != array[0].shape))
                 or (isinstance(x, float) and np.isnan(x))
                 for x in array
             ):
+                if np.lib.NumpyVersion(np.__version__) >= '2.0.0b1':
+                    return np.array(array, dtype=object)
                 return np.array(array, copy=False, dtype=object)
-        return np.array(array, copy=False)
+        if np.lib.NumpyVersion(np.__version__) >= '2.0.0b1':
+            return np.array(array)
+        else:
+            return np.array(array, copy=False)
 
 
 class PandasArrowExtractor(BaseArrowExtractor[pd.DataFrame, pd.Series, pd.DataFrame]):


src/datasets/formatting/formatting.py
```diff
@@ -187,14 +187,20 @@ def _arrow_array_to_numpy(self, pa_array: pa.Array) -> np.ndarray:
...
                 (isinstance(x, np.ndarray) and (x.dtype == object or x.shape != array[0].shape))
                 or (isinstance(x, float) and np.isnan(x))
                 for x in array
             ):
+                if np.lib.NumpyVersion(np.__version__) >= '2.0.0b1':
+                    return np.array(array, dtype=object)
                 return np.array(array, copy=False, dtype=object)
-        return np.array(array, copy=False)
+        if np.lib.NumpyVersion(np.__version__) >= '2.0.0b1':
+            return np.array(array)
```
lhoestq, line 201: I think it should be `asarray` to avoid copies when possible no ?
```suggestion
            return np.asarray(array)
```

KennethEnevoldsen: You are completely right

KennethEnevoldsen: I believe we can actually remove the if else statement and just use asarray() with older versions of numpy as well. That is at least my reading of the [docs](https://numpy.org/doc/1.26/reference/generated/numpy.asarray.html)

lhoestq: they state that asarray used to have more overhead than copy=False. So maybe keep copy=False for numpy 1.x and use asarray for 2.x

src/datasets/formatting/formatting.py
```diff
@@ -187,14 +187,20 @@ def _arrow_array_to_numpy(self, pa_array: pa.Array) -> np.ndarray:
...
                 array: List = pa_array.to_numpy(zero_copy_only=zero_copy_only).tolist()
+
         if len(array) > 0:
             if any(
                 (isinstance(x, np.ndarray) and (x.dtype == object or x.shape != array[0].shape))
                 or (isinstance(x, float) and np.isnan(x))
                 for x in array
             ):
+                if np.lib.NumpyVersion(np.__version__) >= '2.0.0b1':
+                    return np.array(array, dtype=object)
```
lhoestq, line 198: last nit
```suggestion
                    return np.asarray(array, dtype=object)
```

src/datasets/formatting/formatting.py
```diff
@@ -187,14 +187,20 @@ def _arrow_array_to_numpy(self, pa_array: pa.Array) -> np.ndarray:
...
+
         if len(array) > 0:
             if any(
                 (isinstance(x, np.ndarray) and (x.dtype == object or x.shape != array[0].shape))
                 or (isinstance(x, float) and np.isnan(x))
                 for x in array
             ):
+                if np.lib.NumpyVersion(np.__version__) >= '2.0.0b1':
+                    return np.asarray(array, dtype=object)
                 return np.array(array, copy=False, dtype=object)
```
NeilGirdhar, line 199: I think you can make this uncoditional, and just use `asarray(array,...)` for both cases.  `asarray` has the right default no-copy behavior.

src/datasets/formatting/formatting.py
```diff
@@ -187,14 +187,20 @@ def _arrow_array_to_numpy(self, pa_array: pa.Array) -> np.ndarray:
...
-        return np.array(array, copy=False)
+        if np.lib.NumpyVersion(np.__version__) >= '2.0.0b1':
+            return np.asarray(array)
+        else:
+            return np.array(array, copy=False)
```
NeilGirdhar, line 203: same here.

lhoestq (approved):

PR #6983, opened 2024-06-19
Remove metrics

albertvillanova:
Remove all metrics, as part of the 3.0 release.

Note they are deprecated since 2.5.0 version.


lhoestq (approved):
Awesome ! You can also remove all the dependencies from https://github.com/huggingface/datasets/blob/main/additional-tests-requirements.txt (except torchdata)

Issue #6980, opened 2024-06-18
Support NumPy 2.0


NeilGirdhar:
### Feature request

Support NumPy 2.0.

### Motivation

NumPy introduces the Array API, which bridges the gap between machine learning libraries.  Many clients of HuggingFace are eager to start using the Array API.

Besides that, NumPy 2 provides a cleaner interface than NumPy 1.

### Tasks

NumPy 2.0 was released for testing so that libraries could ensure compatibility [since mid-March](https://github.com/numpy/numpy/issues/24300#issuecomment-1986815755).  What needs to be done for HuggingFace to support Numpy 2?

- [x] Fix use of `array`: https://github.com/huggingface/datasets/pull/6976
- [ ] Remove [NumPy version limit](https://github.com/huggingface/datasets/pull/6975): https://github.com/huggingface/datasets/pull/6991

PR #6991, opened 2024-06-22
Unblock NumPy 2.0

NeilGirdhar:
Fixes https://github.com/huggingface/datasets/issues/6980

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,3 +96,30 @@ jobs:
       - name: Test with pytest
         run: |
           python -m pytest -rfExX -m ${{ matrix.test }} -n 2 --dist loadfile -sv ./tests/
+
+  test_py310_numpy2:
+    needs: check_code_quality
+    strategy:
+      matrix:
+        test: ['unit']
+        os: [ubuntu-latest, windows-latest]
+        deps_versions: [deps-latest]
+    continue-on-error: false
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+      - name: Upgrade pip
+        run: python -m pip install --upgrade pip
+      - name: Install uv
+        run: pip install --upgrade uv
+      - name: Install dependencies
+        run: uv pip install --system "datasets[tests_numpy2] @ ."
+      - name: Test with pytest
+        run: |
+          python -m pytest -rfExX -m ${{ matrix.test }} -n 2 --dist loadfile -sv ./tests/
--- a/setup.py
+++ b/setup.py
@@ -176,7 +176,7 @@ TESTS_REQUIRE = [
     "sqlalchemy",
     "s3fs>=2021.11.1",  # aligned with fsspec[http]>=2021.11.1; test only on python 3.7 for now
     "protobuf<4.0.0",  # 4.0.0 breaks compatibility with tensorflow<2.12
-    "tensorflow>=2.6.0",  # Issue installing 2.16.0; we rely on other dependencies pinning numpy < 2
+    "tensorflow>=2.6.0",  # Issue installing 2.16.0 with Python 3.8; we rely on other dependencies pinning numpy < 2
     "tiktoken",
     "torch>=2.0.0",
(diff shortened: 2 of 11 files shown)


NeilGirdhar:
@albertvillanova Any chance we could get this in before the next release?  Everything depending on HuggingFace has their NumPy upgrade blocked.

albertvillanova (commented):
The CI integration tests for Python 3.10 are failing.

albertvillanova (changes requested):
Note that the CI tests for Python 3.8 are OK because they use numpy 1.24.4: numpy 2.0.0 requires Python >= 3.10

albertvillanova:
The incompatible libraries are:
- faiss-cpu 1.8.0.post1 requires numpy<2.0,>=1.0, but you have numpy 2.0.0 which is incompatible.
- tensorflow 2.16.2 requires numpy<2.0.0,>=1.23.5; python_version <= "3.11", but you have numpy 2.0.0 which is incompatible.
- transformers 4.42.3 requires numpy<2.0,>=1.17, but you have numpy 2.0.0 which is incompatible.

NeilGirdhar:
Why is it installing numpy 2 if the dependencies don't support it?

NeilGirdhar:
For me, I'm getting:
```
❯ uv pip install --system "datasets[tests] @ ."
Found existing alias for "uv pip install". You should use: "pipi"
Resolved 119 packages in 934ms
   Built datasets @ file:///Users/neil/src/datasets
Prepared 1 package in 1.28s
Uninstalled 1 package in 10ms
Installed 2 packages in 17ms
 - datasets==2.20.1.dev0 (from file:///Users/neil/src/datasets)
 + datasets==2.20.1.dev0 (from file:///Users/neil/src/datasets)
 + numpy==1.26.4
```

albertvillanova:
Which version on Python do you have?

NeilGirdhar:
3.12.4  I'll try on 3.10 now.

albertvillanova:
Please, note that I obtained  the previous incompatible libraries in my local environment, by forcing the update of numpy.

albertvillanova:
In the Python 3.10 CI, the situation is different:
- for example, they install an older version of tensorflow (2.14.0), where probably the constraint on numpy was not yet implemented. See the details: https://github.com/huggingface/datasets/actions/runs/9879100332/job/27306903343?pr=6991
```
> uv pip install --system "datasets[tests] @ ."
...
 + faiss-cpu==1.8.0
...
  + numpy==2.0.0
...
 + tensorflow==2.14.0
```

albertvillanova:
The main point is that we cannot support numpy 2.0 until tensorflow and faiss do.

NeilGirdhar:
> Alternatively, we should ignore/select tests depending on the installed versions.

That works.

Alternatively, you could depend on tensorflow 2.16.2 (etc.) for the tests?

albertvillanova:
Yes, I was thinking of a workaround solution.

The issue I see is that our CI will not test numpy 2.0 indeed.

NeilGirdhar:
> The issue I see is that our CI will not test numpy 2.0 indeed.

Right, that's the advantage of the test skipping you wanted, I see your point.

albertvillanova:
Without testing Numpy 2.0, we do not know if there are some other parts in the code broken.

albertvillanova:
> I guess the solution is to write tests that don't depend on tensorflow, etc., but still use numpy.
That is my point. What we cannot do is just blindly support Numpy 2.0 without knowing its consequences. We need to test it:
- to know if our core code works with it
- to know what optional libraries are incompatible

For example, while testing locally, I have discovered that librosa is also incompatible with numpy-2.0, due to its dependency on soxr...

albertvillanova:
While testing locally, I have also discovered that pytorch does not support Numpy 2.0 on Windows platforms.

albertvillanova:
I am adding Numpy 2.0 tests to your PR if you don't mind, before merging this PR.

NeilGirdhar:
Awesome, thank you!  Please let me know if I need to do anything.

albertvillanova:
Now we test numpy 2.0 in the `test_py310_numpy2` CI tests: https://github.com/huggingface/datasets/actions/runs/9907254874/job/27370545495?pr=6991
```
 + numpy==2.0.0
```

albertvillanova (approved):
Thank you.

PR #6993, opened 2024-06-24
less script docs

lhoestq:
+ mark as legacy in some parts of the docs since we'll not build new features for script datasets

### DIFF
--- a/docs/source/audio_dataset.mdx
+++ b/docs/source/audio_dataset.mdx
@@ -14,8 +14,6 @@ There are several methods for creating and sharing an audio dataset:
 
 * Create an audio dataset repository with the `AudioFolder` builder. This is a no-code solution for quickly creating an audio dataset with several thousand audio files.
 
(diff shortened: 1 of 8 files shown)


albertvillanova (approved):
Thanks.

mayorblock:
Hi @lhoestq,

I was confused by `legacy` prefix added to the [image data loading](https://huggingface.co/docs/datasets/main/en/image_dataset#legacy-loading-script) script section. I have a custom image dataset and have looked through the documentation to find something similar but can't find a good alternative What is now the recommend way to create a custom image dataset then? I want the HF format but will not host it on the hub.

Apologies in advance if this is the wrong place to ask such questions...

lhoestq:
We stopped making new features for datasets with scripts for obvious security reasons, that's why they are marked as "legacy". What is blocking you from hosting on HF ?

mayorblock:
Hi, thanks for the prompt answer :) I am working on proprietary datasets for the company where I am employed. We want to keep the data in-house but would like to investigate the use of the HF ecosystem.

PR #6999, opened 2024-06-25
Remove tasks

albertvillanova:
Remove tasks, as part of the 3.0 release.


src/datasets/info.py
```diff
@@ -146,7 +143,6 @@ class DatasetInfo:
     features: Optional[Features] = None
     post_processed: Optional[PostProcessedInfo] = None
     supervised_keys: Optional[SupervisedKeysData] = None
-    task_templates: Optional[List[TaskTemplate]] = None
```
lhoestq, line 149: Maybe we can leave this (or ignore it in a post_init) since otherwise it will break ILSVRC/imagenet-1k and ylecun/mnist and many other datasets.

We could also keep the task classes to be imported and instantiated without errors but still remove all their methods like align_feature since they are unused

albertvillanova: So we keep these deprecated classes and parameters for datasets 3.0...

lhoestq: Alternatively we could just hardcode a patch in dataset-viewer to keep supporting imagenet and mnist until they are converted to no-code datasets ?

lhoestq: Actually breaking this will be in the right direction of making people stop using code datasets. I'm just concerned that the mnist repo will stop working but if ylecun doesn't merge the PR to convert the dataset to parquet we can still hardcode something for the viewer

(and imagenet we can merge by ourselves)

albertvillanova: I understand your concerns, but at the same time I would push to not keep the deprecated tasks (since version 2.13.0, more than a year ago) in the new major version.

So I would propose, before making the next release:
- Identify all the datasets using tasks
- Open PRs to convert them to Parquet
- Wait 1/2 weeks for the owners to merge before forcing the merge ourselves for "maintenance" reasons
- Only then, make the release

I can work on that with a script.

lhoestq: Sounds good, though no need to convert all the datasets - just the ones that are important

Issue #7000, opened 2024-06-25
IterableDataset: Unsupported ScalarType BFloat16


stoical07:
### Describe the bug

`IterableDataset.from_generator` crashes when using BFloat16:

```
  File "/usr/local/lib/python3.11/site-packages/datasets/utils/_dill.py", line 169, in _save_torchTensor
    args = (obj.detach().cpu().numpy(),)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^

TypeError: Got unsupported ScalarType BFloat16
```

### Steps to reproduce the bug

```python
import torch
from datasets import IterableDataset


def demo(x):
    yield {"x": x}


x = torch.tensor([1.], dtype=torch.bfloat16)

dataset = IterableDataset.from_generator(
    demo,
    gen_kwargs=dict(x=x),
)

example = next(iter(dataset))
print(example)
```

### Expected behavior

Code sample should print:
```python
{'x': tensor([1.], dtype=torch.bfloat16)}
```

### Environment info

```
datasets==2.20.0
torch==2.2.2
```

stoical07:
@lhoestq Thank you for merging #6607, but unfortunately the issue persists for `IterableDataset` :pensive:

lhoestq:
Hi ! I opened https://github.com/huggingface/datasets/pull/7002 to fix this bug

stoical07:
Amazing, thank you so much @lhoestq! :pray:

PR #7002, opened 2024-06-25
Fix dump of bfloat16 torch tensor

lhoestq:
close https://github.com/huggingface/datasets/issues/7000

### DIFF
--- a/src/datasets/utils/_dill.py
+++ b/src/datasets/utils/_dill.py
@@ -162,11 +162,17 @@ def _save_torchTensor(pickler, obj):
     import torch  # type: ignore
 
     # `torch.from_numpy` is not picklable in `torch>=1.11.0`
-    def create_torchTensor(np_array):
-        return torch.from_numpy(np_array)
+    def create_torchTensor(np_array, dtype=None):
+        tensor = torch.from_numpy(np_array)
+        if dtype:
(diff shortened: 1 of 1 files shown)


src/datasets/utils/_dill.py
```diff
@@ -162,11 +162,17 @@ def _save_torchTensor(pickler, obj):
     import torch  # type: ignore
 
     # `torch.from_numpy` is not picklable in `torch>=1.11.0`
-    def create_torchTensor(np_array):
-        return torch.from_numpy(np_array)
+    def create_torchTensor(np_array, dtype=None):
+        tensor = torch.from_numpy(np_array)
+        if dtype:
+            tensor = tensor.type(torch.bfloat16)
+        return tensor
```
stoical07, line 169: For future compatibility? :smile:

```suggestion
    def create_torchTensor(np_array, dtype=None):
        tensor = torch.from_numpy(np_array)
        if dtype:
            tensor = tensor.type(dtype)
        return tensor
```

lhoestq: ah good catch, I went too fast x)

lhoestq: fixed at https://github.com/huggingface/datasets/pull/7003

Issue #6900, opened 2024-05-15
[WebDataset] KeyError with user-defined `Features` when a field is missing in an example


lhoestq:
reported at https://huggingface.co/datasets/ProGamerGov/synthetic-dataset-1m-dalle3-high-quality-captions/discussions/1

```
File "/src/services/worker/.venv/lib/python3.9/site-packages/datasets/packaged_modules/webdataset/webdataset.py", line 109, in _generate_examples
    example[field_name] = {"path": example["__key__"] + "." + field_name, "bytes": example[field_name]}
```

ProGamerGov:
@lhoestq How difficult of fix is this?

lhoestq:
It shouldn't be difficult, I think it's just a matter of adding the missing fields from `self.config.features` in `example` here:

https://github.com/huggingface/datasets/blob/768cb35ede5a6c35fa7545aa3671f3e321c96440/src/datasets/packaged_modules/webdataset/webdataset.py#L113-L116

lhoestq:
Yup this is the solution !

```python
            for field_name in all_field_names:
                if field_name not in example:
                    example[field_name] = None
```

ProGamerGov:
@lhoestq Awesome, thanks! I made a PR with the fixes

PR #7004, opened 2024-06-26
Fix WebDatasets KeyError for user-defined Features when a field is missing in an example

ProGamerGov:
Fixes: https://github.com/huggingface/datasets/issues/6900

Not sure if this needs any addition stuff before merging

### DIFF
--- a/src/datasets/packaged_modules/webdataset/webdataset.py
+++ b/src/datasets/packaged_modules/webdataset/webdataset.py
@@ -109,8 +109,12 @@ class WebDataset(datasets.GeneratorBasedBuilder):
         audio_field_names = [
             field_name for field_name, feature in self.info.features.items() if isinstance(feature, datasets.Audio)
         ]
+        all_field_names = list(self.config.features.keys())
         for tar_idx, (tar_path, tar_iterator) in enumerate(zip(tar_paths, tar_iterators)):
             for example_idx, example in enumerate(self._get_pipeline_from_tar(tar_path, tar_iterator)):
+                for field_name in all_field_names:
+                    if field_name not in example:
+                        example[field_name] = None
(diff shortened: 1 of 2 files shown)


src/datasets/packaged_modules/webdataset/webdataset.py
```diff
@@ -109,8 +109,12 @@ def _generate_examples(self, tar_paths, tar_iterators):
...
             field_name for field_name, feature in self.info.features.items() if isinstance(feature, datasets.Audio)
         ]
+        all_field_names = list(self.config.features.keys())
         for tar_idx, (tar_path, tar_iterator) in enumerate(zip(tar_paths, tar_iterators)):
             for example_idx, example in enumerate(self._get_pipeline_from_tar(tar_path, tar_iterator)):
+                for field_name in all_field_names:
+                    if field_name not in example:
+                        example[field_name] = None
                 for field_name in image_field_names + audio_field_names:
                     example[field_name] = {"path": example["__key__"] + "." + field_name, "bytes": example[field_name]}
```
lhoestq, line 119:
```suggestion
                    if example[field_name] is not None:
                        example[field_name] = {"path": example["__key__"] + "." + field_name, "bytes": example[field_name]}
```

lhoestq (commented):
Great ! I added a small suggestion to handle the case where the missing field is an image or an audio sample.

Could you also add a small test in `test_webdataset.py` ?
I believe you can reuse the same code as in `test_image_webdataset` and simply pass `features=` to `Webdataset()` and check if the examples have the extra features as None. Then we'll be good to merge !

lhoestq (approved):
Great ! I updated the test :) the rest of the CI failures are unrelated to this PR

ProGamerGov:
@lhoestq Thank you!

PR #7015, opened 2024-07-01
add split argument to Generator

piercus:
## Actual

When creating a multi-split dataset using generators like 

```python
datasets.DatasetDict({
  "val": datasets.Dataset.from_generator(
      generator=generator_val,
      features=features
  ),
  "test": datasets.Dataset.from_generator(
      generator=generator_test,
      features=features,
  )
})
```

It displays (for both test and val)

```
Generating train split
```

## Expected

I would like to be able to improve this behavior by doing

```python
datasets.DatasetDict({
  "val": datasets.Dataset.from_generator(
      generator=generator_val,
      features=features,
      split="val"
  ),
  "test": datasets.Dataset.from_generator(
      generator=generator_test,
      features=features,
      split="test"
  )
})
```

It would display

```
Generating val split
```

and 

```
Generating test split
```

## Proposal

Current PR is adding an explicit `split` argument and replace the implicit "train" split in the following classes/function : 
* Generator
* from_generator
* AbstractDatasetInputStream
* GeneratorDatasetInputStream

Please share your feedbacks

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -1068,6 +1068,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         keep_in_memory: bool = False,
         gen_kwargs: Optional[dict] = None,
         num_proc: Optional[int] = None,
+        split: Optional[NamedSplit] = None,
         **kwargs,
     ):
         """Create a Dataset from a generator.
@@ -1088,6 +1089,8 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                 Number of processes when downloading and generating the dataset locally.
                 This is helpful if the dataset is made of multiple files. Multiprocessing is disabled by default.
                 If `num_proc` is greater than one, then all list values in `gen_kwargs` must be the same length. These values will be split between calls to the generator. The number of shards will be the minimum of the shortest list in `gen_kwargs` and `num_proc`.
+            split (`str`, defaults to `"train"`):
+                Split name to be assigned to the dataset.
 
                 <Added version="2.7.0"/>
             **kwargs (additional keyword arguments):
@@ -1126,6 +1129,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
             keep_in_memory=keep_in_memory,
             gen_kwargs=gen_kwargs,
             num_proc=num_proc,
+            split=split,
             **kwargs,
         ).read()
 
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -3618,6 +3618,57 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
             yield rank, True, shard
 
     @transmit_format
+    @fingerprint_transform(inplace=False)
+    def batch(
+        self,
+        batch_size: int,
+        drop_last_batch: bool = False,
+        num_proc: Optional[int] = None,
+        new_fingerprint: Optional[str] = None,
+    ) -> "Dataset":
+        """
+        Group samples from the dataset into batches.
+
+        Args:
+            batch_size (`int`):
+                The number of samples in each batch.
+            drop_last_batch (`bool`, defaults to `False`):
+                Whether to drop the last incomplete batch.
+            num_proc (`int`, *optional*, defaults to `None`):
+                Max number of processes when generating cache. Already cached shards are loaded sequentially.
+            new_fingerprint (`str`, *optional*, defaults to `None`):
+                The new fingerprint of the dataset after transform.
+                If `None`, the new fingerprint is computed using a hash of the previous fingerprint, and the transform arguments.
+
+        Returns:
+            [`Dataset`]: A new Dataset where each item is a batch of multiple samples from the original dataset.
+
+        Example:
+
+        ```py
+        >>> from datasets import load_dataset
+        >>> ds = load_dataset("rotten_tomatoes", split="train")
+        >>> batched_ds = ds.batch(batch_size=4)
+        >>> batched_ds[0]
+        {'text': ['compassionately explores the seemingly irreconcilable situation...', ...],  # 4 items
+        'label': [1, 1, 1, 1]}
+        ```
+        """
+
+        def batch_fn(example):
+            return {k: [v] for k, v in example.items()}
+
+        return self.map(
+            batch_fn,
+            batched=True,
+            batch_size=batch_size,
+            drop_last_batch=drop_last_batch,
+            num_proc=num_proc,
+            new_fingerprint=new_fingerprint,
+            desc="Batching examples",
+        )
+
+    @transmit_format
     @fingerprint_transform(
         inplace=False, ignore_kwargs=["load_from_cache_file", "cache_file_name", "desc"], version="2.0.1"
     )
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -1068,7 +1068,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         keep_in_memory: bool = False,
         gen_kwargs: Optional[dict] = None,
         num_proc: Optional[int] = None,
-        split: Optional[NamedSplit] = None,
+        split: NamedSplit = Split.TRAIN,
         **kwargs,
     ):
         """Create a Dataset from a generator.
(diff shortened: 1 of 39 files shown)


src/datasets/io/generator.py
```diff
@@ -15,6 +15,7 @@ def __init__(
         streaming: bool = False,
         gen_kwargs: Optional[dict] = None,
         num_proc: Optional[int] = None,
+        split: Optional[NamedSplit] = None,
```
albertvillanova, line 18: Again, this should default to `Split.TRAIN` and therefore no `Optional`.

src/datasets/io/generator.py
```diff
@@ -23,20 +24,22 @@ def __init__(
             keep_in_memory=keep_in_memory,
             streaming=streaming,
             num_proc=num_proc,
+            split=split,
```
albertvillanova, line 27: I think this is not necessary. See my comment below.

src/datasets/io/generator.py
```diff
@@ -23,20 +24,22 @@ def __init__(
...
             gen_kwargs=gen_kwargs,
+            split=split,
             **kwargs,
         )
 
     def read(self):
         # Build iterable dataset
         if self.streaming:
-            dataset = self.builder.as_streaming_dataset(split="train")
+            dataset = self.builder.as_streaming_dataset(split=self.split)
```
albertvillanova, line 42: Instead of defining and using the `split` parameter as an attribute of the `GeneratorDatasetInputStream` class (and `AbstractDatasetInputStream` and children), we can do differently.

We can define the `split` parameter as an attribute of `GeneratorConfig` and here use it as: `self.builder.config.split`.

src/datasets/io/generator.py
```diff
@@ -52,6 +55,6 @@ def read(self):
                 num_proc=self.num_proc,
             )
             dataset = self.builder.as_dataset(
-                split="train", verification_mode=verification_mode, in_memory=self.keep_in_memory
+                split=self.split, verification_mode=verification_mode, in_memory=self.keep_in_memory
```
albertvillanova, line 58: Same as above, use `self.builder.config.split`.

src/datasets/io/sql.py
```diff
@@ -48,7 +48,7 @@ def read(self):
 
         # Build dataset for splits
         dataset = self.builder.as_dataset(
-            split="train", verification_mode=verification_mode, in_memory=self.keep_in_memory
+            split=self.split, verification_mode=verification_mode, in_memory=self.keep_in_memory
```
albertvillanova, line 51: I would not add the `split` parameter to the `SqlDatasetReader` because it is unrelated to the objective of this PR.

src/datasets/iterable_dataset.py
```diff
@@ -2062,6 +2062,7 @@ def from_generator(
         generator: Callable,
         features: Optional[Features] = None,
         gen_kwargs: Optional[dict] = None,
+        split: Optional[NamedSplit] = None,
```
albertvillanova, line 2065: Same comments as before.

src/datasets/iterable_dataset.py
```diff
@@ -2074,7 +2075,8 @@ def from_generator(
                 Keyword arguments to be passed to the `generator` callable.
                 You can define a sharded iterable dataset by passing the list of shards in `gen_kwargs`.
                 This can be used to improve shuffling and when iterating over the dataset with multiple workers.
-
+            split(`str`, default="train"):
+                Split name to be assigned to the dataset.
```
albertvillanova, line 2079: Same comments as before.

src/datasets/packaged_modules/generator/generator.py
```diff
@@ -22,11 +22,21 @@ def __post_init__(self):
...
+    def __init__(
+        self,
+        split: Optional[datasets.NamedSplit] = None,
+        **kwargs,
+    ):
+        self.split = split if split is not None else datasets.Split.TRAIN
+        return super().__init__(
+            **kwargs,
+        )
+
```
albertvillanova, line 34: Do not add the `split` parameter as an attribute of `Generator` class, but as an attribute of `GeneratorConfig` instead.

src/datasets/packaged_modules/generator/generator.py
```diff
@@ -22,11 +22,21 @@ def __post_init__(self):
...
     def _info(self):
         return datasets.DatasetInfo(features=self.config.features)
 
     def _split_generators(self, dl_manager):
-        return [datasets.SplitGenerator(name=datasets.Split.TRAIN, gen_kwargs=self.config.gen_kwargs)]
+        return [datasets.SplitGenerator(name=self.split, gen_kwargs=self.config.gen_kwargs)]
```
albertvillanova, line 39: Use `self.config.split` instead.

tests/test_arrow_dataset.py
```diff
@@ -3898,12 +3902,13 @@ def test_dataset_from_generator_keep_in_memory(keep_in_memory, data_generator, t
...
     default_expected_features = {"col_1": "string", "col_2": "int64", "col_3": "float64"}
+    split = NamedSplit("validation")
     expected_features = features.copy() if features else default_expected_features
     features = (
         Features({feature: Value(dtype) for feature, dtype in features.items()}) if features is not None else None
     )
-    dataset = Dataset.from_generator(data_generator, features=features, cache_dir=cache_dir)
-    _check_generator_dataset(dataset, expected_features)
+    dataset = Dataset.from_generator(data_generator, features=features, cache_dir=cache_dir, split=split)
+    _check_generator_dataset(dataset, expected_features, split)
```
albertvillanova, line 3912: I would add a specific `test_dataset_from_generator_split` with a parametrized `split` values, such as not passing any value, passing `NamedSplit("train")`, passing literal `"train"`, passing other NamedSplit, etc.

albertvillanova (changes requested):
Thanks for your proposed contribution @piercus.

This is a nice one!

Some comments below. Basically, I would propose to define the `split` parameter just as an attribute of `GeneratorConfig` instead of `Generator`, `GeneratorDatasetInputStream`, `AbstractDatasetInputStream` and `SqlDatasetReader`.

src/datasets/arrow_dataset.py
```diff
@@ -1088,6 +1089,8 @@ def from_generator(
                 Number of processes when downloading and generating the dataset locally.
                 This is helpful if the dataset is made of multiple files. Multiprocessing is disabled by default.
                 If `num_proc` is greater than one, then all list values in `gen_kwargs` must be the same length. These values will be split between calls to the generator. The number of shards will be the minimum of the shortest list in `gen_kwargs` and `num_proc`.
+            split (`str`, defaults to `"train"`):
+                Split name to be assigned to the dataset.
```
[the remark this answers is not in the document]
piercus: just used  `<Added version="2.21.0"/>`, please cross-check

src/datasets/iterable_dataset.py
piercus: I also added `<Added version="2.21.0"/>` please cross-check

tests/test_arrow_dataset.py
piercus: `test_dataset_from_generator_split` added, still i have impacted `_check_generator_dataset` to share the same generic check everywhere

piercus:
@albertvillanova thanks for the review, please take a look

src/datasets/iterable_dataset.py
```diff
@@ -2074,7 +2075,10 @@ def from_generator(
                 Keyword arguments to be passed to the `generator` callable.
                 You can define a sharded iterable dataset by passing the list of shards in `gen_kwargs`.
                 This can be used to improve shuffling and when iterating over the dataset with multiple workers.
+            split(`NamedSplit`, default=Split.TRAIN):
```
albertvillanova, line 2078: Nit (to be aligned with the other docstring; also you forgot to add a blank space after `split` and before `(`):
```suggestion
            split ([`NamedSplit`], defaults to `Split.TRAIN`)
```

tests/test_arrow_dataset.py
```diff
@@ -3903,7 +3904,20 @@ def test_dataset_from_generator_features(features, data_generator, tmp_path):
...
+@pytest.mark.parametrize(
+    "split",
+    [None, NamedSplit("train"), "train", NamedSplit("foo"), "foo"],
+)
+def test_dataset_from_generator_split(split, data_generator, tmp_path):
+    cache_dir = tmp_path / "cache"
+    default_expected_split = "train"
+    expected_features = {"col_1": "string", "col_2": "int64", "col_3": "float64"}
+    expected_split = split if split else default_expected_split
+    dataset = Dataset.from_generator(data_generator, cache_dir=cache_dir, split=split)
```
albertvillanova, line 3919: The CI is failing because in Python, "not passing any value" is different from "passing None value".

In this test, when the value of `split` is None, you should not pass it to `Dataset.from_generator`, so that you check its default value.

albertvillanova (changes requested):
Thanks! Good work.

Just a fix of the non-passing test and a nit.

src/datasets/iterable_dataset.py
```diff
@@ -2074,7 +2075,10 @@ def from_generator(
                 Keyword arguments to be passed to the `generator` callable.
                 You can define a sharded iterable dataset by passing the list of shards in `gen_kwargs`.
                 This can be used to improve shuffling and when iterating over the dataset with multiple workers.
+            split ([`NamedSplit`], defaults to `Split.TRAIN`)
```
albertvillanova, line 2078: My bad, I forgot the colon:
```suggestion
            split ([`NamedSplit`], defaults to `Split.TRAIN`):
```

piercus:
@albertvillanova please take a look

albertvillanova (approved):
Thanks for your contribution!

Note the CI action to generate the docs is failing due to an unrelated issue: https://github.com/huggingface/datasets/actions/runs/9887484572/job/27309892176?pr=7015

Therefore, if we do not want to break the generation of docs, this other PR should be merged before yours:
- #7036

PR #7025, opened 2024-07-04
feat: support non streamable arrow file binary format

kmehant:
Support Arrow files (`.arrow`) that are in non streamable binary file formats.

### DIFF
--- a/src/datasets/packaged_modules/arrow/arrow.py
+++ b/src/datasets/packaged_modules/arrow/arrow.py
@@ -43,7 +43,11 @@ class Arrow(datasets.ArrowBasedBuilder):
             if self.info.features is None:
                 for file in itertools.chain.from_iterable(files):
                     with open(file, "rb") as f:
-                        self.info.features = datasets.Features.from_arrow_schema(pa.ipc.open_stream(f).schema)
+                        try:
+                            reader = pa.ipc.open_stream(f)
+                        except pa.lib.ArrowInvalid:
+                            reader = pa.ipc.open_file(f)
+                    self.info.features = datasets.Features.from_arrow_schema(reader.schema)
                     break
             splits.append(datasets.SplitGenerator(name=split_name, gen_kwargs={"files": files}))
         return splits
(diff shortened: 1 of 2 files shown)


kmehant:
requesting review - @albertvillanova @lhoestq

src/datasets/packaged_modules/arrow/arrow.py
```diff
@@ -42,8 +42,12 @@ def _split_generators(self, dl_manager):
             # Infer features if they are stored in the arrow schema
             if self.info.features is None:
                 for file in itertools.chain.from_iterable(files):
-                    with open(file, "rb") as f:
-                        self.info.features = datasets.Features.from_arrow_schema(pa.ipc.open_stream(f).schema)
+                    data_memory_map = pa.memory_map(file)
```
lhoestq, line 45: Memory mapping is only available for local files, however in streaming mode `file` is a URL (and `open` is extended to work with URLs and returns a valid `f`).

Could you make it work using `f` rather than `data_memory_map ` ?

kmehant: Updated @lhoestq thanks

kmehant:
Thank you for the review.

> Before we merge could you also add a test in tests/packaged_modules/test_arrow.py ?

> I noticed it's pretty empty right now compared to test_json.py or test_csv.py though, maybe I can take care of it next week if needed

@lhoestq Would you like to take that up? since it needs adding some test data and I see no supportive examples for simple data formats - parquet pandas etc. Thanks

kmehant:
@lhoestq rebased the PR, if there is anything pending on this PR, It would be really helpful to have this feature into datasets, please let me know thanks.

kmehant:
@lhoestq 

Have added the unit test to generate tables for both the arrow formats - file and streaming.

Let me know if we have any docs changes as well. Thanks

Screenshot 2024-07-25 at 7 04 26 PM

kmehant:
@lhoestq any update on this thread? Thanks

prince14322:
Timely PR!
Can we please look into this?

albertvillanova (approved):
Thank you for the useful enhancement and the test!

PR #7039, opened 2024-07-11
Fix export to JSON when dataset larger than batch size

### LINKED ISSUE #7037: A bug of Dataset.to_json() function
### Describe the bug

When using the Dataset.to_json() function, an unexpected error occurs if the parameter is set to lines=False. The stored data should be in the form of a list, but it actually turns into multiple lists, which causes an error when reading the data again.
The reason is that to_json() writes to the file in several segments based on the batch size. This is not a problem when lines=True, but it is incorrect when lines=False, because writing in several times will produce multiple lists(when len(dataset) > batch_size).

try this code:
```python
import json

train_dataset = load_dataset("Anthropic/hh-rlhf", data_dir="harmless-base")["train"]
output_path = "./harmless-base_hftojs.json"
print(len(train_dataset))
train_dataset.to_json(output_path, lines=False, force_ascii=False, indent=2)

with open(output_path, encoding="utf-8") as f:
    data = json.loads(f.read())

```
it raise error: json.decoder.JSONDecodeError: Extra data: line 4003 column 1 (char 1373709)

Extra square brackets have appeared here:
image

### Expected behavior

The code runs normally.

### Environment info

datasets=2.20.0

albertvillanova:
Fix export to JSON when dataset larger than batch size.

Fix #7037.

### DIFF
--- a/tests/test_arrow_dataset.py
+++ b/tests/test_arrow_dataset.py
@@ -4041,10 +4041,11 @@ def test_dataset_from_sql_keep_in_memory(keep_in_memory, sqlite_path, tmp_path, 
     _check_sql_dataset(dataset, expected_features)
 
 
+@pytest.mark.parametrize("batch_size", [None, 5])
 @pytest.mark.parametrize("lines", [True, False])
-def test_dataset_to_json(lines, dataset, tmp_path):
+def test_dataset_to_json(lines, batch_size, dataset, tmp_path):
     file_path = tmp_path / "test_path.jsonl"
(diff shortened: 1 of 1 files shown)


albertvillanova:
The test before confirms the bug.

There are different possible solutions to this issue:
- the easiest would be to write multiple JSON files, one for each batch; this solution can be done in parallel if `num_proc` is passed
- alternatively, we could tweak the writing and remove the extra `[` and `]` characters; this solution will only be valid if `orient="records"`
- others?

I prefer the first solution.

PR #7042, opened 2024-07-12
Improved the tutorial by adding a link for loading datasets

AmboThom:
Improved the tutorial by letting readers know about loading datasets with common files and including a link. I left the local files section alone because the methods were already listed with code snippets.


lhoestq (approved):
thanks !

Issue #6279, opened 2023-10-05
Batched IterableDataset


lneukom:
### Feature request

Hi, 

could you add an implementation of a batched `IterableDataset`. It already support an option to do batch iteration via `.iter(batch_size=...)` but this cannot be used in combination with a torch `DataLoader` since it just returns an iterator.

### Motivation

The current implementation loads each element of a batch individually which can be very slow in cases of a big batch_size. I did some experiments [here](https://discuss.huggingface.co/t/slow-dataloader-with-big-batch-size/57224) and using a batched iteration would speed up data loading significantly.

### Your contribution

N/A

VascoSch92:
This is exactly what I was looking for. It would also be very useful for me :-)

zhh210:
This issue is really smashing the selling point of HF datasets... The only workaround I've found so far is to create a customized IterableDataloader.

For example I've a HF dataset `dt_train` with `len(dt_train) == 1M`. Using plain DataLoader is extremely slow:
```
%%time
dl_train = DataLoader(dt_train, batch_size=128, shuffle = True)
for batch in dl_train:
    pass
``` 

```
CPU times: user 24min 35s, sys: 704 ms, total: 24min 36s
Wall time: 24min 37s
```

Workaround by running a customized wrapper:
```
%%time
from torch.utils.data import DataLoader, IterableDataset

class Dataset2Iterable(IterableDataset):
    """
    Wrapper to use a HF dataset as pytorch IterableDataset to speed up data loading.
    """
    def __init__(self, dataset, batch_size=1, shuffle=True):
        super(Dataset2Iterable).__init__()
        self.dataset = dataset
        self.batch_size = batch_size
        self.shuffle = shuffle

    def __iter__(self):
        if self.shuffle: self.dataset.shuffle()
        return self.dataset.iter(batch_size=self.batch_size)

dl_test = DataLoader(Dataset2Iterable(dt_test, batch_size = 128), batch_size=1, num_workers=0)
for n in range(2):
    for batch in dl_test:
        pass
```
The speed still is slower than using tensorflow's loader but improved a lot than previous code:
```
CPU times: user 3min 4s, sys: 0 ns, total: 3min 4s
Wall time: 3min 4s
```
Note that the way I implemented `Dataset2Iterable` will only work with `num_workers=0`.

jaketae:
[...]

PR #7054, opened 2024-07-19
Add batching to `IterableDataset`

lappemic:
I've taken a try at implementing a batched `IterableDataset` as requested in issue #6279. This PR adds a new `BatchedExamplesIterable` class and a `.batch()` method to the `IterableDataset` class.

The main changes are:
1. A new `BatchedExamplesIterable` that groups examples into batches.
2. A `.batch()` method for `IterableDataset` to easily create batched versions.
3. Support for shuffling and sharding to work with PyTorch DataLoader and multiple workers.

I'm not sure if this is exactly what you had in mind and also have not fully tested it atm, so I'd really appreciate your feedback. Does this seem like it's heading in the right direction? I'm happy to make any changes or explore different approaches if needed.

Pinging @lhoestq


lhoestq:
Cool ! Thanks for diving into it :)

Your implementation is great for and indeed supports shuffling and batching, you just need to additionally account for state_dict (for dataset [checkpointing+resuming](https://huggingface.co/docs/datasets/main/en/use_with_pytorch#checkpoint-and-resume))

That being said, I believe the implementation can be made simpler by relying on `IterableDataset.map()` which already implements all this. Maybe something like

```python

def batch(self, batch_size: int, drop_last_batch: bool = False) -> "IterableDataset":
    def batch(unbatched: dict[str, list]) -> dict[str, list]:
        return {k: [v] for k, v in unbatched}

    return self.map(batch, batched=True, batch_size=batch_size, drop_last_batch=drop_last_batch)
```

And this way no need to reimplement everything !

(my only small concern is that it's not an Arrow-optimized function so it requires the examples to be manipulated as python objects even if the original data is in Arrow format (e.g. when streaming Parquet files) but it's not a big deal and we can see later if we need to optimize this)

lappemic:
Thanks a lot for the feedback @lhoestq! I definitely could have saved some time looking into it properly first. 😅 

Implemented the `.batch()` method, added a proper docsrtring for documentation, and added tests.

Let me know what you think and if this needs some update.

src/datasets/iterable_dataset.py
```diff
@@ -2885,6 +2885,28 @@ def _resolve_features(self):
...
 
+    def batch(self, batch_size: int, drop_last_batch: bool = False) -> "IterableDataset":
+        """
+        Group samples from the dataset into batches.
+
+        Args:
+            batch_size (int):
+                The number of samples in each batch.
+            drop_last_batch (bool, defaults to `False`):
+                Whether to drop the last incomplete batch.
```
lhoestq, line 2896:
```suggestion
            batch_size (`int`): The number of samples in each batch.
            drop_last_batch (`bool`, defaults to `False`): Whether to drop the last incomplete batch.
```

docs/source/about_mapstyle_vs_iterable.mdx
```diff
@@ -205,6 +205,43 @@ for epoch in range(n_epochs):
...
+When set to `True`, it will discard the last batch if it's smaller than the specified `batch_size`. 
+This can be useful in scenarios where your downstream processing requires all batches to be of the same size:
+
+```python
+batched_dataset = dataset.batch(batch_size=32, drop_last_batch=True)
+```
+
+Batching with `IterableDataset` combines the benefits of processing data in larger chunks with the memory efficiency of streaming data. 
+This makes it an excellent choice for working with large datasets that don't fit in memory, especially when used with deep learning frameworks that expect batched inputs.
+
```
lhoestq, line 244: feels a bit redundant

```suggestion
```

docs/source/about_mapstyle_vs_iterable.mdx
```diff
@@ -205,6 +205,43 @@ for epoch in range(n_epochs):
         pass
 ```
 
+## Batching data in IterableDataset
```
lhoestq, line 208: this can go in stream.mdx instead ?

Btw in the future it would be cool to have `Dataset.batch()` as well :)

lhoestq (commented):
Nice thanks ! I added a few suggestions before we merge.

lappemic:
Thanks for the feedbak @lhoestq!

Applied it and referenced the `batched=True` option in the `map` function and highlighted the difference. Hope i got this right.

lhoestq (approved):
lgtm !

PR #7057, opened 2024-07-22
Update load_hub.mdx

severo:


lhoestq (approved):
cool, let's aim to have "subsets" everywhere, and "configure the subsets" when it comes to the YAML (that uses the fields "configs" / "config_name" ) :)

PR #7064, opened 2024-07-23
Add `batch` method to `Dataset` class

### LINKED ISSUE #7063: Add `batch` method to `Dataset`
### Feature request

Add a `batch` method to the Dataset class, similar to the one recently implemented for `IterableDataset` in PR #7054.

### Motivation

A batched iteration speeds up data loading significantly (see e.g. #6279)

### Your contribution

I plan to open a PR to implement this.

lappemic:
This PR introduces a new `batch` method to the `Dataset` class, aligning its functionality with the `IterableDataset.batch()` method (implemented in #7054). The implementation uses as well the existing `map` method for efficient batching of examples.

Key changes:
- Add `batch` method to `Dataset` class in `arrow_dataset.py`
- Utilize `map` method for batching

Closes #7063

Once the approach is approved, i will create the tests and update the documentation.

### DIFF
--- a/docs/source/process.mdx
+++ b/docs/source/process.mdx
@@ -546,6 +546,36 @@ The following example shows how you can use `torch.distributed.barrier` to synch
 ...     torch.distributed.barrier()
 ```
 
+## Batch
+
+The [`~Dataset.batch`] method allows you to group samples from the dataset into batches. This is particularly useful when you want to create batches of data for training or evaluation, especially when working with deep learning models.
+
+Here's an example of how to use the `batch()` method:
+
+```python
+>>> from datasets import load_dataset
+>>> dataset = load_dataset("rotten_tomatoes", split="train")
+>>> batched_dataset = dataset.batch(batch_size=4)
+>>> next(iter(batched_dataset))
+{'text': ['the rock is destined to be the 21st century\'s new " conan " and that he\'s going to make a splash even greater than arnold schwarzenegger , jean-claud van damme or steven segal .',
+        'the gorgeously elaborate continuation of " the lord of the rings " trilogy is so huge that a column of words cannot adequately describe co-writer/director peter jackson\'s expanded vision of j . r . r . tolkien\'s middle-earth .',
+        'effective but too-tepid biopic',
+        'if you sometimes like to go to the movies to have fun , wasabi is a good place to start .'],
+'label': [1, 1, 1, 1]}
+```
+
+The `batch()` method accepts the following parameters:
+
+- `batch_size` (`int`): The number of samples in each batch.
+- `drop_last_batch` (`bool`, defaults to `False`): Whether to drop the last incomplete batch if the dataset size is not divisible by the batch size.
+
+Note that `Dataset.batch()` returns a new [`Dataset`] where each item is a batch of multiple samples from the original dataset. This is different from the batched processing in [`~Dataset.map`], which applies a function to batches but returns a dataset with the same number of items as the original.
+
+<Tip>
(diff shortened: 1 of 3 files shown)


lhoestq:
Looks good to me ! :)

you might want to add the `map` num_proc argument as well, for people who want to make it run faster

lappemic:
Thanks for the feedback @lhoestq! The last commits include:
- Adding the `num_proc` parameter to `batch`
- Adding tests similar to the one done for `IterableDataset.batch()`
- Updated the documentation -> I think they are actually misplaced in the `Stream` page. But could not find a better place atm. Where would you put this documentation?

WDYT?

lhoestq:
You can put the documentation in process.mdx :)

lappemic:
I reset the head to the commit before I added the `Dataset.batch()` documentation to `stream.mdx` and instead added the documentation to `process.mdx`.

docs/source/process.mdx
```diff
@@ -546,6 +546,37 @@ The following example shows how you can use `torch.distributed.barrier` to synch
...
+
+The [`~Dataset.batch`] method allows you to group samples from the dataset into batches. This is particularly useful when you want to create batches of data for training or evaluation, especially when working with deep learning models.
+
+Here's an example of how to use the `batch()` method:
+
+```python
+>>> from datasets import load_dataset
+>>> dataset = load_dataset("rotten_tomatoes", split="train")
+>>> batched_dataset = dataset.batch(batch_size=4)
+>>> next(iter(batched_dataset))
```
lhoestq, line 559:
```suggestion
>>> batched_dataset[0]
```

docs/source/process.mdx
```diff
@@ -546,6 +546,37 @@ The following example shows how you can use `torch.distributed.barrier` to synch
...
+'label': [1, 1, 1, 1]}
+```
+
+The `batch()` method accepts the following parameters:
+
+- `batch_size` (`int`): The number of samples in each batch.
+- `drop_last_batch` (`bool`, defaults to `False`): Whether to drop the last incomplete batch if the dataset size is not divisible by the batch size.
+- `num_proc` (`int`, optional, defaults to `None`): The number of processes to use for multiprocessing. If None, no multiprocessing is used. This can significantly speed up batching for large datasets.
+
+Note that `Dataset.batch()` returns a new [`Dataset`] where each item is a batch of multiple samples from the original dataset. This is different from the batched processing in [`~Dataset.map`], which applies a function to batches but returns a dataset with the same number of items as the original.
```
lhoestq, line 573:
```suggestion
Note that `Dataset.batch()` returns a new [`Dataset`] where each item is a batch of multiple samples from the original dataset. If you want to process data in batches, you should use a batched [`~Dataset.map`] directly, which applies a function to batches but the output dataset is unbatched.
```

docs/source/process.mdx
```diff
@@ -546,6 +546,37 @@ The following example shows how you can use `torch.distributed.barrier` to synch
...
+
+<Tip>
+The batch() method is also available for [IterableDataset] objects. While Dataset.batch() returns a new [Dataset] with batched examples, IterableDataset.batch() returns an [IterableDataset] that yields batches on-the-fly. This makes IterableDataset.batch() more memory-efficient for very large datasets. For more information on batching with [IterableDataset], refer to the [Batch section in Stream](https://huggingface.co/docs/datasets/main/en/stream#batch) documentation.
+</Tip>
+
```
lhoestq, line 578: not needed imo

```suggestion
```

lhoestq (approved):
LGTM thanks ! the CI failures are unrelated to your PR

PR #7068, opened 2024-07-24
Fix prepare_single_hop_path_and_storage_options

albertvillanova:
Fix `_prepare_single_hop_path_and_storage_options`:
- Do not pass HF authentication headers and HF user-agent to non-HF HTTP URLs
- Do not overwrite passed `storage_options` nested values:
  - Before, when passed 
    ```DownloadConfig(storage_options={"https": {"client_kwargs": {"raise_for_status": True}}})```, 
    it was overwritten to 
    ```{"https": {"client_kwargs": {"trust_env": True}}}```
  - Now, the result combines both: 
    ```{"https": {"client_kwargs": {"trust_env": True, "raise_for_status": True}}}```

### DIFF
--- a/src/datasets/utils/file_utils.py
+++ b/src/datasets/utils/file_utils.py
@@ -1173,9 +1173,9 @@ def _prepare_single_hop_path_and_storage_options(
                     urlpath += "&confirm=" + v
                     cookies = response.cookies
                     storage_options[protocol] = {"cookies": cookies, **storage_options.get(protocol, {})}
-        # Fix Google Drive URL to avoid Virus scan warning
-        if "drive.google.com" in urlpath and "confirm=" not in urlpath:
-            urlpath += "&confirm=t"
+            # Fix Google Drive URL to avoid Virus scan warning
+            if "confirm=" not in urlpath:
+                urlpath += "&confirm=t"
         if urlpath.startswith("https://raw.githubusercontent.com/"):
             # Workaround for served data with gzip content-encoding: https://github.com/fsspec/filesystem_spec/issues/389
             headers = storage_options[protocol].setdefault("headers", {})
--- a/src/datasets/utils/file_utils.py
+++ b/src/datasets/utils/file_utils.py
@@ -1159,12 +1159,10 @@ def _prepare_single_hop_path_and_storage_options(
         }
     else:
         storage_options = {}
-    if storage_options:
-        storage_options = {protocol: storage_options}
     if protocol in {"http", "https"}:
-        storage_options[protocol] = {
+        storage_options = {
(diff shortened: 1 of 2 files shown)


src/datasets/utils/file_utils.py
```diff
@@ -1146,7 +1146,7 @@ def _prepare_single_hop_path_and_storage_options(
     Storage options are formatted in the form {protocol: storage_options_for_protocol}
     """
     token = None if download_config is None else download_config.token
-    if urlpath.startswith(config.HF_ENDPOINT) and "/resolve/" in urlpath:
+    if urlpath.startswith(config.HF_ENDPOINT):
```
lhoestq, line 1149: not a big deal but you can revert this no ? if `/resolve/` is not present, the URL points to a web page on HF or an API endpoint that can't be converted to a `hf://` link

src/datasets/utils/file_utils.py
```diff
@@ -1159,40 +1159,34 @@ def _prepare_single_hop_path_and_storage_options(
...
-        storage_options[protocol] = {
-            "headers": {
-                **get_authentication_headers_for_url(urlpath, token=token),
-                "user-agent": get_datasets_user_agent(),
-            },
-            "client_kwargs": {"trust_env": True},  # Enable reading proxy env variables.
-            **(storage_options.get(protocol, {})),
-        }
+    if protocol in {"http", "https"}:
+        client_kwargs = storage_options.setdefault("client_kwargs", {})
```
lhoestq, line 1163: you might want to modify a copy of the user's `storage_options`

src/datasets/utils/file_utils.py
albertvillanova: Thanks for pointing this out.

I wrongly assumed that all and HTTP URL was either convertible to hf:// or a non-HF URL.

Now you make me realize that there is a third option: an HF URL not convertible to hf://
- I think in this case we should pass HF authentication headers and HF user-agent

Do you agree, @lhoestq?

src/datasets/utils/file_utils.py
albertvillanova: Definitely! Thanks for the insight! :sweat_smile:

src/datasets/utils/file_utils.py
lhoestq: for urls non convertible to hf:// we can let the user pass HTTP storage options IMO to keep things simple

albertvillanova: Done.

src/datasets/utils/file_utils.py
albertvillanova: Done.

lhoestq (approved):
great !

PR #7096, opened 2024-08-09
Automatically create `cache_dir` from `cache_file_name`

ringohoffman:
You get a pretty unhelpful error message when specifying a `cache_file_name` in a directory that doesn't exist, e.g. `cache_file_name="./cache/data.map"`

```python
import datasets

cache_file_name="./cache/train.map"
dataset = datasets.load_dataset("ylecun/mnist")
dataset["train"].map(lambda x: x, cache_file_name=cache_file_name)
```

```
FileNotFoundError: [Errno 2] No such file or directory: '/.../cache/tmp48r61siw'
```

It is simple enough to create and I was expecting that this would have been the case.


ringohoffman:
Hi @albertvillanova, is this PR looking okay to you? Anything else you'd like to see?

lhoestq (approved):
cool thanks !

PR #6996, opened 2024-06-25
Remove deprecated code

albertvillanova:
Remove deprecated code, as part of the 3.0 release.

PR #7105, opened 2024-08-15
Use `huggingface_hub` cache

lhoestq:
wip

- use `hf_hub_download()` from `huggingface_hub` for HF files
- `datasets` cache_dir is still used for:
  - caching datasets as Arrow files (that back `Dataset` objects)
  - extracted archives, uncompressed files
  - files downloaded via http (datasets with scripts)
- I removed code that were made for http files (and also the dummy_data / mock_download_manager stuff that happened to rely on them and have been legacy for a while now)


src/datasets/utils/file_utils.py
```diff
@@ -190,32 +169,52 @@ def cached_path(
...
+                    force_download=download_config.force_download,
+                    proxies=download_config.proxies,
+                )
+            except (
+                huggingface_hub.utils.RepositoryNotFoundError,
+                huggingface_hub.utils.EntryNotFoundError,
+                huggingface_hub.utils.RevisionNotFoundError,
+                huggingface_hub.utils.GatedRepoError,
+            ) as e:
+                raise FileNotFoundError(str(e)) from e
```
lhoestq, line 201: that's the important part cc @Wauplin

src/datasets/utils/file_utils.py
```diff
@@ -190,32 +169,52 @@ def cached_path(
 
     if is_remote_url(url_or_filename):
         # URL, so get it from the cache (downloading if necessary)
+        # HF HTTP URLs are prepared as to hf:// URLs
```
Wauplin, line 172: not sure how to interpret this comment

src/datasets/utils/file_utils.py
```diff
@@ -190,32 +169,52 @@ def cached_path(
...
-            download_desc=download_config.download_desc,
-            disable_tqdm=download_config.disable_tqdm,
-        )
+        if url_or_filename.startswith("hf://"):
+            resolved_path = huggingface_hub.HfFileSystem(
+                endpoint=config.HF_ENDPOINT, token=download_config.token
+            ).resolve_path(url_or_filename)
+            try:
+                output_path = huggingface_hub.HfApi(
+                    endpoint=config.HF_ENDPOINT,
```
Wauplin, line 182:
```suggestion
```

This is already the default value in `huggingface_hub` (parsed from the same environment variable)

src/datasets/utils/file_utils.py
```diff
@@ -190,32 +169,52 @@ def cached_path(
...
+                    token=download_config.token,
+                    library_name="datasets",
+                    library_version=__version__,
+                    user_agent=get_datasets_user_agent(download_config.user_agent),
+                ).hf_hub_download(
```
Wauplin, line 187: Top!

src/datasets/utils/file_utils.py
```diff
@@ -190,32 +169,52 @@ def cached_path(
...
-            max_retries=download_config.max_retries,
-            token=download_config.token,
-            ignore_url_params=download_config.ignore_url_params,
-            storage_options=storage_options,
-            download_desc=download_config.download_desc,
-            disable_tqdm=download_config.disable_tqdm,
-        )
+        if url_or_filename.startswith("hf://"):
+            resolved_path = huggingface_hub.HfFileSystem(
+                endpoint=config.HF_ENDPOINT, token=download_config.token
```
Wauplin, line 178:
```suggestion
                token=download_config.token
```

This is already the default value in huggingface_hub (parsed from the same environment variable)

src/datasets/utils/file_utils.py
```diff
@@ -1172,7 +913,7 @@ def _prepare_single_hop_path_and_storage_options(
         client_kwargs = storage_options.pop("client_kwargs", {})
         storage_options["client_kwargs"] = {"trust_env": True, **client_kwargs}  # Enable reading proxy env variables
         if "drive.google.com" in urlpath:
-            response = http_head(urlpath)
+            response = requests.head(urlpath, timeout=10)
```
Wauplin, line 912: Same comment as above about `requests.head` vs `get_session().head`. Maybe worth doing a pass on the `datasets` codebase in a separate PR

Wauplin (commented):
Thanks for working on this @lhoestq! :tada: :tada: :tada: 

I did a first pass and left a few minor comments. Looks good!

src/datasets/utils/file_utils.py
lhoestq: `datasets` users can modify `config. HF_ENDPOINT ` so I'd rather keep it

src/datasets/utils/file_utils.py
lhoestq: I'll make it clearer. Basically even if someone passes a HTTP HF URL like `https://huggingface.co/datasets/.../data.csv` then it will be converted to `hf://datasets/.../data.csv`, that we can parse to get repo_id/revision/etc

src/datasets/utils/file_utils.py
lhoestq: they are the only uses of `requests` in `datasets` left :) I'll use `get_session().head` though it's not a big deal imo

src/datasets/utils/file_utils.py
Wauplin: ah I see, thanks :+1:

src/datasets/utils/file_utils.py
Wauplin: Do you mean users monkey-patching a constant value at runtime without using the environment variable? I feel this is not something we should promote/support

lhoestq: `dataset-viewer` does it in its tests to switch between prod and testing endpoints :p

Wauplin: ok ok, maybe a topic for a separate PR then. It still feels wrong to me to handle endpoints in various places (both in `huggingface_hub` and in `datasets`)

src/datasets/utils/file_utils.py
Wauplin: great to know!

severo:
Nice

Capture d’écran 2024-08-19 à 15 25 00

Wauplin (approved):
Thanks @lhoestq! I thought the switch would be a much more complex process but happy to realize it's not! :smile: I re-reviewed the PR and it looks good to me -to the extent of my knowledgeable-. Better to have other pairs of eyes for this one :)

lhoestq:
fyi the CI failure on test_py310_numpy2 is unrelated to this PR (it's a dependency install failure)

albertvillanova (commented):
The CI error on test_py310_numpy2 has been temporarily fixed by:
- #7114

src/datasets/commands/dummy_data.py
```diff
@@ -1,468 +0,0 @@
-import fnmatch
```
albertvillanova, line 1: The removal of deprecated code has been addressed in a separate dedicated PR:
- #6996: https://github.com/huggingface/datasets/pull/6996/commits/5d4687e4ee1dba2b14e59117492d7f75ea66800a

albertvillanova (commented):
The removal of deprecated code has been addressed in a separate dedicated PR:
-

albertvillanova (approved):
Thanks.

julien-c:
yay! is this in a shipped release?

lhoestq:
we can do one in the coming days once @albertvillanova is back

albertvillanova:
We have made a release and this feature is now included.

PR #7106, opened 2024-08-16
Rename LargeList.dtype to LargeList.feature

albertvillanova:
Rename `LargeList.dtype` to `LargeList.feature`.

Note that `dtype` is usually used for NumPy data types ("int64", "float32",...): see `Value.dtype`.

However, `LargeList` attribute (like `Sequence.feature`) expects a `FeatureType` instead.


lhoestq (approved):
good idea

PR #7110, opened 2024-08-20
Fix ConnectionError for gated datasets and unauthenticated users

### LINKED ISSUE #7109: ConnectionError for gated datasets and unauthenticated users
Since the Hub returns dataset info for gated datasets and unauthenticated users, there is dead code: https://github.com/huggingface/datasets/blob/98fdc9e78e6d057ca66e58a37f49d6618aab8130/src/datasets/load.py#L1846-L1852

We should remove the dead code and properly handle this case: currently we are raising a `ConnectionError` instead of a `DatasetNotFoundError` (as before).

See:
- https://github.com/huggingface/dataset-viewer/issues/3025
- https://github.com/huggingface/huggingface_hub/issues/2457

albertvillanova:
Fix `ConnectionError` for gated datasets and unauthenticated users. See:
- https://github.com/huggingface/dataset-viewer/issues/3025

Note that a recent change in the Hub returns dataset info for gated datasets and unauthenticated users, instead of raising a `GatedRepoError` as before. See:
- https://github.com/huggingface/huggingface_hub/issues/2457

This PR adds an additional check (/auth-check) for gated datasets and raises `DatasetNotFoundError` for unauthenticated users, as it was the case before the change in the Hub.
- Fix suggested by @Pierrci (thanks @Wauplin for pointing it out).

Fix #7109.

### DIFF
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -82,7 +82,7 @@ from .utils.file_utils import (
     relative_to_absolute_path,
     url_or_path_join,
 )
-from .utils.hub import check_auth_for_gated_repo, hf_dataset_url
+from .utils.hub import check_auth, hf_dataset_url
 from .utils.info_utils import VerificationMode, is_small_dataset
 from .utils.logging import get_logger
 from .utils.metadata import MetadataConfigs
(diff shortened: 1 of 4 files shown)


albertvillanova:
Note that the CI error is unrelated to this PR and should be addressed in another PR.

Wauplin (approved):
Looks good to me! We might add support for the `/auth-check` endpoint in `huggingface_hub` directly in the future.

Pierrci:
lgtm!

PR #7133, opened 2024-08-30
remove filecheck to enable symlinks

fschlatt:
Enables streaming from local symlinks #7083 

@lhoestq


src/datasets/data_files.py
```diff
@@ -385,9 +385,8 @@ def resolve_pattern(
         glob_kwargs["expand_info"] = False
     matched_paths = [
         filepath if filepath.startswith(protocol_prefix) else protocol_prefix + filepath
-        for filepath, info in fs.glob(pattern, detail=True, **glob_kwargs).items()
-        if info["type"] == "file"
-        and (xbasename(filepath) not in files_to_ignore)
+        for filepath in fs.glob(pattern, **glob_kwargs).items()
```
lhoestq, line 388: (fix)
```suggestion
        for filepath in fs.glob(pattern, **glob_kwargs)
```

lhoestq:
The CI is failing, looks like it breaks imagefolder loading.

I just checked fsspec internals and maybe instead we can detect symlink by checking `islink` and `size` to make sure it's a file
```python
if info["type"] == "file" or (info.get("islink") and info["size"])
```

lhoestq:
hmm actually `size` doesn't seem to filter symlinked directories, we need another way

fschlatt:
Does fsspec perhaps allow resolving symlinks? Something like https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve

lhoestq:
there is `info["destination"]` in case of a symlink, so maybe


```python
if info["type"] == "file" or (info.get("islink") and info.get("destination") and os.path.isfile(info["destination"]))
```

fschlatt:
I've added a fix which works with some temporary test files locally 

`(info["type"] == "file" or (info.get("islink") and os.path.isfile(os.path.realpath(filepath))))`

lhoestq (approved):
Great ! lgtm :)

Issue #7142, opened 2024-09-08
Specifying datatype when adding a column to a dataset.


varadhbhatnagar:
### Feature request

There should be a way to specify the datatype of a column in `datasets.add_column()`. 

### Motivation

To specify a custom datatype, we have to use `datasets.add_column()` followed by `datasets.cast_column()` which is slow for large datasets. Another workaround is to pass a `numpy.array()` of desired type to the `datasets.add_column()` function.

IMO this functionality should be natively supported.

https://discuss.huggingface.co/t/add-column-with-a-particular-type-in-datasets/95674

### Your contribution

I can submit a PR for this.

varadhbhatnagar:
#self-assign

PR #7143, opened 2024-09-08
Modify add_column() to optionally accept a FeatureType as param

varadhbhatnagar:
[Open Issue](https://github.com/huggingface/datasets/issues/7142)

**Before (Add + Cast)**:

```
from datasets import load_dataset, Value
ds = load_dataset("rotten_tomatoes", split="test")
lst = [i for i in range(len(ds))]

ds = ds.add_column("new_col", lst)
# Assigns int64 to new_col by default
print(ds.features)

ds = ds.cast_column("new_col", Value(dtype="uint16", id=None))
print(ds.features)
```

**Before (Numpy Workaround)**:

```
import numpy as np
ds = load_dataset("rotten_tomatoes", split="test")
lst = [i for i in range(len(ds))]

ds = ds.add_column("new_col", np.array(lst, dtype=np.uint16))
print(ds.features)
```

**After**:
```
from datasets import load_dataset, Value
ds = load_dataset("rotten_tomatoes", split="test")
lst = [i for i in range(len(ds))]
val = Value(dtype="uint16", id=None))
ds = ds.add_column("new_col", lst, feature=val)
print(ds.features)
```

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -5613,7 +5613,9 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
 
     @transmit_format
     @fingerprint_transform(inplace=False)
-    def add_column(self, name: str, column: Union[list, np.array], new_fingerprint: str):
+    def add_column(
+        self, name: str, column: Union[list, np.array], new_fingerprint: str, feature: Optional[FeatureType] = None
+    ):
         """Add column to Dataset.
 
         <Added version="1.7"/>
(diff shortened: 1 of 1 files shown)


varadhbhatnagar:
Requesting review @lhoestq 
I will also update the docs if this looks good.

lhoestq:
Cool ! maybe you can rename the argument `feature` and with type `FeatureType` ? This way it would work the same way as `.cast_column()` ?

lhoestq (commented):
LGTM ! Can you just add the `feature` argument to the docstrings before we merge ?

varadhbhatnagar:
@lhoestq done!

varadhbhatnagar:
@lhoestq anything pending on this?

lhoestq (approved):
Thanks ! LGTM :)

Issue #6880, opened 2024-05-07
Webdataset: KeyError: 'png' on some datasets when streaming


lhoestq:
reported at https://huggingface.co/datasets/tbone5563/tar_images/discussions/1

```python
>>> from datasets import load_dataset
>>> ds = load_dataset("tbone5563/tar_images")
Downloading data: 100%
 1.41G/1.41G [00:48<00:00, 17.2MB/s]
Downloading data: 100%
 619M/619M [00:11<00:00, 57.4MB/s]
Generating train split: 
 970/0 [00:02<00:00, 534.94 examples/s]
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
[/usr/local/lib/python3.10/dist-packages/datasets/builder.py](https://localhost:8080/#) in _prepare_split_single(self, gen_kwargs, fpath, file_format, max_shard_size, split_info, check_duplicate_keys, job_id)
   1747                 _time = time.time()
-> 1748                 for key, record in generator:
   1749                     if max_shard_size is not None and writer._num_bytes > max_shard_size:

7 frames
[/usr/local/lib/python3.10/dist-packages/datasets/packaged_modules/webdataset/webdataset.py](https://localhost:8080/#) in _generate_examples(self, tar_paths, tar_iterators)
    108                 for field_name in image_field_names + audio_field_names:
--> 109                     example[field_name] = {"path": example["__key__"] + "." + field_name, "bytes": example[field_name]}
    110                 yield f"{tar_idx}_{example_idx}", example

KeyError: 'png'

The above exception was the direct cause of the following exception:

DatasetGenerationError                    Traceback (most recent call last)
[<ipython-input-2-8e0fbb7badc9>](https://localhost:8080/#) in <cell line: 3>()
      1 from datasets import load_dataset
      2 
----> 3 ds = load_dataset("tbone5563/tar_images")

[/usr/local/lib/python3.10/dist-packages/datasets/load.py](https://localhost:8080/#) in load_dataset(path, name, data_dir, data_files, split, cache_dir, features, download_config, download_mode, verification_mode, ignore_verifications, keep_in_memory, save_infos, revision, token, use_auth_token, task, streaming, num_proc, storage_options, trust_remote_code, **config_kwargs)
   2607 
   2608     # Download and prepare data
-> 2609     builder_instance.download_and_prepare(
   2610         download_config=download_config,
   2611         download_mode=download_mode,

[...]

PR #6888, opened 2024-05-09
Support WebDataset containing file basenames with dots

albertvillanova:
Support WebDataset containing file basenames with dots.

Fix #6880.

PR #7144, opened 2024-09-11
Fix key error in webdataset

ragavsachdeva:
I was running into

```
example[field_name] = {"path": example["__key__"] + "." + field_name, "bytes": example[field_name]}
KeyError: 'png'
```

The issue is that a filename may have multiple "." e.g. `22.05.png`. Changing `split` to `rsplit` fixes it.

Related https://github.com/huggingface/datasets/issues/6880

### DIFF
--- a/src/datasets/packaged_modules/webdataset/webdataset.py
+++ b/src/datasets/packaged_modules/webdataset/webdataset.py
@@ -29,7 +29,7 @@ class WebDataset(datasets.GeneratorBasedBuilder):
         streaming_download_manager = datasets.StreamingDownloadManager()
         for filename, f in tar_iterator:
             if "." in filename:
-                example_key, field_name = filename.split(".", 1)
+                example_key, field_name = filename.rsplit(".", 1)
                 if current_example and current_example["__key__"] != example_key:
                     yield current_example
                     current_example = {}


lhoestq:
hi ! What version of `datasets` are you using ? Is this issue also happening with `datasets==3.0.0` ?
Asking because we made sure to replicate the official webdataset logic, which is to use the latest dot as separator between the sample base name and the key

ragavsachdeva:
Hi, yes this is still a problem on `datasets==3.0.0`.

I was using `datasets=2.20.0` and in that version you get the key error.

I just upgraded to `datasets==3.0.0` and in this version, you do not get a key error because it sets all keys to none by default   in `_generate_examples` function:

```python
if field_name not in example:
    example[field_name] = None
```

However, the behaviour is still incorrect, This if condition is triggered because the filename is not split properly and it returns the data as `None` when it shouldn't.
 
Using `rsplit` instead of `split` fixes this and in general, should not a breaking change I believe.

albertvillanova:
Hi  @ragavsachdeva,

We already had this discussion in the issue you have linked:
- #6880
- even I proposed a PR with your proposed fix:
  - #6888

However, we decided not to implement this feature because it is NOT aligned with the behavior of the `webdataset` library:
> The prefix of a file is all directory components of the file plus the file name component up to the *first* “.” in the file name.
```python
In [1]: import webdataset as wds

In [2]: wds.tariterators.base_plus_ext("22.05.png")
Out[2]: ('22', '05.png')
```

ragavsachdeva:
Ah, my apologies I missed https://github.com/huggingface/datasets/pull/6888 (clearly didn't do my due diligence). It's such a weird convention to have though. My keys are `/some/path/22.0/1.1.png` and it splits them at  `/some/path/22` and `.0/1.1.png`(!) I'm okay with this PR not being merged though. Thanks for your time.

lhoestq:
Actually our code should not split as `.0/1.1.png` - even webdataset handles this correctly via their regex `^((?:.*/|)[^.]+)[.]([^/]*)$` here:

https://github.com/webdataset/webdataset/blob/87bd5aa41602d57f070f65a670893ee625702f2f/webdataset/tariterators.py#L36

ragavsachdeva:
Oh.. the intention with that regex is to capture "multi-part" extensions e.g. `.tar.gz`. Makes sense. So `rsplit` isn't the solution then and neither is `split`. This expression makes so much more sense. Nice find! I'm assuming you'll add a patch?

albertvillanova:
Issue addressed by:
- #7151

phunc20:
Maybe the libray should suggest/warn users about **file naming**, in particular, file naming **with dots**. Indeed, otherwise a naive naming approach like mentioned above would lead to

```python
ValueError: The TAR archives of the dataset should be in WebDataset format, but the files in the archive don't share the same prefix or the same types.
```

with the following code

```ipython
In [1]: from datasets import load_dataset

In [2]: !tar tvf easy_trouble/001.tar
drwxr-xr-x phunc20/phunc20   0 2025-01-15 17:57 001/
-rw-r--r-- phunc20/phunc20  20 2025-01-15 17:54 001/1.2.jpg
-rw-r--r-- phunc20/phunc20  20 2025-01-15 17:53 001/1.1.jpg
-rw-r--r-- phunc20/phunc20  20 2025-01-15 17:57 001/2.1.jpg
-rw-r--r-- phunc20/phunc20  20 2025-01-15 17:57 001/2.2.jpg
-rw-r--r-- phunc20/phunc20  20 2025-01-15 17:57 001/2.3.jpg

In [29]: dataset = load_dataset("webdataset", data_dir="easy_trouble")
```

BTW, @ragavsachdeva , why would you find including a `.tar.gz` inside a `.tar` makes sense? I personally find switching `base_plus_ext` into sth similar to the following makes more sense, though.

```python
from pathlib import Path

def base_plus_ext(path):
    """Split off all file extensions.

    Returns base, allext.
    """
    p = Path(path)
    base = (p.parent / p.stem).as_posix()
    ext = p.suffix.split(".")[-1]
    return base, ext
```

PR #7165, opened 2024-09-24
fix increase_load_count

lhoestq:
it was failing since 3.0 and therefore not updating download counts on HF or in our dashboard


lhoestq:
I tested a few load_dataset and they do show up in download stats now

albertvillanova:
Thanks for having noticed and fixed.

PR #7181, opened 2024-09-29
Fix datasets export to JSON

varadhbhatnagar:


varadhbhatnagar:
Linked Issue: #7037
Ideas: #7039

varadhbhatnagar:
@albertvillanova any early feedback?

varadhbhatnagar:
`orient` = "split" can also be handled. I will add the changes soon

lhoestq:
Thanks for diving into this ! I don't think we want the JSON export to be that complex though, especially if people can do `ds.to_pandas().to_json(orient=...)`. Maybe we can just raise an error and suggest users to use pandas ? And also note that it loads the full dataset in memory so it's mainly for small scale datasets. The only acceptable option for large scale datasets is probably just JSON Lines anyway since it enables streaming.

lhoestq:
Yes indeed. Though I think it's fine since using something else than orient="lines" is only suitable/useful for small datasets. Or you know a case where a big dataset need to be in a format that is not orient="lines" ?

varadhbhatnagar:
@lhoestq Let me close this PR and open another one where I will add an error message, as suggested here.

> Thanks for diving into this ! I don't think we want the JSON export to be that complex though, especially if people can do `ds.to_pandas().to_json(orient=...)`. Maybe we can just raise an error and suggest users to use pandas ? And also note that it loads the full dataset in memory so it's mainly for small scale datasets. The only acceptable option for large scale datasets is probably just JSON Lines anyway since it enables streaming.

varadhbhatnagar:
Addressed here: #7273 
@lhoestq

PR #7182, opened 2024-09-30
Support features in metadata configs

albertvillanova:
Support features in metadata configs, like:
```
configs:
  - config_name: default
    features:
      - name: id
        dtype: int64
      - name:  name
        dtype: string
      - name: score
        dtype: float64
```

This will allow to avoid inference of data types.

Currently, we allow passing this information in the `dataset_info` (instead of `configs`) field, but this is not intuitive and it is not properly documented.

TODO:
- [ ] Document usage


albertvillanova:
The CI issue is unrelated:
- #7183

lhoestq (approved):
lgtm, let's see if it works and we can add it to the datasets hub documentation for the viewer (e.g. "how to enforce certain column types in the viewer")

Issue #7206, opened 2024-10-08
Slow iteration for iterable dataset with numpy formatting for array data


alex-hh:
### Describe the bug

When working with large arrays, setting with_format to e.g. numpy then applying map causes a significant slowdown for iterable datasets.

### Steps to reproduce the bug

```python
features=Features(**{"array0": Array3D((None, 10, 10), dtype="float32"), "array1": Array3D((None,10,10), dtype="float32")})
dataset = Dataset.from_dict({f"array{i}": [np.zeros((x,10,10), dtype=np.float32) for x in [2000,1000]*25] for i in range(2)}, features=features)```

Then
```
ds = dataset.to_iterable_dataset()
ds = ds.with_format("numpy").map(lambda x: x)
t0 = time.time()
for ex in ds:
    pass
t1 = time.time()
print(t1-t0)
```

takes 27 s, whereas

```python
ds = dataset.to_iterable_dataset()
ds = ds.with_format("numpy")
ds = dataset.to_iterable_dataset()
t0 = time.time()
for ex in ds:
    pass
t1 = time.time()
print(t1 - t0)
```

takes ~1s

### Expected behavior

Map should not introduce a slowdown when formatting is enabled.

### Environment info

3.0.2

PR #7207, opened 2024-10-08
apply formatting after iter_arrow to speed up format -> map, filter for iterable datasets

alex-hh:
I got to this by hacking around a bit but it seems to solve #7206 

I have no idea if this approach makes sense or would break something else? 

Could maybe work on a full pr if this looks reasonable @lhoestq ? I imagine the same issue might affect other iterable dataset methods?


lhoestq (commented):
Interesting ! Let's validate if we get the same speed on those 2 ?
1. Dataset.to_iterable_dataset -> numpy format
1. Dataset.to_iterable_dataset -> numpy format -> batched map identity (batched could be quite important)

src/datasets/iterable_dataset.py
```diff
@@ -926,7 +926,7 @@ def __init__(
 
     @property
     def iter_arrow(self):
-        if self.formatting and self.formatting.format_type == "arrow":
+        if self.formatting and (self.formatting.format_type == "arrow" or self.ex_iterable.iter_arrow):
```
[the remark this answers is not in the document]
alex-hh: agreed, I wasn't really understanding the flow - have updated with a more direct attempt

alex-hh:
With updated version:

```python
import numpy as np
import time
from datasets import Dataset, Features, Array3D

features=Features(**{"array0": Array3D((None, 10, 10), dtype="float32"), "array1": Array3D((None,10,10), dtype="float32")})
dataset = Dataset.from_dict({f"array{i}": [np.zeros((x,10,10), dtype=np.float32) for x in [2000,1000]*25] for i in range(2)}, features=features)
```

```python
ds = dataset.to_iterable_dataset()
ds = ds.with_format("numpy").map(lambda x: x, batched=True, batch_size=10)
t0 = time.time()
for ex in ds:
    pass
t1 = time.time()
```
Total time: < 0.01s

```python
ds = dataset.to_iterable_dataset()
ds = ds.with_format("numpy").map(lambda x: x, batched=False)
t0 = time.time()
for ex in ds:
    pass
t1 = time.time()
```
Time: ~0.02 s

```python
ds = dataset.to_iterable_dataset()
ds = ds.with_format("numpy")
t0 = time.time()
for ex in ds:
    pass
t1 = time.time()
```
Time: ~0.02s

alex-hh:
also now working for filter:

```python
filtered_examples = []
ds = dataset.to_iterable_dataset()
ds = ds.with_format("numpy").filter(lambda x: [arr.shape[0]==2000 for arr in x["array0"]], batch_size=10, batched=True)
t0 = time.time()
for ex in ds:
    filtered_examples.append(ex)
t1 = time.time()
```

```python
filtered_examples = []
ds = dataset.to_iterable_dataset()
ds = ds.with_format("numpy").filter(lambda x: x["array0"].shape[0]==2000, batched=False)
t0 = time.time()
for ex in ds:
    filtered_examples.append(ex)
t1 = time.time()
```

src/datasets/iterable_dataset.py
```diff
@@ -2211,6 +2250,7 @@ def map(
         remove_columns: Optional[Union[str, List[str]]] = None,
         features: Optional[Features] = None,
         fn_kwargs: Optional[dict] = None,
+        format_outputs: bool = True,
```
lhoestq, line 2253: is this useful ?

lhoestq:
(the distributed tests failing in the CI are unrelated)

src/datasets/iterable_dataset.py
alex-hh: was useful for optimising performance in my case - 

idea is it allows the user to determine the formatting of the examples returned by map (within the mapped function) rather than via self.formatting, which determines the formatting of the inputs to the map function

if map returns lists / arbitrary types then re-applying the formatter to the outputs can be expensive

However, haven't thought about how to handle self.formatting for downstream dataset transformations...might require something better than this kwarg

lhoestq: You can still do `ds = ds.map().with_format(None)` and no formatting will be applied to the output

alex-hh:
There also appears to be an issue with chaining filter and map bc filter iter_arrow only returns _iter_arrow if arrow formatting is applied (and vv presumably)

src/datasets/iterable_dataset.py
```diff
@@ -1209,57 +1243,62 @@ def _iter(self):
...
-            format_dict = (
-                formatter.recursive_tensorize if isinstance(formatter, TensorFormatter) else cast_to_python_objects
+        formatter = get_formatter(self.formatting.format_type) if self.formatting else None
+        if self.formatting and self.ex_iterable.iter_arrow:
+            # we still want to use an arrow iterator, yielding single batches of size self.batch_size
+            # to which the formatter can be applied
+            ex_iterable = RebatchedArrowExamplesIterable(
+                self.ex_iterable, batch_size=self.batch_size if self.batched else 1, drop_last_batch=False
             )
+            batched_examples_iterator = formatted_arrow_examples_iterator(ex_iterable, formatter, batched=self.batched)
```
lhoestq, line 1254: maybe name it `input_iterator ` or something like that since it's not necessarily batched ?

src/datasets/iterable_dataset.py
```diff
@@ -1209,57 +1243,62 @@ def _iter(self):
...
+                    function_args.append([current_idx + i for i in range(batch_len)])
                 mask = self.function(*function_args, **self.fn_kwargs)
                 # yield one example at a time from the batch
-                for key_example, to_keep in zip(key_examples_list, mask):
+                examples = _batch_to_examples(batch)
+                # TODO: nicer way to handle keys?
+                if not self.formatting:
+                    keys = combined_key.split("_")
+                else:
+                    keys = [combined_key] * len(mask)
```
lhoestq, line 1283: this should be fine
```suggestion
                keys = [combined_key] * len(mask)
```

src/datasets/iterable_dataset.py
alex-hh: perfect - will remove the kwarg

lhoestq:
> issue with chaining filter and map bc filter iter_arrow only returns _iter_arrow if arrow formatting is applied (and vv presumably)

Maybe related to this issue ?

```python
ds = Dataset.from_dict({"a": range(10)}).to_iterable_dataset()
ds = ds.with_format("arrow").map(lambda x: x, features=Features({"a": Value("string")})).with_format(None)
print(list(ds))  # yields integers instead of strings
```

src/datasets/iterable_dataset.py
alex-hh: updated

lhoestq:
I feel like we could get rid of TypedExampleIterable altogether and apply formatting with feature conversion with `formatted_python_examples_iterator ` and `formatted_arrow_examples_iterator`

btw you can pass `features=` in `get_formatter()` to get a formatter that does the feature conversion at the same time as formatting

alex-hh:
> I feel like we could get rid of TypedExampleIterable altogether and apply formatting with feature conversion with formatted_python_examples_iterator  and formatted_arrow_examples_iterator

Oh nice didn't know about the feature support in get_formatter. Haven't thought through whether this works but would a FormattedExampleIterable (with feature conversion) be able to solve this and fit the API better?

lhoestq:
> Oh nice didn't know about the feature support in get_formatter. Haven't thought through whether this works but would a FormattedExampleIterable (with feature conversion) be able to solve this and fit the API better?

Yes this is surely the way to go actually !

alex-hh:
ok i've fixed the chaining issue with my last two commits.

Will see if I can refactor into a FormattedExampleIterable

The other issue you posted seems to be unrelated (maybe something to do with how Values are decoded?)

alex-hh:
updated with FormattedExamplesIterable.

It might be possible to do a wider refactor and use FormattedExamplesIterable elsewhere.

But I'd prefer not to try that rn.

src/datasets/iterable_dataset.py
```diff
@@ -1218,13 +1265,19 @@ def __init__(
 
     @property
     def iter_arrow(self):
-        if self.formatting and self.formatting.format_type == "arrow":
+        # if self.formatting is not None, we format, filter, then return arrow iterator
+        # if formatting is unset, we don't, as that would require filter fn to accept arrow format
+        if self.formatting and self.ex_iterable.iter_arrow:
```
lhoestq, line 1270: I think it's ok to not return iter_arrow if the formatting is not "arrow", this way we decode the data only once in the pipeline (e.g. if it contains image/audio data)

alex-hh:
Thinking about this in the context of #e/datasets/issues/7210 - am wondering if it would make sense for Features to define their own extraction arrow->object logic? e.g. Arrays should *always* be extracted with NumpyArrowExtractor, not only in case with_format is set to numpy (which a user can easily forget to do)

lhoestq:
> Thinking about this in the context of https://github.com/huggingface/datasets/issues/7210 - am wondering if it would make sense for Features to define their own extraction arrow->object logic? e.g. Arrays should always be extracted with NumpyArrowExtractor, not only in case with_format is set to numpy (which a user can easily forget or not know to do)

For `ArrayND` they already implement `to_pylist` to decode arrow data and it can be updated to return a numpy array (see the `ArrayExtensionArray` class for more details)

src/datasets/iterable_dataset.py
```diff
@@ -2277,11 +2372,17 @@ def with_format(
...
+            features=self.features,
+            token_per_repo_id=self._token_per_repo_id,
+        )
         return IterableDataset(
-            ex_iterable=self._ex_iterable,
+            ex_iterable=ex_iterable,
             info=self._info.copy(),
             split=self._split,
-            formatting=FormattingConfig(format_type=type),
+            formatting=None,  # formatting is applied in ex_iterable
```
lhoestq, line 2385: Originally we tried to keep `formatting` as an attribute of the `IterableDataset` so that people can change the formatting without having to create nested examples iterables that would apply a different formatting every time.

What is the rationale for removing it from here ?

Note that applying a formatting on data already in the right format is cheap.

src/datasets/iterable_dataset.py
```diff
@@ -2396,15 +2514,19 @@ def map(
...
+            return_features=features,
         )
         info = self.info.copy()
         info.features = features
         return IterableDataset(
             ex_iterable=ex_iterable,
             info=info,
             split=self._split,
-            formatting=self._formatting,
+            formatting=None,  # formatting is applied in mapped ex_iterable; no need to re-apply it in IterableDataset
```
lhoestq, line 2529: you mean formatting is already applied to the input, but not to the output afaik ?

src/datasets/iterable_dataset.py
```diff
@@ -941,7 +974,8 @@ def __init__(
         self.with_indices = with_indices
         self.input_columns = input_columns
         self.fn_kwargs = fn_kwargs or {}
-        self.formatting = formatting
+        self.formatting = formatting  # required for iter_arrow
+        self._return_features = return_features
```
lhoestq, line 978: btw you can call this `features` (to be aligned with .map() for which `features=` corresponds to the output features)

src/datasets/iterable_dataset.py
```diff
@@ -1076,6 +1114,12 @@ def _iter(self):
                 if self.remove_columns:
                     for c in self.remove_columns:
                         del transformed_example[c]
+                if self._return_features:
+                    for c in self._return_features.keys():
+                        if c not in transformed_example:
+                            transformed_example[c] = None
+                    # TODO: check types
+                    transformed_example = self._return_features.decode_example(transformed_example)
```
lhoestq, line 1122: is it expected that it is only applied to the unbatched case ?

lhoestq (commented):
I left a few more comments, this is so close to be ready !

src/datasets/iterable_dataset.py
alex-hh: yes - i guess what i was thinking is that if a user calls

```
ds.with_format(format_type).map(function)
```

then they are responsible for choosing the format applied to the mapped function outputs:

if the mapped function does nothing, then the formatting will be format_type.

it doesn't seem ideal to re-format the users outputs (providing no way of overriding this).

e.g if i do with_format("numpy").map(fn) where fn extracts some python object from numpy arrays, I don't want to call np.asarray(obj) on the resulting object

src/datasets/iterable_dataset.py
alex-hh: c.f. comment https://github.com/huggingface/datasets/pull/7207#discussion_r1800014191

my thinking was that this way the formatting is never `lost'; it only gets overridden if the user overrides it.

there could indeed be some nested cases that i didn't consider though, can you think of some?

src/datasets/iterable_dataset.py
alex-hh: updated

src/datasets/iterable_dataset.py
alex-hh: nope! fixed thanks for spotting

src/datasets/iterable_dataset.py
lhoestq: By nested I actually meant consecutive, I had in mind that this code would be a no-op:
```python
format = ds.format
ds = ds.with_format("np")
ds = ds.with_format(**format)
```

(actually just noticed `IterableDataset.format` is not implemented yet, though it should be and work like `Dataset.format`, but anyway the point is successive `with_format` would override the previous ones since it's an operation that is done "on top" of the dataset rather than a processing function)

src/datasets/iterable_dataset.py
lhoestq: > it doesn't seem ideal to re-format the user's outputs (providing them with no way of overriding this).
> e.g if i do with_format("numpy").map(fn) where fn extracts some python object from numpy arrays, I don't want to have np.asarray(obj) being called on the resulting object by the numpy formatter

You can keep in `self._formatting` because `np.asarray(obj)` is cheap on an object that is already a numpy array (zero copy). And if someone does this
```python
ds.with_format(format_type).map(function).with_format(another_type)
```
then np.asarray would not be called on the output since we can just replace the `IterableDataset._formatting` with the new one

alex-hh: I think my question around this was that there is no way to just prevent any formatting happening on the outputs of map, because the user can't set _formatting to None (not what with_format(None) does).

So if map already takes care of formatting it risks introducing an expensive re-formatting step.

lhoestq: Maybe `with_format(None)` should imply no formatting (contrary to the current logic which is python formatting) ? would that work ?

alex-hh: I think so!

lhoestq: ok let's do this then, I think it should do the job ! feel free to ping me if there is anything blocking

alex-hh: Kind of still feel like setting formatting to None should be default behaviour for map? it's such an easy trap to fall into and can be expensive with any kind of complex transformation in map

src/datasets/iterable_dataset.py
alex-hh: reverted this case so with_format.with_format can still no op

src/datasets/iterable_dataset.py
lhoestq: > On reflection, kind of still feel like setting formatting to None should be default behaviour for map? it's such an easy trap to fall into to not do it and it can be expensive with any kind of complex transformation in map

It would make it inconsistent with `Dataset.map` unfortunately :/
Also I've rarely seen people going from one format to another between maps

alex-hh: Looking at the Dataset.map code, is it right that formatting doesn't get re-applied to output of map?

Preventing that could be another way to fix the iterable dataset case? (with_format None also seems fine but affects a bit more things which makes me slightly nervous...but can do what you think best?)

alex-hh: @lhoestq what is the intended chaining logic? Should with_format.map.with_format override the original with_format and change the formatting of the inputs to map? Or is it intended to apply to the outputs?

Realise I haven't been clear about this in my head so far, and know your comment above was somewhat related but wanted to be sure I understand!

lhoestq: No problem, this is not as easy as it seems though you've been doing great ! Anyway, to answer your questions:

> is it right that formatting doesn't get re-applied to output of Dataset.map?

MappedExamplesIterable doesn't apply formatting, but the IterableDataset does apply formatting if the users did set one.

> e.g. would it potentially be possible to completely remove format_dict in IterableDataset.iter in the case that ex_iterable.iter_arrow evaluates to False, since in this case extraction+formatting has presumably already occurred?

the output of map can be anything in MappedExamplesIterable, it's the IterableDataset's job to format it if the user asks for a specific format. If there is no format set by the user, we yield the example as is (the data can even contain arbitrary python objects). E.g. if a user wants to do some processing from numpy arrays and return arbitrary python objects they should do `ds = ds.with_format("np").map(fn).with_format(None)`

So no I don't think we can remove the format_dict in IterableDataset (or we would have to add FormattedExamplesIterable after each MappedExamplesIterable but this doesn't allow disabling formatting after a map)

> @lhoestq what is the intended chaining logic? Should with_format.map.with_format override the original with_format and change the formatting of the inputs to map? Or is it intended to apply to the outputs?

the second with_format sets the format of the data yielded when iterating on the IterableDataset - it doens't change the input to map

alex-hh: so, just to be sure, the default behaviour of with_format().map() should be to format both the inputs and the outputs of map?

only asking because it looked like for the non-iterable Dataset.map formatting only applies to the inputs but wasn't sure I was understanding that correctly.

lhoestq: the output of Dataset.map is written as Arrow data on disk, but when the data is accessed it's formatted using the Dataset format type (map doesn't change the format of the dataset)

that's why a code likes this still returns numpy arrays:
```python
>>> ds = with_format("np").map(...)
>>> ds[0]  # dict of numpy arrays
```
and it can be reset to python objects with
```python
>>> ds = with_format("np").map(...).with_format("python")
>>> ds[0]  # dict of python objects
```

alex-hh: ok got it thanks!

alex-hh:
@lhoestq after all that im not sure my concern about with_format(None) was well-founded - I didn't appreciate that the python formatter tries to do nothing to python objects including numpy arrays.

I think this PR is ready for review as is.

alex-hh:
@lhoestq I've updated to make compatible with latest changes on main, please let me know if there's anything else I can do!

lhoestq:
Hi Alex, I will be less available from today and for a week. I'll review your PR and play with it once I come back if you don't mind !

lhoestq (approved):
I took the liberty to add a test and do slight changes to fix other tests :) it looks all good now and is much faster than before ! This PR was also needed to further improve polars/pandas/arrow support

alex-hh:
thanks for the reviews and extensions, happy to see this merged !

Issue #7208, opened 2024-10-09
Iterable dataset.filter should not override features


alex-hh:
### Describe the bug

When calling filter on an iterable dataset, the features get set to None

### Steps to reproduce the bug

import numpy as np
import time
from datasets import Dataset, Features, Array3D

```python
features=Features(**{"array0": Array3D((None, 10, 10), dtype="float32"), "array1": Array3D((None,10,10), dtype="float32")})
dataset = Dataset.from_dict({f"array{i}": [np.zeros((x,10,10), dtype=np.float32) for x in [2000,1000]*25] for i in range(2)}, features=features)
ds = dataset.to_iterable_dataset()
orig_column_names = ds.column_names
ds = ds.filter(lambda x: True)
assert ds.column_names == orig_column_names
```

### Expected behavior

Filter should preserve features information

### Environment info

3.0.2

lhoestq:
closed by https://github.com/huggingface/datasets/pull/7209, thanks @alex-hh !

PR #7209, opened 2024-10-09
Preserve features in iterable dataset.filter

alex-hh:
Fixes example in #7208 - I'm not sure what other checks I should do? @lhoestq 

I also haven't thought hard about the concatenate / interleaving example iterables but think this might work assuming that features are either all identical or None?

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -130,6 +130,10 @@ class _BaseExamplesIterable:
     def iter_arrow(self) -> Optional[Callable[[], Iterator[Tuple[Key, pa.Table]]]]:
         return None
 
+    @property
+    def is_typed(self) -> bool:
+        return False
+
     def shuffle_data_sources(self, generator: np.random.Generator) -> "_BaseExamplesIterable":
         """
         Either shuffle the shards/sources of the dataset, or propagate the shuffling to the underlying iterable.
@@ -393,6 +397,10 @@ class RebatchedArrowExamplesIterable(_BaseExamplesIterable):
     def iter_arrow(self):
         return self._iter_arrow
 
+    @property
+    def is_typed(self):
+        return self.ex_iterable.is_typed
+
     def _init_state_dict(self) -> dict:
         self._state_dict = {
             "ex_iterable": self.ex_iterable._init_state_dict(),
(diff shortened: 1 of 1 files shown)


src/datasets/iterable_dataset.py
```diff
@@ -929,6 +961,10 @@ def iter_arrow(self):
         if self.formatting and self.formatting.format_type == "arrow":
             return self._iter_arrow
 
+    @property
+    def is_typed(self):
+        return self.ex_iterable.is_typed
```
lhoestq, line 966: we can't know the output of a mapped iterable, so it should be False no ?
```suggestion
        return False
```

lhoestq:
Yes your assumption on concatenate/interleave is ok imo.

It seems the TypedExamplesIterable can slow down things, it should take formatting into account to not convert numpy arrays to python lists

right now it's slow:

```python
>>> ds = Dataset.from_dict({"a": np.zeros((1000, 32, 32))}).to_iterable_dataset().with_format("np")
>>> filtered_ds = ds.filter(lambda x: True)
>>> %time sum(1 for _ in ds)
CPU times: user 175 ms, sys: 8.1 ms, total: 183 ms
Wall time: 184 ms
1000
>>> %time sum(1 for _ in filtered_ds)
CPU times: user 4.1 s, sys: 8.41 ms, total: 4.1 s
Wall time: 4.12 s
1000
```

src/datasets/iterable_dataset.py
alex-hh: oh yes - should be False

still, wondering what happens if user passes explicit features != None to map:
in that case should the feature conversion be applied or not? If it should be applied I think returning False here works, if it shouldn't we may need to handle that somehow

lhoestq: If we pass features != None to map(), the ExampleIterable is not typed so the IterableDataset will apply the feature conversion in `IterableDataset.__iter__`. And if there is a subsequent map(), it will use a TypedExampleIterable as input to the subsequent map function so the feature conversion is also applied.

alex-hh:
> It seems the TypedExamplesIterable can slow down things, it should take formatting into account to not convert numpy arrays to python lists

Should be fixed by #7207 I hope!

lhoestq (approved):
in any case, LGTM now :)

PR #7229, opened 2024-10-15
handle config_name=None in push_to_hub

alex-hh:
This caught me out - thought it might be better to explicitly handle None?


lhoestq:
not sure it's a good idea, we always need a config name so better have the correct default and not support None (which could lead to think it doesn't have a config name, while it does)

PR #7252, opened 2024-10-25
Add IterableDataset.shard()

lhoestq:
Will be useful to distribute a dataset across workers (other than pytorch) like spark

I also renamed `.n_shards` -> `.num_shards` for consistency and kept the old name for backward compatibility. And a few changes in internal functions for consistency as well (rank, world_size -> num_shards, index)

Breaking change: the new default for `contiguous` in `Dataset.shard()` is `True`, but imo not a big deal since I couldn't find any usage of `contiguous=False` internally (we always do contiguous=True for map-style datasets since its more optimized) or in the wild


jonathanasdf:
Is there some way to get this to work for pytorch dataloader workers?

eg. start with a single sharded IterableDataset.from_generator(), then reshard before calling map() to do expensive processing over multiple workers

PR #7284, opened 2024-11-08
support for custom feature encoding/decoding

alex-hh:
Fix for https://github.com/huggingface/datasets/issues/7220 as suggested in discussion, in preference to #7221 

(only concern would be on effect on type checking with custom feature types that aren't covered by FeatureType?)


alex-hh:
@lhoestq

PR #7294, opened 2024-11-18
Remove `aiohttp` from direct dependencies

akx:
The dependency is only used for catching an exception from other code. That can be done with an import guard.


lhoestq (approved):
great ! I moved it to optional dependencies

PR #7307, opened 2024-12-05
refactor: remove unnecessary else

HarikrishnanBalagopal:


ruidazeng (commented):
I think both`if` or `elif` achieve the same functionality.

Issue #7356, opened 2025-01-06
How about adding a feature to pass the key when performing map on DatasetDict?


jp1924:
### Feature request

Add a feature to pass the key of the DatasetDict when performing map

### Motivation

I often preprocess using map on DatasetDict. 
Sometimes, I need to preprocess train and valid data differently depending on the task. 
So, I thought it would be nice to pass the key (like train, valid) when performing map on DatasetDict. 

What do you think?

### Your contribution

I can submit a pull request to add the feature to pass the key of the DatasetDict when performing map.

jp1924:
@lhoestq 
If it's okay with you, can I work on this?

jp1924:
@lhoestq 
Thanks for the response! 
Let me clarify what I'm looking for with an example:

Currently, we need to write separate processing functions or call .map() separately:
```python
# Current approach
def process_train(example):
    # Training-specific processing
    return example

def process_valid(example):
    # Validation-specific processing
    return example

ds["train"] = ds["train"].map(process_train)
ds["valid"] = ds["valid"].map(process_valid)
```

What I'm proposing is to have a single processing function that knows which split it's processing:

```python
# Proposed feature
def process(example, split_key):
    if split_key == "train":
        # Training-specific processing
    elif split_key == "valid":
        # Validation-specific processing
    return example

# Using with_key=True to pass the split information
ds = ds.map(process, with_key=True)
```

This becomes particularly useful when:
1. The processing logic is heavily shared between splits but needs minor adjustments
2. You want to maintain the processing logic in one place for better maintainability
3. The processing function is complex and you want to avoid duplicating code

I've created a draft PR implementing this: https://github.com/huggingface/datasets/pull/7240/files

lhoestq:
I see ! I think it makes sense, and it's more readable than doing something like this:
```python
from functools import partial
ds = DatasetDict({key: ds[key].map(partial(process, split_key=key)) for key in ds})
```

PS: you named the argument `with_key` in your example, but it might be even clearer with it's named `with_split` maybe no ?

lhoestq:
Sure !

PR #7368, opened 2025-01-13
Add with_split to DatasetDict.map

jp1924:
#7356


jp1924:
Can you check this out, @lhoestq?

jp1924:
cc @lhoestq @albertvillanova

jp1924:
@lhoestq

jp1924:
@lhoestq

src/datasets/dataset_dict.py
```diff
@@ -897,30 +901,34 @@ def map(
...
-                    fn_kwargs=fn_kwargs,
-                    num_proc=num_proc,
-                    desc=desc,
-                )
-                for k, dataset in self.items()
-            }
-        )
+
+        if with_split and "split" in fn_kwargs:
+            raise ValueError("The key 'split' is reserved for the split name and cannot be passed in fn_kwargs.")
```
lhoestq, line 906: to avoid this issue we use positional arguments for indices/rank to avoid this issue (this way users can name their parameters arbitrarily), maybe we can use this trick: https://stackoverflow.com/a/66274908 ?

jp1924: @lhoestq 
Ah, I see what you mean.

So what you're saying is that there may be people who were already using fn_kwagrs to do splits, and my implementation may cause conflicts? I agree.

So let's use partial instead of fn_kwargs to solve the problem? Sounds like a good idea!

So I redefined partial and made it like this, can you take a look?

```python
class bind(partial):
    def __call__(self, *fn_args, **fn_kwargs):
        return self.func(*fn_args, *self.args, **fn_kwargs)
```

src/datasets/dataset_dict.py
```diff
@@ -795,7 +797,7 @@ def map(
         writer_batch_size: Optional[int] = 1000,
         features: Optional[Features] = None,
         disable_nullable: bool = False,
-        fn_kwargs: Optional[dict] = None,
+        fn_kwargs: dict = {},
```
lhoestq, line 800: I think you can revert this change and keep it as None by default :)

It's also good practice to avoid having lists or dicts as default parameters, since they can be modified inadvertently

src/datasets/dataset_dict.py
```diff
@@ -897,30 +902,38 @@ def map(
...
-                )
-                for k, dataset in self.items()
-            }
-        )
+
+        if with_split:
+
+            class bind(partial):
+                def __call__(self, *fn_args, **fn_kwargs):
+                    return self.func(*fn_args, *self.args, **fn_kwargs)
```
lhoestq, line 910: can you define this outside the DatasetDict class ? this way it can also be pickled

lhoestq (commented):
Great ! just added minor comments but LGTM

before we merge can you also ad the same for IterableDatasetDict ?

jp1924:
@lhoestq

jp1924:
@lhoestq please...

lhoestq (approved):
LGTM ! and sorry for the delay !

jp1924:
Thank you so much for reviewing this PR! Have a great weekend~

PR #7370, opened 2025-01-14
Support faster processing using pandas or polars functions in `IterableDataset.map()`

lhoestq:
Allow super fast processing using pandas or polars functions in `IterableDataset.map()` by adding support to pandas and polars formatting in `IterableDataset`

```python
import polars as pl
from datasets import Dataset

ds = Dataset.from_dict({"i": range(10)}).to_iterable_dataset()
ds = ds.with_format("polars")
ds = ds.map(lambda df: df.with_columns(pl.col("i"), pl.col("i").add(1).alias("i+1")), batched=True)
ds = ds.with_format(None)
print(next(iter(ds)))
# {'i': 0, 'i+1': 1}
```

It leverages arrow's zero-copy features from/to pandas and polars.

related to https://github.com/huggingface/datasets/issues/3444 #6762

### DIFF
--- a/docs/source/process.mdx
+++ b/docs/source/process.mdx
@@ -647,7 +647,7 @@ The [`~Dataset.with_format`] function also changes the format of a column, excep
 
 <Tip>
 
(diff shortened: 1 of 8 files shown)


lhoestq:
merging this and will make some docs and communications around using polars for optimizing data processing :)

PR #7434, opened 2025-03-04
Refactor Dataset.map to reuse cache files mapped with different num_proc

ringohoffman:
Fixes #7433

This refactor unifies num_proc is None or num_proc == 1 and num_proc > 1; instead of handling them completely separately where one uses a list of kwargs and shards and the other just uses a single set of kwargs and self, by wrapping the num_proc == 1 case in a list and making the difference just whether or not you use a pool, you set up either case to be able to load each other cache_files just by changing num_shards; num_proc == 1 can sequentially load the shards of a dataset mapped num_shards > 1 and sequentially map any missing shards

Other than the structural refactor, the main contribution of this PR is get_existing_cache_file_map, which uses a regex of cache_file_name and suffix_template to find existing cache files, grouped by their num_shards; using this data structure, we can reset num_shards to an existing set of cache files, and load them accordingly

PR #7435, opened 2025-03-04
Refactor `string_to_dict` to return `None` if there is no match instead of raising `ValueError`

ringohoffman:
Making this change, as encouraged here:

* https://github.com/huggingface/datasets/pull/7434#discussion_r1979933054

instead of having the pattern of using `try`-`except` to handle when there is no match, we can instead check if the return value is `None`; we can also assert that the return value should not be `None` if we know that should be true

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -3184,9 +3184,11 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                     del kwargs["shard"]
             else:
                 logger.info(f"Loading cached processed dataset at {format_cache_file_name(cache_file_name, '*')}")
-            assert None not in transformed_shards, (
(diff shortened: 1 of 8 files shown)


ringohoffman:
cc: @lhoestq

ringohoffman:
I am going to rebase #7434 onto this branch. Then we can merge this one first if you approve, and then #7434.

ringohoffman:
@lhoestq any thoughts here?

ringohoffman:
It looks like I was unsafely asserting that `source_url_fields is not None` in `image.py`, `video.py` and `audio.py` (which did not correspond to the `except ValueError` like was their previously). I've changed it to handle `source_url_fields is None`.

ringohoffman:
Can we re-run CI on this one?

ringohoffman:
Sweet! These failures are looking spurious due to connectivity issues. Can the failing run be retried?

ringohoffman:
@lhoestq Sorry to double ping, but can this PR be reviewed? I think it is ready!

PR #7437, opened 2025-03-06
Use pyupgrade --py39-plus for remaining files

cyyever:


cyyever:
@lhoestq Have a look?

PR #7478, opened 2025-03-26
update fsspec 2025.3.0

peteski22:
It appears there have been two releases of fsspec since this dependency was last updated, it would be great if Datasets could be updated so that it didn't hold back the usage of newer fsspec versions in consuming projects.

PR based on https://github.com/huggingface/datasets/pull/7352


lhoestq (approved):
thanks ! the CI failures are unrelated so we can merge :)

Issue #7373, opened 2025-01-16
Excessive RAM Usage After Dataset Concatenation concatenate_datasets


sam-hey:
### Describe the bug

When loading a dataset from disk, concatenating it, and starting the training process, the RAM usage progressively increases until the kernel terminates the process due to excessive memory consumption.

https://github.com/huggingface/datasets/issues/2276

### Steps to reproduce the bug

```python
from datasets import  DatasetDict, concatenate_datasets

dataset = DatasetDict.load_from_disk("data")

...
...

combined_dataset = concatenate_datasets(
        [dataset[split] for split in dataset]
    )

#start SentenceTransformer training
```

### Expected behavior

I would not expect RAM utilization to increase after concatenation. Removing the concatenation step resolves the issue

### Environment info

sentence-transformers==3.1.1
datasets==3.2.0

python3.10

sam-hey:
![Image](https://github.com/user-attachments/assets/b6f8bcbd-44af-413e-bc06-65380eb0f746)

Adding a img from memray

nepfaff:
I'm having the same issue where concatenation seems to use a huge amount of RAM.

```python
# Load all chunks and concatenate them into a final dataset.
        chunk_datasets = [
            Dataset.load_from_disk(file, keep_in_memory=False)
            for file in tqdm(chunk_files, desc="Loading chunk datasets")
        ]
        logging.info("Concatenating chunk datasets...")
        final_dataset = concatenate_datasets(chunk_datasets)
```

This is a real issue for me as the final dataset is a few terabytes in size. I'm using datasets version `3.1.0`

sam-hey:
I did have a short look, the error seems to be from `memory_map` not being closed. 

https://github.com/huggingface/datasets/blob/5f8d2ad9a1b0bccfd962d998987228addfd5be9f/src/datasets/table.py#L48-L50

PR #7489, opened 2025-03-29
fix: loading of datasets from Disk(#7373)

sam-hey:
Fixes dataset loading from disk by ensuring that memory maps and streams are properly closed. 

For more details, see https://github.com/huggingface/datasets/issues/7373.

### DIFF
--- a/src/datasets/table.py
+++ b/src/datasets/table.py
@@ -47,7 +47,7 @@ def _in_memory_arrow_table_from_buffer(buffer: pa.Buffer) -> pa.Table:
 
 def _memory_mapped_record_batch_reader_from_file(filename: str) -> pa.RecordBatchStreamReader:
     memory_mapped_stream = pa.memory_map(filename)
-    return pa.ipc.open_stream(memory_mapped_stream)
+    return pa.ipc.open_stream(memory_mapped_stream), memory_mapped_stream
 
 
 def read_schema_from_file(filename: str) -> pa.Schema:
@@ -61,8 +61,9 @@ def read_schema_from_file(filename: str) -> pa.Schema:
 
 
 def _memory_mapped_arrow_table_from_file(filename: str) -> pa.Table:
-    opened_stream = _memory_mapped_record_batch_reader_from_file(filename)
+    opened_stream, memory_mapped_stream = _memory_mapped_record_batch_reader_from_file(filename)
     pa_table = opened_stream.read_all()
+    memory_mapped_stream.close()
     return pa_table
 
 
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -5191,8 +5191,13 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
 
     @staticmethod
     def _generate_tables_from_cache_file(filename: str):
-        for batch_idx, batch in enumerate(_memory_mapped_record_batch_reader_from_file(filename)):
-            yield batch_idx, pa.Table.from_batches([batch])
+        reader, mmap_stream = _memory_mapped_record_batch_reader_from_file(filename)
+        try:
+            for batch_idx, batch in enumerate(reader):
+                yield batch_idx, pa.Table.from_batches([batch])
+        finally:
+            reader.close()
+            mmap_stream.close()
 
     def to_iterable_dataset(self, num_shards: Optional[int] = 1) -> "IterableDataset":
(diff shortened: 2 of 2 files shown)


sam-hey:
@nepfaff Could you confirm if this fixes the issue for you? I checked Memray, and everything looked good on my end.

nepfaff:
Will aim to get to this soon. I don't have a rapid testing pipeline setup but need to wait for some AWS nodes to become free

nepfaff:
I now set up a small experiment:

```python
# Log initial RAM usage
    process = psutil.Process(os.getpid())
    initial_ram = process.memory_info().rss / (1024 * 1024)  # Convert to MB
    logging.info(f"Initial RAM usage: {initial_ram:.2f} MB")

    chunk_datasets = [
        Dataset.load_from_disk(dataset_path, keep_in_memory=False) for _ in range(N)
    ]
    combined_dataset = concatenate_datasets(chunk_datasets)
    print(len(combined_dataset))

    # Log final RAM usage
    final_ram = process.memory_info().rss / (1024 * 1024)  # Convert to MB
    ram_diff = final_ram - initial_ram

    logging.info(f"Final RAM usage: {final_ram:.2f} MB")
    logging.info(f"RAM usage increase: {ram_diff:.2f} MB")
```

The RAM usage is linearly correlated with `N` on datasets master!

For my dataset:
- N=5 => RAM usage increase: 26302.91 MB
- N=10 => RAM usage increase: 52315.18 MB
- N=20 => RAM usage increase: 104510.65 MB

Unfortunately, your patch doesn't seem to change this:
```bash
pip install git+https://github.com/sam-hey/datasets.git@fix/concatenate_datasets
pip list | grep datasets
datasets                 3.5.1.dev0
```
Gives exactly the same RAM statistics.

sam-hey:
Thanks a lot, @nepfaff, for taking a look at this! It seems that `concatenate_datasets()` is fixed with this PR. I can also confirm that loading a large number of files requires significant memory. However, as I understand it, this is expected/a bug since the memory consumption stems from `pa.memory_map()`, which returns a memory-mapped file.

This behavior might be related to this bug: https://github.com/apache/arrow/issues/34423

src/datasets/table.py
```diff
@@ -61,8 +86,10 @@ def read_schema_from_file(filename: str) -> pa.Schema:
 
 
 def _memory_mapped_arrow_table_from_file(filename: str) -> pa.Table:
-    opened_stream = _memory_mapped_record_batch_reader_from_file(filename)
+    opened_stream, memory_mapped_stream = _memory_mapped_record_batch_reader_from_file(filename)
     pa_table = opened_stream.read_all()
+    opened_stream.close()
+    memory_mapped_stream.close()
     return pa_table
```
lhoestq, line 93: is it really ok to close the memory map, given the memory mapped table is still in use ?

lhoestq: I think `read_all()` doesn't load the data in memory here, it loads buffers that are memory mapped from disk

lasuomela: I can confirm that this works:

```
import pyarrow as pa

BATCH_SIZE = 10000
NUM_BATCHES = 1000

schema = pa.schema([pa.field('nums', pa.int32())])

# Write in stream format
with pa.OSFile('bigfile.arrow', 'wb') as sink:
    with pa.ipc.new_stream(sink, schema) as writer:
        for _ in range(NUM_BATCHES):
            batch = pa.record_batch([pa.array(range(BATCH_SIZE), type=pa.int32())], schema)
            writer.write(batch)

# Read the stream back. The memory map is closed after read_all()
with pa.memory_map('bigfile.arrow', 'rb') as source:
    loaded_stream = pa.ipc.open_stream(source)
    table = loaded_stream.read_all()

print("LEN:", table.num_rows)
print("RSS: {}MB".format(pa.total_allocated_bytes() >> 20))

# Read the first batch. Memory consumption is 0, but we can still read the data.
print("")
print("First batch:")
print(table[0][0:BATCH_SIZE])
```

Out:
```
LEN: 10000000
RSS: 0MB

First batch:
[
  [
    0,
    1,
    2,
    3,
    4,
    ...
    9995,
    9996,
    9997,
    9998,
    9999
  ]
]
```

sam-hey: @lasuomela Thanks a lot for checking! I’m currently on vacation and without a laptop to verify it myself.
@lhoestq Would this be sufficient proof for you?

lhoestq:
Great ! have you tested that it indeed fixes the memory issue ?

PR #7525, opened 2025-04-17
Fix indexing in split commit messages

klamike:
When a large commit is split up, it seems the commit index in the message is zero-based while the total number is one-based. I came across this running `convert-to-parquet` and was wondering why there was no `6-of-6` commit. This PR fixes that by adding one to the commit index, so both are one-based.

Current behavior:
Screenshot 2025-04-17 at 1 00 17 PM


lhoestq:
Hi ! this is expected and is coherent with other naming conventions in `datasets` such as parquet shards naming

Issue #7513, opened 2025-04-15
MemoryError while creating dataset from generator


simonreise:
### Describe the bug

# TL:DR

`Dataset.from_generator` function passes all of its arguments to `BuilderConfig.create_config_id`, including `generator` function itself. `BuilderConfig.create_config_id` function tries to hash all the args, which can take a large amount of time or even cause MemoryError if the dataset processed in a generator function is large enough.

Maybe we should pop `generator` from `config_kwargs_to_add_to_suffix` before hashing to avoid it.

# Full description

I have a pretty large spatial imagery dataset that is generated from two xbatcher.BatchGenerators via custom `dataset_generator` function that looks like this if simplified:
```
def dataset_generator():
    for index in samples:
        data_dict = {
            "key": index,
            "x": x_batches[index].data,
            "y":  y_batches[index].data,
        }
        yield data_dict
```
Then I use `datasets.Dataset.from_generator` to generate the dataset itself.
```
# Create dataset
ds = datasets.Dataset.from_generator(
    dataset_generator,
    features=feat,
    cache_dir=(output / ".cache"),
)
```

It works nicely with pretty small data, but if the dataset is huge and barely fits in memory, it crashes with memory error:
<details>
<summary>Full stack trace</summary>

```
File [C:\ProgramData\miniforge3\envs\geo\Lib\site-packages\remote_sensing_processor\segmentation\semantic\tiles.py:248](file:///C:/ProgramData/miniforge3/envs/geo/Lib/site-packages/remote_sensing_processor/segmentation/semantic/tiles.py#line=247), in generate_tiles(x, y, output, tile_size, shuffle, split, x_dtype, y_dtype, x_nodata, y_nodata)
    245         yield data_dict
    247 # Create dataset
--> 248 ds = datasets.Dataset.from_generator(
    249     dataset_generator,
    250     features=feat,
    251     cache_dir=(output / ".cache"),
    252 )
    254 # Save dataset
    255 ds.save_to_disk(output / name)

File [C:\ProgramData\miniforge3\envs\geo\Lib\site-packages\datasets\arrow_dataset.py:1105](file:///C:/ProgramData/miniforge3/envs/geo/Lib/site-packages/datasets/arrow_dataset.py#line=1104), in Dataset.from_generator(generator, features, cache_dir, keep_in_memory, gen_kwargs, num_proc, split, **kwargs)
   1052 """Create a Dataset from a generator.
   1053 
   1054 Args:
   (...)   1101 ```
[...]

PR #7533, opened 2025-04-23
Add custom fingerprint support to `from_generator`

simonreise:
This PR adds `dataset_id_suffix` parameter to 'Dataset.from_generator' function.

`Dataset.from_generator` function passes all of its arguments to `BuilderConfig.create_config_id`, including generator function itself. `BuilderConfig.create_config_id` function tries to hash all the args, which can take a large amount of time or even cause MemoryError if the dataset processed in a generator function is large enough.

This PR allows user to pass a custom fingerprint (`dataset_id_suffix`) to be used as a suffix in a dataset name instead of the one generated by hashing the args.

This PR is a possible solution of #7513


lhoestq:
This is great !

What do you think of passing `config_id=` directly to the builder instead of just the suffix ? This would be a power user argument though, or for internal use. And in from_generator the new argument can be `fingerprint=` as in `Dataset.__init__()`

The `config_id` can be defined using something like `config_id = "default-fingerprint=" + fingerprint`

I feel ike this could make the Dataset API more coherent if we avoid introducing a new argument while we can juste use `fingerprint=`

src/datasets/builder.py
```diff
@@ -141,6 +141,7 @@ def create_config_id(
         self,
         config_kwargs: dict,
         custom_features: Optional[Features] = None,
+        fingerprint: Optional[str] = None,
```
lhoestq, line 144: no need to modify create_config_id() imo, you can just use the user-defined config_id in the `__init__`

and in src/datasets/io/generator.py you can pass `Generator(..., config_id=fingerpring)`

lhoestq (commented):
added more comments

src/datasets/packaged_modules/generator/generator.py
```diff
@@ -10,6 +10,7 @@ class GeneratorConfig(datasets.BuilderConfig):
     gen_kwargs: Optional[dict] = None
     features: Optional[datasets.Features] = None
     split: datasets.NamedSplit = datasets.Split.TRAIN
+    fingerprint: Optional[str] = None
```
lhoestq, line 13: no need for this line imo

simonreise: Removed it.

ArjunJagdale:
I looked into this issue and the original cause makes total sense — hashing a large generator is clearly inefficient and fragile for big datasets.

PR #7533 looks like a robust and flexible solution! It cleanly separates the fingerprinting responsibility by letting users pass `fingerprint=` (now `config_id=`), which avoids hashing heavy objects like generators but still preserves caching logic.

simonreise:
@lhoestq could you please re-review the changes I made?

simonreise:
@lhoestq ping
I also added a simple test for the `fingerprint` parameter

simonreise:
@lhoestq could you please review the PR? I implemented the requested changes and added a test for the added `fingerprint` arg

src/datasets/io/generator.py
```diff
@@ -32,6 +33,7 @@ def __init__(
             generator=generator,
             gen_kwargs=gen_kwargs,
             split=split,
+            config_id=fingerprint,
```
lhoestq:
```suggestion
            config_id="default-fingerprint=" + fingerprint,
```

lhoestq (approved):
lgtm :) I just did some minor edits

Issue #7475, opened 2025-03-25
IterableDataset's state_dict shard_example_idx is always equal to the number of samples in a shard


bruno-hays:
### Describe the bug

I've noticed a strange behaviour with Iterable state_dict: the value of shard_example_idx is always equal to the amount of samples in a shard.

### Steps to reproduce the bug

I am reusing the example from the doc
```python
from datasets import Dataset

ds = Dataset.from_dict({"a": range(6)}).to_iterable_dataset(num_shards=1)
state_dict = None
# Iterate through the dataset and print examples
for idx, example in enumerate(ds):
    print(example)
    if idx == 2:
        state_dict = ds.state_dict()
        print("checkpoint")
        break
print(state_dict)
```
Returns:
```
{'a': 0}
{'a': 1}
checkpoint
{'examples_iterable': {'shard_idx': 0, 'shard_example_idx': 6, 'type': 'ArrowExamplesIterable'}, 'epoch': 0}
```

### Expected behavior

shard_example_idx should be 2 instead of 6
If we run with num_shards=2, then shard_example_idx is 3 instead of 2 and so on.

### Environment info

- `datasets` version: 3.4.1
- Platform: macOS-14.6.1-arm64-arm-64bit
- Python version: 3.12.9
- `huggingface_hub` version: 0.29.3
- PyArrow version: 19.0.1
- Pandas version: 2.2.3
- `fsspec` version: 2024.12.0

Harry-Yang0518:
Hey, I’d love to work on this issue but I am a beginner, can I work it with you?

bruno-hays:
Hello. I'm sorry but I don't have much time to get in the details for now.
Have you managed to reproduce the issue with the code provided ?
If you want to work on it, you can self-assign and ask @lhoestq for directions

Harry-Yang0518:
Hi Bruno, I am trying to reproduce it this later in this week and let you know what I found.

Harry-Yang0518:
#self_assign

Harry-Yang0518:
#self-assign

lhoestq:
Good catch, I tried and if the dataset is bigger (e.g. `range(9999)`) it returns `1000` somehow

Harry-Yang0518:
Hello @lhoestq,

[...]

PR #7539, opened 2025-04-27
Fix IterableDataset state_dict shard_example_idx counting

Harry-Yang0518:
# Fix IterableDataset's state_dict shard_example_idx reporting

## Description

This PR fixes issue #7475 where the `shard_example_idx` value in `IterableDataset`'s `state_dict()` always equals the number of samples in a shard, even if only a few examples have been consumed.

The issue is in the `_iter_arrow` method of the `ArrowExamplesIterable` class where it updates the `shard_example_idx` state by the full length of the batch (`len(pa_table)`) even when we're only partway through processing the examples. 

## Changes

Modified the `_iter_arrow` method of `ArrowExamplesIterable` to:
1. Track the actual number of examples processed
2. Only increment the `shard_example_idx` by the number of examples actually yielded
3. Handle partial batches correctly

## How to Test

I've included a simple test case that demonstrates the fix:

```python
from datasets import Dataset

# Create a test dataset
ds = Dataset.from_dict({"a": range(6)}).to_iterable_dataset(num_shards=1)

# Iterate through part of the dataset
for idx, example in enumerate(ds):
    print(example)
    if idx == 2:  # Stop after 3 examples (0, 1, 2)
        state_dict = ds.state_dict()
        print("Checkpoint state_dict:", state_dict)
        break

# Before the fix, the output would show shard_example_idx: 6
# After the fix, it shows shard_example_idx: 3, correctly reflecting the 3 processed examples
```

## Implementation Details

1. Added logic to track the number of examples actually seen in the current shard
2. Modified the state update to only count examples actually yielded
3. Improved handling of partial batches and skipped examples

This fix ensures that checkpointing and resuming works correctly with exactly the expected number of examples, rather than skipping ahead to the end of the batch.

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -317,16 +317,40 @@ class ArrowExamplesIterable(_BaseExamplesIterable):
 
     def _iter_arrow(self):
         shard_idx_start = self._state_dict["shard_idx"] if self._state_dict else 0
-        for gen_kwags in islice(_split_gen_kwargs(self.kwargs, max_num_jobs=self.num_shards), shard_idx_start, None):
+        kwargs_with_shuffled_shards = (
+            _shuffle_gen_kwargs(self.generator, self.kwargs) if hasattr(self, "generator") else self.kwargs
+        )
+
+        for gen_kwags in islice(
+            _split_gen_kwargs(kwargs_with_shuffled_shards, max_num_jobs=self.num_shards), shard_idx_start, None
+        ):
             shard_example_idx_start = self._state_dict["shard_example_idx"] if self._state_dict else 0
             shard_example_idx = 0
+
+            examples_seen_in_current_shard = 0
+
             for key, pa_table in self.generate_tables_fn(**gen_kwags):
-                shard_example_idx += len(pa_table)
-                if shard_example_idx <= shard_example_idx_start:
+                batch_size = len(pa_table)
+
+                if shard_example_idx + batch_size <= shard_example_idx_start:
+                    shard_example_idx += batch_size
                     continue
+
+                if shard_example_idx < shard_example_idx_start:
+                    offset = shard_example_idx_start - shard_example_idx
+                    pa_table = pa_table.slice(offset)
+                    examples_seen_in_current_shard = offset
+
                 if self._state_dict:
-                    self._state_dict["shard_example_idx"] += len(pa_table)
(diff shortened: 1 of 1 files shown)


src/datasets/iterable_dataset.py
```diff
@@ -317,16 +317,40 @@ def __iter__(self):
 
     def _iter_arrow(self):
         shard_idx_start = self._state_dict["shard_idx"] if self._state_dict else 0
-        for gen_kwags in islice(_split_gen_kwargs(self.kwargs, max_num_jobs=self.num_shards), shard_idx_start, None):
+        kwargs_with_shuffled_shards = (
+            _shuffle_gen_kwargs(self.generator, self.kwargs) if hasattr(self, "generator") else self.kwargs
```
lhoestq, line 321: is this expected ? I think ShuffledDataSourcesArrowExamplesIterable has its own `_iter_arrow()` implementation

src/datasets/iterable_dataset.py
```diff
@@ -317,16 +317,40 @@ def __iter__(self):
...
+        )
+
+        for gen_kwags in islice(
+            _split_gen_kwargs(kwargs_with_shuffled_shards, max_num_jobs=self.num_shards), shard_idx_start, None
+        ):
             shard_example_idx_start = self._state_dict["shard_example_idx"] if self._state_dict else 0
             shard_example_idx = 0
+
+            examples_seen_in_current_shard = 0
```
lhoestq, line 330: how is it different from shard_example_idx ?

src/datasets/iterable_dataset.py
```diff
@@ -317,16 +317,40 @@ def __iter__(self):
...
+                batch_size = len(pa_table)
+
+                if shard_example_idx + batch_size <= shard_example_idx_start:
+                    shard_example_idx += batch_size
                     continue
+
+                if shard_example_idx < shard_example_idx_start:
+                    offset = shard_example_idx_start - shard_example_idx
+                    pa_table = pa_table.slice(offset)
+                    examples_seen_in_current_shard = offset
```
lhoestq, line 342: is this needed ? we always yield full tables, so it's unlikely we end up with a `shard_example_idx` that doesn't land exactly on a table boundary (except if the dataset state is manually crafted maybe)

lhoestq (commented):
Cool ! I left some comments :)

Feel free to also update your branch to include CI fixes from `main`

Issue #7536, opened 2025-04-24
[Errno 13] Permission denied: on `.incomplete` file


ryan-clancy:
### Describe the bug

When downloading a dataset, we frequently hit the below Permission Denied error. This looks to happen (at least) across datasets in HF, S3, and GCS.

It looks like the `temp_file` being passed [here](https://github.com/huggingface/datasets/blob/main/src/datasets/utils/file_utils.py#L412) can sometimes be created with `000` permissions leading to the permission denied error (the user running the code is still the owner of the file). Deleting that particular file and re-running the code with 0 changes will usually succeed.

Is there some race condition happening with the [umask](https://github.com/huggingface/datasets/blob/main/src/datasets/utils/file_utils.py#L416), which is process global, and the [file creation](https://github.com/huggingface/datasets/blob/main/src/datasets/utils/file_utils.py#L404)?

```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.venv/lib/python3.12/site-packages/datasets/load.py:2084: in load_dataset
    builder_instance.download_and_prepare(
.venv/lib/python3.12/site-packages/datasets/builder.py:925: in download_and_prepare
    self._download_and_prepare(
.venv/lib/python3.12/site-packages/datasets/builder.py:1649: in _download_and_prepare
    super()._download_and_prepare(
.venv/lib/python3.12/site-packages/datasets/builder.py:979: in _download_and_prepare
    split_generators = self._split_generators(dl_manager, **split_generators_kwargs)
.venv/lib/python3.12/site-packages/datasets/packaged_modules/folder_based_builder/folder_based_builder.py:120: in _split_generators
    downloaded_files = dl_manager.download(files)
.venv/lib/python3.12/site-packages/datasets/download/download_manager.py:159: in download
    downloaded_path_or_paths = map_nested(
.venv/lib/python3.12/site-packages/datasets/utils/py_utils.py:514: in map_nested
    _single_map_nested((function, obj, batched, batch_size, types, None, True, None))
.venv/lib/python3.12/site-packages/datasets/utils/py_utils.py:382: in _single_map_nested
    return [mapped_item for batch in iter_batched(data_struct, batch_size) for mapped_item in function(batch)]
.venv/lib/python3.12/site-packages/datasets/download/download_manager.py:206: in _download_batched
    return thread_map(
[...]

Issue #5799, opened 2023-04-27
Files downloaded to cache do not respect umask


albertvillanova:
As reported by @stas00, files downloaded to the cache do not respect umask:
```bash
$ ls -l /path/to/cache/datasets/downloads/
-rw------- 1 uername    username    150M Apr 25 16:41 5e646c1d600f065adaeb134e536f6f2f296a6d804bd1f0e1fdcd20ee28c185c6 
```

Related to:
- #2065

PR #7547, opened 2025-05-01
Avoid global umask for setting file mode.

ryan-clancy:
This PR updates the method for setting the permissions on `cache_path` after calling `shutil.move`. The call to `shutil.move` may not preserve permissions if the source and destination are on different filesystems. Reading and resetting umask can cause race conditions, so directly read what permissions were set for the `temp_file` instead.

This fixes https://github.com/huggingface/datasets/issues/7536.

### DIFF
--- a/src/datasets/utils/file_utils.py
+++ b/src/datasets/utils/file_utils.py
@@ -13,6 +13,7 @@ import os
 import posixpath
 import re
 import shutil
+import stat
 import sys
 import tarfile
 import time
@@ -411,11 +412,14 @@ def get_from_cache(
             # GET file object
             fsspec_get(url, temp_file, storage_options=storage_options, desc=download_desc, disable_tqdm=disable_tqdm)
 
+        # get the permissions of the temp file
+        temp_file_mode = stat.S_IMODE(os.stat(temp_file.name).st_mode)
+
         logger.info(f"storing {url} in cache at {cache_path}")
         shutil.move(temp_file.name, cache_path)
-        umask = os.umask(0o666)
-        os.umask(umask)
-        os.chmod(cache_path, 0o666 & ~umask)
+
+        # make sure make sure permissions on cache_path are the same as temp_file, shutil.move may not preserve them
+        os.chmod(cache_path, temp_file_mode)
 
         logger.info(f"creating metadata file for {cache_path}")
         meta = {"url": url, "etag": etag}


src/datasets/utils/file_utils.py
```diff
@@ -411,11 +412,14 @@ def temp_file_manager(mode="w+b"):
             # GET file object
             fsspec_get(url, temp_file, storage_options=storage_options, desc=download_desc, disable_tqdm=disable_tqdm)
 
+        # get the permissions of the temp file
+        temp_file_mode = stat.S_IMODE(os.stat(temp_file.name).st_mode)
```
lhoestq, line 416: this doesn't necessarily return the user's umask no ?

we need the user's umask since some people use `datasets` with shared disks and use group permissions (see #2157 #5799)

ryan-clancy: When looking more at the history of this code, I am not sure the umask block below is needed at all anymore and could just be deleted.

https://github.com/huggingface/datasets/blob/f5f6c7ecad3cbd05c2fdf954491af57b78df5ca5/src/datasets/utils/file_utils.py#L582-L626 is the code just after the umask was added and `NamedTemporaryFile` was used to create the files. `NamedTemporaryFile` always set 600 permissions and didn't respect umask so umask code was added to fix the permissions.

https://github.com/huggingface/datasets/commit/d536e37b21a6dd5c122b6d8113994ec50846c5b5 removed the `NamedTemporaryFile` and moved to `open()`, which respects the umask.

We should just be able to remove this `temp_file_mode` + updated chmod lines and then remove:
```
umask = os.umask(0o666)
os.umask(umask)
os.chmod(cache_path, 0o666 & ~umask)
```

Thoughts?

ryan-clancy: @lhoestq - I updated the PR to remove this outdated code, seems to work as expected and umask is respected (see output below from looking at files produced from `tests/test_file_utils.py`).

If this looks good to you, can we get this merged for `3.5.2`?

Default umask:
```
$ ls -la | grep a6ecb420baa10e46bf84ed0960dee4e9093fee1aa58dc0
-rw-r--r--@ 1 ryan  staff   39 May  5 13:14 a6ecb420baa10e46bf84ed0960dee4e9093fee1aa58dc0035db883c7050ae847
-rw-r--r--@ 1 ryan  staff   35 May  5 13:14 a6ecb420baa10e46bf84ed0960dee4e9093fee1aa58dc0035db883c7050ae847.json
-rw-r--r--@ 1 ryan  staff    0 May  5 13:14 a6ecb420baa10e46bf84ed0960dee4e9093fee1aa58dc0035db883c7050ae847.lock
```

Modified to `umask 002`:
```
$ ls -la | grep a6ecb420baa10e46bf84ed0960dee4e9093fee1aa58dc0
-rw-rw-r--@ 1 ryan  staff   39 May  5 13:15 a6ecb420baa10e46bf84ed0960dee4e9093fee1aa58dc0035db883c7050ae847
-rw-rw-r--@ 1 ryan  staff   35 May  5 13:15 a6ecb420baa10e46bf84ed0960dee4e9093fee1aa58dc0035db883c7050ae847.json
-rw-rw-r--@ 1 ryan  staff    0 May  5 13:15 a6ecb420baa10e46bf84ed0960dee4e9093fee1aa58dc0035db883c7050ae847.lock
```

lhoestq (approved):
Great !

PR #7553, opened 2025-05-06
Rebatch arrow iterables before formatted iterable

lhoestq:
close https://github.com/huggingface/datasets/issues/7538 and https://github.com/huggingface/datasets/issues/7475

PR #7557, opened 2025-05-07
check for empty _formatting

winglian:
Fixes a regression from #7553 breaking shuffling of iterable datasets 
Screenshot 2025-05-07 at 9 16 52 AM


lhoestq:
Thanks for reporting and for the fix ! I tried to reorganize the condition in your PR but didn't get the right permission so. I ended up merging https://github.com/huggingface/datasets/pull/7558 directly so I can make a release today - I hope you don't mind

PR #7565, opened 2025-05-09
add check if repo exists for dataset uploading

Samoed:
Currently, I'm uploading datasets for `MTEB`. Some of them have a lot of splits (more than 20), and I'm encountering the error:`Too many requests for https://huggingface.co/datasets/repo/create`.

It seems this issue occurs because each time a split is uploaded, the dataset tries to recreate itself. I've added a check to ensure that if the dataset already exists, it won't attempt to recreate it.


Samoed:
@lhoestq Can you review, please? I don't think that errors in CI are related to my changes

Samoed:
Similar implementation done in https://github.com/huggingface/datasets/pull/7708

PR #7576, opened 2025-05-21
Fix regex library warnings

emmanuel-ferdman:
# PR Summary
This small PR resolves the regex library warnings showing starting Python3.11:
```python
DeprecationWarning: 'count' is passed as positional argument
```


emmanuel-ferdman:
@lhoestq could you please take a look into this small PR when you get the chance? thanks 🙌

PR #7609, opened 2025-06-12
Update `_dill.py` to use `co_linetable` for Python 3.10+ in place of `co_lnotab`

qgallouedec:
Not 100% about this one, but it seems to be recommended.

```
/fsx/qgallouedec/miniconda3/envs/trl/lib/python3.12/site-packages/datasets/utils/_dill.py:385: DeprecationWarning: co_lnotab is deprecated, use co_lines instead.
```
Tests pass locally. And the warning is gone with this change.

https://peps.python.org/pep-0626/#backwards-compatibility

### DIFF
--- a/src/datasets/utils/_dill.py
+++ b/src/datasets/utils/_dill.py
@@ -359,7 +359,7 @@ elif config.DILL_VERSION.release[:3] in [
 
         if hasattr(obj, "co_endlinetable"):  # python 3.11a (20 args)
             args = (
-                obj.co_linetable,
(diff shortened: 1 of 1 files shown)


lhoestq:
not 100% sure either, I tried removing unnecessary checks - let me know if they sound good to you otherwise I'll revert

qgallouedec:
I can't reproduce the warning anymore... 🤦🏻‍♂️

qgallouedec:
Ah now I can reproduce!, and I can confirm that the warning is gone when you apply the change in this PR

Issue #6046, opened 2023-07-17
Support proxy and user-agent in fsspec calls


lhoestq:
Since we switched to the new HfFileSystem we no longer apply user's proxy and user-agent.

Using the HTTP_PROXY and HTTPS_PROXY environment variables works though since we use aiohttp to call the HF Hub.

lhoestq:
You can reply "#self-assign" to this issue to automatically get assigned to it :)
Let me know if you have any questions or if I can help

zutarich:
#2289

zutarich:
Actually i am quite new to figure it out how everything goes and done 

> You can reply "#self-assign" to this issue to automatically get assigned to it :)
> Let me know if you have any questions or if I can help

when i wrote #self-assign it automatically got converted to some number is it correct or i have done it some wrong way, I am quite new to open source thus wanna try to learn and explore it

zutarich:
#2289 #self-assign

lhoestq:
Ah yea github tries to replace the #self-assign with an issue link. I guess you can try to copy-paste instead to see if it works

zutarich:
thanks a lot @lhoestq ! though i have a very lil idea of the issue, i am new. as i said before,  but gonnna try my best shot to do it.
can you please suggest some tips or anything from your sid ehow basically we approach it will be really helpfulll.
Will try my best!

lhoestq:
The HfFileSystem from the `huggingface_hub` package can already read the HTTP_PROXY and HTTPS_PROXY environment variables. So the remaining thing missing is the `user_agent` that the user may include in a `DownloadConfig` object. The user agent can be used for regular http calls but also calls to the HfFileSystem.

- for http, the `user_agent` isn't passed from `DownloadConfig` to `get_datasets_user_agent` in `_prepare_single_hop_path_and_storage_options` in `streaming_download_manager.py` so we need to include it
- for HfFileSystem I think it requires a PR in https://github.com/huggingface/huggingface_hub to include it in the `HfFileSystem.__init__`

ArjunJagdale:
Hi @lhoestq 👋🏼

Is anyone currently working on this? If not, I'd like to pick it up.

As I understand it:
- The `user_agent` from `DownloadConfig` isn't currently passed to `get_datasets_user_agent()` inside `_prepare_single_hop_path_and_storage_options`.
- I'll update that function to include the correct `user-agent` in the `headers`.
[...]

PR #7631, opened 2025-06-21
Pass user-agent from DownloadConfig into fsspec storage_options

ArjunJagdale:
Fixes part of issue #6046

### Problem
The `user-agent` defined in `DownloadConfig` was not passed down to fsspec-based filesystems like `HfFileSystem`, which prevents proper identification/tracking of client requests.

### Solution
Added support for injecting the `user-agent` into `storage_options["headers"]` within `_prepare_single_hop_path_and_storage_options()` based on the `protocol`.

Now, when using `hf://`, `http://`, or `https://`, the custom user-agent is passed automatically.

### Code Location
Modified:
- `src/datasets/utils/file_utils.py`

Used `get_datasets_user_agent(...)` to ensure proper formatting and fallback logic.

### DIFF
--- a/src/datasets/utils/file_utils.py
+++ b/src/datasets/utils/file_utils.py
@@ -30,6 +30,8 @@ import fsspec
 import huggingface_hub
 import huggingface_hub.errors
 import requests
+from .file_utils import get_datasets_user_agent  # If needed
+from ..download.download_config import DownloadConfig
 from fsspec.core import strip_protocol, url_to_fs
 from fsspec.utils import can_be_local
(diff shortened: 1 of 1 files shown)


ArjunJagdale:
- This PR assumes that `HfFileSystem` in `huggingface_hub` supports receiving `headers` in `storage_options`. If not, a follow-up PR can be opened to add this support to `HfFileSystem.__init__`.
- No test was added for this since it’s a config passthrough. If needed, I’d be happy to add one.

Issue #7632, opened 2025-06-23
Graceful Error Handling for cast_column("image", Image(decode=True)) in Hugging Face Datasets


ganiket19:
### Feature request

Currently, when using dataset.cast_column("image", Image(decode=True)), the pipeline throws an error and halts if any image in the dataset is invalid or corrupted (e.g., truncated files, incorrect formats, unreachable URLs). This behavior disrupts large-scale processing where a few faulty samples are common.
reference : https://discuss.huggingface.co/t/handle-errors-when-loading-images-404-corrupted-etc/50318/5
                  https://discuss.huggingface.co/t/handling-non-existing-url-in-image-dataset-while-cast-column/69185
                  
Proposed Feature

Introduce a mechanism (e.g., a continue_on_error=True flag or global error handling mode) in Image(decode=True) that:

Skips invalid images and sets them as None, or

Logs the error but allows the rest of the dataset to be processed without interruption.

Example Usage

from datasets import load_dataset, Image

dataset = load_dataset("my_dataset")
dataset = dataset.cast_column("image", Image(decode=True, continue_on_error=True))

Benefits

Ensures robust large-scale image dataset processing.

Improves developer productivity by avoiding custom retry/error-handling code.

Aligns with best practices in dataset preprocessing pipelines that tolerate minor data corruption.

Potential Implementation Options

Internally wrap the decoding in a try/except block.

Return None or a placeholder on failure.

Optionally allow custom error callbacks or logging.

### Motivation

Robustness: Large-scale image datasets often contain a small fraction of corrupt files or unreachable URLs. Halting on the first error forces users to write custom workarounds or preprocess externally.

Simplicity: A built-in flag removes boilerplate try/except logic around every decode step.

Performance: Skipping invalid samples inline is more efficient than a two-pass approach (filter then decode).

### Your contribution

1. API Change
Extend datasets.features.Image(decode=True) to accept continue_on_error: bool = False.

2. Behavior
If continue_on_error=False (default), maintain current behavior: any decode error raises an exception.
If continue_on_error=True, wrap decode logic in try/except:
On success: store the decoded image.
On failure: log a warning (e.g., via logging.warning) and set the field to None (or a sentinel value).

3. Optional Enhancements
Allow a callback hook:
Image(decode=True, continue_on_error=True, on_error=lambda idx, url, exc: ...)
Emit metrics or counts of skipped images.

ArjunJagdale:
Hi! This is now handled in PR #7638

[...]

PR #7638, opened 2025-06-24
Add ignore_decode_errors option to Image feature for robust decoding #7612

### LINKED ISSUE #7612: Provide an option of robust dataset iterator with error handling
### Feature request

Adding an option to skip corrupted data samples when __iter__ is called. Currently the datasets behavior is throwing errors if the data sample if corrupted and let user aware and handle the data corruption. When I tried to try-catch the error at user level, the iterator will raise StopIteration when I called next() again. 

The way I try to do error handling is: (This doesn't work, unfortunately)

```
        # Load the dataset with streaming enabled
        dataset = load_dataset(
            "pixparse/cc12m-wds", split="train", streaming=True
        )
        # Get an iterator from the dataset
        iterator = iter(dataset)

        while True:
            try:
                # Try to get the next example
                example = next(iterator)
                
                # Try to access and process the image
                image = example["jpg"]
                pil_image = Image.fromarray(np.array(image))
                pil_image.verify()  # Verify it's a valid image file

            except StopIteration:  # Code path 1
                print("\nStopIteration was raised! Reach the end of dataset")
                raise StopIteration

            except Exception as e:  # Code path 2
                errors += 1
                print("Error! Skip this sample")
                cotinue
            else:
                successful += 1
```

This is because the `IterableDataset` already throws an error (reaches Code path 2). And if I continue call next(), it will hit Code path 1 because the inner iterator of `IterableDataset`([code](https://github.com/huggingface/datasets/blob/89bd1f971402acb62805ef110bc1059c38b1c8c6/src/datasets/iterable_dataset.py#L2242)) as been stopped, so I can not skip the corrupted data sample. Would also love to hear any suggestions about creating a robust dataloader.

Thanks for your help in advance!

## Public dataset corruption might be common
A lot of users would use public dataset, and the public dataset might contains some corrupted data, especially for dataset with image / video etc.  I totally understand it's dataset owner and user's responsibility to ensure the data integrity / run data cleaning or preprocessing, but it would be easier for developers who would use the dataset

## Use cases
For example, a robust dataloader would be easy for users who want to try quick tests on different dataset, and chose one dataset which fits their needs. So user could use IterableDataloader with `stream=True` to use the dataset easily without downloading and removing corrupted data samples from the dataset.

### Your contribution

The error handling might not trivial and might need more careful design.

ArjunJagdale:
This PR implements support for robust image decoding in the `Image` feature, as discussed in issue #7612.

## 🔧 What was added
- A new boolean field: `ignore_decode_errors` (default: `False`)
- If set to `True`, any exceptions during decoding will be caught, and `None` will be returned instead of raising an error

```python
features = Features({
    "image": Image(decode=True, ignore_decode_errors=True),
})
````

This enables robust iteration over potentially corrupted datasets — especially useful when streaming datasets like WebDataset or image-heavy public sets where sample corruption is common.

## 🧪 Behavior

* If `ignore_decode_errors=False` (default), decoding behaves exactly as before
* If `True`, decoding errors are caught, and a warning is emitted:

  ```
  [Image.decode_example] Skipped corrupted image: ...
  ```

## 🧵 Linked issue

Closes #7612

Let me know if you'd like a follow-up test PR. Happy to write one!

### DIFF
--- a/src/datasets/features/image.py
+++ b/src/datasets/features/image.py
@@ -84,6 +84,10 @@ class Image:
 
     mode: Optional[str] = None
     decode: bool = True
+    
+    # addition - 1
+    ignore_decode_errors: bool = False
+
     id: Optional[str] = field(default=None, repr=False)
     # Automatically constructed
     dtype: ClassVar[str] = "PIL.Image.Image"
@@ -132,23 +136,10 @@ class Image:
                 f"An image sample should have one of 'path' or 'bytes' but they are missing or None in {value}."
             )
 
-    def decode_example(self, value: dict, token_per_repo_id=None) -> "PIL.Image.Image":
+    def decode_example(self, value: dict, token_per_repo_id=None) -> Optional["PIL.Image.Image"]:
         """Decode example image file into image data.
 
-        Args:
-            value (`str` or `dict`):
-                A string with the absolute image file path, a dictionary with
-                keys:
-
-                - `path`: String with absolute or relative image file path.
-                - `bytes`: The bytes of the image file.
-            token_per_repo_id (`dict`, *optional*):
-                To access and decode
-                image files from private repositories on the Hub, you can pass
-                a dictionary repo_id (`str`) -> token (`bool` or `str`).
-
-        Returns:
-            `PIL.Image.Image`
+        Returns None if `ignore_decode_errors=True` and decoding fails.
         """
         if not self.decode:
             raise RuntimeError("Decoding is disabled for this feature. Please use Image(decode=True) instead.")
@@ -159,14 +150,15 @@ class Image:
         else:
             raise ImportError("To support decoding images, please install 'Pillow'.")
 
-        if token_per_repo_id is None:
-            token_per_repo_id = {}
+        try:
+            if token_per_repo_id is None:
+                token_per_repo_id = {}
 
-        path, bytes_ = value["path"], value["bytes"]
-        if bytes_ is None:
-            if path is None:
-                raise ValueError(f"An image should have one of 'path' or 'bytes' but both are None in {value}.")
-            else:
+            path, bytes_ = value["path"], value["bytes"]
+
+            if bytes_ is None:
+                if path is None:
+                    raise ValueError(f"An image should have one of 'path' or 'bytes' but both are None in {value}.")
                 if is_local_path(path):
                     image = PIL.Image.open(path)
                 else:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by ArjunJagdale** [comment]
cc @lhoestq

**Comment by Seas0** [comment]
I think splitting the error handling for the main image decoding process and the metadata decoding process is possibly a bit nicer, as some images do render correctly, but their metadata might be invalid and cause the pipeline to fail, which I've encountered recently as in #7668.

The [`decode_image`](https://docs.pytorch.org/vision/main/generated/torchvision.io.decode_image.html) function in `torchvision` handles similar cases by using the `apply_exif_orientation` flag to turn off the exif metadata processing entirely.

**Comment by ArjunJagdale** [comment]
> I think splitting the error handling for the main image decoding process and the metadata decoding process is possibly a bit nicer, as some images do render correctly, but their metadata might be invalid and cause the pipeline to fail, which I've encountered recently as in #7668.
> 
> The [`decode_image`](https://docs.pytorch.org/vision/main/generated/torchvision.io.decode_image.html) function in `torchvision` handles similar cases by using the `apply_exif_orientation` flag to turn off the exif metadata processing entirely.

@lhoestq & @Seas0 — that makes total sense.

You're right: currently, if EXIF metadata like `.getexif()` fails (due to malformed tags), the whole image gets dropped even if it renders correctly — not ideal.

To address this, I'm planning to split the EXIF handling into a separate `try/except` block, like:

```python
try:
    exif = image.getexif()
    if exif.get(PIL.Image.ExifTags.Base.Orientation) is not None:
        image = PIL.ImageOps.exif_transpose(image)
except Exception as exif_err:
    if self.ignore_decode_errors:
        warnings.warn(f"[Image.decode_example] Skipped EXIF metadata: {exif_err}")
    else:
        raise
```

ArjunJagdale:
> I think splitting the error handling for the main image decoding process and the metadata decoding process is possibly a bit nicer, as some images do render correctly, but their metadata might be invalid and cause the pipeline to fail, which I've encountered recently as in #7668.
> The [`decode_image`](https://docs.pytorch.org/vision/main/generated/torchvision.io.decode_image.html) function in `torchvision` handles similar cases by using the `apply_exif_orientation` flag to turn off the exif metadata processing entirely.
 
 @lhoestq & @Seas0 — that makes total sense.
 
Currently, if EXIF metadata like `.getexif()` fails (due to malformed tags), the whole image gets dropped even if it renders correctly — not ideal.
 
To address this, I'm planning to split the EXIF handling into a separate `try/except` block, like:
```python
try:
    exif = image.getexif()
    if exif.get(PIL.Image.ExifTags.Base.Orientation) is not None:
        image = PIL.ImageOps.exif_transpose(image)
except Exception as exif_err:
    if self.ignore_decode_errors:
        warnings.warn(f"[Image.decode_example] Skipped EXIF metadata: {exif_err}")
    else:
        raise
```

So that, Valid but EXIF-broken images will still be returned & EXIF failures will be skipped only if ignore_decode_errors=True. 

Sounds good??

ArjunJagdale:
With the recent EXIF decoding isolation logic added, this PR now fully addresses:

- ✅ #7612 – Robust iteration over corrupt samples (especially useful in `.streaming=True`)
- ✅ #7632 – Graceful handling of invalid image files when using `cast_column(..., Image(...))`
- ✅ #7668 – Broken EXIF metadata no longer crashes decoding; images are returned if usable

All decoding errors (including `.getexif()` and image file loading) are now skipped with a warning when `ignore_decode_errors=True`. This enables safe, scalable image preprocessing pipelines.

PR #7645, opened 2025-06-25
`ClassLabel` docs: Correct value for unknown labels

l-uuz:
This small change fixes the documentation to to be compliant with what happens in `encode_example`.
https://github.com/huggingface/datasets/blob/e71b0b19d79c7531f9b9bea7c09916b5f6157f42/src/datasets/features/features.py#L1126-L1129


l-uuz:
@lhoestq Can you take a look at this PR? It seems to me that docs is incorrect at that point.

lhoestq (approved):
yes thanks !

Issue #7066, opened 2024-07-23
One subset per file in repo ?


lhoestq:
Right now we consider all the files of a dataset to be the same data, e.g.
```
single_subset_dataset/
├── train0.jsonl
├── train1.jsonl
└── train2.jsonl
```
but in cases like this, each file is actually a different subset of the dataset and should be loaded separately
```
many_subsets_dataset/
├── animals.jsonl
├── trees.jsonl
└── metadata.jsonl
```

It would be nice to detect those subsets automatically using a simple heuristic. For example we can group files together if their paths names are the same except some digits ?

ArjunJagdale:
Hi @lhoestq! I’ve opened a PR that addresses this issue

PR #7646, opened 2025-06-26
Update data_files.py #7066

ArjunJagdale:
Temp #7066

### DIFF
--- a/src/datasets/data_files.py
+++ b/src/datasets/data_files.py
@@ -19,7 +19,30 @@ from .utils import logging
 from .utils import tqdm as hf_tqdm
 from .utils.file_utils import _prepare_path_and_storage_options, is_local_path, is_relative_path, xbasename, xjoin
 from .utils.py_utils import string_to_dict
+import re
+from collections import defaultdict
+from pathlib import Path
 
+def group_files_by_subset(filepaths):
+    """
+    Groups files by subset according to a heuristic:
+    - Files whose names only differ by digits or shard suffixes are grouped together.
+    - Others are placed in separate groups.
+    """
+    def normalize(filename):
+        # Remove trailing numbers or known sharding conventions
+        name = Path(filename).stem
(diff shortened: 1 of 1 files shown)


ArjunJagdale:
It adds automatic grouping of files into subsets based on their root name (e.g., `train0.jsonl`, `train1.jsonl` → `"train"`), as discussed above. The logic is integrated into `FolderBasedBuilder` and is fully tested + documented.

Let me know if any changes are needed — happy to iterate!

lhoestq:
Hi ! I believe the subsets need to be instantiated here as `configs` - not `splits` (which are meant for train/validation/test):

https://github.com/huggingface/datasets/blob/ef762e664a2a1675368ed7a203b0ac8cecca6e19/src/datasets/load.py#L647-L662

Also the subset names should probably be inferred only from the parquet/csv/json files and not from png/jpeg/wav/mp4 etc. WDYT ?

ArjunJagdale:
> Hi ! I believe the subsets need to be instantiated here as `configs` - not `splits` (which are meant for train/validation/test):
> 
> https://github.com/huggingface/datasets/blob/ef762e664a2a1675368ed7a203b0ac8cecca6e19/src/datasets/load.py#L647-L662
> 
> Also the subset names should probably be inferred only from the parquet/csv/json files and not from png/jpeg/wav/mp4 etc. WDYT ?

Thanks a lot for the review!

You're absolutely right — treating subsets as separate configs instead of overloaded splits makes much more sense. If that approach sounds good to you, I can move the grouping logic to `load.py`, where configs are instantiated, and revise the PR to emit one `BuilderConfig` per grouped subset.

Also totally agree on limiting grouping to structured file types — I’d scope this to `.json`, `.jsonl`, `.csv`, and `.parquet`.

Let me know if this direction sounds good, and I’ll get started on the changes right away!

ArjunJagdale:
Hi! @lhoestq!

PR #7658, opened 2025-06-30
Fix: Prevent loss of info.features and column_names in IterableDatasetDict.map when features is None

### LINKED ISSUE #7568: `IterableDatasetDict.map()` call removes `column_names` (in fact info.features)
When calling IterableDatasetDict.map(), each split’s IterableDataset.map() is invoked without a features argument. While omitting the argument isn’t itself incorrect, the implementation then sets info.features = None, which destroys the original feature schema. Since IterableDataset.column_names relies on info.features, it ends up broken (empty or erroring).

**Reproduction**

1.  Define an IterableDatasetDict with a non-None features schema.
2.  my_iterable_dataset_dict contains "text" column.
3. Call:
```
new_dict = my_iterable_dataset_dict.map(
    function=my_fn,
    with_indices=False,
    batched=True,
    batch_size=16,
)
```
4. Observe
```
new_dict["train"].info.features  # {'text': Value(dtype='string', id=None)}
new_dict["train"].column_names   # ['text']
```
5. Call:
```
new_dict = my_iterable_dataset_dict.map(
    function=my_fn,
    with_indices=False,
    batched=True,
    batch_size=16,
    remove_columns=["foo"]
)
```
6. Observe:
```
new_dict["train"].info.features  # → None
new_dict["train"].column_names   # → None
```
5. Internally, in dataset_dict.py this loop omits features ([code](https://github.com/huggingface/datasets/blob/b9efdc64c3bfb8f21f8a4a22b21bddd31ecd5a31/src/datasets/dataset_dict.py#L2047C5-L2056C14)):
```
for split, dataset in self.items():
    dataset_dict[split] = dataset.map(
        function=function,
        with_indices=with_indices,
        input_columns=input_columns,
        batched=batched,
        batch_size=batch_size,
        drop_last_batch=drop_last_batch,
        remove_columns=remove_columns,
        fn_kwargs=fn_kwargs,
-       # features omitted → defaults to None
    )
```
7. Then inside IterableDataset.map() ([code](https://github.com/huggingface/datasets/blob/b9efdc64c3bfb8f21f8a4a22b21bddd31ecd5a31/src/datasets/iterable_dataset.py#L2619C1-L2622C37)) correct `info.features` is replaced by features which is None:
```
info = self.info.copy()
info.features = features  # features is None here
return IterableDataset(..., info=info, ...)
```

**Suggestion**:
It looks like this replacement was added intentionally but maybe should be done only if `features` is not None.

ArjunJagdale:
This PR fixes a bug where calling `IterableDatasetDict.map()` or `IterableDataset.map()` with the default `features=None` argument would overwrite the existing `info.features` attribute with `None`. This, in turn, caused the resulting dataset to lose its schema, breaking downstream usage of attributes like `column_names`.

Why
Previously, the code would always set `info.features = features`, even if `features` was `None`. When mapping with removal of columns or other transformations, this led to the destruction of the schema and caused failures in code that relied on the dataset schema being present.

How
We now update `info.features` only if `features` is not `None`. This preserves the original schema unless the user explicitly provides a new one.

Reference
Fixes #7568

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -2713,8 +2713,8 @@ class IterableDataset(DatasetInfoMixin):
             features=features,
         )
         info = self.info.copy()
-        if features is not None:
-            info.features = features
+        info.features = features or self.info.features
+        
         return IterableDataset(
             ex_iterable=ex_iterable,
             info=info,


ArjunJagdale:
Hi!
I haven’t included a test for this change, as the fix is quite small and targeted.
Please let me know if you’d like a test for this case or if you’d prefer to handle it during review.
Thanks!

lhoestq:
we can't know in advance the `features` after map() (it transforms the data !), so you can reuse the `features` from `info.features`

ArjunJagdale:
I'll the patch as suggested — `info.features = features` or `self.info.features` — to ensure schema preservation while keeping the logic simple and explicit. WDYT?

lhoestq:
info.features should be None in the general case, and replaced by the user's `features` if it's passed explicitly with `map(..., features=...)`

https://github.com/huggingface/datasets/issues/7568 is not an issue we can fix

ArjunJagdale:
> info.features should be None in the general case, and replaced by the user's `features` if it's passed explicitly with `map(..., features=...)`
> 
> #7568 is not an issue we can fix

Thanks for the clarification! Totally makes sense now — I understand that features=None is the expected behavior post-map() unless explicitly passed, and that preserving old schema by default could lead to incorrect assumptions.
Closing this one — appreciate the feedback as always

Issue #7660, opened 2025-06-30
AttributeError: '_tqdm_cls' object has no attribute '_lock'


Hypothesis-Z:
### Describe the bug

`AttributeError: type object 'tqdm' has no attribute '_lock'`

It occurs when I'm trying to load datasets in thread pool.

Issue https://github.com/huggingface/datasets/issues/6066 and PR https://github.com/huggingface/datasets/pull/6067 https://github.com/huggingface/datasets/pull/6068 tried to fix this. 

### Steps to reproduce the bug

Will have to try serveral times to reproduce the error cause it is concerned with threads.
 
1. save some datasets for test
```pythonfrom datasets import Dataset, DatasetDict
import os

os.makedirs("test_dataset_shards", exist_ok=True)

for i in range(10):
    data = Dataset.from_dict({"text": [f"example {j}" for j in range(1000000)]})
    data = DatasetDict({'train': data})
    data.save_to_disk(f"test_dataset_shards/shard_{i}")
```

2. load them in a thread pool 
```python
from datasets import load_from_disk
from tqdm import tqdm
from concurrent.futures import ThreadPoolExecutor, as_completed
import glob

datas = glob.glob('test_dataset_shards/shard_*')
with ThreadPoolExecutor(max_workers=10) as pool:
    futures = [pool.submit(load_from_disk, it) for it in datas]
datas = []
for future in tqdm(as_completed(futures), total=len(futures)):
    datas.append(future.result())
```

### Expected behavior

no exception raised

### Environment info

datasets==2.19.0
python==3.10

Hypothesis-Z:
Deleting a class (**not instance**) attribute might be invalid in this case, which is `tqdm` doing in `ensure_lock`.

```python
from tqdm import tqdm as old_tqdm

class tqdm1(old_tqdm):
     def __delattr__(self, attr):
        try:
            super().__delattr__(attr)
        except AttributeError:
            if attr != '_lock':
                print(attr)
                raise

class Meta(type):
    def __delattr__(cls, name):
        if name == "_lock":
            return  
        return super().__delattr__(name)
    
class tqdm2(old_tqdm, metaclass=Meta):
    pass

del tqdm2._lock
del tqdm1._lock # error
```

https://github.com/huggingface/datasets/blob/e71b0b19d79c7531f9b9bea7c09916b5f6157f42/src/datasets/utils/tqdm.py#L104-L122

Hypothesis-Z:
[...]

PR #7661, opened 2025-07-01
fix del tqdm lock error

Hypothesis-Z:
fixes https://github.com/huggingface/datasets/issues/7660


lhoestq:
let's see which solution is found at https://github.com/huggingface/huggingface_hub/pull/3286 and do the same maybe ?

PR #7690, opened 2025-07-18
HDF5 support

klamike:
This PR adds support for (almost) arbitrary HDF5 file(s) by converting them to Arrow. The heavy lifter here is `datasets.features.features.numpy_to_pyarrow_listarray`.

Haven't had time to add tests yet, but maintainer edits are allowed 😉 Also, there are a few things I am not sure about regarding endianness/dtypes which I have marked with `FIXME`.

### DIFF
--- a/src/datasets/packaged_modules/hdf5/hdf5.py
+++ b/src/datasets/packaged_modules/hdf5/hdf5.py
@@ -1,0 +1,192 @@
+import itertools
+from dataclasses import dataclass
+from typing import Dict, List, Optional
+
+import h5py
+import numpy as np
+import pyarrow as pa
+
+import datasets
+from datasets.table import table_cast
+
+
+logger = datasets.utils.logging.get_logger(__name__)
+
+EXTENSIONS = [".h5", ".hdf5"]
+
+
+@dataclass
+class HDF5Config(datasets.BuilderConfig):
+    """BuilderConfig for HDF5."""
+
+    batch_size: Optional[int] = None
+    columns: Optional[List[str]] = None
+    features: Optional[datasets.Features] = None
+
+    def __post_init__(self):
+        super().__post_init__()
+
+
+class HDF5(datasets.ArrowBasedBuilder):
+    """ArrowBasedBuilder that converts HDF5 files to Arrow tables using the HF extension types."""
+
+    BUILDER_CONFIG_CLASS = HDF5Config
+
+    def _info(self):
+        if (
+            self.config.columns is not None
+            and self.config.features is not None
+            and set(self.config.columns) != set(self.config.features)
+        ):
+            raise ValueError(
+                "The columns and features argument must contain the same columns, but got ",
+                f"{self.config.columns} and {self.config.features}",
+            )
+        return datasets.DatasetInfo(features=self.config.features)
+
+    def _split_generators(self, dl_manager):
+        if not self.config.data_files:
+            raise ValueError(f"At least one data file must be specified, but got data_files={self.config.data_files}")
+        dl_manager.download_config.extract_on_the_fly = True
+        data_files = dl_manager.download_and_extract(self.config.data_files)
+        splits = []
+        for split_name, files in data_files.items():
+            if isinstance(files, str):
+                files = [files]
+
+            files = [dl_manager.iter_files(file) for file in files]
+            # Infer features from first file
+            if self.info.features is None:
+                for first_file in itertools.chain.from_iterable(files):
+                    with h5py.File(first_file, "r") as h5:
+                        dataset_map = _traverse_datasets(h5)
+                        features_dict = {}
+                        for path, dset in dataset_map.items():
+                            feat = _infer_feature_from_dataset(dset)
+                            features_dict[path] = feat
+                        self.info.features = datasets.Features(features_dict)
+                    break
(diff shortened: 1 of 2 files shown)


klamike:
This is ready for review now. Note that it doesn't support *all* HDF5 files (and I don't think that's worth attempting)... the biggest assumption is that the first dimension of each dataset corresponds to rows in the split.

klamike:
A few to-dos which I think can be left for future PRs (which I am happy to do/help with -- just this one is already huge 😄 ):
- [Enum types](https://docs.h5py.org/en/stable/special.html#enumerated-types)
- HDF5 [io](https://github.com/huggingface/datasets/tree/main/src/datasets/io)
- [dataset-viewer](https://github.com/huggingface/dataset-viewer) support (not sure if changes are needed with the way it is written now)

setup.py
```diff
@@ -166,6 +166,7 @@
     "aiohttp",
     "elasticsearch>=7.17.12,<8.0.0",  # 8.0 asks users to provide hosts or cloud_id when instantiating ElasticSearch(); 7.9.1 has legacy numpy.float_ which was fixed in https://github.com/elastic/elasticsearch-py/pull/2551.
     "faiss-cpu>=1.8.0.post1",  # Pins numpy < 2
+    "h5py",  # FIXME: probably needs a lower bound
```
klamike, line 169:
```suggestion
    "h5py>=2.3",  # FIXME: probably needs a lower bound
```
Probably the most recent thing we need is the vlen/complex/compound, which [according to the docs was added in 2.3](https://docs.h5py.org/en/stable/special.html#:~:text=As%20of%20version%202.3%2C%20h5py%20fully%20supports%20HDF5%20enums%20and%20VL%20types.). Version as of writing is 3.14

src/datasets/packaged_modules/hdf5/hdf5.py
```diff
@@ -0,0 +1,375 @@
...
+
+    def _cast_table(self, pa_table: pa.Table) -> pa.Table:
+        if self.info.features is not None:
+            relevant_features = {
+                col: self.info.features[col] for col in pa_table.column_names if col in self.info.features
+            }
+            has_zero_dims = any(_has_zero_dimensions(feature) for feature in relevant_features.values())
+            # FIXME: pyarrow.lib.ArrowInvalid: list_size needs to be a strict positive integer
+            if not has_zero_dims:
+                pa_table = table_cast(pa_table, self.info.features.arrow_schema)
```
klamike, line 122: This is not ideal -- `table_cast` does not support zero-dim arrays, so I skip calling it if there are any zero-dim arrays in the table.

src/datasets/packaged_modules/hdf5/hdf5.py
```diff
@@ -0,0 +1,375 @@
...
+                        if dset.shape[0] != num_rows:
+                            raise ValueError(
+                                f"Dataset '{path}' length {dset.shape[0]} differs from {num_rows} in file '{file}'"
+                            )
+                    effective_batch = batch_size_cfg or self._writer_batch_size or num_rows
+                    for start in range(0, num_rows, effective_batch):
+                        end = min(start + effective_batch, num_rows)
+                        batch_dict = {}
+                        for path, dset in dataset_map.items():
+                            arr = dset[start:end]
```
klamike, line 161: Here is where the actual read from disk happens

klamike:
@lhoestq any interest in merging this? Let me know if I can do anything to make reviewing it easier!

lhoestq:
Sorry for the delay, I'll review your PR soon :)

klamike:
Thanks for the review @lhoestq! Rebased on main and incorporated most of your suggestions.

I believe the only one left is the zero-dim handling with `table_cast`...

klamike:
@lhoestq is 2c4bfba what you meant?

src/datasets/packaged_modules/hdf5/hdf5.py
```diff
@@ -0,0 +1,375 @@
...
+    """Create separate features for each field in compound data."""
+    field_names = list(dset.dtype.names)
+    logger.info(
+        f"Compound dataset '{base_path}' (dtype: {dset.dtype}) flattened into {len(field_names)} columns: {field_names}"
+    )
+
+    features = {}
+    for field_name in field_names:
+        field_dtype = dset.dtype[field_name]
+        field_path = f"{base_path}_{field_name}"
```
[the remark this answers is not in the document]
lhoestq: Looks good ! btw with `ds = ds.flatten()` you can get a flat structure with columns named `"position.x"`, `"position.y"` etc.

klamike:
Awesome! Yes, I'm happy to help with the docs. Would appreciate any pointers, we can discuss in #7740.

It does look like there was a CI test failure, though it seems unrelated?
```
FAILED tests/test_dataset_dict.py::test_dummy_datasetdict_serialize_fs - ValueError: Protocol not known: mock
FAILED tests/test_arrow_dataset.py::test_dummy_dataset_serialize_fs - ValueError: Protocol not known: mock
```
Also, what do you think of the todos in https://github.com/huggingface/datasets/pull/7690#issuecomment-3105391677 ? In particular I think support in dataset-viewer would be nice.

lhoestq:
Cool ! Yeah the failure is unrelated

Regarding the Viewer, it should work out of the box when it's updated with the next version of `datasets` :)

PR #7701, opened 2025-07-26
Update fsspec max version to current release 2025.7.0

rootAvish:
Diffusers currently asks for a max fsspec version of `2025.3.0`. This change updates it to the current latest version. This change is mainly required to resolve conflicts with other packages in an environment. In my particular case, `aider-chat` which is a part of my environment installs `2025.5.1` which is incompatible with `datasets`.


rootAvish:
@lhoestq I ran the test suite locally and while some tests were failing those failures are present on the main branch too. Could you please review and trigger the CI?

Shiti:
Which release will this be available in ? I'm running into this issue with `datasets=3.6.0`

Issue #7700, opened 2025-07-25
[doc] map.num_proc needs clarification


sfc-gh-sbekman:
https://huggingface.co/docs/datasets/v4.0.0/en/package_reference/main_classes#datasets.Dataset.map.num_proc
```
num_proc (int, optional, defaults to None) — Max number of processes when generating cache. Already cached shards are loaded sequentially.
```

for batch:
```
num_proc (int, optional, defaults to None): The number of processes to use for multiprocessing. If None, no multiprocessing is used. This can significantly speed up batching for large datasets.
```

So what happens to `map` - is it the same API as `batch` wrt `num_proc` - so only if `None` then no multiprocessing is used.

Let's update the doc to be unambiguous.

context: debugging a failing `map`

Thank you!

PR #7702, opened 2025-07-26
num_proc=0 behave like None, num_proc=1 uses one worker (not main process) and clarify num_proc documentation

tanuj-rai:
Fixes issue #7700

This PR makes num_proc=0 behave like None in Dataset.map(), disabling multiprocessing.
It improves UX by aligning with DataLoader(num_workers=0) behavior.
The num_proc docstring is also updated to clearly explain valid values and behavior.

@SunMarc

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -5615,7 +5615,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
             total=num_shards,
             desc=desc,
         )
-        with contextlib.nullcontext() if num_proc is None or num_proc <= 1 else Pool(num_proc) as pool:
+        with contextlib.nullcontext() if num_proc is None or num_proc < 1 else Pool(num_proc) as pool:
             update_stream = (
                 Dataset._push_parquet_shards_to_hub_single(**kwargs_iterable[0])
                 if pool is None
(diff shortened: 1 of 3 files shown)


lhoestq:
I think we can support num_proc=0 and make it equivalent to `None` to make it simpler

tanuj-rai:
> I think we can support num_proc=0 and make it equivalent to `None` to make it simpler

Thank you @lhoestq for reviewing it. Please let me know if anything needs to be updated further.

src/datasets/arrow_dataset.py
```diff
@@ -3007,8 +3007,12 @@ def map(
...
                 Keyword arguments to be passed to `function`.
             num_proc (`int`, *optional*, defaults to `None`):
-                Max number of processes when generating cache. Already cached shards are loaded sequentially.
-            suffix_template (`str`):
+                    The number of processes to use for multiprocessing.  
+                    - If `None` or `0`, no multiprocessing is used and the operation runs in the main process.  
+                    - If greater than `1`, multiple worker processes are used to process data in parallel.  
+                    - Setting `num_proc=0` is not supported and will raise a `ValueError`.
+                     Note: The function passed to `map()` must be picklable for multiprocessing to work correctly (i.e., defined at the top level of a module, not inside another function or class).
+                           suffix_template (`str`):
```
lhoestq, line 3015:
```suggestion
                 The number of processes to use for multiprocessing.  
                - If `None` or `0`, no multiprocessing is used and the operation runs in the main process.  
                - If greater than `1`, one or multiple worker processes are used to process data in parallel.  
                 Note: The function passed to `map()` must be picklable for multiprocessing to work correctly (i.e., prefer functions defined at the top level of a module, not inside another function or class).
             suffix_template (`str`):
```

lhoestq (approved):
thanks ç

PR #7704, opened 2025-07-26
Fix map() example in datasets documentation: define tokenizer before use

### LINKED ISSUE #7703: [Docs] map() example uses undefined `tokenizer` — causes NameError
## Description

The current documentation example for `datasets.Dataset.map()` demonstrates batched processing but uses a `tokenizer` object without defining or importing it. This causes an error every time it's copied.

Here is the problematic line:

```python
# process a batch of examples
>>> ds = ds.map(lambda example: tokenizer(example["text"]), batched=True)
```
This assumes the user has already set up a tokenizer, which contradicts the goal of having self-contained, copy-paste-friendly examples. 

## Problem
Users who copy and run the example as-is will encounter:

```python
NameError: name 'tokenizer' is not defined
```

This breaks the flow for users and violates HuggingFace's documentation principle that examples should "work as expected" when copied directly. 

## Proposal
Update the example to include the required tokenizer setup using the Transformers library, like so:

```python
from transformers import AutoTokenizer

tokenizer = AutoTokenizer.from_pretrained("bert-base-uncased")
ds_tokenized = ds.map(lambda example: tokenizer(example["text"]), batched=True)
```
This will help new users understand the workflow and apply the method correctly.

Sanjaykumar030:
## Problem
The current datasets.Dataset.map() example in the documentation demonstrates batched processing using a tokenizer object without defining or importing it. This causes a NameError when users copy and run the example as-is, breaking the expected seamless experience.

## Correction
This PR fixes the issue by explicitly importing and initializing the tokenizer using the Transformers library (AutoTokenizer.from_pretrained("bert-base-uncased")), making the example self-contained and runnable without errors.
This will help new users understand the workflow and apply the method correctly.

Closes #7703

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -232,26 +232,19 @@ to this dataset.
 
 The syntax for Example docstrings can look as follows:
 
-```
-    Example:
+Example:
 
-    ```py
+   ``` python
     >>> from datasets import load_dataset
+    >>> from transformers import AutoTokenizer
     >>> ds = load_dataset("cornell-movie-review-data/rotten_tomatoes", split="validation")
     >>> def add_prefix(example):
     ...     example["text"] = "Review: " + example["text"]
     ...     return example
     >>> ds = ds.map(add_prefix)
-    >>> ds[0:3]["text"]
-    ['Review: compassionately explores the seemingly irreconcilable situation between conservative christian parents and their estranged gay and lesbian children .',
-        'Review: the soundtrack alone is worth the price of admission .',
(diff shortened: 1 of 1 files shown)


Sanjaykumar030:
Hi @lhoestq, just a gentle follow-up on this doc fix PR (#7704). Let me know if any changes are needed — happy to update.
Hope this improvement helps users run the example without confusion!

Sanjaykumar030:
Hi @lhoestq, just a gentle follow-up on this doc fix PR (#7704). Let me know if any changes are needed — happy to update.
Hope this improvement helps users run the example without confusion!

lhoestq:
the modified file is the readme of the docs, not about map() specifically

PR #7713, opened 2025-07-31
Update cli.mdx to refer to the new "hf" CLI

evalstate:
Update to refer to `hf auth login`


lhoestq (approved):
great ! It would be cool to update the rest of the docs with `hf` instead of `huggingface-cli`

PR #7718, opened 2025-08-01
add support for pyarrow string view in features

onursatici:


src/datasets/features/features.py
```diff
@@ -110,6 +110,8 @@ def _arrow_to_datasets_dtype(arrow_type: pa.DataType) -> str:
         return "string"
     elif pyarrow.types.is_large_string(arrow_type):
         return "large_string"
+    elif pyarrow.types.is_string_view(arrow_type):
+        return "string_view"
```
onursatici, line 114: `pa.string_view` exists, so in the complementing function`string_to_arrow("string_view")` will return `pa.StringView` correctly

onursatici:
@lhoestq who do you think would be the best to have a look at this? Any pointers would be appreciated, thanks!

lhoestq:
Hi ! what's the rationale for supporting string view ? I'm afraid it can complexify the typing logic without much value

onursatici:
Hi @lhoestq ! I mainly want to be able to create features by using `Features.from_arrow_schema(dataset_schema)` on an arrow dataset with string view columns, currently there is no easy way to do this, and string_view is becoming an increasingly common data type for string columns in arrow. Thanks for having a look!

PR #7734, opened 2025-08-09
Fixing __getitem__ of datasets which behaves inconsistent to documentation when setting _format_type to None

awagen:
Setting _format_type to None, should return plain python object but as of 4.0.0 returns Column. This fails in libs such as sentencetransformers (such as in generation of hard negatives) where plain python is expected.


lhoestq:
this breaking change is actually expected, happy to help with a fix in sentencetransformers to account for this

awagen:
Thank you for the context. I thought this was a mismatch do the documentation. Good to know it was intentional. No worries, can add a PR to sentence transformers.

PR #7737, opened 2025-08-13
docs: Add column overwrite example to batch mapping guide

Sanjaykumar030:
This PR adds a complementary example showing the **column-overwriting** pattern, which is both more direct and more flexible for many transformations.

### Proposed Change

The original `remove_columns` example remains untouched. Below it, this PR introduces an alternative approach that overwrites an existing column during batch mapping.

This teaches users a core `.map()` capability for in-place transformations without extra intermediate steps.

**New Example:**

> ```python
> >>> from datasets import Dataset
> >>> dataset = Dataset.from_dict({"a": [0, 1, 2]})
> # Overwrite "a" directly to duplicate each value
> >>> duplicated_dataset = dataset.map(
> ...     lambda batch: {"a": [x for x in batch["a"] for _ in range(2)]},
> ...     batched=True
> ... )
> >>> duplicated_dataset
> Dataset({
>     features: ['a'],
>     num_rows: 6
> })
> >>> duplicated_dataset["a"]
> [0, 0, 1, 1, 2, 2]
> ```

### DIFF
--- a/docs/source/about_map_batch.mdx
+++ b/docs/source/about_map_batch.mdx
@@ -44,7 +44,7 @@ For example, here’s how to duplicate every row in the dataset by overwriting c
 ```py
 >>> from datasets import Dataset
 >>> dataset = Dataset.from_dict({"a": [0, 1, 2]})
-# Overwrites the existing "a" column with duplicated values
+# overwrites the existing "a" column with duplicated values
 >>> duplicated_dataset = dataset.map(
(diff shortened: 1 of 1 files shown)


Sanjaykumar030:
Hi @lhoestq, just a gentle follow-up on this PR.

lhoestq (approved):
sounds good !

Issue #7751, opened 2025-08-27
Dill version update


Navanit-git:
### Describe the bug

Why the datasets is not updating the dill ? 
Just want to know if I update the dill version in dill what will be the repucssion. 
For now in multiplaces I have to update the library like process requirequire dill 0.4.0 so why not datasets. 
Adding a pr too. 

### Steps to reproduce the bug

.

### Expected behavior

.

### Environment info

.

Navanit-git:
#7752

lhoestq:
related: #7510

PR #7752, opened 2025-08-27
Fix: Update Dill Version in Setup py

Navanit-git:
Currently the DIll version is less than 3.9 and now major libraries like Multiprocess, gepa requires Dill version as 0.4.0 and this is making a conflict in installation. So added this small PR to update the DIll.


Navanit-git:
https://github.com/huggingface/datasets/issues/7751

lhoestq:
same as https://github.com/huggingface/datasets/pull/7763: some tests need to be fixed to support 0.4.0

lhoestq:
fixed by #7763

Issue #7594, opened 2025-06-05
Add option to ignore keys/columns when loading a dataset from jsonl(or any other data format)


avishaiElmakies:
### Feature request

Hi, I would like the option to ignore keys/columns when loading a dataset from files (e.g. jsonl).

### Motivation

I am working on a dataset which is built on jsonl. It seems the dataset is unclean and a column has different types in each row. I can't clean this or remove the column (It is not my data and it is too big for me to clean and save on my own hardware). 
I would like the option to just ignore this column when using `load_dataset`, since i don't need it.
I tried to look if this is already possible but couldn't find a solution. if there is I would love some help. If it is not currently possible, I would love this feature

### Your contribution

I don't think I can help this time, unfortunately.

lhoestq:
Good point, I'd be in favor of having the `columns` argument in `JsonConfig` (and the others) to align with `ParquetConfig` to let users choose which columns to load and ignore the rest

avishaiElmakies:
Is it possible to ignore columns when using parquet?

lhoestq:
Yes, you can pass `columns=...` to load_dataset to select which columns to load, and it is passed to `ParquetConfig` :)

avishaiElmakies:
Ok, i didn't know that. 
Anyway, it would be good to add this to others

ArjunJagdale:
Hi @lhoestq 

I'd like to take this up!

As you suggested, I’ll extend the support for the columns parameter (currently used in ParquetConfig) to JsonConfig as well. This will allow users to selectively load specific keys/columns from .jsonl (or .json) files and ignore the rest — solving the type inconsistency issues in unclean datasets.

ArjunJagdale:
Hi @avishaiElmakies and @lhoestq 

Just wanted to let you know that this is now implemented in #7594
As suggested, support for the `columns=...` argument (previously available for Parquet) has now been extended to **JSON and JSONL** loading via `load_dataset(...)`. You can now load only specific keys/columns and skip the rest — which should help in cases where some fields are unclean, inconsistent, or just unnecessary.

### ✅ Example:

```python
from datasets import load_dataset

dataset = load_dataset("json", data_files="your_data.jsonl", columns=["id", "title"])
print(dataset["train"].column_names)
# Output: ['id', 'title']
```

### 🔧 Summary of changes:

* Added `columns: Optional[List[str]]` to `JsonConfig`
* Updated `_generate_tables()` to filter selected columns
[...]

PR #7754, opened 2025-09-04
Add columns support to JSON loader

ArjunJagdale:
New fix to #7594
This PR adds support for the columns argument in the JSON dataset builder.
Added columns parameter to JsonConfig.
Applied column filtering after table creation, filling missing columns with None.

Extended tests to cover:
- Selecting a subset of columns
- Handling missing requested columns
- Column selection on list-of-strings case

### DIFF
--- a/src/datasets/packaged_modules/json/json.py
+++ b/src/datasets/packaged_modules/json/json.py
@@ -1,7 +1,7 @@
 import io
 import itertools
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


fish4terrisa-MSDSM:
Kinda strange this one is never merged and never even got any attention.
Maybe you can make another pull request again? The code base didnt change much.
This is really useful.

PR #7755, opened 2025-09-05
Support pathlib.Path for feature input

Joshua-Chin:
This PR adds support for specifying image, video, audio, and pdf features using `pathlib.Path`.

### DIFF
--- a/src/datasets/features/pdf.py
+++ b/src/datasets/features/pdf.py
@@ -95,7 +95,7 @@ class Pdf:
         if isinstance(value, str):
             return {"path": value, "bytes": None}
         elif isinstance(value, Path):
-            return {"path": str(value.absolute), "bytes": None}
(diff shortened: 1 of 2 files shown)


src/datasets/features/pdf.py
```diff
@@ -92,6 +94,8 @@ def encode_example(self, value: Union[str, bytes, bytearray, dict, "pdfplumber.p
 
         if isinstance(value, str):
             return {"path": value, "bytes": None}
+        elif isinstance(value, Path):
+            return {"path": str(value.absolute), "bytes": None}
```
lhoestq, line 98:
```suggestion
            return {"path": str(value.absolute()), "bytes": None}
```

lhoestq (approved):
awesome ! I just did a minor fix

PR #7762, opened 2025-09-11
Parquet: use data page v2 for efficient page pruning

lhoestq:
This is needed to enable page pruning with DataFusion, which will be useful for the Dataset Viewer.

Indeed page pruning with DataFusion allows to download only certain pages of a row group, reducing the I/O required to read just a few rows.

But while data page v1 generally works, it's not easy with DataFusion to do page pruning on datasets with nested data. This is because rows can span multiple pages in v1, contrary to v2.

cc @severo for viz

### DIFF
--- a/src/datasets/arrow_writer.py
+++ b/src/datasets/arrow_writer.py
@@ -745,12 +745,15 @@ class ArrowWriter:
 
 
 class ParquetWriter(ArrowWriter):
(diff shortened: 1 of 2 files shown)


lhoestq:
Closing this since it looks like the page offset index is enough :)

PR #7778, opened 2025-09-17
[FIX] force spawning pool for MacOS

burtenshaw:
This PR gets multiprocessing to work on mac os:

```python

ds = load_dataset("fka/awesome-chatgpt-prompts", split="train").take(100)

ds = ds.map(lambda x: x, num_proc=4)

ds.push_to_hub("burtenshaw/dataset-test", num_proc=4)
```

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -55,6 +55,7 @@ from typing import (
 )
 
 import fsspec
+import multiprocessing as mp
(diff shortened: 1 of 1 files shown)


lhoestq:
After more discussions on slack, we can switch the default to spawn.

Let's use multiprocess instead of multiprocessing and maybe add a check to apply this only on Macos ?

Issue #7745, opened 2025-08-22
Audio mono argument no longer supported, despite class documentation


jheitz:
### Describe the bug

Either update the documentation, or re-introduce the flag (and corresponding logic to convert the audio to mono)

### Steps to reproduce the bug

Audio(sampling_rate=16000, mono=True) raises the error 
TypeError: Audio.__init__() got an unexpected keyword argument 'mono'

However, in the class documentation, is says:
Args:
        sampling_rate (`int`, *optional*):
            Target sampling rate. If `None`, the native sampling rate is used.
        mono (`bool`, defaults to `True`):
            Whether to convert the audio signal to mono by averaging samples across
            channels.
        [...]

### Expected behavior

The above call should either work, or the documentation within the Audio class should be updated

### Environment info

- `datasets` version: 4.0.0
- Platform: Linux-5.15.0-124-generic-x86_64-with-glibc2.35
- Python version: 3.12.11
- `huggingface_hub` version: 0.34.4
- PyArrow version: 21.0.0
- Pandas version: 2.3.2
- `fsspec` version: 2025.3.0

REDXtremee:
i wanna work on this issue so can you assign it to me?

REDXtremee:
I want to solve this problem can you please assign it to me
and also can you please guide whether the mono parameter is required to be re-added or the documentation needs an update?

jbbqqf:
Hi! Triaging older issues — I think the docstring half of this is already resolved.

On current `main` (`datasets==4.8.6.dev0`), the `Audio` feature's `__init__` accepts `sampling_rate`, `num_channels`, `decode`, `stream_index`, `id` — and the docstring documents `num_channels` (not `mono`) with the explanation that `1` = mono, `2` = stereo, `None` = preserve source. There is no `mono` argument anywhere in `src/datasets/features/audio.py` and no leftover `mono` mention in the docstring.

I verified by grepping `src/datasets/features/audio.py` for `mono`: only one match remains, and it's the descriptive prose in the `num_channels` arg doc, not a parameter. So the doc-vs-code mismatch reported here is gone.

Would you mind closing this out? If `Audio(num_channels=1)` is still misbehaving for you in a way the docstring doesn't predict, that would be a separate bug worth a fresh issue (and there's already a related one for stereo at #8005, fwiw).

(Disclosure: I drafted this comment with help from Claude Code while triaging stale issues; the grep above was run against current `main`.)

PR #7785, opened 2025-09-22
Fix Audio docstring by removing unsupported mono argument

tanuj-rai:
This PR fixes issue #7745.

Who can review:
@lhoestq


lhoestq:
I think we can keep the arg and add the missing torch.mean() in the Audio.decode_example method

tanuj-rai:
> I think we can keep the arg and add the missing torch.mean() in the Audio.decode_example method

Thank you @lhoestq. I will add torch.mean().

lhoestq:
fixed by #7840

PR #7795, opened 2025-09-29
Add pyarrow's binary view to features

delta003:
Basically https://github.com/huggingface/datasets/pull/7718 just for binary view instead of string view


delta003:
@lhoestq 🙏

PR #7820, opened 2025-10-15
Keep hffs cache in workers when streaming

lhoestq:
(and also reorder the hffs args to improve caching)

When using `DataLoader(iterable_dataset, num_workers=...)` the dataset is pickled and passed to the worker. However previously the resulting dataset would be in a process with an empty hffs cache. By keeping the cache attached to `IterableDataset`, the cached hffs instances are pickled with the dataset and re-populates the cache in the DataLoader workers

this requires https://github.com/huggingface/huggingface_hub/pull/3443 to work effectively though, otherwise the unpickled hffs cache would start empty

cc @andimarafioti @ltmeyer

### DIFF
--- a/src/datasets/utils/file_utils.py
+++ b/src/datasets/utils/file_utils.py
@@ -895,8 +895,8 @@ def _prepare_single_hop_path_and_storage_options(
             storage_options["headers"] = {"Accept-Encoding": "identity", **headers}
     elif protocol == "hf":
         storage_options = {
-            "token": token,
             "endpoint": config.HF_ENDPOINT,
+            "token": token,
             **storage_options,
         }
     if storage_options:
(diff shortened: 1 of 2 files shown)


src/datasets/utils/file_utils.py
```diff
@@ -895,8 +895,8 @@ def _prepare_single_hop_path_and_storage_options(
             storage_options["headers"] = {"Accept-Encoding": "identity", **headers}
     elif protocol == "hf":
         storage_options = {
-            "token": token,
             "endpoint": config.HF_ENDPOINT,
+            "token": token,
```
andimarafioti: same, weird that you need this :S

andimarafioti (commented):
I'll try this :)

PR #7823, opened 2025-10-17
Fix random seed on shuffle and interleave_datasets

### LINKED ISSUE #7567: interleave_datasets seed with multiple workers
### Describe the bug

Using interleave_datasets with multiple dataloader workers and a seed set causes the same dataset sampling order across all workers.

Should the seed be modulated with the worker id?

See above

### Expected behavior

See above

### Environment info

- `datasets` version: 3.5.1
- Platform: macOS-15.4.1-arm64-arm-64bit
- Python version: 3.12.9
- `huggingface_hub` version: 0.30.2
- PyArrow version: 19.0.1
- Pandas version: 2.2.3
- `fsspec` version: 2024.12.0

CloseChoice:
closes #7567

Add `shift_rngs` method to `ExamplesIterable` that is called directly after sharding. If a generator is available (not the case for all subclasses) we update the seed of the generator by shifting by the worker_id.

~This is just the fix for `shuffle`, in the corresponding issue `interleave_datasets` is mentioned as well, which won't be fixed with this approach.~

EDIT: This is a fix for `shuffle` and `interleave_datasets`. Adding recursivity to `shift_rngs` solved `interleave_datasets` as well. Not sure though if this is completely safe or if we could destroy something with that. I don't think so but could be wrong and appreciate some guidance from the maintainers. I also checked, on a single_worker we are always handing over `index=0` so that case preserves the seed the user specified.

### DIFF
--- a/tests/test_iterable_dataset.py
+++ b/tests/test_iterable_dataset.py
@@ -1598,6 +1598,32 @@ def test_iterable_dataset_interleave_dataset_with_multiple_workers():
         assert len(set(values)) != 1, "Make sure not all values are identical"
 
 
+@require_torch
+def test_iterable_dataset_interleave_dataset_deterministic_across_iterations():
+    # GH 7567
+    from torch.utils.data import DataLoader
+
+    def gen(shard, value):
+        for i in range(50):
+            yield {"value": value, "id": i}
+
+    num_workers = 10
+    ds = [
+        IterableDataset.from_generator(gen, gen_kwargs={"shard": list(range(num_workers)), "value": i})
+        for i in range(5)
+    ]
+    ds = interleave_datasets(ds, probabilities=[1 / len(ds)] * len(ds), seed=1234)
+    dataloader = DataLoader(ds, batch_size=None, num_workers=num_workers)
+
+    # First iteration
+    first_result = list(dataloader)
+
+    # Second iteration
+    second_result = list(dataloader)
+
+    assert first_result == second_result, "Results should be identical across iterations when using same seed"
+
+
 @pytest.mark.parametrize("batch_size", [4, 5])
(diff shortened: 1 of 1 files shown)


lhoestq:
Cool ! To avoid unwanted side effects it could be implemented for every class instead of using `hasattr`, and also return a new instance instead of modifying it in place. Some other functions work like that already like `shuffle_data_sources` and `shard_data_sources`.

Returning a new object is actually quite important otherwise iterating on the dataset multiple times would shift the RNGs every time

CloseChoice:
> Cool ! To avoid unwanted side effects it could be implemented for every class instead of using `hasattr`, and also return a new instance instead of modifying it in place. Some other functions work like that already like `shuffle_data_sources` and `shard_data_sources`.
> 
> Returning a new object is actually quite important otherwise iterating on the dataset multiple times would shift the RNGs every time

Thanks for the review. 

I managed to return instances of `_BaseExamplesIterable` instead of modifying inplace and this shifts the seed correctly even in the case of multiple iterations (see additional test). But I kept the `hasattr` for the reason that we have chained ex_iterables (so an ex_iterable can have other ex_iterables) and for the child ex_iterable it is not clear if they do have a generator attribute (e.g. `ArrowExamplesIterable`, `RebatchedArrowExamplesIterable`, `SelectColumnsIterable`, `StepExamplesIterable`, `SkipExamplesIterable`, `RepeatExamplesIterable` all don't have that). We could implement a generic method `shift_rngs` on the baseclass and then just return the ex_iterable and get rid of `hasattr` but I'd prefer the `hasattr` solution here. Let me now what you think about that.

lhoestq (approved):
I see, it sounds good to me as is, we can improve later if needed

Issue #7804, opened 2025-10-09
Support scientific data formats


lhoestq:
List of formats and libraries we can use to load the data in `datasets`:

- [ ] DICOMs: pydicom
- [x] NIfTIs: nibabel
- [ ] WFDB: wfdb

cc @zaRizk7 for viz

Feel free to comment / suggest other formats and libs you'd like to see or to share your interest in one of the mentioned format

stefanches7:
Please add the support for `Zarr`! That's what we use in the Bioimaging community. It is crucial, because raw upload of a *single* bio image can take _terrabytes in memory_!

The python library would be `bioio` or `zarr`:
- [ ] Zarr: `bioio` or `zarr`

See a [Zarr example](https://ome.github.io/ome-ngff-validator/?source=https://uk1s3.embassy.ebi.ac.uk/bia-integrator-data/S-BIAD845/796b9fb8-f4ec-4c4b-bfc3-5cb00ccf19fe/796b9fb8-f4ec-4c4b-bfc3-5cb00ccf19fe.zarr)

cc @joshmoore

cakiki:
@stefanches7 `zarr` is already usable with the hf hub as an array store. See this example from the [docs](https://huggingface.co/docs/huggingface_hub/en/guides/hf_file_system):

```python
import numpy as np
import zarr

embeddings = np.random.randn(50000, 1000).astype("float32")

# Write an array to a repo
with zarr.open_group("hf://my-username/my-model-repo/array-store", mode="w") as root:
   foo = root.create_group("embeddings")
   foobar = foo.zeros('experiment_0', shape=(50000, 1000), chunks=(10000, 1000), dtype='f4')
   foobar[:] = embeddings

# Read an array from a repo
with zarr.open_group("hf://my-username/my-model-repo/array-store", mode="r") as root:
   first_row = root["embeddings/experiment_0"][0]
```

Is there additional functionality that would not be covered by this?

stefanches7:
@cakiki I think some tiling capabilities, as well as metadata / labels handling. Consult ome-zarr doc here: https://ome-zarr.readthedocs.io/en/stable/python.html
Visualization would be the cherry on the top. 

cc @joshmoore @lubianat @St3V0Bay: curious what you think

cakiki:
zarr-specific dataset viewer would be very cool

bruAristimunha:
[...]

PR #7835, opened 2025-10-28
Add DICOM support

CloseChoice:
supports #7804
Add support for the dicom file format. 

This PR follows PR #7815 and PR #7325 closely.
Remarkable differences:
I made sure that we can load all of pydicom's test data, and encountered the `force=True` parameter that we explicitly support here. This allows to trying to load corrupted dicom files, we explicitly test this!

There is one dataset with all of dicom's test data on huggingface which can be loaded using this branch with the following script:
```python
from datasets import Features, ClassLabel
from datasets.features import Dicom

features = Features({
    "dicom": Dicom(force=True),  # necessary to be able to load one corrupted file
    "label": ClassLabel(num_classes=2)
})

ds = load_dataset("TobiasPitters/dicom-sample-dataset",
                  features=features)

error_count = 0

for idx, item in enumerate(ds["test"]):
    dicom = item["dicom"]

    try:
        print(f"Type: {type(dicom)}")
        if hasattr(dicom, 'PatientID'):
            print(f"PatientID: {dicom.PatientID}")
        if hasattr(dicom, 'StudyInstanceUID'):
            print(f"StudyInstanceUID: {dicom.StudyInstanceUID}")
        if hasattr(dicom, 'Modality'):
            print(f"Modality: {dicom.Modality}")
    except Exception as e:
        error_count += 1
        print(e)

print(f"Finished processing with {error_count} errors.")
```

todo: 
- [x] add docs (will do so soon)


lhoestq:
Awesome ! For the docs should we rename https://huggingface.co/docs/datasets/nifti_dataset to medical_imaging_dataset and have both DICOM and NIfTI together or have separate pages in you opinion ?

CloseChoice:
> Awesome ! For the docs should we rename https://huggingface.co/docs/datasets/nifti_dataset to medical_imaging_dataset and have both DICOM and NIfTI together or have separate pages in you opinion ?

Makes sense, is more intuitive for the user and the pages as proposed in this branch have a lot of overlap. I would then structure it in such a way to write some brief things about medical imaging, then introduce the formats (so basically concatenating the two pages together and removing duplicates).

CloseChoice:
Pls don't merge currently, since we'll need an `embed_storage` function in here as well. See
https://github.com/huggingface/datasets/pull/7815#issuecomment-3494094692 and the following conversation

CloseChoice:
@lhoestq, this is ready for a first round of review.

PR #7840, opened 2025-11-02
Add num channels to audio

### LINKED ISSUE #7837: mono parameter to the Audio feature is missing
According to the docs, there is a "mono" parameter to the Audio feature, which turns any stereo into mono. In practice the signal is not touched and the mono parameter, even though documented, does not exist.

https://github.com/huggingface/datasets/blob/41c05299348a499807432ab476e1cdc4143c8772/src/datasets/features/audio.py#L52C1-L54C22

CloseChoice:
Fixes #7837

We currently have the [mono attribute for Audio documented](https://github.com/huggingface/datasets/blob/41c05299348a499807432ab476e1cdc4143c8772/src/datasets/features/audio.py#L52C1-L54C22) but not used anywhere resulting in confusion for users. Since torchcodec does not know this attribute I suggest using `num_channels` (currently supported `None` (leave unchanged), mono: `1`, stereo: `2`).

I could also add a mono attribute but found that to be more confusing for developers and would restrict us if at any point in the future more than 2 channels are supported.

### DIFF
--- a/src/datasets/features/audio.py
+++ b/src/datasets/features/audio.py
@@ -51,7 +51,11 @@ class Audio:
             Target sampling rate. If `None`, the native sampling rate is used.
         num_channels (`int`, *optional*):
              The desired number of channels of the samples. By default, the number of channels of the source is used.
-             Currently `None` (number of channels of the source), `1` (mono) or `2` (stereo) channels are supported.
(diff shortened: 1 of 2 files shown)


lhoestq (approved):
lgtm !

PR #7854, opened 2025-11-06
[Distributed] split_dataset_by_node() gives the same number of examples for each node

lhoestq:
this works:

```python
import torch.distributed as dist
from datasets import IterableDataset
from collections import Counter

def g(shards):
    for shard in shards:
        # shards don't have the same length
        num_examples = 3 + shard 
        for i in range(num_examples):
            yield {"shard": f"{shard=}", "i": i}

if __name__ == "__main__":
    dist.init_process_group(backend="gloo")
    rank, world_size = dist.get_rank(), dist.get_world_size()
    num_shards = 6
    ds = IterableDataset.from_generator(g, gen_kwargs={"shards": list(range(num_shards))})
    ds = split_dataset_by_node(ds, rank=rank, world_size=world_size)
    # Check that each rank has the same number of examples
    # and show the number of examples per shard and per rank
    counter = Counter(ds["shard"])
    print(f"# {rank=}\ttotal={counter.total()}\t{counter}", flush=True)

    # torchrun --nproc_per_node 2 script.py
    # rank=0        total=16        Counter({'shard=4': 7, 'shard=2': 5, 'shard=0': 4})
    # rank=1        total=16        Counter({'shard=3': 6, 'shard=5': 6, 'shard=1': 4})

```

TODO: make it work with DataLoader (communicate with main process to know when the node runs out of data ?)

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -2110,6 +2110,95 @@ class FormattedExamplesIterable(_BaseExamplesIterable):
         return self.ex_iterable.num_shards
 
 
+class SyncedDistributedExamplesIterable(_BaseExamplesIterable):
+    def __init__(
+        self,
+        ex_iterable: _BaseExamplesIterable,
+        rank: int,
+        world_size: int,
+        stopping_strategy: Literal["first_exhausted", "all_exhausted"] = "first_exhausted",
+    ):
+        super().__init__()
+        self.ex_iterable = ex_iterable
+        self.rank = rank
+        self.world_size = world_size
+        self.stopping_strategy = stopping_strategy
+        # if undersampling ("first_exhausted"), we stop as soon as one dataset is exhausted
+        # if oversampling ("all_exhausted"), we stop as soons as every dataset is exhausted, i.e as soon as every samples of every dataset has been visited at least once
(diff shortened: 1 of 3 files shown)


src/datasets/iterable_dataset.py
```diff
@@ -2110,6 +2110,95 @@ def num_shards(self) -> int:
...
+        import torch.distributed as dist
+
+        is_exhausted = torch.zeros(self.world_size, dtype=torch.bool)
+        while True:
+            for key, example in self.ex_iterable:
+                yield key, example
+                dist.all_reduce(is_exhausted)
+                if self.bool_strategy_func(is_exhausted):
+                    return
+            is_exhausted[self.rank] = True
```
LTMeyer: I'm wondering if this doesn't yield one sample more if the dataset is exhausted.

lhoestq: good point, let me see

lhoestq:
Making this work with multiple workers could create a lot of communication for not a lot of benefits, considering you can simply use `Join()` to let nodes shutdown when they run out of data while the other nodes continue training: https://docs.pytorch.org/docs/stable/distributed.algorithms.join.html

PR #7884, opened 2025-11-27
Fix 7846: add_column and add_item erroneously(?) require new_fingerprint parameter 

sajmaru:
Summary of the change:

Made new_fingerprint optional (Optional[str] = None) in Dataset.add_column and Dataset.add_item
Added a simple test to verify both methods work without providing a fingerprint

Why this change is safe:

The Dataset constructor already handles fingerprint=None by generating a new fingerprint automatically
No internal logic is broken — if a user provides a fingerprint, it’s still used as before
The change only affects the function signature, making it more user-friendly without changing any functionality

### DIFF
--- a/tests/test_arrow_dataset.py
+++ b/tests/test_arrow_dataset.py
@@ -4787,6 +4787,7 @@ def test_polars_round_trip():
 
 def test_add_column():
     from datasets import Dataset
+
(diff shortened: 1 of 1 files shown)


sajmaru:
@lhoestq can you help review this pull request?

PR #7889, opened 2025-11-29
fix(tests): stabilize flaky Hub LFS integration test

The-Obstacle-Is-The-Way:
## Problem

`test_push_dataset_dict_to_hub_overwrite_files` intermittently fails with:
```
BadRequestError: LFS pointer pointed to a file that does not exist
```

This has been causing the `deps-latest` integration tests to fail on main (visible in recent CI runs). I ran into this while working on the BIDS loader PR and dug into the root cause.

## Root Cause

Two race conditions in the test:

1. **LFS propagation timing** - Rapid successive `push_to_hub` calls don't wait for Hub to fully propagate LFS objects between pushes
2. **Repo name reuse** - The second test scenario reused the same repo name from scenario 1, creating a race between deletion and recreation

## Solution

- Add `_wait_for_repo_ready()` helper that polls `list_repo_files` to ensure the repo is consistent before subsequent operations
- Use a unique repo name (`ds_name_2`) for the second scenario, eliminating the delete/create race entirely

## Testing

All 4 integration test variants now pass:
- ✅ `ubuntu-latest, deps-latest` (was failing)
- ✅ `ubuntu-latest, deps-minimum`
- ✅ `windows-latest, deps-latest` (was failing)
- ✅ `windows-latest, deps-minimum`

Validated on fork: https://github.com/The-Obstacle-Is-The-Way/datasets/pull/4

## Related

- #7600 (push_to_hub concurrency)
- #6392 (push_to_hub connection robustness)

cc @lhoestq - small fix but should help CI reliability

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -6005,19 +6005,25 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                             create_pr=create_pr,
                         )
                     except HfHubHTTPError as err:
-                        if (
-                            err.__context__
-                            and isinstance(err.__context__, HfHubHTTPError)
-                            and err.__context__.response.status_code == 409
-                        ):
+                        response = (
+                            err.response if err.response is not None else getattr(err.__context__, "response", None)
+                        )
+                        status_code = response.status_code if response is not None else None
+                        if status_code == 409:
                             # 409 is Conflict (another commit is in progress)
                             time.sleep(sleep_time)
                             logger.info(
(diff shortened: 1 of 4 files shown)


lhoestq:
Maybe `push_to_hub()` can retry on error 400 when committing to the Hub instead ? This way we make sure `push_to_hub()` works without having to add the waiting step

The-Obstacle-Is-The-Way:
Thanks for the pointer @lhoestq! Updated to add LFS 400 retry handling directly in `push_to_hub()` across all implementations (Dataset, DatasetDict, IterableDataset, IterableDatasetDict). Reverted the test-side waits.

PR #7890, opened 2025-11-29
Fix: Auto-create TMPDIR directory when it doesn't exist (Issue #7877)

### LINKED ISSUE #7877: work around `tempfile` silently ignoring `TMPDIR` if the dir doesn't exist
This should help a lot of users running into `No space left on device` while using `datasets`. Normally the issue is is that `/tmp` is too small and the user needs to use another path, which they would normally set as `export TMPDIR=/some/big/storage`

However, the `tempfile` facility that `datasets` and `pyarrow` use is somewhat broken. If the path doesn't exist it'd ignore it and fall back to using `/tmp`. Watch this:

```
$ export TMPDIR='/tmp/username' 

$ python -c "\
import os
import tempfile
print(os.environ['TMPDIR'])
print(tempfile.gettempdir())"
/tmp/username
/tmp
```

Now let's ensure the path exists:
```
$ export TMPDIR='/tmp/username' 
$ mkdir -p $TMPDIR
$ python -c "\
import os
import tempfile
print(os.environ['TMPDIR'])
print(tempfile.gettempdir())"
/tmp/username
/tmp/username
```

So I recommend `datasets` do either of the 2:
1. assert if `$TMPDIR` dir doesn't exist, telling the user to create it
2. auto-create it

The reason for (1) is that I don't know why `tempdir` doesn't auto-create the dir - perhaps some security implication? I will let you guys make the decision, but the key is not to let things silently fall through and the user puzzling why no matter what they do they can't break past `No space left on device` while using `datasets`

Thank you.

I found this via https://stackoverflow.com/questions/37229398/python-tempfile-gettempdir-does-not-respect-tmpdir while trying to help a colleague to solve this exact issue.

ada-ggf25:
# Fix: Auto-create TMPDIR directory when it doesn't exist

Fixes #7877

## Description

This PR fixes issue #7877 by implementing automatic creation of the `TMPDIR` directory when it is set but doesn't exist. Previously, `tempfile.mkdtemp()` would silently ignore `TMPDIR` and fall back to `/tmp` if the specified directory didn't exist, causing confusion for users experiencing "No space left on device" errors.

## Problem

When users set `TMPDIR` to a non-existent directory (e.g., `export TMPDIR=/some/big/storage`), Python's `tempfile` module silently ignores it and falls back to the default temporary directory (`/tmp`). This leads to:

- Users unable to use their specified temporary directory
- Silent failures that are difficult to debug
- Continued "No space left on device" errors even after setting `TMPDIR`

## Solution

The fix automatically creates the `TMPDIR` directory if it is set but doesn't exist, ensuring that:

1. Users' `TMPDIR` settings are respected
2. Clear logging is provided when the directory is created
3. Graceful fallback with warnings if directory creation fails
4. The fix works even if `tempfile.gettempdir()` was already called and cached

### Implementation (`src/datasets/fingerprint.py`)

- Modified `_TempCacheDir.__init__()` to check if `TMPDIR` environment variable is set
- Added logic to auto-create the directory if it doesn't exist using `os.makedirs()`
- Added informative logging when directory is created
- Added warning logging when directory creation fails, with graceful fallback
- Added path normalisation to handle path resolution issues
- Explicitly pass `dir` parameter to `tempfile.mkdtemp()` to ensure TMPDIR is respected

### Tests (`tests/test_fingerprint.py`)

Added comprehensive test coverage:

- `test_temp_cache_dir_with_tmpdir_nonexistent`: Tests auto-creation of non-existent TMPDIR
- `test_temp_cache_dir_with_tmpdir_existing`: Tests behaviour when TMPDIR already exists
- `test_temp_cache_dir_without_tmpdir`: Tests default behaviour when TMPDIR is not set
- `test_temp_cache_dir_tmpdir_creation_failure`: Tests error handling when directory creation fails

Also fixed incomplete `test_fingerprint_in_multiprocessing` test that was missing implementation.

## Testing

- All existing tests pass
- New tests added for TMPDIR handling scenarios
- Code formatted with `make style`
- No linter errors
- Manual testing confirms the fix works as expected

## Example Usage

Before this fix:
```bash
$ export TMPDIR='/tmp/username'  # Directory doesn't exist
$ python -c "import tempfile; print(tempfile.gettempdir())"
/tmp  # Silently falls back, ignoring TMPDIR
```

After this fix:
```bash
$ export TMPDIR='/tmp/username'  # Directory doesn't exist
$ python -c "from datasets.fingerprint import get_temporary_cache_files_directory; print(get_temporary_cache_files_directory())"
# Directory is automatically created and used
# Log: "Created TMPDIR directory: /tmp/username"
```

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

---

**Note**: This implementation follows the approach suggested in the issue, automatically creating the TMPDIR directory when it doesn't exist, which provides the best user experience whilst maintaining security (we only create directories explicitly specified by the user via environment variable).

### DIFF
--- a/src/datasets/fingerprint.py
+++ b/src/datasets/fingerprint.py
@@ -61,18 +61,16 @@ class _TempCacheDir:
                     os.makedirs(tmpdir, exist_ok=True)
                     logger.info(f"Created TMPDIR directory: {tmpdir}")
                 except OSError as e:
-                    logger.warning(
+                    raise OSError(
                         f"TMPDIR is set to '{tmpdir}' but the directory does not exist and could not be created: {e}. "
-                        f"Falling back to default temporary directory."
-                    )
-                    tmpdir = None
+                        "Please create it manually or unset TMPDIR to fall back to the default temporary directory."
+                    ) from e
             # If tmpdir exists, verify it's actually a directory and writable
             elif not os.path.isdir(tmpdir):
-                logger.warning(
+                raise OSError(
                     f"TMPDIR is set to '{tmpdir}' but it is not a directory. "
-                    f"Falling back to default temporary directory."
+                    "Please point TMPDIR to a writable directory or unset it to fall back to the default temporary directory."
                 )
-                tmpdir = None
 
         # Explicitly pass the directory to mkdtemp to ensure TMPDIR is respected
         # This works even if tempfile.gettempdir() was already called and cached


stas00 (commented):
Thank you for implementing this, @ada-ggf25 

I see there are a few more uses of `tempfile` in `datasets` besides the one treated here - should they be treated in the same way as well?

ada-ggf25:
> Thank you for implementing this, @ada-ggf25
> 
> I see there are a few more uses of `tempfile` in `datasets` besides the one treated here - should they be treated in the same way as well?

I had a look through the other tempfile call sites:
- In `arrow_dataset.py` the code always pass an explicit `dir=cache_dir` when using `NamedTemporaryFile`, so TMPDIR is not involved there.
- In `search.py` the `NamedTemporaryFile` is only used to generate a random Elasticsearch index name, it doesn’t control where big cache files are written, so it doesn’t have the same “No space left on device” impact.

Given that, the fingerprint temp directory is the only place where respecting TMPDIR is part of the user‑visible API for disk usage. If you’d like, I’m happy to also thread TMPDIR through the `search.py` name generation, but I think this PR should keep scoped to the place that actually writes large temporary files.

stas00:
Thank you for checking the other instances of tempfile use, @ada-ggf25

Issue #7852, opened 2025-11-06
Problems with NifTI


CloseChoice:
### Describe the bug

There are currently 2 problems with the new NifTI feature:
1. dealing with zipped files, this is mentioned and explained [here](https://github.com/huggingface/datasets/pull/7815#issuecomment-3496199503)
2. when uploading via the `niftifolder` feature, the resulting parquet only contains relative paths to the nifti files:

```bash
table['nifti']
<pyarrow.lib.ChunkedArray object at 0x798245d37d60>
[
  -- is_valid: all not null
  -- child 0 type: binary
    [
      null,
      null,
      null,
      null,
      null,
      null
    ]
  -- child 1 type: string
    [
      "/home/tobias/programming/github/datasets/nifti_extracted/T1.nii",
      "/home/tobias/programming/github/datasets/nifti_extracted/T2-interleaved.nii",
      "/home/tobias/programming/github/datasets/nifti_extracted/T2.nii",
      "/home/tobias/programming/github/datasets/nifti_extracted/T2_-interleaved.nii",
      "/home/tobias/programming/github/datasets/nifti_extracted/T2_.nii",
      "/home/tobias/programming/github/datasets/nifti_extracted/fieldmap.nii"
    ]
]
```
instead of containing bytes. The code is copy pasted from PDF, so I wonder what is going wrong here.

### Steps to reproduce the bug

see the linked comment

### Expected behavior

downloading should work as smoothly as for pdf

### Environment info

- `datasets` version: 4.4.2.dev0
- Platform: Linux-6.14.0-33-generic-x86_64-with-glibc2.39
- Python version: 3.12.3
- `huggingface_hub` version: 0.35.3
- PyArrow version: 21.0.0
- Pandas version: 2.3.3
- `fsspec` version: 2025.9.0

lhoestq:
> 2. when uploading via the niftifolder feature, the resulting parquet only contains relative paths to the nifti files:

what did you use to upload the dataset ? iirc push_to_hub() does upload the bytes as well, but to_parquet() doesn't

CloseChoice:
> > 2. when uploading via the niftifolder feature, the resulting parquet only contains relative paths to the nifti files:
> 
> what did you use to upload the dataset ? iirc push_to_hub() does upload the bytes as well, but to_parquet() doesn't

I used `push_to_hub` but the problem is that the nifti feature does not have an `embed_storage` function

Issue #6790, opened 2024-04-07
PyArrow 'Memory mapping file failed: Cannot allocate memory' bug


lasuomela:
### Describe the bug

Hello,

I've been struggling with a problem using Huggingface datasets caused by PyArrow memory allocation. I finally managed to solve it, and thought to document it since similar issues have been raised here before (https://github.com/huggingface/datasets/issues/5710, https://github.com/huggingface/datasets/issues/6176).

In my case, I was trying to load ~70k dataset files from disk. This triggered an (uninformative) exception around 64k loaded files:
```
  File "pyarrow/io.pxi", line 1053, in pyarrow.lib.memory_map
  File "pyarrow/io.pxi", line 1000, in pyarrow.lib.MemoryMappedFile._open
  File "pyarrow/error.pxi", line 154, in pyarrow.lib.pyarrow_internal_check_status
  File "pyarrow/error.pxi", line 91, in pyarrow.lib.check_status
OSError: Memory mapping file failed: Cannot allocate memory
```
Despite system RAM usage being very low. After a lot of digging around, I discovered that my Ubuntu machine had a limit on the maximum number of memory mapped files in `/proc/sys/vm/max_map_count` set to 65530, which was causing my data loader to crash. Increasing the limit in the file (`echo <new_mmap_size> | sudo tee /proc/sys/vm/max_map_count`) made the issue go away.

While this isn't a bug as such in either Datasets or PyArrow, this behavior can be very confusing to users. Maybe this should be mentioned in documentation? I suspect the other issues raised here about memory mapping OOM errors could actually be consequence of system configuration.

Br,
Lauri

### Steps to reproduce the bug

```
import numpy as np
import pyarrow as pa
import tqdm

# Write some data to disk
arr = pa.array(np.arange(100))
schema = pa.schema([
    pa.field('nums', arr.type)
])
with pa.OSFile('arraydata.arrow', 'wb') as sink:
    with pa.ipc.new_file(sink, schema=schema) as writer:
        batch = pa.record_batch([arr], schema=schema)
        writer.write(batch)

# Number of times to open the memory map
nums = 70000

# Read the data back
arrays = [pa.memory_map('arraydata.arrow', 'r') for _ in tqdm.tqdm(range(nums))]
```

### Expected behavior

No errors.

### Environment info

datasets: 2.18.0
pyarrow: 15.0.0

jxmorris12:
[...]

Issue #6686, opened 2024-02-21
Question: Is there any way for uploading a large image dataset?


zhjohnchan:
I am uploading an image dataset like this:
```
dataset = load_dataset(
    "json",
    data_files={"train": "data/custom_dataset/train.json", "validation": "data/custom_dataset/val.json"},
)
dataset = dataset.cast_column("images", Sequence(Image()))
dataset.push_to_hub("StanfordAIMI/custom_dataset", max_shard_size="1GB")
```
where it takes a long time in the `Map` process. Do you think I can use multi-processing to map all the image data to the memory first? For the `Map()` function, I can set `num_proc`. But for `push_to_hub` and `cast_column`, I can not find it.

Thanks in advance!

Best,

dirtycomputer:
```
import pandas as pd
from datasets import Dataset, Image

# Read the CSV file
data = pd.read_csv("XXXX.csv")

# Create a Hugging Face Dataset
dataset = Dataset.from_pandas(data)
dataset = dataset.cast_column("file_name", Image())

# Upload to Hugging Face Hub (make sure authentication is set up)
dataset.push_to_hub("XXXXX"")
```

stuck in "Casting the dataset
![截屏2024-05-02 11 44 50](https://github.com/huggingface/datasets/assets/48406770/dc012dc5-16f6-4fd5-9e02-1b705c552c5b)
"

PR #7896, opened 2025-12-03
fix: force contiguous copy for sliced list arrays in embed_array_storage

### LINKED ISSUE #7894: embed_table_storage crashes (SIGKILL) on sharded datasets with Sequence() nested types
## Summary

`embed_table_storage` crashes with SIGKILL (exit code 137) when processing sharded datasets containing `Sequence()` nested types like `Sequence(Nifti())`. Likely affects `Sequence(Image())` and `Sequence(Audio())` as well.

The crash occurs at the C++ level with no Python traceback.

### Related Issues

- #7852 - Problems with NifTI (closed, but related embedding issues)
- #6790 - PyArrow 'Memory mapping file failed' (potentially related)
- #7893 - OOM issue (separate bug, but discovered together)

### Context

Discovered while uploading the [Aphasia Recovery Cohort (ARC)](https://openneuro.org/datasets/ds004884) neuroimaging dataset to HuggingFace Hub. Even after fixing the OOM issue (#7893), this crash blocked uploads.

Working implementation with workaround: [arc-aphasia-bids](https://github.com/The-Obstacle-Is-The-Way/arc-aphasia-bids)

## Reproduction

```python
from datasets import Dataset, Features, Sequence, Value
from datasets.features import Nifti
from datasets.table import embed_table_storage

features = Features({
    "id": Value("string"),
    "images": Sequence(Nifti()),
})

ds = Dataset.from_dict({
    "id": ["a", "b"],
    "images": [["/path/to/file.nii.gz"], []],
}).cast(features)

# This works fine:
table = ds._data.table.combine_chunks()
embedded = embed_table_storage(table)  # OK

# This crashes with SIGKILL:
shard = ds.shard(num_shards=2, index=0)
shard_table = shard._data.table.combine_chunks()
embedded = embed_table_storage(shard_table)  # CRASH - no Python traceback
```

## Key Observations

| Scenario | Result |
|----------|--------|
| Single `Nifti()` column | Works |
| `Sequence(Nifti())` on full dataset | Works |
| `Sequence(Nifti())` after `ds.shard()` | **CRASHES** |
| `Sequence(Nifti())` after `ds.select([i])` | **CRASHES** |
| Crash with empty Sequence `[]` | **YES** - not file-size related |

## Workaround

Convert shard to pandas and recreate the Dataset to break internal Arrow references:

```python
shard = ds.shard(num_shards=num_shards, index=i, contiguous=True)

# CRITICAL: Pandas round-trip breaks problematic references
shard_df = shard.to_pandas()
fresh_shard = Dataset.from_pandas(shard_df, preserve_index=False)
fresh_shard = fresh_shard.cast(ds.features)

# Now embedding works
table = fresh_shard._data.table.combine_chunks()
embedded = embed_table_storage(table)  # OK!
```

## Disproven Hypotheses

| Hypothesis | Test | Result |
|------------|------|--------|
| PyArrow 2GB binary limit | Monkey-patched `Nifti.pa_type` to `pa.large_binary()` | Still crashed |
| Memory fragmentation | Called `table.combine_chunks()` | Still crashed |
| File size issue | Tested with tiny NIfTI files | Still crashed |

## Root Cause Hypothesis

When `ds.shard()` or `ds.select()` creates a subset, the resulting Arrow table retains internal references/views to the parent table. When `embed_table_storage` processes nested struct types like `Sequence(Nifti())`, these references cause a crash in the C++ layer.

The pandas round-trip forces a full data copy, breaking these problematic references.

## Environment

- datasets version: main branch (post-0.22.0)
- Platform: macOS 14.x ARM64 (may be platform-specific)
- Python: 3.13
- PyArrow: 18.1.0

## Notes

This may ultimately be a PyArrow issue surfacing through datasets. Happy to help debug further if maintainers can point to where to look in the embedding logic.

The-Obstacle-Is-The-Way:
## Summary

Fixes SIGKILL crash in `embed_array_storage` when processing sliced/sharded datasets with nested types like `Sequence(Nifti())` or `Sequence(Image())`.

**Root cause**: When `ds.shard()` or `ds.select()` creates a sliced view, `array.values` on a sliced `ListArray` returns values with internal offset references. For nested types, PyArrow's C++ layer can crash (SIGKILL, exit code 137) when materializing these sliced nested structs.

**Fix**: Force a contiguous copy via `pa.concat_arrays([array])` when the array has a non-zero offset before processing list/large_list arrays.

## Changes

- Add offset check in `embed_array_storage` for list/large_list arrays
- Force contiguous copy when `array.offset > 0` to break internal references
- Add regression tests for sliced arrays with Image, Nifti, and LargeList types

## Test plan

- [x] Added `tests/features/test_embed_storage_sliced.py` with 3 tests:
  - `test_embed_array_storage_sliced_list_image`
  - `test_embed_array_storage_sliced_list_nifti`
  - `test_embed_array_storage_sliced_large_list`
- [x] All tests verify `embedded.offset == 0` (contiguous result)
- [x] All tests pass locally
- [x] ruff check passes

## Context

This was discovered while uploading a 270GB neuroimaging dataset (ARC) with `Sequence(Nifti())` columns. The process crashed with SIGKILL (no Python traceback) when `embed_table_storage` was called on sharded data.

Workaround that confirmed the fix: pandas round-trip (`shard.to_pandas()` → `Dataset.from_pandas()`) which forces a contiguous copy.

Fixes #7894
Related: #6686, #7852, #6790

### DIFF
--- a/src/datasets/table.py
+++ b/src/datasets/table.py
@@ -2125,7 +2125,6 @@ def embed_array_storage(array: pa.Array, feature: "FeatureType", token_per_repo_
     # When ds.shard() or ds.select() creates a sliced view, array.values returns
     # values with internal offset references that can cause PyArrow's C++ layer
     # to crash when processing nested types like Sequence(Nifti()).
-    # See: https://github.com/huggingface/datasets/issues/XXXX
     if pa.types.is_list(array.type) or pa.types.is_large_list(array.type):
         if hasattr(array, "offset") and array.offset > 0:
             array = pa.concat_arrays([array])
--- a/tests/features/test_embed_storage_sliced.py
+++ b/tests/features/test_embed_storage_sliced.py
@@ -1,7 +1,7 @@
 """Tests for embed_array_storage with sliced/sharded arrays.
 
-Regression tests for https://github.com/huggingface/datasets/issues/XXXX
-(SIGKILL in embed_array_storage when processing sliced/sharded Arrow tables)
+Regression tests for SIGKILL crash when processing sliced/sharded Arrow tables
+with nested types like Sequence(Nifti()) or Sequence(Image()).
 """
 
 import pyarrow as pa
@@ -81,6 +81,8 @@ class TestEmbedArrayStorageSliced:
         # This should NOT crash with SIGKILL
         embedded = embed_array_storage(sliced, List(Nifti()))
 
+        # The fix should make the result contiguous (offset = 0)
+        assert embedded.offset == 0, "Result should be contiguous after fix"
         assert len(embedded) == 2
         # Verify bytes were embedded
         assert embedded[0].as_py()[0]["bytes"] is not None
@@ -111,4 +113,10 @@ class TestEmbedArrayStorageSliced:
         # This should NOT crash with SIGKILL
         embedded = embed_array_storage(sliced, LargeList(Image()))
 
+        # The fix should make the result contiguous (offset = 0)
+        assert embedded.offset == 0, "Result should be contiguous after fix"
         assert len(embedded) == 2
+        # Item 0 of sliced = Item 1 of original (has 2 images)
+        assert len(embedded[0].as_py()) == 2
+        # Verify bytes were embedded
+        assert embedded[0].as_py()[0]["bytes"] is not None


lhoestq:
Hi @The-Obstacle-Is-The-Way, I haven't had a chance to try your repro at https://github.com/huggingface/datasets/issues/7894#issuecomment-3618888430 but I tried the test functions in your PRs and they all pass without your change, is it expected ? (I ran the tests without the checks on the offsets)

The-Obstacle-Is-The-Way:
Hi @lhoestq, you're right - the tests pass without the fix. 

I think the SIGKILL crash is scale-dependent. As I documented in #7894, it only manifests with large, real-world data (the 273GB ARC dataset) and other large dataset upload issues, not with synthetic test files:

| Data | Crash? |
|------|--------|
| Synthetic 64³ NIfTI | ✅ No crash |
| Real ARC (273GB) | ❌ SIGKILL |

The tests verify that the fix's mechanism *works* (sliced arrays get made contiguous), but they can't reproduce the crash because that requires the full-scale data.

I understand this makes it harder to review - the fix is essentially defensive against a crash that's difficult to reproduce in CI. A few options:

1. **Keep as-is** - the fix is low-risk (`pa.concat_arrays` on sliced list arrays) and I've confirmed it resolves the real-world crash
2. **Simplify tests** - remove the offset assertions if they're confusing, keep just the "doesn't crash" aspect
3. **Close** - if the crash can't be reproduced in CI, I understand if this doesn't meet the bar for merging

Happy to do whatever you think is best. Thank you for your time and dedication!

PR #7899, opened 2025-12-08
Add inspect_ai eval logs support

lhoestq:
Support for .eval log files from inspect_ai

They are actually ZIP files according to the source code at https://github.com/UKGovernmentBEIS/inspect_ai/blob/main/src/inspect_ai/log/_log.py

Unfortunately their format can't be converted to Parquet, so I had to JSON-encode all the nested values

```python
ds = load_dataset("dvilasuero/kimi-bfcl")
```

this will enable the Viewer for datasets like https://huggingface.co/datasets/dvilasuero/kimi-bfcl

original tweet for context: https://x.com/dvilasuero/status/1996936988176343220?s=20

cc @dvsrepo @julien-c @davanstrien


davanstrien:
Very cool! 

Any reason not to directly use inspect for loading/converting from the binary format to JSON? https://inspect.aisi.org.uk/reference/inspect_ai.log.html#convert_eval_logs

lhoestq:
The format is simple enough to not have to rely on an additional dependency :)

Issue #7863, opened 2025-11-13
Support hosting iceberg / vortex / lance datasets on huggingface hub


pavanramkumar:
### Feature request

Huggingface datasets has great support for large tabular datasets in parquet with large partitions. I would love to see two things in the future: 

- equivalent support for `lance`, `vortex`, `iceberg`, `zarr` (in that order) in a way that I can stream them using the datasets library
- more fine-grained control of streaming, so that I can stream at the partition / shard level

### Motivation

I work with very large `lance` datasets on S3 and often require random access for AI/ML applications like multi-node training. I was able to achieve high throughput dataloading on a lance dataset with ~150B rows by building distributed dataloaders that can be scaled both vertically (until i/o and CPU are saturated), and then horizontally (to workaround network bottlenecks). 

Using this strategy I was able to achieve 10-20x the throughput of the streaming data loader from the `huggingface/datasets` library.

I realized that these would be great features for huggingface to support natively

### Your contribution

I'm not ready yet to make a PR but open to it with the right pointers!

rjurney:
Kudos!

lhoestq:
So cool ! Would love to see support for lance :)

pavanramkumar:
@lhoestq thanks for your support! Any suggestions across `datasets` or `huggingface_hub` projects to make this happen?

I just noticed this blog post: https://huggingface.co/blog/streaming-datasets

Do you know if `hfFileSystem` from `huggingface_hub` is flexible enough to accommodate lance? I don't want to `open` and scan a file, I want to create generators with the `lance.dataset.to_batches()` from each fragment (partition) that I can iterate over in a distributed dataloader.

Ideally, something like this should just work:

```
import lance
lance_ds_path = f"hf://datasets/{dataset_id}/{path_in_repo}.lance"
ds = lance.dataset(lance_ds_path)
fragments = ds.get_fragments()
fragment_generators = []
for fragment in fragments:
  fragment_generators = fragment.to_batches()
```

Looking at the huggingface blog post, I think we might need a PR into `pyarrow` to create a `LanceFragmentScanOptions` class that subclasses [pyarrow.dataset.FragmentScanOptions](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.FragmentScanOptions.html#pyarrow.dataset.FragmentScanOptions) cc @prrao87, @changhiskhan

lhoestq:
[...]

PR #7913, opened 2025-12-24
Add lance format support

eddyxu:
Add lance format as one of the `packaged_modules`.

```py
import datasets

ds = datasets.load_dataset("org/lance_repo", split="train")

# Or

ds = datasets.load_dataset("./local/data.lance")
```

### DIFF
--- a/src/datasets/packaged_modules/lance/lance.py
+++ b/src/datasets/packaged_modules/lance/lance.py
@@ -118,7 +118,10 @@ class Lance(datasets.ArrowBasedBuilder):
                     )
                 )
             else:
-                lance_files = [lance.file.LanceFileReader(file, storage_options=storage_options) for file in files]
+                lance_files = [
+                    lance.file.LanceFileReader(file, storage_options=storage_options, columns=self.config.columns)
+                    for file in files
+                ]
                 if self.info.features is None:
                     pa_schema = lance_files[0].metadata().schema
                 splits.append(


eddyxu:
Mentioned https://github.com/huggingface/datasets/issues/7863 as well

zhe-thoughts:
@pdames for vis

lhoestq:
Cool ! I notice the current implementation doesn't support streaming because of the symlink hack.

I believe you can do something like this instead:

```python
def _generate_tables(self, paths: list[str]):
    for path in paths:
        ds = lance.dataset(path)
        for frag_idx, fragment in enumerate(ds.get_fragments()):
            for batch_idx, batch in enumerate(
                fragment.to_batches(columns=self.config.columns, batch_size=self.config.batch_size)
            ):
                table = pa.Table.from_batches([batch])
                table = self._cast_table(table)
                yield Key(frag_idx, batch_idx), table
```

note that path can be a local one, but also a `hf://` URI

eddyxu:
@lhoestq Take another look?

lhoestq:
I took the liberty to make a few changes :)

Now I believe we should be good:
- both local and streaming work fine
- both dataset and single files work fine
- all files are properly downloaded now than all files and metadata files are included in config.data_files
- sharding is supported:
  - dataset: one shard = one fragment
  - single files: one shard = one file 
- streaming dataset resuming works fine thanks to Key()
- the two hacks are visible and with TODOs to remove them when possible
  1. remove the revision in HF uris since only "main" is supported
  2. write proper _version/* files since lance doesn't work if they are symlinks

I think this PR is ready, just let me know what you think before we merge 🚀 

The next steps are:
- open a PR in this repository to document Lance support in `datasets`
- open a PR in https://github.com/huggingface/hub-docs to add `pylance` to the list of integrated library on HF, and have some documentation on how to use it with datasets on HF (here is an example [PR](https://github.com/huggingface/hub-docs/pull/1892))
- open a PR in https://github.com/huggingface/huggingface.js to add Lance as a supported dataset library on the HF website (here is an example [PR](https://github.com/huggingface/huggingface.js/pull/1870))

Feel free to start some drafts (I noticed there are great examples in your HF account now !), I'll be happy to review :)

And once Lance is available in huggingface.js and docs are ready we'll be ready to enable the Dataset Viewer and Lance code snippets on HF !

src/datasets/packaged_modules/lance/lance.py
```diff
@@ -0,0 +1,163 @@
...
+        if fragments:
+            for frag_idx, fragment in enumerate(fragments):
+                for batch_idx, batch in enumerate(
+                    fragment.to_batches(columns=self.config.columns, batch_size=self.config.batch_size)
+                ):
+                    table = pa.Table.from_batches([batch])
+                    yield Key(frag_idx, batch_idx), self._cast_table(table)
+        else:
+            for file_idx, lance_file in enumerate(lance_files):
+                for batch_idx, batch in enumerate(lance_file.read_all(batch_size=self.config.batch_size).to_batches()):
```
eddyxu: should we support columns pushdown here?

eddyxu: I see. Lets 🚢 !!

PR #7919, opened 2025-12-28
Fix load_from_disk progress bar with redirected stdout

### LINKED ISSUE #7918: datasets.load_from_disk doesn't show progress bar
### Describe the bug

This is the inverse of the bug at [https://github.com/huggingface/datasets/issues/7030](https://github.com/huggingface/datasets/issues/7030), i.e. that `datasets.load_from_disk(path)` displays no progress bar. My dataset has > 16 files in it.

I am redirecting stdout as I capture the log, could this have something to do with it? All other progress bars work fine though except for HF dataset progress bars.

```py
with contextlib.redirect_stdout(log_file), contextlib.redirect_stderr(log_file):
    datasets.load_from_disk(path)
```

### Expected behavior

The progress bar should show when loading a dataset.

### Environment info

Python 3.13.9
Datasets 4.4.1
macOS Tahoe 26.2

omarfarhoud:
Fixes #7918 

## Problem
When using `load_from_disk()` with `contextlib.redirect_stdout()`, the progress bar was not showing even for datasets with >16 files.

## Root Cause  
The `disable` parameter was set to `None` which triggers TTY auto-detection. This fails when stdout is redirected, causing the progress bar to be hidden.

## Solution
Changed `disable=len(state["_data_files"]) <= 16 or None` to `disable=len(state["_data_files"]) <= 16` to force the progress bar to show for datasets with >16 files, regardless of stdout redirection.

## Testing
Verified that progress bars now appear correctly both with and without stdout redirection for datasets with >16 shards.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -167,6 +167,7 @@ TESTS_REQUIRE = [
     "elasticsearch>=7.17.12,<8.0.0",  # 8.0 asks users to provide hosts or cloud_id when instantiating ElasticSearch(); 7.9.1 has legacy numpy.float_ which was fixed in https://github.com/elastic/elasticsearch-py/pull/2551.
     "faiss-cpu>=1.8.0.post1",  # Pins numpy < 2
     "h5py",
+    "pylance",
     "jax>=0.3.14; sys_platform != 'win32'",
     "jaxlib>=0.3.14; sys_platform != 'win32'",
     "lz4; python_version < '3.14'",  # python 3.14 gives ImportError: cannot import name '_compression' from partially initialized module 'lz4.frame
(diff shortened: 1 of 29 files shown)


lhoestq:
this seems to contradict the comment that says 

> set `disable=None` rather than `disable=False` by default to disable progress bar when no TTY attached

I believe the right approach is to do the same as in https://github.com/huggingface/huggingface_hub/pull/2698

omarfarhoud:
> this seems to contradict the comment that says
> 
> > set `disable=None` rather than `disable=False` by default to disable progress bar when no TTY attached
> 
> I believe the right approach is to do the same as in [huggingface/huggingface_hub#2698](https://github.com/huggingface/huggingface_hub/pull/2698)

Updated to check TQDM_POSITION=-1 to force-enable progress bars in cloud environments, 
following the same pattern as huggingface_hub#2698.

omarfarhoud:
Moved the TQDM_POSITION check to the tqdm class in utils/tqdm.py so all progress bars 
in the codebase have consistent behavior. Thanks for the suggestion!

omarfarhoud:
@lhoestq thanks again for the suggestion. I’ve applied it and everything should now be consistent across all tqdm usage. Happy to adjust anything else if needed.

PR #7923, opened 2025-12-31
feat(fasta): add lightweight FASTA file format support

behroozazarkhalili:
## Summary

This PR adds support for loading FASTA files directly with `load_dataset()`, addressing feedback from #7851.

FASTA is a text-based format for representing nucleotide sequences (DNA/RNA) or peptide sequences (proteins), widely used in bioinformatics.

## Key Features

- **Zero external dependencies** - Uses a lightweight pure Python parser based on [readfq.py](https://github.com/lh3/readfq) by Heng Li
- **Streaming support** - Generator-based parsing for memory efficiency with large genomic files
- **Compression support** - Automatic detection and handling of gzip, bzip2, and xz compressed files via magic bytes
- **Large sequence support** - Uses `large_string` Arrow type to handle viral genomes and long sequences (fixes UTF-8 overflow)
- **Adaptive batching** - `max_batch_bytes` parameter (default 256MB) prevents Parquet page size errors with very large sequences

## Technical Decisions (Addressing #7851 Feedback)

| Concern | Solution |
|---------|----------|
| Long sequences → UTF-8 overflow (@apcamargo, @UriNeri) | Uses `pa.large_string()` for sequence column |
| BioPython is overkill (@apcamargo) | Pure Python parser based on Heng Li's readfq.py |
| Parquet page size limit i32::MAX (@UriNeri) | Adaptive dual-threshold batching with `max_batch_bytes` |

## Columns

| Column | Type | Description |
|--------|------|-------------|
| `id` | string | Sequence identifier (first word after `>`) |
| `description` | string | Full description line (everything after id) |
| `sequence` | large_string | The biological sequence (DNA/RNA/protein) |

## Supported Extensions

`.fa`, `.fasta`, `.fna`, `.ffn`, `.faa`, `.frn` (and compressed variants)

## Usage

```python

# Load FASTA file
dataset = load_dataset("fasta", data_files="sequences.fasta")

# Load with column filtering
dataset = load_dataset("fasta", data_files="sequences.fa", columns=["id", "sequence"])

# Load gzipped file
dataset = load_dataset("fasta", data_files="sequences.fa.gz")

# Configure batching for very large genomes
dataset = load_dataset("fasta", data_files="genome.fasta", max_batch_bytes=128*1024*1024)
```

## Test Plan

- [x] Basic FASTA loading (3 sequences, multi-line)
- [x] Multiple extension support (.fa, .fasta, .fna, .ffn, .faa, .frn)
- [x] Compression formats (gzip, bz2, xz)
- [x] Long sequences with `large_string` type
- [x] Column filtering
- [x] Batch size configuration
- [x] Byte-based batching (`max_batch_bytes`)
- [x] Large genome handling (simulated 50KB sequences)
- [x] Empty description handling
- [x] Multiple files loading
- [x] Custom feature casting

All 22 tests passing.

cc: @georgia-hf

### DIFF
--- a/docs/source/loading.mdx
+++ b/docs/source/loading.mdx
@@ -267,6 +267,30 @@ To load remote WebDatasets via HTTP, pass the URLs instead:
 >>> dataset = load_dataset("webdataset", data_files={"train": urls}, split="train", streaming=True)
 ```
 
+### FASTA
+
+[FASTA](https://en.wikipedia.org/wiki/FASTA_format) is a text-based format for representing nucleotide sequences (DNA/RNA) or peptide sequences (proteins), widely used in bioinformatics.
+
+To load a FASTA file:
+
+```py
+>>> from datasets import load_dataset
+>>> dataset = load_dataset("fasta", data_files="sequences.fasta")
+```
+
+This returns a dataset with the following columns:
+- `id`: Sequence identifier (first word after `>`)
+- `description`: Full description line (everything after the id)
+- `sequence`: The biological sequence (DNA/RNA/protein)
+
+You can also load only specific columns:
+
+```py
+>>> dataset = load_dataset("fasta", data_files="sequences.fa", columns=["id", "sequence"])
+```
+
+FASTA files with various extensions are supported: `.fa`, `.fasta`, `.fna` (nucleic acid), `.ffn` (nucleotide of gene regions), `.faa` (amino acid), and `.frn` (non-coding RNA). Compressed files (gzip, bzip2, xz) are automatically detected and decompressed.
+
 ## Remote files
 
 If you have remote files likely stored as a `csv`, `json`, `txt`, `parquet` or any supported format, the [`load_dataset`] function can load load them if you specify their remote paths:
--- a/docs/source/loading.mdx
+++ b/docs/source/loading.mdx
@@ -292,6 +292,30 @@ To load remote WebDatasets via HTTP, pass the URLs instead:
 >>> dataset = load_dataset("webdataset", data_files={"train": urls}, split="train", streaming=True)
 ```
 
+### FASTA
+
+[FASTA](https://en.wikipedia.org/wiki/FASTA_format) is a text-based format for representing nucleotide sequences (DNA/RNA) or peptide sequences (proteins), widely used in bioinformatics.
+
+To load a FASTA file:
+
+```py
+>>> from datasets import load_dataset
+>>> dataset = load_dataset("fasta", data_files="sequences.fasta")
+```
+
+This returns a dataset with the following columns:
+- `id`: Sequence identifier (first word after `>`)
+- `description`: Full description line (everything after the id)
+- `sequence`: The biological sequence (DNA/RNA/protein)
+
+You can also load only specific columns:
+
+```py
+>>> dataset = load_dataset("fasta", data_files="sequences.fa", columns=["id", "sequence"])
+```
+
+FASTA files with various extensions are supported: `.fa`, `.fasta`, `.fna` (nucleic acid), `.ffn` (nucleotide of gene regions), `.faa` (amino acid), and `.frn` (non-coding RNA). Compressed files (gzip, bzip2, xz) are automatically detected and decompressed.
+
 ## Remote files
 
 If you have remote files likely stored as a `csv`, `json`, `txt`, `parquet` or any supported format, the [`load_dataset`] function can load load them if you specify their remote paths:
--- a/docs/source/loading.mdx
+++ b/docs/source/loading.mdx
@@ -307,6 +307,9 @@ This returns a dataset with the following columns:
 - `id`: Sequence identifier (first word after `>`)
 - `description`: Full description line (everything after the id)
 - `sequence`: The biological sequence (DNA/RNA/protein)
+- `record`: A [`BioSequence`] with `format="fasta"` storing each record's original bytes, decoded to a Biopython `SeqRecord` when accessed.
+
+Reading the parsed columns requires no Biopython; exclude `record` with `columns`, or use `dataset.cast_column("record", BioSequence(decode=False))` (imported from `datasets`) to access its `{"bytes": ..., "path": None}` storage without decoding.
 
 You can also load only specific columns:
 
--- a/src/datasets/packaged_modules/__init__.py
+++ b/src/datasets/packaged_modules/__init__.py
@@ -10,6 +10,7 @@ from .cache import cache
 from .conll import conll
 from .csv import csv
 from .eval import eval
+from .fasta import fasta
 from .hdf5 import hdf5
 from .iceberg import iceberg
 from .imagefolder import imagefolder
@@ -44,6 +45,7 @@ def _hash_python_lines(lines: list[str]) -> str:
 # get importable module names and hash for caching
 _PACKAGED_DATASETS_MODULES = {
     "csv": (csv.__name__, _hash_python_lines(inspect.getsource(csv).splitlines())),
+    "fasta": (fasta.__name__, _hash_python_lines(inspect.getsource(fasta).splitlines())),
     "json": (json.__name__, _hash_python_lines(inspect.getsource(json).splitlines())),
     "pandas": (pandas.__name__, _hash_python_lines(inspect.getsource(pandas).splitlines())),
     "parquet": (parquet.__name__, _hash_python_lines(inspect.getsource(parquet).splitlines())),
@@ -81,6 +83,13 @@ _PACKAGED_DATASETS_MODULES_2_15_HASHES = {
 _EXTENSION_TO_MODULE: dict[str, tuple[str, dict]] = {
     ".csv": ("csv", {}),
     ".tsv": ("csv", {"sep": "\t"}),
+    # FASTA biological sequence formats
+    ".fa": ("fasta", {}),
+    ".fasta": ("fasta", {}),
+    ".fna": ("fasta", {}),  # FASTA nucleic acid
+    ".ffn": ("fasta", {}),  # FASTA nucleotide of gene regions
+    ".faa": ("fasta", {}),  # FASTA amino acid
+    ".frn": ("fasta", {}),  # FASTA non-coding RNA
     ".json": ("json", {}),
     ".jsonl": ("json", {}),
     # ndjson is no longer maintained (see: https://github.com/ndjson/ndjson-spec/issues/35#issuecomment-1285673417)
--- a/src/datasets/packaged_modules/__init__.py
+++ b/src/datasets/packaged_modules/__init__.py
@@ -90,6 +90,7 @@ _EXTENSION_TO_MODULE: dict[str, tuple[str, dict]] = {
     ".ffn": ("fasta", {}),  # FASTA nucleotide of gene regions
     ".faa": ("fasta", {}),  # FASTA amino acid
     ".frn": ("fasta", {}),  # FASTA non-coding RNA
+    ".afa": ("fasta", {}),  # aligned FASTA (multiple sequence alignment)
     ".json": ("json", {}),
     ".jsonl": ("json", {}),
     # ndjson is no longer maintained (see: https://github.com/ndjson/ndjson-spec/issues/35#issuecomment-1285673417)
--- a/src/datasets/packaged_modules/__init__.py
+++ b/src/datasets/packaged_modules/__init__.py
@@ -10,6 +10,7 @@ from .cache import cache
 from .conll import conll
 from .csv import csv
 from .eval import eval
+from .fasta import fasta
 from .hdf5 import hdf5
 from .iceberg import iceberg
 from .imagefolder import imagefolder
@@ -45,6 +46,7 @@ def _hash_python_lines(lines: list[str]) -> str:
 # get importable module names and hash for caching
 _PACKAGED_DATASETS_MODULES = {
     "csv": (csv.__name__, _hash_python_lines(inspect.getsource(csv).splitlines())),
+    "fasta": (fasta.__name__, _hash_python_lines(inspect.getsource(fasta).splitlines())),
     "json": (json.__name__, _hash_python_lines(inspect.getsource(json).splitlines())),
     "pandas": (pandas.__name__, _hash_python_lines(inspect.getsource(pandas).splitlines())),
     "parquet": (parquet.__name__, _hash_python_lines(inspect.getsource(parquet).splitlines())),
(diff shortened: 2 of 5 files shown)


lhoestq:
As fo the other PR I think we can have biopython as an optional dependency for the parsing

Also I mentioned in https://github.com/huggingface/datasets/pull/7851 that maybe it would be nice to have a dedicated type for this kind of data and decode the data as `SeqRecord` objects from biopython, which are nicer than simple python dictionaries for practitioners. wdyt ?

apcamargo:
Is there any advantage in using biopython for parsing? It tends to be much slower than Heng Li's native parser (on top of adding an extra dependency)

lhoestq:
Mostly convenience for users, since they get python objects they are used to.
And since the biopython open source community seems to be growing it will also provide access to future developments / features.

But let me know if you think otherwise

behroozazarkhalili:
Thanks! I agree with decoding to BioPython objects (`SeqRecord` here) rather than plain dicts — it gives practitioners something they can process/convert directly, and aligns with where the community already contributes.

Proposed approach for this PR (and the sibling FASTQ/PDB/mmCIF ones):
- Add **`biopython` as an optional dependency**. Parsing/decoding returns `SeqRecord` objects when it's installed.
- Keep a lightweight fallback (the existing zero-dependency parser) for when `biopython` isn't present, so the base install stays lean — this addresses @apcamargo's and @UriNeri's point on #7851 that BioPython is heavy/slow for plain FASTA reads.
- On the **dedicated feature type**: I'd be glad to add one (decoding to `SeqRecord`) — should it live alongside the existing decodable features (like `Audio`/`Image`), or would you prefer to start with `biopython` optional + objects and add the dedicated type as a follow-up?

Also noted your suggestion to **merge #7923 + #7924** since FASTA/FASTQ are so similar — happy to consolidate them into one PR if you prefer. Will hold the PDB/mmCIF "one row = one structure" restructuring for those PRs specifically.

apcamargo:
Two additional suggestions:

- I'm not a big fan on requiring specific extensions. Is this needed? If so, you should include `.afa`, which is a common extension for alignments.
- Support for ZSTD (which is becoming very popular for FASTA files) can be added when the user has Python 3.14 or higher. You can find an example here: https://gist.github.com/apcamargo/d039aa04a2cbbcbb14e2d34a0963b862#file-fancy_fasta_reader-py-L7-L35

behroozazarkhalili:
Thanks @apcamargo, both good points.

**Extensions:** I'll add `.afa`. On dropping the allowlist entirely — the extension list also feeds the packaged-module auto-detection (how `load_dataset` picks the FASTA builder for a folder of files), so I can't remove it outright without breaking that inference. What I can do is relax it (broaden the set, and not hard-fail on unlisted extensions when the builder is selected explicitly). This also ties into the open question above re: decoding via BioPython — if we go that route, format detection leans on content/handle rather than the filename, which makes the extension list less load-bearing. I'll align the final behavior with whatever we land on there.

**ZSTD:** Agreed — it's increasingly common for FASTA. Python 3.14 added `compression.zstd` to the stdlib, so I can add it as a version-gated branch in `_open_file` (alongside the existing gzip/bz2/xz magic-byte detection) without a new dependency. Thanks for the gist reference.

lhoestq:
Your proposed approach looks good !

> Add biopython as an optional dependency. Parsing/decoding returns SeqRecord objects when it's installed.

yes sound sgood

> Keep a lightweight fallback (the existing zero-dependency parser) for when biopython isn't present, so the base install stays lean — this addresses @apcamargo's and @UriNeri's point on https://github.com/huggingface/datasets/pull/7851 that BioPython is heavy/slow for plain FASTA reads.
> On the dedicated feature type: I'd be glad to add one (decoding to SeqRecord) — should it live alongside the existing decodable features (like Audio/Image), or would you prefer to start with biopython optional + objects and add the dedicated type as a follow-up?

Users can disable decoding anyways, and in this case they get the raw files bytes that they can parse themselves. Let's use biopython and have the dedicated type

behroozazarkhalili:
Great — I'll start implementing the biopython-optional + dedicated-type approach across the bio-format PRs. Three things I'd like your call on so I name the public API right the first time:

1. **Feature-type naming.** I'm planning a decodable feature (like `Audio`/`Image`) that stores raw bytes and decodes to a biopython object when biopython is installed, with `decode=False` giving raw bytes. Since FASTA/FASTQ decode to `Bio.SeqRecord` and PDB/mmCIF decode to `Bio.PDB.Structure`, do you prefer (a) two types, e.g. `BioSequence` (SeqRecord) and `Structure` (Bio.PDB), or (b) one general `Biopython`/`SeqRecord`-style type parameterized by format? I'll follow whichever you prefer.

2. **Merge #7923 + #7924?** You mentioned FASTA and FASTQ are similar enough to be one PR — happy to consolidate them into #7923 and close #7924, or keep them separate. Your call.

3. **Where should the feature type live** — added to core `src/datasets/features/` now (registered in `features/__init__.py`), or start with biopython-optional decoding inside the packaged modules and add the dedicated feature type as a follow-up PR?

I'll wire the shared foundation (optional `biopython` dep + availability flag) meanwhile since that's name-independent.

behroozazarkhalili:
Pushed `6d3e5d3a5`. On ZSTD: rather than adding another codec branch, I removed the magic-byte `_open_file` entirely and let the loader use the plain `open()` (which becomes `xopen` under the streaming patch). `datasets`' extraction layer already recognises zstd (magic `28B52FFD` → `ZstdExtractor`), so zstd support comes through that path for free — verified end-to-end with a `.fasta.zst` file. This also matches @lhoestq's note on the GenBank PR that `download_and_extract` already handles decompression.

I went with the existing `zstandard`-based extractor rather than the stdlib `compression.zstd` on 3.14, since `datasets` already standardises on `zstandard` library-wide and this needs no version gate. Compressed test fixtures now use the chained `<protocol>://<inner>::<outer>` URI, with the protocol derived from `_get_extraction_protocol`. 24 tests pass, gzip/bz2/xz coverage unchanged, plus a new zstd test guarded by `require_zstandard`.

behroozazarkhalili:
Two corrections to my last comment, and some measurements.

**I was wrong about zstd being free.** I said `datasets` standardizes on `zstandard` library-wide. It doesn't. `zstandard` appears only in `TESTS_REQUIRE`, never in `REQUIRED_PKGS`, and without it installed a `.fasta.zst` fails with `ValueError: Compression type zstd not supported`. Protocol detection still returns `zstd`, but fsspec cannot service it. My end-to-end check passed only because I had installed the package.

**@apcamargo was right about 3.14.** I checked against CPython 3.14.0rc3 and `compression.zstd` is there, with `ZstdFile` and `open`. The two paths are complementary. Today zstd works on any supported Python through the optional `zstandard` package via the shared extraction layer, so the loader still needs no codec branch of its own. On 3.14+ the stdlib could serve the same bytes with no third-party dependency. I am happy to add the stdlib path behind a version check if you want it.

Removing `_open_file` does not depend on either point. It rests on `download_and_extract` already decompressing, which is what the GenBank thread established.

**biopython vs the native parser.** @lhoestq, you asked for pushback on this and nobody had measured it. biopython is 12.4 MB installed. Parsing with `Bio.SeqIO` against the readfq-based parser, median of 5 runs, biopython is 2.20x slower on 20,000 records of 300 bp (6.0 MB) and 2.14x slower on 5,000 records of 5,000 bp (23.9 MB). The cost is consistent at roughly 2x. That supports the design you approved, with biopython optional for `SeqRecord` decoding and readfq kept as the fallback so users who only want sequences skip both the install and the slowdown.

**Merging #7923 and #7924.** After normalizing the format names, the two modules share 188 of 233 lines, or 81%. `fastq.py` also still carries the `_open_file` shim I removed here. Folding FASTQ into this PR would fix the duplication and the shim together. Want me to do that and close #7924?

behroozazarkhalili:
`.afa` is in as of `765895dc8`.

On whether requiring specific extensions is needed at all, it turns out the loader never enforces them. `Fasta.EXTENSIONS` is declared but nothing reads it, so an unlisted extension does not hard-fail. The only place extensions matter is `_EXTENSION_TO_MODULE` in `packaged_modules/__init__.py`, which drives auto-detection when you point `load_dataset` at a directory without naming a builder.

Verified both halves against a FASTA file named `sequences.weirdext`:

```python
# builder named explicitly -> loads, extension irrelevant
load_dataset("fasta", data_files="sequences.weirdext", split="train")   # 2 records, ids ['s1', 's2']

# same file, auto-detection -> DataFilesNotFoundError
load_dataset("/that/directory", split="train")
```

So the allowlist only governs the case where you point at a directory and do not say which loader to use. It never blocks an explicit request. If you want it broader anyway, adding entries to that map is cheap. Otherwise I would leave it as is, since the same map is what stops a directory of mixed files from being read as FASTA by accident.

behroozazarkhalili:
@lhoestq one open question is blocking the feature-type work on this PR and on #7924, #7925, #7926, so I want to reduce it to a yes or no rather than leave it sitting.

You approved biopython plus a dedicated decodable type on 2026-07-01. What is still undecided is how many types and what they are called, which I asked on 2026-07-03. Since a feature class has to be registered by name in `features/__init__.py`, I cannot write the storage, the decode contract, or the docs without committing to that name.

Here is what I propose to build, following the shape the existing decodable types already have. Every one of `Audio`, `Image`, `Video`, `Pdf`, `Nifti` and `Mesh` is a single class in its own `src/datasets/features/<name>.py`, and each one covers exactly one decoded object type. `Mesh` is the closest thing to an exception and it unions `trimesh.Trimesh` with `trimesh.Scene`, which are two spellings of one concept rather than two concepts.

`Bio.SeqRecord.SeqRecord` and `Bio.PDB.Structure.Structure` are two concepts. They come from different biopython subpackages, so they want different optional imports, and no caller wants a type whose return depends on which file it was handed. So, two types:

| Type | Module | `decode_example` returns | Used by |
|---|---|---|---|
| `BioSequence` | `features/bio_sequence.py` | `Bio.SeqRecord.SeqRecord` | FASTA #7923, FASTQ #7924, GenBank #7951 |
| `BioStructure` | `features/bio_structure.py` | `Bio.PDB.Structure.Structure` | mmCIF #7925, PDB #7926 |

Both store `struct<bytes: binary, path: string>` and both raise under `decode=False`, matching `Audio` and `Image`, so a user without biopython gets raw bytes and parses them however they like.

I will start on that on Monday 2026-08-31 unless you tell me otherwise, so a silent thread stops costing four PRs. If you would rather have one parameterized type, or different names, say the word and I will switch before anything is registered.

Two smaller decisions from the same comment, both still open and neither blocking:

1. Should the type live in core `src/datasets/features/` now, or should decoding start inside the packaged modules with the feature type as a follow-up PR? I would rather put it in core now, because the packaged-module version would be thrown away.
2. Merge #7924 into this PR and close it? After normalizing format names the two modules share 188 of 233 lines. I am happy either way.

behroozazarkhalili:
Separate from the naming question above: none of the five bio-format PRs have ever run CI, so there is no test evidence attached to any of them for a reviewer to look at.

To be precise about what "no CI" means here, since the PR header reads the same either way. The GitHub checks API returns `total_count: 0` for all five, and the combined status is `pending` with zero statuses. That is not a failing suite, it is a suite that has not started.

| PR | branch | mergeable | check runs |
|---|---|---|---|
| #7923 | `feat/fasta-support` | true | 0 |
| #7924 | `feat/fastq-support` | true | 0 |
| #7925 | `feat/mmcif-support` | true | 0 |
| #7926 | `feat/pdb-support` | true | 0 |
| #7951 | `feat/genbank-support` | true | 0 |

`ci.yml` triggers on `pull_request` against `main`, which all five target, so the workflow config looks right and the runs appear to be waiting on approval. A sample of twelve recently opened PRs shows the same pattern: the maintainer-authored one has runs, one outside-contributor PR has runs, and the remaining ten have none, including several from the same author as the one that does. So this reads as the outside-contributor approval gate rather than anything specific to these branches.

Could you approve the workflow runs when you get a chance? All five branches are currently even with `main` at `cd6a34065` and report `mergeable: true`, so they should run clean, and I would rather find out from CI than have you take my word for the local runs.

If it makes the runs cheaper, I can rebase or split anything first.

lhoestq:
Yes your suggested two types look good to me :)

I also ran the CI

behroozazarkhalili:
Thanks, that settles the naming. I opened #8542 with both types against `main`, since you had asked whether they should live in core `features/` now or come as a follow-up, and building them in core avoided writing a packaged-module version that would be thrown away.

#8542 is the foundation the five format PRs sit on: `BioSequence` for FASTA, FASTQ and GenBank, `BioStructure` for mmCIF and PDB. It has 22 tests of its own and does not touch any loader, so it can go in independently of the five.

Two follow-ups.

**CI on the other four.** The run you kicked off covers #7923 only, which is now green at 14 of 14 with `mergeable_state: clean`. #7924, #7925, #7926 and #7951 still report zero check runs, so approving those four would give you test evidence on all of them. All four are even with `main` at `cd6a34065`.

**Merging #7924 into #7923.** You raised this in June and I would like to go ahead with it. After normalizing the format names the two modules share 188 of 233 lines, so keeping them apart means maintaining the same reader twice, and `fastq.py` still carries a `_open_file` shim I already removed here. My plan is to fold FASTQ into this PR, keep both extension sets in `_EXTENSION_TO_MODULE`, and close #7924 pointing at this one. Say if you would rather keep them separate and I will leave both as they are.

I will start wiring the loaders to `BioSequence` once #8542 has had a look, so the decode path lands on top of an agreed type rather than alongside it.

behroozazarkhalili:
Follow-up after a second review pass (own probes plus a cross-model adversarial review; findings reproduced by execution first). Head is f61aee3f6.

- Whitespace inside a sequence line was kept as sequence (`"A C-G*"` stayed as is). Biopython's `SimpleFastaParser` removes it, and this parser now does too.
- Lines starting with `;` are comments in Pearson FASTA and were appended to the sequence; they are skipped.
- `columns=[]` passed validation and failed later with `IndexError`; it raises `ValueError` up front. `columns=` now projects a user-supplied `features` schema.

Tests: 27 (1 skipped), ruff clean.

behroozazarkhalili:
CI note: the full upstream workflow (ubuntu and windows, py3.10 to 3.14, minimum and latest deps) passed on this exact head (f61aee3f6) on my fork: https://github.com/behroozazarkhalili/datasets/actions/runs/33910055715. Approving the workflow run here should be a formality.

behroozazarkhalili:
Adapted to the merged `BioSequence` type as suggested in #8542, and rebased on main (head 6ffb7ea83).

Each row now has a `record` column of type `BioSequence(format="fasta")` holding that record's bytes exactly as they appear in the file, next to the parsed columns, which still need no dependency. With biopython installed `record` decodes to a `SeqRecord`; without it you can drop the column through `columns` or cast it with `decode=False`. Column selection is now validated once at construction (a user-supplied `features` schema selects its own columns, duplicates are rejected, a caller-supplied `info` is kept), and the test fixtures write exact bytes so the Windows lane compares the same record bytes as Linux.
Two behaviours worth knowing, neither changed here: Biopython's plain `fasta` parser keeps Pearson `;` comment lines inside the sequence while the parsed `sequence` column drops them, and bare-CR line endings decode wrongly because `BioSequence.decode_example` reads the bytes without universal newlines. I will open a small follow-up PR for the second one.

Full upstream workflow on this head on my fork: https://github.com/behroozazarkhalili/datasets/actions/runs/34526782812

lhoestq (approved):
lgtm !

PR #7924, opened 2025-12-31
Add lightweight FASTQ file format support

behroozazarkhalili:
## Summary

This PR adds support for loading FASTQ files directly with `load_dataset()`.

FASTQ is an extension of FASTA that includes quality scores for each base, widely used for storing output from high-throughput sequencing instruments.

### Key Features
- **Zero external dependencies** - Pure Python parser based on [readfq.py](https://github.com/lh3/readfq) by Heng Li
- **Quality score support** - Preserves per-base quality scores as ASCII-encoded strings
- **Streaming support** - Generator-based parsing for memory efficiency with large NGS files
- **Compression support** - Automatic detection of gzip, bzip2, and xz compressed files
- **Large sequence support** - Uses `large_string` for both sequence and quality columns
- **Parquet-safe batching** - Dual-threshold batching (batch_size + max_batch_bytes) prevents page size errors

### Columns
| Column | Type | Description |
|--------|------|-------------|
| `id` | string | Sequence identifier (first word after `@`) |
| `description` | string | Full description line (everything after id) |
| `sequence` | large_string | The nucleotide sequence |
| `quality` | large_string | ASCII-encoded quality scores (Phred+33 by default) |

### Supported Extensions
`.fq`, `.fastq` (and compressed variants: `.fq.gz`, `.fastq.gz`, `.fq.bz2`, `.fq.xz`)

### Usage
```python

# Load FASTQ file
dataset = load_dataset("fastq", data_files="reads.fastq")

# Load gzipped file
dataset = load_dataset("fastq", data_files="reads.fq.gz")

# Filter columns
dataset = load_dataset("fastq", data_files="reads.fq", columns=["sequence", "quality"])
```

### Quality Score Format
Quality scores use Sanger/Illumina 1.8+ encoding (Phred+33):
- ASCII character `\!` (33) = quality 0
- ASCII character `I` (73) = quality 40

### Testing
- 22 comprehensive tests covering basic loading, multi-line sequences, compression, batching, schema types, and edge cases
- All tests passing
- Linting clean

### References
- Follows pattern established in #7923 (FASTA support)
- Parser based on: https://github.com/lh3/readfq
- Addresses feedback from #7851

cc: @georgia-hf

### DIFF
--- a/docs/source/loading.mdx
+++ b/docs/source/loading.mdx
@@ -267,6 +267,34 @@ To load remote WebDatasets via HTTP, pass the URLs instead:
 >>> dataset = load_dataset("webdataset", data_files={"train": urls}, split="train", streaming=True)
 ```
 
+### FASTQ
+
+[FASTQ](https://en.wikipedia.org/wiki/FASTQ_format) is a text-based format for storing nucleotide sequences together with their quality scores, widely used for high-throughput sequencing data.
+
+Load a FASTQ file:
+
+```py
+>>> from datasets import load_dataset
+>>> dataset = load_dataset("fastq", data_files="reads.fq")
+```
+
+The loader returns four columns:
+
+| Column | Type | Description |
+|--------|------|-------------|
+| `id` | string | Sequence identifier (first word after `@`) |
+| `description` | string | Full description line (everything after id) |
+| `sequence` | large_string | The nucleotide sequence |
+| `quality` | large_string | ASCII-encoded quality scores |
+
+Load only specific columns:
+
+```py
+>>> dataset = load_dataset("fastq", data_files="reads.fq", columns=["sequence", "quality"])
+```
+
+Compressed files (.fq.gz, .fastq.gz, .fq.bz2, .fq.xz) are automatically detected and decompressed.
+
 ## Remote files
 
 If you have remote files likely stored as a `csv`, `json`, `txt`, `parquet` or any supported format, the [`load_dataset`] function can load load them if you specify their remote paths:
--- a/docs/source/loading.mdx
+++ b/docs/source/loading.mdx
@@ -292,6 +292,34 @@ To load remote WebDatasets via HTTP, pass the URLs instead:
 >>> dataset = load_dataset("webdataset", data_files={"train": urls}, split="train", streaming=True)
 ```
 
+### FASTQ
+
+[FASTQ](https://en.wikipedia.org/wiki/FASTQ_format) is a text-based format for storing nucleotide sequences together with their quality scores, widely used for high-throughput sequencing data.
+
+Load a FASTQ file:
+
+```py
+>>> from datasets import load_dataset
+>>> dataset = load_dataset("fastq", data_files="reads.fq")
+```
+
+The loader returns four columns:
+
+| Column | Type | Description |
+|--------|------|-------------|
+| `id` | string | Sequence identifier (first word after `@`) |
+| `description` | string | Full description line (everything after id) |
+| `sequence` | large_string | The nucleotide sequence |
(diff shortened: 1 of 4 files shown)


lhoestq:
same comment as in https://github.com/huggingface/datasets/pull/7923

> As fo the other PR I think we can have biopython as an optional dependency for the parsing
> 
> Also I mentioned in https://github.com/huggingface/datasets/pull/7851 that maybe it would be nice to have a dedicated type for this kind of data and decode the data as `SeqRecord` objects from biopython, which are nicer than simple python dictionaries for practitioners. wdyt ?

(btw since the two formats are similar this could also be one single PR)

behroozazarkhalili:
Pushed `bb43d20`. Removed the magic-byte `_open_file` here for the same reason as the GenBank thread: the loader calls `download_and_extract()` with `extract_on_the_fly`, so files are already decompressed in the real load path. The loader now uses the plain `open()`, which becomes `xopen` under the streaming patch, matching the text and json packaged modules.

The gzip test previously fed a `.fq.gz` path straight to `_generate_tables`, bypassing `dl_manager`. It now builds the chained compression URI (`<protocol>://<inner>::<outer>`), with the protocol read from `_get_extraction_protocol` rather than a hardcoded table, so the test tracks the loader's real behavior. 22 tests pass.

@lhoestq, on your suggestion that FASTA and FASTQ could be one PR: after normalizing the format names the two modules share 188 of 233 lines, or 81%. Happy to fold this into #7923 and close this one if you prefer.

behroozazarkhalili:
Follow-up after a second review pass (own probes plus a cross-model adversarial review; each finding reproduced by execution first). Head is f933c6180.

The readfq-style parser accepted corrupt input silently: a truncated file yielded a final record with a short quality string, an over-long quality line passed through, and a file with no `+` line was yielded as one long sequence with an empty quality. All three reached the Arrow table as misaligned columns. The parser now raises `ValueError` naming the record when the separator is missing, when the `+` line repeats an identifier that differs from the header, or when quality and sequence lengths differ.

Also: `columns=[]` passed validation and later failed with `IndexError`; it raises `ValueError` up front. `columns=` now projects a user-supplied `features` schema instead of failing on a field-name mismatch.

Not changed: trailing whitespace in the description is stripped, as Biopython's `FastqGeneralIterator` does. Tests: 28, ruff clean.

behroozazarkhalili:
CI note: the full upstream workflow (ubuntu and windows, py3.10 to 3.14, minimum and latest deps) passed on this exact head (f933c6180) on my fork: https://github.com/behroozazarkhalili/datasets/actions/runs/33910049879. Approving the workflow run here should be a formality.

behroozazarkhalili:
Adapted to the merged `BioSequence` type as suggested in #8542, and rebased on main (head 3a1330ee4).

Each row now has a `record` column of type `BioSequence(format="fastq")` holding that record's bytes exactly as they appear in the file, next to the parsed columns, which still need no dependency. With biopython installed `record` decodes to a `SeqRecord`; without it you can drop the column through `columns` or cast it with `decode=False`. Column selection is now validated once at construction (a user-supplied `features` schema selects its own columns, duplicates are rejected, a caller-supplied `info` is kept), and the test fixtures write exact bytes so the Windows lane compares the same record bytes as Linux.
One behaviour worth knowing, not changed here: bare-CR line endings decode wrongly because `BioSequence.decode_example` reads the bytes without universal newlines; a small follow-up PR will fix that in the feature type.

Full upstream workflow on this head on my fork: https://github.com/behroozazarkhalili/datasets/actions/runs/34526799043

lhoestq (approved):
lgtm :)

Issue #7934, opened 2026-01-08
xPath cannot handle hdfs:///xxxx properly


li-yi-dong:
### Describe the bug

_as_str('hdfs:///xxxx') would return hdfs://xxxx. Removing one / and making the path invalid.

For the use case like
```
ds = load_dataset(
    "parquet",
    data_files={
        "train": "hdfs:///user/path/to/data/train*.parquet",
    },
    streaming=True,
    storage_options={
        "host": "hostname",  
    }
)
```
would get 
```
  File "/usr/local/lib/python3.11/site-packages/datasets/load.py", line 1511, in load_dataset
    return builder_instance.as_streaming_dataset(split=split)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/datasets/builder.py", line 1193, in as_streaming_dataset
    splits_generators = {sg.name: sg for sg in self._split_generators(dl_manager)}
                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/datasets/packaged_modules/parquet/parquet.py", line 123, in _split_generators
    with open(file, "rb") as f:
         ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/datasets/streaming.py", line 73, in wrapper
    return function(*args, download_config=download_config, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/datasets/utils/file_utils.py", line 963, in xopen
    file_obj = fsspec.open(file, mode=mode, *args, **kwargs).open()
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/fsspec/core.py", line 508, in open
    out = open_files(
          ^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/fsspec/core.py", line 295, in open_files
    fs, fs_token, paths = get_fs_token_paths(
                          ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/fsspec/core.py", line 672, in get_fs_token_paths
[...]

PR #7935, opened 2026-01-09
Bug fix: Add HDFS hostname to protocol prefix

li-yi-dong:
For HDFS url with hostname like `hdfs://hostname/user/xxx`, the function `resolve_pattern` would drop the hostname, and outputs  `hdfs:///user/xxx`. This may break later file operations by trying to connect to wrong HDFS cluster.

### DIFF
--- a/src/datasets/data_files.py
+++ b/src/datasets/data_files.py
@@ -358,6 +358,8 @@ def resolve_pattern(
         else (fs.protocol if isinstance(fs.protocol, str) else fs.protocol[0])
     )
     protocol_prefix = protocol + "://" if protocol != "file" else ""
+    if protocol == "hdfs" and "host" in fs.storage_options:
+        protocol_prefix += f"{fs.storage_options['host']}"
     glob_kwargs = {}
     if protocol == "hf":
         # 10 times faster glob with detail=True (ignores costly info like lastCommit)


lhoestq:
Hi ! is it related to https://github.com/huggingface/datasets/issues/7934 ?

It's not clear to me why the protocol would need this, given hostname should be present in `pattern` already

```python
resolve_pattern("hdfs://hostname/user/xxx", ...)
```

li-yi-dong:
> Hi ! is it related to #7934 ?
> 
> It's not clear to me why the protocol would need this, given hostname should be present in `pattern` already
> 
> ```python
> resolve_pattern("hdfs://hostname/user/xxx", ...)
> ```

It's related to #7934 in a subttle way. In my use case, I need to specify the hdfs hostname. In theory, I can do it by
```python
ds = load_dataset(
    "parquet",
    data_files={
        "train": "hdfs://hostname/xxx*.parquet",
    },
    streaming=True,
)
```
or
```python
ds = load_dataset(
    "parquet",
    data_files={
        "train": "hdfs:///xxx*.parquet",
    },
    streaming=True,
    storage_options={
        "host": "hostname"
    }
)
```
None of them work.
The first one does not work due to what this PR trying to fix, and the second one due to #7934.

Yes, `resolve_pattern` would be called like `resolve_pattern("hdfs://hostname/user/xxx", ...)`, but its out put would be like `hdfs:///user/xxx`, no hostname in it. This output would be passed to later file operation like `fsspec.open()`. It needs the hostname in the url to find the HDFS cluster correctly.

li-yi-dong:
@lhoestq 
Hi! Is there any concern?🙃

lhoestq:
I see, I think the path forward is to fix https://github.com/huggingface/datasets/issues/7934 which sounds like an actual xPath bug, while resolve_pattern dropping the hostname comes from fsspec HDFS implementation that we should probably try to follow

li-yi-dong:
Fixing #7934 alone can solve my problem. 

But I don't think fsspec intends to drop the hostname. Function `resolve_pattern` here is supposed to convert a pattern to absolute file paths, and keeping the protocol intouched. `fs.glob` just returns the absolute paths to files, of which no hostname should in the result. The problem is how the function `resolve_pattern` reconstructs the whole path, ignoring the HDFS hostname in the protocol.

From another point of view, in `resolve_pattern` `fs.glob` is call with `hdfs://hostname/user/xxx` but latter `fs.open` is called with `hdfs:///user/xxx`, which is inconsistent.

PR #7951, opened 2026-01-19
feat: Add GenBank file format support for biological sequence data

behroozazarkhalili:
## Summary
Add native support for loading GenBank (.gb, .gbk, .genbank) files, a standard format for biological sequence data with annotations maintained by NCBI.

## Changes
- Add `genbank` packaged module with pure Python state machine parser
- Register GenBank extensions in `_PACKAGED_DATASETS_MODULES` and `_EXTENSION_TO_MODULE`
- Add comprehensive test suite (28 tests)

## Features
- **Metadata parsing**: LOCUS, DEFINITION, ACCESSION, VERSION, KEYWORDS, ORGANISM, taxonomy
- **Feature parsing**: Structured JSON output with location parsing (complement, join)
- **Sequence parsing**: ORIGIN section with automatic length calculation
- **Compression support**: gzip, bz2, xz via magic bytes detection
- **Memory efficiency**: Dual-threshold batching (batch_size + max_batch_bytes)
- **Large sequences**: Uses `large_string` Arrow type for sequences/features

## Usage
```python

# Load GenBank files
ds = load_dataset("genbank", data_files="sequences.gb")

# With options
ds = load_dataset("genbank", data_files="*.gbk", 
                  columns=["sequence", "organism", "features"],
                  parse_features=True)
```

## Test plan
- [x] All 28 unit tests pass
- [x] Tests cover: basic loading, multi-record, compression, feature parsing, column filtering, batching, schema types

### DIFF
--- a/src/datasets/packaged_modules/genbank/genbank.py
+++ b/src/datasets/packaged_modules/genbank/genbank.py
@@ -1,0 +1,480 @@
+"""GenBank file loader for biological sequence data with annotations.
+
+GenBank is a text-based format for storing nucleotide or protein sequences together with
+their annotations and metadata, widely used in bioinformatics and maintained by NCBI.
+
+This implementation uses a lightweight pure Python state machine parser,
+requiring zero external dependencies.
+"""
+
+import bz2
+import gzip
+import itertools
+import json
+import lzma
+import re
+from dataclasses import dataclass
+from typing import Optional
+
+import pyarrow as pa
+
+import datasets
+from datasets.builder import Key
+from datasets.features.features import require_storage_cast
+from datasets.table import table_cast
+
+
+logger = datasets.utils.logging.get_logger(__name__)
+
+
+# Conservative limit to stay well under Parquet's i32::MAX page limit (~2GB)
+# Using 256MB as default since Parquet compresses data and we want headroom
+DEFAULT_MAX_BATCH_BYTES = 256 * 1024 * 1024  # 256 MB
+
+
+# Parser states for the GenBank state machine
+class ParserState:
+    HEADER = "HEADER"
+    FEATURES = "FEATURES"
+    ORIGIN = "ORIGIN"
+    COMPLETE = "COMPLETE"
+
+
+@dataclass
+class GenBankConfig(datasets.BuilderConfig):
+    """BuilderConfig for GenBank files.
+
+    Args:
+        features: Dataset features (optional, will be inferred if not provided).
+        batch_size: Maximum number of records per batch. Works in conjunction with
+            max_batch_bytes - a batch is flushed when either limit is reached.
+        max_batch_bytes: Maximum cumulative bytes per batch. This prevents Parquet
+            page size errors when dealing with very large sequences. Set to None
+            to disable byte-based batching.
+        columns: Subset of columns to include. Options: ["locus_name", "accession",
+            "version", "definition", "organism", "taxonomy", "keywords", "sequence",
+            "features", "length", "molecule_type"].
+        parse_features: Whether to parse FEATURES section into structured JSON.
+            If False, stores raw text.
+    """
+
+    features: Optional[datasets.Features] = None
+    batch_size: int = 10000
+    max_batch_bytes: Optional[int] = DEFAULT_MAX_BATCH_BYTES
+    columns: Optional[list[str]] = None
+    parse_features: bool = True
+
+    def __post_init__(self):
+        super().__post_init__()
+
+
+class GenBank(datasets.ArrowBasedBuilder):
+    """Dataset builder for GenBank files."""
+
+    BUILDER_CONFIG_CLASS = GenBankConfig
+
+    # All supported GenBank extensions
+    EXTENSIONS: list[str] = [".gb", ".gbk", ".genbank"]
+
+    # All available columns
+    ALL_COLUMNS: list[str] = [
+        "locus_name",
+        "accession",
+        "version",
+        "definition",
+        "organism",
+        "taxonomy",
+        "keywords",
+        "sequence",
+        "features",
+        "length",
+        "molecule_type",
+    ]
+
+    def _info(self):
+        return datasets.DatasetInfo(features=self.config.features)
+
+    def _split_generators(self, dl_manager):
+        """Generate splits from data files.
+
+        The `data_files` kwarg in load_dataset() can be a str, List[str],
+        Dict[str,str], or Dict[str,List[str]].
+
+        If str or List[str], then the dataset returns only the 'train' split.
+        If dict, then keys should be from the `datasets.Split` enum.
+        """
+        if not self.config.data_files:
+            raise ValueError(f"At least one data file must be specified, but got data_files={self.config.data_files}")
+        dl_manager.download_config.extract_on_the_fly = True
+        data_files = dl_manager.download_and_extract(self.config.data_files)
+        splits = []
+        for split_name, files in data_files.items():
+            if isinstance(files, str):
+                files = [files]
+            files = [dl_manager.iter_files(file) for file in files]
+            splits.append(datasets.SplitGenerator(name=split_name, gen_kwargs={"files": files}))
+        return splits
+
+    def _cast_table(self, pa_table: pa.Table) -> pa.Table:
+        """Cast Arrow table to configured features schema."""
+        if self.config.features is not None:
+            schema = self.config.features.arrow_schema
+            if all(not require_storage_cast(feature) for feature in self.config.features.values()):
+                pa_table = pa_table.cast(schema)
+            else:
+                pa_table = table_cast(pa_table, schema)
+            return pa_table
+        return pa_table
+
+    def _open_file(self, filepath: str):
+        """Open file with automatic compression detection based on magic bytes.
+
+        Supports gzip, bzip2, and xz/lzma compression formats.
+        """
+        with open(filepath, "rb") as f:
+            magic = f.read(6)
+
+        if magic[:2] == b"\x1f\x8b":  # gzip magic number
+            return gzip.open(filepath, "rt", encoding="utf-8")
+        elif magic[:3] == b"BZh":  # bzip2 magic number
+            return bz2.open(filepath, "rt", encoding="utf-8")
+        elif magic[:6] == b"\xfd7zXZ\x00":  # xz magic number
+            return lzma.open(filepath, "rt", encoding="utf-8")
+        else:
+            return open(filepath, "r", encoding="utf-8")
+
+    def _parse_feature_location(self, location_str: str) -> dict:
+        """Parse a GenBank feature location string into a structured dict.
+
+        Examples:
+            "100..200" -> {"start": 100, "end": 200, "strand": 1}
+            "complement(100..200)" -> {"start": 100, "end": 200, "strand": -1}
+            "join(1..100,200..300)" -> {"start": 1, "end": 300, "strand": 1, "parts": [[1,100],[200,300]]}
+        """
+        location = {"strand": 1}
+
+        # Check for complement
+        if location_str.startswith("complement("):
+            location["strand"] = -1
+            location_str = location_str[11:-1]  # Remove "complement(" and ")"
+
+        # Check for join
+        if location_str.startswith("join("):
+            location_str = location_str[5:-1]  # Remove "join(" and ")"
+            parts = []
+            for part in location_str.split(","):
+                part = part.strip()
+                if ".." in part:
+                    start, end = part.split("..")
+                    # Handle < and > symbols for partial sequences
+                    start = int(start.lstrip("<>"))
+                    end = int(end.lstrip("<>"))
+                    parts.append([start, end])
+            if parts:
+                location["parts"] = parts
+                location["start"] = parts[0][0]
+                location["end"] = parts[-1][1]
+            return location
+
+        # Simple location
+        if ".." in location_str:
+            start, end = location_str.split("..")
+            location["start"] = int(start.lstrip("<>"))
+            location["end"] = int(end.lstrip("<>"))
+        elif location_str.isdigit():
+            location["start"] = int(location_str)
+            location["end"] = int(location_str)
+
+        return location
+
+    def _parse_genbank(self, fp):
+        """State machine parser for GenBank format.
+
+        GenBank format has several sections:
+        - LOCUS: Contains name, length, molecule type, etc.
+        - DEFINITION: Description of the sequence
+        - ACCESSION: Database accession number
+        - VERSION: Version with GI number
+        - KEYWORDS: Associated keywords
+        - SOURCE/ORGANISM: Organism information and taxonomy
+        - FEATURES: Detailed annotations
+        - ORIGIN: The actual sequence data
+        - // : Record terminator
+
+        Args:
+            fp: File-like object opened in text mode.
+
+        Yields:
+            Dict with parsed record fields for each GenBank record.
+        """
+        state = ParserState.HEADER
+        record = self._new_record()
+        current_feature = None
+        current_qualifier_key = None
+        current_qualifier_value = []
+        features_list = []
+
+        for line in fp:
+            # Record terminator
+            if line.startswith("//"):
+                # Finalize any pending feature
+                if current_feature is not None:
+                    if current_qualifier_key is not None:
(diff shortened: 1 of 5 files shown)


src/datasets/packaged_modules/genbank/genbank.py
```diff
@@ -0,0 +1,480 @@
...
+            start, end = location_str.split("..")
+            location["start"] = int(start.lstrip("<>"))
+            location["end"] = int(end.lstrip("<>"))
+        elif location_str.isdigit():
+            location["start"] = int(location_str)
+            location["end"] = int(location_str)
+
+        return location
+
+    def _parse_genbank(self, fp):
```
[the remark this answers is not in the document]
behroozazarkhalili: Thanks @lhoestq. I've split `_parse_genbank` into the per-section handlers I mentioned: a small `_FeatureAccumulator` now owns the multi-line FEATURES/qualifier bookkeeping (it was duplicated in four spots), and the state machine is broken into `_handle_header_line` / `_handle_features_line` / `_handle_origin_line`, so `_parse_genbank` itself is now a ~40-line dispatch loop. Behavior is unchanged — all existing tests pass. Hope that's easier to review; happy to adjust further.

behroozazarkhalili:
Hi @lhoestq — quick status. The `_parse_genbank` refactor we discussed is done: the state machine is split into per-section handlers (`_handle_header_line` / `_handle_features_line` / `_handle_origin_line`) with a small `_FeatureAccumulator` owning the multi-line FEATURES/qualifier bookkeeping, so `_parse_genbank` itself is now a short dispatch loop and much easier to review. Behavior is unchanged and all existing tests pass.

I've also rebased the branch onto the latest `main` so it's current. Ready for another look whenever you have time — happy to adjust further.

src/datasets/packaged_modules/genbank/genbank.py
```diff
@@ -0,0 +1,495 @@
...
+        """
+        with open(filepath, "rb") as f:
+            magic = f.read(6)
+
+        if magic[:2] == b"\x1f\x8b":  # gzip magic number
+            return gzip.open(filepath, "rt", encoding="utf-8")
+        elif magic[:3] == b"BZh":  # bzip2 magic number
+            return bz2.open(filepath, "rt", encoding="utf-8")
+        elif magic[:6] == b"\xfd7zXZ\x00":  # xz magic number
+            return lzma.open(filepath, "rt", encoding="utf-8")
```
lhoestq: I think you don't need thjis. Indeed you call download_and_extract() so files are already extracted/uncompressed

src/datasets/packaged_modules/genbank/genbank.py
```diff
@@ -0,0 +1,495 @@
...
+        batch_size: Maximum number of records per batch. Works in conjunction with
+            max_batch_bytes - a batch is flushed when either limit is reached.
+        max_batch_bytes: Maximum cumulative bytes per batch. This prevents Parquet
+            page size errors when dealing with very large sequences. Set to None
+            to disable byte-based batching.
+        columns: Subset of columns to include. Options: ["locus_name", "accession",
+            "version", "definition", "organism", "taxonomy", "keywords", "sequence",
+            "features", "length", "molecule_type"].
+        parse_features: Whether to parse FEATURES section into structured JSON.
+            If False, stores raw text.
```
lhoestq: You can remove this argument and always parse IMO.

There is a new `datasets.Json()` type for this kind of data that you can use btw

src/datasets/packaged_modules/genbank/genbank.py
```diff
@@ -0,0 +1,495 @@
...
+        "taxonomy",
+        "keywords",
+        "sequence",
+        "features",
+        "length",
+        "molecule_type",
+    ]
+
+    def _info(self):
+        return datasets.DatasetInfo(features=self.config.features)
```
lhoestq: if the features are always the same you can pass them here to DatasetInfo (and use self.config.features if the user wants to override them for some reason)

lhoestq (commented):
Great thanks for the changes ! Are there datasets on Hugging Face already I can use to test your implementation ?

src/datasets/packaged_modules/genbank/genbank.py
behroozazarkhalili: Done — removed the `parse_features` argument; FEATURES is now always parsed. I also switched the `features` column to the new `datasets.Json()` type as you suggested, so it decodes to a parsed Python object instead of a JSON string. Verified the round-trip stores a single JSON encoding (no double-encoding) and reads back as a list of feature dicts.

src/datasets/packaged_modules/genbank/genbank.py
behroozazarkhalili: Done. The schema is always the same, so I declared a canonical `DEFAULT_FEATURES` and pass it to `DatasetInfo` in `_info()`. `self.config.features` still overrides it if a user wants to, and when `columns=` is set the features are projected to that subset. Column validation now also happens at config time as a side effect (fail-fast).

src/datasets/packaged_modules/genbank/genbank.py
behroozazarkhalili: Happy to remove the magic-byte decompression in `_open_file` — just want to confirm one thing first. The unit tests exercise `_generate_tables` directly (e.g. `test_genbank_gzipped/bz2/xz`) without going through `dl_manager`, so they rely on `_open_file` decompressing. If I drop it and trust `download_and_extract()` + `extract_on_the_fly`, those direct-call tests need to route through the dl_manager (or be dropped). Do you want it fully removed (and the tests reworked accordingly), or kept as a lightweight fallback for the direct-call path? I'll follow your preference.

behroozazarkhalili:
On test datasets: I searched the Hub and couldn't find a public dataset hosting **raw** GenBank files (`.gb`/`.gbk`/`.genbank`) — the `Hack90/ncbi_genbank_part_*` series exists but it's already converted to parquet, so it won't exercise this loader. If it's useful, I can upload a small GenBank sample dataset (a handful of real NCBI records) to the Hub as a canonical test fixture. Want me to do that? In the meantime the PR's own `tests/packaged_modules/test_genbank.py` fixtures cover single/multi-record, complex feature locations, compression, batching, and column subsetting.

src/datasets/packaged_modules/genbank/genbank.py
lhoestq: if you use _generate_tables in the tests directly without the dl_manager, you can replace `file.gb.gz` with `gz://file.gb::file.gb.gz` and the regular open() function and it should work

(open() is extended to support chained uris for compression/archives)

lhoestq:
I think having one on the Hub that we can point to in the documentation would be useful for the community. This way they have an actual example they can try and replicate for their cases

src/datasets/packaged_modules/genbank/genbank.py
behroozazarkhalili: Done in `b5aa245`. Removed `_open_file`; the loader now uses plain `open()` (which becomes `xopen` under the streaming patch), matching the text/json modules.

One small correction for the tests: the chained-URI protocol is `gzip` rather than `gz` — `gz://` raises `Protocol not known` since fsspec registers the codec as `gzip` (and `bz2`/`xz` for the others). So the direct-`_generate_tables` tests now build the URI as `gzip://sequence.gb::/path/sequence.gb.gz`. I derive the protocol from `_get_extraction_protocol` in the fixtures so they track the loader's real behavior rather than hardcoding it. All 27 tests pass with the gzip/bz2/xz coverage preserved through the real `xopen` path.

behroozazarkhalili:
Created one: **https://huggingface.co/datasets/ermiaazarkhalili/datasets-genbank-test**

It has 3 real, complete NCBI records — `U49845` (yeast), `V00662` (human mitochondrion, stored **gzipped** to show compressed loading works), and `X04370` (124 kb) — plus a card with a runnable example. Verified end-to-end:
```python
ds = load_dataset("ermiaazarkhalili/datasets-genbank-test", data_files="*.gb*", split="train")
# 3 rows; FEATURES decode to structured objects (Json()), and the gzipped file loads via xopen
```
Happy to move it under a different org or reference it from the docs page — let me know where you'd like it linked.

behroozazarkhalili:
An adversarial review pass over this branch turned up five parser defects. All five are reachable from valid GenBank files, so I reproduced each one against the branch before touching anything, then fixed them in 5452d7f with a regression test apiece.

| Defect | Observed on the branch before the fix | Expected |
|---|---|---|
| A qualifier key that repeats within one feature kept only its last value | `db_xref` = `"TAX:2"` | `["TAX:1", "TAX:2"]` |
| An operator location wrapped across lines was parsed from its first fragment alone | `join(1..10,20..30,`⏎`40..50)` gave `end=30`, 2 parts | `end=50`, 3 parts |
| A `DEFINITION` wrapped onto a second line was truncated | `"First line of a long definition that"` | both lines |
| Any indented line after `ORGANISM` was appended to `taxonomy` | `"Eukaryota; Fungi;; Ascomycota. Wrapped author line here;"` | `"Eukaryota; Fungi; Ascomycota."` |
| A protein `LOCUS` line produced no molecule type | `molecule_type` = `""` | `"protein"` |

The first one matters most in practice: `/db_xref` repeats on almost every annotated NCBI record, and `/note`, `/EC_number` and `/inference` repeat routinely too, so cross-references were being dropped silently rather than failing loudly.

The header and location defects share a cause worth naming, because it explains why the fix is a small refactor rather than three patches. GenBank is a continuation format: a value may wrap onto following lines indented into the value column, so what a line means depends on which field is currently open. The parser tracked that for qualifiers but inferred it elsewhere from downstream evidence. Its header rule was `line.startswith("            ") and record["organism"]`, which uses "organism has been set" as a proxy for "we are inside ORGANISM" and stays true for the rest of the record, which is how `REFERENCE`, `AUTHORS` and `JOURNAL` text ended up inside `taxonomy`. The open field is now explicit: `_HeaderState.active_field` for the header, and a balanced-parenthesis buffer for locations. A keyword this loader does not read clears the active field, so its continuations are dropped instead of leaking into whatever came before.

One change needs your call, since it alters the shape of the data rather than just fixing a value. Preserving repeated qualifiers means a key can no longer hold a bare string, so `features[i]["qualifiers"]` goes from `{key: str}` to `{key: list[str]}`. I chose consistent lists over a `str`-or-`list` union: the union pushes an isinstance check onto every consumer, and lists match `SeqFeature.qualifiers` in Biopython, which is where this format family is heading anyway. Two existing assertions in the test file are updated for the new shape. If you would rather keep scalars for single-valued keys, say so and I will switch.

32 tests pass on `tests/packaged_modules/test_genbank.py`, `ruff check` and `ruff format` are clean.

src/datasets/packaged_modules/genbank/genbank.py
behroozazarkhalili: Agreed, and removed in b5aa24561. `_split_generators` sets `extract_on_the_fly` and calls `download_and_extract`, so the magic-byte branch was decompressing a second time. The callsite is now a plain `with open(file, encoding="utf-8")`, which the streaming patch turns into `xopen`, so gzip, bz2, xz and zstd all arrive through the shared extraction layer instead of a codec branch in this module.

src/datasets/packaged_modules/genbank/genbank.py
behroozazarkhalili: Both done in a4ead401c. The `parse_features` argument is gone and FEATURES is always parsed, and the column is now typed `datasets.Json()`. One detail worth flagging: the storage schema declares that column as `pa.json_()` directly rather than as a string, because building it as a string and then casting to `Json()` encoded the payload twice, so the value came back as a JSON string of a JSON string.

src/datasets/packaged_modules/genbank/genbank.py
behroozazarkhalili: Done in a4ead401c. The schema is a class-level `DEFAULT_FEATURES`, and `_info` passes it to `DatasetInfo`, using `self.config.features` when the user overrides it, exactly as you described. `_cast_table` now casts to that same resolved schema rather than inferring one per batch.

src/datasets/packaged_modules/genbank/genbank.py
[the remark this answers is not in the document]
behroozazarkhalili: I wrote it. It is not adapted from another project or ported from existing code, so there is nothing owed in attribution; the format handling is written against the NCBI GenBank specification directly.

On the length, you were right, and it has changed since you looked. `_parse_genbank` was one long function doing header, features and sequence in a single loop. It is now a dispatch loop of 39 lines that hands each line to a handler for its section:

- `_handle_header_line` for LOCUS through ORGANISM, with `_HeaderState` tracking which field a wrapped line continues
- `_handle_features_line` for the FEATURES table, with `_FeatureAccumulator` holding the multi-line state of one feature
- `_handle_origin_line` for the sequence block

That split is `f03561118`. Each piece is now reviewable on its own, and the state that used to live in local variables is named.

Worth saying plainly, since it bears on how much you trust the parser: an adversarial review pass over this branch afterwards found five defects in it, all reachable from valid GenBank files, and all now fixed with a regression test each. Repeated qualifiers such as `/db_xref` kept only their last value; a `join(...)` location wrapped across lines was parsed from its first fragment alone; a wrapped `DEFINITION` was truncated; any indented line after `ORGANISM` was appended to `taxonomy`, so `REFERENCE` and `AUTHORS` text leaked into it; and a protein `LOCUS` line produced an empty `molecule_type`. Details are in the comment above.

The suite is 32 tests. If the parser is still more than you want to review by eye, I am happy to walk through any single handler.

behroozazarkhalili:
Follow-up on the parser after a second review pass (own edge-case probes plus a cross-model adversarial review of the diff, every finding reproduced by execution before changing code). Head is 6029c5e81.

Fixed:
- `order(...)` locations raised `ValueError: too many values to unpack`. Locations are now parsed recursively, so `complement()` can wrap or sit inside `join()`/`order()`, commas split only at the top parenthesis level, and a remote reference such as `AB000001.1:1..10` keeps its range. A join whose parts lie on different strands reports `strand: null`, as Biopython's `CompoundLocation` does.
- Wrapped qualifier values joined without a space (`"alpha betagamma"`). They now join with a space, except `/translation` which concatenates directly; this is the rule in `Bio.GenBank.Scanner`. Quote state is tracked, so a continuation line starting with `/` inside an open quoted value is no longer read as a new qualifier.
- `BASE COUNT` inside the feature table produced a feature of type `COUNT`, and a `CONTIG` line was folded into the sequence. Column-0 keywords are now handled as keywords in both states.
- LOCUS molecule types with a strandedness prefix (`ds-DNA`) read as empty.
- `columns=` combined with a user `features` schema was not projected.

Schema additions, following Biopython's representation since the `features` column already mirrors `SeqFeature`:
- every location carries `start_partial` / `end_partial` (the `<` / `>` markers were being dropped);
- new columns `secondary_accessions` (list, includes continuation lines) and `contig`;
- a qualifier with no value (`/pseudo`) is stored as `[""]` rather than `["true"]`;
- a `features` schema naming a subset of columns now selects exactly those columns.

Tests: 47 in `tests/packaged_modules/test_genbank.py`, ruff clean.

One item left as documented behavior rather than changed: a single record larger than `max_batch_bytes` is emitted in a batch of its own, since an empty batch cannot absorb it.

behroozazarkhalili:
CI note: the full upstream workflow (ubuntu and windows, py3.10 to 3.14, minimum and latest deps) passed on this exact head (6029c5e81) on my fork: https://github.com/behroozazarkhalili/datasets/actions/runs/33910033565. Approving the workflow run here should be a formality.

behroozazarkhalili:
Adapted to the merged `BioSequence` type as suggested in #8542, and rebased on main (head 6383f3bf0).

Each row now has a `record` column of type `BioSequence(format="genbank")` holding that record's bytes exactly as they appear in the file, next to the parsed columns, which still need no dependency. With biopython installed `record` decodes to a `SeqRecord`; without it you can drop the column through `columns` or cast it with `decode=False`. Column selection is now validated once at construction (a user-supplied `features` schema selects its own columns, duplicates are rejected, a caller-supplied `info` is kept), and the test fixtures write exact bytes so the Windows lane compares the same record bytes as Linux.

Full upstream workflow on this head on my fork: https://github.com/behroozazarkhalili/datasets/actions/runs/34526817575

lhoestq (approved):
aaaand lgtm !

PR #7974, opened 2026-01-31
Fix duplicate kwargs in load_dataset_builder

KOKOSde:
Avoid passing duplicate keyword arguments to `load_dataset_builder`.

Some module factories provide values in `builder_kwargs` (e.g. `base_path`), and users can also pass the same keys via `config_kwargs`, which raises:
`TypeError: ... got multiple values for keyword argument ...`.

Fix: if `config_kwargs` is provided, drop overlapping keys from `builder_kwargs` (keep the user-provided values).

Includes a regression test.

### DIFF
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -1149,6 +1149,14 @@ def load_dataset_builder(
     dataset_name = builder_kwargs.pop("dataset_name", None)
     info = dataset_module.dataset_infos.get(config_name) if dataset_module.dataset_infos else None
 
+    # Avoid passing duplicate keyword arguments to the builder.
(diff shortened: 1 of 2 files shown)


KOKOSde:
Quick ping on this one. Happy to update anything if you want changes.

KOKOSde:
Hi @lhoestq, quick follow-up on this small regression fix for duplicate builder kwargs. Happy to adjust the approach or test if you prefer a different direction.

lhoestq:
Hi ! this issue which was reported here https://github.com/huggingface/datasets/pull/8066 was fixed in https://github.com/huggingface/datasets/pull/8068

PR #7979, opened 2026-02-02
Use temp files in push_to_hub to save memory

lhoestq:
write parquet data to temp files on disk prior to upload to save memory

this is enabled for for datasets loaded using streaming=True/False


HaukurPall:
And they are never deleted?

HaukurPall:
Just to be clear @lhoestq, by not deleting the files on successful shard push they will accumulate and eventually fill the disk. By accumulating files, stream processing is effectively broken. Why stream files if you end up with all of them on disk?

Is there no review process in place for PRs? Even an automated one would have gotten this.

lhoestq:
arf what an oversight, thanks for reporting I'll do a patch release asap

lhoestq:
4.6.1 is released with a fix !

PR #7982, opened 2026-02-02
Fix unstable tokenizer fingerprinting (enables map cache reuse)

KOKOSde:
Fix unstable dataset fingerprinting when hashing `PreTrainedTokenizerFast`.

Some tokenizers backed by `tokenizers.Tokenizer` mutate runtime settings (padding/truncation) when called, which can change the serialized state and make dataset fingerprints unstable. That prevents `.map(load_from_cache_file=True)` from reusing cache files.

Fix: when hashing, temporarily disable backend padding/truncation so runtime settings don’t affect the fingerprint, then restore the original settings.

Includes a regression test showing `Hasher.hash(tokenizer)` stays stable after calling the tokenizer.

### DIFF
--- a/src/datasets/utils/_dill.py
+++ b/src/datasets/utils/_dill.py
@@ -223,6 +223,8 @@ def _save_transformersPreTrainedTokenizerBase(pickler, obj):
     state = obj.__dict__.copy()
     if "cache" in state and isinstance(state["cache"], dict):
         state["cache"] = {}
+    if "deprecation_warnings" in state and isinstance(state["deprecation_warnings"], dict):
+        state["deprecation_warnings"] = {}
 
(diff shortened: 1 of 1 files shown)


KOKOSde:
Hi! It looks like the GitHub Actions check suites for this PR are in `action_required` (no workflows actually ran). This is usually due to fork workflow approval.

Could a maintainer please approve/run the workflows so CI can execute? Happy to address anything CI flags once it runs.

KOKOSde:
Quick ping on this one. Happy to update anything if you want changes.

lhoestq:
cc @itazap and @ArthurZucker for viz: this is the trick we use to hash a tokenizer

this way the hash doesn't change if the state of the tokenizer changed

PR #7987, opened 2026-02-04
Fix index out of bound error with original_shard_lengths.

jonathanasdf:
I have gotten the following error

```
    original_shard_lengths[original_shard_id] += 1
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
IndexError: list index out of range
```

Not sure what causes it, but this fixes the error. This may not be the proper fix for the root cause though.


lhoestq:
this was fixed in https://github.com/huggingface/datasets/pull/8023, sorry I didn't see your PR and took the liberty of fixing this bug as part of this improvement for `from_generator`

PR #7997, opened 2026-02-08
fix: Dataset.map writer initialization when first examples return None

### LINKED ISSUE #7990: Dataset.map crashes when first examples return None and later examples return dict — writer not initialized

veeceey:
Fixes #7990

## Summary

When Dataset.map is called and the first examples processed return None, the writer is never properly initialized, causing a ValueError.

## Changes

- Modified _map_single to initialize the writer early if the first batch returns empty results
- Ensures writer is set before the first call to writer.write_batch

## Test Plan

- Added test case that reproduces the bug
- Verified the fix resolves the issue
- Existing tests still pass

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -3673,7 +3673,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                     _time = time.time()
                     for i, example in iter_outputs(shard_iterable):
                         if update_data:
-                            if i == 0:
+                            if writer is None:
                                 buf_writer, writer, tmp_file = init_buffer_and_writer()
                                 stack.enter_context(writer)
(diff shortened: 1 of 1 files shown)


veeceey:
Closing in favor of #7996 which addresses the same issue with the same fix.

Issue #8010, opened 2026-02-19
audiofolder / folder-based loader rejects metadata.csv when file_name column is inferred as large_string


Koosh0610:
### Describe the bug

When loading a dataset with load_dataset("audiofolder", data_dir=...) and a metadata.csv (with a file_name column), loading can fail with:

`ValueError: file_name or *_file_name must be present as dictionary key (with type string) in metadata files`

even when the CSV clearly has a file_name column with string values.

**Cause**

In folder_based_builder.py, the metadata schema is inferred from the CSV via pandas to Arrow. For string columns, Arrow often uses the large_string type. The check only accepts Value("string") and does not accept Value("large_string"), so validation fails even though the column is a valid string column.

### Steps to reproduce the bug
1. Create a folder with:
   - `metadata.csv` with columns: `file_name`, `transcript`, `duration` (and audio `.wav` files referenced by `file_name`).
2. Run:
```
   from datasets import load_dataset
   ds = load_dataset("audiofolder", data_dir="/path/to/folder")
```

### Expected behavior

Metadata CSV with a `file_name` column (inferred as either `string` or `large_string`) should be accepted, since both represent string data.

### Environment info

- `datasets` version: 4.5.0
- Platform: Linux-6.8.0-100-generic-x86_64-with-glibc2.39
- Python version: 3.12.3
- `huggingface_hub` version: 0.36.0
- PyArrow version: 23.0.0
- Pandas version: 3.0.0
- `fsspec` version: 2025.10.0

Koosh0610:
Proposing a PR for this [here](https://github.com/huggingface/datasets/pull/8011)

PS: I didn't find a contributing documentation. Please recommend any changes in issue or PR if any. Thank you!

PR #8011, opened 2026-02-19
fix: condition for file_name feature check

Koosh0610:
## Description
Accept `Value("large_string")` in addition to `Value("string")` for the `file_name` (or `*_file_name`) metadata key in folder-based builders (e.g. `audiofolder`).

## Problem
When loading a dataset with `load_dataset("audiofolder", data_dir=...)` and a `metadata.csv` file, the metadata schema is inferred via pandas to Arrow. For string columns, Arrow often uses the `large_string` type, so `Features.from_arrow_schema()` yields `Value("large_string")` for the `file_name` column. The builder only checks for `Value("string")`, so validation fails with:

`ValueError: file_name or *_file_name must be present as dictionary key (with type string) in metadata files`

Users then have to convert metadata to JSONL (which produces Arrow `string`) or work around the loader.

## Solution
Treat `Value("large_string")` as valid for the `file_name` / `*_file_name` key, since it is still a string type and behaves the same for resolution of audio/image paths. This matches the fact that when metadata is loaded from CSV (via pandas and then converted to Arrow), string columns are often inferred as Arrow `large_string`, so accepting it keeps CSV metadata working without format changes.

### DIFF
--- a/src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
+++ b/src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
@@ -216,7 +216,8 @@ class FolderBasedBuilder(datasets.GeneratorBasedBuilder):
                     out = type(feature)()
                     for key in feature:
                         if (key == "file_name" or key.endswith("_file_name")) and (
-                            feature[key] == datasets.Value("string") or feature[key] == datasets.Value("large_string")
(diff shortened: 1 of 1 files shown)


Koosh0610:
Hi @lhoestq, have proposed a fix for [#8010](https://github.com/huggingface/datasets/issues/8010)

PR #8033, opened 2026-02-28
perf: use deque for async map task/index queues

giulio-leone:
## Problem

`arrow_dataset.py` and `iterable_dataset.py` drain `indices` and `tasks` lists front-to-back via `.pop(0)` during parallel async map operations. With `MAX_NUM_RUNNING_ASYNC_MAP_FUNCTIONS_IN_PARALLEL` items in flight, each `.pop(0)` is **O(n)** making the drain loop **O(n²)**.

## Solution

Switch both `indices` and `tasks` to `collections.deque` with `.popleft()` for **O(1)** front removal.

## Changes

- `src/datasets/arrow_dataset.py`: Import `deque`, type `tasks` and `indices` as deque, use `.popleft()`
- `src/datasets/iterable_dataset.py`: Import `deque`, type `tasks`, `indices`, and `_owned_loops_and_tasks` entry as deque, use `.popleft()`

## Testing

- Syntax verified via `ast.parse()` on both files

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -35,7 +35,7 @@ import tempfile
 import time
 import warnings
 import weakref
-from collections import Counter, defaultdict
+from collections import Counter, defaultdict, deque
 from collections.abc import Iterable, Iterator, Mapping
 from collections.abc import Sequence as Sequence_
 from copy import deepcopy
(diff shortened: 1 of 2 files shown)


giulio-leone:
Friendly ping — CI is green and this is ready for review. Happy to address any feedback. Thanks!

PR #8042, opened 2026-03-05
Fix schema enforcement in streaming _convert_to_arrow

HaukurPall:
When the iter_arrow chain is broken (e.g. after .map() without with_format("arrow")), _convert_to_arrow falls back to pa.Table.from_pylist() which infers types from values. If early batches contain None or empty lists, Arrow infers null/list<null> types that conflict with later batches containing real data, causing ArrowInvalid schema mismatch errors.

Add an optional features parameter to _convert_to_arrow that applies cast_table_to_features to each produced table, mirroring what ArrowWriter.write_table does in the map-style path.

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -144,6 +144,7 @@ def _convert_to_arrow(
     iterable: Iterable[tuple[Key, dict]],
     batch_size: int,
     drop_last_batch: bool = False,
+    features: Optional[Features] = None,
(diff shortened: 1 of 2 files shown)


HaukurPall:
I'm going to close this PR as I have a more complete branch which fixes feature handling in the streaming case. The `features` handling in the streaming case is btw completely broken at the moment, it just happens to work in some cases. I will add more test cases to show it in a later PR.

PR #8043, opened 2026-03-05
Add features parameter to IterableDatasetDict.map

HaukurPall:
IterableDataset.map accepts a features parameter to declare the output schema, but IterableDatasetDict.map did not expose it. This meant users of IterableDatasetDict had no way to preserve feature metadata through map operations.


HaukurPall:
I'm going to close this PR as it depends on correct `features` handling in `IterableDataset`, which is not done correctly at the moment.

PR #8045, opened 2026-03-05
Write IterableDataset to parquet incrementally instead of materializing entire shard in memory

HaukurPall:
`IterableDataset.to_parquet()` materialized the entire shard into memory via
`pa.concat_tables(list(...))` before writing. For datasets with large shards and multiprocessing (push_to_hub with num_proc > 0) the memory requirements can be really large.

This PR writes parquet incrementally by iterating Arrow batches and writing each
directly to `pq.ParquetWriter`, reducing the memory requirement from shard size to something around row group size (with lots of caveats).

**Note**

A thing to note is that this code path is only executed when `features` are set on the `IterableDataset`, in f.ex. `map` or the dataset is `cast` before `push_to_hub`.

Changes:
- Extend `ParquetDatasetWriter` to accept both `Dataset` and `IterableDataset`.
- `IterableDataset.to_parquet()` now delegates directly to `ParquetDatasetWriter`
- Use `pq.ParquetWriter` as a context manager for proper cleanup on errors
- Falls back to materialization only when `features` is `None` (schema required upfront)

That being said, I'm not entirely convinced that extending the ParquetDatasetWriter to support IterableDataset is the correct approach or even if there are other footguns related to CDC or other stuff.

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -4466,10 +4466,9 @@ class IterableDataset(DatasetInfoMixin):
                     features=shard.features,
                 )
             shard_path_in_repo = f"{data_dir}/{split}-{index:05d}-of-{num_shards:05d}.parquet"
-            tmp_file = tempfile.NamedTemporaryFile(suffix=".parquet", delete=False)
-            try:
+            with tempfile.NamedTemporaryFile(suffix=".parquet") as tmp_file:
                 shard.to_parquet(tmp_file)
-                tmp_file.close()
+                tmp_file.flush()
                 parquet_metadata = pq.read_metadata(tmp_file.name)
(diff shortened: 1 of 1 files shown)


HaukurPall:
I've tested this change along with the changes from my other PRs (https://github.com/HaukurPall/datasets/tree/all-fixes) and this change **dramatically** reduces the memory requirement per process.

Context:

I'm working with audio datasets, the audio is embedded in the parquet. When loading a 1GB shard with embedded mp3 audio, decoding the audio and writing it back as wav, the shard size increases to around 9GB. I run this processing with num_proc > 1, so the total memory required is large. With this change, the memory required per process drops from ~25GB to ~4GB for this use-case.

HaukurPall:
@lhoestq any feedback?

PR #8050, opened 2026-03-07
Fix IterableDataset features being set to None on chained filter (Fixes #8037)

### LINKED ISSUE #8037: IterableDataset.filter() chaining fails due to features being lost when is_typed=True
### Describe the bug

Chaining multiple .filter() calls on an IterableDataset raises a TypeError: 'NoneType' object is not a mapping on the second .filter() call.

In `IterableDataset.filter()`, the internal `ex_iterable` is wrapped with FormattedExamplesIterable:

```python
ex_iterable = FormattedExamplesIterable(
    ex_iterable,
    formatting=self._formatting,
    features=None if ex_iterable.is_typed else self._info.features,
    token_per_repo_id=self._token_per_repo_id,
)
```

After the first .filter(), the resulting ex_iterable becomes typed (is_typed=True) because a ===MASK=== column is added. On the second .filter(), since is_typed=True, features=None is passed to FormattedExamplesIterable. This causes .features to return None, and FilteredExamplesIterable.__init__ fails when trying to unpack it.

The fix should preserve the existing features when is_typed=True, e.g.:

```python
features=ex_iterable.features if ex_iterable.is_typed else self._info.features
```

```python

ds = load_dataset("bigcode/the-stack-v2", "Scala", split="train", streaming=True)
ds = ds.filter(lambda row: row["src_encoding"] == "UTF-8")
ds = ds.filter(lambda row: row["branch_name"] == "refs/heads/master")  # TypeError here
```

## Error traceback
```
File ".../datasets/iterable_dataset.py", line 2975, in filter
    ex_iterable = FilteredExamplesIterable(
File ".../datasets/iterable_dataset.py", line 1606, in __init__
    features = Features({**ex_iterable.features, self.mask_column_name: Value("bool")})
TypeError: 'NoneType' object is not a mapping
```

### Expected behavior

The error should not raised in this context.

### Environment info

datasets: 4.5.0

Sanket291:
This PR implements the solution suggested in #8037.

It updates the features argument in IterableDataset.filter() to preserve existing features when ex_iterable.is_typed is True. This prevents the TypeError: 'NoneType' object is not a mapping from being raised when chaining multiple .filter() calls.

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -3404,7 +3404,7 @@ class IterableDataset(DatasetInfoMixin):
             ex_iterable = FormattedExamplesIterable(
                 ex_iterable,
                 formatting=self._formatting,
-                features=None if ex_iterable.is_typed else self._info.features,
+                features=ex_iterable.features if ex_iterable.is_typed else self._info.features,
                 token_per_repo_id=self._token_per_repo_id,
             )
 


lhoestq:
Thanks for the PR ! The original issue poster actually opened the same PR https://github.com/huggingface/datasets/pull/8053 but with additional tests, so I took the liberty of merging theirs - I hope you don't mind

PR #8062, opened 2026-03-12
Fix iterable dataset dict num proc shards

HaukurPall:
If the number of shards for some split is lower than num_proc the IterableDatasetDict.push_to_hub will crash.

This handling of num_proc is as close to identical as the handling in IterableDataset.push_to_hub


lhoestq:
I believe this was fixed in https://github.com/huggingface/datasets/pull/8068

PR #8066, opened 2026-03-14
fix: avoid duplicate kwargs collisions in load_dataset_builder

### LINKED ISSUE #4910: Identical keywords in build_kwargs and config_kwargs lead to TypeError in load_dataset_builder()
## Describe the bug
In `load_dataset_builder()`, `build_kwargs` and `config_kwargs` can contain the same keywords leading to a TypeError("type object got multiple values for keyword argument "xyz"). 

I ran into this problem with the keyword: `base_path`. It might happen with other kwargs as well. I think a quickfix would be 
```python
builder_cls = import_main_class(dataset_module.module_path)
builder_kwargs = dataset_module.builder_kwargs
data_files = builder_kwargs.pop("data_files", data_files)
config_name = builder_kwargs.pop("config_name", name)
hash = builder_kwargs.pop("hash")
base_path = builder_kwargs.pop("base_path")
```
and then pass base_path into `builder_cls`.

## Steps to reproduce the bug
```python
load_dataset("rotten_tomatoes", base_path="./sample_data")
```

## Expected results
The docs state: `**config_kwargs` — Keyword arguments to be passed to the [BuilderConfig](https://huggingface.co/docs/datasets/v2.4.0/en/package_reference/builder_classes#datasets.BuilderConfig) and used in the [DatasetBuilder](https://huggingface.co/docs/datasets/v2.4.0/en/package_reference/builder_classes#datasets.DatasetBuilder).

So I would expect to be able to pass the base_path into `load_dataset()`. 
## Actual results
TypeError("type object got multiple values for keyword argument "base_path"). 

## Environment info

- `datasets` version: 2.4.0
- Platform: macOS-12.5-arm64-arm-64bit
- Python version: 3.8.9
- PyArrow version: 9.0.0

biefan:
## Summary
- avoid duplicate keyword collisions when `config_kwargs` overlap with module-provided `builder_kwargs` in `load_dataset_builder`
- this fixes cases like passing `base_path=...` where `base_path` is already present in `builder_kwargs`
- add a regression test using a local dataset directory to ensure overriding kwargs no longer raises

## Reproduction
Before this change:
```python
load_dataset_builder("/path/to/local/data", base_path="/path/to/local/data")
# TypeError: ... got multiple values for keyword argument 'base_path'
```

## Validation
- `uv run --python 3.11 --with-editable . --with pytest --with setuptools -m pytest tests/test_load.py::test_load_dataset_builder_config_kwargs_can_override_builder_kwargs tests/test_load.py::test_load_dataset_builder_for_absolute_data_dir -q`
- `uv run --python 3.11 --with ruff --with setuptools ruff check src/datasets/load.py tests/test_load.py`

Fixes #4910

### DIFF
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -1145,6 +1145,10 @@ def load_dataset_builder(
     )
     dataset_name = builder_kwargs.pop("dataset_name", None)
     info = dataset_module.dataset_infos.get(config_name) if dataset_module.dataset_infos else None
+    # Some kwargs (e.g. base_path) can be present in both builder_kwargs and config_kwargs.
+    # Remove duplicates to avoid "multiple values for keyword argument" when instantiating builders.
+    for key in config_kwargs:
+        builder_kwargs.pop(key, None)
 
     if (
         path in _PACKAGED_DATASETS_MODULES
--- a/tests/test_load.py
+++ b/tests/test_load.py
@@ -858,6 +858,11 @@ def test_load_dataset_builder_for_relative_data_dir(complex_data_dir):
         assert len(builder.config.data_files["test"]) > 0
 
 
+def test_load_dataset_builder_config_kwargs_can_override_builder_kwargs(complex_data_dir):
+    builder = datasets.load_dataset_builder(complex_data_dir, base_path=complex_data_dir)
+    assert isinstance(builder, DatasetBuilder)
+
+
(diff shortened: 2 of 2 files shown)


lhoestq:
I believe this was fixed in https://github.com/huggingface/datasets/pull/8068

PR #8084, opened 2026-03-21
Add reshard support for JSON Lines and CSV file formats

muyihao:
This PR extends the IterableDataset.reshard() functionality to support JSON Lines and CSV file formats, building on the foundation laid in #7992.

### Summary
- JSON Lines : Resharding splits files by line boundaries (only when num_shards is specified)
- CSV : Resharding splits files by line boundaries (only when num_shards is specified)
- Parquet : Already supported via row groups (enhanced to support num_shards parameter)

### Usage
```python

# JSON Lines
ds = load_dataset("json", data_files=["data.jsonl"], streaming=True)
resharded_ds = ds['train'].reshard(num_shards=100)

# CSV
ds = load_dataset("csv", data_files=["data.csv"], streaming=True)
resharded_ds = ds['train'].reshard(num_shards=100)

# Parquet
ds = load_dataset("parquet", data_files=["data.parquet"], streaming=True)
resharded_ds = ds['train'].reshard(num_shards=100)
```

### DIFF
--- a/src/datasets/packaged_modules/csv/csv.py
+++ b/src/datasets/packaged_modules/csv/csv.py
@@ -279,7 +279,7 @@ class Csv(datasets.ArrowBasedBuilder):
                     file, start_byte, end_byte = file_item
                 else:
                     file, start_byte, end_byte = file_item, 0, None
-
+                file_path = file
                 with open(file, "rb") as f:
                     header_line = f.readline()
                     header_end_pos = f.tell()
@@ -320,5 +320,5 @@ class Csv(datasets.ArrowBasedBuilder):
                         # logger.warning('\n'.join(str(pa_table.slice(i, 1).to_pydict()) for i in range(pa_table.num_rows)))
                         yield Key(shard_idx, batch_idx), self._cast_table(pa_table)
                 except ValueError as e:
-                    logger.error(f"Failed to read file '{file}' with error {type(e)}: {e}")
+                    logger.error(f"Failed to read file '{file_path}' with error {type(e)}: {e}")
                     raise
--- a/src/datasets/packaged_modules/csv/csv.py
+++ b/src/datasets/packaged_modules/csv/csv.py
@@ -1,3 +1,4 @@
+import io
 import os
 from dataclasses import dataclass
 from typing import Any, Callable, Optional, Union
@@ -181,16 +182,8 @@ class Csv(datasets.ArrowBasedBuilder):
     def _generate_shards(self, base_files, files_iterables):
         yield from base_files
 
-    def _estimate_line_count(self, file, max_bytes=20 << 20):
-        """Estimate the number of lines in a file by reading the first max_bytes (default 20MB).
-        
-        Args:
-            file: Path to the file
-            max_bytes: Maximum bytes to read (default 20MB)
-            
-        Returns:
-            Estimated line count
-        """
+    def _estimate_byte_size_per_line(self, file, max_bytes=20 << 20):
+        """Estimate the byte size per line by reading the first max_bytes (default 20MB)."""
         try:
             with open(
                 file,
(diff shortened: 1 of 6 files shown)


muyihao:
@lhoestq , hi, lhoestq, could you please help review this PR? Thanks!

src/datasets/packaged_modules/csv/csv.py
```diff
@@ -180,7 +180,60 @@ def _cast_table(self, pa_table: pa.Table) -> pa.Table:
...
+                    with open(
+                        file,
+                        "r",
+                        encoding=self.config.encoding or "utf-8",
+                        errors=self.config.encoding_errors or "strict",
+                    ) as f:
+                        # Read the file to count lines
+                        line_count = 0
+                        while f.readline():
+                            line_count += 1
```
lhoestq: this can be expensive on big files, maybe limit to the first 20MB to estimate the total number of lines ?

we could also assume that the files contain the same kind of data, so maybe there is no need to check every single file

src/datasets/packaged_modules/csv/csv.py
```diff
@@ -180,7 +180,60 @@ def _cast_table(self, pa_table: pa.Table) -> pa.Table:
...
+            # No resharding - return the original gen_kwargs
+            for base_file, files_iterable in zip(base_files, files_iterables):
+                for file in files_iterable:
+                    yield {
+                        "base_files": [base_file],
+                        "files_iterables": [[file]],
+                        "shard_start_line": [0],
+                        "shard_end_line": [None],
+                    }
+            return
```
lhoestq: IIUC when num_shards is None then it should shard maximally.

In the case of the CSV loader, the minimum shard size is defined by `self.config.chunksize` which is 10_000 lines by default. It could make sense to define maximum sharding as aiming for `self.config.chunksize` lines per shard, WDYT ?

src/datasets/packaged_modules/csv/csv.py
```diff
@@ -191,13 +244,27 @@ def _generate_tables(self, base_files, files_iterables):
...
+        if isinstance(shard_end_line, list):
+            shard_end_line = shard_end_line[0]
+
         for shard_idx, files_iterable in enumerate(files_iterables):
             for file in files_iterable:
-                csv_file_reader = pd.read_csv(file, iterator=True, dtype=dtype, **self.config.pd_read_csv_kwargs)
+                # Handle shard_start_line and shard_end_line
+                read_csv_kwargs = self.config.pd_read_csv_kwargs.copy()
+                if shard_start_line > 0:
+                    read_csv_kwargs["skiprows"] = shard_start_line
```
lhoestq: `skiprows` may lead some shard to require downloading a lot of unnecessary data to skip the rows, maybe you could skip bytes instead. Something along those lines maybe ?

```python
if skip_bytes is not None:
    with open(file, "rb") as f:
        header = f.readline()
        f.seek(skip_bytes)
        lines = (f.read(n_bytes) + f.readline()).splitlines()
        lines[0] = header
    file = io.BytesIO("\n".join(lines))
csv_file_reader = pd.read_csv(file, iterator=True, dtype=dtype, **read_csv_kwargs)
```

e.g. this dataset has multiple CSV files of 4GB each: https://huggingface.co/datasets/Koala-36M/Koala-36M-v1

src/datasets/packaged_modules/csv/csv.py
```diff
@@ -180,7 +180,60 @@ def _cast_table(self, pa_table: pa.Table) -> pa.Table:
...
+        # Only reshard if num_shards is specified
+        if num_shards is None:
+            # No resharding - return the original gen_kwargs
+            for base_file, files_iterable in zip(base_files, files_iterables):
+                for file in files_iterable:
+                    yield {
+                        "base_files": [base_file],
+                        "files_iterables": [[file]],
+                        "shard_start_line": [0],
+                        "shard_end_line": [None],
```
lhoestq: it's probably more practical to have them together like this:

```suggestion
                        "base_files": [base_file],
                        "files_iterables": [[(file, 0, None)]],
```

this way later you can do

```python
        for shard_idx, files_iterable in enumerate(files_iterables):
            for file in files_iterable:
                shard_start_line, shard_end_line = None, None
                if isinstance(file, tuple):
                    file, shard_start_line, shard_end_line = file
```

lhoestq: alternatively you could keep it this way, but in that case you need to zip() on files_iterables:

```suggestion
                        "base_files": [base_file],
                        "files_iterables": [[file]],
                        "shard_start_lines": [0],
                        "shard_end_lines": [None],
```


```python
        shard_start_lines = shard_start_lines or [None] * len(base_files)
        shard_end_lines = shard_start_lines or [None] * len(base_files)
        for base_file, files_iterable, shard_start_line, shard_end_line in zip(base_files, files_iterables, shard_start_lines, shard_end_lines):
            for file in files_iterable:
```

src/datasets/packaged_modules/json/json.py
```diff
@@ -127,12 +127,78 @@ def _cast_table(self, pa_table: pa.Table, json_field_paths=()) -> pa.Table:
...
+            # No resharding - return the original gen_kwargs
+            for base_file, files_iterable in zip(base_files, files_iterables):
+                for file in files_iterable:
+                    yield {
+                        "base_files": [base_file],
+                        "files_iterables": [[file]],
+                        "shard_start_line": [0],
+                        "shard_end_line": [None],
+                    }
+            return
```
lhoestq: when sharding maximally we could use the fact that `self.config.chunksize` is 10MB by default and aim for this maybe

(yes it has the same name as CSV but is in bytes, not lines. This collision is not ideal and I guess comes from the fact that pandas uses chunksize in lines for CSV and pyarrow uses chunksize in bytes for json)

lhoestq (commented):
Cool ! I left a few comments. Mostly on CSV but this also applies to JSON since the logic is almost the same. Let me know what you think :)

src/datasets/packaged_modules/csv/csv.py
muyihao: That's a great point. Iterating through massive files just for a line count is indeed a bottleneck.

src/datasets/packaged_modules/csv/csv.py
muyihao: Makes sense. I'll refactor this part to

src/datasets/packaged_modules/csv/csv.py
muyihao: cool, I will try with it

src/datasets/packaged_modules/json/json.py
muyihao: got it

muyihao:
> Cool ! I left a few comments. Mostly on CSV but this also applies to JSON since the logic is almost the same. Let me know what you think :)

@lhoestq Hi, lhoestq, I've made the updates accordingly, could you please take another look? 😊

muyihao:
Hi @lhoestq , just a gentle ping on this. I've re-verified the byte-offset logic for CSV/JSONL and confirmed it handles large files efficiently. Let me know if you have any concerns or would like me to adjust the implementation further. Looking forward to your feedback!

PR #8111, opened 2026-03-31
add sample_by="document" support for jsonl files

cfahlgren1:
With more agentic traces (_ie Claude Code / Codex / Pi etc_) being published to the Hub as JSONL files, it would be nice to easily support loading / viewing them in Dataset Viewer. 

Datasets like https://huggingface.co/datasets/cfahlgren1/agent-sessions contain one JSONL file per session. Without sample_by, each line becomes a separate row, losing the file-level grouping. With sample_by="document", each file is preserved as a single row:

```python
  ds = load_dataset("cfahlgren1/agent-sessions", sample_by="document")
  # Each row is one session file
```

  Also configurable via YAML metadata:

```yaml
  ---
  configs:
    - config_name: default
      data_files: "*.jsonl"
      sample_by: "document"
  ---
```

### DIFF
--- a/src/datasets/packaged_modules/json/json.py
+++ b/src/datasets/packaged_modules/json/json.py
@@ -50,6 +50,7 @@ class JsonConfig(datasets.BuilderConfig):
     chunksize: int = 10 << 20  # 10MB
     newlines_in_values: Optional[bool] = None
     on_mixed_types: Optional[Literal["use_json"]] = "use_json"
+    sample_by: Literal["line", "document"] = "line"
 
     def __post_init__(self):
         super().__post_init__()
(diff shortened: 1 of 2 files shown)


lhoestq (commented):
I think we should decode the JSON data instead of yielding them as text. Also we could name the column "documents" instead of "text".

Moreover since I expect agent traces to have a fixed schema we need to use the main code path based on `paj.read_json` which handles mixed types. We could simply yield Key(..., pa.concat_tables(tables)) once per file instead of at every batch

julien-c:
would the auto-detection of agent harness (claude vs. codex vs. opencode etc) and other metadata, live in datasets in your opinion, @lhoestq? or in dataset-server, or in the frontend?

cfahlgren1:
> would the auto-detection of agent harness (claude vs. codex vs. opencode etc) and other metadata, live in datasets in your opinion, @lhoestq? or in dataset-server, or in the frontend?

would be nice personally if we could have it live in datasets-server so we can do less parsing in frontend and add some sort of detection of trace datasets 

image

cfahlgren1:
closing this in favor of https://github.com/huggingface/datasets/pull/8113

PR #7953, opened 2026-01-19
#5354: replace list with Sequence in from_parquet type hints

ashmi8:
\This PR replaces `list` type hints with `Sequence` in `from_parquet` to improve type checking.

Note: Local pytest errors on Python 3.13 due to removal of `distutils` are unrelated to this change.

PR #8112, opened 2026-04-01
fix: replace List with Sequence in type hints for public API functions (#5354)

abdulhayykhan:
## Summary

Fixes #5354

Replaces `list[X]` with `Sequence[X]` in function argument type annotations 
across the public API. `List` is invariant while `Sequence` is covariant, 
which means passing a `list` to a parameter typed as `List` causes mypy errors, 
even though lists are valid sequences.

## Changes

- `src/datasets/arrow_dataset.py` — `path_or_paths`, `columns`, `input_columns`, `remove_columns`
- `src/datasets/iterable_dataset.py` — same parameters across equivalent methods  
- `src/datasets/combine.py` — `datasets` and `probabilities` parameters
- `src/datasets/dataset_dict.py` — `column_names` and `columns` parameters

## What was NOT changed

- Docstrings (separate concern)
- Internal variables and function bodies
- Return types
- Non-public/internal helpers

## Notes

PR #7953 addressed only `from_parquet` in `arrow_dataset.py` and had merge 
conflicts. This PR covers the broader public API consistently.


abdulhayykhan:
Updated — public Dataset.map(), Dataset.filter(), IterableDataset.map() and IterableDataset.filter() signatures now also use Sequence[str] consistently. Thanks for the pointer.

abdulhayykhan:
Note: PR #8067 by @biefan also addresses path_or_paths in arrow_dataset.py. This PR covers a broader scope across 4 files and additional parameters (columns, input_columns, remove_columns, map(), filter()). Happy to coordinate with @biefan or adjust scope based on maintainer preference.

PR #8113, opened 2026-04-01
Parse agent traces

lhoestq:
Supports claude_code, codex and pi

JSON Lines files are parsed so one file = one row

The resulting row has three fields: harness, session_id, traces, and file_name with features:

```python
AGENT_TRACES_FEATURES = datasets.Features(
    {
        "harness": datasets.Value("string"),
        "session_id": datasets.Value("string"),
        "traces": datasets.List(datasets.Json()),
        "file_name": datasets.Value("string"),
    }
)
```
cc @cfahlgren1


cfahlgren1 (approved):
parsing seems to match

Issue #8116, opened 2026-04-02
`to_json()` serializes `Json()` feature as escaped string instead of native JSON object


ain-soph:
### Describe the bug

When a dataset column uses the `Json()` feature type, calling `to_json()` writes the value as an escaped JSON string (e.g., `"{\"cat_a\":1,\"cat_b\":2}"`) instead of a native JSON object (e.g., `{"cat_a":1,"cat_b":2}`).

This causes two problems:

1. The output JSON file is not interoperable with other tools (pandas, jq, etc.) — they see a string, not an object.
2. Round-trip is broken: `load_dataset("json", ...)` without specifying `features=` infers the column as `Value('string')` instead of `Json()` or even a struct, because the file contains escaped strings.

In contrast, when the same dict data is stored as a struct type (auto-inferred), `to_json()` correctly writes it as a native JSON object.

### Steps to reproduce the bug

```
from datasets import Dataset, Features, Json, load_dataset

# Create dataset with Json() feature
data = [
    {"scores": {"cat_a": 1, "cat_b": 2}},
    {"scores": {"cat_x": 3, "cat_y": 4, "cat_z": 5}},
    {"scores": {"cat_a": 6, "cat_m": 7}},
]
ds = Dataset.from_list(data, features=Features({"scores": Json()}))

print("In-memory features:", ds.features)
# {'scores': Json(decode=True)}

print("In-memory data:", ds[0])
# {'scores': {'cat_a': 1, 'cat_b': 2}}

# Save to JSON
ds.to_json("test_json_feature.json")
```

1. Inspect the saved file `test_json_feature.json`:

```javascript
{"scores":"{\"cat_a\":1,\"cat_b\":2}"}
{"scores":"{\"cat_x\":3,\"cat_y\":4,\"cat_z\":5}"}
{"scores":"{\"cat_a\":6,\"cat_m\":7}"}
```

The `scores` field is a __escaped string__, not a native JSON object.

2. Reload without specifying features:

```python
ds_loaded = load_dataset("json", data_files="test_json_feature.json", split="train")
print("Loaded features:", ds_loaded.features)
# {'scores': Value('string')}

print("Loaded data:", ds_loaded[0])
# {'scores': '{"cat_a":1,"cat_b":2}'}  — it's a string, not a dict!
```

3. For comparison, struct type writes correctly:

```python
ds_struct = Dataset.from_list(data)
ds_struct.to_json("test_struct_feature.json")
```

[...]

PR #8122, opened 2026-04-04
fix: decode Json() values before calling DataFrame.to_json() (#8116)

Brianzhengca:
# Summary
fixes (https://github.com/huggingface/datasets/issues/8116)
Basically the issue is that if a dataset column uses Json() as a feature type, calling to_json() on that dataset causes a raw json escaped string to be exported, without decoding it first. 
### root cause
`to_json` exports feature values in their stored form by default, which means that the raw json escaped string would be exported
### changes made
I tried fixing this minimally. before calling pandas `to_json()`, the writer now decodes the batch with `PandasFeaturesDecoder`. That makes Json() values serialize as native JSON. 
### test plan
- [x] reproduced this error locally
- [x] verified the fix with direct JsonDatasetWriter smoke tests and by confirming it reuses the existing PandasFeaturesDecoder path correctly.

### DIFF
--- a/src/datasets/io/json.py
+++ b/src/datasets/io/json.py
@@ -133,7 +133,9 @@ class JsonDatasetWriter:
             indices=self.dataset._indices,
         )
         batch = batch.to_pandas()
-        batch = PandasFeaturesDecoder(self.dataset.features).decode_batch(batch)
+        for json_field_path in get_json_field_paths_from_feature(self.dataset.features):
+            col, *json_field_subpath = json_field_path
+            batch[col] = batch[col].apply(partial(json_encode_field, json_field_path=json_field_subpath))
         json_str = batch.to_json(path_or_buf=None, orient=orient, lines=lines, **to_json_kwargs)
         if not json_str.endswith("\n"):
             json_str += "\n"
--- a/src/datasets/io/json.py
+++ b/src/datasets/io/json.py
@@ -1,13 +1,5 @@
 import multiprocessing
 import os
-from typing import BinaryIO, Optional, Union
-
-import fsspec
-
-from .. import Dataset, Features, NamedSplit, config
-from ..formatting import query_table
-from ..formatting.formatting import PandasFeaturesDecoder
-from ..packaged_modules.json.json import Json
 from functools import partial
 from typing import BinaryIO, Optional, Union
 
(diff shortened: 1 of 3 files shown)


src/datasets/io/json.py
```diff
@@ -131,7 +132,9 @@ def _batch_json(self, args):
             key=slice(offset, offset + self.batch_size),
             indices=self.dataset._indices,
         )
-        json_str = batch.to_pandas().to_json(path_or_buf=None, orient=orient, lines=lines, **to_json_kwargs)
+        batch = batch.to_pandas()
+        batch = PandasFeaturesDecoder(self.dataset.features).decode_batch(batch)
```
lhoestq:
```suggestion
        for json_field_path in get_json_field_paths_from_feature(self.dataset.features):
            col, *json_field_subpath = json_field_path
            batch[col] = batch[col].apply(partial(json_encode_field, json_field_path=json_field_subpath))
```

src/datasets/io/json.py
```diff
@@ -6,6 +6,7 @@
 
 from .. import Dataset, Features, NamedSplit, config
 from ..formatting import query_table
+from ..formatting.formatting import PandasFeaturesDecoder
 from ..packaged_modules.json.json import Json
 from ..utils import tqdm as hf_tqdm
```
lhoestq:
```suggestion
from functools import partial
from typing import BinaryIO, Optional, Union

import fsspec

from .. import Dataset, Features, NamedSplit, config
from ..formatting import query_table
from ..formatting.formatting import PandasFeaturesDecoder
from ..packaged_modules.json.json import Json
from ..utils import tqdm as hf_tqdm
from ..utils.json import get_json_field_paths_from_feature, json_encode_field
```

lhoestq (commented):
Great ! The `PandasFeaturesDecoder` may return non-JSON-compatible objects like PIL Images, so I just added a suggestion to decode the JSON objects only using some utility functions. Lmk how this sounds :)

src/datasets/io/json.py
Brianzhengca: Cool! I overlooked this issue. will commit this fix.

Brianzhengca:
Just committed the fix, please let me know if there are any other issues

ain-soph:
Thanks for the quick fix!

lhoestq (approved):
lgtm ! I added the decoding function that was missing, nothing major since it's literally the inverse of the encoding one

PR #8126, opened 2026-04-08
Fix batching for table-formatted datasets

### LINKED ISSUE #8075: `.batch()` error on formatted datasets
The `.batch()` method currently assumes the input (batch) is always a dictionary, which causes errors when it isn't. This can happen with formatted datasets, since formats like `"pyarrow"`, `"pandas"` (only affects `IterableDataset`), and `"polars"` return tables/dataframes instead of dictionaries.

For example:
```python
from datasets import IterableDataset, Dataset
list(IterableDataset.from_dict({"a": [1, 2, 3, 4]}).with_format("pyarrow").batch(2))
# AttributeError: 'pyarrow.lib.Table' object has no attribute 'items'
```

Ideally, the result should be the same whether the format is applied before or after batching, i.e., the following should hold for all the format types:
```python
assert list(IterableDataset.from_dict({"a": [1, 2, 3, 4]}).with_format(format_type).batch(2)) == list(IterableDataset.from_dict({"a": [1, 2, 3, 4]}).batch(2).with_format(format_type))
assert list(Dataset.from_dict({"a": [1, 2, 3, 4]}).with_format(format_type).batch(2)) == list(Dataset.from_dict({"a": [1, 2, 3, 4]}).batch(2).with_format(format_type))
```

bluehyena:
## Summary
Fix #8075

Fix `.batch()` on table-formatted `Dataset` and `IterableDataset`.

Before this change, calling `.batch()` on datasets formatted as `pyarrow`, `pandas`, or `polars` could fail because the batching path assumed dict-like inputs. This updates batching to use an Arrow-based path for table-style formats, so batching works regardless of whether the table format is applied before or after `.batch()`. These two forms now behave equivalently:

```python
dataset.with_format(format_type).batch(n)
```

and

```python
dataset.batch(n).with_format(format_type)
```

for `pyarrow`, `pandas`, and `polars`.

## What changed

- added a shared `_batch_arrow_table` helper to build batched Arrow tables
- updated `Dataset.batch()` to route table-formatted batching through an Arrow `.map(...)` path
- updated `IterableDataset.batch()` to do the same and then restore the original table format
- added regression tests for `pyarrow`, `pandas`, and `polars` on both `Dataset` and `IterableDataset`

## Tests

- `python -m pytest tests/test_arrow_dataset.py -k "test_dataset_batch or test_dataset_batch_with_table_format or test_dataset_batch_with_polars_format" -q`
- `python -m pytest tests/test_iterable_dataset.py -k "test_iterable_dataset_batch or test_iterable_dataset_batch_with_table_format or test_iterable_dataset_batch_with_polars_format" -q`
- `python -m ruff check src/datasets/arrow_dataset.py src/datasets/iterable_dataset.py src/datasets/table.py tests/test_arrow_dataset.py tests/test_iterable_dataset.py`

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -113,6 +113,7 @@ from .table import (
     InMemoryTable,
     MemoryMappedTable,
     Table,
+    _batch_arrow_table,
     _memory_mapped_record_batch_reader_from_file,
     cast_array_to_feature,
     concat_tables,
@@ -4098,6 +4099,17 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
 
         def batch_fn(example):
             return {k: [v] for k, v in example.items()}
+
+        if self._format_type in ("arrow", "pandas", "polars"):
+            return self.with_format("arrow").map(
+                _batch_arrow_table,
+                batched=True,
+                batch_size=batch_size,
+                drop_last_batch=drop_last_batch,
+                num_proc=num_proc,
+                new_fingerprint=new_fingerprint,
+                desc="Batching examples",
+            )
 
         return self.map(
             batch_fn,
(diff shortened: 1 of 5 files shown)


lhoestq (approved):
lgtm !

PR #8133, opened 2026-04-14
use full repo-relative file path instead of filename for agent traces

cfahlgren1:
We want to include full path of the file so it's easier to fetch the full trace from repo

  - Store the full repo-relative path in `file_name` instead of just the basename
  - Extracts the path from the `hf://` origin URL set by the download manager

  Before: `file_name = "614b8d50-097f-4e36-aee0-1fdd053335e7.jsonl"`
  After:  `file_name = "sessions/claude/614b8d50-097f-4e36-aee0-1fdd053335e7.jsonl"`

### DIFF
--- a/src/datasets/packaged_modules/json/json.py
+++ b/src/datasets/packaged_modules/json/json.py
@@ -1,5 +1,4 @@
 import io
-import os
 from dataclasses import dataclass
 from typing import Literal, Optional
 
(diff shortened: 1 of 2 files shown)


src/datasets/packaged_modules/json/json.py
cfahlgren1: maybe it would make more sense to update the `file_name` to `file_path` as a breaking change, but wasn't 100% sure, up to you!

**Note:** we don't use `file_name` in moon yet so would be fine if we wanted to do this

lhoestq:
thanks ! let me push some adjustments, and yea let's use file_path instead

Issue #4096, opened 2022-04-05
Add support for streaming Zarr stores for hosted datasets


jacobbieker:
**Is your feature request related to a problem? Please describe.**
Lots of geospatial data is stored in the Zarr format. This format works well for n-dimensional data and coordinates, and can have good compression. Unfortunately, HF datasets doesn't support streaming in data in Zarr format as far as I can tell. Zarr stores are designed to be easily streamed in from cloud storage, especially with xarray and fsspec. Since geospatial data tends to be very large, and on the order of TBs of data or 10's of TBs of data for a single dataset, it can be difficult to store the dataset locally for users. Just adding Zarr stores with HF git doesn't work well (see https://github.com/huggingface/datasets/issues/3823) as Zarr splits the data into lots of small chunks for fast loading, and that doesn't work well with git. I've somewhat gotten around that issue by tarring each Zarr store and uploading them as a single file, which seems to be working (see https://huggingface.co/datasets/openclimatefix/gfs-reforecast for example data files, although the script isn't written yet). This does mean that streaming doesn't quite work though. On the other hand, in https://huggingface.co/datasets/openclimatefix/eumetsat_uk_hrv we stream in a Zarr store from a public GCP bucket quite easily. 

**Describe the solution you'd like**
A way to upload Zarr stores for hosted datasets so that we can stream it with xarray and fsspec. 

**Describe alternatives you've considered**
Tarring each Zarr store individually and just extracting them in the dataset script -> Downside this is a lot of data that probably doesn't fit locally for a lot of potential users.
Pre-prepare examples in a format like Parquet -> Would use a lot more storage, and a lot less flexibility, in the eumetsat_uk_hrv, we use the one Zarr store for multiple different configurations.

albertvillanova:
Hi @jacobbieker, thanks for your request and study of possible alternatives.

We are very interested in finding a way to make `datasets` useful to you.

Looking at the Zarr docs, I saw that among its storage alternatives, there is the ZIP file format: https://zarr.readthedocs.io/en/stable/api/storage.html#zarr.storage.ZipStore

This might be convenient for many reasons:
- On the one hand, we avoid the Git issue with huge number of small files: chunks files are compressed into a single ZIP file
- On the other hand, the ZIP file format is specially suited for streaming data because it allows random access to its component files (i.e. it supports random access to its chunks)

Anyway, I think that a Python loading script will be necessary: you need to implement additional logic to select certain chunks (based on date or other criteria).

[...]

PR #8135, opened 2026-04-15
Add Zarr dataset support: lazy array feature, folder builder, and store-as-table module

Elsword016:
## TL;DR

This PR adds support for **plain Zarr** to `datasets`, split into three complementary pieces:

1. a new `Zarr` **feature** that stores only a path in Arrow and lazily opens the store on access (arrays only, by design — see "Design contract" below);
2. a **`zarrfolder`** packaged module that builds an index dataset from a folder of `.zarr` stores (local or remote);
3. the **`zarr` packaged module** (adopted from #7983, with co-author credit) that reads a single Zarr store as a dataset, row-per-chunk.

This PR follows the restructure suggested in the thread: basic Zarr modelling (compatible with #7983) for Zarr as a whole, with OME-Zarr multiscale support and upload/training helpers deferred to stacked follow-up PRs (see Next Steps).

Initially, I started working on this for my lab's microscopy and connectomics data, and it's at a point where a PR is necessary for everyone to use and to get feedback/improvements. The core design is intentionally lightweight: dataset rows store only a path to a Zarr store, and array data is opened lazily only when the user accesses slices, patches, or metadata. The standard `load_dataset(..., streaming=True)` experience is preserved.

## Motivation

Many scientific and biomedical datasets are not naturally represented as files that can be decoded eagerly into memory. Microscopy volumes, geospatial arrays, medical images, and dense labels are often stored as chunked N-dimensional arrays that are hundreds of GBs or TBs, served from object storage or the Hub, and consumed by reading selected chunks, slices, or training patches. Zarr is a common storage format for these workloads because it stores arrays as independently readable chunks inside a hierarchical store.

Before this PR, users could not represent these datasets directly in `datasets` without writing custom loading code. A typical workflow required users to solve all of the following themselves:

- keep dataset rows small while pointing to very large Zarr stores;
- open local, remote, and Hub-backed Zarr stores lazily;
- read only the chunks needed for a slice, ROI, or patch;
- preserve the standard `load_dataset(..., streaming=True)` experience;
- work around Hub upload constraints for stores with many small chunk files.

## Related issues

- **#4096** : streaming support for hosted Zarr stores: this PR provides the streaming/access foundation (lazy `hf://` opening); the companion upload path (`push_to_hub_zarr`, Zarr v3 sharding) is a planned follow-up.
- **#7863** : native support for cloud-native formats: this PR covers the Zarr part; Lance, Vortex, and Iceberg are out of scope.

### 1. `Zarr` feature (array-only)

Adds a new `Zarr` feature type that stores only a path in Arrow:

```py
{"path": "hf://datasets/username/my-zarr-data@main/sample_001.zarr"}
```

When decoded, the feature returns a lazy `ZarrArrayProxy`. The proxy does not open the store during dataset construction, iteration, or row decoding — it opens on first access (`shape`, `dtype`, `__getitem__`, `iter_patches`, `random_patch`). Values are pickle-safe (important for multiprocessing data loaders), and `decode=False` is supported, matching other modality-aware features.

Supported inputs: local paths, `pathlib.Path`, dicts with a `path` key, and Zarr arrays when a path can be recovered from the store.

Store opening covers:
- local directories;
- `hf://datasets/...` repositories (fsspec-backed, chunked reads, no whole-store download), with private-repo token support via the existing `token_per_repo_id` decode path;
- `hf://buckets/...` object storage;
- custom `storage_options` for other fsspec backends.

Access is cached: a store registry avoids re-opening the same store per row, and a 256 MiB chunk cache (LRU, optional) avoids re-fetching hot chunks.

### 2. `zarrfolder` packaged module

Builds an index dataset from a directory of `.zarr` stores — one row per store:

- local `.zarr` directory discovery;
- remote `data_dir` support via `hf://` / URL file lists (`load_dataset("zarrfolder", data_dir=...)` now passes remote directories through);
- optional label inference from parent folder names;
- optional metadata integration from `metadata.csv`, `metadata.jsonl`, or `metadata.parquet` as the index;
- columns from all rows are unioned, with missing values filled.

### 3. `zarr` packaged module (adopted from #7983)

A single Zarr store can be loaded as a dataset directly:

```py
load_dataset("zarr", data_dir="hf://datasets/username/store.zarr", streaming=True)
```

- Zarr v2 and v3, including consolidated metadata;
- config options: `group`, `features` override, `batch_size`, `consolidated`, `rows_per_shard`, `target_num_shards` (automatic shard sizing via a counting pass, with a progress bar);
- plain-array stores handled through the shared `_select_arrays_by_num_rows` helper.

This module is taken from #7983 and was slightly adapted to avoid duplication. The commit carries `Co-authored-by` credit. A best-effort streaming benchmark is included at `benchmarks/benchmark_zarr_streaming.py` (compares streaming vs non-streaming iteration).

### 4. Supporting changes

- `data_files.py`: store-root matching so Zarr stores (`*.zarr` and equivalent) are recognized as data files without enumerating every internal chunk file;
- `setup.py`: optional `zarr` extra (`datasets[zarr]`) with the Zarr/fsspec deps, plus test requirements;
- `load.py`: remote `data_dir` passthrough for folder-style builders;
- docs: `scientific_load.mdx` (new), `stream.mdx` and `_toctree.yml` updated.

## Design (arrays in the default plugin, groups in the OME plugin)

Following the precedent of napari's Zarr plugins where the default plugin handles plain arrays and the OME plugin (`napari-ome-zarr`) owns the group spec — the default `Zarr` feature resolves **arrays only**:

- a path to a plain array → `ZarrArrayProxy`;
- a path to a **group** → clear `ValueError` explaining the store is a Zarr group, not an array, and pointing to the array sub-path (`store.zarr/array_name`), to `load_dataset("zarr", ...)`, or to the upcoming OME integration.

## Next Steps

I will add the following as stacked PRs after this one is reviewed properly:

- **OME-Zarr**: multiscale levels, axes/scale, `default_plane`, thumbnail helpers, NGFF v0.4/v0.5 metadata — this is where the group/multiscale contract lives.
- **Tooling**: `push_to_hub_zarr` (bucket-first upload, file counting, optional Zarr v3 sharding), `load_zarr_dataset` (fast local discovery), `ZarrCollator` (patch-based training batches).
- **Full demo notebook and docs page** showcasing the complete workflow.

(Lance, Vortex, and Iceberg from #7863 remain out of scope.)

The recommended production workflow remains: upload raw stores to a data repository, build a small index dataset (`zarrfolder` or a manual `Dataset.from_dict`), load it streaming, and `cast_column(..., Zarr())`.

## Validation

New tests (115 across three test files):

- `tests/features/test_zarr.py` — 59: encode/decode, lazy open, local/`hf://datasets`/`hf://buckets` stores, token-aware path handling, `decode=False`, pickle round-trips, patch iteration/sampling, cache behavior, group error path, sub-path into groups;
- `tests/packaged_modules/test_zarrfolder.py` — 35: discovery, labels, metadata-as-index, remote data_dir, column unions;
- `tests/packaged_modules/test_zarr.py` — 21: config validation, v2/v3 loading, store-root loading, features override, shard counting/sizing (adopted from #7983).

## Summary of user-facing APIs

- `datasets.Zarr` feature → `ZarrArrayProxy` (array-only, lazy)
- `load_dataset("zarrfolder", data_dir=...)` — index dataset from a folder of stores
- `load_dataset("zarr", data_dir=...)` — store-as-dataset, row-per-chunk (co-authored with #7983)
- `datasets[zarr]` extra

## Notes
If #7983 merges first, this PR rebases trivially the module files are identical; the feature and zarrfolder commits then sit directly on top.


lhoestq:
How does this PR related to the existing one at https://github.com/huggingface/datasets/pull/7983 ?
They could be complementary, maybe it's worth checking how we can take the best of both PRs ?

Elsword016:
Hi @lhoestq, thanks for your reply. 

My understanding is that #7983 models a **Zarr store as a dataset**: arrays in the selected/root group become columns, axis 0 becomes rows, and `_generate_row_ranges` produces chunk-aligned row shards so streaming is efficient. It also adds `_CountableBuilderMixin` for accurate row counts, v2 consolidated + v3 support, `.zarr` root detection in `resolve_pattern`, and `hf://` support through fsspec `storage_options`. That seems like the right model for row-oriented / tabular-ish Zarr stores.

My PR models a **Zarr store as the value of a dataset column**. Each row stays lightweight, storing only a path, and decodes lazily into a `ZarrArrayProxy`, `ZarrGroupProxy`, or `OmeZarrProxy`. That is the workflow **I’m targeting for the bioimaging /scientific imaging community, where one sample may itself be a large N-D array or OME-Zarr multiscale image, and axis 0 is often a  z/time/channel rather than the dataset row dimension.**

Beyond the feature this PR adds:

- [OME-NGFF](https://ngff.openmicroscopy.org/) v0.4 and v0.5 multiscale support (`levels`, `thumbnail`, `axes`, `scale`, `channel_names`). This needs nested/group-aware access, which #7983 explicitly defers.
- `zarrfolder` / local discovery for folder-of-stores workflows, where each `.zarr` store becomes one row.
- `push_to_hub_zarr(...)` with optional Zarr v3 `ShardingCodec` rechunking, to address the "hundreds of thousands of chunk files" Hub upload problem from #4096.
- Patch helpers (`iter_patches`, `random_patch`) and `ZarrCollator` for training workflows.
- `token_per_repo_id` handling for dataset rows that reference private Hub repos.

Also, in #7983’s `Zarr` is a packaged builder used through `load_dataset("zarr", ...)`, while this PR’s `Zarr` is a feature type registered in `_FEATURE_TYPES`. 

The real overlap seems narrower: fsspec / `hf://` opening, Zarr v2/v3 metadata detection, and `.zarr` store-root handling. Those would be worth consolidating into shared helpers so there is one place to fix hf-fsspec or Zarr metadata edge cases. I’d be happy to reuse the store-opening path from #7983 where it makes sense.

So I see the combined design as:

- `load_dataset("zarr", ...)` from #7983 for **“the Zarr store is the dataset”.**
- `datasets.Zarr` from this PR for **“the Zarr store is a lazy value in a dataset row”.**
- Shared internal helpers for store resolution/opening, optional dependency handling, and fsspec/HF path support.

I’m happy to rebase on top of #7983, or if you have any other suggestions, I'm happy to work on that side as well. I hope this explanation helps.

lubianat:
Hi,

I work with the team developing the OME-Zarr sepc and it is exciting to see this PR for OME-Zarr support on HF. Thank you @Elsword016 and @lhoestq! 

If you need some kind of support from the OME-Zarr side (e.g. additional sample data) in the future, just let us know.  

Also, e keep a thread on a forum for announcing updates related to OME-Zarr (NGFF is basically a synonym): 
https://forum.image.sc/t/ngff-weekly-dev-update-thread/110810. Feel free to announce the PR there if you want extra pairs of eyes.

lubianat:
A nod to the new Buckets offering in HF; 

Copying @Elsword016's https://huggingface.co/datasets/LuciexJune/ome-vip-dataset to a bucket enables validating the dataset: 

https://ome.github.io/ome-ngff-validator/?source=https://huggingface.co/buckets/tiagolubiana/ome-vip-dataset-bucket/resolve

And streaming/visualizing the bytes:
https://biongff.github.io/vizarr/?source=https://huggingface.co/buckets/tiagolubiana/ome-vip-dataset-bucket/resolve

I added also another OME-Zarr 0.5 dataset for reference (though not with a linked index):
 
https://huggingface.co/datasets/tiagolubiana/bbbc032

validation: https://ome.github.io/ome-ngff-validator/?source=https://huggingface.co/buckets/tiagolubiana/bbbc032-bucket/resolve

viewing: https://biongff.github.io/vizarr/?source=https://huggingface.co/buckets/tiagolubiana/bbbc032-bucket/resolve

Elsword016:
Hi @lubianat, apologies for the late response. Thank you for your response, I would appreciate any help that I can get. I am solo dev on this one and scope is quite big and I am really happy to see the response. Would love work alongside, if you need me to do something feel free to reach out.

Elsword016:
# Update: what changed since the initial version


## 1. Storage Bucket support

The original upload path only knew regular dataset repositories. Large Zarr stores don't fit that model well: a single OME-Zarr image can easily contain hundreds of thousands of tiny chunk files (the old upload problem from [#4096](https://github.com/huggingface/datasets/issues/4096)), and dataset repos enforce per-directory file limits that make those uploads fail or crawl. The earlier version of this PR tried to solve that by rechunking stores into Zarr v3 with a `ShardingCodec` before uploading fewer, bigger files.

**What changed.** [`push_to_hub_zarr`](https://github.com/huggingface/datasets/blob/main/src/datasets/utils/zarr_utils.py) is now **bucket-first**. If you pass `repo_id=" buckets/<namespace>/<bucket_name>"`, the store is uploaded to an HF Storage Bucket instead, which is purpose-built for object-store-style data: resumable uploads and no per-directory file limits. The bucket is created automatically if it doesn't exist:

```python
from datasets.utils.zarr_utils import push_to_hub_zarr

push_to_hub_zarr(
    local_path="/path/to/sample.zarr",
    repo_id="buckets/username/my-zarr-data",
)
# 'hf://buckets/username/my-zarr-data/sample.zarr'
```

Automatic v3 sharding had several downsides: it rewrites every byte of the store (expensive for large images), it changes the store's chunk layout, which can surprise downstream tools that assume the original chunking, and it makes the uploaded artifact differ from what the user tested locally. Buckets solve the actual problem (file-count limits) without touching the data. So stores are now uploaded **as-is**: if a regular repo upload would exceed `file_limit` (default 10,000), we print a warning recommending a Storage Bucket and proceed with the resumable `upload_large_folder` backend instead of silently transforming the data. Malformed bucket ids are rejected early with a clear `buckets/<namespace>/<bucket_name>` error message.

Bucket-hosted stores stream directly through the `Zarr` feature, no download or auth required:

```python
from datasets import Dataset, Features, Zarr

ds = Dataset.from_dict(
    {"scan": ["hf://buckets/username/my-zarr-data/sample_001.zarr"]},
    features=Features({"scan": Zarr()}),
)
ds[0]["scan"].shape  # (1937, 2048, 2048)
```
---

## 2. NGFF 0.5 spec compliance: `default_plane` instead of `thumbnail`

The original preview used `thumbnail()`, which I assumed returned the store's "small 2D image". On the demo bucket that assumption turned out to be both wrong and slow: `thumbnail()` actually returns the **entire lowest-resolution level as an N-D array**

Resolution levels in [NGFF](https://ngff.openmicroscopy.org/) are N-D arrays (2–5 dimensions, in `time → channel → space` order), not 2D images; axes that aren't downsampled keep scale 1.0. There is **no "thumbnail" concept** in the spec, so there is no spec-defined 2D view. What the spec ecosystem does define is `omero.rdefs.defaultT` / `omero.rdefs.defaultZ`: the "default" 2D plane of the image.

A new `OmeZarrProxy.default_plane(level=-1)` method:

- reads `omero.rdefs` from the OME metadata,
- maps `defaultT` → time index and `defaultZ` → z index,
- rescales the z index when the level *has* downsampled z (e.g. `defaultZ=14` at full resolution becomes `z=7` at half-res), so the requested plane is the same physical plane at every level,
- falls back to `[0]` slices when `omero.rdefs` is absent, and works for 2D stores unchanged.

```python
plane = proxy.default_plane(level=-1)  # (64, 64) for this store, fetched lazily
plane.shape
# (64, 64)
```

`thumbnail()` is still available, but its docstring now explains the NGFF caveat so nobody trips on it again. Interactive previews (`_repr_html_` in notebooks) and the demo notebook use `default_plane` now.

---

## 3. Performance: parallel chunk reads, store reuse, chunk caching

All reads go through a `cached_array_getitem` path with three layers:

1. **Store registry** — one open handle per `(path, token, storage_options)`, reused across samples and batches instead of reopening the store per read. This also made the `token_per_repo_id` handling (private repos) uniform: the token is part of the registry key.
2. **Chunk LRU cache** — decoded chunks are cached in memory (256 MiB by default, configurable via `DATASETS_ZARR_CHUNK_CACHE_SIZE`), so overlapping patches across a batch fetch and decode each chunk once.
3. **Parallel fetch** — this is the one that fixed the preview hang. Zarr's `sync()` runs all reads on a single global IO loop, so sequential code was queuing up thousands of IO round-trips. Multi-chunk requests are now dispatched to a `ThreadPoolExecutor` (up to 8 workers, all sharing zarr's single loop, which is the intended usage) and results are reassembled in order. 

For training-style access there's also `ZarrCollator`, which samples random patches directly from a streaming dataset:

```python
from datasets import Zarr, load_dataset
from datasets.utils.zarr_utils import ZarrCollator

stream_ds = load_dataset("username/my-zarr-index", split="train", streaming=True)
stream_ds = stream_ds.cast_column("scan", Zarr())
train_ds = stream_ds.batch(batch_size=8).map(ZarrCollator(patch_size=(1, 128, 128)))
```

---

## 4. Hardening pass 

Fixed some of the failures left in the previous iterations:

- **`Zarr.encode_example` stored garbage for non-extractable stores.** When a zarr object had no extractable store path (e.g. an in-memory store), the encoder silently wrote `str(value)` into the Arrow file — a useless, misleading path. It now raises a clear `ValueError` instead of writing bad data.
- **Standalone proxies could not be unpickled.** `ZarrArrayProxy` / `ZarrGroupProxy` / `OmeZarrProxy` pickled fine but left their handles `None` on unpickle, which broke HF pipelines that pickle feature values. All three now reopen their store eagerly in `__setstate__` (via a shared `_reopen_zarr_node` helper that also validates the node type), so a pickled proxy is fully usable standalone.
- **`zarrfolder` used only the first split's metadata.** When a folder had multiple splits, feature columns were inferred from the first split only, so other splits could declare extra metadata columns and break at materialization. Columns are now merged as a **union across splits**.
- **Silently skipped or dangling metadata rows.** Rows without `file_name`/`zarr_file_name`, and rows pointing at `.zarr` paths that aren't among the discovered stores, are now skipped **with a warning** instead of failing or silently producing nothing. Metadata columns that aren't declared in the user's `features` are dropped with a warning too.
- **Remote `zarrfolder` now reads metadata files as the index.** `load_dataset("zarrfolder", data_dir="hf://buckets/...")` (and `s3://`, `file://`) previously returned bare rows; remote discovery ignored any `metadata.csv`/`metadata.jsonl`/`metadata.parquet` sitting next to the stores. Discovery now picks those files up, so a folder of stores with a metadata file needs **no separate manifest dataset**: the metadata file *is* the index, with one row per store joined on `file_name`.
- **Malformed bucket ids crashed deep in the upload path.** Now validated up front with a clear error message (covered above).

One known limitation is documented rather than fixed: the store-registry LRU can evict a store that a proxy still references (reference-counted eviction is a future improvement).

---

## 5. Demo notebook + docs

The notebook (`notebooks/zarr_demo.ipynb`) now walks the whole workflow with **live cells** against the public demo bucket: lazy loading, multiscale levels, `default_plane` previews, physical-coordinate ROI overlays, and matplotlib visualization cells (a 2×2 patch montage, a multiscale + physical-ROI figure, and a default-plane overlay with a scale-bar-accurate rectangle). [`docs/source/zarr_dataset.mdx`](https://github.com/huggingface/datasets/blob/main/docs/source/zarr_dataset.mdx) was rewritten for the bucket-first workflow: create a store → upload to a bucket or repo → build a lightweight index (manifest) dataset → `load_dataset(..., streaming=True)` and cast to `Zarr()`.

---

## 6. Branch hygiene

The branch has been cleaned up substantially: the history was rebuilt onto current `main` (the previous version carried ~80 stale commits from the fork's old lineage, including already-merged upstream work and other PRs' unmerged features), and upstream `main` has been merged in to clear the merge conflicts (config flags, packaged-module registry, and test decorators). The PR is now **18 commits + 1 merge commit**, fully mergeable, with CI pending. Test suite: **177 tests passing** across `tests/features/test_zarr.py`, `tests/packaged_modules/test_zarrfolder.py`, and `tests/utils/test_zarr_utils.py`, covering encoding/decoding, storage options, proxy semantics, OME-Zarr v0.4/v0.5, `default_plane`, parallel multi-chunk reads, pickle round-trips, metadata handling, and bucket upload validation.

---

## Still open/next steps

- As discussed earlier, I'd still be happy to consolidate the shared store-opening / fsspec / `hf://` helpers with [#7983](https://github.com/huggingface/datasets/pull/7983) so there's a single place to fix Zarr metadata edge cases — the two PRs are complementary in my view (`load_dataset("zarr", ...)` = store-as-dataset, `datasets.Zarr` = store-as-lazy-column-value).

Once again thanks to @lubianat for the demo bucket [`tiagolubiana/ome-vip-dataset-bucket`](https://huggingface.co/buckets/tiagolubiana/ome-vip-dataset-bucket) I used this to test and validate the features.

Happy to take more comments/feedbacks. I know the scope of the PR is quite big, so I would also be happy if I need to split.

docs/source/about_dataset_features.mdx
```diff
@@ -228,3 +228,44 @@ Another example with tool calling data and the `on_mixed_types="use_json"` argum
 >>> ds[0][1]["tool_calls"][0]["function"]["arguments"]
 {"room": "living room", "state": "on"}
 ```
+
+## Zarr feature
+
+Use [`Zarr`] for path-based lazy access to Zarr and OME-Zarr stores.
```
joshmoore: Moving forward it might be best to separate these two types. You can look at prior art with the napari plugins. There's a default (i.e. batteries included) Zarr plugin which is expecting solely an array. There's the napari-ome-zarr plugin which knows the full OME-Zarr spec. The behavior for an arbitrary Zarr group is not well-defined.

lubianat:
Thanks @Elswod016!

Maybe the best way forward is to divide this in more reviewable packages?

E.g. take #7983 and make it compatible with your modelling for Zarr as a whole (not OME-Zarr)

Trying to keep code/test to a size where other people could more easily review, even if not all the edges are covered.

After agreement on the basic modelling (e.g. working together directly with the author of #7983), the OME-Zarr/multiscales and other more specific features would be easier to review. Once the path is set for _Zarr_ support, it is easier to get _OME_ people to commit time IMO.

Something like https://docs.github.com/en/pull-requests/get-started/about-stacked-prs, even if that particular feature is not used.

In general, if the PR is smaller and Zarr as a whole, the chances of moving it forward faster are higher

joshmoore (commented):
Sorry, @Elsword016. I started reading but ran out of time. Perhaps you are already starting to address my main comment (below) in #7983 so better to send this now rather than wait. Best, ~J.

docs/source/about_dataset_features.mdx
Elsword016: Hi @joshmoore, thank you for the review:

So here's what I've done instead (will push it soon), which I think will match your suggestion:

- Trim and rewrite #8135 body down to the plain-Zarr foundation, exactly along the lines you suggested: the Zarr feature is now array-only, plus zarrfolder, plus #7983's zarr module folded in with (Co-authored-by credit preserved). The diff is roughly halved; OME-Zarr/multiscales and the upload/training tooling are deferred to stacked follow-up PRs.

- So even if #7983 merges first, my PR rebases trivially (the module files are identical; feature + zarrfolder then sit directly on top). If he reviews #8135 instead, everything is already in one place. Either path costs a single review.

I haven't got any response from KOKOSde, but when/if he returns, nothing is lost; his PR can still merge first, and I rebase, or the module splits back out; attribution is in the commit either way.

I don't know if I should nudge the HF maintainers on #7983 for a final pass on the May 3 changes and approving CI so it can run.

Elsword016:
I've restructured this PR along the lines discussed in the thread:

- Scope: I have reduced the scope to plain Zarr to serve as a foundation, which includes: the array-only Zarr feature (per Josh's comment), the zarrfolder builder, and the zarr module from #7983 (folded in with proper author credits). The diff is now reduced significantly. I have also rewritten the main body of the PR to reflect the changes. 
-
- Next steps: I will fold in the OME-Zarr/multiscales and other utilities like uploading/training once this PR is reviewed as stacked PRs. 

Since I haven't received any reply from the author on #7983, I've decided to move forward. If you'd prefer to merge #7983 first, I can rebase this PR. The module files are identical, and the feature + zarrfolder commits sit directly on top of them. Otherwise, everything plain-Zarr is here in one reviewable place.

@lubianat @joshmoore @lhoestq .It's ready for review, and I will be happy to adjust anything you need.

PR #8137, opened 2026-04-16
Fix: decode JSON type before to_list or to_dict is called

### LINKED ISSUE #8131: datasets.to_list: Behaviour change in version 4.7.0 with the addition of JSON type
### Describe the bug

There is a change in the approach to decoding JSON types, added in https://github.com/huggingface/datasets/commit/d560b58e1f1ec96591a95a1780ee118f5eea6c74. Since 4.7.0, to_list() returns raw JSON strings for columns stored as Json(), while direct access via __getitem__ returns dicts. 

In versions prior to 4.7.0, to_list() returned dicts.

1. Have datasets >= 4.7.0 installed (e.g 8.4.3)
2. Clear the datasets cache - important to reproduce.
3. Run the following:
```
ds = load_dataset("google/IFEval", split="train") 
print(type(ds[0]["kwargs"][0]))  # <class 'dict'>
print(type(ds.to_list()[0]["kwargs"][0]))  # <class 'str'>
```
With datasets < 4.7.0 (e.g 4.5.0) both return dict. 

(Taken from https://github.com/UKGovernmentBEIS/inspect_evals/issues/1416)

### Expected behavior

to_list() and to_dict() doesn't perfom feature decoding, but with the addition of the Json() type, there seems to be a regression where to_list() should return dicts but now no longer does. This would break older implementations where the code assumes data can be accessed this way. 

I'm not sure what fix would be the nicest/ in alignment with the dataset package. From broadest to narrowest options:
1. decode elements of the list in the same way _get_item does - (loading all images into memory)
  would be a heavy behaviour change
2. Do a target fix for Json()
```
  def to_list(self) -> list:                                                                                                                                                                         
      pylist = query_table(
          table=self._data,                                                                                                                                                                          
          key=slice(0, len(self)),                                                                                                                                                                 
          indices=self._indices,              
      ).to_pylist()
       
      # Identify columns that are JSON type with decode set to true         
      if self._info.features:                                                                                                                                                               
          json_columns = [
              col for col, feature in self._info.features.items()                                                                                                                                    
              if isinstance(feature, Json) and feature.decode                                                                                                                                      
          ]                   
          if json_columns:                
              for row in pylist:
                  for col in json_columns:                                                                                                                                                           
                      if col in row and isinstance(row[col], str):
                          row[col] = ujson_loads(row[col])                                                                                                                                           
      return pylist 
```
3. Put a warning message in to notify of this regression, to help with debugging. 

Happy to put a PR in with the preferred option.

### Environment info

- `datasets` version: 4.8.3
- Python version: 3.13.7
- `huggingface_hub` version: 1.4.1
- PyArrow version: 23.0.1
- Pandas version: 2.2.3
- `fsspec` version: 2025.9.0

ItsTania:
**Motivation**:
There is a change in the approach to decoding JSON types, added in https://github.com/huggingface/datasets/commit/d560b58e1f1ec96591a95a1780ee118f5eea6c74. Since 4.7.0, to_list() returns raw JSON strings for columns stored as Json(), while direct access via getitem returns dicts.

In versions prior to 4.7.0, to_list() returned dicts.

**Added context**:
- Closes https://github.com/huggingface/datasets/issues/8131
- On it, the maintainer suggested taking a similar approach to https://github.com/huggingface/datasets/pull/8122

This PR contains:
- An update to the to_list and to_dict functions for the arrow dataset inspired by a (and using the utils developed for) [similar fix](https://github.com/huggingface/datasets/pull/8122) to another problem caused by the addition of the JSON type.
- Unit tests specifying the intended outcome (which was how dictionaries were processed or handled prior to v4.7.0).

**Additional questions/considerations**:
- This PR restore behaviour prior to the breaking change introduced by the addition of the JSON datatype in v4.7.0. However, does mean we now decode one data type (JSON) and not others. We could decide to decode elements of the list in the same way _get_item does but this would be a big change (i.e memory concerns with images and audio etc etc)
- I implemented the PR with a lazy import - just checking that's in line with the repo's design/ vibe?

### DIFF
--- a/tests/test_arrow_dataset.py
+++ b/tests/test_arrow_dataset.py
@@ -37,7 +37,6 @@ from datasets.features import (
     Json,
     LargeList,
     List,
-    Sequence,
     Translation,
     TranslationVariableLanguages,
     Value,
--- a/tests/test_arrow_dataset.py
+++ b/tests/test_arrow_dataset.py
@@ -37,6 +37,7 @@ from datasets.features import (
     Json,
     LargeList,
     List,
+    Sequence,
     Translation,
     TranslationVariableLanguages,
     Value,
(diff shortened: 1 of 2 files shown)


MattFisher:
@lhoestq Could this be reviewed and merged if it's a suitable solution?

tests/test_arrow_dataset.py
```diff
@@ -37,6 +37,7 @@
     Json,
     LargeList,
     List,
+    Sequence,
```
lhoestq:
```suggestion
```

lhoestq (approved):
lgtm ! sorry for the delay

lhoestq:
`datasets` 4.8.5 is out and includes the fix ! :)

PR #8141, opened 2026-04-17
Fix Image feature to dynamically select pyarrow storage type for large binary/string inputs

### LINKED ISSUE #5717: Errror when saving to disk a dataset of images
### Describe the bug

Hello!

I have an issue when I try to save on disk my dataset of images. The error I get is:

```
  File "/home/jplu/miniconda3/envs/image-xp/lib/python3.10/site-packages/datasets/arrow_dataset.py", line 1442, in save_to_disk
    for job_id, done, content in Dataset._save_to_disk_single(**kwargs):
  File "/home/jplu/miniconda3/envs/image-xp/lib/python3.10/site-packages/datasets/arrow_dataset.py", line 1473, in _save_to_disk_single
    writer.write_table(pa_table)
  File "/home/jplu/miniconda3/envs/image-xp/lib/python3.10/site-packages/datasets/arrow_writer.py", line 570, in write_table
    pa_table = embed_table_storage(pa_table)
  File "/home/jplu/miniconda3/envs/image-xp/lib/python3.10/site-packages/datasets/table.py", line 2268, in embed_table_storage
    arrays = [
  File "/home/jplu/miniconda3/envs/image-xp/lib/python3.10/site-packages/datasets/table.py", line 2269, in <listcomp>
    embed_array_storage(table[name], feature) if require_storage_embed(feature) else table[name]
  File "/home/jplu/miniconda3/envs/image-xp/lib/python3.10/site-packages/datasets/table.py", line 1817, in wrapper
    return pa.chunked_array([func(chunk, *args, **kwargs) for chunk in array.chunks])
  File "/home/jplu/miniconda3/envs/image-xp/lib/python3.10/site-packages/datasets/table.py", line 1817, in <listcomp>
    return pa.chunked_array([func(chunk, *args, **kwargs) for chunk in array.chunks])
  File "/home/jplu/miniconda3/envs/image-xp/lib/python3.10/site-packages/datasets/table.py", line 2142, in embed_array_storage
    return feature.embed_storage(array)
  File "/home/jplu/miniconda3/envs/image-xp/lib/python3.10/site-packages/datasets/features/image.py", line 269, in embed_storage
    storage = pa.StructArray.from_arrays([bytes_array, path_array], ["bytes", "path"], mask=bytes_array.is_null())
  File "pyarrow/array.pxi", line 2766, in pyarrow.lib.StructArray.from_arrays
  File "pyarrow/array.pxi", line 2961, in pyarrow.lib.c_mask_inverted_from_obj
TypeError: Mask must be a pyarrow.Array of type boolean
```

My dataset is around 50K images, is this error might be due to a bad image?

Thanks for the help.

```python
dataset = load_dataset("imagefolder", data_dir="/path/to/dataset")
dataset["train"].save_to_disk("./myds", num_shards=40)
```

### Expected behavior

Having my dataset properly saved to disk.

### Environment info

- `datasets` version: 2.11.0
- Platform: Linux-5.15.90.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
- Python version: 3.10.10
- PyArrow version: 11.0.0
- Pandas version: 2.0.0

kasperg3:
Add dynamic pyarrow storage type selection to the Image feature so that pa.large_binary() and pa.large_string() inputs are preserved natively instead of being cast down to pa.binary() / pa.string()
Introduce _get_target_pa_type() to detect large types from incoming storage and _to_struct_storage() to build struct arrays with the correct field types
Both cast_storage and embed_storage now adapt to the input storage type

# Motivation
When PyArrow internally promotes columns to large_binary or large_string (which happens for large datasets), the Image feature would either attempt to cast them back down, failing for data exceeding 2 GB, or produce type mismatches that cause TypeError: Mask must be a pyarrow. Array of type boolean during embed_storage. This change makes the Image feature dynamically match the storage type of the input, avoiding both cast errors and type mismatches.

# Related Issues
Closes https://github.com/huggingface/datasets/issues/5717

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,9 @@
 *.lock
 !dvc.lock
 
+# Git worktrees
+.worktrees/
+
 # Extracted dummy data
 datasets/**/dummy_data-zip-extracted/
 
--- a/.gitignore
+++ b/.gitignore
@@ -2,9 +2,6 @@
 *.lock
 !dvc.lock
 
-# Git worktrees
-.worktrees/
-
 # Extracted dummy data
 datasets/**/dummy_data-zip-extracted/
 
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -1838,13 +1838,19 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         if self.list_indexes():
             raise ValueError("please remove all the indexes using `dataset.drop_index` before saving a dataset")
 
+        from .arrow_writer import get_arrow_writer_batch_size_from_features, get_writer_batch_size_from_data_size
+
+        dataset_nbytes = self._estimate_nbytes()
         if num_shards is None:
-            dataset_nbytes = self._estimate_nbytes()
             max_shard_size = convert_file_size_to_int(max_shard_size or config.MAX_SHARD_SIZE)
             num_shards = int(dataset_nbytes / max_shard_size) + 1
             num_shards = max(num_shards, num_proc or 1)
             # if we have only a few large samples, we should only create as many shards as samples
             num_shards = min(len(self.data), num_shards)
+
+        writer_batch_size = get_arrow_writer_batch_size_from_features(
+            self.features
+        ) or get_writer_batch_size_from_data_size(len(self), dataset_nbytes)
 
         fs: fsspec.AbstractFileSystem
         fs, _ = url_to_fs(dataset_path, **(storage_options or {}))
@@ -1898,6 +1904,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                 "shard": self.shard(num_shards=num_shards, index=shard_idx, contiguous=True),
                 "fpath": posixpath.join(dataset_path, f"data-{shard_idx:05d}-of-{num_shards:05d}.arrow"),
                 "storage_options": storage_options,
+                "writer_batch_size": writer_batch_size,
             }
             for shard_idx in range(num_shards)
         )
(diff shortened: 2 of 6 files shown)


lhoestq:
Hi ! if the goal to fix the linked issue maybe we can find a simpler approach.

While your PR seems to solve it, it needs extra careful attention to not add regressions for image datasets, in particular when it comes to data with mixes of both binary and large_binary

e.g.
- concatenating image datasets should always work without copies
- having a stateful feature type should not break stateless functions like the ones that append new record batches or row groups to arrow or parquet files

kasperg3:
I agree. I am currently performing additional tests, and I found that my implementation introduced a bug where the first batch in an Arrow shard is written before we know it's too large, which introduces a schema mismatch error.

I will continue to work on a solution that takes your notes into account. 

Thank you!

kasperg3:
I have been giving this PR a thorough consideration. The underlying errors that I see when I have arrow chunks larger than 2GB are a type mismatch between Array and ChunkedArray, which is caused by chunking the binaries

I think a proper solution is to make table.py chunk-aware, so that we allow embed_storage to return either pa.StructArray or pa.ChunkedArray of StructArray, and follow arrows recommendations: https://github.com/apache/arrow/blob/main/docs/source/cpp/arrays.rst#size-limitations-and-recommendations
This will result in embed_storage and table.py becoming quite a lot more complex, though. 

Alternatively, we can allow for parsing a large_binary bool save_to_disk() / push_to_hub(), and make the user opt in for the large_binary type at creation. This will require a binary to large_binary promotion when concatenating the datasets. This might be an alright middle ground without too many code changes, though it will break compatibility when opting in on large_binaries. 


I am unsure of all the edge cases that can happen, so I am not confident in commenting on the backward compatibility of these approaches. Let me know what you think @lhoestq.

lhoestq:
Hmm even if we return ChunkedArray, I think the ArrowWriter will try to combine_chunks() before writing the record batch

The original issue seems to be about save_to_disk specifically, maybe we can focus on that one. In particular it seems it currently uses a fixed batch size:

https://github.com/huggingface/datasets/blob/0e267a7c50f3618ae94bc5e9095a00b82570f197/src/datasets/arrow_dataset.py#L1941-L1951

This explains why in some cases a record batch can be >2GB. I guess a simple fix can be to adapt the batch size automatically using the same method as for push_to_hub:

https://github.com/huggingface/datasets/blob/0e267a7c50f3618ae94bc5e9095a00b82570f197/src/datasets/arrow_dataset.py#L5930-L5933

kasperg3:
Okay, I will revert the current changes, implement the changes, and test whether they solve my problem. 

Your suggestion seems to fix the issue where the write batches get too large, but it will not fix the issue when a single row is larger than 2GB. This is likely an extreme edge case. 

I will reopen and mark ready for review the PR when it is ready.

PR #8148, opened 2026-04-23
Add Apache Iceberg format support

frankliee:
# Add Apache Iceberg format support 

## Motivation

  Apache Iceberg is the most widely adopted open table format for data lakes, supported by Databricks,
  Snowflake, AWS Glue, Dremio, and others. A large amount of ML training data lives in Iceberg tables.
  Currently, users must manually export Iceberg data to Parquet before loading it into HuggingFace Datasets —
  this PR removes that friction.

fix this (https://github.com/huggingface/datasets/issues/7863)

##  Usage

  Users pass a pre-configured pyiceberg Catalog object and a table identifier:

  from pyiceberg.catalog.sql import SqlCatalog

  catalog = SqlCatalog("my_catalog", uri="sqlite:///catalog.db", warehouse="/tmp/warehouse")

## Basic loading
  ds = load_dataset("iceberg", catalog=catalog, table="db.my_table")

## Column selection + row filtering (predicate pushdown)
  ds = load_dataset("iceberg", catalog=catalog, table="db.my_table",
                    columns=["text", "label"],
                    filters=[("label", ">", 0)])

## Multiple splits from different tables
  ds = load_dataset("iceberg", catalog=catalog,
                    table={"train": "db.train", "test": "db.test"})

## Time travel via snapshot_id
  ds = load_dataset("iceberg", catalog=catalog, table="db.my_table",
                    snapshot_id=7051729674881785648)

## Streaming
  ds = load_dataset("iceberg", catalog=catalog, table="db.my_table", streaming=True)

  Works with any pyiceberg-supported catalog backend (REST, Hive, Glue, SQL, etc.) — the builder is agnostic
  to how the catalog is configured.

 ## Design decisions

  - Catalog object passed in, not constructed internally. Iceberg catalog configuration varies widely across
  backends (REST, Hive, Glue, SQL each have different auth/connection params). Rather than re-implementing a
  "catalog factory" inside the builder, users bring their own catalog — similar to how the sql builder accepts
   an existing SQLAlchemy connection. This keeps the builder simple and forward-compatible with new catalog
  types.
  - No _EXTENSION_TO_MODULE registration. Unlike file-based formats (Parquet, Lance, CSV), Iceberg tables are
  addressed via catalog + table identifier, not file extensions. Users must specify "iceberg" explicitly as
  the path argument.
  - create_config_id override for fingerprinting. Catalog objects (containing SQLAlchemy engines, connection
  pools, etc.) are not picklable by dill. The override replaces the catalog with a stable string
  representation ("{ClassName}_{name}") before hashing.
  - _CountableBuilderMixin for fast row counting. Uses scan.plan_files() metadata to count rows without
  reading data files.

### DIFF
--- a/src/datasets/packaged_modules/iceberg/iceberg.py
+++ b/src/datasets/packaged_modules/iceberg/iceberg.py
@@ -1,0 +1,173 @@
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Dict, List, Optional, Union
+
+import pyarrow as pa
+
+import datasets
+from datasets.builder import Key
+from datasets.features import Features
+from datasets.table import table_cast
+
+
+if TYPE_CHECKING:
+    from pyiceberg.catalog import Catalog
+    from pyiceberg.expressions import BooleanExpression
+    from pyiceberg.table import FileScanTask
+
+logger = datasets.utils.logging.get_logger(__name__)
+
+
+@dataclass
+class IcebergConfig(datasets.BuilderConfig):
+    """BuilderConfig for Apache Iceberg format.
+
+    Args:
+        catalog (`pyiceberg.catalog.Catalog`):
+            A pre-configured pyiceberg Catalog object.
+        table (`str` or `Dict[str, str]`):
+            Iceberg table identifier, e.g. ``"db.my_table"``.
+            Pass a dict to map split names to table identifiers,
+            e.g. ``{"train": "db.train", "test": "db.test"}``.
+        features (`Features`, *optional*):
+            Cast the data to these features.
+        columns (`List[str]`, *optional*):
+            List of columns to load; others are ignored.
+        filters (`str` or `BooleanExpression`, *optional*):
+            Row filter with predicate pushdown. Accepts a SQL-style string
+            (``"col > 1 AND col2 == 'foo'"``), or a pyiceberg
+            ``BooleanExpression`` object. Parsed by pyiceberg internally.
+        batch_size (`int`, defaults to ``131072``):
+            Number of rows per RecordBatch when reading.
+        snapshot_id (`int`, *optional*):
+            Load a specific snapshot for time-travel queries.
+    """
+
+    catalog: Optional["Catalog"] = None
+    table: Optional[Union[str, Dict[str, str]]] = None
+    features: Optional[datasets.Features] = None
+    columns: Optional[List[str]] = None
+    filters: Optional[Union[str, "BooleanExpression"]] = None
+    batch_size: int = 131072
+    snapshot_id: Optional[int] = None
+
+    def __post_init__(self):
+        super().__post_init__()
+        if self.catalog is None:
+            raise ValueError("`catalog` must be a pyiceberg Catalog object, but got None.")
+        if self.table is None:
+            raise ValueError("`table` must be specified, e.g. table='db.my_table'")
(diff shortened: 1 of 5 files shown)


src/datasets/packaged_modules/iceberg/iceberg.py
```diff
@@ -0,0 +1,138 @@
...
+
+            # Infer features from Arrow schema if not user-provided
+            if self.info.features is None:
+                arrow_schema = scan.projection().as_arrow()
+                self.info.features = datasets.Features.from_arrow_schema(arrow_schema)
+
+            splits.append(
+                datasets.SplitGenerator(
+                    name=split_name,
+                    gen_kwargs={"scan": scan},
```
lhoestq: Do you think we can have a list here instead ? This would enable parallel processing/streaming

e.g. one scan object per file maybe

lhoestq (commented):
Awesome ! I just have one comment:

src/datasets/packaged_modules/iceberg/iceberg.py
frankliee: Thanks for the suggestion! I've refactored the implementation to support num_proc > 1 parallel processing:

  Changes:

  1. List-based gen_kwargs: _split_generators now passes tasks = list(scan.plan_files()) as a list in gen_kwargs, which allows _split_gen_kwargs to distribute FileScanTask objects across
   workers automatically.
  2. Picklable scan_context: Instead of passing the unpicklable scan object, I extract a tuple of individually-serializable components (table_metadata, io, projected_schema, row_filter,
  case_sensitive, limit) that can reconstruct an ArrowScan reader in each worker.
  3. Drop catalog after use: At the end of _split_generators, both self.config.catalog and self.config_kwargs["catalog"] are set to None so the builder itself can be pickled when sent to
   child processes. The catalog is no longer needed after planning — all reading state lives in scan_context.
  4. Per-task reading: _generate_tables now iterates over its assigned tasks list and uses ArrowScan.to_record_batches([task]) for each one, yielding Key(task_idx, batch_idx) for proper
  shard-level parallelism.

frankliee: @lhoestq 
Thanks for the review! I’ve addressed this comment in the latest commit.
Could you please take another look when you have a chance?

tests/packaged_modules/test_iceberg.py
```diff
@@ -0,0 +1,157 @@
+import numpy as np
+import pyarrow as pa
+import pytest
+from pyiceberg.catalog.sql import SqlCatalog
+from pyiceberg.schema import Schema
+from pyiceberg.types import DoubleType, FloatType, ListType, LongType, NestedField, StringType
```
lhoestq: can you move the imports inside the test functions and decorate them with a `@require_pyiceberg` ? this way people can run the test suite even if they don't have all the test dependencies

lhoestq (commented):
Great ! I have one last comment before we merge:

tests/packaged_modules/test_iceberg.py
frankliee: Thanks, and I have added `@require_pyiceberg` and `@require_not_windows` on iceberg test cases, since iceberg tests does not support on window.

lhoestq (approved):
lgtm !

PR #7929, opened 2026-01-02
Raise early for invalid `revision` in `load_dataset`

Scott-Simmons:
Solves https://github.com/huggingface/datasets/issues/7928

Raise early for invalid revisions

PR #8156, opened 2026-04-26
Fix revision-aware cache lookup in load_dataset

### LINKED ISSUE #7928: `load_dataset` `revision` param not respected when fetching from cache
### Describe the bug

`datasets.load_dataset` `revision` semantics are a bit inconsistent when the dataset is not found on the huggingface hub. When fetching the latest cached version of the dataset, the `revision` argument is ignored, so long as any cached versions of the dataset already exist in the HF cache.

```python
import datasets
datasets.load_dataset(
    "sentientfutures/ahb",
    "dimensions",
    split="train",
    revision="main"
)
# would expect some error to raise here
datasets.load_dataset(
    "sentientfutures/ahb",
    "dimensions",
    split="train",
    revision="invalid_revision"
)
```

### Expected behavior

On the second call to `datasets.load_dataset` in the 'steps to reproduce the bug' example, expect something like:

```sh
raise DatasetNotFoundError(
datasets.exceptions.DatasetNotFoundError: Revision 'invalid_revision' doesn't exist for dataset 'sentientfutures/ahb' on the Hub.
```

### Environment info

- `datasets` version: 4.4.1
- Platform: Linux-6.2.0-39-generic-x86_64-with-glibc2.37
- Python version: 3.12.12
- `huggingface_hub` version: 0.36.0
- PyArrow version: 22.0.0
- Pandas version: 2.2.3
- `fsspec` version: 2025.9.0

Jatin-Shihora:
Fixes #7928

When `load_dataset()` falls back to the local cache, for example when offline or when the Hub is unreachable, it ignores the `revision` argument and simply returns the most recently modified cached version. This can lead to unexpected behavior when a specific revision is requested.

This happens because `CachedDatasetModuleFactory` neither accepted nor used the `revision` parameter, and `dataset_module_factory` did not pass it along either.

In this PR, I’ve updated the cache handling so that a specific revision can be respected when loading from cache.

In particular, SHA-based revisions are now matched against cached commit directories. If a matching commit is found, the corresponding cached version is returned. If not, a `DatasetNotFoundError` is raised, listing the available cached SHAs.

For branch or tag names such as `main` or `v1.0`, I was not able to reliably resolve them without access to the Hub. In such cases, a `DatasetNotFoundError` is raised with a clear message suggesting the use of a commit SHA instead.

The existing behavior is preserved when `revision=None`, meaning the latest cached version is still used by default.

Also, I fixed the inner `except Exception` block in `dataset_module_factory`. Previously, it would swallow `DatasetNotFoundError` raised by the cache logic and fall back to the outer Hub-related error. Now, the correct error is propagated.

Test I did:

- [x] Verified that a full SHA present in cache returns the correct version
- [x] Verified that a short SHA prefix resolves correctly when unique
- [x] Confirmed that missing SHAs raise `DatasetNotFoundError` with available SHAs listed
- [x] Confirmed that branch and tag names raise an error with an offline hint
- [x] Ensured that `revision=None` continues to return the latest cached version
- [x] Confirmed that `test_load_dataset_invalid_revision_with_cache` (from PR #7929) still passes


Jatin-Shihora:
Hi @lhoestq just a friendly follow-up on the review when you have a chance. Thanks!

ebarkhordar (commented):
The cache tree is `<namespace___name>/<config_id>/<version>/<hash>`, and for a Hub dataset the last component is the commit SHA (`DatasetModule(module_path, self.commit_hash, ...)`, load.py:717 and :791). So one commit produces one directory per cached config, and `matching_paths` counts configs, not commits:

```python
# cache: ns___name/mrpc/0.0.0/<sha>  and  ns___name/sst2/0.0.0/<sha>
CachedDatasetModuleFactory("ns/name", cache_dir=root, revision="a"*40).get_module()
# DatasetNotFoundError: Revision 'aaaa...' is ambiguous for dataset 'ns/name':
#   it matches multiple cached commits. Provide a longer commit SHA prefix.
```

The SHA is already the full 40 characters, so the suggested remedy does not exist, and any multi-config dataset (glue, xtreme, and so on) with two configs cached becomes unloadable offline at a revision that is present. `revision=None` still serves it. The same input with only `mrpc` cached returns the module fine, so the trigger is purely the second config.

Selecting on the last path component and then choosing the entry for the requested config would avoid it. That also removes the need to guess: `config_name` currently comes from `parts[-3]`, which is a config_id rather than a config name whenever config kwargs or custom features were used. Your own fixture in tests/test_load.py:1346 has that shape, and the value propagates:

```python
# cache: polinaeterna___audiofolder_two_configs_in_metadata/v2-374bfde4f55442bc/0.0.0/<sha>
# builder_kwargs -> {..., 'config_name': 'v2-374bfde4f55442bc', 'version': '0.0.0'}
```

The config there is named `v2`.

Both measured at 31eaeb0 in a clean container, calling `CachedDatasetModuleFactory.get_module()` directly. The cache trees were built by hand to the layout `_find_hash_in_cache` globs, since reaching the Hub was not an option here, so the directory shapes are worth a second look before you act on this. Neither existing test would show it: both are `@pytest.mark.integration` and use a single-config repo.

PR #8161, opened 2026-04-29
Pass library_name/version to HfApi in dataset push and delete paths

davanstrien:
Adds `library_name="datasets"` and `library_version=__version__` to 12 bare `HfApi(endpoint=..., token=...)` instantiations across `hub.py`, `arrow_dataset.py`, `iterable_dataset.py`, and `dataset_dict.py`. Commits made via `push_to_hub` / `delete_from_hub` currently report a User-Agent of `unknown/None; hf_hub/X.Y.Z; ...`; with this change they report as `datasets/X.Y.Z; ...`.

This completes a convention the repo already follows in `load.py` and `utils/file_utils.py`. Motivated by some recent UA-attribution work on the Hub side (happy to share more context on Slack).

Opened as draft to let CI verify nothing breaks before requesting review.

## Test plan
- [ ] CI green
- [ ] Optional manual smoke: `Dataset.from_dict({"x": [1]}).push_to_hub(...)` against a staging repo, confirm commit UA reports `datasets/<version>`.

cc @Wauplin to check this makes sense in your opinion!

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -75,7 +75,7 @@ from huggingface_hub.utils import EntryNotFoundError, HfHubHTTPError, Repository
 from packaging import version
 from tqdm.contrib.concurrent import thread_map
 
-from . import config
+from . import __version__, config
 from .arrow_reader import ArrowReader
 from .arrow_writer import ArrowWriter, OptimizedTypedSequence
 from .data_files import sanitize_patterns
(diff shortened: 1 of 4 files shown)


Wauplin (commented):
> cc @Wauplin to check this makes sense in your opinion!

Definitely makes sense yes!

To prevent from forgetting in the future, we could have a `_build_hf_api` helper in the Hub utilities that returns an instanciated `HfApi` with correct library/library_version (and parametrized endpoint/token)

lhoestq (approved):

PR #8166, opened 2026-04-30
Fix storage_options lookup for streaming Lance datasets

### LINKED ISSUE #8164: Lance: HF auth token dropped for private repos in streaming mode (broken `storage_options` lookup)
### Describe the bug

When `Lance._split_generators` receives `hf://` URIs - i.e. anywhere a `StreamingDownloadManager` is in play (`load_dataset(..., streaming=True)`, `get_dataset_split_names`, and the dataset-viewer worker) - the HF auth token is silently dropped before reaching `lance.dataset(...)`. Lance then attempts unauthenticated access, the Hub returns 401, and the user sees `LanceError(IO): Generic PermissionDenied`.

Non-streaming `load_dataset(repo, token=...)` is **not** affected: the regular `DownloadManager` materializes files to the local cache via `huggingface_hub` before `_split_generators` runs, so `files[0]` is a local path and `storage_options=None` becomes harmless. Public Lance repos and parquet-converted private repos are also unaffected.

**Root cause.** `src/datasets/packaged_modules/lance/lance.py`, line 121:

```python
storage_options = dl_manager.download_config.storage_options.get(files[0].split("://", 0)[0] + "://")
```

Two errors compound:

1. **`split("://", 0)`**: the second argument is `maxsplit`, and `0` means *do not split*. So `[0]` is the entire URL string, not the protocol. (Compare `sql.py:60`, which uses `split("://")[0] + "://"` correctly.)
2. **`+ "://"`**: even with the split fixed, the dict is keyed `"hf"`, not `"hf://"`. `download_config.py:78` writes `self.storage_options["hf"] = {...}`, and the same `lance.py` file uses the correct key on line 106: `storage_options.get("hf", {})`.

Concretely, with `files[0] = "hf://datasets/user/repo/data.lance/_versions/1-abc.manifest"`, the lookup key becomes `"hf://datasets/user/repo/data.lance/_versions/1-abc.manifest://"`, which never exists in `storage_options`, so `.get(...)` returns `None`. That `None` is then passed to both `lance.dataset(...)` (line 125) and `lance.file.LanceFileReader(...)` (line 150), so every code path through `_split_generators` is affected when files remain remote.

**Suggested fix:**

```diff
-            storage_options = dl_manager.download_config.storage_options.get(files[0].split("://", 0)[0] + "://")
+            storage_options = dl_manager.download_config.storage_options.get(files[0].split("://", 1)[0])
```

For `hf://` URIs the loader now retrieves `{"endpoint": "...", "token": "..."}` and forwards it to Lance — the dict shape Lance/object_store accepts for the `hf://` scheme. For local paths with no `://`, `split("://", 1)[0]` returns the full path and `.get(...)` returns `None`, matching today's behavior; no regression.

A regression test that calls `_split_generators` with a `StreamingDownloadManager` whose `download_config.storage_options["hf"]` is populated, and asserts the dict reaches a mocked `lance.dataset`, would prevent recurrence. `tests/packaged_modules/test_lance.py` currently only exercises local filesystem paths, which is why this slipped through CI.

**Scope.** Introduced in #7913 (initial Lance support, commit `06b6e020c`); affects every released `datasets` version with Lance support. User-facing impact: streaming reads of private Lance datasets, and the dataset-viewer on any private Lance dataset that hasn't been parquet-converted. Workaround: trigger parquet conversion on the dataset, or use non-streaming `load_dataset` and accept the local-cache materialization.

Requires a private Lance dataset on the Hub.

```python
ds = load_dataset(
    "user/private-lance-dataset",
    streaming=True,
    token="hf_...",
)
```

Result:

```
File ".../datasets/packaged_modules/lance/lance.py", line 125, in _split_generators
    lance_datasets = [lance.dataset(uri, storage_options=storage_options) for uri in lance_dataset_uris]
ValueError: LanceError(IO): Generic PermissionDenied error: PermissionDenied (permanent) at list,
  context: { uri: https://huggingface.co/api/datasets/user/private-lance-dataset/tree/main/<name>.lance/_versions?expand=True&recursive=True,
  response: Parts { status: 401, ..., "x-error-message": "Invalid username or password." } ... }
```

The same failure occurs through `get_dataset_split_names`, which is the path the dataset-viewer worker takes:

```python
from datasets import get_dataset_split_names
get_dataset_split_names("user/private-lance-dataset", token="hf_...")
# -> SplitsNotFoundError: The split names could not be parsed from the dataset config.
#    (caused by the same LanceError(IO): PermissionDenied above)
```

For comparison, the non-streaming form **works**, because files are downloaded to the local cache before `_split_generators` is called:

```python
load_dataset("user/private-lance-dataset", token="hf_...")  # works
```

### Expected behavior

In streaming mode and via `get_dataset_split_names`, the token from `download_config.storage_options["hf"]` should be forwarded to `lance.dataset(uri, storage_options=...)` so Lance can authenticate against the private repo, just as the non-streaming path does via the local cache.

### Environment info

- `datasets` version: 4.8.5
- Platform: Linux-6.1.0-44-amd64-x86_64-with-glibc2.42
- Python version: 3.12.13
- `huggingface_hub` version: 1.12.2
- PyArrow version: 24.0.0
- Pandas version: 3.0.2
- `fsspec` version: 2026.2.0

ericjaebeom:
In `Lance._split_generators`, the `storage_options` lookup used `files[0].split("://", 0)[0] + "://"`, which has two bugs that compound: `split("://", 0)` does not split at all (maxsplit=0), and the `storage_options` dict is keyed by the bare scheme (`"hf"`), not `"hf://"`. The lookup therefore always returned `None`, dropping the HF token before it reached `lance.dataset` and causing 401s for streaming reads of private Lance repos (and the dataset-viewer worker path via `get_dataset_split_names`).

Use `split("://", 1)[0]` to extract the scheme, matching the key written in `DownloadConfig` and the pattern used by `sql.py:60`.

Fixes #8164.

### DIFF
--- a/src/datasets/packaged_modules/lance/lance.py
+++ b/src/datasets/packaged_modules/lance/lance.py
@@ -118,7 +118,7 @@ class Lance(datasets.ArrowBasedBuilder, datasets.builder._CountableBuilderMixin)
 
         splits: list[datasets.SplitGenerator] = []
         for split_name, files in data_files.items():
-            storage_options = dl_manager.download_config.storage_options.get(files[0].split("://", 0)[0] + "://")
+            storage_options = dl_manager.download_config.storage_options.get(files[0].split("://", 1)[0])
 
             lance_dataset_uris = resolve_dataset_uris(files)
             if lance_dataset_uris:


lhoestq (approved):
good catch !

lhoestq:
CI failure is unrelated and is due to https://github.com/pandas-dev/pandas/issues/65462, merging

PR #8170, opened 2026-05-03
fix(map): fix progress bar exceeding total when load_from_cache_file=False

### LINKED ISSUE #8167: [BUG] Abnormal progress bar in dataset.map when load_from_cache_file=False
### Describe the bug

datasets                                 4.8.5

```python
from datasets import load_dataset, load_from_disk

data = load_from_disk('pretrain_cached_dataset/train')

def dummy_map(example):
    return example

dataset = data.map(dummy_map)
dataset = data.map(dummy_map, load_from_cache_file=True)   # works fine
dataset = data.map(dummy_map, load_from_cache_file=True, num_proc=4)   # works fine
dataset = data.map(dummy_map, load_from_cache_file=False)  # does not work

```

Image

Image

-

### Expected behavior

-

### Environment info

-

Nitin-Rajasekar:
## Summary

When `load_from_cache_file=False`, the progress bar in `Dataset.map()` displayed a count that exceeded the dataset size (e.g. "800 examples" for a 400-row dataset). 

The bug: `pbar_initial` was computed using `len(existing_cache_files)`, which counts cache files on disk regardless of whether they would be used. With `load_from_cache_file=False`, those files exist but are ignored, yet the bar was initialized to the full dataset size as if all work was already done. The actual processing then added on top of that, pushing the counter past the total and causing tqdm to drop the percentage display entirely.

Fix: use `num_shards - len(unprocessed_kwargs_per_job)` instead, which is 0 when `load_from_cache_file=False`, so the bar starts and counts correctly.

## Issue

Fixes #8167

## Local verification

`py -3.14 -m pytest tests/test_arrow_dataset.py -k "test_map_load_from_cache_file_false_progress_bar_starts_at_zero" -v`

2 passed

## Risk

Low risk, a one-line change to how `pbar_initial` is computed. No behaviour changes outside of the progress bar display.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -3591,7 +3591,8 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                 )
 
             pbar_total = len(self)
-            pbar_initial = len(existing_cache_files) * pbar_total // num_shards
+            num_shards_done = num_shards - len(unprocessed_kwargs_per_job)
+            pbar_initial = num_shards_done * pbar_total // num_shards
             if batched and drop_last_batch:
                 batch_size = batch_size or 1
                 pbar_initial = pbar_initial // num_shards // batch_size * num_shards * batch_size
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -3500,8 +3500,6 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
 
             num_shards = min(existing_cache_file_map, key=select_existing_cache_files)
 
-        existing_cache_files = existing_cache_file_map[num_shards]
-
         def format_cache_file_name(
             cache_file_name: Optional[str],
             rank: Union[int, Literal["*"]],  # noqa: F722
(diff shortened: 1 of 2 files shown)


Jintao-Huang:
hello! any update?

Nitin-Rajasekar:
> hello! any update?

Fixed a formatting issue and pushed an update. The workflows are awaiting approval...

lhoestq (approved):
lgtm !

Nitin-Rajasekar:
Hey @lhoestq, thanks for the approval!

I looked into the 2 failing CI checks, and they seem unrelated to this PR..

- The `deps-latest` failures are caused by a breaking change in `huggingface_hub`, which no longer accepts single-segment repo IDs like `"dummy"` and `"paws"`. This affects existing tests rather than the change introduced here.
* The Windows runner also shows additional `401 Unauthorized` errors, which appear to be CI/environment-related.

The `deps-minimum` variants pass successfully, and my test (`test_map_load_from_cache_file_false_progress_bar_starts_at_zero`) passes in all runs.

Would you be able to merge this PR? I’m happy to make any additional changes if needed.

lhoestq:
sure, the errors sound unrelated to this pr indeed :)

PR #8207, opened 2026-05-18
fix docs: use AutoImageProcessor for vision model examples

wali-reheman:
swaps AutoFeatureExtractor for AutoImageProcessor in the image preprocessing section of use_dataset.mdx — aligns with the transformers library direction where vision feature extractor references were already replaced in their own docs (see PRs #20111 and #20501). audio section is untouched since AutoFeatureExtractor is still correct there.


vitoriall:
Thanks for the clear explanation! Good catch keeping AutoFeatureExtractor 
in the audio section since it's still valid there. The vision section 
update looks correct to me.

Issue #6913, opened 2024-05-22
Column order is nondeterministic when loading from JSON


albertvillanova:
As reported by @meg-huggingface, the order of the JSON object keys is not preserved while loading a dataset from a JSON file with a list of objects.

For example, when loading a JSON files with a list of objects, each with the following ordered keys:
- [ID, Language, Topic], 

the resulting dataset may have columns:
- [ID, Topic, Language], or
- [Topic, Language, ID], or
- [Topic, ID, Language],...

This issue is caused by the use of a Python set (which does not preserve the order):
https://github.com/huggingface/datasets/blob/60d21efbc01e15d0b596ac1072750cbecd91548a/src/datasets/packaged_modules/json/json.py#L168
introduced in
- #5772

PR #8209, opened 2026-05-18
Preserve float columns when JSON loader uses `field=`

### LINKED ISSUE #6937: The JSON loader implicitly coerces floats to integers
The JSON loader implicitly coerces floats to integers.

The column values `[0.0, 1.0, 2.0]` are coerced to `[0, 1, 2]`.
See CI error in dataset-viewer: https://github.com/huggingface/dataset-viewer/actions/runs/9290164936/job/25576926446
```
 =================================== FAILURES ===================================
___________________________ test_statistics_endpoint ___________________________

normal_user_public_json_dataset = 'DVUser/tmp-dataset-17170199043860'

    def test_statistics_endpoint(normal_user_public_json_dataset: str) -> None:
        dataset = normal_user_public_json_dataset
        config, split = get_default_config_split()
        statistics_response = poll_until_ready_and_assert(
            relative_url=f"/statistics?dataset={dataset}&config={config}&split={split}",
            check_x_revision=True,
            dataset=dataset,
        )
    
        content = statistics_response.json()
        assert len(content) == 3
        assert sorted(content) == ["num_examples", "partial", "statistics"], statistics_response
        statistics = content["statistics"]
        num_examples = content["num_examples"]
        partial = content["partial"]
    
        assert isinstance(statistics, list), statistics
        assert len(statistics) == 6
        assert num_examples == 4
        assert partial is False
    
        string_label_column = statistics[0]
        assert "column_name" in string_label_column
        assert "column_statistics" in string_label_column
        assert "column_type" in string_label_column
        assert string_label_column["column_name"] == "col_1"
        assert string_label_column["column_type"] == "string_label"  # 4 unique values -> label
        assert isinstance(string_label_column["column_statistics"], dict)
        assert string_label_column["column_statistics"] == {
            "nan_count": 0,
            "nan_proportion": 0.0,
            "no_label_count": 0,
            "no_label_proportion": 0.0,
            "n_unique": 4,
            "frequencies": {
                "There goes another one.": 1,
                "Vader turns round and round in circles as his ship spins into space.": 1,
                "We count thirty Rebel ships, Lord Vader.": 1,
                "The wingman spots the pirateship coming at him and warns the Dark Lord": 1,
            },
        }
    
        int_column = statistics[1]
        assert "column_name" in int_column
        assert "column_statistics" in int_column
        assert "column_type" in int_column
        assert int_column["column_name"] == "col_2"
        assert int_column["column_type"] == "int"
        assert isinstance(int_column["column_statistics"], dict)
        assert int_column["column_statistics"] == {
            "histogram": {"bin_edges": [0, 1, 2, 3, 3], "hist": [1, 1, 1, 1]},
            "max": 3,
            "mean": 1.5,
            "median": 1.5,
            "min": 0,
            "nan_count": 0,
            "nan_proportion": 0.0,
            "std": 1.29099,
        }
    
        float_column = statistics[2]
        assert "column_name" in float_column
        assert "column_statistics" in float_column
        assert "column_type" in float_column
        assert float_column["column_name"] == "col_3"
>       assert float_column["column_type"] == "float"
E       AssertionError: assert 'int' == 'float'
E         - float
E         + int

tests/test_14_statistics.py:72: AssertionError

FAILED tests/test_14_statistics.py::test_statistics_endpoint - AssertionError: assert 'int' == 'float'
  - float
  + int
``` 

This bug was introduced after:
- #6914

We have reported the issue to pandas:
- https://github.com/pandas-dev/pandas/issues/58866

LeSingh1:
Closes #6937.

When `load_dataset("json", data_files=..., field="data", ...)` is used, columns whose values are all integer-valued floats (`[0.0, 1.0, 2.0]`) get silently coerced to `int64`. Repro:

```python
import tempfile, json

with tempfile.NamedTemporaryFile(mode="w", suffix=".json", delete=False) as f:
    json.dump({"data": [{"col": v} for v in [0.0, 1.0, 2.0]]}, f)

ds = load_dataset("json", data_files=f.name, field="data", split="train")
print(ds.features)  # before: {'col': Value('int64')}
                    # after:  {'col': Value('float64')}
```

The underlying cause is `pd.read_json(..., dtype_backend="pyarrow")` (tracked upstream at pandas-dev/pandas#58866). The path that hits it is the one introduced in #6914 to preserve column insertion order. The dataset-viewer statistics regression in the issue is a direct consequence.

This PR replaces the `pandas.read_json` -> `pa.Table.from_pandas` path in the `field=` branch with a small `_arrow_table_from_field` helper that builds the Arrow table directly from the already-parsed Python object. PyArrow's own JSON inference preserves `float64`, and CPython dict iteration order gives us the `#6914` column-insertion-order invariant for free (no pandas roundtrip needed).

The helper handles:

- list of dicts (the common case): collect keys in insertion order, build a column-major dict
- list of scalars: wrap in a single-column table named after the configured feature, falling back to "text" (matches the prior `df.columns == [0]` rename)
- dict of lists (column-major payload): pass through to `pa.Table.from_pydict`
- empty list with features supplied: emit empty columns matching the configured feature names so downstream `_cast_table` aligns

The other JSON loading path (raw JSON Lines via `paj.read_json`) does not have this bug and is unchanged.

There is an older dormant PR for this issue (#7635 from June 2025, no reviews). I went a different direction because that PR scans the resulting DataFrame and re-casts float-looking int columns back to float after the fact. By that point pandas has already converted the values to Python ints, so the `isinstance(x, float)` check there does not actually detect them, and the scan adds an O(rows) Python pass to a hot path. Sidestepping pandas entirely is simpler and faster.

Tests in `tests/packaged_modules/test_json.py`:

- `test_json_field_path_preserves_float_columns`: the exact #6937 repro, asserts `col` stays float64
- `test_json_field_path_preserves_float_columns_alongside_ints`: list-of-dicts payload with mixed int / float / int-valued-float columns, each type preserved
- `test_json_field_path_preserves_float_columns_with_dict_of_lists`: dict-of-lists field payload
- `test_json_field_path_preserves_column_order_with_list_of_dicts`: column insertion order check (sanity vs the #6914 invariant)

`pytest tests/packaged_modules/test_json.py` shows 33 passed, 0 failed (29 pre-existing + 4 new).

### DIFF
--- a/src/datasets/packaged_modules/json/json.py
+++ b/src/datasets/packaged_modules/json/json.py
@@ -4,7 +4,7 @@ import os
 from dataclasses import dataclass
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import Literal, Optional
+from typing import Any, Literal, Optional
 
 import pandas as pd
 import pyarrow as pa
@@ -36,6 +36,48 @@ def pandas_read_json(path_or_buf, **kwargs):
     if datasets.config.PANDAS_VERSION.major >= 2:
         kwargs["dtype_backend"] = "pyarrow"
     return pd.read_json(path_or_buf, **kwargs)
+
+
+def _arrow_table_from_field(
+    dataset: Any, features: Optional["datasets.Features"] = None, field: Optional[str] = None
+) -> pa.Table:
+    # Build directly from the parsed Python object instead of routing through
+    # pandas.read_json. Pandas with the pyarrow dtype_backend coerces columns
+    # like [0.0, 1.0, 2.0] to int64, dropping the float semantics (#6937,
+    # pandas-dev/pandas#58866). PyArrow's own inference preserves float64,
+    # and CPython dict key iteration order already gives us the
+    # column-insertion-order invariant from #6914.
+    if isinstance(dataset, list):
+        if not dataset:
+            if features is not None:
+                return pa.Table.from_pydict({name: [] for name in features})
+            return pa.Table.from_pydict({})
+        if not isinstance(dataset[0], dict):
+            # List of scalars; mirror the prior `df.columns == [0]` rename.
+            if features is None:
+                return pa.Table.from_pydict({"text": dataset})
+            # A scalar list is a single column, so an over-specified `features`
+            # cannot be satisfied. Pandas used to catch this when the rename
+            # assigned to `df.columns`; keep it an error rather than returning
+            # the extra columns as all-null.
+            if len(features) != 1:
+                raise ValueError(
+                    f"Field {field!r} is a list of scalars, which maps to a single column, "
+                    f"but `features` declares {len(features)}: {list(features)}"
+                )
+            return pa.Table.from_pydict({next(iter(features)): dataset})
+        keys: dict[str, None] = {}
+        for row in dataset:
+            for key in row.keys():
+                keys.setdefault(key, None)
+        column_order = list(keys)
+        mapping = {col: [row.get(col) for row in dataset] for col in column_order}
+        return pa.Table.from_pydict(mapping)
+    if isinstance(dataset, dict):
+        return pa.Table.from_pydict(dataset)
+    raise ValueError(
+        f"Cannot build a table from the JSON field at type {type(dataset).__name__}"
+    )
 
 
 class FullReadDisallowed(Exception):
@@ -162,10 +204,9 @@ class Json(datasets.ArrowBasedBuilder):
                         dataset = ujson_loads(f.read())
                     # We keep only the field we are interested in
                     dataset = dataset[self.config.field]
-                    df = pandas_read_json(io.StringIO(ujson_dumps(dataset)))
-                    if df.columns.tolist() == [0]:
-                        df.columns = list(self.config.features) if self.config.features else ["text"]
-                    pa_table = pa.Table.from_pandas(df, preserve_index=False)
+                    pa_table = _arrow_table_from_field(
+                        dataset, self.config.features, self.config.field
+                    )
                     yield Key(shard_idx, 0), self._cast_table(pa_table)
 
                 # If the files are agent traces (one row = one file except for hermes which can have multiple sessions per file)
--- a/tests/packaged_modules/test_json.py
+++ b/tests/packaged_modules/test_json.py
@@ -807,3 +807,184 @@ def test_json_load_dataset_without_droid_marker_stays_ordinary_json(tmp_path):
 
     assert dataset.column_names == ["type", "id", "version", "timestamp", "message"]
     assert dataset[0]["type"] == "session_start"
+
+
+# Regression tests for #6937 (float-to-int coercion in the field= JSON loader).
+
+
+@pytest.fixture
+def json_file_with_integer_valued_floats_field(tmp_path):
+    path = tmp_path / "file_with_int_valued_floats_field.json"
+    data = textwrap.dedent(
+        """\
+        {
+            "data": [
+                {"col": 0.0},
+                {"col": 1.0},
+                {"col": 2.0}
+            ]
+        }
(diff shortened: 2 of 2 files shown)


ebarkhordar (commented):
I ran this at `6a301ab` against its base `b7cb10b` in a clean container: 22 JSON shapes through `field=`, the 20 list-shaped ones again through the no-field path as a control, and 7 more with an explicit `features=`. Three of those 49 comparisons differ. Everything else matches, including nested structs, ragged keys, column order, ints past 2^53 and the dict-of-lists form.

| input under `field="data"` | base | this PR |
|---|---|---|
| `[{"col":0.0},{"col":1.0},{"col":2.0}]` | `int64` | `float64` |
| `[{"col":1.0},{"col":null}]` | `int64` | `float64` |
| `[1,2,3]` with `features` of 2 fields | `ValueError` | 3 rows, second column all null |

The third row is the one I would want a decision on. On base, `df.columns = list(self.config.features)` enforced an arity match, so an over-specified `features=` surfaced as `ValueError: Length mismatch: Expected axis has 1 elements, new values have 2 elements`. `_arrow_table_from_field` takes `next(iter(features))` and ignores the rest, so the declared column that cannot be filled now comes back silently null instead:

```
features=Features({"num": Value("int64"), "extra": Value("string")})
# base: DatasetGenerationError <- ValueError: Length mismatch ...
# head: [{'num': 1, 'extra': None}, {'num': 2, 'extra': None}, {'num': 3, 'extra': None}]
```

It needs a scalar list under `field=` plus an over-specified `features=`, so the reach is narrow, but it turns a loud error into quiet null data. If dropping that check is deliberate, a comment saying so would help; if not, raising on the scalar-list branch when `len(features) > 1` restores it.

Two smaller notes.

`test_json_field_path_preserves_float_columns_alongside_ints` constructs `Json()` with no `field=`, so it covers the path this PR does not touch. Running the four new tests against base source, that one and the column-order test both pass; only `preserves_float_columns` and `preserves_float_columns_with_dict_of_lists` fail. The other two do not guard the change.

Row 2 of the table is a fix beyond the int-valued-float case in #6937 and currently has no test. `[{"col": 1.0}, {"col": null}]` gives `int64` on base and `float64` here, which is the same silent corruption in a shape the issue does not mention.

Verification: both trees on `PYTHONPATH` in a stock `python:3.11-slim` image, nothing installed from the source trees, `datasets.__file__` confirmed pointing at each tree per run. `tests/packaged_modules/test_json.py` is 28 passed, 11 skipped at the PR head. I did not exercise any streaming or remote path, or a multi-file shard set.

LeSingh1:
@ebarkhordar this is a genuinely useful review — thank you for running it properly. All three findings were right and are fixed in fd006fe.

**Row 3, the silent nulls.** A real regression I introduced. `df.columns = list(features)` was doing double duty as an arity check, and `next(iter(features))` dropped it. Not deliberate, so I restored it as an explicit error rather than a comment:

```
ValueError: Field 'data' is a list of scalars, which maps to a single column,
but `features` declares 2: ['num', 'extra']
```

**The test that did not guard.** Correct — `preserves_float_columns_alongside_ints` built `Json()` with no `field=`, so it exercised the untouched path. Its fixture is now wrapped in a `data` field and the test constructs `Json(field="data")`.

**Row 2, floats with nulls.** Added `preserves_float_columns_with_nulls` for `[{"col": 1.0}, {"col": null}]`.

Teeth-checked by reverting only `json.py`: 5 of the 6 now fail on base, including `alongside_ints` and both new tests. The sixth, `preserves_column_order_with_list_of_dicts`, passes on base by design — it is not a regression test for #6937, it guards my rewrite against breaking the #6913 ordering invariant. Happy to rename it if that is unclear.

Full file: 30 passed, 11 skipped.

PR #8215, opened 2026-05-21
Warn instead of raise when user-provided data_files yields a subset

### LINKED ISSUE #7867: NonMatchingSplitsSizesError when loading partial dataset files
### Describe the bug

When loading only a subset of dataset files while the dataset's README.md contains split metadata, the system throws a NonMatchingSplitsSizesError . This prevents users from loading partial datasets for quick validation in cases of poor network conditions or very large datasets.

1. Use the Hugging Face `datasets` library to load a dataset with only specific files specified
2. Ensure the dataset repository has split metadata defined in README.md
3. Observe the error when attempting to load a subset of files

```python
# Example code that triggers the error

book_corpus_ds = load_dataset(
    "SaylorTwift/the_pile_books3_minus_gutenberg",
    name="default",
    data_files="data/train-00000-of-00213-312fd8d7a3c58a63.parquet",
    split="train",
    cache_dir="./data"
)
```

### Error Message
```
  File "/Users/QingGo/code/llm_learn/src/data/clean_cc_bc.py", line 13, in <module>
    book_corpus_ds = load_dataset(
        "SaylorTwift/the_pile_books3_minus_gutenberg",
    ...
  File "/Users/QingGo/code/llm_learn/.venv/lib/python3.13/site-packages/datasets/utils/info_utils.py", line 77, in verify_splits
    raise NonMatchingSplitsSizesError(str(bad_splits))
datasets.exceptions.NonMatchingSplitsSizesError: [{'expected': SplitInfo(name='train', num_bytes=106199627990.47722, num_examples=192661, shard_lengths=None, dataset_name=None), 'recorded': SplitInfo(name='train', num_bytes=454897326, num_examples=905, shard_lengths=None, dataset_name='the_pile_books3_minus_gutenberg')}]
```

### Expected behavior

When loading partial dataset files, the system should:
1. Skip the `NonMatchingSplitsSizesError` validation, OR
2. Only log a warning message instead of raising an error

### Environment info

- `datasets` version: 4.3.0
- Platform: macOS-15.7.1-arm64-arm-64bit-Mach-O
- Python version: 3.13.2
- `huggingface_hub` version: 0.36.0
- PyArrow version: 22.0.0
- Pandas version: 2.3.3
- `fsspec` version: 2025.9.0

adityasingh2400:
Fixes #7867.

NonMatchingSplitsSizesError currently fires whenever the loaded split size differs from the expected size, including when the user explicitly passed data_files for a known subset of the dataset. The only user-side workaround is verification_mode='no_checks', which silences ALL checks rather than just the split-size one.

This PR downgrades the split-size mismatch to a UserWarning when data_files was explicitly provided by the caller. Other mismatch paths (corrupted download, wrong size for full download) remain hard errors.

### DIFF
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -949,7 +949,11 @@ class DatasetBuilder:
             dl_manager.manage_extracted_files()
 
         if verification_mode == VerificationMode.BASIC_CHECKS or verification_mode == VerificationMode.ALL_CHECKS:
-            verify_splits(self.info.splits, split_dict)
+            verify_splits(
+                self.info.splits,
+                split_dict,
+                user_provided_data_files=getattr(self.config, "data_files", None) is not None,
+            )
 
         # Update the info object with the splits.
         self.info.splits = split_dict
--- a/src/datasets/utils/info_utils.py
+++ b/src/datasets/utils/info_utils.py
@@ -1,6 +1,5 @@
 import enum
 import os
-import warnings
 from typing import Optional
 
 from huggingface_hub.utils import insecure_hashlib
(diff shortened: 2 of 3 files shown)


lhoestq:
I think we can simply not raise or warn anything at all when the user provides the data_files. The number of examples is expected to be likely different

adityasingh2400:
Makes sense, done in 2162efc. verify_splits now returns early when the user passed data_files, so there is no raise and no warning at all in that case. The early return also covers the split name checks, since custom data_files can define a different set of splits. I removed the warning path and updated the tests accordingly.

PR #8223, opened 2026-05-27
DOC: document missing save_infos and max_depth parameters

kratos0718:
## What this fixes

Two public API functions were missing parameters from their docstrings:

### 1. `load_dataset()` — missing `save_infos`

The `save_infos` parameter has been in the function signature but was absent from the `Args` section of the docstring. When `save_infos=True`, it overrides `verification_mode` and forces it to `VerificationMode.ALL_CHECKS`, running full checksums/size/splits verification.

### 2. `Dataset.flatten()` — missing `max_depth`

The `max_depth` parameter (default `16`) controls how many nesting levels are flattened, but was not documented in the `Args` section.

## Files changed

- `src/datasets/load.py` — added `save_infos` to `load_dataset` docstring
- `src/datasets/arrow_dataset.py` — added `max_depth` to `Dataset.flatten` docstring

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -2322,6 +2322,9 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
             new_fingerprint (`str`, *optional*):
                 The new fingerprint of the dataset after transform.
                 If `None`, the new fingerprint is computed using a hash of the previous fingerprint, and the transform arguments.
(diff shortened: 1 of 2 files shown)


kratos0718:
Friendly ping — this has been open since 27 May.

It documents two parameters that already exist in the signature but were missing from the docstring (`save_infos` and `max_depth`), so it's docs-only with no behaviour change and still applies cleanly.

Entirely understand if docs-only PRs sit low in the queue — happy to close it if you'd rather not carry it.

PR #8228, opened 2026-05-30
fix: stable fingerprint for closures capturing non-deterministic state

### LINKED ISSUE #7986: `Dataset.map()` causes cache miss/fingerprint change when closure captures self containing non-deterministic state.

kamaleshpanda:
Fixes #7986

When `.map()` uses a closure capturing `self` with non-deterministic state 
(like UUIDs or loggers), the fingerprint changes every run causing cache misses.

Fix: only hash primitive attributes of captured objects instead of the full object.

Added a regression test in `tests/test_fingerprint.py`.

### DIFF
--- a/src/datasets/fingerprint.py
+++ b/src/datasets/fingerprint.py
@@ -7,7 +7,7 @@ import weakref
 from functools import wraps
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Optional, Union
-
+import types
 import numpy as np
 import xxhash
 
@@ -208,9 +208,39 @@ class Hasher:
         for x in value:
             m.update(x)
         return m.hexdigest()
-
+    @staticmethod
+    def _hash_callable(func):
+        code = func.__code__
+        freevars = {}
+        if func.__closure__:
+            for name, cell in zip(code.co_freevars, func.__closure__):
+                try:
+                    val = cell.cell_contents
+                # if the free variable is an object (like self),
+                # only keep its __dict__ filtered to simple types
+                    if hasattr(val, "__dict__"):
+                        freevars[name] = {
+                        k: v for k, v in val.__dict__.items()
+                        if isinstance(v, (int, float, str, bool, type(None)))
+                        }
+                    else:
+                        freevars[name] = val
+                except ValueError:
+                    freevars[name] = "<empty cell>"
+        return (code.co_code, code.co_consts, code.co_varnames, freevars)
+    
     @classmethod
     def hash(cls, value: Any) -> str:
+        if (
+        isinstance(value, types.FunctionType)
+        and value.__closure__
+        and any(
+            hasattr(cell.cell_contents, "__dict__")
+            for cell in value.__closure__
+            if hasattr(cell, "cell_contents")
+        )
+        ):
+            value = cls._hash_callable(value)
         return cls.hash_bytes(dumps(value))
 
     def update(self, value: Any) -> None:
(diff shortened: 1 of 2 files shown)


lhoestq:
I feel like this would cause headaches to many people, causing cache reuses when the function actually changed

kamaleshpanda:
You're right, thank you for pointing that out. The current approach is too 
broad — ignoring all non-primitive attributes could cause incorrect cache 
reuse when meaningful objects like tokenizers change between runs.

What approach would you suggest? Should we instead warn the user when a 
closure captures a non-deterministic object, or only apply this when the 
user explicitly opts in?

PR #8229, opened 2026-06-01
Skip origin metadata fetch when streaming=True

### LINKED ISSUE #8197: Feature Request / Discussion: Skip origin_metadata fetch when streaming=True?
Hi team!

Thanks!

TheDerbiedOne:
Fixes #8197

When using `load_dataset(..., streaming=True, data_files=...)`, `DataFilesList.from_patterns` currently fetches origin metadata (ETag/mtime) for each resolved file even though the streaming pipeline doesn't use the local cache/fingerprint path.

This PR adds an opt-out (`skip_origin_metadata`) for data-file resolution and uses it in `load_dataset_builder(..., streaming=True)` so we skip the origin metadata fetch in streaming mode, reducing init latency and avoiding unnecessary network calls.

Includes a unit test that asserts `_get_origin_metadata` is not called for streaming.

### DIFF
--- a/src/datasets/data_files.py
+++ b/src/datasets/data_files.py
@@ -581,10 +581,16 @@ class DataFilesList(list[str]):
         base_path: Optional[str] = None,
         allowed_extensions: Optional[list[str]] = None,
         download_config: Optional[DownloadConfig] = None,
+        *,
+        skip_origin_metadata: bool = False,
     ) -> "DataFilesList":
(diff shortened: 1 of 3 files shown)


yuxin00j:
Hi! Thanks for putting this PR together. Following up on the discussion in #8197 with the maintainers, instead of only applying this automatically for streaming, we should expose skip_origin_metadata=False as a public parameter (e.g., passing it through load_dataset into DatasetBuilder). Could you amend the PR?

Issue #5805, opened 2023-04-28
Imrove `Create a dataset` tutorial 


polinaeterna:
Our [tutorial on how to create a dataset](https://huggingface.co/docs/datasets/create_dataset) is a bit misleading. 
1. In **Folder-based builders** section it says that we have two folder-based builders as standard builders, but we also have similar builders (that can be created from directory with data of required format) for `csv`, `json/jsonl`, `parquet` and `txt` files. We have info about these loaders in separate [guide for loading](https://huggingface.co/docs/datasets/loading#local-and-remote-files) but it's worth briefly mentioning them in the beginning tutorial because they are more common and for consistency. Would be helpful to add the link to the full guide.
2. **From local files** section lists methods for creating a dataset from in-memory data which are also described in [loading guide](https://huggingface.co/docs/datasets/loading#inmemory-data).  

Maybe we should actually rethink and restructure this tutorial somehow.

isunitha98selvan:
I can work on this. The link to the tutorial seems to be broken though @polinaeterna.

benjaminbrown038:
#self-assign

polinaeterna:
@isunitha98selvan would be great, thank you! which link are you talking about? I think it should work: https://huggingface.co/docs/datasets/create_dataset

AmboThom:
Hey @polinaeterna, I have a pull request for this issue. Can you review and see if it needs any changes?

zanvari:
Hi, I'd be interested in working on this issue if it is still available. I can review the tutorial and propose documentation improvements to clarify the relationship between dataset creation and dataset loading workflows.

krishna2426:
is this assigned , i would like to work on this

LeonxLJX:
I'd like to take this one (`Improve `Create a dataset` tutorial`). I'll look into the root cause and follow up with a PR. (claiming via @LeonxLJX)

AmboThom:
Welcome to the

LeonxLJX:
I'd like to take this one (`Improve `Create a dataset` tutorial`). I'll look into the root cause and follow up with a PR. (claiming via @LeonxLJX)

ShamikOfficial:
Hi! Opening a PR for this issue.

Updated the create dataset tutorial to mention csv/json/parquet/txt folder builders and cross-link the loading and repository structure guides.

#self-assign

LeonxLJX:
[...]

PR #8235, opened 2026-06-04
Clarify dataset creation vs loading workflows in create_dataset tutorial

zanvari:
This PR clarifies the distinction between dataset creation and dataset loading workflows in the create_dataset tutorial.

Changes:
- Added a note directing users to the loading guide.
- Clarified that `from_` methods create datasets from local Python objects rather than local files.

Addresses #5805.


lhoestq (approved):
thx !

PR #8236, opened 2026-06-04
Fix iterable skip over full Arrow blocks

my17th2:
## What does this PR do?

Fixes `IterableDataset.skip(n)` for streaming datasets when the underlying iterable uses Arrow batches and `n` skips one or more complete Arrow blocks.

Previously, after a full Arrow block was counted as skipped, `_iter_arrow()` continued into the partial-slice branch and yielded rows from a block that should have been fully skipped.

## What was the issue?

`SkipExamplesIterable._iter_arrow()` handles skipping in two cases:

1. the current Arrow table is fully skipped
2. only the beginning of the current Arrow table is skipped, and the rest is yielded

The bug was that case 1 did not stop after marking the table as skipped. So the same table then fell through into case 2.

In other words, a table could first be counted as "already skipped", but then still be sliced and yielded.

For example, if Arrow tables have 4 rows each and we call `skip(6)`:

- table 1 has rows `[0, 1, 2, 3]` and should be fully skipped
- table 2 has rows `[4, 5, 6, 7]`, so only `[4, 5]` should be skipped and `[6, 7]` should be yielded

Before this PR, after table 1 was counted as skipped, the code kept processing table 1 and yielded part of it. This is why skipped rows could appear in the output.

This PR adds `continue` after a table is fully skipped, so the code moves directly to the next Arrow table.

## Tests

- `PYTHONPATH=src pytest tests/test_iterable_dataset.py::test_skip_arrow_examples_iterable -q`

The regression test covers skipping within a block, exactly one block, across blocks, and beyond the dataset length.

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -2041,6 +2041,7 @@ class SkipExamplesIterable(_BaseExamplesIterable):
                 skipped += len(pa_table)
                 if self._state_dict:
                     self._state_dict["skipped"] = skipped
+                continue
             if skipped + 1 <= self.n:
                 offset = self.n - skipped
                 skipped = self.n
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -2041,8 +2041,7 @@ class SkipExamplesIterable(_BaseExamplesIterable):
                 skipped += len(pa_table)
                 if self._state_dict:
                     self._state_dict["skipped"] = skipped
-                continue
-            if skipped + 1 <= self.n:
+            elif skipped + 1 <= self.n:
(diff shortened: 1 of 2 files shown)


src/datasets/iterable_dataset.py
```diff
@@ -2041,6 +2041,7 @@ def _iter_arrow(self):
                 skipped += len(pa_table)
                 if self._state_dict:
                     self._state_dict["skipped"] = skipped
+                continue
             if skipped + 1 <= self.n:
```
lhoestq:
```suggestion
            elif skipped + 1 <= self.n:
```

lhoestq (approved):
lgtm ! applying a minor change for consistency with take()

PR #8243, opened 2026-06-05
Fix JSON loader schema inference for files starting with a UTF-8 BOM (#8241)

### LINKED ISSUE #8241: load_dataset("json", ...) infers a different schema when the JSONL file starts with a UTF-8 BOM
### Describe the bug

When a JSONL file starts with a UTF-8 BOM (`\xef\xbb\xbf`), `datasets.load_dataset("json", data_files=...)` infers a different schema than when the same file is loaded without the BOM. Specifically, the loader's "mixed-struct-types" pre-scan — which normally promotes columns with heterogeneous nested keys to a `Json` extension type so each dict is preserved as-is — is silently skipped, and PyArrow's normal schema unification kicks in instead. This materializes `null` values for fields that were absent in the source data, changing the user-visible content of records.

The root cause is in `datasets/packaged_modules/json/json.py` (the JSON builder). On the first batch, the loader parses each line with `ujson_loads` to detect mixed struct schemas:

```python
try:
    examples = [ujson_loads(line) for line in batch.splitlines()]
except ValueError:
    # the file is likely not JSON Lines and may contain one single multi-line JSON object
    pass
else:
    json_field_paths += find_mixed_struct_types_field_paths(examples)
```

`ujson` rejects the leading BOM with `ValueError('Expected object or value')`, the broad `except ValueError: pass` swallows it, and `find_mixed_struct_types_field_paths` is never called. PyArrow's `paj.read_json`, on the other hand, tolerates the BOM and goes on to parse the file successfully — but with a unified `list<struct<...>>` schema where every element must conform to the same struct, producing spurious `null`s for absent fields.

This is a silent correctness issue: no warning is logged, and downstream code that checks `key in record` (or that is sensitive to extra `None` values, e.g. validators or API payloads) will behave differently depending on whether the file happens to start with a BOM.

Two minimal JSONL files with identical content — except `bom.jsonl` begins with a UTF-8 BOM.
**`nobom.jsonl`**

```json
{"id": 1, "tags": [{"name": "a"}, {"name": "b", "weight": 0.5}]}
{"id": 2, "tags": [{"name": "c"}]}
{"id": 3, "tags": [{"name": "d", "weight": 0.9}]}
```

**`bom.jsonl`** is byte-for-byte identical except for the leading 3 bytes:

```text
$ xxd bom.jsonl   | head -1
00000000: efbb bf7b 2269 6422 3a20 312c 2022 7461  ...{"id": 1, "ta
$ xxd nobom.jsonl | head -1
00000000: 7b22 6964 223a 2031 2c20 2274 6167 7322  {"id": 1, "tags"
```
**Repro script:**

```python
import pathlib
import datasets

HERE = pathlib.Path(__file__).parent

for name in ("nobom", "bom"):
    ds = datasets.load_dataset(
        "json",
        data_files={"train": str(HERE / f"{name}.jsonl")},
    )["train"]
    print(f"=== {name} ===")
    print("schema  :", ds.features)
    print("row 0   :", ds[0])
    print("tags[0] :", ds[0]["tags"][0])
    print()
```

**Output:**

```text
=== nobom ===
schema  : {'id': Value('int64'), 'tags': List(Json(decode=True))}
row 0   : {'id': 1, 'tags': [{'name': 'a'}, {'name': 'b', 'weight': 0.5}]}
tags[0] : {'name': 'a'}

=== bom ===
schema  : {'id': Value('int64'), 'tags': List({'name': Value('string'), 'weight': Value('float64')})}
row 0   : {'id': 1, 'tags': [{'name': 'a', 'weight': None}, {'name': 'b', 'weight': 0.5}]}
tags[0] : {'name': 'a', 'weight': None}
```

Note that the BOM version injects `'weight': None` into the first tag, even though the source line contains `{"name": "a"}` with no `weight` key.

[repro.zip](https://github.com/user-attachments/files/28643388/repro.zip)

### Expected behavior

A leading UTF-8 BOM should not change the inferred schema or the content of loaded records. Either:

1. Strip the BOM (or otherwise normalize the input) before running the mixed-struct-types pre-scan, so the same code path is taken regardless of BOM, **or**
2. At minimum, log a warning when the pre-scan is skipped due to a `ValueError`, so silently divergent schema inference doesn't go unnoticed.

In both files above, the expected result is the `nobom` output: `tags` inferred as `List(Json(decode=True))` and each dict preserved exactly as written.

### Environment info

```text
- `datasets` version: 4.8.5
- Platform: Linux-6.6.87.2-microsoft-standard-WSL2-x86_64-with-glibc2.39
- Python version: 3.12.3
- `huggingface_hub` version: 1.12.2
- PyArrow version: 21.0.0
- Pandas version: 2.3.3
- `fsspec` version: 2026.2.0
```

archievi:
Fixes #8241.

## What's the bug

When a JSONL file starts with a UTF-8 BOM (`\xef\xbb\xbf`), `load_dataset("json", ...)` infers a different schema than the same file without the BOM.

The JSON builder's first-batch "mixed-struct-types" pre-scan parses each line with `ujson` to decide whether to preserve heterogeneous nested dicts as a `Json` type:

```python
try:
    examples = [ujson_loads(line) for line in batch.splitlines()]
except ValueError:
    pass
else:
    json_field_paths += find_mixed_struct_types_field_paths(examples)
```

`ujson` rejects the leading BOM with `ValueError`, the broad `except ValueError` swallows it, and `find_mixed_struct_types_field_paths` is never called. PyArrow's `read_json` tolerates the BOM and instead unifies the schema, materializing spurious `null`s for fields absent from individual records. So the user-visible content silently changes depending on whether the file happens to start with a BOM.

## Fix

Strip a leading UTF-8 BOM from the first batch (guarded to the start of the file) so the pre-scan and PyArrow operate on the same bytes.

## Test

Added a regression test: a BOM-prefixed variant of the existing `jsonl_file_with_lists_of_dicts_of_varying_keys` fixture, asserting it produces the same schema (`EXPECTED_LISTS_OF_DICTS_WITH_VARYING_KEYS`) as the non-BOM file. It fails on `main` and passes with this change.

`ruff check`, `ruff format`, and `pytest tests/packaged_modules/test_json.py::test_json_generate_tables` all pass locally.

### DIFF
--- a/src/datasets/packaged_modules/json/json.py
+++ b/src/datasets/packaged_modules/json/json.py
@@ -1,3 +1,4 @@
+import codecs
 import io
 import os
 from dataclasses import dataclass
@@ -208,6 +209,13 @@ class Json(datasets.ArrowBasedBuilder):
                             batch = f.read(self.config.chunksize)
                             if not batch:
                                 break
+                            # A leading UTF-8 BOM makes the ujson pre-scan below raise
+                            # ValueError, silently skipping mixed-struct detection, while
+                            # PyArrow tolerates the BOM -- so the inferred schema differed
+                            # depending on whether the file started with a BOM. Strip it so
+                            # both see the same bytes (a BOM only appears at the start of the file).
+                            if shard_idx == 0 and batch_idx == 0 and batch.startswith(codecs.BOM_UTF8):
+                                batch = batch[len(codecs.BOM_UTF8) :]
                             if batch.startswith(b"["):
                                 if not allow_full_read:
                                     raise FullReadDisallowed()
--- a/tests/packaged_modules/test_json.py
+++ b/tests/packaged_modules/test_json.py
@@ -252,6 +252,22 @@ def jsonl_file_with_lists_of_dicts_of_varying_keys(tmp_path):
     return str(filename)
 
 
+@pytest.fixture
+def jsonl_file_with_lists_of_dicts_of_varying_keys_and_bom(tmp_path):
+    # Same content as jsonl_file_with_lists_of_dicts_of_varying_keys, but the file
+    # starts with a UTF-8 BOM (written via the utf-8-sig codec).
+    filename = tmp_path / "file_with_bom.jsonl"
+    data = textwrap.dedent(
+        """\
+        {"col_1": [{"a": 0}, {"b": 0}]}
+        {"col_1": [{"c": 0}, {"d": 0}]}
+        """
+    )
+    with open(filename, "w", encoding="utf-8-sig") as f:
+        f.write(data)
+    return str(filename)
+
+
 _messages = [
     {"role": "user", "content": "Turn on the living room lights and play my electronic music playlist."},
     {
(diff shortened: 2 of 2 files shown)


lhoestq (approved):
lgtm !

PR #8249, opened 2026-06-08
Catch RecursionError when decoding deeply nested mesh files (#8246)

### LINKED ISSUE #8246: RecursionError when loading GLB file with deeply nested JSON in Mesh feature
## Description

Loading a GLB file with deeply nested JSON chunk causes a RecursionError, crashing the application.

## Step to Reproduce

```python
from datasets import Dataset, Mesh
import tempfile, os, struct

def make_malicious_glb(depth=5000):
    json_chunk = (b'{"a":' * depth + b'1' + b'}' * depth)
    magic = b'glTF'
    version = struct.pack('<I', 2)
    json_len = struct.pack('<I', len(json_chunk))
    json_type = struct.pack('<I', 0x4E4F534A)
    total_len = 12 + 8 + len(json_chunk)
    length = struct.pack('<I', total_len)
    return magic + version + length + json_len + json_type + json_chunk

glb_bytes = make_malicious_glb(depth=5000)

with tempfile.TemporaryDirectory() as tmpdir:
    fpath = os.path.join(tmpdir, 'test.glb')
    with open(fpath, 'wb') as f:
        f.write(glb_bytes)
    
    ds = Dataset.from_dict({'mesh': [fpath]}).cast_column('mesh', Mesh())
    result = ds[0]['mesh']
```

**Expected:** Proper error message
**Actual:** RecursionError crash

## Suggested Fix

Catch RecursionError in Mesh.decode_example() and raise ValueError instead.

archievi:
Fixes #8246.

A GLB whose glTF JSON chunk is deeply nested (malformed or malicious) makes `trimesh.load` exceed Python's recursion limit while parsing it (`RecursionError: maximum recursion depth exceeded while decoding a JSON array`), crashing `Mesh.decode_example` with an uncaught error.

This routes both `trimesh.load` call sites through a small `_load_mesh` helper that catches `RecursionError` and re-raises it as a clear `ValueError`, so a malformed/adversarial mesh produces an actionable error instead of crashing the whole load. `htmlTreeAsString`-style decoding for valid meshes is unchanged.

Added a regression test that builds a GLB with a deeply-nested JSON chunk and asserts `decode_example` raises `ValueError` (it fails with the uncaught `RecursionError` without the fix).

Verified locally: `ruff check` / `ruff format` clean, and the new test passes.

### DIFF
--- a/src/datasets/features/mesh.py
+++ b/src/datasets/features/mesh.py
@@ -108,9 +108,7 @@ class Mesh:
         if not self.decode:
             raise RuntimeError("Decoding is disabled for this feature. Please use Mesh(decode=True) instead.")
 
-        if config.TRIMESH_AVAILABLE:
-            import trimesh
-        else:
+        if not config.TRIMESH_AVAILABLE:
             raise ImportError("To support decoding meshes, please install 'trimesh'.")
 
         if token_per_repo_id is None:
@@ -124,7 +122,7 @@ class Mesh:
                 file_type = _infer_mesh_file_type(path)
                 if file_type is None:
                     raise ValueError("A mesh path should have a .glb, .ply, or .stl extension.")
-                return trimesh.load(path, file_type=file_type)
+                return _load_mesh(path, file_type)
             source_url = path.split("::")[-1]
             pattern = (
                 config.HUB_DATASETS_URL if source_url.startswith(config.HF_ENDPOINT) else config.HUB_DATASETS_HFFS_URL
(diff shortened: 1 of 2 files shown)


archievi:
Gentle bump on this one — it's a small, self-contained fix for the `RecursionError` on deeply nested mesh files (#8246), with a regression test, and CI is green. Happy to rebase or adjust if anything's needed. Thanks for taking a look!

PR #8254, opened 2026-06-09
Prefer torchvision to decode images instead of PIL

NicolasHug:
This PR uses `torchvision.io.decode_image` instead of `PIL.open()` to decode images when the `torch` formatter is used.

TorchVision is typically faster than PIL to decode images, and this also saves the `pil_to_tensor` conversion. See https://github.com/huggingface/transformers/pull/45195 for benchmarks on `transfomers` where the same changes were applied.

----

Very open to any feedback regarding the implem. I went for creating a new `TorchFeaturesDecoder` class which mirrors what exists for `pandas`. There are other ways to enable this, like making the `Image` class 'formatter-aware'.

As discussed offline, I considered making `Image.decode_example()` return a lazy `PIL.Image` (no .load()) so the torch formatter could extract the path/bytes and pass them to torchvision. But this doesn't work cleanly because `PIL.Image.open(path)` without .load() keeps OS file handles open, which caused https://github.com/huggingface/datasets/issues/3985. Additionally, EXIF/mode destroys lazy state:[ `decode_example()` needs to apply EXIF transpose and mode conversion](https://github.com/NicolasHug/datasets/blame/e88a484cf05869e13f535e097ff116ffb00b2608/src/datasets/features/image.py#L194-L198). Both trigger `.load()` and produce a new image that loses filename and fp. So we can't both do EXIF/mode AND keep the image lazy.

### DIFF
--- a/src/datasets/features/features.py
+++ b/src/datasets/features/features.py
@@ -1474,7 +1474,7 @@ def decode_nested_example(
     Args:
         decoder_overrides: optional dict mapping feature types to custom decode functions.
             When the python type of a leaf feature matches a key, the override function is called
-            as ``override_fn(obj, schema)`` instead of ``schema.decode_example(obj, ...)``.
+            as ``override_fn(obj, schema, token_per_repo_id=...)`` instead of ``schema.decode_example(obj, ...)``.
     """
     _kwargs = {"token_per_repo_id": token_per_repo_id, "decoder_overrides": decoder_overrides}
     # Nested structures: we allow dict, list/tuples, sequences
@@ -1519,7 +1519,7 @@ def decode_nested_example(
             return None
         if decoder_overrides:
             if override_fn := decoder_overrides.get(type(schema)):
-                return override_fn(obj, schema)
+                return override_fn(obj, schema, token_per_repo_id=token_per_repo_id)
         # we pass the token to read and decode files from private repositories in streaming mode
         return schema.decode_example(obj, token_per_repo_id=token_per_repo_id)
     return obj
--- a/src/datasets/features/features.py
+++ b/src/datasets/features/features.py
@@ -1462,17 +1462,28 @@ def encode_nested_example(schema, obj, level=0):
     return obj
 
 
-def decode_nested_example(schema, obj, token_per_repo_id: Optional[dict[str, Union[str, bool, None]]] = None):
+def decode_nested_example(
+    schema, obj, token_per_repo_id: Optional[dict[str, Union[str, bool, None]]] = None, decoder_overrides=None
+):
     """Decode a nested example.
     This is used since some features (in particular Audio and Image) have some logic during decoding.
 
     To avoid iterating over possibly long lists, it first checks (recursively) if the first element that is not None or empty (if it is a sequence) has to be decoded.
     If the first element needs to be decoded, then all the elements of the list will be decoded, otherwise they'll stay the same.
+
+    Args:
+        decoder_overrides: optional dict mapping feature types to custom decode functions.
+            When a leaf feature matches a key (via isinstance), the override function is called
+            as ``override_fn(obj, schema)`` instead of ``schema.decode_example(obj, ...)``.
     """
+    _kwargs = {"token_per_repo_id": token_per_repo_id, "decoder_overrides": decoder_overrides}
     # Nested structures: we allow dict, list/tuples, sequences
     if isinstance(schema, dict):
         return (
-            {k: decode_nested_example(sub_schema, sub_obj) for k, (sub_schema, sub_obj) in zip_dict(schema, obj)}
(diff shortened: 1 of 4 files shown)


lhoestq:
Sounds good ! Maybe a way to make it work even on List(Image()) would be to integrate it in decode_nested_example() directly (or copy and adapt decode_nested_example())

E.g. decode_nested_example() could accept an argument that makes it use `_decode_image_for_torch` for Image() type instead of `Image.decode_example`

This way the TorchDecoder can call `decode_nested_example(self.features[column_name], value, overrides={Image: _decode_image_for_torch})` or something like that

Other that that looks good to me ! Excited by the speed ups this will bring to dataloaders 🚀 

cc @vasqu for viz, this speeds up datasets loaded with

```python
ds = load_dataset(...).with_format("torch")
# much faster than before
for image_tensor in ds["image"]:
    ...
```

src/datasets/features/features.py
```diff
@@ -1471,17 +1462,28 @@ def encode_nested_example(schema, obj, level=0):
...
 
     To avoid iterating over possibly long lists, it first checks (recursively) if the first element that is not None or empty (if it is a sequence) has to be decoded.
     If the first element needs to be decoded, then all the elements of the list will be decoded, otherwise they'll stay the same.
+
+    Args:
+        decoder_overrides: optional dict mapping feature types to custom decode functions.
+            When a leaf feature matches a key (via isinstance), the override function is called
+            as ``override_fn(obj, schema)`` instead of ``schema.decode_example(obj, ...)``.
     """
+    _kwargs = {"token_per_repo_id": token_per_repo_id, "decoder_overrides": decoder_overrides}
```
NicolasHug, line 1479: Note the new `token_per_repo_id` which is passed down now - I think this is a drive-by bug fix, but not entirely sure

src/datasets/features/features.py
```diff
@@ -1494,27 +1496,33 @@ def decode_nested_example(schema, obj, token_per_repo_id: Optional[dict[str, Uni
...
             return list(obj)
     # Object with special decoding:
     elif hasattr(schema, "decode_example") and getattr(schema, "decode", True):
+        if obj is None:
+            return None
+        if decoder_overrides:
+            for type_, override_fn in decoder_overrides.items():
+                if isinstance(schema, type_):
+                    return override_fn(obj, schema)
         # we pass the token to read and decode files from private repositories in streaming mode
```
lhoestq, line 1523: simpler ?

```suggestion
            if override_fn := decoder_overrides.get(type(schema)):
                return override_fn(obj, schema)
```

we don't do feature types subclasses so it's fine

src/datasets/features/features.py
```diff
@@ -1471,17 +1462,28 @@ def encode_nested_example(schema, obj, level=0):
...
+):
     """Decode a nested example.
     This is used since some features (in particular Audio and Image) have some logic during decoding.
 
     To avoid iterating over possibly long lists, it first checks (recursively) if the first element that is not None or empty (if it is a sequence) has to be decoded.
     If the first element needs to be decoded, then all the elements of the list will be decoded, otherwise they'll stay the same.
+
+    Args:
+        decoder_overrides: optional dict mapping feature types to custom decode functions.
+            When a leaf feature matches a key (via isinstance), the override function is called
```
lhoestq:
```suggestion
            When the python type of leaf feature matches a key, the override function is called
```

src/datasets/formatting/torch_formatter.py
```diff
@@ -15,24 +15,138 @@
...
+        value: dict with "bytes" and "path" keys from Arrow extraction
+        image_feature: the Image feature instance (carries mode/decode info)
+
+    Returns:
+        torch.Tensor in CHW format
+    """
+    import torch
+    from torchvision.io import decode_image
+
+    path, bytes_ = value["path"], value["bytes"]
```
lhoestq, line 65: sometimes bytes_ is None and in this case we need to read the bytes from path

see the code in Image.decode_example:

https://github.com/huggingface/datasets/blob/11947f5f44ea9a8cb101dd4d32c8a5675104d0ff/src/datasets/features/image.py#L170-L190

(I missed that in my first review sry)

lhoestq (commented):
Great ! I left some additional comments to address and I then I think we can merge

src/datasets/formatting/torch_formatter.py
NicolasHug: I think we handle both bytes and local paths just below within the `try:` block. Did you mean that we should also handle the case of non-local paths?

lhoestq: Ah yes local path is handled indeed. So only remote path is missing: you can use xopen() to support remote paths

PR #8255, opened 2026-06-09
Support hermes traces

lhoestq:
Parse hermes agent traces export using teich

supports both single session export and multi-sessions export

```bash
# single session
>>> hermes sessions export session.jsonl --session-id 20250305_091523_a1b2c3d4
# multi-sessions
>>> hermes sessions export sessions.jsonl
```

+ bump teich to 0.1.2

breaking change: "trace" is now Json() instead of List(Json())

example using [dacorvo/hf-hub-session-hermes-traces](https://huggingface.co/datasets/dacorvo/hf-hub-session-hermes-traces):

```python
In [1]: from datasets import load_dataset

In [2]: ds = load_dataset("dacorvo/hf-hub-session-hermes-traces", split="train")

In [3]: ds
Out[3]: 
Dataset({
    features: ['harness', 'session_id', 'prompt', 'messages', 'tools', 'metadata', 'sent_at', 'num_user_messages', 'num_tool_calls', 'trace', 'file_path'],
    num_rows: 228
})
```

this also makes me think we should document more how to use `teich` with `datasets`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -194,7 +194,7 @@ TESTS_REQUIRE = [
     "torchcodec>=0.7.0; python_version < '3.14'",  # minium version to get windows support, torchcodec doesn't have wheels for 3.14 yet
     "nibabel>=5.3.1",
     "trimesh>=4.10.0",
-    "teich==0.1.1a76",
+    "teich==0.1.2",
 ]
 
 NUMPY2_INCOMPATIBLE_LIBRARIES = [
--- a/src/datasets/packaged_modules/json/json.py
+++ b/src/datasets/packaged_modules/json/json.py
@@ -1,6 +1,7 @@
 import io
 import os
 from dataclasses import dataclass
+from datetime import datetime, timezone
 from pathlib import Path
 from typing import Literal, Optional
 
@@ -145,7 +146,7 @@ class Json(datasets.ArrowBasedBuilder):
             if self.info.features == AGENT_TRACES_FEATURES:
                 is_agent_traces = True
                 if datasets.config.TEICH_AVAILABLE:
-                    from teich import convert_trace_to_training_example
+                    from teich import convert_traces_to_training_data
                 else:
                     raise ImportError("To support decoding agent traces, please install 'teich'.")
             json_field_paths = get_json_field_paths_from_feature(self.info.features)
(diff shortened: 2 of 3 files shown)


cfahlgren1 (approved):
tested with https://huggingface.co/datasets/cfahlgren1/hermes-agent-trace-samples-2026-06-05

cfahlgren1:
do we want to use teich for trace detection too to simplify somethings?

```python
from teich import detect_trace_type
```

will just want to make sure it works well for existing trace datasets

lhoestq:
I think it doesn't differentiate openclaw and pi yet so I'm keeping it like this for now, but in the next update I plan to use detect_trace_type and use a fork/PR of teich. They can surely be interested in this improvement too

cc @armand0e for viz

armand0e:
Thanks for bringing me in the loop here.

Hermes and Openclaw were the two biggest unknowns I had for teich, given the lack of working trace examples on huggingface. If this native hermes trace support works:
```
# single session
>>> hermes sessions export session.jsonl --session-id 20250305_091523_a1b2c3d4
# multi-sessions
>>> hermes sessions export sessions.jsonl
```
then teich should be working as expected with hermes. As for openclaw I just never got around to that implementation yet, mainly because I'm not familar with how openclaw does their session storage, but if it's just a native trace format I can at least implement it into `detect_trace_type` to avoid that pi detection mixup. Thanks again!

cfahlgren1:
> Thanks for bringing me in the loop here.
> 
> Hermes and Openclaw were the two biggest unknowns I had for teich, given the lack of working trace examples on huggingface. If this native hermes trace support works:
> 
> ```
> # single session
> >>> hermes sessions export session.jsonl --session-id 20250305_091523_a1b2c3d4
> # multi-sessions
> >>> hermes sessions export sessions.jsonl
> ```
> 
> then teich should be working as expected with hermes. As for openclaw I just never got around to that implementation yet, mainly because I'm not familar with how openclaw does their session storage, but if it's just a native trace format I can at least implement it into `detect_trace_type` to avoid that pi detection mixup. Thanks again!

Yeah openclaw uses Pi under the hood. We just check for the openclaw directory within the Pi cwd to label it differently. Everything else remains the same as Pi.

https://github.com/huggingface/datasets/blob/35a7f29c88a085a3658aa239b58f8a3af6a1ef53/src/datasets/packaged_modules/json/json.py#L402

PR #8264, opened 2026-06-12
docs: sync four docstrings with their actual signatures

DaoyuanLi2816:
## What does this PR do?

Audits docstring Args sections against actual signatures and fixes three drifts:

### `Dataset.train_test_split`

The Args section documents `train_cache_file_name` / `test_cache_file_name` — **neither is a parameter**; the real names are `train_indices_cache_file_name` / `test_indices_cache_file_name`. Following the docs (`ds.train_test_split(train_cache_file_name=...)`) raises `TypeError`. Renamed the two entries — their descriptions already describe the indices cache, so only the names change.

### `DatasetBuilder.as_dataset`

Drops the `verification_mode` entry: the parameter no longer exists (the signature is `(split, in_memory)`), so the entry documents an argument that can't be passed.

### `table.cast_table_to_schema`

The Args entry documents `features ([Features])` — carried over from the sibling `cast_table_to_features`; this function's parameter is `schema` (a `pa.Schema`). Renamed and retyped the entry.

Docstring-only; +4 / −6 across 3 files, no behavior change.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -5017,10 +5017,10 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
             load_from_cache_file (`Optional[bool]`, defaults to `True` if caching is enabled):
                 If a cache file storing the splits indices
                 can be identified, use it instead of recomputing.
-            train_cache_file_name (`str`, *optional*):
+            train_indices_cache_file_name (`str`, *optional*):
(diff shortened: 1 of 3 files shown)


DaoyuanLi2816:
Added one more: `DatasetBuilder.as_dataset` documented a `verification_mode` parameter it doesn't accept (signature is `(self, split=None, in_memory=False)`) — dropped the stale entry. Same drift class as the others.

PR #8268, opened 2026-06-13
Add force_sample_level to split_dataset_by_node

### LINKED ISSUE #8253: Feature request: Add a streaming_shard operator for early sample-level sharding when file-level sharding is insufficient
### Proposed API

It would be useful to provide a first-class streaming sharding operator in `datasets`, for example:

```python
dataset = datasets.load_dataset(..., streaming=True, split="train")

dataset = dataset.streaming_shard(
    num_shards=world_size,
    index=rank,
)
```

or:

```python
dataset = datasets.streaming_shard(
    dataset,
    num_shards=world_size,
    index=rank,
)
```

Then users could write:

```python
dataset = datasets.load_dataset(..., streaming=True, split="train")

dataset = dataset.streaming_shard(
    num_shards=world_size,
    index=rank,
)

dataset = dataset.map(tokenize_fn)
dataset = dataset.map(build_labels_or_loss_mask)
dataset = dataset.filter(...)
```

This would move rank-aware sample sharding earlier in the streaming pipeline, before expensive transformations.

### Why this is needed

Existing file-level sharding works well when the dataset has enough physical shards. However, it is not enough when:

```text
num_physical_files < global_world_size
```

or when physical files cannot be split.

For example, if we have:

```text
global_world_size = 64
num_physical_files = 4
```

then file-level assignment alone cannot give each rank a distinct physical shard. Many ranks would either read the same files or receive no file-level shard. In practice, the only reliable way to divide the dataset is sample-level sharding, such as what `IterableDatasetShard` does.

The problem is that `IterableDatasetShard` is applied at the training/dataloader layer. This means users cannot easily place the shard boundary before expensive dataset transformations.

### Expected behavior

For a streaming dataset yielding:

```python
[0, 1, 2, 3, 4, 5, 6, 7]
```

with:

```python
num_shards = 2
```

the result could be:

```python
rank 0 -> [0, 2, 4, 6]
rank 1 -> [1, 3, 5, 7]
```

This is sample-level logical sharding and does not depend on the number of physical files.

A batch-aware mode may also be useful to align with `IterableDatasetShard` semantics:

```python
dataset.streaming_shard(
    num_shards=world_size,
    index=rank,
    batch_size=batch_size,
    split_batches=False,
    drop_last=False,
)
```

### Current workaround

The current workaround is to rely on `IterableDatasetShard`:

```python
dataset = datasets.load_dataset(..., streaming=True, split="train")

dataset = dataset.map(tokenize_fn)
dataset = dataset.map(build_labels_or_loss_mask)

dataset = IterableDatasetShard(
    dataset,
    batch_size=batch_size,
    num_processes=world_size,
    process_index=rank,
)
```

This is correct from a data partitioning perspective, but inefficient because the sharding happens after the previous dataset transformations.

### Desired pattern

The desired pattern is:

```python
dataset = datasets.load_dataset(..., streaming=True, split="train")

dataset = dataset.streaming_shard(
    num_shards=world_size,
    index=rank,
)

dataset = dataset.map(tokenize_fn)
dataset = dataset.map(build_labels_or_loss_mask)
```

This way, each rank only preprocesses the examples it will actually consume.

### Use cases

This would be useful for:

1. Large-scale LLM pretraining with a small number of very large JSONL / Parquet files.
2. SFT pipelines where tokenization, chat template application, and loss-mask construction are expensive.
3. Multimodal training where image/video metadata parsing or remote object loading should not be repeated across ranks.
4. Distributed jobs where the number of source files is smaller than `global_world_size`.
5. Data sources where physical repartitioning is expensive or controlled by an upstream production pipeline.

### Relationship to existing APIs

This feature is related to:

* `accelerate.data_loader.IterableDatasetShard`, which provides sample-level sharding at the dataloader/training layer.
* Datasets file-level sharding, which works well when there are enough physical source shards.
* `to_iterable_dataset(num_shards=...)`, which can create iterable shards when converting a map-style dataset.

The proposed feature is different because it provides logical sample-level sharding directly in the `datasets` streaming pipeline, before downstream transformations such as `map`, `filter`, tokenization, or loss-mask construction.

### Open questions

1. Should the operator exactly match `IterableDatasetShard` semantics?
2. Should it support both sample-level and batch-aware sharding?
3. Should it support both strided and contiguous logical sharding?
4. How should it interact with `shuffle(buffer_size=...)` and `set_epoch()`?
5. Should this be implemented as a method on `IterableDataset`, or as a utility function under `datasets.distributed`?
6. Should it expose options such as `drop_last`, `even_batches`, and `split_batches` to match distributed dataloader behavior?

### Summary

When physical file-level sharding is impossible or when the number of source files is smaller than the global world size, distributed streaming training has to rely on `IterableDatasetShard` for sample-level data partitioning. However, this sharding happens too late, after expensive dataset transformations may already have been executed.

A `streaming_shard` operator in `datasets` would allow users to apply logical rank-aware sharding earlier in the streaming pipeline, reducing duplicated preprocessing work and improving efficiency for large-scale distributed training.

### Motivation

When using `datasets.load_dataset(..., streaming=True)` for distributed training, Hugging Face Datasets can benefit from physical file-level sharding if the dataset is already split into enough source files.

However, in many real-world training pipelines, this assumption does not hold.

For example:

* the dataset may contain only a few very large JSONL / Parquet files;
* the number of physical files may be smaller than the global distributed world size;
* the source files may be generated by an upstream pipeline and cannot be repartitioned easily;
* the files may be stored on remote storage such as HDFS or object storage, where rewriting or repartitioning is expensive;
* some file formats or data sources do not provide a convenient way to split one physical file into multiple independent streaming shards.

In these cases, file-level sharding is not sufficient. The training job has to fully rely on `accelerate.data_loader.IterableDatasetShard` or similar dataloader-level logic to split the data across ranks.

This works functionally, but it happens too late in the pipeline.

A common pipeline looks like this:

```python
dataset = datasets.load_dataset(..., streaming=True, split="train")

dataset = dataset.map(tokenize_fn)
dataset = dataset.map(build_labels_or_loss_mask)
dataset = dataset.filter(...)
```

Then the distributed dataloader wraps the dataset with `IterableDatasetShard`.

The issue is that every rank may execute the upstream streaming pipeline before the dataloader-level shard decides which examples should be kept by the current rank. For expensive preprocessing steps such as tokenization, chat template application, multimodal metadata loading, filtering, or loss-mask construction, this may cause duplicated CPU work and unnecessary I/O pressure.

### Your contribution

I propose adding a `streaming_shard` operator that can be applied before expensive streaming dataset transformations.

I’m happy to work on this feature if the maintainers think it makes sense.

muyihao:
Allow users to explicitly request sample-level sharding for streaming `IterableDataset` even when `num_shards % world_size == 0`. Default behavior is unchanged.

When `num_physical_files < world_size`, or when shards divide evenly but are too few or too imbalanced for shard-level sharding, applying sample-level sharding before expensive `map`/`filter` avoids duplicating those transformations across nodes:

    ds = split_dataset_by_node(ds, rank, world_size, force_sample_level=True)
    ds = ds.map(tokenize_fn)  # only the examples this rank consumes

Map-style datasets ignore the flag.

Closes #8253

### DIFF
--- a/docs/source/use_with_pytorch.mdx
+++ b/docs/source/use_with_pytorch.mdx
@@ -250,9 +250,16 @@ Each node is assigned a chunk of data, e.g. rank 0 is given the first chunk of t
 
 For iterable datasets:
 
-If the dataset has a number of shards that is a factor of `world_size` (i.e. if `dataset.num_shards % world_size == 0`),
+By default, if the dataset has a number of shards that is a factor of `world_size` (i.e. if `dataset.num_shards % world_size == 0`),
 then the shards are evenly assigned across the nodes, which is the most optimized.
 Otherwise, each node keeps 1 example out of `world_size`, skipping the other examples.
+
+When the number of physical shards is small or imbalanced relative to `world_size`, you can pass `force_sample_level=True` to force sample-level splitting even when shard counts divide evenly. Each node then keeps 1 example out of `world_size`, which avoids duplicating expensive transformations applied after `split_dataset_by_node` across nodes:
+
+```python
+ds = split_dataset_by_node(ds, rank=rank, world_size=world_size, force_sample_level=True)
+ds = ds.map(tokenize_fn)  # only runs on the examples this rank will actually consume
+```
 
 This can also be combined with a `torch.utils.data.DataLoader` if you want each node to use multiple workers to load the data.
 
--- a/src/datasets/distributed.py
+++ b/src/datasets/distributed.py
@@ -7,7 +7,12 @@ from .iterable_dataset import IterableDataset, _split_by_node_iterable_dataset
 DatasetType = TypeVar("DatasetType", Dataset, IterableDataset)
 
 
-def split_dataset_by_node(dataset: DatasetType, rank: int, world_size: int) -> DatasetType:
+def split_dataset_by_node(
+    dataset: DatasetType,
+    rank: int,
+    world_size: int,
+    force_sample_level: bool = False,
+) -> DatasetType:
     """
     Split a dataset for the node at rank `rank` in a pool of nodes of size `world_size`.
 
@@ -15,12 +20,19 @@ def split_dataset_by_node(dataset: DatasetType, rank: int, world_size: int) -> D
 
     Each node is assigned a chunk of data, e.g. rank 0 is given the first chunk of the dataset.
     To maximize data loading throughput, chunks are made of contiguous data on disk if possible.
+    The `force_sample_level` argument is ignored for map-style datasets.
 
     For iterable datasets:
 
-    If the dataset has a number of shards that is a factor of `world_size` (i.e. if `dataset.num_shards % world_size == 0`),
-    then the shards are evenly assigned across the nodes, which is the most optimized.
-    Otherwise, each node keeps 1 example out of `world_size`, skipping the other examples.
+    By default, if the dataset has a number of shards that is a factor of `world_size` (i.e. if
+    `dataset.num_shards % world_size == 0`), then the shards are evenly assigned across the nodes,
+    which is the most optimized. Otherwise, each node keeps 1 example out of `world_size`, skipping
+    the other examples.
+
+    Pass `force_sample_level=True` to force sample-level splitting regardless of shard count
+    (each node keeps 1 example out of `world_size`). This is useful when the number of physical
+    shards is small or imbalanced relative to `world_size`, so that expensive transformations applied
+    after `split_dataset_by_node` are not duplicated across nodes.
 
     > [!WARNING]
     > If you shuffle your iterable dataset in a distributed setup, make sure to set a fixed `seed` in [`IterableDataset.shuffle`]
@@ -33,6 +45,9 @@ def split_dataset_by_node(dataset: DatasetType, rank: int, world_size: int) -> D
             Rank of the current node.
         world_size (`int`):
             Total number of nodes.
+        force_sample_level (`bool`, defaults to `False`):
+            For iterable datasets, force sample-level splitting even when shards divide evenly across
+            nodes. Ignored for map-style datasets.
 
     Returns:
         [`Dataset`] or [`IterableDataset`]: The dataset to be used on the node at rank `rank`.
@@ -40,4 +55,6 @@ def split_dataset_by_node(dataset: DatasetType, rank: int, world_size: int) -> D
     if isinstance(dataset, Dataset):
         return _split_by_node_map_style_dataset(dataset, rank=rank, world_size=world_size)
     else:
-        return _split_by_node_iterable_dataset(dataset, rank=rank, world_size=world_size)
+        return _split_by_node_iterable_dataset(
+            dataset, rank=rank, world_size=world_size, force_sample_level=force_sample_level
+        )
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -2426,6 +2426,7 @@ class FormattedExamplesIterable(_BaseExamplesIterable):
 class DistributedConfig:
     rank: int
     world_size: int
+    force_sample_level: bool = False
 
 
 def _maybe_add_torch_iterable_dataset_parent_class(cls):
@@ -2666,8 +2667,11 @@ class IterableDataset(DatasetInfoMixin):
 
     @property
     def num_shards(self) -> int:
-        if self._distributed and self._ex_iterable.num_shards % self._distributed.world_size == 0:
-            return self._ex_iterable.num_shards // self._distributed.world_size
+        if self._distributed:
+            world_size = self._distributed.world_size
+            divisible = self._ex_iterable.num_shards % world_size == 0
+            if divisible and not self._distributed.force_sample_level:
+                return self._ex_iterable.num_shards // world_size
         return self._ex_iterable.num_shards
 
     @property
@@ -2753,7 +2757,8 @@ class IterableDataset(DatasetInfoMixin):
         if self._distributed:
             rank = self._distributed.rank
             world_size = self._distributed.world_size
-            if ex_iterable.num_shards % world_size == 0:
+            divisible = ex_iterable.num_shards % world_size == 0
+            if divisible and not self._distributed.force_sample_level:
                 if self._is_main_process():
                     num_shards_per_node = ex_iterable.num_shards // world_size
                     plural = "s" if num_shards_per_node > 1 else ""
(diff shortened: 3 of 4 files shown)


muyihao:
@lhoestq , hi, could you please help review this PR? Thanks!

trungminhdo4-glitch:
I tested the `force_sample_level=True` path on the exact current PR head
(`564ba50d`) because I wanted to verify the example in the PR description.

With one deterministic 240-record local JSONL file and `world_size=4`, each
rank retained the expected 60 strided records, but each rank still:

- read the full 30,183-byte source,
- parsed all 240 records, and
- invoked the `map` callback 240 times.

I got the same result both on the exact PR head and with this patch applied
cleanly to current `main`. `map`-before-split and split-before-`map` also
produced the same callback count; the same held for `filter`.

From tracing the iterable construction, it looks like
`_prepare_ex_iterable_for_iteration()` wraps the existing `_ex_iterable`
with `StepExamplesIterable`, while `map()` / `filter()` have already placed
their wrappers inside that iterable. So in this path the effective order is
still roughly:

`source read/parse -> map/filter -> StepExamplesIterable selection`

rather than selection before the expensive callback.

As a separate experiment, I tried record-aligned physical byte ranges for
the same local uncompressed JSONL case. Across four workers, the worker
ranges together read and parsed the file once and reproduced the exact
ordered records without gaps or duplicates. The simple planner requires one
additional boundary-planning scan, so I see this as a complementary
physical-I/O optimization rather than a replacement for the logical
sample-level API in this PR.

I can share the minimal reproduction if useful.

muyihao:
@trungminhdo4-glitch Thanks for testing this and for the detailed investigation.

Just to make sure I understand the behavior correctly, could you share the exact order in which you constructed the dataset and called split_dataset_by_node() and map() in your reproduction?

For example, did you test:

```python
ds = split_dataset_by_node(
    ds, rank, world_size, force_sample_level=True
)
ds = ds.map(tokenize_fn)
```

as well as:

```python
ds = ds.map(tokenize_fn)
ds = split_dataset_by_node(
    ds, rank, world_size, force_sample_level=True
)
```
I’m particularly interested in the first case, since the intended use case of this PR is to apply sample-level sharding before the expensive map/filter transformations.

If the first case still invokes the map callback for all samples, then it indeed looks like StepExamplesIterable is being applied after the existing map wrapper internally, as you described.

trungminhdo4-glitch:
Yes — I tested split-before-map explicitly on the exact PR head `564ba50d`:

```python
ds = IterableDataset.from_generator(source_generator)
ds = split_dataset_by_node(
    ds, rank, world_size, force_sample_level=True
)
ds = ds.map(counting_map)
list(ds)
```

For one deterministic local JSONL source (240 records, 30,183 bytes) and `world_size=4`, the runtime result was:

| rank | retained | source records parsed | map calls | source bytes read |
|---:|---:|---:|---:|---:|
| 0 | 60 | 240 | 240 | 30,183 |
| 1 | 60 | 240 | 240 | 30,183 |
| 2 | 60 | 240 | 240 | 30,183 |
| 3 | 60 | 240 | 240 | 30,183 |

The retained IDs were the expected strided sets (`rank, rank+4, ..., rank+236`). So split-before-map still invokes the expensive callback for all 240 source records on every rank.

I also tested map-before-split; it behaved the same: 60 retained, 240 source records parsed, 240 map calls, and 30,183 source bytes read per rank. The corresponding split-before-filter path likewise evaluated the filter for all 240 source records per rank.

The current source explains the ordering: `map()` puts `MappedExamplesIterable` around `self._ex_iterable` ([line 3605](https://github.com/huggingface/datasets/blob/564ba50d328a61b9bff9ebb8cfe4912cc8fcd351/src/datasets/iterable_dataset.py#L3567-L3605)); `split_dataset_by_node()` preserves that iterable and sets `DistributedConfig` ([distributed.py lines 52–60](https://github.com/huggingface/datasets/blob/564ba50d328a61b9bff9ebb8cfe4912cc8fcd351/src/datasets/distributed.py#L52-L60)); iteration then wraps the existing iterable in `StepExamplesIterable` ([lines 2748–2780](https://github.com/huggingface/datasets/blob/564ba50d328a61b9bff9ebb8cfe4912cc8fcd351/src/datasets/iterable_dataset.py#L2748-L2780)). The effective order is therefore `source/read -> map/filter -> StepExamplesIterable -> retained samples`, for both public construction orders.

Minimal reproducer: the two snippets above, using a local 240-line JSONL generator that increments counters on each source `readline()`, JSON parse, and callback invocation.

muyihao:
@trungminhdo4-glitch Thanks again for the detailed reproduction and root-cause analysis — you were right about the effective iterable ordering.
I’ve updated the implementation on the latest PR head. When force_sample_level=True, StepExamplesIterable is now inserted into the iterable pipeline at the point where split_dataset_by_node() is called, rather than being deferred until iteration through DistributedConfig.
As a result:
 - split -> map/filter: the callback now runs only for the samples retained by each rank (e.g. 60 calls per rank for 240 samples with world_size=4);
- map/filter -> split: the callback still runs for all samples, as expected, since the transformation is upstream of the split;
- sequential source reads/parsing still process all source records on every rank. This change avoids duplicated downstream transformations, but does not implement physical source-I/O partitioning.

I’ve also added you as a co-author for your contribution. If you have time, it would be great if you could rerun your reproducer against the latest head and confirm the new callback counts. Thanks!

trungminhdo4-glitch:
Confirmed independently on the latest head
`133bb52416dc63cdeac4804e854832f48ae25e69`.

Using the same deterministic 240-record local JSONL fixture with
`world_size=4`, I now get per rank:

- `split -> map`: 60 retained, **60 map calls**, 240 source records
  read/parsed
- `map -> split`: 60 retained, **240 map calls**, 240 source records
  read/parsed
- `split -> filter`: 60 retained, **60 filter calls**, 240 source records
  read/parsed

Across all four ranks the retained IDs still have exact coverage: 240 total,
no duplicates, no missing records.

So the latest change fixes the duplicated downstream `map`/`filter` work
exactly as intended. Sequential source reading/parsing remains full-source
per rank, consistent with the scope you described.

Thanks for the update and for including me as a co-author.

muyihao:
@lhoestq Hi! We’ve updated the PR based on the recent discussion and pushed the latest changes. Could you please take a look at the current implementation when you have a chance?

We’d really appreciate a review. Thanks!

muyihao:
@trungminhdo4-glitch A bit unfortunate for #8268 after waiting for a while, but perhaps #8565 has found a better way to address the same issue. Glad to see it moving forward!

lhoestq:
Yea I did a batch of reviews a few days ago and I must have missed your PR, sorry

At least the issue is fixed now, thanks again for the discussions in the issue and for diving into it

PR #8277, opened 2026-06-17
fix traces streaming

lhoestq:
requires https://github.com/TeichAI/teich/pull/12 as well


armand0e:
Thanks for the PR. Implemented them in the latest version: 0.2.5

PR #8284, opened 2026-06-26
Hash frozensets deterministically

sarathfrancis90:
While looking at how transforms get fingerprinted I noticed `frozenset` isn't hashed deterministically, unlike `set`.

`set` has a custom reducer (`_save_set`) that pickles its elements sorted, so the fingerprint is stable across runs. `frozenset` has no such reducer and falls back to the default one, which keeps the iteration order. Because Python randomizes the string hash seed per process, a transform that captures a `frozenset` of strings gets a different fingerprint every session, so the cache never hits.

Repro:
```python
# prints a different hash on every run (PYTHONHASHSEED is random by default)
print(Hasher.hash(frozenset("abcdefghij")))
```

I registered the same sorted reducer for `frozenset` that `set` already uses (falling back to sorting by `Hasher.hash` for unorderable elements). Added a test that hashes a frozenset in two subprocesses with different `PYTHONHASHSEED` and checks the hashes match, mirroring the existing `set` coverage.

### DIFF
--- a/src/datasets/utils/_dill.py
+++ b/src/datasets/utils/_dill.py
@@ -146,6 +146,21 @@ def _save_set(pickler, obj):
     log(pickler, "# Se")
 
 
+@pklregister(frozenset)
+def _save_frozenset(pickler, obj):
+    log(pickler, f"Fr: {obj}")
+    try:
+        # Faster, but fails for unorderable elements
+        args = (sorted(obj),)
+    except Exception:  # TypeError, decimal.InvalidOperation, etc.
+        from datasets.fingerprint import Hasher
+
+        args = (sorted(obj, key=Hasher.hash),)
+
+    pickler.save_reduce(frozenset, args, obj=obj)
+    log(pickler, "# Fr")
+
+
 def _save_regexPattern(pickler, obj):
     import regex  # type: ignore
 
--- a/tests/test_fingerprint.py
+++ b/tests/test_fingerprint.py
@@ -2,6 +2,7 @@ import json
 import os
 import pickle
 import subprocess
+import sys
 from functools import partial
 from pathlib import Path
 from tempfile import gettempdir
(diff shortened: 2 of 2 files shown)


Rodrigo-Palma:
I hit the same problem independently and ended up with the same reducer, so here is the
measurement I have, in case it helps this along.

**The test is a real regression test.** Your `PYTHONHASHSEED` pair fails on `main` and
passes with the patch, on a set of 10 strings:

```
main       PYTHONHASHSEED=0 -> 08706786801f3a5e   PYTHONHASHSEED=1 -> 9408e6f1b30d3f9b
with patch PYTHONHASHSEED=0 -> 878165f46e7e4822   PYTHONHASHSEED=1 -> 878165f46e7e4822
```

Worth noting for reviewers that a `Pool(1)` test in the style of `test_set_stable` would
not have caught this on Linux: the forked child inherits the parent's hash seed. The
subprocess with an explicit seed is the right shape.

**It does not move any existing fingerprint.** `set` and `dict` hash identically before
and after (`Hasher.hash(set_of_8_strings)` is `e5bef1d1eae9255f` in both), so no cached
transform is invalidated by this change. That seemed the main risk worth ruling out.

**One residual, if you want it in scope.** `sorted(obj)` succeeds when the elements are
themselves sets, but subset comparison is only a partial order, so the result still
depends on the input iteration order:

```
frozenset({frozenset("ab"), frozenset("cd")})   seed 1 -> d4bf0863569dc792   seed 2 -> 94bd85deae050311
set({frozenset("ab"), frozenset("cd")})         seed 1 -> 3fe3d860281ba7dd   seed 2 -> 3f873e13efc470f5   (main, today)
```

The second line is `main` as it stands, so this is pre-existing for `set` and not
something your patch introduces, and no cache can be hitting for those values today
either way. Making it deterministic means sorting by `Hasher.hash` unconditionally, which
would change existing `set` fingerprints, so it looks like a separate decision rather
than something to fold in here.

Also, `_save_set` and `_save_frozenset` are now the same eight lines twice; a shared
helper for the sorted elements would keep the two from drifting apart. Entirely optional.

PR #8288, opened 2026-06-29
Fix duplicate word in HubBucketDatasetModuleFactory docstring

Sreekant13:
Removes a duplicated "a" in the `HubBucketDatasetModuleFactory` docstring in `src/datasets/load.py`:

"...data files of a a Storage Bucket." → "...data files of a Storage Bucket."

Documentation-only change; no code behavior affected.

### DIFF
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -843,7 +843,7 @@ class CachedDatasetModuleFactory(_DatasetModuleFactory):
 
 class HubBucketDatasetModuleFactory(_DatasetModuleFactory):
     """
-    Get the module of a dataset loaded from data files of a a Storage Bucket.
+    Get the module of a dataset loaded from data files of a Storage Bucket.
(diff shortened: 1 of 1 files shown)


Sreekant13:
Hi @lhoestq, gentle nudge on this docstring typo fix whenever you have a moment. Thanks for your work on datasets.

Sreekant13:
Hi @lhoestq, gentle nudge on this docstring typo fix whenever you have a moment. I also have #8315 open, a small size_str(0) fix with a regression test, if you are batching reviews. Thanks for your work on datasets.

Sreekant13:
Hi @lhoestq, a final check-in on this one. It is only a one-line docstring typo fix, so if it is not worth a review cycle I am glad to close it. If you would like it in, it is ready and rebased. Either way is fine. (The more substantive sibling is #8315, which fixes `size_str(0)` returning "Unknown size" instead of "0 bytes".)

PR #8291, opened 2026-06-30
Add `skip_origin_metadata` flag to optimize streaming startup time

### LINKED ISSUE #8197: Feature Request / Discussion: Skip origin_metadata fetch when streaming=True?

yuxin00j:
Fix #8197
This PR introduces the `skip_origin_metadata` kwarg to skip costly origin metadata fetching (ETags/mtime) when it's not needed, improving streaming initialization latency. This saves number of files `fs.info` calls per `load_dataset` call.

### Changes
- **Kwarg Plumbing**: Routes the `skip_origin_metadata` kwarg from `load_dataset(...)` safely down to `DatasetBuilder` while stripping it securely from kwargs (via `.pop()`) so it's not leaked into underlying configuration classes that don't expect it (such as `CsvConfig()`).
- **Streaming Validations**:
  - Emits a warning when `skip_origin_metadata=True` is passed to `load_dataset` without `streaming=True`.
  - Disables `download_and_prepare()` entirely when `builder._skip_origin_metadata` is true by raising a `ValueError("This function is not intended for streaming.")`.
- **Test Coverage**: Added robust test coverage spanning `tests/test_data_files.py` and `tests/test_load.py`:
  - `test_DataFilesList_from_patterns_skip_origin_metadata`
  - `test_DataFilesList_from_patterns_skip_origin_metadata_raises_FileNotFoundError`
  - `test_load_dataset_builder_skip_origin_metadata` (covers kwarg mapping & ValueError)
  - `test_load_dataset_skip_origin_metadata_warning`

### Metrics
We ran a streaming dataset (1650 shards) benchmark across 8 processes against GCS. 
1. Before PR (main branch)
Performance: Iteration execution times were highly variable, scaling anywhere between ~4.0s and ~10.8s depending.
fs.info() Calls: 1,651 per process (13,208 total across 8 processes).

2. After PR (pr-8291, with --skip-origin-metadata)
Performance: Execution times were extremely stable, consistently landing around ~1.7s.
fs.info() Calls: 1 per process (8 total across 8 processes).

### Impact 
By bypassing the unconditional ETag/origin metadata fetches during dataset builder initialization on the streaming path, we reduced the fs.info() calls from over 13,000 down to just 8. 

On average, this structural optimization reduced the initialization latency from 5.55s to 1.77s (~3.13x faster), entirely removing the extensive delays previously chained to API overhead/variability.

### DIFF
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -276,7 +276,7 @@ class DatasetBuilder:
             None means that the ArrowWriter will use its default value.
         skip_origin_metadata (`bool`, *optional*):
             Whether to skip fetching origin metadata (ETag/mtime) for the data files.
-            Defaults to `False`. When `True`, initialization may be faster, but it skips the file existence check.
+            It should only be true when streaming is true.
         **config_kwargs (additional keyword arguments): Keyword arguments to be passed to the corresponding builder
             configuration class, set on the class attribute [`DatasetBuilder.BUILDER_CONFIG_CLASS`]. The builder
             configuration class is [`BuilderConfig`] or a subclass of it.
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -274,6 +274,9 @@ class DatasetBuilder:
             It defines the number of samples that are kept in memory before writing them
             and also the length of the arrow chunks.
             None means that the ArrowWriter will use its default value.
+        skip_origin_metadata (`bool`, *optional*):
+            Whether to skip fetching origin metadata (ETag/mtime) for the data files.
+            Defaults to `False`. When `True`, initialization may be faster, but it skips the file existence check.
         **config_kwargs (additional keyword arguments): Keyword arguments to be passed to the corresponding builder
             configuration class, set on the class attribute [`DatasetBuilder.BUILDER_CONFIG_CLASS`]. The builder
             configuration class is [`BuilderConfig`] or a subclass of it.
@@ -313,6 +316,7 @@ class DatasetBuilder:
         storage_options: Optional[dict] = None,
         writer_batch_size: Optional[int] = None,
         config_id: Optional[str] = None,
+        skip_origin_metadata: bool = False,
         **config_kwargs,
     ):
         # DatasetBuilder name
@@ -330,6 +334,7 @@ class DatasetBuilder:
                 sanitize_patterns(data_files),
                 base_path=base_path,
                 download_config=DownloadConfig(token=token, storage_options=self.storage_options),
+                skip_origin_metadata=skip_origin_metadata,
             )
 
         # Prepare config: DatasetConfig contains name, version and description but can be extended by each dataset
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -779,7 +779,7 @@ class DatasetBuilder:
         >>> builder.download_and_prepare("s3://my-bucket/my_rotten_tomatoes", storage_options=storage_options, file_format="parquet")
         ```
         """
-        if getattr(self, "_skip_origin_metadata", False):
+        if self._skip_origin_metadata:
             raise ValueError("This function is not intended for streaming.")
 
         output_dir = output_dir if output_dir is not None else self._cache_dir
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -200,10 +200,17 @@ class BuilderConfig:
         else:
             return self.name
 
-    def _resolve_data_files(self, base_path: str, download_config: DownloadConfig) -> None:
+    def _resolve_data_files(
+        self,
+        base_path: str,
+        download_config: DownloadConfig,
+        skip_origin_metadata: bool = False,
+    ) -> None:
         if isinstance(self.data_files, DataFilesPatternsDict):
             base_path = xjoin(base_path, self.data_dir) if self.data_dir else base_path
-            self.data_files = self.data_files.resolve(base_path, download_config)
+            self.data_files = self.data_files.resolve(
+                base_path, download_config, skip_origin_metadata=skip_origin_metadata
+            )
 
 
 class DatasetBuilder:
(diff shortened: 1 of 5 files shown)


src/datasets/load.py
zhixiangli, line 408: `skip_origin_metadata: bool = False` is needed for those support remote streaming.

zhixiangli, line 418: Add `self.skip_origin_metadata = skip_origin_metadata`

zhixiangli, line 446: Add `skip_origin_metadata=self.skip_origin_metadata`

zhixiangli, line 511: Same as `LocalDatasetModuleFactory`

zhixiangli, line 555: Same as `LocalDatasetModuleFactory`

zhixiangli, line 843: Same as `LocalDatasetModuleFactory`

zhixiangli, line 974: The `skip_origin_metadata` parameter is required and passes those values to the factories that support remote streaming.

zhixiangli, line 1342: Include `dataset_module_factory`

src/datasets/builder.py
zhixiangli, line 713: If `self._skip_origin_metadata` is true, we should raise an error, as this function is not intended for streaming.

zhixiangli, line 576: Should we pass `skip_origin_metadata` to `_resolve_data_files`?

yuxin00j: Done.

yuxin00j: Done. There's a check added in `load_dataset` method below.

src/datasets/load.py
yuxin00j: Done.

yuxin00j: Done.

yuxin00j: Done.

yuxin00j: Done.

yuxin00j: Done.

yuxin00j: Done.

yuxin00j: Done.

yuxin00j: Done.

src/datasets/load.py
```diff
@@ -1330,6 +1341,7 @@ def load_dataset_builder(
         data_dir=data_dir,
         data_files=data_files,
         cache_dir=cache_dir,
+        skip_origin_metadata=skip_origin_metadata,
```
zhixiangli: nit: why not `skip_origin_metadata=config_kwargs.get("skip_origin_metadata", False)`

src/datasets/builder.py
```diff
@@ -773,6 +779,9 @@ def download_and_prepare(
         >>> builder.download_and_prepare("s3://my-bucket/my_rotten_tomatoes", storage_options=storage_options, file_format="parquet")
         ```
         """
+        if getattr(self, "_skip_origin_metadata", False):
```
zhixiangli: nit: `self._skip_origin_metadata`

tests/test_load.py
```diff
@@ -1125,6 +1125,11 @@ def test_load_streaming_private_dataset_with_zipped_data(hf_token, hf_private_da
     assert next(iter(ds)) is not None
```
zhixiangli, line 1127: Please add tests for the ALL changed.

yuxin00j (commented):
I have pushed a fix for the remaining open comments:

1. Regarding `.get()` vs `.pop()`: Using `.get()` leaves `skip_origin_metadata` inside `config_kwargs`, which then gets passed to the `CsvConfig.__init__` (and others), causing a `TypeError`. I reverted back to `.pop()` and explicitly passed `skip_origin_metadata` to the `builder_cls` call.
2. Changed to use `self._skip_origin_metadata` directly in `builder.py`.
3. The newly added tests (`test_load_dataset_builder_skip_origin_metadata` and `test_load_dataset_skip_origin_metadata_warning`) cover the changes across `load.py` and `builder.py`.

src/datasets/builder.py
yuxin00j: Fixed! Updated exactly to use `self._skip_origin_metadata` directly.

src/datasets/load.py
yuxin00j: Fixed in the latest commit! I've inlined the `.get()` call here and then properly extracted it with `.pop()` when instantiating the builder later to prevent kwargs leakage.

tests/test_load.py
yuxin00j: Done.

yuxin00j:
Hi @lhoestq, could you take a look at this PR? Thanks.

ankitaluthra1:
@yuxin00j Can you please add how much is this change improving streaming initialization latency ?

yuxin00j:
> @yuxin00j Can you please add how much is this change improving streaming initialization latency ?

Added in the description.

src/datasets/builder.py
```diff
@@ -279,6 +281,9 @@ class DatasetBuilder:
             It defines the number of samples that are kept in memory before writing them
             and also the length of the arrow chunks.
             None means that the ArrowWriter will use its default value.
+        skip_origin_metadata (`bool`, *optional*):
+            Whether to skip fetching origin metadata (ETag/mtime) for the data files.
+            It should only be true when streaming is true.
```
ankitaluthra1: @yuxin00j Consider adding more details on when it should be `true` like for any cloud storage and any type of bucket. Is there any other scenario in which this should be set `false`

yuxin00j: Done.

trinadhkotturu:
@lhoestq for your review

PR #8301, opened 2026-07-03
Fix lance auth

lhoestq:
fix two bugs
- `TypeError: argument 'storage_options': 'None' is not an instance of 'str'` because "token": None is not supported in storage_options
- download_config.storage_options was empty and therefore the token for private datasets was not passed to lance


davanstrien (approved):
thanks for getting to this so quick!

Issue #8296, opened 2026-06-30
[security] Symlink-Following Arbitrary File Write via Archive Extraction in huggingface/datasets


AAtomical:
## Summary

A symlink-following vulnerability exists in the archive extraction subsystem of `huggingface/datasets` that allows a local attacker with write access to a shared HuggingFace cache directory to write arbitrary files with the victim's privileges, by pre-planting a symlink at a predictable extraction output path.

The root cause is that `Extractor.extract()` in `src/datasets/utils/extract.py` calls `shutil.rmtree(output_path, ignore_errors=True)` which **silently fails** when `output_path` is a symlink (Python ≥3.8 documented behavior), then proceeds to extract archive contents through the surviving symlink via `os.makedirs(output_path, exist_ok=True)` + `extractall()`.

The existing `safemembers()` defense (CVE-2007-4559 zip-slip mitigation) only validates archive **member names** — it is completely blind to the output directory itself being a symlink. The extraction output path is fully predictable (`sha256` with no salt).

In shared-cache deployments (CI runners, multi-user ML servers, NFS mounts, Docker volumes), this yields **arbitrary file write as the victim** → privilege escalation / code execution.
### Vulnerable function: `Extractor.extract()`

```python
# src/datasets/utils/extract.py, line 432-444
@classmethod
def extract(cls, input_path, output_path, extractor_format):
    os.makedirs(os.path.dirname(output_path), exist_ok=True)
    lock_path = str(Path(output_path).with_suffix(".lock"))
    with FileLock(lock_path):
        shutil.rmtree(output_path, ignore_errors=True)   # ← silent no-op on symlink
        extractor = cls.extractors[extractor_format]
        return extractor.extract(input_path, output_path) # ← follows the symlink
```

### Sink: format-specific extractors

```python
# TarExtractor.extract(), line 128-132 (identical pattern in Zip/Rar/7z)
@staticmethod
def extract(input_path, output_path):
    os.makedirs(output_path, exist_ok=True)                                    # ← follows symlink
    tar_file = tarfile.open(input_path)
    tar_file.extractall(output_path, members=TarExtractor.safemembers(...))     # ← writes to target
```

### Why `shutil.rmtree` fails silently

Since Python 3.8 ([bpo-35699](https://bugs.python.org/issue35699)), `shutil.rmtree()` raises `OSError` when the path is a symbolic link to a directory. With `ignore_errors=True`, this exception is swallowed. The symlink, its target, and all target contents survive completely intact.

[...]

PR #8303, opened 2026-07-03
Fix symlink-following arbitrary file write in archive extraction

AAtomical:
Extractor.extract() called shutil.rmtree(output_path, ignore_errors=True) before extracting. Since Python 3.8, shutil.rmtree() raises OSError on a symlink pointing to a directory, which ignore_errors=True silently swallows. An attacker with write access to a shared cache directory could pre-plant a symlink at the predictable extraction output path; rmtree would no-op on it, and extraction would then follow the symlink and write archive contents into the attacker-chosen target, yielding arbitrary file write as the victim.

The existing safemembers() zip-slip mitigation only validates member names and is blind to the output directory itself being a symlink.

Remove the symlink itself before extracting so contents can no longer be written through it.
#8296

### DIFF
--- a/src/datasets/utils/extract.py
+++ b/src/datasets/utils/extract.py
@@ -439,6 +439,8 @@ class Extractor:
         # Prevent parallel extractions
         lock_path = str(Path(output_path).with_suffix(".lock"))
         with FileLock(lock_path):
+            if os.path.islink(output_path):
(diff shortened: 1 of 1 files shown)


lhoestq (approved):
lgtm thx !

PR #8315, opened 2026-07-09
Fix size_str(0) returning "Unknown size" instead of "0 bytes"

Sreekant13:
`size_str` uses `if not size_in_bytes:` to return `"Unknown size"`, which also catches `0`. But `0` is a valid size (0 bytes), and the docstring states that only `None` means "Unknown size". So a 0-byte size (e.g. an empty split) currently renders as "Unknown size":

```python
>>> from datasets.utils.py_utils import size_str
>>> size_str(0)
'Unknown size'   # expected '0 bytes'
```
This checks `size_in_bytes is None` instead. Added a regression test covering `None`, `0`, and non-zero sizes.

### DIFF
--- a/src/datasets/utils/py_utils.py
+++ b/src/datasets/utils/py_utils.py
@@ -79,7 +79,7 @@ def size_str(size_in_bytes):
         size_in_bytes: `int` or `None`, the size, in bytes, that we want to
             format as a human-readable size string.
     """
-    if not size_in_bytes:
+    if size_in_bytes is None:
         return "Unknown size"
 
     _NAME_LIST = [("PiB", 2**50), ("TiB", 2**40), ("GiB", 2**30), ("MiB", 2**20), ("KiB", 2**10)]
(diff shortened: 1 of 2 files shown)


Sreekant13:
Hi @lhoestq, gentle nudge on this one when you get a chance. It fixes size_str(0) returning "Unknown size" instead of "0 bytes", with a regression test. Thanks!

Sreekant13:
Hi @lhoestq, a gentle check-in on this one whenever you have capacity.

`size_str(0)` returns "Unknown size" because the whole-number branch is gated on a truthy check, so a genuine zero falls through to the same message as `None`. The docstring says only `None` is unknown. The fix is `is None` rather than truthiness, so a 0-byte file renders as "0 bytes", with a regression test.

Mentioning it here rather than opening a second thread: #8288 is also still open, a one-word duplicate typo in the `HubBucketDatasetModuleFactory` docstring. That one is trivial and can be closed just as easily if you would rather not carry it.

No rush on either, and thank you for the work on datasets.

PR #8317, opened 2026-07-10
fix(json): produce valid JSON when lines=False spans multiple batches (#7037)

### LINKED ISSUE #7037: A bug of Dataset.to_json() function

sridhar-3009:
## Summary

Fixes #7037.

`Dataset.to_json(lines=False)` raised `NotImplementedError` (and before the guard was added, silently produced _invalid_ JSON) whenever `batch_size < num_rows` — i.e., the default case for any dataset larger than `DEFAULT_MAX_BATCH_SIZE` (1000 rows).

**Root cause:** Each batch was serialised to its own JSON array (`[{...}, ...]`) and written back-to-back in the file, producing e.g. `[{...}][{...}]` — syntactically invalid.

**Fix (two changes in `src/datasets/io/json.py`):**

1. **`write()`** — Narrow the `NotImplementedError` guard to dict-producing orientations (`split`, `index`, `columns`, `table`) only. Array-producing orientations (`records`, `values`) can now be batched safely.

2. **`_write()`** — When `lines=False` and `orient in ("records", "values")`, strip the outer `[` / `]` from each batch's bytes and emit a single wrapping `[...]`, streaming with no extra memory overhead. Dict-producing orientations fall through to the unchanged write-as-is path.

As a bonus, the single-proc and multiprocessing iteration loops are unified via `contextlib.nullcontext`, removing ~10 lines of duplication.

**Verified:**

```python
import io, json
from datasets import Dataset
from datasets.io.json import JsonDatasetWriter

ds = Dataset.from_dict({"id": list(range(25)), "value": [f"v{i}" for i in range(25)]})
buf = io.BytesIO()
JsonDatasetWriter(ds, buf, batch_size=10, lines=False).write()  # 3 batches
buf.seek(0)
result = json.loads(buf.read())   # previously raised NotImplementedError / produced invalid JSON
assert len(result) == 25          # single valid JSON array with all rows
```

## Test plan

- [x] New parametrised test `test_dataset_to_json_lines_false_batched` covers `orient="records"` and `orient="values"` with `batch_size < num_rows`
- [x] Full `TestJsonDatasetWriter` suite passes (20 tests; 3 pre-existing errors for `shared_datadir` fixture from missing `pytest-datadir` plugin, unrelated to this change)

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -4036,7 +4036,14 @@ class IterableDataset(DatasetInfoMixin):
         Returns:
             `IterableDataset`
         """
-        return self.map(partial(add_column_fn, name=name, column=column), with_indices=True)
+        original_features = self._info.features.copy() if self._info.features else None
+        ds_iterable = self.map(partial(add_column_fn, name=name, column=column), with_indices=True)
+        if original_features is not None:
+            col_table = pa.table({name: column})
+            new_features = original_features.copy()
+            new_features[name] = Features.from_arrow_schema(col_table.schema)[name]
+            ds_iterable._info.features = new_features
+        return ds_iterable
 
     def rename_column(self, original_column_name: str, new_column_name: str) -> "IterableDataset":
         """
--- a/src/datasets/io/json.py
+++ b/src/datasets/io/json.py
@@ -104,9 +104,12 @@ class JsonDatasetWriter:
         if compression not in [None, "infer", "gzip", "bz2", "xz"]:
             raise NotImplementedError(f"`datasets` currently does not support {compression} compression")
 
-        if not lines and self.batch_size < self.dataset.num_rows:
+        # Array-producing orientations ("records", "values") support batching when lines=False;
+        # dict-producing orientations ("split", "index", "columns", "table") do not.
+        if not lines and orient not in ("records", "values") and self.batch_size < self.dataset.num_rows:
             raise NotImplementedError(
-                "Output JSON will not be formatted correctly when lines = False and batch_size < number of rows in the dataset. Use pandas.DataFrame.to_json() instead."
+                f"Output JSON will not be formatted correctly when lines=False, orient={orient!r}, "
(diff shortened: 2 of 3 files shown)


src/datasets/iterable_dataset.py
```diff
@@ -4036,7 +4036,14 @@ def add_column(self, name: str, column: Union[list, np.array]) -> "IterableDatas
         Returns:
             `IterableDataset`
         """
-        return self.map(partial(add_column_fn, name=name, column=column), with_indices=True)
+        original_features = self._info.features.copy() if self._info.features else None
+        ds_iterable = self.map(partial(add_column_fn, name=name, column=column), with_indices=True)
+        if original_features is not None:
+            col_table = pa.table({name: column})
```
Sanjays2402, line 4042: `add_column` currently accepts a 2-D NumPy column because `add_column_fn` feeds one row at a time into Arrow. This new `pa.table({name: column})` fails eagerly with `ArrowInvalid: only handle 1-dimensional arrays`; wrapping it as `list(column)` preserves that input while still inferring a list feature.

PR #8318, opened 2026-07-10
Vectorize interleave_datasets index generation (probabilities + first/all_exhausted)

saforem2:
## What

`_interleave_map_style_datasets` (used by `interleave_datasets` for map-style `Dataset`s) builds the output index list in a **pure-Python `for`-loop, one iteration per output row**, when `probabilities` is given:

```python
for source_idx in iter_random_indices():
    if bool_strategy_func(is_exhausted):
        break
    indices.append(current_index[source_idx] + offsets[source_idx])
    current_index[source_idx] += 1
    if current_index[source_idx] >= lengths[source_idx]:
        is_exhausted[source_idx] = True
        current_index[source_idx] = 0
```

For large interleaves this loop dominates runtime. Interleaving e.g. [`nvidia/OpenMathInstruct-2`](https://huggingface.co/datasets/nvidia/OpenMathInstruct-2) (~14M rows) with two smaller datasets under `stopping_strategy="all_exhausted"` produces ~93M output rows and the index build alone takes **~90 min** — almost entirely Python interpreter overhead, since the RNG is already drawn in batches (`rng.choice(..., size=1000)`); it is not compute-bound.

Notably, the sibling `probabilities is None` `all_exhausted` branch in the same function is **already vectorized** with numpy (`np.mod`/offset). This PR brings the probabilities-given `first_exhausted` and `all_exhausted` branches to parity.

## How

Reproduce the exact same algorithm with numpy:
1. Replay the identical draw sequence — `np.random.default_rng(seed)` drawn in the same 1000-sized `rng.choice(n, p=probabilities)` blocks — until the stop condition can be evaluated.
2. Find the stop position from each source's `length`-th occurrence: `min` over sources for `first_exhausted` (stop when *any* source is exhausted), `max` for `all_exhausted` (stop when *every* source is).
3. Map each source's k-th appearance to `(k % length) + offset` (the same rolling-window-on-exhaustion behavior), vectorized per source.

`all_exhausted_without_replacement` keeps the explicit loop — its skip-on-exhaustion semantics make the output length data-dependent, so it isn't a simple function of the draw sequence.

## Bit-identical output

Because the RNG is consumed in exactly the same way and the index mapping is the same, output is bit-identical for a fixed `seed`:
- The existing hardcoded tests `test_interleave_datasets_probabilities` (`[10, 11, 20, 12, 0, 21, 13]`) and `test_interleave_datasets_probabilities_oversampling_strategy` (16-element sequence) pass **unchanged**.
- 80 randomized `(num_datasets, lengths, probabilities, seed)` cases across both `first_exhausted` and `all_exhausted` match the previous implementation exactly.
- Added `test_interleave_datasets_probabilities_is_deterministic_and_balanced` (parametrized over strategy + seed) as a regression guard.

## Benchmark

3-source mix, ~93M output rows (`all_exhausted`, probabilities given): **~90 min → ~5 s**.

## Notes

- No public API or semantics change; same results, same `_fingerprint`.
- `first_exhausted` / `all_exhausted` only; `all_exhausted_without_replacement` and the `probabilities is None` paths are untouched.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -7210,15 +7210,63 @@ def _interleave_map_style_datasets(
         # We have to keep the indices to their respective dataset offsets and to flatten to effectively interleave the datasets
         indices = (indices + offsets).flatten().tolist()
 
-    else:
-        # boolean array indicating if at index i if the dataset_i has been fully exhausted
-        is_exhausted = np.full(len(lengths), False)
+    elif stopping_strategy in ("first_exhausted", "all_exhausted"):
+        # Vectorized equivalent of the per-example Python loop below (kept for
+        # `all_exhausted_without_replacement`). For `first_exhausted` /
+        # `all_exhausted` the loop just appends, for each randomly drawn source,
+        # that source's next index with wrap-around on exhaustion (rolling
+        # window), and stops when the first (any) / every (all) source has been
+        # drawn at least `length` times. That is fully vectorizable, giving a
+        # large speedup for big outputs (e.g. ~90 min -> ~5 s for a ~93M-row
+        # interleave) while producing bit-identical output for a fixed `seed`,
+        # since we consume the RNG in the exact same 1000-sized `choice` blocks.
+        lengths_arr = np.asarray(lengths, dtype=np.int64)
+        n_datasets = len(lengths)
+        rng = np.random.default_rng(seed)
 
-        # if undersampling ("first_exhausted"), we stop as soon as one dataset is exhausted
-        # if oversampling ("all_exhausted"), we stop as soons as every dataset is exhausted, i.e as soon as every samples of every dataset has been visited at least once
-        bool_strategy_func = (
-            np.all if (oversampling or stopping_strategy == "all_exhausted_without_replacement") else np.any
-        )
+        # Draw source indices in 1000-sized blocks (matching the original
+        # iter_random_indices) until the stopping condition can be evaluated,
+        # i.e. until enough sources have reached their length.
+        blocks = []
+        counts = np.zeros(n_datasets, dtype=np.int64)
+        reached = np.zeros(n_datasets, dtype=bool)
+        while not (reached.any() if not oversampling else reached.all()):
+            block = rng.choice(n_datasets, size=1000, p=probabilities)
+            blocks.append(block)
+            counts += np.bincount(block, minlength=n_datasets)
+            reached = counts >= lengths_arr
+        draws = np.concatenate(blocks)
+
+        # For each source, the draw-position of its `length`-th occurrence (when
+        # it becomes exhausted). The original loop checks the stop condition
+        # BEFORE appending, so the last kept draw is at:
+        #   - first_exhausted: the earliest such position (min over sources that
+        #     reached length) minus... no: stop fires as soon as ANY source is
+        #     exhausted, which happens right AFTER that source's length-th draw.
+        #     So we keep draws up to and including that length-th occurrence.
+        #   - all_exhausted: the latest such position (max), inclusive.
+        exhaust_pos = np.full(n_datasets, -1, dtype=np.int64)
+        for s in range(n_datasets):
+            occ = np.flatnonzero(draws == s)
+            if len(occ) >= lengths[s]:
+                exhaust_pos[s] = occ[lengths[s] - 1]
+        valid_pos = exhaust_pos[exhaust_pos >= 0]
+        stop_at = valid_pos.min() if not oversampling else valid_pos.max()
+        used = draws[: stop_at + 1]
+
+        # Map each source's k-th appearance to its k-th index with wrap-around:
+        # concatenated row = (k % length) + offset.
+        indices_arr = np.empty(len(used), dtype=np.int64)
+        for s in range(n_datasets):
+            pos = np.flatnonzero(used == s)
+            indices_arr[pos] = (np.arange(len(pos)) % lengths[s]) + offsets[s]
+        indices = indices_arr.tolist()
+
+    else:
+        # all_exhausted_without_replacement: each sample appears exactly once, so
+        # an exhausted source is skipped rather than wrapped -- the output length
+        # is not simply the draw count, so keep the explicit per-example loop.
+        is_exhausted = np.full(len(lengths), False)
 
         def iter_random_indices():
             """Get an infinite iterator that randomly samples the index of the source to pick examples from."""
@@ -7229,24 +7277,19 @@ def _interleave_map_style_datasets(
         current_index = [0] * len(datasets)
         indices = []
         for source_idx in iter_random_indices():
-            # If no oversampling, we stop as soon as a dataset has ran out of examples (np.any)
-            # Otherwise, we stop as soon as every dataset has ran out of examples (np.all)
-            if bool_strategy_func(is_exhausted):
-                # the stopping condition was reached, let's stop
+            # Stop as soon as every dataset has run out of examples (np.all).
+            if np.all(is_exhausted):
                 break
 
-            # let's add the example at the current index of the `source_idx`-th dataset
-            # For without replacement sampling we additionally need to make sure the current source is not exhausted to not oversample.
-            if stopping_strategy != "all_exhausted_without_replacement" or not is_exhausted[source_idx]:
+            # Add the example at the current index of the source_idx-th dataset,
+            # unless that source is already exhausted (no oversampling).
+            if not is_exhausted[source_idx]:
                 indices.append(current_index[source_idx] + offsets[source_idx])
                 current_index[source_idx] += 1
 
-            # we've ran out of examples for the current dataset, let's update our boolean array and bring the current_index back to 0
+            # Mark exhaustion; do not reset the index (no replacement).
             if current_index[source_idx] >= lengths[source_idx]:
                 is_exhausted[source_idx] = True
-                # We don't want to reset the iterator when stopping strategy is without replacement.
-                if stopping_strategy != "all_exhausted_without_replacement":
-                    current_index[source_idx] = 0
 
     return concatenated_datasets.select(indices, **kwargs)
 
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -7222,45 +7222,56 @@ def _interleave_map_style_datasets(
         # since we consume the RNG in the exact same 1000-sized `choice` blocks.
         lengths_arr = np.asarray(lengths, dtype=np.int64)
         n_datasets = len(lengths)
-        rng = np.random.default_rng(seed)
 
-        # Draw source indices in 1000-sized blocks (matching the original
-        # iter_random_indices) until the stopping condition can be evaluated,
-        # i.e. until enough sources have reached their length.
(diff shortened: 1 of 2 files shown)


src/datasets/arrow_dataset.py
```diff
@@ -7210,16 +7210,64 @@ def _interleave_map_style_datasets(
...
+        # that source's next index with wrap-around on exhaustion (rolling
+        # window), and stops when the first (any) / every (all) source has been
+        # drawn at least `length` times. That is fully vectorizable, giving a
+        # large speedup for big outputs (e.g. ~90 min -> ~5 s for a ~93M-row
+        # interleave) while producing bit-identical output for a fixed `seed`,
+        # since we consume the RNG in the exact same 1000-sized `choice` blocks.
+        lengths_arr = np.asarray(lengths, dtype=np.int64)
+        n_datasets = len(lengths)
+        rng = np.random.default_rng(seed)
+
```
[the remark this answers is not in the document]
saforem2: Good catch, fixed in 71b03df. An empty source made `blocks` stay empty and `np.concatenate([])` raise; the previous (loop) implementation also crashed here, with a cryptic `IndexError: Index N out of range`. Now raises a clear `ValueError` naming the empty dataset indices, for both `first_exhausted` and `all_exhausted` (an empty source is degenerate either way, and silently dropping it would change results). Added a parametrized test.

src/datasets/arrow_dataset.py
```diff
@@ -7210,16 +7210,64 @@ def _interleave_map_style_datasets(
...
+        draws = np.concatenate(blocks)
+
+        # For each source, the draw-position of its `length`-th occurrence (when
+        # it becomes exhausted). The original loop checks the stop condition
+        # BEFORE appending, so the last kept draw is at:
+        #   - first_exhausted: the earliest such position (min over sources that
+        #     reached length) minus... no: stop fires as soon as ANY source is
+        #     exhausted, which happens right AFTER that source's length-th draw.
+        #     So we keep draws up to and including that length-th occurrence.
+        #   - all_exhausted: the latest such position (max), inclusive.
```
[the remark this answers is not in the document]
saforem2: Fixed in 71b03df -- removed the "minus... no:" thinking-out-loud and replaced it with a clear final statement: a source is exhausted right after its `length`-th draw, so the inclusive stop position is the earliest such position over sources for `first_exhausted` (any) and the latest for `all_exhausted` (all).

src/datasets/arrow_dataset.py
```diff
@@ -7210,16 +7210,64 @@ def _interleave_map_style_datasets(
...
+        exhaust_pos = np.full(n_datasets, -1, dtype=np.int64)
+        for s in range(n_datasets):
+            occ = np.flatnonzero(draws == s)
+            if len(occ) >= lengths[s]:
+                exhaust_pos[s] = occ[lengths[s] - 1]
```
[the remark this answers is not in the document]
saforem2: I benchmarked the argsort-based single-pass approach against the current per-source `np.flatnonzero`, and the per-source version is actually faster in both the few-dataset and many-dataset cases, so I've kept it.

At 93M draws (dummy lengths, index-mapping section only):

| case | per-source `flatnonzero` (current) | argsort + bincount |
|---|---|---|
| 3 datasets | **1.47s** | 5.22s |
| 50 datasets | **7.58s** | 12.13s |

The `O(n_datasets * n_draws)` scaling is real in complexity terms, but the constant factors favor `flatnonzero`: it's a cheap fully-vectorized compare-and-pack, whereas the argsort does an `O(n_draws log n_draws)` sort of a 93M-element array that dominates even at 50 datasets. `flatnonzero`'s boolean temporaries are `n_draws` bools each (transient, one source at a time), so peak memory stays modest.

`flatnonzero` would only lose for a very large number of datasets (roughly where `n_datasets` approaches `log(n_draws)`); interleave mixes are typically a handful of sources, so this stays on the fast side in practice. Happy to switch to the argsort form if you'd prefer the better asymptotic behavior for pathological `n_datasets` regardless -- but for the common case the current code is the faster of the two.

saforem2:
Hi @lhoestq — gentle ping on this one when you have a moment.

It vectorizes the `probabilities`-given index generation in `_interleave_map_style_datasets` (the `first_exhausted` / `all_exhausted` branches), which currently builds the index list in a pure-Python per-row loop. This brings it to parity with the already-vectorized `probabilities=None` branch in the same function.

- **Bit-identical output** for a fixed seed (added/extended tests assert new == old across seeds, lengths, probabilities, and n-datasets).
- **~90 min → ~5 s** to build a real 93M-row interleaved mix (tulu-3 + OpenMathInstruct-2 + ultrachat at given weights) — the per-row loop was pure interpreter overhead, not compute.

Looks like CI is awaiting maintainer approval to run (external contributor). Happy to rebase or adjust anything — thanks!

saforem2:
Following up here -- tagging a couple of folks who've worked in this exact function, in case you have a moment to take a look:

@prathamk-tw -- you added the `all_exhausted_without_replacement` branch in #7955, which is the one branch this PR intentionally leaves as the explicit per-row loop (its skip-on-exhaustion semantics make the output length data-dependent, so it isn't a pure function of the draw sequence). The `first_exhausted` / `all_exhausted` branches are what get vectorized here.

@CloseChoice -- since you worked on seed handling for interleave in #7823, you might be interested in the correctness argument: the vectorized path replays the identical RNG draw sequence (`np.random.default_rng(seed)`, same 1000-sized `rng.choice(n, p=probabilities)` blocks), so output is bit-identical for a fixed seed. The existing hardcoded tests pass unchanged, plus 80 randomized cases match the old implementation exactly.

Net effect is ~90 min -> ~5 s to build a real ~93M-row mix, with no API or semantics change. Happy to adjust anything -- thanks!

src/datasets/arrow_dataset.py
```diff
@@ -7210,16 +7210,75 @@ def _interleave_map_style_datasets(
...
+        lengths_arr = np.asarray(lengths, dtype=np.int64)
+        n_datasets = len(lengths)
+
+        # Empty sources: a length-0 dataset can never be sampled to its length,
+        # so the stopping condition is ill-defined. The previous implementation
+        # crashed here with a cryptic `IndexError: Index N out of range` (it
+        # sampled the empty source and indexed into it). Raise a clear error
+        # instead -- for both strategies, since an empty source is degenerate
+        # either way and silently dropping it would change results.
+        if np.any(lengths_arr == 0):
```
Sanjays2402: the empty-source ValueError is stricter than what it replaces. today a zero-length dataset only blows up if it actually gets drawn, so `interleave_datasets([d3, d0], probabilities=[1.0, 0.0], seed=7)` with first_exhausted returns [0,1,2] fine (checked against the current loop). after this it raises. worth gating the error on `probabilities[i] > 0` rather than on length alone?

Sanjays2402: the empty-source ValueError also fires when the empty dataset has probability 0, which used to work fine: with lengths [3, 0] and probabilities [1.0, 0.0] the old loop never draws source 1, so it just returns [0, 1, 2]. gating on `lengths_arr == 0` alone turns that into a hard error. `(lengths_arr == 0) & (np.asarray(probabilities) > 0)` keeps the cryptic-IndexError fix without breaking the zero-probability case.

lhoestq:
can you run `make style` to fix code formatting so we can run the CI ?

src/datasets/arrow_dataset.py
saforem2: Good catch, thanks — you're right, and it's fixed in f58810731.

I confirmed the regression against the pre-vectorization loop: `lengths=[3, 0]`, `probabilities=[1.0, 0.0]`, `first_exhausted` returns `[0, 1, 2]` on the old loop, and gating the error on `lengths_arr == 0` alone turned that into a hard error.

Rather than only relaxing the raise, I made "probability 0 ⇒ never drawn" explicit throughout the branch, since gating the error alone still left two crashes:

1. **The error gate** is now `length == 0 and probability > 0`, as you suggested.
2. **The stop condition** now only considers drawable sources. A length-0 source satisfies `counts >= lengths` (`0 >= 0`) immediately, so it made `reached.any()` true on the first block — with `lengths=[5000, 0]`, `p=[1.0, 0.0]` the draw loop stopped after 1000 draws before source 0 had reached its length, leaving `valid_pos` empty and `.min()` raising.
3. **The index mapping / `exhaust_pos` loops** skip non-drawable sources. Without that, a zero-probability source has `len(occ) == 0 >= lengths[s] == 0`, so `occ[lengths[s] - 1]` indexes `occ[-1]` into an empty array → `IndexError`.

One extra thing surfaced while testing: under `all_exhausted`, a probability-0 source can never be exhausted, and the **old loop hung forever** (not just for empty sources — `lengths=[3, 5]`, `p=[1.0, 0.0]`, `all_exhausted` never terminates). That now raises a clear `ValueError` instead of spinning. Happy to drop that part if you'd rather keep this PR strictly to the vectorization.

Verified bit-identical to the old loop across 400 randomized `(n_datasets, lengths, probabilities, seed)` cases over both strategies, and added regression tests covering the zero-probability empty source, the zero-probability non-empty source, and the `all_exhausted` raise.

saforem2: Fixed in f58810731 — see the [detailed reply on your follow-up thread](https://github.com/huggingface/datasets/pull/8318#discussion_r3836157715).

Short version: you were right that gating on length alone regressed the working `probabilities=[1.0, 0.0]` case, so the raise is now gated on `length == 0 and probability > 0`. Zero-probability sources are also excluded from the stop condition and the index mapping (both of which still crashed otherwise), and `all_exhausted` with a probability-0 source now raises instead of hanging forever as the old loop did. Thanks for catching this.

saforem2:
Done — `make style` applied in f58810731, and `make quality` now passes clean:

```
ruff check tests src benchmarks utils setup.py  # linter
All checks passed!
ruff format --check tests src benchmarks utils setup.py # formatter
232 files already formatted
```

(The linter needed one manual fix first: a `dict(...)` call in the new test tripped `C408`, which has no safe autofix, so `ruff check --fix` exited non-zero and `make style` stopped before reaching the formatter. Rewrote it as a dict literal.)

While in here I also addressed @Sanjays2402's outstanding review on the empty-source `ValueError` — it was gated on length alone, which regressed the previously-working case of a **zero-probability** empty source (`lengths=[3, 0]`, `probabilities=[1.0, 0.0]` under `first_exhausted` returns `[0, 1, 2]` on the old loop). Zero-probability sources are now consistently treated as never-drawn: excluded from the error gate, the stop condition, and the index mapping. One related pre-existing bug surfaced too — under `all_exhausted` a probability-0 source can never be exhausted and the old loop **hung forever**; that now raises a clear `ValueError`. Details [here](https://github.com/huggingface/datasets/pull/8318#discussion_r3836157715).

Re-verified bit-identical output vs. the pre-vectorization loop across 400 randomized `(n_datasets, lengths, probabilities, seed)` cases over both strategies; the existing hardcoded sequence tests pass unchanged, and `pytest tests/test_arrow_dataset.py -k interleave` is 15 passed.

CI should be unblocked now — thanks for taking a look!

saforem2:
@lhoestq I think this should be good to go now!

PR #8323, opened 2026-07-11
Raise on Dataset.rename_columns name collision with an existing column

roli-lpci:
### What

`Dataset.rename_column` (singular) already rejects a new column name that collides with an
existing column:

```python
>>> ds.rename_column("foo", "bar")  # "bar" already exists
ValueError: New column name bar already in the dataset. ...
```

`Dataset.rename_columns` (plural, dict-based) has no equivalent check. Renaming a column to a
name that collides with an existing, untouched column silently produces a `Dataset` whose
underlying Arrow table has two columns with the same name, while `dataset.features` (built
from a Python dict) collapses to one entry for that name. The two become inconsistent, and in
the reproduction below the renamed column's data becomes unreachable through the normal read
paths (`ds[:]`, `to_dict()`, `to_pandas()`, iteration, ...) with no exception raised anywhere.

```python
from datasets import Dataset

ds = Dataset.from_dict({"foo": [1, 2, 3], "bar": [10, 20, 30], "baz": [100, 200, 300]})
ds2 = ds.rename_columns({"foo": "bar"})

ds2.column_names   # ['bar', 'bar', 'baz']  -- two Arrow columns named "bar"
ds2.features       # {'bar': ..., 'baz': ...} -- only one "bar" entry
ds2[:]             # {'bar': [10, 20, 30], 'baz': [...]}  -- foo's [1, 2, 3] is gone
```

Since `DatasetDict.rename_columns` forwards directly to `Dataset.rename_columns` per split,
the same corruption is reachable from `DatasetDict` too.

### Fix

Add the same collision check `rename_column` already performs, generalized to the mapping
case: a new name is rejected only if it collides with a column that isn't itself being renamed
away in the same call (so swaps like `{"a": "b", "b": "a"}` and no-op renames like
`{"a": "a"}` keep working).

```python
colliding_new_columns = set(column_mapping.values()) & (
    set(dataset.column_names) - set(column_mapping.keys())
)
if colliding_new_columns:
    raise ValueError(
        f"New column names {colliding_new_columns} already in the dataset. "
        f"Please choose column names which are not already in the dataset. "
        f"Current columns in the dataset: {dataset._data.column_names}"
    )
```

### Test

Added to `tests/test_arrow_dataset.py::BaseDatasetTest.test_rename_columns`:
- Asserts `ValueError` on a colliding rename (`{"col_1": "col_2"}` where `col_2` already
  exists and isn't being renamed).
- Asserts a swap (`{"col_1": "col_2", "col_2": "col_1"}`) still succeeds, to confirm the fix
  doesn't over-reject legitimate renames.

Confirmed the added assertions fail on `main` with `AssertionError: ValueError not raised`
(not a generic error) and pass with the fix (`pytest tests/test_arrow_dataset.py -k rename`:
2 failed on the unmodified baseline, 4 passed with the fix). A prior full run of
`tests/test_arrow_dataset.py` with this change: 360 passed, 61 skipped, 1 pre-existing
unrelated failure (a PyTorch-formatting test that fails identically on `main` because torch
isn't installed in the test env).

### Compatibility

Pure Python, no new imports, no version-gated syntax. Only affects the specific case where a
`rename_columns` mapping's target name collides with an existing column outside the mapping;
that call previously produced silently-corrupted output, so this is a bug fix, not a behavior
users could be relying on for correct results.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -2655,6 +2655,16 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         if empty_new_columns:
             raise ValueError(f"New column names {empty_new_columns} are empty.")
 
+        colliding_new_columns = set(column_mapping.values()) & (
+            set(dataset.column_names) - set(column_mapping.keys())
+        )
+        if colliding_new_columns:
+            raise ValueError(
+                f"New column names {colliding_new_columns} already in the dataset. "
+                f"Please choose column names which are not already in the dataset. "
+                f"Current columns in the dataset: {dataset._data.column_names}"
+            )
+
         def rename(columns):
             return [column_mapping[col] if col in column_mapping else col for col in columns]
 
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -2655,9 +2655,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         if empty_new_columns:
             raise ValueError(f"New column names {empty_new_columns} are empty.")
 
-        colliding_new_columns = set(column_mapping.values()) & (
-            set(dataset.column_names) - set(column_mapping.keys())
-        )
+        colliding_new_columns = set(column_mapping.values()) & (set(dataset.column_names) - set(column_mapping.keys()))
         if colliding_new_columns:
             raise ValueError(
                 f"New column names {colliding_new_columns} already in the dataset. "
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -110,9 +110,10 @@ def _rename_columns_fn(example: dict, column_mapping: dict[str, str]):
         raise ValueError(
             f"Error when renaming {list(column_mapping)} to {list(column_mapping.values())}: columns {set(column_mapping) - set(example)} are not in the dataset."
         )
-    if any(col in example for col in column_mapping.values()):
+    colliding_new_columns = set(column_mapping.values()) & (set(example) - set(column_mapping))
+    if colliding_new_columns:
         raise ValueError(
-            f"Error when renaming {list(column_mapping)} to {list(column_mapping.values())}: columns {set(example) - set(column_mapping.values())} are already in the dataset."
+            f"Error when renaming {list(column_mapping)} to {list(column_mapping.values())}: columns {colliding_new_columns} are already in the dataset."
         )
     return {
         new_column_name: example[original_column_name]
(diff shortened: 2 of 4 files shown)


src/datasets/arrow_dataset.py
```diff
@@ -2655,6 +2655,16 @@ def rename_columns(self, column_mapping: dict[str, str], new_fingerprint: Option
         if empty_new_columns:
             raise ValueError(f"New column names {empty_new_columns} are empty.")
 
+        colliding_new_columns = set(column_mapping.values()) & (
+            set(dataset.column_names) - set(column_mapping.keys())
+        )
+        if colliding_new_columns:
+            raise ValueError(
+                f"New column names {colliding_new_columns} already in the dataset. "
```
Sanjays2402, line 2661: the streaming sibling has the same hole but worse: `IterableDataset.rename_columns` does no check at all, and `_rename_columns_fn` tests `any(col in example for col in column_mapping.values())` before the old columns are removed, so on 5.0.1 even the swap case your test blesses here (`{"a": "b", "b": "a"}`) blows up at iteration with `columns set() are already in the dataset.` — and `{"a": "b"}` silently gives `features == {"b": ...}` while erroring only once you iterate. worth mirroring this eager check there so both paths fail the same way?

Sanjays2402:
the swap case you added a test for only works on Dataset — the streaming side diverges. IterableDataset.rename_columns goes through _rename_columns_fn, which raises if `any(col in example for col in column_mapping.values())`, so `{"col_1": "col_2", "col_2": "col_1"}` blows up at iteration time even though remove_columns would have cleared both. worth aligning that check the same way you did the collision check here (ignore new names that are themselves being renamed away), otherwise the two APIs now disagree on what a valid mapping is.

src/datasets/arrow_dataset.py
```diff
@@ -2655,6 +2655,16 @@ def rename_columns(self, column_mapping: dict[str, str], new_fingerprint: Option
         if empty_new_columns:
             raise ValueError(f"New column names {empty_new_columns} are empty.")
 
+        colliding_new_columns = set(column_mapping.values()) & (
```
Sanjays2402: IterableDataset.rename_columns has no equivalent guard, so the two APIs diverge after this: streaming still only catches the collision inside _rename_columns_fn at iteration time, and that check is `any(col in example for col in column_mapping.values())`, which also rejects the swap case you added a test for here ({"col_1": "col_2", "col_2": "col_1"} — both new names are present in the example). Worth mirroring the eager check in iterable_dataset.py so the swap works and the collision fails at call time rather than on first next()?

roli-lpci:
Addressed the streaming parity feedback and rebased this branch onto current `main` in `8b113d45f`.

- `IterableDataset` swap renames now ignore destination names that are themselves being renamed away.
- Collisions with untouched columns still raise; when features are known they now raise eagerly at `rename_columns()`, while unknown-feature streams reject during iteration.
- Added regression coverage for both swap and collision behavior with unknown and resolved features.

Validation on the refreshed branch:

- `pytest -q tests/test_iterable_dataset.py -k rename_column` — 4 passed
- `pytest -q tests/test_arrow_dataset.py -k rename_columns` — 2 passed
- Ruff check and format check on all four affected files — passed

I also reproduced both reported streaming failures against the unpatched behavior before applying the fix.

PR #8325, opened 2026-07-12
Fix path traversal via metadata file_name in folder-based builders

### LINKED ISSUE #8324: Path traversal / arbitrary file read via the `file_name` metadata field in `*folder` dataset builders (no `trust_remote_code`)
### Describe the bug

reported via email on 9 June 2026 - no acknowledgement or response:

Package: datasets (PyPI). Affected file: src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
Affected Versions: confirmed on 5.0.0 (PyPI) and HEAD 63e103e.
CVSS: CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:U/C:H/I:N/A:N  (~7.1 High when the exfil/re-embed chain applies; Medium for local read only)
CWE: CWE-22 Path Traversal

## Summary

The built-in `imagefolder` / `audiofolder` / `videofolder` / `pdffolder` (`*folder`) builders read a `file_name` value verbatim from a dataset's `metadata.jsonl`/`.csv`/`.parquet` and join it to the dataset directory without neutralizing `..` or absolute paths. A malicious dataset can therefore make `load_dataset(...)` open and read files **outside** the dataset directory. No `trust_remote_code` is required (these are packaged builders).

## Details

`src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py` (~L404-405):

```python
file_relpath = os.path.normpath(item).replace("\\", "/")   # item = attacker-controlled file_name
item = os.path.join(downloaded_metadata_dir, file_relpath)  # join escapes the dataset dir
```

`os.path.normpath` canonicalizes separators but does **not** strip a leading `../`, so `../../../../etc/passwd` survives, and an absolute value escapes entirely. No `commonpath`/`relpath`/`realpath` containment check exists. The resolved path is stored in the `Image`/`Audio`/`Pdf` feature and then read:
- `features/image.py:175` `PIL.Image.open(path)` (lazy read on row access), and
- `features/image.py:311` `xopen(path,"rb").read()` in `embed_storage` — the unconditional raw-byte read used by `save_to_disk` / `push_to_hub`.

## PoC 
`metadata.jsonl` in an `imagefolder` dataset:
```json
{"file_name": "../victim/secret.png", "text": "x"}
```
```python
d = load_dataset("imagefolder", data_dir=malicious_dir, split="train")
d[0]["image"]   # the image OUTSIDE the dataset dir
```
Observed: `OUT-OF-DATASET READ: True` (loaded the out-of-directory image). With `file_name: "../../../../../../etc/passwd"`, the path resolves to `/etc/passwd`; the `embed_storage` path (invoked by `save_to_disk`/`push_to_hub`) pulled the bytes of `/etc/passwd` into the saved output (`EXFIL CONFIRMED`).

## Impact

Loading a malicious dataset reads arbitrary local files; when the victim re-saves/re-publishes (`save_to_disk`/`push_to_hub`) the dataset, arbitrary local file contents are silently embedded into the output and exfiltrated. Triggered by the ordinary `load_dataset("imagefolder", data_dir=...)` and identically for a remote Hub dataset `load_dataset("user/dataset")`.

## Net-new

No matching CVE/GHSA. Distinct from the tarfile (CVE-2007-4559) mitigations in `utils/extract.py`, which were tested separately and correctly block `..`/symlink escapes — the flaw here is the unsanitized `file_name` metadata join, not archive extraction.

## Remediation

Reject `file_name` values that are absolute or contain `..` (require the resolved path to remain under the dataset/metadata directory via `os.path.realpath` + `os.path.commonpath`, or restrict to `os.path.basename`-relative paths within the tree).

### Expected behavior

-

### Environment info

-

Kaif10:
Fix #8324

### Vulnerability

The folder-based builders (`imagefolder`, `audiofolder`, `videofolder`, `pdffolder`) resolve each metadata row's `file_name` by joining it to the metadata file's directory with no containment check. A malicious dataset can set `file_name` to a `../`-traversal path or an absolute path, causing the builder to open and read files outside the dataset directory when the dataset is loaded — arbitrary file read (CWE-22).

### Fix

`FolderBasedBuilder._generate_examples` now rejects, with a `ValueError` naming the offending `file_name`, any value that is absolute or uses `..` traversal to escape the directory containing the metadata file. The validation is performed on the relative reference so it applies uniformly to local dataset directories and to the fsspec URLs used when reading from downloaded archives; `os.path.realpath` / `os.path.commonpath` were deliberately not used on the joined path because those functions are not fsspec-aware (unlike the streaming-patched `os.path.join`) and would corrupt archive URLs such as `zip://...::...`. For genuine local paths, symlinks are additionally resolved and containment is verified with `os.path.commonpath`. The containment root is the metadata file's own directory, matching the existing resolution base and preserving all legitimate relative references into subdirectories.

### Tests

Added regression tests in `tests/packaged_modules/test_folder_based_builder.py`: `../` traversal rejected (three variants, including `subdir/../../` forms), absolute path rejected, and legitimate same-dir / `subdir/` / `./subdir/` references still load. These fail on the pre-fix code and pass with the fix. `test_folder_based_builder.py` (51 passed) and `test_imagefolder.py` (36 passed).

Follows the containment approach of #8303.

*AI assistance (Claude) was used in preparing this change; it was reviewed and verified locally.*

### DIFF
--- a/src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
+++ b/src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
@@ -2,7 +2,6 @@ import collections
 import io
 import itertools
 import os
-import posixpath
 from dataclasses import dataclass
 from typing import Any, Callable, Iterator, Optional, Union
 
@@ -432,17 +431,7 @@ class FolderBasedBuilder(datasets.GeneratorBasedBuilder):
                                     f"and parent-directory ('..') traversal that escape the dataset directory are "
                                     f"not allowed."
                                 )
-                            resolved_reference = os.path.join(downloaded_metadata_dir, file_relpath)
-                            # Defense in depth: confirm the resolved reference stays inside the
-                            # metadata directory. This works for local dirs (resolving symlinks) as
-                            # well as for the fsspec URLs (possibly chained with "::") produced when
-                            # reading from a downloaded archive. See `_metadata_reference_is_contained`.
-                            if not _metadata_reference_is_contained(downloaded_metadata_dir, resolved_reference):
-                                raise ValueError(
-                                    f"Invalid metadata file_name '{item}': the resolved path escapes the "
-                                    f"dataset directory containing the metadata file."
-                                )
-                            item = resolved_reference
+                            item = os.path.join(downloaded_metadata_dir, file_relpath)
                     return item
 
                 for pa_metadata_table in self._read_metadata(downloaded_metadata_file, metadata_ext=metadata_ext):
@@ -469,47 +458,6 @@ class FolderBasedBuilder(datasets.GeneratorBasedBuilder):
                         if len(pa_table) == 0:
                             continue
                     yield Key(shard_idx, sample_idx), sample
-
-
-def _metadata_reference_is_contained(downloaded_metadata_dir: str, resolved_reference: str) -> bool:
-    """Return whether ``resolved_reference`` stays inside ``downloaded_metadata_dir``.
-
-    ``downloaded_metadata_dir`` is the directory containing the metadata file and
-    ``resolved_reference`` is a metadata ``file_name`` joined onto it. Both may be plain local
-    paths or fsspec URLs, and either may be a chained URL that uses the "::" hop separator to
-    navigate a downloaded archive, e.g.:
-
-    * ``/path/to/local/metadata_dir``
-    * ``hf://path/to/remote/metadata_dir``
-    * ``zip://metadata_dir::/path/to/local/archive.zip``
-    * ``zip://metadata_dir::hf://path/to/remote/archive.zip``
-
-    Joining a relative ``file_name`` only extends the first "::" component (the path *inside* the
-    archive, or the whole reference when there is no archive); the remaining components describe
-    the fixed container and must not change. So containment is checked on the first component,
-    while every subsequent component must stay identical.
-    """
-    root_parts = downloaded_metadata_dir.split("::")
-    reference_parts = resolved_reference.split("::")
-    # The container part(s) after the first hop are fixed by the download step; a `file_name`
-    # must never alter them.
-    if root_parts[1:] != reference_parts[1:]:
-        return False
-    root, reference = root_parts[0], reference_parts[0]
-    if "://" in root:
-        # fsspec URL component: normalize textually with posixpath. os.path must not be used here
-        # because on Windows it rewrites "/" as "\\" and collapses the "//" of the URL scheme.
-        root_norm = posixpath.normpath(root)
-        reference_norm = posixpath.normpath(reference)
-        return reference_norm == root_norm or reference_norm.startswith(root_norm + "/")
-    # Genuine local path: resolve symlinks, then confirm containment.
-    real_root = os.path.realpath(root)
-    real_reference = os.path.realpath(reference)
-    try:
-        return os.path.commonpath([real_root, real_reference]) == real_root
-    except ValueError:
-        # commonpath raises ValueError for paths on different drives (Windows).
-        return False
 
 
 def _nested_apply(item: Any, feature_path: _VisitPath, func: Callable[[Any, _VisitPath], Any]):
--- a/src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
+++ b/src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
@@ -402,7 +402,41 @@ class FolderBasedBuilder(datasets.GeneratorBasedBuilder):
                     elif len(feature_path) == 0:
                         if item is not None:
                             file_relpath = os.path.normpath(item).replace("\\", "/")
+                            # Guard against path traversal (CWE-22): a crafted `file_name` such as
+                            # "../../etc/passwd" or an absolute path must not be able to escape the
+                            # metadata file's directory and read arbitrary files on the host. The
+                            # check is performed on the relative reference (not the joined result)
+                            # so that it works both for local directories and for the fsspec URLs
+                            # (e.g. "zip://...::...") used when reading from downloaded archives.
+                            if (
+                                os.path.isabs(item)
+                                or os.path.isabs(file_relpath)
+                                or file_relpath == ".."
+                                or file_relpath.startswith("../")
+                            ):
+                                raise ValueError(
+                                    f"Invalid metadata file_name '{item}': `file_name` must be a relative path "
+                                    f"pointing inside the directory containing the metadata file. Absolute paths "
+                                    f"and parent-directory ('..') traversal that escape the dataset directory are "
+                                    f"not allowed."
+                                )
                             item = os.path.join(downloaded_metadata_dir, file_relpath)
+                            # For local paths, additionally resolve symlinks and confirm containment.
+                            # Skipped for fsspec URLs (which contain "://") since realpath is not
+                            # aware of them and would corrupt the URL.
+                            if "://" not in item:
+                                real_root = os.path.realpath(downloaded_metadata_dir)
+                                real_path = os.path.realpath(item)
+                                try:
+                                    is_contained = os.path.commonpath([real_root, real_path]) == real_root
+                                except ValueError:
+                                    # commonpath raises ValueError for paths on different drives (Windows)
+                                    is_contained = False
+                                if not is_contained:
+                                    raise ValueError(
+                                        f"Invalid metadata file_name '{item}': the resolved path escapes the "
+                                        f"dataset directory containing the metadata file."
+                                    )
                     return item
 
                 for pa_metadata_table in self._read_metadata(downloaded_metadata_file, metadata_ext=metadata_ext):
--- a/src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
+++ b/src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
@@ -2,6 +2,7 @@ import collections
 import io
 import itertools
 import os
+import posixpath
 from dataclasses import dataclass
 from typing import Any, Callable, Iterator, Optional, Union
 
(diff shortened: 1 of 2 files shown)


src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
```diff
@@ -402,7 +402,41 @@ def set_feature(item, feature_path: _VisitPath):
...
+                                    f"Invalid metadata file_name '{item}': `file_name` must be a relative path "
+                                    f"pointing inside the directory containing the metadata file. Absolute paths "
+                                    f"and parent-directory ('..') traversal that escape the dataset directory are "
+                                    f"not allowed."
+                                )
                             item = os.path.join(downloaded_metadata_dir, file_relpath)
+                            # For local paths, additionally resolve symlinks and confirm containment.
+                            # Skipped for fsspec URLs (which contain "://") since realpath is not
+                            # aware of them and would corrupt the URL.
+                            if "://" not in item:
```
lhoestq: you can point to local files using `file://../..` or `local://../..` so this check is maybe not enough

maybe we need to forbid `://` altogether ? (except for zip://<file_name>::<valid_relative_path>)

lhoestq (commented):
Thanks for reporting and for the fix, I just have one comment:

src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
Kaif10: You're right — `file://` and `local://` (and any other fsspec scheme) resolve to local files and slip past the containment check, since the old guard skipped it whenever `file_name` contained `://`. On Linux the scheme also survives `normpath`, so this is a real bypass.

I've switched to forbidding `://` in the `file_name` value entirely, as you suggested. Legitimate archive loads are unaffected: their URL is `zip://<file_name>::<container>`, and the scheme is always on the container (`downloaded_metadata_dir`), never on the `file_name` itself — so I now key the realpath containment check on the metadata directory rather than the joined path. Added regression tests for `file://` / `local://` payloads and confirmed the existing zip-archive tests (streaming and non-streaming) still pass.

(Disclosure: I used AI assistance while preparing this change.)

src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
```diff
@@ -401,8 +401,54 @@ def set_feature(item, feature_path: _VisitPath):
...
+                                    f"Invalid metadata file_name '{item}': `file_name` must be a relative path "
+                                    f"pointing inside the directory containing the metadata file. URL schemes "
+                                    f"(e.g. 'file://', 'local://') are not allowed."
+                                )
                             file_relpath = os.path.normpath(item).replace("\\", "/")
+                            if (
+                                os.path.isabs(item)
+                                or os.path.isabs(file_relpath)
+                                or file_relpath == ".."
+                                or file_relpath.startswith("../")
```
lhoestq, line 426: if the path is like `dummy/../../../` it can still escape

src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
```diff
@@ -401,8 +401,54 @@ def set_feature(item, feature_path: _VisitPath):
...
+                                    f"not allowed."
+                                )
                             item = os.path.join(downloaded_metadata_dir, file_relpath)
+                            # For local metadata directories, additionally resolve symlinks and confirm
+                            # containment. Skipped when the metadata dir is itself an fsspec URL (which
+                            # contains "://", e.g. "zip://...::..." from a downloaded archive) since
+                            # realpath is not scheme-aware and would corrupt the URL.
+                            if "://" not in downloaded_metadata_dir:
+                                real_root = os.path.realpath(downloaded_metadata_dir)
+                                real_path = os.path.realpath(item)
```
lhoestq: maybe you can have a check on both a chained url/path `zip://...::...` and a single url/path if you split `downloaded_metadata_dir` and `item` on `::` to get only the first part before passing to  `commonpath()` ?

we will need to make sure this works for these schemes:
- `/path/to/local/metadata_dir`
- `hf://path/to/remote/metadata_dir`
- `zip://metadata_dir::/path/to/local`
- `zip://metadata_dir::hf://path/to/remote`

lhoestq (commented):
Thanks ! I added a few more comments

Kaif10:
Thanks, both addressed.

**1. `dummy/../../../`:** this is actually already caught — `file_name` is run through `os.path.normpath` before the check, so `dummy/../../../etc/passwd` normalizes to `../../etc/passwd` and is rejected by the existing `..` guard. I added explicit regression tests (`dummy/../../../etc/passwd`, `subdir/../../../../outside/secret.txt`) so it stays that way.

**2. General containment check:** done as you suggested — I split both `downloaded_metadata_dir` and the joined reference on `::` and validate the first component, since a relative `file_name` only ever extends that part (the path inside the archive); the trailing container component(s) must stay identical, which also rejects tampering with the archive path. URL components are normalized with `posixpath` (os.path rewrites `/`→`\` and collapses `//` on Windows); genuine local paths still go through `realpath` + `commonpath` to resolve symlinks. This now runs for all four metadata-dir forms instead of being skipped for URLs:
- `/path/to/local/metadata_dir`
- `hf://path/to/remote/metadata_dir`
- `zip://metadata_dir::/path/to/local/archive.zip`
- `zip://metadata_dir::hf://path/to/remote/archive.zip`

Added a test covering all four (legit relative passes, sibling-dir escape and container tampering rejected), and the existing scheme-rejection and zip-archive loading tests still pass.

(Disclosure: I used AI assistance while preparing this change.)

Kaif10:
Hi @lhoestq — I pushed the revision addressing both points: `file_name` values containing a scheme (`file://`, `local://`, etc.) are now rejected outright, and the containment check runs for every metadata-dir form (local, `hf://`, and the two `zip://…::…` chained variants) by splitting on `::` and comparing the relevant path component, with `dummy/../../../` covered by tests. The scheme-rejection, containment, and existing zip-archive tests all pass. Let me know if there's anything else you'd like adjusted.

Kaif10:
@lhoestq gentle ping — I pushed the revised version that (1) forbids fsspec schemes in `file_name`, (2) keeps the `dummy/../../../` case covered, and (3) generalizes the containment check across the four metadata-dir forms by splitting on `::` (local, `hf://`, and both `zip://…::…` chained variants), with tests for each. Does that address your concerns, or is there anything else you'd like changed? Happy to iterate.

lhoestq:
ah yes `dummy/../../../` was already covered indeed, good catch

lhoestq:
That also means the second check `_metadata_reference_is_contained` is maybe not useful. I see it doesn't impact the tests at least.

Kaif10:
Good question on the second check — I dug into it. The `_metadata_reference_is_contained` realpath check does guard against a case the string checks can't: a symlink escape. If a subdirectory of the dataset is a symlink pointing outside the dataset root, a `file_name` like `evil_link/secret.txt` looks like a perfectly innocent relative path — no scheme, not absolute, no `..` — so the string checks pass it, but it resolves to a file outside the dataset. Only the realpath/commonpath containment check catches that.

I confirmed it end-to-end: with the containment check in place the reference is rejected ("the resolved path escapes the dataset directory"); with it removed, the escape succeeds and the read resolves to the file outside the dataset dir. I've added a regression test (`test_data_files_with_metadata_symlink_escape_is_rejected`) that asserts on that specific behavior, so it fails if the check is ever dropped. It creates a real symlink and skips where symlink creation needs privileges (e.g. Windows without Developer Mode), so it still runs on CI.

You were right that the earlier `dummy/../../../` case is already covered by the `..` check — the containment check earns its keep specifically on symlinks.

src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
```diff
@@ -401,8 +402,47 @@ def set_feature(item, feature_path: _VisitPath):
...
+                            # Defense in depth: confirm the resolved reference stays inside the
+                            # metadata directory. This works for local dirs (resolving symlinks) as
+                            # well as for the fsspec URLs (possibly chained with "::") produced when
+                            # reading from a downloaded archive. See `_metadata_reference_is_contained`.
+                            if not _metadata_reference_is_contained(downloaded_metadata_dir, resolved_reference):
+                                raise ValueError(
+                                    f"Invalid metadata file_name '{item}': the resolved path escapes the "
+                                    f"dataset directory containing the metadata file."
+                                )
+                            item = resolved_reference
```
lhoestq:
```suggestion
                            item = os.path.join(downloaded_metadata_dir, file_relpath)
```

src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
```diff
@@ -2,6 +2,7 @@
 import io
 import itertools
 import os
+import posixpath
```
lhoestq:
```suggestion
```

src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
```diff
@@ -431,6 +471,47 @@ def set_feature(item, feature_path: _VisitPath):
...
+    # Genuine local path: resolve symlinks, then confirm containment.
+    real_root = os.path.realpath(root)
+    real_reference = os.path.realpath(reference)
+    try:
+        return os.path.commonpath([real_root, real_reference]) == real_root
+    except ValueError:
+        # commonpath raises ValueError for paths on different drives (Windows).
+        return False
+
+
```
lhoestq:
```suggestion
```

tests/packaged_modules/test_folder_based_builder.py
```diff
@@ -11,6 +13,7 @@
 from datasets.packaged_modules.folder_based_builder.folder_based_builder import (
     FolderBasedBuilder,
     FolderBasedBuilderConfig,
+    _metadata_reference_is_contained,
```
lhoestq:
```suggestion
```

tests/packaged_modules/test_folder_based_builder.py
```diff
@@ -449,6 +452,206 @@ def test_data_files_with_wrong_metadata_file_name(cache_dir, tmp_path, auto_text
...
+    data_files = DataFilesDict.from_patterns(get_data_patterns(str(data_dir)), data_dir.as_posix())
+    autofolder = DummyFolderBasedBuilder(data_files=data_files, cache_dir=cache_dir)
+    gen_kwargs = autofolder._split_generators(StreamingDownloadManager())[0].gen_kwargs
+    # The string checks accept this relative path; only the realpath-based containment check
+    # rejects it, with this specific message. Asserting on that message ensures the test would
+    # fail (the file would be read) if the containment check were removed.
+    with pytest.raises(ValueError, match="the resolved path escapes the dataset directory"):
+        list(autofolder._generate_examples(**gen_kwargs))
+
+
```
lhoestq:
```suggestion
```

tests/packaged_modules/test_folder_based_builder.py
```diff
@@ -449,6 +452,206 @@ def test_data_files_with_wrong_metadata_file_name(cache_dir, tmp_path, auto_text
...
+        sibling_head = head.rsplit("/", 1)[0] + "/evil_sibling"
+        escaping = sibling_head + metadata_dir[len(head) :]
+        assert not _metadata_reference_is_contained(metadata_dir, escaping), f"sibling escape allowed for {name}"
+
+        # for chained URLs, tampering with the container part (after "::") must be rejected
+        if "::" in metadata_dir:
+            tampered = legit.split("::")[0] + "::/attacker/other.zip"
+            assert not _metadata_reference_is_contained(metadata_dir, tampered), f"container tamper allowed for {name}"
+
+
```
lhoestq:
```suggestion
```

lhoestq (approved):
I see, thx ! I'll merge the PR without syminks for now to keep things simple and since the TAR and RAR extractor are fixed to not allow symlinks already, ZIP in python simply doesn't support them at all, and 7z in python natively disallow them. Happy to discuss this further in a separate issue though

PR #8335, opened 2026-07-15
Fix column loss in horizontal concatenation of arrow-backed IterableDatasets

Vedant-Agarwal:
## Bug

`concatenate_datasets([...], axis=1)` over arrow-backed `IterableDataset`s (e.g. from `.to_iterable_dataset()`) silently drops the **first** dataset's columns and duplicates the **last** dataset's column.

In `HorizontallyConcatenatedMultiSourcesExamplesIterable._iter_arrow`, the merged table is built by appending each later source's columns onto the running accumulator `new_pa_table`. But it appended onto `pa_table` — the loop variable left over from the fetch loop just above, which holds the *last* source's table — and reassigned `new_pa_table` on each iteration instead of accumulating onto it.

## Reproduction

```python

a = Dataset.from_dict({"a": [1, 2, 3]}).to_iterable_dataset()
b = Dataset.from_dict({"b": [4, 5, 6]}).to_iterable_dataset()
print(list(concatenate_datasets([a, b], axis=1)))
# Expected: [{'a': 1, 'b': 4}, {'a': 2, 'b': 5}, {'a': 3, 'b': 6}]
# Actual (before fix): column 'a' is lost (nulled) and 'b' is duplicated,
#                      which then fails casting to the merged features
```

The non-arrow `__iter__` path is correct (it uses `new_example.update(...)`); only the arrow path was affected. That's why the existing `test_concatenate_datasets_axis_1*` tests — which use `ExamplesIterable` and never hit `_iter_arrow` — didn't catch it.

## Fix

Accumulate onto `new_pa_table` instead of `pa_table` (one line).

## Test

Added `test_concatenate_datasets_axis_1_arrow_backed`, which exercises the arrow-backed path and fails without the fix.

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -1176,7 +1176,7 @@ class HorizontallyConcatenatedMultiSourcesExamplesIterable(_BaseExamplesIterable
                         new_pa_table = table
                     else:
                         for name, col in zip(table.column_names, table.columns):
-                            new_pa_table = pa_table.append_column(name, col)
+                            new_pa_table = new_pa_table.append_column(name, col)
                 new_key = "_".join(str(key) for key in keys)
                 yield new_key, new_pa_table
             else:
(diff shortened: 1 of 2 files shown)


Vedant-Agarwal:
Closing this — on review the fix only covers the equal-length case. The arrow (`_iter_arrow`) path also needs to pad different-length sources with nulls the way the non-arrow path does, otherwise horizontal concat of unequal-length sources raises `ArrowInvalid`. I'd rather rework it to handle that completely before it's reviewed. Apologies for the noise.

PR #8338, opened 2026-07-16
Include file size in dataset fingerprint

yuxin00j:
### Motivation
Currently, the dataset cache fingerprints rely entirely on the `ETag`/`etag`/`mtime` metadata fetched during `fs.info()`. If a file's byte size changes on the remote end—and the remote backend fails to reliably update the ETag—the dataset fingerprint stays improperly static. This leads to insidious bugs where an old, structurally distinct dataset cache continues mapping to the new remote file.

### Summary of Changes
* **Modified `_get_single_origin_metadata` in `src/datasets/data_files.py`**:
  Checks if `size` is present in the underlying filesystem's `info` payload, and appends it directly to the ETag string (`{ETag}-{size}`). If the size doesn't exist, it elegantly falls back to the original `ETag` string structure.
* **Updated `tests/test_data_files.py`**:
  Implemented `test_get_single_origin_metadata_includes_size` using a mocked `url_to_fs` patch to guarantee that the `size` byte integer correctly concatenates to the final string tuple emitted by `_get_single_origin_metadata()`.

### Testing
Tested unit test logic behavior locally by verifying that the mock tuple returns `("my_etag_123-1024",)`.

### DIFF
--- a/tests/test_data_files.py
+++ b/tests/test_data_files.py
@@ -2,7 +2,7 @@ import copy
 import os
 from pathlib import Path
 from typing import List
-from unittest.mock import patch
+from unittest.mock import MagicMock, patch
 
 import fsspec
 import pytest
@@ -15,6 +15,7 @@ from datasets.data_files import (
     DataFilesPatternsDict,
     DataFilesPatternsList,
     _get_data_files_patterns,
+    _get_single_origin_metadata,
     _is_inside_unrequested_special_dir,
     _is_unrequested_hidden_file_or_is_inside_unrequested_hidden_dir,
     get_data_patterns,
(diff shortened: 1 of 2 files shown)


tests/test_data_files.py
```diff
@@ -690,3 +690,17 @@ def test_get_data_patterns_from_directory_with_the_word_data_twice(tmp_path):
     data_file.touch()
     data_file_patterns = get_data_patterns(repo_dir.as_posix())
     assert data_file_patterns == {"train": ["data/train-[0-9][0-9][0-9][0-9][0-9]-of-[0-9][0-9][0-9][0-9][0-9]*.*"]}
+
```
Sanjays2402, line 694: these imports land after roughly 690 lines of test code, so the repo’s selected E402 check will fail. `patch` is already imported at the top; `MagicMock` and `_get_single_origin_metadata` need to move into the existing import blocks.

yuxin00j: Fixed.

lhoestq:
Great to have extra robustness, though I'm curious: in what situation would a file change and the ETag stay the same ?

yuxin00j:
> Great to have extra robustness, though I'm curious: in what situation would a file change and the ETag stay the same ?

Hi @lhoestq , this happens with appendable objects in GCS Rapid buckets. When data is appended to an existing object in a Rapid bucket, the object's size increases, but its etag and object generation remain unchanged. Including the file size in the metadata tuple ensures that datasets invalidates the cache when data is appended to these files.

PR #8339, opened 2026-07-16
Make the dataset fingerprint independent of Arrow chunking

### LINKED ISSUE #8327: `Dataset.from_pandas` allocates ~1 MB per Arrow chunk, Out of Memory errors
### Describe the bug

`Dataset.from_pandas` memory usage scales with the chunk count of the backing Arrow array (around 1 MB per chunk). 

`Dataset.shuffle()` followed by `to_pandas()` produces string columns with one chunk per row, so a 25k-row DataFrame holding ~32 MB of text allocates ~30 GB on the way back through `from_pandas` and the process gets killed silently. Encountered in Jupyter.

`pyarrow.Table.from_pandas` on the identical DataFrame completes in ~140 MB, so the amplification is likely in Dataset.

The following allocates ~30GB.
Open Activity Monitor (or equivalent), observe memory pressure when running the following code:

```python
from datasets import Dataset

texts = ['lorem ipsum dolor sit amet consectetur ' * 34 for _ in range(25_000)]  # ~1.3 KB each
ds = Dataset.from_dict({'text': texts, 'label': [0] * 25_000})

df = ds.shuffle(seed=42).to_pandas()
df['text'].array._pa_array.num_chunks   # 25k, chunk per row

Dataset.from_pandas(df)
```

Measurements:

| scenario, 25k rows × ~1.3 KB | chunks | result |
|---|---|---|
| `shuffle()` → `to_pandas()` → `from_pandas()` | 25,000 | OOM |
| same, 6,000 rows | 6,000 | OOM (>6.5 GB) |
| no shuffle → `from_pandas()` | 25 | OK, 139 MB |
| `shuffle()` → `flatten_indices()` → `from_pandas()` | 25 | OK, 184 MB |
| shuffled df → `pa.Table.from_pandas(df)` | 25,000 | OK, 141 MB |
| `Dataset.from_dict({'text': df['text'].tolist(), ...})` | — | OK, 468 MB |

### Expected behavior

Memory proportional to data size, independent of chunk layout (ex. `pyarrow.Table.from_pandas`).

Possible workaround is to `flatten_indices()` before `to_pandas()`, or `from_dict` with `.tolist()` columns.

### Environment info

- `datasets` 5.0.0, `pandas` 3.0.4, `pyarrow` 24.0.0, Python 3.12.13
- macOS 14.6, Apple M3 Pro (arm64), 36 GB RAM

SuryanshSS1011:
Fixes #8327.

As discussed in the issue, the OOM in `Dataset.from_pandas` comes from `generate_fingerprint`, not from the Arrow conversion. Fingerprinting hashes `dataset.__dict__`, and the underlying `pa.Table` (and the `InMemoryTable` wrapping it) pickles chunk-by-chunk, so the serialized size and the hashing cost scale with the number of chunks rather than the amount of data. `shuffle().to_pandas()` produces one chunk per row, which is where it really blows up.

Following @lhoestq's suggestion to make the table hash chunk-count-independent (rather than calling `combine_chunks()` on the stored data), this PR does two things:

- `InMemoryTable` now serializes via its underlying `table` only. The `_batches`/`_offsets` index is derived and rebuilt in `__init__`, so there's no need to pickle it. This mirrors how `MemoryMappedTable` and `ConcatenationTable` already define `__getstate__`/`__setstate__`.
- `pa.Table` and `pa.ChunkedArray` get dill reducers that serialize a chunk-count-independent form. They combine one column at a time, so memory stays bounded and a table with the same data hashes the same regardless of how it's chunked.

The combine is transient and per-column, so it never materializes a full extra copy of the table.

### Result

On the repro from the issue (25k rows shuffled to one chunk per row), `Dataset.from_pandas` goes from a ~1.5 GB allocation to a few MB. I measured RSS on the scaled 2000-chunk case: the delta across `from_pandas` dropped from ~1500 MB to ~4 MB.

The fingerprint is now stable across different chunkings of the same data and still distinguishes different data. `pickle` round-trips of a dataset reconstruct correctly. `tests/test_fingerprint.py` and `tests/test_table.py` pass (the two failures I see, `test_hash_torch_compiled_module` and `test_move_script_doesnt_change_hash`, also fail on `main` in my environment and are unrelated).

I added `test_hash_arrow_table_is_independent_of_chunking` to cover the invariant.

One thing that I would flag is that this changes the fingerprint values for existing datasets, so it will invalidate caches once. This seems unavoidable for a fix that changes how the table is hashed.

### DIFF
--- a/src/datasets/utils/_dill.py
+++ b/src/datasets/utils/_dill.py
@@ -157,7 +157,7 @@ def _save_arrowTable(pickler, obj):
     # regardless of chunking. See
     # https://github.com/huggingface/datasets/issues/8327.
     def create_arrowTable(schema, columns):
-        return pa.table({field.name: column for field, column in zip(schema, columns)}).cast(schema)
+        return pa.Table.from_arrays(columns, schema=schema)
 
     log(pickler, f"Ta: {obj}")
     args = (obj.schema, [column.combine_chunks() for column in obj.columns])
--- a/src/datasets/utils/_dill.py
+++ b/src/datasets/utils/_dill.py
@@ -19,6 +19,7 @@ from io import BytesIO
 from types import CodeType, FunctionType
 
 import dill
+import pyarrow as pa
 from packaging import version
 
 from .. import config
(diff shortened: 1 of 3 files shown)


src/datasets/utils/_dill.py
```diff
@@ -146,6 +147,37 @@ def _save_set(pickler, obj):
...
+def _save_arrowTable(pickler, obj):
+    # pyarrow's default pickle serializes each chunk's buffers separately, so the
+    # pickled size (and therefore the fingerprint cost) scales with the number of
+    # chunks rather than the amount of data. Serialize a chunk-count-independent
+    # form instead: combine each column's chunks one at a time (bounded memory,
+    # never a full-table copy) so identical data produces identical bytes
+    # regardless of chunking. See
+    # https://github.com/huggingface/datasets/issues/8327.
+    def create_arrowTable(schema, columns):
+        return pa.table({field.name: column for field, column in zip(schema, columns)}).cast(schema)
```
Sanjays2402: Arrow allows duplicate field names, but this dict comprehension collapses them. `dumps()` will then fail in `.cast(schema)` because the reconstructed table has fewer columns; `pa.Table.from_arrays(columns, schema=schema)` preserves duplicates.

SuryanshSS1011: Thanks for the review and catching this! 

You're right that keying the reconstruction on `field.name` collapses duplicate columns, so I've switched to `pa.Table.from_arrays(columns, schema=schema)` which is positional and keeps the duplicates, as you suggested.

lhoestq (approved):
lgtm :)

PR #8345, opened 2026-07-19
Include archive path in WebDataset TAR read errors

### LINKED ISSUE #7280: Add filename in error message when ReadError or similar occur
Please update error messages to include relevant information for debugging when loading datasets with `load_dataset()` that may have a few corrupted files. 

Whenever downloading a full dataset, some files might be corrupted (either at the source or from downloading corruption). 

However the errors often only let me know it was a tar file if `tarfile.ReadError` appears on the traceback, and I imagine similarly for other file types. 

This makes it really hard to debug which file is corrupted, and when dealing with very large datasets, it shouldn't be necessary to force download everything again.

sankalpsthakur:
Fixes #7280

## What changed

WebDataset catches `tarfile.ReadError` around TAR processing and re-raises with `{tar_path!r}`. The original error remains on `__cause__`.

Covers failures while advancing the archive iterator and while reading a member payload (feature inference and example generation).

Origin is included in `tracked_str.__repr__` (often an `hf://` URL), so any loader that reprs a tracked path gets the remote location. Userinfo and query strings are stripped so presigned URLs and basic-auth credentials are not exposed in tracebacks.

## Before

```text
ReadError: empty file
```

## After

```text
ReadError: Failed to read TAR archive '/path/to/corrupted-shard-00042.tar' (origin=hf://datasets/org/name@main/data/corrupted-shard-00042.tar): empty file
```

## Tests

- corrupt first shard (feature inference)
- valid shard then corrupt shard (example generation)
- truncated member during `f.read()`
- original `ReadError` preserved as cause
- Windows-safe path matching
- `tracked_str.__repr__` includes hf:// origin
- userinfo/query stripped from origin in repr

## Validation

- pytest: `tests/test_track.py` and WebDataset ReadError tests passed
- `ruff check` / `ruff format --check`: passed

### AI/LLM disclosure

- AI coding tools (including Grok and/or Codex agent-assisted editing) were used to help draft or modify code and this PR description.

### DIFF
--- a/src/datasets/packaged_modules/webdataset/webdataset.py
+++ b/src/datasets/packaged_modules/webdataset/webdataset.py
@@ -1,6 +1,7 @@
 import io
 import json
 import re
+import tarfile
 from itertools import islice
 from typing import Any, Callable
 
@@ -28,6 +29,14 @@ class WebDataset(datasets.GeneratorBasedBuilder):
 
     @classmethod
     def _get_pipeline_from_tar(cls, tar_path, tar_iterator):
+        try:
+            yield from cls._get_pipeline_from_tar_without_error_context(tar_path, tar_iterator)
+        except tarfile.ReadError as error:
+            # Avoid tracked_str.__repr__, which can include a presigned origin URL in the traceback.
+            raise tarfile.ReadError(f"Failed to read TAR archive {str(tar_path)!r}: {error}") from error
+
+    @classmethod
+    def _get_pipeline_from_tar_without_error_context(cls, tar_path, tar_iterator):
         current_example = {}
         for filename, f in tar_iterator:
             example_key, field_name = base_plus_ext(filename)
--- a/src/datasets/packaged_modules/webdataset/webdataset.py
+++ b/src/datasets/packaged_modules/webdataset/webdataset.py
@@ -4,6 +4,7 @@ import re
 import tarfile
 from itertools import islice
 from typing import Any, Callable
+from urllib.parse import urlsplit, urlunsplit
 
 import numpy as np
 import pyarrow as pa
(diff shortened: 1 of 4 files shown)


sankalpsthakur:
CI waiting on maintainer workflow approval (`action_required`). Local WebDataset ReadError-path tests are green.

sankalpsthakur:
related: this covers the WebDataset TAR ReadError path from #7280; broader `_iter_from_urlpath` remainder left for a follow-up.

lhoestq:
I would still include the origin, in most cases it's a hf:// URLs that points to the file online which is useful for debugging

sankalpsthakur:
Updated — errors now include both the archive path and the hf:// origin.

src/datasets/packaged_modules/webdataset/webdataset.py
```diff
@@ -28,6 +30,22 @@ class WebDataset(datasets.GeneratorBasedBuilder):
...
+            yield from cls._get_pipeline_from_tar_without_error_context(tar_path, tar_iterator)
+        except tarfile.ReadError as error:
+            # Include both the resolved local path and origin (e.g. hf://…) for debugging.
+            # Prefer str()/get_origin() over tracked_str.__repr__ so formatting stays explicit.
+            archive = str(tar_path)
+            if isinstance(tar_path, tracked_str):
+                origin = tar_path.get_origin()
+                if origin != archive:
+                    raise tarfile.ReadError(
+                        f"Failed to read TAR archive {archive!r} (origin={origin}): {error}"
```
Sanjays2402: Could this strip the query string and userinfo before putting `origin` in the exception? `_download_single` records the arbitrary input URL as the origin, so presigned URLs or basic-auth credentials will end up in logs and tracebacks here.

sankalpsthakur: Good call — origin in the error now drops userinfo/query so presigned URLs and basic-auth don't leak. Path/hf:// host kept for debugging.

lhoestq:
feel free to do it in `tracked_str.__repr__` directly insetad of having a custom logic just for webdataset, this way it will benefit other loaders as well

sankalpsthakur:
Good idea — I'll move this into `tracked_str.__repr__` in a follow-up so other loaders benefit too.

sankalpsthakur:
Went ahead and implemented this now rather than deferring to a follow-up (061468af0): the origin sanitization now lives in `tracked_str.__repr__` (via a small `_sanitize_origin` helper), so any loader that reprs a tracked path gets the safe origin — userinfo and query strings are stripped while scheme/host/path (e.g. `hf://…`) are kept for debugging. WebDataset's `except tarfile.ReadError` block is back to a one-liner and its local sanitizer is gone. The existing WebDataset read-error tests cover the behaviour unchanged.

sankalpsthakur:
Sanitization is in tracked_str.__repr__ now — ready for another look when you have a moment.

PR #8348, opened 2026-07-21
Optimize dataset loading by reusing metadata in `data_files.py`

### LINKED ISSUE #8197: Feature Request / Discussion: Skip origin_metadata fetch when streaming=True?

yuxin00j:
Fix #8197

This PR addresses the performance overhead during the data files pattern resolution phase by reusing the filesystem metadata that is already fetched by `fs.glob(..., detail=True)`. Previously, this metadata was discarded, requiring a subsequent and redundant `fs.info()` network call in `_get_single_origin_metadata` for every matched file to retrieve the ETag or mtime.
### Key Changes
*   **Introduced `_resolve_pattern`**: Added an internal function that mirrors the original `resolve_pattern` logic but returns a list of tuples `(filepath, info)` containing the pre-fetched metadata.
*   **Maintained Public API**: Re-implemented the public `resolve_pattern` as a backward-compatible wrapper around `_resolve_pattern` that strips out the metadata and only returns the file paths. This keeps the metadata handling private to the module, addressing previous review feedback.
*   **Optimized Metadata Retrieval**: Updated `_get_single_origin_metadata` to accept the `(data_file, info)` tuple. It now checks the pre-fetched `info` dict for `ETag`, `etag`, or `mtime`. If found, it returns the metadata immediately, bypassing the network call entirely.
*   **Safe Fallback Mechanism**: If the relevant metadata keys are absent from the `info` dict (or for specific remote filesystem edge cases), it safely falls back to the existing `fs.info()` behavior.
*   **Added Unit Tests**: Included explicit unit tests to verify that `fs.info()` is bypassed when metadata is successfully passed from the glob operation, and to ensure the fallback logic is triggered correctly when it is not.

### Metrics
We ran a streaming dataset benchmark (1,650 shards) across 8 processes against GCS across 5 iterations:

1. **Before PR (`main` branch)**
   - **Performance / Latency**: Execution times were highly variable due to network throttling, ranging between a min of **`3.75s`** and a max of **`10.11s`** across individual processes.
   - **Average Latency**: **`4.43s`** across all runs (per-run averages ranged from `3.86s` to `5.47s`).
   - **`fs.info()` Calls**: 1,651 per process (**13,208 total** across 8 processes).

2. **After PR (`optimize-metadata` branch)**
   - **Performance / Latency**: Execution times were extremely stable, ranging strictly between **`1.59s`** and **`2.00s`**.
   - **Average Latency**: **`1.73s`** across all runs.
   - **`fs.info()` Calls**: 1 per process (**8 total** across 8 processes).

### Impact 
By eliminating 1,650 redundant `fs.info()` network calls per process during initialization, we reduced total filesystem info calls from 13,208 down to just 8. 

This structural change reduces startup latency from **`4.43s`** to **`1.73s`** (**~2.56x faster**) and completely removes the 10+ second latency spikes caused by API throttling when loading datasets at scale.

### DIFF
--- a/src/datasets/data_files.py
+++ b/src/datasets/data_files.py
@@ -298,12 +298,12 @@ def _get_data_files_patterns(pattern_resolver: Callable[[str], list[str]]) -> di
     raise FileNotFoundError(f"Couldn't resolve pattern {pattern} with resolver {pattern_resolver}")
 
 
-def resolve_pattern(
+def _resolve_pattern(
     pattern: str,
     base_path: str,
     allowed_extensions: Optional[list[str]] = None,
     download_config: Optional[DownloadConfig] = None,
-) -> list[str]:
+) -> list[tuple[str, dict]]:
     """
     Resolve the paths and URLs of the data files from the pattern passed by the user.
 
@@ -381,16 +381,16 @@ def resolve_pattern(
         filepath = filepath if "://" in filepath else protocol_prefix + filepath
         if rest_hops:
             filepath = "::".join([filepath] + rest_hops)
-        matched_paths.append(filepath)
+        matched_paths.append((filepath, info))
     # ignore .ipynb and __pycache__, but keep /../
     if allowed_extensions is not None:
         out = [
-            filepath
-            for filepath in matched_paths
+            (filepath, info)
+            for filepath, info in matched_paths
             if any("." + suffix in allowed_extensions for suffix in xbasename(filepath).split(".")[1:])
         ]
         if len(out) < len(matched_paths):
-            invalid_matched_files = list(set(matched_paths) - set(out))
+            invalid_matched_files = list({filepath for filepath, _ in matched_paths} - {filepath for filepath, _ in out})
             logger.info(
                 f"Some files matched the pattern '{pattern}' but don't have valid data file extensions: {invalid_matched_files}"
             )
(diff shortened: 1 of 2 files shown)


src/datasets/data_files.py
```diff
@@ -378,30 +400,47 @@ def resolve_pattern(
...
     if allowed_extensions is not None:
-        out = [
-            filepath
-            for filepath in matched_paths
-            if any("." + suffix in allowed_extensions for suffix in xbasename(filepath).split(".")[1:])
-        ]
+        out = []
+        out_metadata = []
+        for i, filepath in enumerate(matched_paths):
+            if any("." + suffix in allowed_extensions for suffix in xbasename(filepath).split(".")[1:]):
```
zhixiangli: Why not perform extension filtering before the matched_metadata calculation?

src/datasets/data_files.py
```diff
@@ -298,12 +298,33 @@ def _get_data_files_patterns(pattern_resolver: Callable[[str], list[str]]) -> di
...
+) -> tuple[list[str], list[SingleOriginMetadata]]: ...
+
+
 def resolve_pattern(
     pattern: str,
     base_path: str,
     allowed_extensions: Optional[list[str]] = None,
     download_config: Optional[DownloadConfig] = None,
-) -> list[str]:
+    return_metadata: bool = False,
```
zhixiangli: Is it possible to avoid exposing the new field publicly?

src/datasets/data_files.py
```diff
@@ -378,30 +400,47 @@ def resolve_pattern(
             continue
         if _is_unrequested_hidden_file_or_is_inside_unrequested_hidden_dir(filepath, fs_pattern):
             continue
+
+        if return_metadata:
+            meta = ()
```
zhixiangli: Does ALL the fsspec filesystems return the etag/mtime through fs.glob()? For those who don't, fallback to parallel fs.info resolution is needed.

yuxin00j: Done.

src/datasets/data_files.py
yuxin00j: Done.

src/datasets/data_files.py
yuxin00j: Done.

src/datasets/data_files.py
```diff
@@ -298,12 +298,36 @@ def _get_data_files_patterns(pattern_resolver: Callable[[str], list[str]]) -> di
     raise FileNotFoundError(f"Couldn't resolve pattern {pattern} with resolver {pattern_resolver}")
 
 
-def resolve_pattern(
+def _get_single_origin_metadata(
+    data_file: str, download_config: Optional[DownloadConfig] = None
+) -> SingleOriginMetadata:
+    data_file, storage_options = _prepare_path_and_storage_options(data_file, download_config=download_config)
+    data_file = data_file.split("::")[0]
```
zhixiangli: This will remove all the `::`?

src/datasets/data_files.py
```diff
@@ -610,20 +677,21 @@ def from_patterns(
...
+                    allowed_extensions=allowed_extensions,
+                    download_config=download_config,
+                    with_metadata=True,
                 )
+                data_files.extend(resolved_paths)
+                origin_metadata.extend(resolved_metadata)
             except FileNotFoundError:
                 if not has_magic(pattern):
                     raise
-        origin_metadata = _get_origin_metadata(data_files, download_config=download_config)
```
zhixiangli, line 626: When having multi patterns, `origin_metadata` calculation was concurrent but now sequentially cross patterns.

src/datasets/data_files.py
```diff
@@ -298,12 +299,36 @@ def _get_data_files_patterns(pattern_resolver: Callable[[str], list[str]]) -> di
     raise FileNotFoundError(f"Couldn't resolve pattern {pattern} with resolver {pattern_resolver}")
 
 
-def resolve_pattern(
+def _get_single_origin_metadata(
+    data_file: str, download_config: Optional[DownloadConfig] = None
+) -> SingleOriginMetadata:
+    data_file, storage_options = _prepare_path_and_storage_options(data_file, download_config=download_config)
+    data_file = data_file.split("::")[-1]
+    fs, _ = url_to_fs(data_file, **storage_options)
```
zhixiangli: ```fs, fs_path = url_to_fs(source_path, **storage_options)```

Use fs_path for both resolve_path() and info(), keep the same behavior as before.

yuxin00j: Modified the pr logic to be consistent with existing code.

src/datasets/data_files.py
yuxin00j: Done.

martindurant:
Without analysing the specific code, I approve the gist of what it being done: file listings always inherently fetch all file information on the backend. This could be cached by filesystems themselves or, as here, explicitly passed around to be queried internally.

yuxin00j:
@lhoestq  Could you take a look at this PR please?

lhoestq:
I think the filesystem can cache them as the HF one does

zhixiangli:
> I think the filesystem can cache them as the HF one does

Thanks @lhoestq, I agree that caching this at the filesystem layer could be valuable longer term.
 
For this path, though, fs.glob(..., detail=True) has already fetched the per-file metadata. Today Datasets discards it and then immediately calls fs.info() for every matched file. This PR just carries that already-fetched metadata through the existing resolution flow, while keeping the current fs.info() fallback when a filesystem does not provide the needed fields.

The existing gcsfs listing cache is local to a filesystem instance, so it does not address the independent-worker case. A shared filesystem metadata cache would need a separate design for persistence, concurrent access, freshness, and invalidation across the relevant APIs. It is worth exploring separately, but I do not think it should block removing these redundant requests in Datasets now, where the improvement benefits all users immediately.

yuxin00j:
Hi @lhoestq, fsspec itself provides a dircache, and it misses this path in a few ways:

* fs.glob doesn't cache. For sharded patterns like data/train-[0-9]..., fsspec passes prefix="train-" to filter server-side. gcsfs intentionally skips caching prefixed listings because they're incomplete.
* info() does not cache either. ls()/find() write to dircache, while info() only reads it.
* It's per-process in memory. The cache doesn't cross distributed ranks.

Since the response from glob(detail=True) already returned this exact metadata, passing it through avoids throwing it away and re-fetching it. We still keep the fs.info() fallback if any fields are missing, so nothing regresses — and startup drops from 4.43s → 1.73s (13,208 → 8 calls across 8 ranks).

PR #8349, opened 2026-07-21
Fix require_storage_embed recursing into require_storage_cast

vineethsaivs:
### Summary

`require_storage_embed` decides whether a (possibly nested) feature needs its data embedded into storage. For nested features it dispatches to the wrong helper and can report the wrong answer.

### Root cause

The recursive branches call `require_storage_cast` instead of recursing into `require_storage_embed`:

```python
def require_storage_embed(feature: FeatureType) -> bool:
    if isinstance(feature, dict):
        return any(require_storage_cast(f) for f in feature.values())
    elif isinstance(feature, LargeList):
        return require_storage_cast(feature.feature)
    elif isinstance(feature, List):
        return require_storage_cast(feature.feature)
    else:
        return hasattr(feature, "embed_storage")
```

The two sibling predicates, `require_storage_cast` and `require_decoding`, both recurse into themselves; only `require_storage_embed` recurses into the wrong function. As a result, a nested feature whose inner type implements `cast_storage` but not `embed_storage` (for example `List(ClassLabel)`) is checked for `cast_storage` and wrongly reported as requiring embedding.

`ArrowWriter._build_writer` uses `require_storage_embed` per column to choose Parquet encoding, so such a column is treated as an embed column and written without compression / with `PLAIN` encoding and no dictionary, producing larger, less optimally encoded output. Types that do implement `embed_storage` (Image, Audio, Video, Pdf) also implement `cast_storage`, so the bug does not drop embedding where it is needed; it only over-reports.

### Fix

Recurse into `require_storage_embed`, mirroring the sibling predicates, so the leaf check is `embed_storage` rather than `cast_storage`.

### Verification

The existing `test_require_storage_embed_with_list_types` only parametrizes over `Audio` (which implements both `cast_storage` and `embed_storage`), so it passes with the bug present. Added `test_require_storage_embed_with_non_embeddable_list_types` covering `List`/`LargeList` of `ClassLabel` (cast_storage but not embed_storage), which returns `True` before the fix and `False` after. Reproduced directly: `require_storage_embed(List(ClassLabel(...)))` returns `True` on the current code and `False` after the change, while `List(Audio())` remains `True`.

### DIFF
--- a/src/datasets/features/features.py
+++ b/src/datasets/features/features.py
@@ -1813,11 +1813,11 @@ def require_storage_embed(feature: FeatureType) -> bool:
         :obj:`bool`
     """
     if isinstance(feature, dict):
-        return any(require_storage_cast(f) for f in feature.values())
+        return any(require_storage_embed(f) for f in feature.values())
     elif isinstance(feature, LargeList):
-        return require_storage_cast(feature.feature)
+        return require_storage_embed(feature.feature)
     elif isinstance(feature, List):
-        return require_storage_cast(feature.feature)
+        return require_storage_embed(feature.feature)
     else:
         return hasattr(feature, "embed_storage")
 
--- a/src/datasets/features/features.py
+++ b/src/datasets/features/features.py
@@ -1796,8 +1796,6 @@ def require_storage_cast(feature: FeatureType) -> bool:
     """
     if isinstance(feature, dict):
         return any(require_storage_cast(f) for f in feature.values())
-    elif isinstance(feature, (list, tuple)):
-        return require_storage_cast(feature[0])
     elif isinstance(feature, LargeList):
         return require_storage_cast(feature.feature)
     elif isinstance(feature, List):
--- a/src/datasets/features/features.py
+++ b/src/datasets/features/features.py
@@ -1796,6 +1796,8 @@ def require_storage_cast(feature: FeatureType) -> bool:
     """
     if isinstance(feature, dict):
         return any(require_storage_cast(f) for f in feature.values())
+    elif isinstance(feature, (list, tuple)):
+        return require_storage_cast(feature[0])
     elif isinstance(feature, LargeList):
         return require_storage_cast(feature.feature)
     elif isinstance(feature, List):
@@ -1814,6 +1816,8 @@ def require_storage_embed(feature: FeatureType) -> bool:
     """
     if isinstance(feature, dict):
         return any(require_storage_embed(f) for f in feature.values())
+    elif isinstance(feature, (list, tuple)):
+        return require_storage_embed(feature[0])
     elif isinstance(feature, LargeList):
         return require_storage_embed(feature.feature)
     elif isinstance(feature, List):
--- a/tests/features/test_features.py
+++ b/tests/features/test_features.py
@@ -1018,6 +1018,13 @@ def test_require_storage_embed_with_list_types(feature):
     assert require_storage_embed(feature)
 
 
+@pytest.mark.parametrize("feature", [LargeList(ClassLabel(names=["a", "b"])), List(ClassLabel(names=["a", "b"]))])
+def test_require_storage_embed_with_non_embeddable_list_types(feature):
+    # ClassLabel implements cast_storage but not embed_storage, so a nested
+    # ClassLabel does not require storage embedding
+    assert require_storage_embed(feature) is False
+
+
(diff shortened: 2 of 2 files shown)


src/datasets/features/features.py
```diff
@@ -1813,11 +1813,11 @@ def require_storage_embed(feature: FeatureType) -> bool:
         :obj:`bool`
     """
     if isinstance(feature, dict):
-        return any(require_storage_cast(f) for f in feature.values())
+        return any(require_storage_embed(f) for f in feature.values())
```
Sanjays2402, line 1816: `FeatureType` still includes list/tuple schemas such as `[Audio()]`, and `get_nested_type` explicitly accepts them, but this helper falls through to `hasattr(list, "embed_storage")` and returns false. That leaves embeddable media in a supported nested representation misclassified; should this recurse through `feature[0]` like `require_decoding` does?

vineethsaivs:
Good catch, thanks. You are right that a bare `[Audio()]` list/tuple is a valid FeatureType (get_nested_type accepts it) and that require_decoding recurses through `feature[0]` while these two helpers did not.

In practice the normal path does not hit it: Features normalizes any raw list/tuple schema (including nested inside a dict) to a List object before these helpers ever see it, and every caller iterates `features.items()`, so the recursion always receives normalized feature objects. But the asymmetry is a real latent gap in the helpers' stated contract, and it existed identically in require_storage_cast, so I added the matching `(list, tuple)` branch to both for parity with require_decoding, plus a test on the raw schema. Now the three helpers agree on the shapes get_nested_type accepts.

src/datasets/features/features.py
```diff
@@ -1796,6 +1796,8 @@ def require_storage_cast(feature: FeatureType) -> bool:
     """
     if isinstance(feature, dict):
         return any(require_storage_cast(f) for f in feature.values())
+    elif isinstance(feature, (list, tuple)):
+        return require_storage_cast(feature[0])
```
lhoestq:
```suggestion
        return any(require_storage_cast(f) for f in feature.values())
```

lhoestq: No need to have it, it's a legacy structure that is not used anymore. We can remove list/tuple from FeatureType entirely now

lhoestq (approved):
lgtm :)

vineethsaivs:
@lhoestq you were right to pull `test_require_storage_cast_and_embed_recurse_into_raw_list_schema` in #8357, and I owe an explanation rather than leaving it as an unexplained removal.

That test could never have passed. It asserted a symmetry between the two helpers that does not exist: this PR added the `(list, tuple)` branch to `require_storage_embed` only, and `require_storage_cast` has never had one, so `require_storage_cast([Audio()])` falls through to `hasattr(list_instance, "cast_storage")`. Running the shipped functions at both revisions:

```
                        require_storage_cast   require_storage_embed
at 959825da  [Audio()]          False                  True
main today   [Audio()]          False                  True
```

So the `assert require_storage_cast(feature)` half was false the moment I wrote it. It reached `main` because no workflow ran on the fork branch (`gh pr checks` on this PR still reports "no checks reported on the branch"), and I did not run the file locally before pushing. That is on me, and it is the second half of the review I should have done rather than the half I did.

On @Sanjays2402's question and your answer, your read holds up when measured. `Features({"a": [Audio()]})` normalises the raw list at construction:

```
>>> Features({"a": [Audio()]})
{'a': List(Audio(sampling_rate=None, decode=True, num_channels=None, stream_index=None))}
```

so nothing arriving at these helpers from a `Features` object is ever a bare `list` or `tuple`. `get_nested_type` still accepts one directly, which is what made it look reachable to me, but that is a lower-level entry point. Which means the `(list, tuple)` branch this PR added to `require_storage_embed` is dead in the path that matters and is now the only place either helper mentions the legacy shape.

Happy to send the removal of those two lines as a small follow-up so the two helpers line up again, or to leave it if you would rather it come out as part of dropping list/tuple from `FeatureType` wholesale. Your call, I will not open it unasked.

The actual fix here is unaffected either way, and `test_require_storage_embed_with_non_embeddable_list_types` still pins it:

```
                                    before #8349   main today
require_storage_embed(LargeList(ClassLabel))   True        False
require_storage_embed(List(ClassLabel))        True        False
require_storage_embed(List(Audio))             True        True
```

PR #8352, opened 2026-07-23
Keep flat numeric columns with nulls numeric in numpy format

### LINKED ISSUE #8351: with_format('numpy') promotes flat float64/int columns with nulls to dtype=object (float32 stays numeric)
### Describe the bug

`NumpyArrowExtractor._arrow_array_to_numpy` (used by `Dataset.with_format("numpy")`) promotes a flat, homogeneous numeric column to `dtype=object` when the column contains nulls, but only for `float64`/`int64` columns. A `float32` column is returned as a numeric array. The result is a silent wrong-result inconsistency: identical data returns a numeric array or an object array depending on the column's numeric dtype.

The object array holds `np.float64(nan)` (not `None`), so nothing is preserved by the object dtype; downstream math and vectorized ops just break.

Root cause: `#3195` added an unguarded `or (isinstance(x, float) and np.isnan(x))` clause to the `any(...)` that decides `dtype=object` in `src/datasets/formatting/formatting.py`. It was meant to catch a null row in an *array* column, but it fires on any scalar float nan. Because `np.float64` is a subclass of Python `float` while `np.float32` is not, only `float64` (and `int64`, which pyarrow promotes to floating when nulls are present) trip it.

```python
import pyarrow as pa
from datasets.formatting.formatting import NumpyArrowExtractor

ext = NumpyArrowExtractor()
for t in (pa.float64(), pa.float32(), pa.int64()):
    tbl = pa.table({"c": pa.array([1.0, None, 3.0], type=t)})
    print(t, "->", ext.extract_column(tbl).dtype)
# float64 -> object   (wrong)
# float32 -> float32  (correct)
# int64   -> object   (wrong)
```

End to end:

```python
from datasets import Dataset
ds = Dataset.from_dict({"c": [1.0, None, 3.0]})   # Value('float64')
print(ds.with_format("numpy")[:]["c"].dtype)      # object, expected a numeric dtype
```

### Expected behavior

A flat homogeneous numeric column with nulls returns a numeric array (nulls as `nan`), consistently across `float32`/`float64`/`int`. `dtype=object` stays reserved for per-row array columns (`ArrayXD`/list columns) that cannot be stacked into a homogeneous numeric array.

### Environment info

- `datasets` version: `main` (5.0.1.dev0)
- Platform: Linux, Python 3.11
- `pyarrow` version: 25.0.0
- `numpy` version: 2.4.6

ebarkhordar:
Fixes #8351

## Root cause

`NumpyArrowExtractor._arrow_array_to_numpy` decides `dtype=object` with:

```python
if any(
    (isinstance(x, np.ndarray) and (x.dtype == object or x.shape != array[0].shape))
    or (isinstance(x, float) and np.isnan(x))
    for x in array
):
```

The second clause was added in #3195 to handle a null row in an *array* column, but it fires on any scalar float nan. For a flat homogeneous numeric column with nulls, `array` is a list of numpy scalars and the null surfaces as `np.float64(nan)`. Since `np.float64` is a subclass of Python `float` but `np.float32` is not, only `float64` (and `int`, which pyarrow promotes to floating when nulls are present) trips the clause, so identical data returns object or numeric depending on the column dtype.

## Fix

Guard the whole promotion on the presence of an ndarray element:

```python
if any(isinstance(x, np.ndarray) for x in array) and any(
    ... same conditions ...
):
```

Invariant: `dtype=object` is only returned when the column actually contains per-row arrays. A column made entirely of scalars (a flat numeric column) can always be stacked into a homogeneous numeric array, so it stays numeric.

This preserves the behavior of every array-column path exactly. When an ndarray is present the guard is `True`, so the inner check is evaluated unchanged; the only behavior that changes is the flat-scalar-with-nan case, which is the bug.

## Verification

Run in a clean `python:3.11-slim` container against this branch (`pyarrow` 25.0.0, `numpy` 2.4.6). Provenance checked via `datasets.formatting.formatting.__file__`.

- `extract_column` on `[1.0, None, 3.0]`: `float64 -> float64`, `float32 -> float32`, `int64 -> float64` (nulls as `nan`). On `main` `float64`/`int64` return `object`.
- Per-row array column with a null row (`list<float64>` `[[1,2],[3,4,5],None]`) still returns `dtype=object`, unchanged.
- Existing `ArrowExtractorTest` suite passes (including `test_numpy_extractor_nested`, which pins the object dtype for nested-array columns).

Two regression tests added:

- `test_numpy_extractor_flat_numeric_with_nulls`: flat float64/float32/int columns with a null stay numeric. Fails on `main` (`dtype('O') != float64`), passes here.
- `test_numpy_extractor_nested_with_nulls`: an array column with a null row still yields `dtype=object` (guards against a regression of the case #3195 added the clause for).

I did not touch a separate, pre-existing issue where a same-shape `list<float32>` column with a null row raises on stacking; that is unrelated to this dtype-consistency fix.

### DIFF
--- a/src/datasets/formatting/formatting.py
+++ b/src/datasets/formatting/formatting.py
@@ -201,7 +201,17 @@ class NumpyArrowExtractor(BaseArrowExtractor[dict, np.ndarray, dict]):
             ):
                 if np.lib.NumpyVersion(np.__version__) >= "2.0.0b1":
                     return np.asarray(array, dtype=object)
-                return np.array(array, copy=False, dtype=object)
+            for first_subarray in array:
+                if isinstance(first_subarray, np.ndarray):
+                    if any(
+                        (isinstance(x, np.ndarray) and (x.dtype == object or x.shape != first_subarray.shape))
+                        or ((isinstance(x, float) and np.isnan(x) or x is None) and first_subarray.ndim > 0)
+                        for x in array
+                    ):
+                        if np.lib.NumpyVersion(np.__version__) >= "2.0.0b1":
+                            return np.asarray(array, dtype=object)
+                        return np.array(array, copy=False, dtype=object)
+                    break
         if np.lib.NumpyVersion(np.__version__) >= "2.0.0b1":
             return np.asarray(array)
         else:
--- a/src/datasets/formatting/formatting.py
+++ b/src/datasets/formatting/formatting.py
@@ -188,7 +188,13 @@ class NumpyArrowExtractor(BaseArrowExtractor[dict, np.ndarray, dict]):
                 array: list = pa_array.to_numpy(zero_copy_only=zero_copy_only).tolist()
 
         if len(array) > 0:
-            if any(
+            # Only promote to dtype=object when the column is made of per-row arrays
+            # (ArrayXD/list columns) that cannot be stacked into a homogeneous numeric
+            # array: ragged shapes, an already-object element, or a null row arriving as
+            # a scalar nan among the ndarrays. A flat homogeneous numeric column whose
+            # nulls surface as scalar nan must stay numeric, so guard the whole check on
+            # the presence of an ndarray element first.
+            if any(isinstance(x, np.ndarray) for x in array) and any(
                 (isinstance(x, np.ndarray) and (x.dtype == object or x.shape != array[0].shape))
                 or (isinstance(x, float) and np.isnan(x))
                 for x in array
--- a/tests/test_formatting.py
+++ b/tests/test_formatting.py
@@ -98,6 +98,36 @@ class ArrowExtractorTest(TestCase):
         self.assertEqual(batch["c"][0].dtype, object)
         self.assertEqual(batch["c"].dtype, object)
 
+    def test_numpy_extractor_flat_numeric_with_nulls(self):
+        # A flat homogeneous numeric column with nulls must stay numeric (nulls -> nan),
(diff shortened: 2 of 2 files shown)


ebarkhordar:
The two failing legs (deps-minimum and py314) both fail on one test, tests/features/test_features.py::test_require_storage_cast_and_embed_recurse_into_raw_list_schema[feature0/feature1], which fails identically on current main (d21c5816). This PR only changes src/datasets/formatting/formatting.py and its test, so that failure is not from this branch. The remaining red entries are matrix jobs the runner cancelled after that one failed. Happy to rebase once main is green.

src/datasets/formatting/formatting.py
lhoestq: minor fix in the case of 
```python
ds = Dataset.from_dict({"i": [None, [1], [2]]}).with_format("np")
```

```suggestion
            for first_subarray in array:
                if isinstance(first_subarray, np.ndarray):
                    if any(
                        (isinstance(x, np.ndarray) and (x.dtype == object or x.shape != first_subarray.shape))
                        or ((isinstance(x, float) and np.isnan(x) or x is None) and first_subarray.ndim > 0)
                        for x in array
                    ):
                        if np.lib.NumpyVersion(np.__version__) >= "2.0.0b1":
                            return np.asarray(array, dtype=object)
                        return np.array(array, copy=False, dtype=object)
                    break
```

lhoestq (approved):
lgtm ! just added a small fix

PR #8353, opened 2026-07-23
Add file_name column option in JSON processing

### LINKED ISSUE #5806: Return the name of the currently loaded file in the load_dataset function.

Par-star:
Fixes #5806

  Adds an optional `return_file_name` parameter to the JSON loader (load_dataset).
  When set to True, a `file_name` column is added with the source file path,
  useful for resuming training from checkpoints by skipping already-used shards.

### DIFF
--- a/src/datasets/packaged_modules/json/json.py
+++ b/src/datasets/packaged_modules/json/json.py
@@ -75,7 +75,13 @@ class Json(datasets.ArrowBasedBuilder):
             )
         if self.config.newlines_in_values is not None:
             raise ValueError("The JSON loader parameter `newlines_in_values` is no longer supported")
-        return datasets.DatasetInfo(features=self.config.features)
+        features = self.config.features
+        # If the caller passed an explicit schema, make sure it accounts for the
+        # extra `file_name` column we append when return_file_name=True, otherwise
+        # `_cast_table` rejects the column since it isn't part of the schema.
+        if self.config.return_file_name and features is not None and "file_name" not in features:
+            features = datasets.Features({**features, "file_name": Value("string")})
+        return datasets.DatasetInfo(features=features)
 
     def _split_generators(self, dl_manager):
         """We handle string, list and dicts in datafiles"""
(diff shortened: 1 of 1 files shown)


src/datasets/packaged_modules/json/json.py
```diff
@@ -139,6 +140,14 @@ def _cast_table(self, pa_table: pa.Table, json_field_paths=()) -> pa.Table:
     def _generate_shards(self, base_files, files_iterables, original_files):
         yield from base_files
 
+    def _add_file_name_column(self, pa_table: pa.Table, file) -> pa.Table:
+        """Append a 'file_name' column with the source file path, when return_file_name=True."""
+        if self.config.return_file_name:
+            pa_table = pa_table.append_column(
```
Sanjays2402, line 152: `return_file_name=True` breaks when the caller also passes `features`: `self.info.features` remains that caller-provided schema, so `_cast_table` rejects this newly appended column because the names don’t match. The effective feature schema needs to include `file_name` too.

Par-star: Thanks for the review! You're right sir.

Pushed the change.

lhoestq:
Note users can already use `state_dict = ds.state_dict()` / `ds.load_state_dict(state_dict)` to checkpoint and resume training, which not only resumes at the right shard, but also at the right sample and taking into account any map()/filter()/shard()/split_dataset_by_node()/etc. operation and which works with any file format.

So I would rather encourage the use of `state_dict = ds.state_dict()` / `ds.load_state_dict(state_dict)` than adding a feature that are redundant and specific to json, wdyt ?

Par-star:
That makes sense, thanks for pointing me to `state_dict()`/`load_state_dict()` — 
I wasn't aware it already handled shard/sample-level resumption across 
different file formats and operations. That's clearly a more robust solution 
than what I proposed.

I'll close this PR. Thanks for the guidance!

PR #8354, opened 2026-07-23
Fix bucket dataset card handling and push metadata accounting

pjh4993:
Four fixes on the `buckets/...` read/write paths, hit by `push_to_hub` → `load_dataset`. The loader one crashes on any dataset with a README; the writer ones corrupt uploaded metadata.

## Key Changes

- **Bucket card loading** (`src/datasets/load.py`)**:** passed the card object instead of its `.data`, crashing on load. Now uses `.data`.
- **Standalone-YAML branch** (`src/datasets/load.py`)**:** the missing-file case wiped the parsed card. Now leaves it intact.
- **Split size** (`src/datasets/iterable_dataset.py`)**:** worker byte count was dropped, so `SplitInfo.num_bytes` was always 0. Now accumulated.
- **`dataset_infos.json`** (`src/datasets/arrow_dataset.py`)**:** serialized the wrong variable, always crashing (`UnboundLocalError` with no README, `TypeError` with one). Now returns the assembled dict for the caller to encode.

## Tests

One hermetic test per fix in `tests/test_buckets.py` (in-memory filesystems / a stubbed shard worker, no Hub), each verified to fail on the pre-fix code — same order as above:

- `test_bucket_module_uses_dataset_card_data_not_card` — guards `src/datasets/load.py`
- `test_bucket_module_preserves_card_when_standalone_yaml_missing` — guards `src/datasets/load.py`
- `test_push_parquet_shards_reports_dataset_nbytes` — guards `src/datasets/iterable_dataset.py`
- `test_get_updated_dataset_card_returns_legacy_infos_as_dict` — guards `src/datasets/arrow_dataset.py`

`pytest tests/test_buckets.py` → `4 passed`; `ruff check`/`ruff format --check` clean.


lhoestq (approved):
good catch !

PR #8147, opened 2026-04-23
Fix iterable map resume state

Brianzhengca:
Fixes: https://github.com/huggingface/datasets/issues/8145
## Root Cause
  `FormattedExamplesIterable.__iter__` ignored `force_convert_to_python` and used Arrow batch iteration, advancing the
  underlying state by a full Arrow chunk before yielding Python examples.

  ## Fix
  Only use Arrow batch iteration when `force_convert_to_python` is false:
  `if self.ex_iterable.iter_arrow and not self.force_convert_to_python:`

  ## Tests
  - [x] Verified the issue reproduction now resumes at `[200, 201, 202, 203, 204]`.

PR #8360, opened 2026-07-24
Rebatch arrow source before formatting in IterableDataset.filter to fix resume data loss

### LINKED ISSUE #8359: IterableDataset.filter() checkpoint/resume silently drops the rest of the in-flight Arrow table
### Describe the bug

Resuming a streaming `IterableDataset.filter()` from a `state_dict()` checkpoint silently drops rows: the remainder of the in-flight Arrow table is skipped. For a single-shard dataset this loses up to a full row group per checkpoint, and for a single-table shard it can drop every unconsumed row.

`filter()` wraps the Arrow source in `FormattedExamplesIterable` but, unlike `map()`, does not first rebatch it to one row. `FormattedExamplesIterable` takes the Arrow branch and pulls a whole table from the source, advancing the shard position to the end of that table while emitting examples one at a time. The default non-batched path never records a rewind checkpoint, so `state_dict()` saves the shard position at the table end and resume skips every row of that table that had not yet been emitted.

Batched `filter()`, `map()`, `skip`, `take`, `select_columns`, `rename_column` and formatting all resume without loss under the same harness; only non-batched `filter()` on an Arrow-backed source is affected. This is the standard `load_dataset(..., streaming=True).filter(...)` + `StatefulDataLoader` training-resume path.

```python
from datasets import Dataset

n = 10000
ds = Dataset.from_dict({"a": list(range(n))}).to_iterable_dataset(num_shards=1).filter(lambda e: True)

it = iter(ds)
seen = [next(it)["a"] for _ in range(1500)]  # consume 1500
state = ds.state_dict()

ds2 = Dataset.from_dict({"a": list(range(n))}).to_iterable_dataset(num_shards=1).filter(lambda e: True)
ds2.load_state_dict(state)
rest = [e["a"] for e in ds2]

print(len(seen) + len(rest))     # 9500, not 10000
print(sorted(set(range(n)) - set(seen + rest))[:5])  # [1500, 1501, 1502, 1503, 1504]
```

Consuming 1500 drops rows 1500..1999 (the rest of the second 1000-row table). Consuming 10 drops 990 rows.

### Expected behavior

`seen + rest` should equal every row exactly once, as it already does for batched `filter()` and for `map()`. No row should be lost across a checkpoint/resume.

### Environment info

- `datasets` version: 5.0.1.dev0 (main, commit 521a590)
- Python: 3.11.15
- pyarrow: 25.0.0
- Reproduced in a clean `python:3.11-slim` container.

ebarkhordar:
## Root cause

`IterableDataset.filter()` wraps the Arrow source in `FormattedExamplesIterable` but, unlike `map()`, does not first wrap it in `RebatchedArrowExamplesIterable(batch_size=1)`. `FormattedExamplesIterable.__iter__` takes the Arrow branch and pulls a whole table from `ArrowExamplesIterable`, advancing `shard_example_idx` to the end of that table while emitting one example at a time. The default non-batched filter never records a rewind checkpoint (`previous_state` is only set on the batched branch of `iter_outputs`), so on resume `ArrowExamplesIterable._iter_arrow` sees the saved position already at the table end and `continue`s past every unemitted row of that table.

Effect on `load_dataset(..., streaming=True).filter(...)` + `StatefulDataLoader` training-resume: each checkpoint drops up to a full row group. A single-table shard drops all unconsumed rows.

`map()` avoids this by rebatching the Arrow source to one row before formatting (#8147); `filter()` was left untouched.

## Fix

Mirror `map()`: when the filter source exposes `iter_arrow`, wrap it in `RebatchedArrowExamplesIterable(batch_size=1)` before `FormattedExamplesIterable`, so the formatter reads a single example ahead and `state_dict()` records the shard position at the emitted example. The `features` argument is now read from the original `self._ex_iterable` since `ex_iterable` may have been reassigned to the rebatched wrapper.

## Verification

Reproduced and fixed in a clean `python:3.11-slim` container at HEAD `521a590` (datasets 5.0.1.dev0, pyarrow 25).

- Differential on the new test `test_iterable_dataset_filter_resume_state_dict[batched, consume]`: on `main` the `batched=False` cases fail (resume loses rows) and the `batched=True` cases pass; on this branch all pass. Batched filter is the control, it already resumed correctly.
- The test consumes N examples, checkpoints, resumes, and asserts `seen + rest == list(range(n))` (every row exactly once) for `consume in {1, 500, 1500}`.
- `tests/test_iterable_dataset.py`: 441 passed, 47 skipped (the 4 deselected `test_interleave_dataset_with_sharding` cases require `torch`, which is not installed; they fail on `import torch` independently of this change).
- `ruff check` and `ruff format --check` (pinned `v0.11.8`) clean on both changed files.

I did not exercise the multi-worker `StatefulDataLoader` sharding path with `torch` installed; the fix is verified through `state_dict()`/`load_state_dict()` directly, which is the mechanism that path uses.

Resolves #8359

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -3702,10 +3702,17 @@ class IterableDataset(DatasetInfoMixin):
         # format and type before filtering
         ex_iterable = self._ex_iterable
         if self._info.features or self._formatting:
+            if ex_iterable.iter_arrow:
+                # Rebatch before formatting so the formatter reads at most one batch ahead
+                # instead of a whole arrow table. Without this, a non-batched filter consumes
+                # a table in full while emitting examples one at a time, so state_dict()
+                # records the shard position at the table end and resume skips every unemitted
+                # row of that table (mirrors `map`, see #8147).
+                ex_iterable = RebatchedArrowExamplesIterable(ex_iterable, batch_size=batch_size if batched else 1)
             ex_iterable = FormattedExamplesIterable(
                 ex_iterable,
                 formatting=self._formatting,
-                features=ex_iterable.features if ex_iterable.is_typed else self._info.features,
+                features=self._ex_iterable.features if self._ex_iterable.is_typed else self._info.features,
                 token_per_repo_id=self._token_per_repo_id,
             )
 
--- a/tests/test_iterable_dataset.py
+++ b/tests/test_iterable_dataset.py
@@ -2965,6 +2965,26 @@ def test_resume_dataloader_twice(num_workers):
     assert consumed + remainder == all_examples
 
 
+@pytest.mark.parametrize("consume", [1, 500, 1500])
+@pytest.mark.parametrize("batched", [False, True])
+def test_iterable_dataset_filter_resume_state_dict(consume, batched):
+    # Resuming an arrow-backed `filter()` must not skip rows: the formatter reads one
+    # example ahead instead of a whole arrow table, so state_dict() records the shard
+    # position at the emitted example and resume replays every unemitted row exactly once.
+    n = 2000
+    filter_fn = (lambda batch: [True] * len(batch["a"])) if batched else (lambda example: True)
+    ds = Dataset.from_dict({"a": list(range(n))}).to_iterable_dataset(num_shards=1).filter(filter_fn, batched=batched)
+    it = iter(ds)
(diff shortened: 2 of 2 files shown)


lhoestq (approved):
good catch, lgtm !

PR #8361, opened 2026-07-24
Use Sequence instead of list for path_or_paths typing

Par-star:
Fixes #5354

## What
Changes the type annotation of `path_or_paths` from `Union[PathLike, list[PathLike]]`
to `Union[PathLike, Sequence[PathLike]]` in:
- `Dataset.from_csv`
- `Dataset.from_json`
- `Dataset.from_parquet`
- `Dataset.from_text`

## Why
`list` is invariant in mypy, so passing a `List[Union[str, bytes, PathLike]]` (or any
other list subtype) to these functions raises a mypy error, even though it works
correctly at runtime:

    error: Argument 1 to "from_parquet" has incompatible type "List[str]";
    expected "Union[..., List[Union[str, bytes, PathLike[Any]]]]"  [arg-type]
    note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
    note: Consider using "Sequence" instead, which is covariant

Since these functions only read/iterate over `path_or_paths` and never mutate it,
`Sequence` is the more accurate and mypy-friendly annotation, matching the standard
`typing` recommendation for read-only sequence arguments.

## Backward compatibility
No runtime behavior change — `Sequence` still accepts lists, tuples, etc., so this is
purely a typing improvement and fully backward compatible.

### DIFF
--- a/src/datasets/packaged_modules/json/json.py
+++ b/src/datasets/packaged_modules/json/json.py
@@ -75,7 +75,13 @@ class Json(datasets.ArrowBasedBuilder):
             )
         if self.config.newlines_in_values is not None:
             raise ValueError("The JSON loader parameter `newlines_in_values` is no longer supported")
-        return datasets.DatasetInfo(features=self.config.features)
+        features = self.config.features
+        # If the caller passed an explicit schema, make sure it accounts for the
+        # extra `file_name` column we append when return_file_name=True, otherwise
+        # `_cast_table` rejects the column since it isn't part of the schema.
+        if self.config.return_file_name and features is not None and "file_name" not in features:
+            features = datasets.Features({**features, "file_name": Value("string")})
+        return datasets.DatasetInfo(features=features)
 
     def _split_generators(self, dl_manager):
         """We handle string, list and dicts in datafiles"""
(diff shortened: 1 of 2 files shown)


src/datasets/packaged_modules/json/json.py
```diff
@@ -139,6 +146,14 @@ def _cast_table(self, pa_table: pa.Table, json_field_paths=()) -> pa.Table:
     def _generate_shards(self, base_files, files_iterables, original_files):
         yield from base_files
 
+    def _add_file_name_column(self, pa_table: pa.Table, file) -> pa.Table:
+        """Append a 'file_name' column with the source file path, when return_file_name=True."""
+        if self.config.return_file_name:
+            pa_table = pa_table.append_column(
+                "file_name", pa.array([str(file)] * len(pa_table), type=pa.string())
```
Sanjays2402, line 153: For remote URLs, `file` here is the DownloadManager cache/extraction path, so `return_file_name=True` returns a machine-local filename instead of the source URL. The agent-trace branch below uses `original_files[shard_idx]` for this; should this helper get that value too?

Par-star: You're right — for remote files, `file` was the
DownloadManager's resolved local cache path, not the source path/URL.

I've updated `_add_file_name_column` to use `original_files[shard_idx]`
consistently across all call sites (the field-specific path, the pandas
fallback path, and the main JSON-lines path). For the agent-traces branch
I reused the `file_path` variable that was already being computed the
same way for `example["file_path"]`, so both stay consistent.

Pushed the fix — let me know if this looks right to you.

PR #8363, opened 2026-07-24
Keep integers on the python read path for fixed-shape ArrayXD columns with nulls

### LINKED ISSUE #8362: ArrayXD fixed-shape integer column with a null row is silently cast to float on the python read path
### Describe the bug

A fixed-shape `ArrayXD` column (for example `Array2D(shape=(1, 2), dtype="int64")`) that contains at least one `None` row is silently cast to float on the python read path. `to_dict()` / `to_pylist()` return every value as a float, and integers above 2**53 lose precision, purely because an unrelated row in the column is null. Single-row access `dataset[i][column]` returns the correct integers, so the output depends on whether another row happens to be null.

```python

feats = Features({"a": Array2D(shape=(1, 2), dtype="int64")})
ds = Dataset.from_dict({"a": [[[10, 20]], [[30, 40]], None]}, features=feats)

print(ds.to_dict()["a"])   # [[[10.0, 20.0]], [[30.0, 40.0]], [[nan, nan]]]
print(ds[0]["a"])          # [[10, 20]]   (correct ints via single-row access)

# precision loss above 2**53:
feats = Features({"a": Array2D(shape=(1, 1), dtype="int64")})
ds = Dataset.from_dict({"a": [[[9007199254740993]], None]}, features=feats)
print(ds.to_dict()["a"])   # [[[9007199254740992.0]], [[nan]]]   (2**53 + 1 altered)
```

### Expected behavior

The non-null rows keep their integer values (and full precision), and the null row is returned as `None`:

```
[[[10, 20]], [[30, 40]], None]
[[[9007199254740993]], None]
```

### Root cause

`ArrayExtensionArray.to_numpy` (fixed-first-dim branch) casts the whole column to `float64` so it can insert `np.nan` for the null rows, and `to_pylist` then returns `numpy_arr.tolist()`, so the float cast reaches the python read path. A dense numpy array cannot hold both integers and a missing marker, but the python path can (`None`), and the dynamic-first-dim branch (`shape=(None, ...)`) already avoids the blanket cast.

### Environment info

- `datasets` 5.0.1.dev0 (HEAD 521a590f), reproduced in a clean `python:3.11-slim` container, numpy 2.x, pyarrow 25.x.

ebarkhordar:
## What does this PR do?

Fixes #8362.

A fixed-shape `ArrayXD` column (`Array2D(shape=(1, 2), dtype="int64")`, and so on) that contains a `None` row is silently cast to float on the python read path: `to_dict()` / `to_pylist()` return every value as a float, and integers above 2**53 are altered, because an unrelated row is null. `dataset[i][column]` returns the correct integers, so the result depends on whether another row happens to be null.

## Root cause

`ArrayExtensionArray.to_numpy` casts a fixed-shape column to `float64` so it can insert `np.nan` for the null rows (a dense numpy array cannot hold both integers and a missing marker). `to_pylist` returns `numpy_arr.tolist()`, so that float cast reaches the python read path, where `None` is available and no cast is needed.

## Fix

In `to_pylist`, when the array is fixed-shape and has nulls, build the list from the nested-list `storage` (`self.storage.to_pylist()`), which already carries the exact values and `None` for the null rows.

The invariant: a non-null value read through the python path must never be altered because another row in the column is null.

Scope kept deliberately narrow:
- The numpy format path is unchanged. `to_numpy` still returns the dense `float64` + `np.nan` array (asserted by the existing `test_array_xd_with_none`), so `with_format("numpy")` behaves exactly as before.
- The dynamic-first-dim branch (`shape=(None, ...)`) is untouched.

## Verification

- New test `test_array_xd_with_none_python_format_keeps_integers`: fails on `main` (`[[nan, nan]] != None`), passes here. It covers integer preservation, the 2**53 + 1 precision case, and asserts the numpy format still returns a dense `float64` + `nan` array.
- `python -m pytest tests/features/test_array_xd.py`: 103 passed.
- `ruff check` and `ruff format --check` clean on the changed files.
- Verified in a clean `python:3.11-slim` container at HEAD 521a590f. Not verified: the torch/tf/jax formatting paths, which this change does not touch (it only affects the python `to_pylist` path).

### DIFF
--- a/src/datasets/features/features.py
+++ b/src/datasets/features/features.py
@@ -835,6 +835,13 @@ class ArrayExtensionArray(pa.ExtensionArray):
         numpy_arr = self.to_numpy(zero_copy_only=zero_copy_only)
         if self.type.shape[0] is None and numpy_arr.dtype == object:
             return [arr.tolist() for arr in numpy_arr.tolist()]
+        elif self.type.shape[0] is not None and self.storage.null_count:
+            # For a fixed-shape array, to_numpy casts the whole column to float64 to
+            # hold np.nan in the null rows. On the python read path that silently
+            # corrupts every non-null value: integers become floats and values above
+            # 2**53 lose precision. The nested-list storage already carries the exact
+            # values and None for the null rows, so build the list from it directly.
+            return self.storage.to_pylist()
         else:
             return numpy_arr.tolist()
 
--- a/tests/features/test_array_xd.py
+++ b/tests/features/test_array_xd.py
@@ -397,6 +397,28 @@ def test_array_xd_with_none():
     assert np.isnan(arr[1]) and np.isnan(arr[3])  # a single np.nan value - np.all not needed
 
 
+def test_array_xd_with_none_python_format_keeps_integers():
+    # Fixed-shape integer column with a null row: the python read path (to_dict /
+    # to_pylist) must keep the non-null values as integers and surface the null as
+    # None, not cast the whole column to float to make room for np.nan.
+    features = datasets.Features({"foo": datasets.Array2D(dtype="int64", shape=(1, 2))})
+    dataset = datasets.Dataset.from_dict({"foo": [[[10, 20]], [[30, 40]], None]}, features=features)
+    assert dataset.to_dict()["foo"] == [[[10, 20]], [[30, 40]], None]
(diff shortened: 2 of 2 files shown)


lhoestq (approved):
lgtm !

PR #8366, opened 2026-07-25
Preserve nullable integer columns in to_json/to_csv/to_sql

### LINKED ISSUE #8365: to_json/to_csv/to_sql cast nullable integer columns to float, corrupting values past 2**53
### Describe the bug

`Dataset.to_json`, `to_csv`, and `to_sql` write a nullable integer column as floating point rather than integers. All three writers (`io/json.py`, `io/csv.py`, `io/sql.py`) call `batch.to_pandas()` with pyarrow's default `integer_object_nulls=False`, which casts any integer column containing a null to `float64`. That drops the integer type on export, and for values beyond `2**53` it also corrupts the value, since those are not representable in `float64`. A column with no nulls is unaffected.

Verified on `main` (030a3e5) with datasets 5.0.1.dev0, pyarrow 25.0.0, pandas 3.0.5:

```python

ds = Dataset.from_dict({"a": [9007199254740993, None, 5]}, features=Features({"a": Value("int64")}))
ds.to_json("out.jsonl", lines=True)
print(open("out.jsonl").read())
```

Output:

```
{"a":9007199254740992.0}
{"a":null}
{"a":5.0}
```

The value is written as a float and is off by one (`9007199254740993` becomes `...992.0`). `to_csv` writes `9007199254740992.0` / `5.0`; `to_sql` stores the column as SQLite `REAL` (`typeof(a)` = `real`) with the same corruption. A `to_json` then `load_dataset("json")` round-trip silently turns `int64` into `float64`.

### Expected behavior

Integer values are written as integers and preserved exactly: `9007199254740993`, `null`, `5`.

### Environment info

- `datasets` main (030a3e5)
- `pyarrow` 25.0.0, `pandas` 3.0.5, Python 3.11

ebarkhordar:
### What this fixes

`Dataset.to_json`, `to_csv` and `to_sql` write a nullable integer column as floating point rather than integers.

Fixes #8365.

### Root cause

The three writers build a pandas frame with a bare `batch.to_pandas()`:

- `io/json.py` `_batch_json`
- `io/csv.py` `_batch_csv`
- `io/sql.py` `_batch_sql`

pyarrow's `to_pandas` defaults to `integer_object_nulls=False`, which casts any integer column that contains a null to `float64`. On export that drops the integer type, and for values beyond `2**53` it corrupts the value, since those are not representable in float64. A `to_json` then `load_dataset("json")` round-trip silently changes `int64` into `float64`.

### Fix

Pass `integer_object_nulls=True` to the three writers' `to_pandas()` calls. A null-containing integer column then materializes as Python integers (object dtype) and is written as integers; the null stays null. Columns with no nulls are untouched by the flag, so their numpy dtype is preserved.

Tradeoff: `integer_object_nulls=True` uses object dtype for integer columns that contain a null, which costs memory and speed for those columns only. An arrow-backed nullable dtype (pandas `Int64` via a `types_mapper`) would avoid object dtype but touches more of the export path. Glad to switch to that approach if you prefer it.

### Verification

Ran in a clean container against this branch (datasets 5.0.1.dev0, pyarrow 25.0.0, pandas 3.0.5):

- Added regression tests for all three writers, parametrized over `int64`/`uint64`/`int32`. They fail on `main` (the column exports as float, e.g. `9007199254740992.0`, and SQLite stores `real`) and pass here (exact integers, SQLite `integer`).
- `tests/io/test_json.py tests/io/test_csv.py tests/io/test_sql.py`: 96 passed, no regressions.
- `ruff check` and `ruff format --check` clean on the changed files.

I did not run the full non-io test suite or the Windows and minimum-deps CI matrix locally.

### DIFF
--- a/src/datasets/io/csv.py
+++ b/src/datasets/io/csv.py
@@ -105,7 +105,7 @@ class CsvDatasetWriter:
             key=slice(offset, offset + self.batch_size),
             indices=self.dataset._indices,
         )
-        csv_str = batch.to_pandas().to_csv(
+        csv_str = batch.to_pandas(integer_object_nulls=True).to_csv(
             path_or_buf=None, header=header if (offset == 0) else False, index=index, **to_csv_kwargs
         )
         return csv_str.encode(self.encoding)
--- a/src/datasets/io/json.py
+++ b/src/datasets/io/json.py
@@ -131,7 +131,7 @@ class JsonDatasetWriter:
             key=slice(offset, offset + self.batch_size),
             indices=self.dataset._indices,
         )
-        batch = batch.to_pandas()
+        batch = batch.to_pandas(integer_object_nulls=True)
         for json_field_path in get_json_field_paths_from_feature(self.dataset.features):
             col, *json_field_subpath = json_field_path
             batch[col] = batch[col].apply(partial(json_decode_field, json_field_path=json_field_subpath))
--- a/src/datasets/io/sql.py
+++ b/src/datasets/io/sql.py
@@ -87,7 +87,7 @@ class SqlDatasetWriter:
             key=slice(offset, offset + self.batch_size),
             indices=self.dataset._indices,
         )
-        df = batch.to_pandas()
+        df = batch.to_pandas(integer_object_nulls=True)
         num_rows = df.to_sql(self.name, self.con, index=index, **to_sql_kwargs)
         return num_rows or len(df)
 
--- a/tests/io/test_csv.py
+++ b/tests/io/test_csv.py
@@ -173,3 +173,19 @@ def test_dataset_to_csv_fsspec(dataset, mockfs):
 
     with fsspec.open(dataset_path, "rb", **mockfs.storage_options) as f:
         assert f.read()
+
+
+@pytest.mark.parametrize(
+    "dtype, big",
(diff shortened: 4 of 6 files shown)


ebarkhordar:
The red Windows legs come from the ffmpeg pin this branch inherited from its base, not from the change itself.

`.github/workflows/ci.yml` here still runs `conda install "ffmpeg=7.0.1" -c conda-forge`; #8369 (b305031c, yesterday) changed that to `ffmpeg=6.1.1`. With ffmpeg 7.0.1 and torchcodec 0.15.0 the Windows job cannot load `libtorchcodec_core7.dll` (WinError 127), so 78 of the 80 failures are audio and video decode tests. The other two are `tests/test_buckets.py` getting a 401 from the buckets API, which a fork PR has no token for. The push run on main started two minutes before this one on the same runner image, installed ffmpeg 6.1.1, and passed `test (unit, windows-latest, deps-latest)`.

The three tests this PR adds passed inside that same failing Windows job. Happy to merge main in if you want the Windows legs re-run, though the runs would need approval again.

lhoestq (approved):
it's all good, the failures are indeed unrelated to your change and they are fixed on main :)

ebarkhordar:
Thanks for the quick turnaround. Good to know the ffmpeg pin is already handled on main.

PR #8368, opened 2026-07-27
Reject duplicate column names in Dataset.select_columns

AbdullahRasheed45:
### The problem

`Dataset.select_columns` checks that every requested column exists, but never checks that the request is free of duplicates:

```python
missing_columns = set(column_names) - set(self._data.column_names)
if missing_columns:
    raise ValueError(...)
```

Passing the same name twice then produces a dataset that is internally inconsistent. PyArrow's `select` honours the repeat and returns two columns, while the features are rebuilt with a dict comprehension that collapses it back to one:

```python
dataset._data = dataset._data.select(column_names)                       # 2 columns
dataset._info.features = Features({col: ... for col in dataset._data.column_names})   # 1 feature
```

On current `main`:

```python
d = Dataset.from_dict({"x": [1, 2], "y": ["a", "b"]})
sel = d.select_columns(["x", "x"])

sel.column_names      # ['x', 'x']
sel.num_columns       # 2
len(sel.features)     # 1     <- disagrees with num_columns
sel["x"]              # KeyError: 'Field "x" exists 2 times in schema'
sel.map(lambda r: r)  # KeyError: 'Field "x" exists 2 times in schema'
sel.to_dict()         # {'x': [1, 2]}   <- one column silently dropped
```

So the call succeeds and hands back an object that raises on its own basic operations. The failure surfaces later, at an unrelated line, with a message that points at the Arrow schema rather than at the `select_columns` call that caused it.

### The change

Raise a `ValueError` when the selection contains duplicates.

This mirrors `rename_columns`, which already rejects the analogous case a few methods up:

> `New column names must all be different, but this column mapping has N duplicates.`

Selecting distinct columns — including reordering, a single string, or an empty list — is unaffected.

### A consistency question for you

`IterableDataset.select_columns(["x", "x"])` does **not** fail today: it silently de-duplicates and yields `{"x": ...}`. So the two classes disagree about what a duplicated selection means.

I chose to raise on `Dataset` rather than de-duplicate, because silently dropping part of what the caller asked for hides a mistake, and because `rename_columns` sets the precedent for rejecting duplicates outright. But that leaves the divergence in place, and aligning the two is a call for you rather than me — I'm happy to follow up either by making `IterableDataset` raise too, or by switching this to de-duplicate instead, whichever you prefer.

Adjacent, not touched here: `remove_columns(["x", "x"])` raises a bare `KeyError: 'x'`, which at least fails rather than corrupting, but is not a clear message either.

### Tests

Added `test_select_columns_duplicates` next to the existing `test_select_columns`, following the same `_create_dummy_dataset` pattern. It covers the plain duplicate and a duplicate mixed among distinct columns, and asserts that a reordered distinct selection still works with `num_columns == len(features)` — the invariant the bug broke.

It fails on `main` in both the in-memory and on-disk parametrizations, and passes with the fix.

`pytest tests/test_arrow_dataset.py -k "select_columns or remove_columns or rename_column or concatenate or flatten"` → **45 passed**. `ruff check` and `ruff format` clean.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -2716,6 +2716,13 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                 f"{self._data.column_names}."
             )
 
+        number_of_duplicates = len(column_names) - len(set(column_names))
+        if number_of_duplicates != 0:
+            raise ValueError(
+                "Selected column names must all be different, but this selection "
+                f"has {number_of_duplicates} duplicates."
+            )
+
         dataset = copy.deepcopy(self)
         dataset._data = dataset._data.select(column_names)
         dataset._info.features = Features({col: self._info.features[col] for col in dataset._data.column_names})
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -4181,6 +4181,13 @@ class IterableDataset(DatasetInfoMixin):
         if isinstance(column_names, str):
             column_names = [column_names]
 
+        number_of_duplicates = len(column_names) - len(set(column_names))
+        if number_of_duplicates != 0:
+            raise ValueError(
+                "Selected column names must all be different, but this selection "
+                f"has {number_of_duplicates} duplicates."
+            )
+
         if self._info:
             info = deepcopy(self._info)
             if self._info.features is not None:
--- a/tests/test_arrow_dataset.py
+++ b/tests/test_arrow_dataset.py
@@ -729,6 +729,23 @@ class BaseDatasetTest(TestCase):
                     self.assertNotEqual(new_dset._fingerprint, fingerprint)
                     assert_arrow_metadata_are_synced_with_dataset_features(new_dset)
 
+    def test_select_columns_duplicates(self, in_memory):
+        # Selecting the same column twice used to build a dataset whose Arrow table had
+        # the column twice while its features had it once. That object then raised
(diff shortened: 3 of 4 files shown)


ebarkhordar (commented):
The duplicate check is the right call, and it holds up on a run. On main, `select_columns(["col_1", "col_1"])` returns a dataset whose Arrow table carries the column twice while `features` has it once, and `d["col_1"]`, `select`, `sort`, `flatten_indices` and `save_to_disk` then all fail with `KeyError: 'Field "col_1" exists 2 times in schema'`. The guard removes that, and the new test actually detects it: deleting just the seven added lines in `arrow_dataset.py` takes `test_select_columns_duplicates` from 2 passed to 2 failed with "ValueError not raised". `tests/test_arrow_dataset.py -k select_columns` stays green (4 passed).

One scope question. `IterableDataset.select_columns` (`iterable_dataset.py:4156`) takes the same argument and is untouched here, so after this PR the two classes answer the same call differently. At this branch:

```
IterableDataset.select_columns(['col_1','col_1'])
  features : ['col_1']
  rows     : [{'col_1': 0}, {'col_1': 1}, {'col_1': 2}]

Dataset.select_columns(['col_1','col_1'])
  ValueError: Selected column names must all be different, but this selection has 1 duplicates.
```

The iterable path dedupes silently instead of corrupting anything, so it is not the same bug, but it does mean the validation is now class-dependent. `DatasetDict` and `IterableDatasetDict` each inherit whichever their element class does (`dataset_dict.py:507` and `:2261`), so the split follows through to them. Is keeping this to `Dataset` deliberate, or would the same check be welcome in the iterable path?

PR #8374, opened 2026-07-30
feat: Qdrant Search Index

### LINKED ISSUE #5698: Add Qdrant as another search index
### Feature request

I'd suggest adding Qdrant (https://qdrant.tech) as another search index available, so users can directly build an index from a dataset. Currently, FAISS and ElasticSearch are only supported: https://huggingface.co/docs/datasets/faiss_es 

### Motivation

ElasticSearch is a keyword-based search system, while FAISS is a vector search library. Vector database, such as Qdrant, is a different tool based on similarity (like FAISS) but is not limited to a single machine. It makes the vector database well-suited for bigger datasets and collaboration if several people want to access a particular dataset.

### Your contribution

I can provide a PR implementing that functionality on my own.

anush008:
Resolves #5698 

## Description

This PR adds Qdrant as a vector search backend alongside FAISS and Elasticsearch. Associated documentation and tests are also included. The Qdrant client dependency is marked optional and installed on demand.

[Qdrant](https://qdrant.tech/) is an open-source vector search engine built for high-performance. It offers extensive [filtering support](https://qdrant.tech/documentation/search/filtering/) with its [filterable HNSW](https://qdrant.tech/articles/filterable-hnsw/) implementation.

I've tested the integration end-to-end against a local Qdrant instance.

### Setup

You can run Qdrant with 

```bash
docker run -p 6333:6333 -p 6334:6334 qdrant/qdrant
```

The dashboard is accessible at http://localhost:6333/dashboard

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@
 | 🔄 **Smart caching** | Never wait for your data to process twice — cached results are automatically reused |
 | 📊 **Multi-framework interoperability** | Native conversion to/from NumPy, Pandas, Polars, Arrow, PyTorch, TensorFlow, JAX, and Spark |
 | 🏎️ **Multi-processing** | Fast parallel data processing with `map(num_proc=N)` |
-| 🔎 **Search & index** | Built-in FAISS and Elasticsearch index support for similarity search |
+| 🔎 **Search & index** | Built-in FAISS, Elasticsearch, and Qdrant index support for similarity search |
 | 📦 **JSON type** | Flexible JSON/structured data support with `Json()` feature type |
 
 # Installation
--- a/docs/source/_redirects.yml
+++ b/docs/source/_redirects.yml
@@ -6,7 +6,8 @@ dataset_streaming: stream
 torch_tensorflow: use_dataset
 splits: loading#slice-splits
 processing: process
-faiss_and_ea: faiss_es
+faiss_es: faiss_es_qdrant
+faiss_and_ea: faiss_es_qdrant
 features: about_dataset_features
 exploring: access
 package_reference/logging_methods: package_reference/utilities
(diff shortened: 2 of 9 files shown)


ebarkhordar (commented):
The Qdrant path works against a real client. I ran `add_qdrant_index`, `get_nearest_examples` and `search_batch` (with and without a `query_filter`) at f53e2763 in a clean container and the results were correct, including with `set_format("numpy")` and `set_format("torch")` active on the dataset.

One suggestion, on the test rather than the implementation. `QdrantIndexTest` drives a `Mock()` client, so no part of the upload or retrieval path actually executes. `qdrant-client` has a local mode that runs the real engine in-process with no server (`QdrantClient(":memory:")`), and `qdrant-client` is already in `TESTS_REQUIRE`, so using it costs nothing in CI.

The specific gap is the point-id mapping, which is what the index rests on: `load_qdrant_index` documents "point IDs equal to the zero-based dataset row numbers", and `search`/`search_batch` hand `int(point.id)` straight back as dataset indices. If `id=offset + i` in `point_generator` were to regress to `id=i` (batch-local index), every assertion in `QdrantIndexTest` still passes, because the ids only ever reach a `Mock` and the query results are hardcoded.

Against a local client the same regression shows up at once. Four rows, `batch_size=2`:

```
[A] id=offset + i    points_in_collection = 4    nearest_to_row3 = ['row3']
[B] id=i             points_in_collection = 2    nearest_to_row3 = ['row1']
```

Duplicate ids overwrite the earlier batch, so the collection silently holds half the rows and retrieval returns the wrong ones with no error.

Payload indexes are the one thing local mode does not support (it warns and no-ops), so the mock is still the right tool for the `create_payload_index` kwargs and ordering assertions and for the `query_filter` propagation check. The suggestion is to add a local-mode case next to it covering retrieval, not to replace what is already there.

Anush008:
The test case issue you mentioned is covered.

PR #8379, opened 2026-07-31
Preserve features after IterableDataset.add_column

### LINKED ISSUE #5752: Streaming dataset looses `.feature` method after `.add_column`

kiwigitops:
## Summary

- preserve known `IterableDataset` features when `add_column` appends a streaming column
- infer the appended column feature from the provided column values
- add a regression test for `add_column` after resolving iterable dataset features

Fixes #5752.

## Testing

- local repro with `IterableDataset.from_generator(...).add_column(...)`
- `python -m pytest tests/test_iterable_dataset.py::test_iterable_dataset_add_column tests/test_iterable_dataset.py::test_iterable_dataset_add_column_preserves_features`
- `python -m pytest tests/test_iterable_dataset.py::test_pickle_after_many_transforms`

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -4061,7 +4061,10 @@ class IterableDataset(DatasetInfoMixin):
         Returns:
             `IterableDataset`
         """
-        return self.map(partial(add_column_fn, name=name, column=column), with_indices=True)
+        features = self.features.copy() if self.features is not None else None
+        if features is not None:
+            features.update(_infer_features_from_batch({name: column}))
+        return self.map(partial(add_column_fn, name=name, column=column), with_indices=True, features=features)
 
     def rename_column(self, original_column_name: str, new_column_name: str) -> "IterableDataset":
         """
--- a/tests/test_iterable_dataset.py
+++ b/tests/test_iterable_dataset.py
@@ -2414,6 +2414,15 @@ def test_iterable_dataset_add_column(dataset_with_several_columns: IterableDatas
     assert "new_column" in new_dataset.column_names
 
 
+def test_iterable_dataset_add_column_preserves_features(dataset_with_several_columns: IterableDataset):
+    dataset = dataset_with_several_columns._resolve_features()
+    new_column = list(range(3 * DEFAULT_N_EXAMPLES))
+    new_dataset = dataset.add_column("new_column", new_column)
+    assert new_dataset.features is not None
+    assert new_dataset.features["new_column"] == Value("int64")
+    assert new_dataset.column_names == [*dataset.column_names, "new_column"]
+
+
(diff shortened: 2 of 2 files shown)


src/datasets/iterable_dataset.py
```diff
@@ -4061,7 +4061,10 @@ def add_column(self, name: str, column: Union[list, np.array]) -> "IterableDatas
         Returns:
             `IterableDataset`
         """
-        return self.map(partial(add_column_fn, name=name, column=column), with_indices=True)
+        features = self.features.copy() if self.features is not None else None
+        if features is not None:
+            features.update(_infer_features_from_batch({name: column}))
```
Sanjays2402, line 4066: inferring off `{name: column}` builds an arrow table from the whole column, which both copies the full column up front and narrows what `add_column` accepts: the signature allows `np.array`, and `pa.Table.from_pydict` raises `ArrowInvalid: only handle 1-dimensional arrays` for a 2D ndarray, so a call that used to just end up untyped now blows up before `map`. inferring from `column[:1]` (and falling back to `None` on ArrowInvalid) would keep that working. also worth mirroring `Dataset.add_column`'s `feature=` arg so the caller can skip inference entirely.

Sanjays2402, line 4066: the docstring types `column` as `Union[list, np.array]`, and a 2-D ndarray (adding an embedding column) goes straight into `pa.Table.from_pydict` here and raises `ArrowInvalid: only handle 1-dimensional arrays`. that used to work since features stayed None. worth wrapping the infer in a try and falling back to features=None, or converting to a list of rows first.

Sanjays2402, line 4066: `_infer_features_from_batch` goes through `pa.Table.from_pydict({name: column})`, which rejects a multi-dim ndarray (`ArrowInvalid: only handle 1-dimensional arrays`), and the signature advertises `np.array`. Before this, a 2-D column survived because inference only ever saw `column[idx]` (a 1-D row) at resolve time, so this turns a working call into a hard error. Worth wrapping the infer in a try and falling back to `features=None`?

Sanjays2402, line 4066: does this break `add_column` for multi-dim numpy columns? the signature says `Union[list, np.array]`, and `_infer_features_from_batch` goes straight to `pa.Table.from_pydict`, which raises ArrowInvalid("only handle 1-dimensional arrays") on a 2-D array. that call used to work since `add_column_fn` only indexes into the column. wrapping the inference in a try/except and falling back to `features=None` would keep the old behavior.

PR #8380, opened 2026-08-01
Speed up default Python row extraction

dexhunter:
## Problem

`PythonArrowExtractor.extract_row` runs once per row on the default read path (every `for row in ds` and every `ds[i]`), and it does more work than it needs to:

```python
def extract_row(self, pa_table: pa.Table) -> dict:
    return _unnest(pa_table.to_pydict())
```

`pa_table` here is always a one-row table. `to_pydict()` builds a one-element list for every column, and `_unnest` then throws those lists away, keeping `[0]` from each. A three-column row allocates three throwaway lists and two dicts.

Separately, `PythonFormatter.format_row` calls `self.python_arrow_extractor()`, which instantiates a new extractor object on every row. `PythonArrowExtractor` is stateless, so it can be built once.

## Fix

Convert only the row that was asked for, and build the stateless extractor once in `__init__`. Nine lines added, four removed.

## Results

Python 3.12.13, pyarrow 25.0.0, single core, at stock defaults. No option pinned away from its default.

| workload | before | after | speedup |
|---|---|---|---|
| `for row in ds` (1 column, 20k rows) | 142.5 ms | 124.5 ms | 1.14x |
| `for row in ds` (3 columns, 20k rows) | 309.4 ms | 236.9 ms | 1.31x |
| `for row in ds` (10 columns, 20k rows) | 809.5 ms | 567.2 ms | 1.43x |
| `for row in ds.select(...)` (5k rows) | 75.0 ms | 57.7 ms | 1.30x |
| `ds[i]` x 5000 (3 columns) | 119.1 ms | 96.2 ms | 1.24x |

The gain scales with column count, since the wasted allocation was per column.

## Tradeoffs

- `ds.map()` is unchanged, and not because the measurement was noisy. `map()` sets `format_kwargs["lazy"] = True`, so `format_row` returns a `LazyRow` and `extract_row` is never reached. I counted the calls: 0 during a `map()` over 2000 rows, 2000 during a full iteration of the same dataset. The before/after timings overlap (374.4-407.2 ms vs 376.7-415.9 ms).
- This only touches the Python format, which is the default one. The numpy, pandas and arrow formatters are untouched.
- Per-row cost here is microseconds. The claim is about iteration throughput, not end-to-end training time.

## Verification

- `tests/test_formatting.py`: 24 passed, 30 skipped, identical to base.
- `tests/test_arrow_dataset.py`: 362 passed, 61 skipped, 1 failed, identical to base. That one failure (`test_dataset_to_iterable_dataset`) reproduces on the unmodified base commit, so it is not from this change.
- `ruff check` and `ruff format --check` clean.
- Row output was compared against the unmodified implementation bit-for-bit, floats as raw IEEE-754 bytes, over freshly seeded datasets. The module's wider surface was probed directly too: `extract_batch`, `extract_column`, empty tables, `query_table` with int, slice, range, list and column-name keys both with and without an indices mapping, every built-in formatter, and the index validators.
- The change came out of an optimization run over this file: https://dashboard.weco.ai/share/ObqFOw-UhlOqE5ZIhbBwd3a1kjv0JHWE

### DIFF
--- a/src/datasets/formatting/formatting.py
+++ b/src/datasets/formatting/formatting.py
@@ -141,7 +141,9 @@ class SimpleArrowExtractor(BaseArrowExtractor[pa.Table, pa.Array, pa.Table]):
 
 class PythonArrowExtractor(BaseArrowExtractor[dict, list, dict]):
     def extract_row(self, pa_table: pa.Table) -> dict:
-        return _unnest(pa_table.to_pydict())
+        # Convert only the single row that is asked for. to_pydict() builds a
+        # one-element list per column and _unnest then throws those lists away.
+        return {name: col[0].as_py() for name, col in zip(pa_table.column_names, pa_table.columns)}
 
     def extract_column(self, pa_table: pa.Table) -> list:
         return pa_table.column(0).to_pylist()
@@ -468,23 +470,26 @@ class PythonFormatter(Formatter[Mapping, list, Mapping]):
     def __init__(self, features=None, lazy=False, token_per_repo_id=None):
         super().__init__(features, token_per_repo_id)
         self.lazy = lazy
+        # PythonArrowExtractor is stateless, so build it once instead of once per
+        # row. format_row() is called for every row of every iteration.
+        self._python_arrow_extractor = self.python_arrow_extractor()
 
     def format_row(self, pa_table: pa.Table) -> Mapping:
         if self.lazy:
             return LazyRow(pa_table, self)
-        row = self.python_arrow_extractor().extract_row(pa_table)
+        row = self._python_arrow_extractor.extract_row(pa_table)
         row = self.python_features_decoder.decode_row(row)
         return row
 
     def format_column(self, pa_table: pa.Table) -> list:
-        column = self.python_arrow_extractor().extract_column(pa_table)
+        column = self._python_arrow_extractor.extract_column(pa_table)
         column = self.python_features_decoder.decode_column(column, pa_table.column_names[0])
         return column
 
     def format_batch(self, pa_table: pa.Table) -> Mapping:
         if self.lazy:
             return LazyBatch(pa_table, self)
-        batch = self.python_arrow_extractor().extract_batch(pa_table)
+        batch = self._python_arrow_extractor.extract_batch(pa_table)
         batch = self.python_features_decoder.decode_batch(batch)
         return batch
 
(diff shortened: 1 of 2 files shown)


ebarkhordar (commented):
`extract_row` no longer goes through `ArrayExtensionArray.to_pylist()` (`features.py:840`), because `col[0].as_py()` reads the storage scalar directly. On `ArrayXD` columns with a dynamic first dimension, that changes behaviour in two ways the description does not mention:

```python

ds = Dataset.from_dict(
    {"a": [[["p", "q", "r"]], [["s", "t", "u"]]]},
    features=Features({"a": Array2D(shape=(None, 3), dtype="string")}),
)
ds[0]          # before: AttributeError: 'list' object has no attribute 'tolist'
               # after:  {'a': [['p', 'q', 'r']]}
ds["a"]        # AttributeError, unchanged
ds[0:2]        # AttributeError, unchanged
ds.to_dict()   # AttributeError, unchanged
```

and with a null row, `Array2D(shape=(None, 3), dtype="int64")`, first row `None`:

```
ds[0]     -> {'a': None}    # was {'a': nan}
ds[0:1]   -> {'a': [nan]}   # unchanged
```

Each change reads as an improvement on its own, but the row path and the column/batch paths now return different values for the same cell. Nothing in the suite catches it: `tests/features/test_array_xd.py` is 103 passed on both sides.

The fixed-shape case is fine: a `(2, 2)` int64 column with one null row and `2**60` in the other round-trips identically before and after, so the exact-integer handling there is not regressed.

Measured in a clean `python:3.12-slim` container, pyarrow 25.0.0, `b7cb10b0e` against your `f88e48e1`, with `datasets.__file__` pointed at each tree.

Is the row/batch split intended? If so, a test pinning it would help. If not, `extract_batch` probably wants the same treatment, or `to_pylist` does.

dexhunter:
Thanks for catching this. The row fast path was bypassing `ArrayExtensionArray.to_pylist()`, which owns ArrayXD shape and null conversion.

I updated `extract_row` to use `to_pylist()[0]` only for ArrayXD columns. Standard columns still use `col[0].as_py()`, so the ordinary fast path is unchanged. I also added a regression test for a dynamic `Array2D` null row.

Checks on the exact pushed commit:

- The dynamic string row and batch paths now both raise the same `AttributeError`.
- The dynamic integer null row and batch paths now both produce `nan`.
- The new regression test fails on the previous PR head and passes with this fix.
- `tests/test_formatting.py`: 25 passed, 30 skipped.
- `tests/features/test_array_xd.py`: 103 passed.
- Ruff format and check are clean.
- The frozen performance evaluator still measures about 3.65 s versus the 4.33 s baseline, so the corrected path keeps roughly a 15% improvement.

The fix is in 1d719ed64f9bcc68771a791d75f50770634aa57b.

PR #8068, opened 2026-03-15
Datasets streaming iterable packaged fixes

Michael-RDev:
This PR ckumps a bunch of fixes across streaming, iterable datasets, and packaged modules, plus workflow/docs cleanups.

A collection of fixes and long-overdue feature completions that kept getting pushed off as TODO(QL) comments. Nothing here is a dramatic architectural change, mostly things that were partially wired up but not quite finished, plus a handful of bugs.

 ## Changes:

### `max_shard_size` is active
The parameter existed on `push_to_hub` for both `IterableDataset` and `DatasetDict` but was commented out. It's now active. If you pass `max_shard_size`, the shard count gets estimated from the dataset's actual byte size. Passing both `max_shard_size` and `num_shards` at the same time raises a `ValueError` since they're two ways of saying the same thing. Also fixed a footgun where `num_proc` could be set higher than the number of shards, we now clamp it, and disable multiprocessing entirely for single-shard datasets since there's nothing to parallelize.

### `iter_arrow` added to additional iterable wrappers

`StepExamplesIterable`, `HorizontallyConcatenatedMultiSourcesExamplesIterable`, `SkipExamplesIterable`, and `TakeExamplesIterable` were all sitting with `# TODO(QL): implement iter_arrow` comments. They now have it, delegating to `_convert_to_arrow(batch_size=1)` and gating on whether the underlying iterable supports it.

### Archive glob patterns now work end-to-end

A few things were broken with prefixed archive paths like `zip://*::https://foo.bar/data.zip`:

- `resolve_pattern` was dropping the archive hop after globbing
- `_extract` was trying to extract paths that were already targeting an archive filesystem
- `_iter_from_urlpaths` didn't expand wildcards before walking

All three are fixed and covered by tests.

### `to_pandas` respects declared features

If your `IterableDataset` had declared features and the inferred Arrow schema didn't match, `to_pandas` would silently return data in the wrong types. It now passes the declared features through and casts when needed, in both the batched and non-batched paths.

### Smaller bug fixes

- **`folder_based_builder`** had a logic error — `large_string` list types in `file_names`/`*_file_names` metadata columns weren't being recognised due to a malformed `or` inside a `List(...)` call. Fixed to a proper `in [...]` check.
- **`WebDataset.VIDEO_EXTENSIONS`** had leading dots on the entries (`.mp4` instead of `mp4`) which broke extension matching.
[...]

PR #8382, opened 2026-08-03
Fix IterableDataset.to_dict() and to_polars() returning an empty generator

### LINKED ISSUE #8381: IterableDataset.to_dict() and to_polars() return an empty generator instead of the data
### Describe the bug

`IterableDataset.to_dict()` and `IterableDataset.to_polars()` each have a bare `yield` in their `batched=True` branch, which makes the whole method a generator function. Calling either one the way its docstring shows returns a generator that yields nothing, and the value the `else` branch returns is swallowed into `StopIteration.value`. Both are annotated `-> Union[dict, Iterator[dict]]` / `-> Union["pl.DataFrame", Iterator["pl.DataFrame"]]`, and the plain call is the only example in each docstring.

The sibling `to_pandas()` had the same shape and was changed to a returned generator expression in #8068; `to_dict` and `to_polars` were left as they were. `to_list()` is unaffected.

```python
from datasets import Dataset

ds = Dataset.from_dict({"col": [0, 1, 2]}).to_iterable_dataset()

print(type(ds.to_dict()))   # <class 'generator'>
print(list(ds.to_dict()))   # []
ds.to_dict()["col"]         # TypeError: 'generator' object is not subscriptable
```

The data is only reachable through the exception:

```python
g = ds.to_dict()
try:
    next(g)
except StopIteration as e:
    print(e.value)          # {'col': [0, 1, 2]}
```

`ds.to_polars()` behaves identically. On the same object, `ds.to_pandas()` returns a DataFrame and `ds.to_list()` returns the rows.

### Expected behavior

`ds.to_dict()` returns `{'col': [0, 1, 2]}` and `ds.to_polars()` returns a `polars.DataFrame`, matching `Dataset.to_dict()` / `Dataset.to_polars()` and the non-iterator half of each return annotation. `batched=True` should keep yielding one batch at a time.

### Environment info

- datasets 5.0.2.dev0, main at `b7cb10b0e`
- python 3.11.15, Linux
- pyarrow 25.0.0, polars 1.43.2, pandas 3.0.5

ebarkhordar:
### The bug

`IterableDataset.to_dict()` and `IterableDataset.to_polars()` each contain a bare `yield` in their `batched=True` branch. A bare `yield` anywhere in a function body makes the whole function a generator function, so both methods return a generator on every call, including the `batched=False` default. That generator yields nothing, and the value the `else` branch returns is swallowed into `StopIteration.value`, so the call shown in each docstring hands back an empty generator rather than the dict or the DataFrame.

`to_pandas()` had the same shape and was changed to a returned generator expression in #8068. `to_dict` and `to_polars` were left alone, so they are the only two `to_*` methods on the class still affected. This applies the same shape to both: `batched=True` stays lazy, and the non-batched branch can now return.

After the change `inspect.isgeneratorfunction` is `False` for all eight `to_*` methods on `IterableDataset`, and the only generator functions left on the class are `__iter__`, `iter` and `_iter_pytorch`.

Fixes #8381

### Verification

- Two new tests in `tests/test_iterable_dataset.py` pin both methods in both modes. Both fail on `main` (`to_dict` returns a generator, `to_polars` returns a generator) and pass on this branch, run in a clean container with the two trees mounted side by side and `datasets.__file__` printed on each run.
- Full `tests/test_iterable_dataset.py`: 445 passed here against 443 on `main`. The only failures on either side are the same four `test_interleave_dataset_with_sharding` cases, which fail identically on `main` and are unrelated to this change.
- `ruff check` and `ruff format --check` over `tests src benchmarks utils setup.py`, the `check_code_quality` job, are clean.
- Not checked: the Windows, deps-minimum, integration and py3.14 legs, and any suite other than `tests/test_iterable_dataset.py`.

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -4501,8 +4501,10 @@ class IterableDataset(DatasetInfoMixin):
         ```
         """
         if batched:
-            for table in self.with_format("arrow").iter(batch_size=batch_size):
-                yield Dataset(table, fingerprint="unset").to_dict()
+            return (
+                Dataset(table, fingerprint="unset").to_dict()
+                for table in self.with_format("arrow").iter(batch_size=batch_size)
+            )
         else:
             table = pa.concat_tables(list(self.with_format("arrow").iter(batch_size=1000)))
             return Dataset(table, fingerprint="unset").to_dict()
@@ -4593,8 +4595,10 @@ class IterableDataset(DatasetInfoMixin):
         ```
         """
         if batched:
-            for table in self.with_format("arrow").iter(batch_size=batch_size):
-                yield Dataset(table, fingerprint="unset").to_polars(schema_overrides=schema_overrides, rechunk=rechunk)
+            return (
+                Dataset(table, fingerprint="unset").to_polars(schema_overrides=schema_overrides, rechunk=rechunk)
+                for table in self.with_format("arrow").iter(batch_size=batch_size)
+            )
         else:
             table = pa.concat_tables(list(self.with_format("arrow").iter(batch_size=1000)))
             return Dataset(table, fingerprint="unset").to_polars(schema_overrides=schema_overrides, rechunk=rechunk)
--- a/tests/test_iterable_dataset.py
+++ b/tests/test_iterable_dataset.py
@@ -1796,6 +1796,29 @@ def test_iterable_dataset_to_pandas_casts_when_schema_mismatch():
     assert len(batches) == 2
 
 
+def test_iterable_dataset_to_dict():
+    dataset = Dataset.from_dict({"col": [0, 1, 2]}).to_iterable_dataset()
+
+    assert dataset.to_dict() == {"col": [0, 1, 2]}
+
+    batches = list(dataset.to_dict(batch_size=2, batched=True))
+    assert batches == [{"col": [0, 1]}, {"col": [2]}]
+
+
+@require_polars
+def test_iterable_dataset_to_polars():
+    import polars as pl
+
+    dataset = Dataset.from_dict({"col": [0, 1, 2]}).to_iterable_dataset()
+
+    df = dataset.to_polars()
+    assert isinstance(df, pl.DataFrame)
+    assert df["col"].to_list() == [0, 1, 2]
+
+    batches = list(dataset.to_polars(batch_size=2, batched=True))
+    assert [batch["col"].to_list() for batch in batches] == [[0, 1], [2]]
+
+
 @require_numpy1_on_windows
 def test_iterable_dataset_from_file(dataset: IterableDataset, arrow_file: str):
     with assert_arrow_memory_doesnt_increase():


ebarkhordar:
No rush on this one. If it helps with review, it is the same change #8068 already made to `to_pandas()`, applied to the two `to_*` methods on `IterableDataset` that still return an empty generator, and I am happy to split `to_polars` into its own PR if you would rather take them one at a time.

shashvat-singham (commented):
Ran the branch — the four `to_dict`/`to_polars`/`to_pandas` tests pass, and I confirmed the bug on `main` first:

```python
>>> ds = Dataset.from_dict({"col": [0, 1, 2]}).to_iterable_dataset()
>>> type(ds.to_dict()).__name__
'generator'
>>> type(ds.to_polars()).__name__
'generator'
>>> type(ds.to_pandas()).__name__
'DataFrame'
```

Worth spelling out the mechanism in the description, because the diff looks like a cosmetic rewrite until you see it: a `yield` anywhere in the body makes the *whole* function a generator function, so the `else` branch's `return` never returned to the caller — it just became the `StopIteration` value of a generator nobody iterated. That's why the non-batched path silently produced an empty generator instead of a dict, and it's a bug you can stare straight at without noticing.

The thing that makes me confident this is the right fix rather than one of several options: `to_pandas` already does exactly this, and is the reason it's the only one of the three that works today:

```python
if batched:
    return (
        Dataset(...).to_pandas()
        for table in self.with_format("arrow").iter(batch_size=batch_size)
    )
else:
    ...
    return Dataset(table, ...).to_pandas()
```

So this brings `to_dict` and `to_polars` in line with the sibling that already had the correct shape, rather than inventing a pattern. Might be worth saying so in the PR body.

One asymmetry I noticed while there, offered only as a question: `to_pandas` also threads `info=info` and runs `maybe_cast_to_declared_features` over each table, and `to_dict`/`to_polars` do neither. I tried to construct a divergence from that (declared `Value("int32")` against inferred data) and couldn't — `to_polars` returns `Int32` either way, presumably because the arrow formatting already carries the features — so it may well be redundant in `to_pandas` rather than missing in the other two. Either way it's out of scope here; just flagging it since the three are otherwise meant to be parallel and someone may want to reconcile them.

ebarkhordar:
Thanks for running it on both sides. The mechanism and the `to_pandas` parallel are the first two paragraphs of the description, so we read it the same way, and the independent reproduction on `main` is worth more than either.

On the `info=` and `maybe_cast_to_declared_features` asymmetry: I have not measured that one, and I agree it is out of scope here. If it turns out to matter it wants its own issue, because the answer decides whether `to_pandas` is doing something redundant or the other two are missing a cast.

shashvat-singham:
Fair — you had both in the description already and I suggested adding what was there. My mistake; I'd read the diff before the prose.

Agreed on the asymmetry being its own issue. For whoever picks it up, the one concrete thing I have is that I couldn't construct a divergence: declaring `Value("int32")` against inferred data, `to_polars` returns `Int32` with or without the cast, presumably because the arrow formatting already carries the features. So the evidence I have points at `to_pandas` being redundant rather than the other two missing something — but that's one case, not a measurement, and it's exactly the kind of thing that deserves its own issue rather than a footnote here.

Nothing blocking from me on this PR.

ebarkhordar:
Closing this one, it's been a month with no movement and I'd rather not leave stale PRs sitting in the queue. I've kept the branch, so reopening is easy if this is still worth doing, though I haven't rechecked it against main since early August.

PR #8387, opened 2026-08-04
Fix `deepcopy` bottleneck in state_dict()

### LINKED ISSUE #8393: CPU Bottleneck in `IterableDataset.state_dict()` when shuffling multiple shards
**Describe the bug**
There is a significant CPU bottleneck in `IterableDataset` dataloading when `buffer_size` and `max_buffer_input_shards` are greater than 1 (i.e. highly parallelized streaming with a shuffle buffer).

Currently, `_BaseExamplesIterable.state_dict()` falls back to performing a full recursive `copy.deepcopy(self._state_dict)`. 
In nested iterator chains (like `BufferShuffledExamplesIterable` wrapping `CyclingMultiSourcesExamplesIterable` wrapping multiple `ArrowExamplesIterable` shards), the `state_dict()` is queried and updated **on every yielded row**. 

Because `copy.deepcopy` relies on Python's `pickle` mechanics for object introspection and memoization, it incurs massive CPU overhead. Furthermore, `pickle` often crashes or fails gracefully when it encounters unpicklable objects like generator states. 
When `max_buffer_input_shards` is scaled up, the dictionary size grows, and the deepcopy completely blocks the CPU, throttling the dataloader significantly.

**Steps to Reproduce**
```python
import datasets

ds = datasets.load_dataset(
    "parquet",
    data_files="gs://your-bucket/benchmark_data/shard_*.parquet",
    split="train",
    streaming=True
)

# Scaling max_buffer_input_shards triggers the bottleneck
ds = ds.shuffle(seed=0, buffer_size=10000, max_buffer_input_shards=10)
ds = split_dataset_by_node(ds, rank=0, world_size=8)
ds = ds.with_format("torch")

loader = DataLoader(
    ds,
    batch_size=8,
    num_workers=4,
    prefetch_factor=2,
    persistent_workers=True,
)

for epoch in range(3):
    ds.set_epoch(epoch)
    for batch in loader:
        pass
```
* **Before fix:** ~103.5 seconds per epoch.
* **After fix:** ~65.8 seconds per epoch.

**Proposed Solution**
We have drafted a fix in PR #8387 that replaces the expensive `deepcopy()` logic with a fast, dynamic `_fast_copy()` helper directly inside `_BaseExamplesIterable.state_dict()`. Because dataset `state_dict`s strictly consist of acyclic standard collections (dict, list) and primitive state objects, `_fast_copy()` manually recursively walks `dict` and `list` structures and natively clones the values, bypassing `pickle` and `deepcopy` entirely.

**Environment info**
- `datasets` version: `main` branch
- Python version: 3.13

yuxin00j:
Fix #8393 
This PR significantly resolves a major CPU bottleneck in `IterableDataset` dataloading when `buffer_size` and `max_buffer_input_shards` are > 1. 

**The Bug:**
Currently, `_BaseExamplesIterable.state_dict()` falls back to performing a full recursive `copy.deepcopy(self._state_dict)`. 
In nested iterator chains (like `BufferShuffledExamplesIterable` wrapping `CyclingMultiSourcesExamplesIterable` wrapping multiple `ArrowExamplesIterable` shards), the `state_dict()` is queried and updated **on every yielded row**. 

Because `copy.deepcopy` relies on Python's `pickle` mechanics for object introspection and memoization, it incurs significant overhead. Worse, `pickle` often crashes when it encounters unpicklable objects like generators. When `mbis` is scaled up and the dictionary size grows, the deepcopy completely blocks the CPU, hiding underneath the shuffle buffer processing. 

**The Fix:**
We replaced the expensive `deepcopy()` logic with a dynamic, lightweight `_fast_copy()` helper directly inside `_BaseExamplesIterable.state_dict()`. 

Because dataset `state_dict`s strictly consist of standard collections and primitive state objects (and importantly, contain no circular references), `_fast_copy()` manually recursively walks `dict` and `list` structures. For any other object, it natively returns the object (or uses its `.copy()` method if available), bypassing `pickle` and `deepcopy` entirely.

This allows us to dynamically serialize state without explicitly overriding `state_dict` in every single dataset subclass, restoring maintainability and drastically improving speed.

**Impact:**
We measured workloads using a high number of concurrent shards (`max_buffer_input_shards=10`):

```python
import datasets

ds = datasets.load_dataset(
    "parquet",
    data_files="gs://your-bucket/benchmark_data/shard_*.parquet",
    split="train",
    streaming=True
)

ds = ds.shuffle(seed=0, buffer_size=10000, max_buffer_input_shards=10)
ds = split_dataset_by_node(ds, rank=0, world_size=8)
ds = ds.with_format("torch")

loader = DataLoader(
    ds,
    batch_size=8,
    num_workers=4,
    prefetch_factor=2,
    persistent_workers=True,
)

for epoch in range(3):
    ds.set_epoch(epoch)
    for batch in loader:
        pass
```

* **Before this PR:** ~103.5 seconds per epoch.
* **After this PR:** ~65.8 seconds per epoch.

This single PR shaves ~38 seconds of pure CPU overhead off the workload, generating a **36% overall speedup**, entirely restoring the expected linear scaling of shuffle buffers when reading across many shards!

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -201,14 +201,25 @@ def shift_ex_examples_rngs(ex_iterable: "_BaseExamplesIterable", value: int) -> 
 
 def _fast_copy(state):
     """Fast acyclic copy of state_dict data structures, bypassing slow pickle/deepcopy introspection."""
-    if isinstance(state, dict):
-        return {k: _fast_copy(v) for k, v in state.items()}
-    elif isinstance(state, list):
-        return [_fast_copy(v) for v in state]
-    elif isinstance(state, tuple):
-        return tuple(_fast_copy(v) for v in state)
-    elif isinstance(state, (int, float, bool, str, bytes, type(None))):
+    # Use type() instead of isinstance() for 2x faster type checks
+    state_type = type(state)
+    if state_type in (int, float, bool, str, bytes, type(None)):
         return state
+    elif state_type is list:
+        return [
+            v if type(v) in (int, float, bool, str, bytes, type(None)) else _fast_copy(v)
+            for v in state
+        ]
+    elif state_type is dict:
+        return {
+            k: (v if type(v) in (int, float, bool, str, bytes, type(None)) else _fast_copy(v))
+            for k, v in state.items()
+        }
+    elif state_type is tuple:
+        return tuple(
+            v if type(v) in (int, float, bool, str, bytes, type(None)) else _fast_copy(v)
+            for v in state
+        )
     elif hasattr(state, "copy") and callable(state.copy):
         return state.copy()
     else:
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -199,6 +199,22 @@ def shift_ex_examples_rngs(ex_iterable: "_BaseExamplesIterable", value: int) -> 
     return set_seed_recursively(ex_iterable)
 
 
+def _fast_copy(state):
+    """Fast acyclic copy of state_dict data structures, bypassing slow pickle/deepcopy introspection."""
+    if isinstance(state, dict):
+        return {k: _fast_copy(v) for k, v in state.items()}
+    elif isinstance(state, list):
+        return [_fast_copy(v) for v in state]
+    elif isinstance(state, tuple):
+        return tuple(_fast_copy(v) for v in state)
+    elif isinstance(state, (int, float, bool, str, bytes, type(None))):
+        return state
+    elif hasattr(state, "copy") and callable(state.copy):
+        return state.copy()
+    else:
+        raise TypeError(f"Unsupported type for _fast_copy: {type(state)}")
+
+
 class _BaseExamplesIterable:
     """Base class for the examples iterable used by an IterableDataset"""
 
@@ -274,23 +290,6 @@ class _BaseExamplesIterable:
 
     def state_dict(self) -> dict:
         if self._state_dict:
-
-            def _fast_copy(state):
-                # Native copy.deepcopy() is extremely slow because it does extensive memoization
-                # (to prevent infinite loops on cyclical references) and dynamic introspection
-                # (checking for __deepcopy__, __reduce__, etc. using the pickle module).
-                # Furthermore, pickle crashes when encountering generators/iterators.
-                # Since we know dataset state_dicts contain strictly acyclic standard collections
-                # (dict, list) and primitive state objects, we can bypass deepcopy overhead
-                # entirely and cleanly clone the exact state needed natively.
-                if isinstance(state, dict):
-                    return {k: _fast_copy(v) for k, v in state.items()}
-                elif isinstance(state, list):
-                    return [_fast_copy(v) for v in state]
-                elif hasattr(state, "copy"):
-                    return state.copy()
-                return state
-
             return _fast_copy(self._state_dict)
         raise RuntimeError("State dict is not initialized, please call ex_iterable._init_state_dict() first.")
 
@@ -2627,8 +2626,7 @@ class IterableDataset(DatasetInfoMixin):
         """
         if not self._state_dict:
             return {}
-        # Since ex_iterable.state_dict() returns a fresh isolated dict, we don't need a full recursive deepcopy.
-        return dict(self._state_dict)
+        return _fast_copy(self._state_dict)
 
     def load_state_dict(self, state_dict: dict) -> None:
         """Load the state_dict of the dataset.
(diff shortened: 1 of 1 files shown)


src/datasets/iterable_dataset.py
```diff
@@ -274,7 +274,24 @@ def _inner_load_state_dict(state, new_state):
 
     def state_dict(self) -> dict:
         if self._state_dict:
-            return deepcopy(self._state_dict)
+
+            def _fast_copy(state):
```
zhixiangli: How about moving this func to module level for reusability?

src/datasets/iterable_dataset.py
```diff
@@ -274,7 +274,24 @@ def _inner_load_state_dict(state, new_state):
...
+                # Since we know dataset state_dicts contain strictly acyclic standard collections
+                # (dict, list) and primitive state objects, we can bypass deepcopy overhead
+                # entirely and cleanly clone the exact state needed natively.
+                if isinstance(state, dict):
+                    return {k: _fast_copy(v) for k, v in state.items()}
+                elif isinstance(state, list):
+                    return [_fast_copy(v) for v in state]
+                elif hasattr(state, "copy"):
+                    return state.copy()
+                return state
```
zhixiangli: We need to throw TypeError for unexpected types, such as those outside of the expected dict, list, tuple, primitive, numpy scalar, and numpy array types.

src/datasets/iterable_dataset.py
```diff
@@ -2608,7 +2625,10 @@ def state_dict(self) -> dict:
         >>> dataloader.load_state_dict(state_dict)  # uses ds.load_state_dict() under the hood
         ```
         """
-        return deepcopy(self._state_dict)
+        if not self._state_dict:
+            return {}
+        # Since ex_iterable.state_dict() returns a fresh isolated dict, we don't need a full recursive deepcopy.
+        return dict(self._state_dict)
```
zhixiangli: _fast_copy?

src/datasets/iterable_dataset.py
zhixiangli, line 289: Could you confirm are there any other `deepcopy` needed to be replaced with `_fast_copy`?

src/datasets/iterable_dataset.py
yuxin00j: Done.

src/datasets/iterable_dataset.py
yuxin00j: Done.

src/datasets/iterable_dataset.py
yuxin00j: Done.

src/datasets/iterable_dataset.py
yuxin00j: All remaining 31 deepcopy(...) calls in src/datasets/iterable_dataset.py fall into two categories where _fast_copy cannot or should not be used:
1. RNG Generators (deepcopy(generator) on np.random.Generator): These internal RNG objects cannot be copied by structural dictionary/list recursion and must retain deepcopy.
2. Dataset Setup Metadata (_distributed, _formatting, _info, or d._ex_iterable in concatenate_datasets): These execute only once during dataset pipeline construction (e.g., calling .map(), .shuffle(), .rename_column()) and do not occur during streaming iteration or checkpointing.

src/datasets/iterable_dataset.py
```diff
@@ -199,6 +199,33 @@ def set_seed_recursively(ex_iterable):
     return set_seed_recursively(ex_iterable)
 
 
+_PRIMITIVE_TYPES = (int, float, bool, str, bytes, type(None))
```
zhixiangli, line 202: nit: `frozenset` should be faster

src/datasets/iterable_dataset.py
```diff
@@ -199,6 +199,33 @@ def set_seed_recursively(ex_iterable):
...
+
+def _fast_copy(state):
+    """Fast acyclic copy of state_dict data structures, bypassing slow pickle/deepcopy introspection."""
+    # Use type() instead of isinstance() for 2x faster type checks
+    state_type = type(state)
+    if state_type in _PRIMITIVE_TYPES:
```
zhixiangli, line 209: nit: perform this primitive type check last, before the else statement.

src/datasets/iterable_dataset.py
```diff
@@ -199,6 +199,33 @@ def set_seed_recursively(ex_iterable):
...
+        return state
+    elif state_type is list:
+        return [
+            v if type(v) in _PRIMITIVE_TYPES else _fast_copy(v)
```
zhixiangli: nit: simplify to `[_fast_copy(v) for v in state]`; the same applies to dict and tuple.

lhoestq (approved):
lgtm :)

PR #8395, opened 2026-08-04
Forward iter_arrow, is_typed and features from RepeatExamplesIterable

### LINKED ISSUE #8394: IterableDataset.repeat() drops the Arrow fast path and widens numpy dtypes
### Describe the bug

`RepeatExamplesIterable` is the only wrapper in `iterable_dataset.py` that does not forward `iter_arrow`, `is_typed` and `features` to the iterable it wraps. Two things follow from that: `.repeat()` drops the Arrow fast path, and under `numpy` format it silently widens the declared dtypes.

`SkipExamplesIterable` (line 2012) and `TakeExamplesIterable` (line 2175) sandwich `RepeatExamplesIterable` (line 2120) and both carry the three-property block; `RepeatExamplesIterable` carries none of them. Walking every `_BaseExamplesIterable` subclass at `b7cb10b`, 12 of the 14 wrap another iterable, 11 of those 12 forward all three, and `RepeatExamplesIterable` is the one that forwards none.

```python

features = Features({"i32": Value("int32"), "f32": Value("float32"), "u8": Value("uint8")})
ds = Dataset.from_dict({"i32": [1, 2], "f32": [1.5, 2.5], "u8": [3, 4]}, features=features)
ds = ds.to_iterable_dataset().with_format("numpy")

print({k: v.dtype for k, v in next(iter(ds)).items()})
print({k: v.dtype for k, v in next(iter(ds.take(2))).items()})
print({k: v.dtype for k, v in next(iter(ds.repeat(2))).items()})
```

```
{'i32': int32,   'f32': float32,   'u8': uint8}     # plain
{'i32': int32,   'f32': float32,   'u8': uint8}     # take(2)
{'i32': int64,   'f32': float64,   'u8': int64}     # repeat(2)
```

Same shape on throughput, 100k rows under `arrow` format:

```
take(100000)  0.500s
repeat(1)     6.644s
```

`repeat` produces a `RepeatExamplesIterable` whose `iter_arrow` is `None` sitting on top of an `ArrowExamplesIterable` whose `iter_arrow` is not, so iteration falls back to the per-example path and the types are re-inferred from Python objects.

### Expected behavior

`.repeat()` repeats rows. It should not change their dtype, and it should not disable the Arrow path when the underlying iterable supports it, the same way `.take()` and `.skip()` do not.

Scope of what I checked: `numpy` format on scalar `Value` columns, single-example iteration. I did not check the torch formatter or non-scalar feature types.

### Environment info

- `datasets` main at `b7cb10b0e38ab8bbefdb142d0279bc85ac8c7ab8` (installed with `pip install -e .`)
- Python 3.12, Linux
- pyarrow 25.0.0, numpy 2.5.1

A PR follows.

ebarkhordar:
`RepeatExamplesIterable` never forwarded `iter_arrow`, `is_typed` or `features` to the iterable it wraps, so `.repeat()` stranded an Arrow-capable child behind a non-Arrow parent: iteration fell back to the per-example path and the dtypes were re-inferred from Python objects.

This copies the three-property block and the `_iter_arrow` loop from `SkipExamplesIterable` and `TakeExamplesIterable`, which sit either side of it. `_iter_arrow` mirrors `__iter__` exactly, including the per-pass reset of the child state dict, so `state_dict()`/`load_state_dict()` behaves the same on both paths.

### Verification

On `main` at `b7cb10b`, Python 3.12, pyarrow 25.0.0, numpy 2.5.1:

- The three new tests fail before the change and pass after. `test_iterable_dataset_repeat_keeps_numpy_dtypes` reports `int64/float64/int64` before against a declared `int32/float32/uint8`; `test_repeat_examples_iterable_arrow` asserts `iter_arrow is not None` and runs both `assert_load_state_dict_resumes_iteration` and `assert_load_state_dict_resumes_arrow_iteration`.
- `tests/test_iterable_dataset.py` goes from 441 passed to 445 passed, with the same 4 pre-existing `test_interleave_dataset_with_sharding` failures on both sides (a missing optional dependency in my environment, unrelated to this change).
- Arrow-format iteration over 100k rows drops from 6.644s to 0.469s, which is parity with `.take(100000)` at 0.465s. `list(ds.repeat(2))` is unchanged.
- `ruff format --check` and `ruff check` are clean on both files.

What I did not check: the torch formatter, and non-scalar feature types such as `Image`. The dtype claim above is scoped to scalar `Value` columns under `numpy` format.

Fixes #8394

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -2131,6 +2131,18 @@ class RepeatExamplesIterable(_BaseExamplesIterable):
         self.ex_iterable = ex_iterable
         self.num_times = num_times
 
+    @property
+    def iter_arrow(self):
+        return self._iter_arrow if self.ex_iterable.iter_arrow else None
+
+    @property
+    def is_typed(self):
+        return self.ex_iterable.is_typed
+
+    @property
+    def features(self):
+        return self.ex_iterable.features
+
     def _init_state_dict(self) -> dict:
         self._state_dict = {
             "repeat_index": 0,
@@ -2145,6 +2157,17 @@ class RepeatExamplesIterable(_BaseExamplesIterable):
             if self.num_times is not None and repeat_index >= max(self.num_times, 0):
                 break
             yield from self.ex_iterable
+            repeat_index += 1
+            if self._state_dict:
+                self._state_dict["repeat_index"] = repeat_index
+                self._state_dict["examples_iterable"] = self.ex_iterable._init_state_dict()
+
+    def _iter_arrow(self):
+        repeat_index = self._state_dict["repeat_index"] if self._state_dict else 0
+        while True:
+            if self.num_times is not None and repeat_index >= max(self.num_times, 0):
+                break
+            yield from self.ex_iterable.iter_arrow()
             repeat_index += 1
             if self._state_dict:
                 self._state_dict["repeat_index"] = repeat_index
--- a/tests/test_iterable_dataset.py
+++ b/tests/test_iterable_dataset.py
@@ -1445,6 +1445,26 @@ def test_repeat_examples_iterable(n, num_times):
             assert next(iterator)[1] == all_examples[i % len(all_examples)], f"iteration {i} failed,"
 
 
+@pytest.mark.parametrize("num_times", [2, 0])
+def test_repeat_examples_iterable_arrow(num_times):
+    base_ex_iterable = ArrowExamplesIterable(generate_tables_fn, {"n": 10})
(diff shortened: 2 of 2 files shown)


ebarkhordar:
No rush on this one. The three properties are byte-identical to the ones `TakeExamplesIterable` and `SkipExamplesIterable` already carry either side of it, and if you would rather keep the change minimal I am happy to drop `_iter_arrow` and leave just `is_typed` and `features`.

lhoestq (approved):
lgtm !

ebarkhordar:
Thanks for the merge. `RepeatExamplesIterable` now carries the same three properties as the `Take` and `Skip` wrappers either side of it.

PR #8407, opened 2026-08-06
Optimize PyArrow block shuffling with dynamic rebatched chunk interleaving

yuxin00j:
When streaming PyArrow datasets with shuffling enabled (`buffer_size > 0`), the `BufferShuffledExamplesIterable._iter_arrow()` method historically relied on a Python-level `list` to maintain the shuffle buffer, replacing elements one-by-one. In a multi-worker `DataLoader` environment, this per-row generator approach caused severe GIL contention, high generator state-transition overhead, and inefficient memory interleaving. 
This PR introduces **Amortized Native C++ Block Shuffling** for PyArrow streaming. It completely replaces the Python 1-row exchange loop with vectorized block shuffling:
- Incoming PyArrow tables are accumulated in memory until the combined size reaches `2 * self.buffer_size`.
- The buffered tables are concatenated via `pyarrow.concat_tables(tables)`.
- The entire block is shuffled in a single vectorized C++ call using `pyarrow.compute.take(buffer_table, indices)`.
- The first `current_len - self.buffer_size` rows are yielded immediately, while the remaining `self.buffer_size` rows are retained as the seed buffer for the next shuffle block.
By eliminating the 1-row `yield` loop, we drastically drop CPU overhead and unlock the full IO throughput potential of PyArrow when combined with `split_dataset_by_node` and multiple dataloader workers.
### Impact
We measured workloads using a high number of concurrent shards (`max_buffer_input_shards=10`) and combined the change in this PR and PR #8387:

```python
import datasets

ds = datasets.load_dataset(
    "parquet",
    data_files="gs://your-bucket/benchmark_data/shard_*.parquet",
    split="train",
    streaming=True
)

ds = ds.shuffle(seed=0, buffer_size=10000, max_buffer_input_shards=10)
ds = split_dataset_by_node(ds, rank=0, world_size=8)
ds = ds.with_format("torch")

loader = DataLoader(
    ds,
    batch_size=8,
    num_workers=4,
    prefetch_factor=2,
    persistent_workers=True,
)

for epoch in range(3):
    ds.set_epoch(epoch)
    for batch in loader:
        pass
```

* **Before this PR:** ~103.5 seconds per epoch.
* **After #8387:** ~65.8 seconds per epoch.
* **After this PR :** ~38.36s seconds per epoch.
**62.9% overall speedup**

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -1978,12 +1978,15 @@ class BufferShuffledExamplesIterable(_BaseExamplesIterable):
             tables.append(pa_table)
             current_len += len(pa_table)
 
-            # Amortize shuffle cost by waiting until buffer is full
+            # Amortize shuffle cost by waiting until buffer is full.
+            # Once the buffer reaches `buffer_size`, we shuffle it, yield half, and keep half.
+            # This maintains a rolling randomized buffer while strictly bounding memory usage.
             if current_len >= self.buffer_size:
                 buffer_table = pa.concat_tables(tables)
                 indices = rng.permutation(current_len)
                 shuffled_table = pc.take(buffer_table, indices)
 
+                # Keep half of the buffer to mix with incoming data on the next iteration
                 keep_rows = self.buffer_size // 2
                 rows_to_yield = current_len - keep_rows
                 yield key, shuffled_table.slice(0, rows_to_yield)
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -1978,17 +1978,18 @@ class BufferShuffledExamplesIterable(_BaseExamplesIterable):
             tables.append(pa_table)
             current_len += len(pa_table)
 
-            # Amortize shuffle cost by waiting until buffer is 2x full
-            if current_len >= 2 * self.buffer_size:
+            # Amortize shuffle cost by waiting until buffer is full
+            if current_len >= self.buffer_size:
                 buffer_table = pa.concat_tables(tables)
                 indices = rng.permutation(current_len)
                 shuffled_table = pc.take(buffer_table, indices)
 
-                rows_to_yield = current_len - self.buffer_size
+                keep_rows = self.buffer_size // 2
+                rows_to_yield = current_len - keep_rows
                 yield key, shuffled_table.slice(0, rows_to_yield)
 
                 tables = [shuffled_table.slice(rows_to_yield)]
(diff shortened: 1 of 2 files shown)


Yonghui-Lee:
The PR changes the shuffle results for a given seed, seems to cause test_iterable_dataset_shuffle_buffer_uses_multiple_input_shards to fail.

src/datasets/iterable_dataset.py
```diff
@@ -1963,21 +1963,37 @@ def __iter__(self):
...
-                mem_buffer.append((key, pa_table))
-        # when we run out of examples, we shuffle the remaining examples in the buffer and yield them
-        rng.shuffle(mem_buffer)
-        yield from mem_buffer
+            last_key = key
+            tables.append(pa_table)
+            current_len += len(pa_table)
+
+            # Amortize shuffle cost by waiting until buffer is 2x full
+            if current_len >= 2 * self.buffer_size:
```
Yonghui-Lee: It kind of exceeds buffer size limit. Will it be better if we wait until buffer full, then keep half and yield half?

yuxin00j: Updated the logic to implement this.

yuxin00j:
> The PR changes the shuffle results for a given seed, seems to cause test_iterable_dataset_shuffle_buffer_uses_multiple_input_shards to fail.

Updated the test.

lhoestq (commented):
Nice improvement ! I think it would make sense to also update the `__iter__` method as well so both `__iter__` and `_iter_arrow` return examples in the same order - otherwise it can be confusing for users IMO

yuxin00j:
> Nice improvement ! I think it would make sense to also update the `__iter__` method as well so both `__iter__` and `_iter_arrow` return examples in the same order - otherwise it can be confusing for users IMO

@lhoestq sure, I have updated the `__iter__` method.

lhoestq (approved):
lgtm ! I could even get a x150 speed boost with a local dataset !

lhoestq:
This is really, really useful btw ! Would it be ok if I share this improvement online and mention you and the GCS team ?

edit: I couldn't wait lol: https://x.com/lhoestq/status/2096966828547277195?s=20

yuxin00j:
> This is really, really useful btw ! Would it be ok if I share this improvement online and mention you and the GCS team ?
> 
> edit: I couldn't wait lol: https://x.com/lhoestq/status/2096966828547277195?s=20

@lhoestq Yes, thanks for the recognition!

PR #8408, opened 2026-08-06
Save the shuffle buffer content in the state dict so resuming is exact

### LINKED ISSUE #7901: ShuffledDataSourcesArrowExamplesIterable cannot properly resume from checkpoint
### Describe the bug

ShuffledDataSourcesArrowExamplesIterable cannot properly resume from checkpoint

1. The reproducible code is as follows：

```
from datasets import Dataset, concatenate_datasets, interleave_datasets
ds = Dataset.from_dict({"a": range(12)}).to_iterable_dataset(num_shards=1)
ds = ds.shuffle(seed=42)

for idx, example in enumerate(ds):
    print(example)
    if idx == 2: #The checkpoint can be loaded correctly only when idx <= 1.
        state_dict = ds.state_dict()
        print("checkpoint")
        break

print("state_dict: ",state_dict)

ds.load_state_dict(state_dict)
print(f"restart from checkpoint")
for example in ds:
    print(example)
```
2. The error message is as follows:
```
{'a': 0}
{'a': 7}
{'a': 6}
checkpoint
state_dict:  {'examples_iterable': {'examples_iterable': {'examples_iterable': {'shard_idx': 1, 'shard_example_idx': 0, 'type': 'ShuffledDataSourcesArrowExamplesIterable'}, 'previous_state': {'shard_idx': 0, 'shard_example_idx': 0, 'type': 'ShuffledDataSourcesArrowExamplesIterable'}, 'batch_idx': 12, 'num_chunks_since_previous_state': 12, 'cropped_chunk_length': 0, 'type': 'RebatchedArrowExamplesIterable'}, 'previous_state': {'examples_iterable': {'shard_idx': 1, 'shard_example_idx': 0, 'type': 'ShuffledDataSourcesArrowExamplesIterable'}, 'previous_state': {'shard_idx': 0, 'shard_example_idx': 0, 'type': 'ShuffledDataSourcesArrowExamplesIterable'}, 'batch_idx': 12, 'num_chunks_since_previous_state': 12, 'cropped_chunk_length': 0, 'type': 'RebatchedArrowExamplesIterable'}, 'batch_idx': 3, 'num_chunks_since_previous_state': 2, 'cropped_chunk_length': 0, 'type': 'RebatchedArrowExamplesIterable'}, 'epoch': 0}
restart from checkpoint
Loading a state dict of a shuffle buffer of a dataset without the buffer content.The shuffle buffer will be refilled before starting to yield new examples.
```
### Expected behavior

I want a correct resume from any checkpoint, but currently the checkpoint can only be loaded correctly when idx <= 1.

### Environment info

datasets Version: 4.4.1

@lhoestq

thejamesgore:
## Overview

There are two problems when resuming a shuffled iterable dataset. The known one from the warning and the discussion in #7901 is that the shuffle buffer content isn't saved in the state dict so you lose roughly `buffer_size` examples every resume.

The second is you can get an example twice. The `RebatchedArrowExamplesIterable` that wraps the pipeline resumes by rewinding to a `previous_state` snapshot and skipping chunks it already delviered which assumes the inner stream is deterministic from its state. The shuffle buffer isn't as it restarts empty with a fresh rng so the rewound rows get reshuffled differently and an already yielded row can come out again. I instrumented it at the issue's k=20: the resume re-reads rows 58 and 158. 158 was still in the buffer but 58 had already been yielded so that's the {58: 2} in the issue. Silent duplicates in a training run that resumes often is worse than the documented loss.

## Fix

Save the buffer content, the rng pick count and the end-of-data shuffle flag in the state dict. Resume restores the buffer and fast-forwards the rng, so it carries on exactly where it left off and both problems go away.

Two costs dealt with: state gets captured once per yielded example, so `state_dict()` keeps the buffer as a tuple instead of deep-copying it. And the buffered arrow tables are usually zero-copy slices of a whole source table, a buffer_size=1000 checkpoint over 50k rows came out at 1.6 GiB. Slices get rewritten as stand-alone tables before entering the buffer, which brings it down to ~400 KiB.

- The repro from #7901: no duplicates, nothing missing. Sweeping the checkpoint over k=1..100 for buffer sizes 4/8/16/32: zero duplicates (was 22/9/3/0).
- Resumed iteration is identical to an uninterrupted run across shard counts, buffer sizes and checkpoint positions, including mid-drain and resume/checkpoint/resume chains.
- Removed the BufferShuffledExamplesIterable exemption in test_no_iter_arrow so the standard resume assertion covers it, plus new resume tests. All of them fail without the fix.
- 501 passed in test_iterable_dataset.py, ruff clean, iteration order unchanged from main for a fixed seed.
- 50k examples with buffer_size=1000: ~2% overhead with state tracking, checkpoint ~400 KiB.

 Resolves #7901

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -1927,57 +1927,99 @@ class BufferShuffledExamplesIterable(_BaseExamplesIterable):
         return self._iter_arrow if self.ex_iterable.iter_arrow else None
 
     def _init_state_dict(self) -> dict:
+        # the buffer state is stored in the same dict as the wrapped iterable's state, so that
+        # state dicts from versions that didn't save the buffer content can still be loaded.
+        # The buffer content is stored as a tuple so that state merging and copying treat it
+        # as an atomic value.
         self._state_dict = self.ex_iterable._init_state_dict()
-        self._original_state_dict = self.state_dict()
+        self._state_dict["shuffle_buffer"] = ()
+        self._state_dict["shuffle_buffer_num_picks"] = 0
+        self._state_dict["shuffle_buffer_end_of_data_shuffled"] = False
         return self._state_dict
 
-    def load_state_dict(self, state_dict: dict) -> dict:
+    def state_dict(self) -> dict:
         if self._state_dict:
-            if state_dict != self._original_state_dict:
-                logger.warning(
-                    "Loading a state dict of a shuffle buffer of a dataset without the buffer content."
-                    "The shuffle buffer will be refilled before starting to yield new examples."
-                )
-        return super().load_state_dict(state_dict)
+            # avoid deep-copying the buffer content: the tuple is immutable and its entries are
+            # only ever replaced, never mutated in place. This runs on every state capture.
+            return {
+                key: (value if key == "shuffle_buffer" else deepcopy(value)) for key, value in self._state_dict.items()
+            }
+        raise RuntimeError("State dict is not initialized, please call ex_iterable._init_state_dict() first.")
+
+    def load_state_dict(self, state_dict: dict) -> dict:
+        if "shuffle_buffer" not in state_dict:
+            # state dict from a version that didn't save the buffer content
+            logger.warning(
+                "Loading a state dict of a shuffle buffer of a dataset without the buffer content."
+                "The shuffle buffer will be refilled before starting to yield new examples."
+            )
+        inner_state_dict = {key: value for key, value in state_dict.items() if not key.startswith("shuffle_buffer")}
+        self.ex_iterable.load_state_dict(inner_state_dict)
+        self._state_dict = self.ex_iterable._state_dict
+        self._state_dict["shuffle_buffer"] = tuple(state_dict.get("shuffle_buffer", ()))
+        self._state_dict["shuffle_buffer_num_picks"] = state_dict.get("shuffle_buffer_num_picks", 0)
+        self._state_dict["shuffle_buffer_end_of_data_shuffled"] = state_dict.get(
+            "shuffle_buffer_end_of_data_shuffled", False
+        )
+        return self._state_dict
 
     @staticmethod
     def _iter_random_indices(rng: np.random.Generator, buffer_size: int, random_batch_size=1000) -> Iterator[int]:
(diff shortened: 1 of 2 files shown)


MohammadSadeghSalehi:
I reported the duplicate this fixes over in #7901, so I checked this branch against the full sweep I had rather than just the one case.

Twenty-seven configurations, varying seed, checkpoint offset, buffer size and shard count, on `pr8408` at `bfcf1eac5`:

```
 seed    k  buf shards |  dupes  missing  exact
   17    5   16      4 |      0        0   True
   17   20   16      4 |      0        0   True
   17   20    8      4 |      0        0   True
   17   20   16      2 |      0        0   True
   ...
27 configurations: 0 with duplicates, 0 with missing rows,
0 not exactly equal to the uninterrupted stream
```

The third column is the strict one: the resumed stream is compared element by element against the uninterrupted stream, so this is exact resume rather than merely the absence of a repeat. On 5.0.1 the same sweep gives duplicates in 3 of 27 and loses `buffer_size` rows in all 27.

The backward-compatible path also behaves. Stripping the new keys out of a checkpoint, which is what a state dict written by an older version looks like, reproduces the old behaviour precisely rather than failing in some new way:

```
Loading a state dict of a shuffle buffer of a dataset without the buffer content....
loaded a pre-fix state dict without raising: 164 rows after resume
duplicates: 1   missing: 17
```

That is the same 1 duplicate and 17 missing rows as the original report, so old checkpoints stay loadable and stay exactly as lossy as they were, which seems like the right call.

One note for anyone reproducing this: the new keys are nested under
`examples_iterable.examples_iterable.*` and `examples_iterable.previous_state.*`, not at the top level of the state dict. My first attempt at the compatibility check stripped only top-level keys, which silently tested nothing and looked like a pass.

Your account of the mechanism also matches what I saw and is more precise than mine. I had stopped at `BufferShuffledExamplesIterable` holding no state of its own and reporting the source position; the part I missed is that the rewind-and-skip in the rebatching layer is only correct while the inner stream replays deterministically, which the buffer breaks by coming back empty with a fresh rng. That explains why it reproduces at some checkpoint offsets and not others, which I could not account for.

PR #7817, opened 2025-10-14
fix: better args passthrough for `_batch_setitems()`

sghng:
In Python 3.14, there's a change in the signature of `_Pickler._batch_setitems`.

It's changed to:

```python
# pickle.py

def _batch_setitems(self, items, obj):
    # Helper to batch up SETITEMS sequences; proto >= 1 only
    save = self.save
    write = self.write
```

To accomodate this, in `dill`, we have this compatibility code:

```python
if sys.hexversion < 0x30E00A1:
    pickler._batch_setitems(iter(source.items()))
else:
    pickler._batch_setitems(iter(source.items()), obj=obj)
```

Thus, the datasets package will emit this error

```
│ /Users/sghuang/mamba/envs/ds/lib/python3.14/site-packages/dill/_dill.py:1262 in save_module_dict │
│                                                                                                  │
│   1259 │   │   if is_dill(pickler, child=False) and pickler._session:                            │
│   1260 │   │   │   # we only care about session the first pass thru                              │
│   1261 │   │   │   pickler._first_pass = False                                                   │
│ ❱ 1262 │   │   StockPickler.save_dict(pickler, obj)                                              │
│   1263 │   │   logger.trace(pickler, "# D2")                                                     │
│   1264 │   return                                                                                │
│   1265                                                                                           │
│                                                                                                  │
│ /Users/sghuang/mamba/envs/ds/lib/python3.14/pickle.py:1133 in save_dict                          │
│                                                                                                  │
│   1130 │   │   print(f"Line number: {inspect.getsourcelines(method)[1]}")                        │
│   1131 │   │   print(f"Full path: {inspect.getmodule(method)}")                                  │
│   1132 │   │   print(f"Class: {method.__qualname__}")                                            │
│ ❱ 1133 │   │   self._batch_setitems(obj.items(), obj)                                            │
│   1134 │                                                                                         │
│   1135 │   dispatch[dict] = save_dict                                                            │
│   1136                                                                                           │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
TypeError: Pickler._batch_setitems() takes 2 positional arguments but 3 were given
[NOTE] when serializing datasets.table.InMemoryTable state
[NOTE] when serializing datasets.table.InMemoryTable object
```

To fix it, we update the signature of the `_batch_setitems` method defined in `utils/_dill.py`.

This fix should be backward compatible, since the compatibility is handled by `dill`.

This should close #7813.

Similar to https://github.com/joblib/joblib/issues/1658.

[...]

PR #8411, opened 2026-08-06
Restore the _legacy_no_dict_keys_sorting branch in Pickler._batch_setitems

### LINKED ISSUE #8410: _legacy_no_dict_keys_sorting has no reader since #7817, so the 2.15.0 legacy cache lookup computes the wrong config_id
### Describe the bug

`Pickler._legacy_no_dict_keys_sorting` has a definition and a writer, but since #7817 it has no reader, so the flag does nothing.

`DatasetBuilder._check_legacy_cache2` patches it on to recompute the `config_id` that 2.15.0 would have produced, because 2.15.0 hashed dicts without sorting their items:

```python
# src/datasets/builder.py:493-495
with patch.object(Pickler, "_legacy_no_dict_keys_sorting", True):
    config_id = self.config.name + "-" + Hasher.hash({"data_files": self.config.data_files})
```

#7817 rewrote `_batch_setitems` to accept the extra argument Python 3.14 passes, and in the same hunk removed the early return that read the flag:

```diff
-    def _batch_setitems(self, items):
-        if self._legacy_no_dict_keys_sorting:
-            return super()._batch_setitems(items)
+    def _batch_setitems(self, items, *args, **kwargs):
         # Ignore the order of keys in a dict
```

That PR is about the 3.14 signature change and says nothing about the legacy branch, so the removal looks incidental rather than deliberate. At HEAD the flag appears exactly twice in the tree, and neither is a read:

```console
$ grep -rn '_legacy_no_dict_keys_sorting' src/
src/datasets/builder.py:494:            with patch.object(Pickler, "_legacy_no_dict_keys_sorting", True):
src/datasets/utils/_dill.py:30:    _legacy_no_dict_keys_sorting = False
```

`_check_legacy_cache2` builds `{cache_dir}/{namespace}___{dataset_name}/{config_id}/0.0.0/{hash}` from that `config_id` and returns it only `if os.path.isdir(legacy_cache_dir)`, so a `config_id` that no longer matches the one 2.15.0 wrote means the old directory is not found and the lookup silently misses.

`data_files` is a dict, and the two branches differ only when its keys are not already in sorted order, which is the common `{"train": ..., "test": ...}` case:

```python
from unittest.mock import patch

from datasets.utils._dill import Pickler

obj = {"train": ["train.csv"], "test": ["test.csv"]}

print("normal:", Hasher.hash(obj))
with patch.object(Pickler, "_legacy_no_dict_keys_sorting", True):
    print("legacy:", Hasher.hash(obj))
```

Run against several versions on Python 3.11, one venv each, the 2.15.0 column being what that release actually produces for the same object:

| datasets | `normal` | `legacy` | matches 2.15.0 |
|---|---|---|---|
| 2.15.0 | `711511d8f1d9bc25` | n/a | n/a |
| 4.3.0 | `cfd3b51f0f8e9fd8` | `711511d8f1d9bc25` | yes |
| 4.4.0 | `cfd3b51f0f8e9fd8` | `cfd3b51f0f8e9fd8` | no |
| main @ b7cb10b0 | `cfd3b51f0f8e9fd8` | `cfd3b51f0f8e9fd8` | no |

4.4.0 is the first release containing #7817 (`git tag --contains f7c8e46ec`), so the behaviour changes exactly at that boundary.

### Expected behavior

Under the patched flag, `Hasher.hash({"train": ..., "test": ...})` should reproduce the 2.15.0 value `711511d8f1d9bc25`, so `_check_legacy_cache2` looks for the directory 2.15.0 actually wrote. At HEAD it returns the sorted-key hash instead.

Two reasonable resolutions, and I do not know which you would prefer:

1. Restore the early return in `_batch_setitems`, keeping the `*args`/`**kwargs` passthrough that #7817 added.
2. Retire the compat path deliberately, dropping the flag and its use in `_check_legacy_cache2`, if the 2.14/2.15 cache layout is no longer worth carrying.

I have a PR ready for option 1, since it is the smaller change and preserves the stated intent of #6514. Happy to switch it to option 2 if you would rather retire the path.

One thing I did not test: I did not build a real 2.15.0 cache directory and load it under HEAD. The hash mismatch above is measured, and the path construction and `os.path.isdir` check are read from the source rather than exercised end to end.

### Environment info

- `datasets` 5.0.2.dev0 (main @ `b7cb10b0e38ab8bbefdb142d0279bc85ac8c7ab8`), plus 4.4.0, 4.3.0 and 2.15.0 from PyPI
- Python 3.11.15, `dill` 0.4.1, Linux x86_64, all runs in clean containers

ebarkhordar:
### Root cause

`Pickler._legacy_no_dict_keys_sorting` lost its only reader in #7817. That PR rewrote `_batch_setitems` to accept the argument Python 3.14 passes, and in the same hunk dropped the early return:

```diff
-    def _batch_setitems(self, items):
-        if self._legacy_no_dict_keys_sorting:
-            return super()._batch_setitems(items)
+    def _batch_setitems(self, items, *args, **kwargs):
```

`_check_legacy_cache2` still patches the flag on to rebuild the 2.15.0 `config_id`, so that lookup now hashes with sorted dict keys and no longer matches the directory 2.15.0 wrote. The flag appears twice in `src/`, as a definition and as that `patch.object` write, and nothing reads it.

### The fix

Restore the branch, keeping the `*args`/`**kwargs` passthrough #7817 added so the 3.14 fix is preserved on both paths.

### Verification

- The two new legacy tests fail on `main` (`assert [('a', 2), ('b', 1)] == [('b', 1), ('a', 2)]`) and pass with this change; the third pins the default order-insensitive behaviour and passes either way.
- `pytest tests/test_fingerprint.py`: 24 passed / 9 skipped before, 26 passed / 9 skipped after, in a clean container on Python 3.11.15 with `dill` 0.4.1.
- With the fix, `Hasher.hash({"train": ["train.csv"], "test": ["test.csv"]})` under the patched flag returns `711511d8f1d9bc25`, which is the value `datasets==2.15.0` actually produces for that object. On `main` it returns the sorted-key hash `cfd3b51f0f8e9fd8`. 4.4.0 is the first release where that regressed, matching `git tag --contains f7c8e46ec`.
- Not checked: I did not build a 2.15.0 cache directory and load it under HEAD. The hash mismatch is measured; the effect on the directory lookup is read from `_check_legacy_cache2`.

If you would rather retire the 2.14/2.15 compat path than restore it, that is the alternative I raised in the issue and I am happy to send that instead.

Fixes #8410

### DIFF
--- a/src/datasets/utils/_dill.py
+++ b/src/datasets/utils/_dill.py
@@ -71,6 +71,8 @@ class Pickler(dill.Pickler):
         dill.Pickler.save(self, obj, save_persistent_id=save_persistent_id)
 
     def _batch_setitems(self, items, *args, **kwargs):
+        if self._legacy_no_dict_keys_sorting:
+            return super()._batch_setitems(items, *args, **kwargs)
         # Ignore the order of keys in a dict
         try:
             # Faster, but fails for unorderable elements
--- a/tests/test_fingerprint.py
+++ b/tests/test_fingerprint.py
@@ -588,3 +588,44 @@ def test_dependency_on_dill():
     # AttributeError: module 'dill._dill' has no attribute 'stack'
     hasher = Hasher()
     hasher.update(lambda x: x)
+
+
+def test_hash_is_insensitive_to_dict_key_order():
+    # The default Pickler sorts dict items, so insertion order must not change the hash
+    assert Hasher.hash({"train": ["train.csv"], "test": ["test.csv"]}) == Hasher.hash(
+        {"test": ["test.csv"], "train": ["train.csv"]}
+    )
+
+
+def test_legacy_no_dict_keys_sorting_preserves_dict_key_order():
+    # `_check_legacy_cache2` patches this flag on to reproduce the 2.15.0 config_id,
+    # which was computed without sorting dict items
+    from datasets.utils._dill import Pickler
+
+    with patch.object(Pickler, "_legacy_no_dict_keys_sorting", True):
+        assert Hasher.hash({"train": ["train.csv"], "test": ["test.csv"]}) != Hasher.hash(
+            {"test": ["test.csv"], "train": ["train.csv"]}
+        )
+
+
+def test_legacy_no_dict_keys_sorting_forwards_extra_args():
+    # The legacy branch must forward *args/**kwargs like the sorting branch does,
+    # because Python 3.14 calls `_batch_setitems(items, obj=obj)`
+    import dill
+
+    from datasets.utils._dill import Pickler
+
+    calls = []
+
+    def fake_batch_setitems(self, items, *args, **kwargs):
+        calls.append((list(items), args, kwargs))
+
+    pickler = Pickler.__new__(Pickler)
+    with patch.object(dill.Pickler, "_batch_setitems", fake_batch_setitems):
+        with patch.object(Pickler, "_legacy_no_dict_keys_sorting", True):
+            pickler._batch_setitems(iter([("b", 1), ("a", 2)]), "positional", keyword=3)
+
+    items, args, kwargs = calls[0]
+    assert items == [("b", 1), ("a", 2)]  # unsorted -> the legacy branch ran
+    assert args == ("positional",)
+    assert kwargs == {"keyword": 3}


ebarkhordar:
No rush on this one. The offer at the bottom of the description still stands: if you would rather retire the 2.14/2.15 compat path than restore the branch, say so and I will send that instead.

ebarkhordar:
Closing this to keep my open PRs down. It has sat a month without CI ever being approved, so nobody has had much to go on.

If the 2.14/2.15 compat path is worth restoring, this is easy to reopen. The issue in #8410 describes the behaviour either way.

PR #8438, opened 2026-08-09
Validate rank and world_size in split_dataset_by_node for IterableDataset

LeSingh1:
`split_dataset_by_node()` behaves differently on the two dataset types for invalid `rank` / `world_size`. The map-style version delegates to `Dataset.shard()`, which rejects them:

```python
def _split_by_node_map_style_dataset(dataset, rank, world_size):
    return dataset.shard(num_shards=world_size, index=rank, contiguous=True)
```

The iterable version just stores them in a `DistributedConfig` without checking:

```python
ds = Dataset.from_dict({"x": range(10)}).to_iterable_dataset(num_shards=2)

len(list(split_dataset_by_node(ds, rank=-1, world_size=2)))   # 10 — the whole dataset
list(split_dataset_by_node(ds, rank=0, world_size=0))         # ZeroDivisionError: integer modulo by zero
list(split_dataset_by_node(ds, rank=5, world_size=2))         # IndexError: list index out of range
```

The `rank=-1` case is the one worth fixing: nothing raises, and the node gets **every** example instead of its shard. In a distributed run that silently duplicates the whole dataset on that node rather than failing. The same calls on a map-style dataset all raise `ValueError` immediately.

## Fix

Check `0 <= rank < world_size` in `_split_by_node_iterable_dataset()`, before the `dataset._distributed` nesting combines the values, so nested calls are validated on the arguments the caller actually passed.

Only the iterable path is touched — the map-style path keeps raising from `Dataset.shard()` exactly as before.

## Test

`test_split_dataset_by_node_iterable_invalid_rank`, parametrized over `(0, 0)`, `(-1, 2)`, `(2, 2)` and `(5, 2)`. All four fail on `main`.

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -5389,6 +5389,8 @@ def _split_by_node_iterable_dataset(dataset: IterableDataset, rank: int, world_s
     Returns:
         [`IterableDataset`]: The iterable dataset to be used on the node at rank `rank`.
     """
+    if not 0 <= rank < world_size:
+        raise ValueError("rank should be in [0, world_size-1]")
     if dataset._distributed:
         rank = world_size * dataset._distributed.rank + rank
         world_size = world_size * dataset._distributed.world_size
--- a/tests/test_distributed.py
+++ b/tests/test_distributed.py
@@ -137,3 +137,12 @@ def test_torch_distributed_run_streaming_with_num_workers(nproc_per_node, num_wo
     """.split()
     cmd = [sys.executable] + distributed_args + args
     execute_subprocess_async(cmd, env=os.environ.copy())
+
+
+@pytest.mark.parametrize("rank, world_size", [(0, 0), (-1, 2), (2, 2), (5, 2)])
+def test_split_dataset_by_node_iterable_invalid_rank(rank, world_size):
+    # invalid ranks used to be accepted: rank=-1 silently returned the whole dataset,
+    # while the map-style version raises for all of them
+    full_ds = Dataset.from_dict({"i": range(17)}).to_iterable_dataset(num_shards=2)
+    with pytest.raises(ValueError, match=r"rank should be in \[0, world_size-1\]"):
+        split_dataset_by_node(full_ds, rank=rank, world_size=world_size)


shashvat-singham (commented):
Checked all four parametrized cases plus a negative `world_size`, against the map-style path each time. The gap was real and the fix closes it:

```
                    eager                                     iterable (this PR)
rank= 0 ws= 0   ValueError: index should be in [0, num_shards-1]   ValueError: rank should be in [0, world_size-1]
rank=-1 ws= 2   ValueError: index should be in [0, num_shards-1]   ValueError: rank should be in [0, world_size-1]
rank= 2 ws= 2   ValueError: index should be in [0, num_shards-1]   ValueError: rank should be in [0, world_size-1]
rank= 5 ws= 2   ValueError: index should be in [0, num_shards-1]   ValueError: rank should be in [0, world_size-1]
rank= 0 ws=-1   ValueError: index should be in [0, num_shards-1]   ValueError: rank should be in [0, world_size-1]
```

Both paths now raise for every invalid combination, including `world_size=-1`, which is not in the parametrize list but falls out of the same `0 <= rank < world_size` check — worth adding as a case since it is the one where the bound is nonsensical rather than merely out of range.

`rank=-1` silently returning the whole dataset was the dangerous one: under DDP that means every node quietly processes every example, so a training run completes with `world_size`× the intended data and no error anywhere. Good catch.

Two notes, neither blocking:

**The messages differ between the two paths.** Eager says `index should be in [0, num_shards-1]`, this says `rank should be in [0, world_size-1]`. Yours is the better message here — the caller passed `rank` and `world_size`, so naming those is more useful than exposing the internal `shard`/`num_shards` vocabulary. Since the point of the PR is bringing the two paths into agreement, it may be worth saying in the description that the wording deliberately does *not* match, so a reviewer does not read it as an oversight.

**`world_size=0` produces a strange bound.** The message interpolates to "rank should be in [0, world_size-1]" literally, but for `world_size=0` the described range is `[0, -1]`, which is empty. Fine as a constant string; would only matter if the values were interpolated in. Mentioning in case you wanted to add a separate `world_size` check with its own message — `world_size` being 0 or negative is arguably a distinct error from `rank` being out of range.

Verified on Windows 11 / Python 3.11.9, `pr-8438`.

PR #8442, opened 2026-08-09
Accept numpy integers as train_size / test_size in train_test_split

LeSingh1:
`Dataset.train_test_split()` uses `isinstance(..., int)` / `isinstance(..., float)` to tell "a number of rows" from "a fraction", and rejects everything else:

```python
if train_size is not None and not isinstance(train_size, (int, float)):
    raise ValueError(f"Invalid value for train_size: {train_size} of type {type(train_size)}")
```

numpy **floats** are subclasses of `float`, so they already work. numpy **integers** are not subclasses of `int`, so they don't:

```python
import numpy as np

ds.train_test_split(test_size=np.float64(0.2))   # works
ds.train_test_split(test_size=np.int64(2))
# ValueError: Invalid value for test_size: 2 of type <class 'numpy.int64'>
```

The asymmetry is accidental, and the error is confusing because the printed value (`2`) looks perfectly valid. numpy integers turn up naturally — `np.floor(0.2 * len(ds)).astype(int)`, a value read out of a numpy array, a pandas `.nunique()`, etc. `sklearn.model_selection.train_test_split`, which this API is modelled on, accepts them.

This PR converts numpy integers to `int` before the checks, leaving the int/float semantics untouched.

Added `tests/test_arrow_dataset.py::test_train_test_split_with_numpy_integer_sizes`, parametrized over `int32`, `int64` and `uint8`, covering both `test_size` and `train_size`. All three fail on `main`.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -5086,6 +5086,13 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         if test_size is None and train_size is None:
             test_size = 0.25
 
+        # numpy integers are not instances of `int` (contrary to numpy floats, which are instances of `float`),
+        # so they wouldn't pass the checks below
+        if isinstance(test_size, np.integer):
+            test_size = int(test_size)
+        if isinstance(train_size, np.integer):
+            train_size = int(train_size)
+
         # Safety checks similar to scikit-learn's ones.
         # (adapted from https://github.com/scikit-learn/scikit-learn/blob/fd237278e895b42abe8d8d09105cbb82dc2cbba7/sklearn/model_selection/_split.py#L1750)
         n_samples = len(self)
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -5086,12 +5086,17 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         if test_size is None and train_size is None:
             test_size = 0.25
 
-        # numpy integers are not instances of `int` (contrary to numpy floats, which are instances of `float`),
-        # so they wouldn't pass the checks below
+        # A numpy scalar is not reliably an instance of the matching Python builtin:
+        # `np.float64` happens to subclass `float`, but `np.float32` does not and no
+        # numpy integer subclasses `int`, so they would fail the checks below.
         if isinstance(test_size, np.integer):
             test_size = int(test_size)
         if isinstance(train_size, np.integer):
             train_size = int(train_size)
+        if isinstance(test_size, np.floating):
+            test_size = float(test_size)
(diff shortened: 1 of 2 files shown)


shashvat-singham (commented):
The integer fix works — `np.int64(2)` and `np.int32(2)` both go from `ValueError` to a correct split on this branch.

But the comment justifying the scope is not right, and it hides a remaining case:

```python
# numpy integers are not instances of `int` (contrary to numpy floats, which are instances of `float`),
# so they wouldn't pass the checks below
```

Only `np.float64` is a `float` subclass. `np.float32` and `np.float16` are not:

```
float64  isinstance(float)=True
float32  isinstance(float)=False
float16  isinstance(float)=False
int64    isinstance(float)=False
```

So the same bug survives for the narrower float types. Measured on this branch:

```
np.int64(2)      -> test=2 train=8
np.int32(2)      -> test=2 train=8
np.float64(0.2)  -> test=2 train=8
np.float32(0.2)  -> ValueError: Invalid value for test_size: 0.20000000298023224 of type <class 'numpy.float32'>
np.float16(0.2)  -> ValueError: Invalid value for test_size: 0.199951171875 of type <class 'numpy.float16'>
```

`np.float32` is not exotic — it is what you get from a `float32` column, from most of the array libraries, and from anything that has been through a torch/JAX tensor. A ratio computed that way hitting "Invalid value for test_size" is the same papercut this PR is fixing for ints.

Symmetric fix, right next to the existing one:

```python
if isinstance(test_size, np.floating):
    test_size = float(test_size)
if isinstance(train_size, np.floating):
    train_size = float(train_size)
```

`np.floating` covers `float16`/`float32`/`float64` and leaves `np.float64` behaving exactly as it does today, so nothing regresses. Then the comment can say what is actually true: numpy scalars are not reliably instances of the corresponding Python builtin — `np.float64` happens to be, `np.float32` is not, and no numpy integer is.

Worth extending the test the same way, since `["int32", "int64", "uint8"]` is already parametrized and adding the float types is nearly free.

Verified on Windows 11 / Python 3.11.9 / numpy 2.4.6, `main` vs `pr-8442`.

LeSingh1:
Good catch, thank you — added in 4be591467, and you were right that the comment was wrong. `np.float64` subclasses `float` but `np.float32` and `np.float16` do not, so the comment now says that instead of implying all numpy floats do.

One wrinkle worth flagging, since it changed how I wrote the test. A narrow float does not hold the ratio exactly: `np.float32(0.2)` is `0.20000000298...`, so on a 10-row dataset it legitimately splits 3/7 where `0.2` splits 2/8. That is correct behaviour rather than something to paper over, so the test asserts the numpy scalar matches `float(value)` rather than a fixed row count. Parametrized over float16/32/64 with two ratios, and all six cases agree with their Python-float equivalent.

Thanks also for the verification tables on the other five — the `main` vs branch comparison is genuinely useful.

PR #8443, opened 2026-08-09
Validate by_column and fix the error message in Dataset.batch

LeSingh1:
Two small problems in `Dataset.batch()`.

**1. The error names the wrong class.**

```python
ds.batch()
# ValueError: IterableDataset.batch() misses `batch_size` or `by_column` argument.
```

`IterableDataset.batch()` has its own copy of the same check, so this occurrence in `arrow_dataset.py` is just a copy-paste that sends the reader to the wrong class.

**2. `by_column` is not validated.**

It is passed straight through to `_batch_accumulate_arrow_table_by_columns`, so a typo only fails once pyarrow tries to read the field, inside a `map` worker:

```python
ds.batch(by_column="labels")
# KeyError: 'Field "labels" does not exist in schema'
```

Every other column-taking method (`map(input_columns=...)`, `map(remove_columns=...)`, `sort`, `select_columns`, `remove_columns`, ...) checks up front and lists the available columns. This PR does the same for `by_column`, for both the `str` and the `list[str]` form.

### Tests

`test_dataset_batch_without_batch_size_or_by_column` (asserts the message starts with `Dataset.batch()` — note a loose `match="Dataset.batch"` would also match `IterableDataset.batch`) and `test_dataset_batch_by_missing_column`, parametrized over a single column name and a list. All three fail on `main`.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -4111,11 +4111,17 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         ```
         """
         if batch_size is None and by_column is None:
-            raise ValueError("IterableDataset.batch() misses `batch_size` or `by_column` argument.")
+            raise ValueError("Dataset.batch() misses `batch_size` or `by_column` argument.")
         if by_column is not None:
             if num_proc:
                 raise NotImplementedError("Multiprocessed batching by column is not implemented yet")
             columns = [by_column] if isinstance(by_column, str) else by_column
+            missing_columns = [column for column in columns if column not in self._data.column_names]
+            if missing_columns:
+                raise ValueError(
+                    f"Column to batch by {missing_columns} not in the dataset. "
+                    f"Current columns in the dataset: {self._data.column_names}"
+                )
             _batch_fn = partial(
                 _batch_accumulate_arrow_table_by_columns, columns=columns, tables_accumulator=[], length=len(self)
             )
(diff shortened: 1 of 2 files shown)


shashvat-singham (commented):
Both halves check out against `main` @ 48b7ee7.

The message really was wrong. `Dataset.batch` raises it at `arrow_dataset.py:4114` and names the wrong class:

```python
raise ValueError("IterableDataset.batch() misses `batch_size` or `by_column` argument.")
```

and the genuine `IterableDataset.batch` copy is at `iterable_dataset.py:4450` with the same text — so this was a copy-paste into `arrow_dataset.py`, and fixing only the `Dataset` side is right. Both now name the class you actually called.

The missing-column check is a real improvement too: `by_column="col_2"` on a dataset without `col_2` previously failed somewhere downstream rather than at the argument. Listing the available columns in the message matches what `_check_column_names` and friends already do.

One suggestion on the test:

```python
with pytest.raises(ValueError, match=r"^Dataset\.batch\(\)"):
```

Anchoring on `^Dataset` is what makes this a regression test for the rename rather than a generic "it raises" test, which is good — but it will also pass if the message degrades to just `"Dataset.batch()"` with the rest dropped. `match=r"^Dataset\.batch\(\) misses"` costs nothing and pins the sentence.

Worth knowing while you are in this method: `batch_size` itself is still unvalidated, so `ds.batch(0)` returns a single batch containing the entire dataset rather than raising (#8472). Your #8446 adds a `_check_batch_size` helper that would cover it. Not this PR's job — just noting the three PRs all touch the same argument and it would be easy to leave the gap behind.

Verified on Windows 11 / Python 3.11.9.

PR #8444, opened 2026-08-09
Fix StopIteration when the JSON data files are empty

LeSingh1:
Feature inference in `json.py` did `next(iter(...))` catching only `FullReadDisallowed`, so an empty `.jsonl` leaked a bare `StopIteration` out of `load_dataset` (the PEP 479 hazard).

This brings `json.py` in line with its neighbours rather than adding a special case: `arrow.py` and `parquet.py` already infer with a `for file in files: ... break` loop, which naturally leaves `info.features` as `None` when there is nothing to read. `json.py` was the odd one out.

**Scope — the non-streaming path still raises.** Only the streaming load is fixed:

```
load_dataset("json", data_files=empty, split="train", streaming=True)   -> [] (fixed here)
load_dataset("json", data_files=empty, split="train")                   -> SchemaInferenceError
```

That is not a regression (on `main` the same call dies earlier with the bare `StopIteration`), and it is the behaviour `text` already has — you genuinely cannot infer a schema from zero examples, and `SchemaInferenceError` says so clearly. After this PR `json` matches `text` on both axes:

| loader | streaming=True | streaming=False |
|---|---|---|
| json (this PR) | ok, 0 rows | SchemaInferenceError |
| text | ok, 0 rows | SchemaInferenceError |
| csv | EmptyDataError | EmptyDataError |

Test covers the streaming case, which is the one whose behaviour changes.

### DIFF
--- a/src/datasets/packaged_modules/json/json.py
+++ b/src/datasets/packaged_modules/json/json.py
@@ -98,10 +98,16 @@ class Json(datasets.ArrowBasedBuilder):
             )
         if self.info.features is None:
             try:
-                pa_table = next(iter(self._generate_tables(**splits[0].gen_kwargs, allow_full_read=False)))[1]
-                self.info.features = datasets.Features.from_arrow_schema(pa_table.schema)
-                if self.config.parse_agent_traces and has_agent_traces_markers(self.info.features):
-                    self.info.features = AGENT_TRACES_FEATURES
+                # the generator is empty when the data files contain no data at all,
+                # in which case the features stay unknown instead of raising StopIteration
+                first_key_and_table = next(
+                    iter(self._generate_tables(**splits[0].gen_kwargs, allow_full_read=False)), None
+                )
+                if first_key_and_table is not None:
+                    pa_table = first_key_and_table[1]
+                    self.info.features = datasets.Features.from_arrow_schema(pa_table.schema)
+                    if self.config.parse_agent_traces and has_agent_traces_markers(self.info.features):
+                        self.info.features = AGENT_TRACES_FEATURES
             except FullReadDisallowed:
                 pass
         return splits
--- a/tests/packaged_modules/test_json.py
+++ b/tests/packaged_modules/test_json.py
@@ -7,6 +7,7 @@ import pytest
 from datasets import Features, Value, load_dataset
 from datasets.builder import InvalidConfigName
 from datasets.data_files import DataFilesList
+from datasets.download.streaming_download_manager import StreamingDownloadManager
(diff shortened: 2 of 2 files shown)


shashvat-singham (commented):
Ran the branch — both new tests pass, and the fix does what it says for the streaming path.

Two things I checked that might be worth folding in.

**It matches the convention the other loaders already use.** `arrow.py` and `parquet.py` do their schema inference with a `for file in files: ... break` loop, which naturally leaves `info.features` as `None` when there's nothing to read. `json.py`'s `next(iter(...))` was the odd one out, so this brings it in line rather than adding a special case. Might be worth a word in the description, since "use the default arg of `next()`" reads as a band-aid until you notice the neighbours already behave this way.

**The non-streaming path still fails, and the PR reads as if it wouldn't.** Only the streaming case is covered by a test:

```python
>>> load_dataset("json", data_files=empty, split="train", streaming=True)
[]                                        # fixed by this PR
>>> load_dataset("json", data_files=empty, split="train")
DatasetGenerationError: ... caused by
SchemaInferenceError: Please pass `features` or at least one example when writing data
```

That's not a regression — on `main` it dies earlier with the bare `StopIteration` — so this PR strictly improves it. But someone reading "Fix StopIteration when the JSON data files are empty" will reasonably expect the non-streaming load to work too, and it doesn't.

For what it's worth I think the current behaviour is *right*, because it's exactly what `text` already does:

```
json  streaming=True  -> ok, 0 rows            (after this PR)
json  streaming=False -> SchemaInferenceError
text  streaming=True  -> ok, 0 rows
text  streaming=False -> SchemaInferenceError
csv   streaming=True  -> EmptyDataError
csv   streaming=False -> EmptyDataError
```

So json now lines up with `text` on both axes, which seems like the outcome you'd want — you genuinely can't infer a schema from zero examples, and `SchemaInferenceError` says so clearly instead of leaking `StopIteration`.

Given that, I'd suggest adding a non-streaming test asserting `SchemaInferenceError` (and maybe that passing explicit `features=` makes it succeed with 0 rows). It pins the intended behaviour, and stops the next person filing "empty JSON still fails" as a fresh bug.

LeSingh1:
Both good points — description rewritten.

You are right that `next(..., default)` reads as a band-aid on its own; noting that `arrow.py` and `parquet.py` already infer with a `for ... break` loop makes it clear this brings `json.py` in line rather than special-casing it.

And the scope point is fair: the title promised more than the diff delivers. The description now states plainly that the non-streaming path still raises `SchemaInferenceError`, with your table showing `json` ending up matching `text` on both axes. I agree that is the right outcome rather than something to fix — you cannot infer a schema from zero examples, and that error says so.

PR #8446, opened 2026-08-09
Reject a non-positive batch_size in the Dataset export methods

LeSingh1:
All the `Dataset.to_*` exports batch with `range(0, len(self), batch_size)`. A negative `batch_size` makes that range empty, so the data is silently dropped — including when writing to a file:

```python
from datasets import Dataset

ds = Dataset.from_dict({"col_1": [0, 1, 2]})

len(list(ds.to_dict(batch_size=-1, batched=True)))     # 0
len(list(ds.to_pandas(batch_size=-1, batched=True)))   # 0

ds.to_csv("out.csv", batch_size=-1)                    # 0   -> out.csv is empty
ds.to_json("out.json", batch_size=-1)                  # 0   -> out.json is empty
ds.to_parquet("out.parquet", batch_size=-1)            # 0
ds.to_sql("t", con, batch_size=-1)                     # 0
```

A "successful" export that writes an empty file is about the worst failure mode available here — the return value is the number of bytes written, and `0` is easy to ignore.

`batch_size=0` is not caught either; it is falsy, so it is silently replaced by `config.DEFAULT_MAX_BATCH_SIZE`.

This mistake is easy to make because `map`, `filter` and `batch` all document `batch_size <= 0` (or `None`) as *"provide the full dataset as a single batch"*. Someone carrying that idiom over to `to_csv` gets an empty file instead.

This PR adds a small `_check_batch_size` helper next to the other `_check_*` helpers and calls it in `to_dict`, `to_pandas`, `to_polars`, `to_csv`, `to_json`, `to_parquet` and `to_sql`, raising the same `ValueError` that `Dataset.iter` raises. `batch_size=None` keeps meaning "use the default".

Note this also turns `batch_size=0` into an error rather than a silent fallback to the default; happy to relax it to `< 0` only if you'd rather keep that.

Added `tests/test_arrow_dataset.py::test_dataset_export_with_non_positive_batch_size`, parametrized over `0` and `-1`; both fail on `main`.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -633,6 +633,12 @@ def _check_column_names(column_names: list[str]):
         raise ValueError(f"The table can't have duplicated columns but columns {duplicated_columns} are duplicated.")
 
 
+def _check_batch_size(batch_size: Optional[int]):
+    """Check the `batch_size` of the methods that export a dataset by batches."""
+    if batch_size is not None and batch_size <= 0:
+        raise ValueError(f"batch_size must be a positive integer, but got {batch_size}.")
+
+
 def _check_valid_indices_value(index, size):
     if (index < 0 and index + size < 0) or (index >= size):
         raise IndexError(f"Index {index} out of range for dataset of size {size}.")
@@ -5371,6 +5377,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         # Dynamic import to avoid circular dependency
         from .io.csv import CsvDatasetWriter
 
+        _check_batch_size(batch_size)
         return CsvDatasetWriter(
             self,
             path_or_buf,
@@ -5414,6 +5421,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                 result[col] = [json_decode_field(row, json_field_subpath) for row in result[col]]
             return result
 
+        _check_batch_size(batch_size)
         if not batched:
             return query_to_dict(slice(0, len(self)))
         else:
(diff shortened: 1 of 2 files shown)


shashvat-singham (commented):
Checked the coverage: all seven `Dataset` export methods that take a `batch_size` get the new check — `to_csv`, `to_dict`, `to_json`, `to_pandas`, `to_polars`, `to_parquet`, `to_sql`. Nothing missed on that list.

Two gaps worth deciding on before this lands, because they leave `batch_size=0` doing three different things depending on which method you call.

**1. `Dataset.batch()` is not covered**, and it is the worst offender:

```python
ds = Dataset.from_dict({"a": list(range(5))})
len(ds.batch(0))    # 1  -- one batch holding everything
len(ds.batch(-1))   # 1
```

Zero-sized batch requested, largest possible batch returned. `_check_batch_size` is exactly the helper it needs, so it is a one-line addition here rather than a separate change later. I filed #8472 with the details.

**2. The `IterableDataset` exports are not covered, and #8451 pushes them the other way.**

#8451 (yours) adds this to `IterableDataset.to_dict`/`to_pandas`/`to_polars`:

```python
batch_size = batch_size if batch_size else config.DEFAULT_MAX_BATCH_SIZE
```

`0` is falsy, so it becomes `1000`. If both PRs land as written:

| call | after #8446 | after #8451 |
|---|---|---|
| `Dataset.to_pandas(batch_size=0)` | `ValueError` | — |
| `IterableDataset.to_pandas(batch_size=0)` | — | silently `1000` |

That is a new eager/streaming divergence, in the same methods #8451 exists to bring back into agreement. Changing that guard to `is not None` (which I also mentioned on #8451) lets the `0` through so it can be rejected, and then wiring `_check_batch_size` into the `IterableDataset` exports too would line all of them up.

Nothing wrong with the code in this PR as it stands — it is a scoping question, and the answer may well be "separate PR". Flagging it because the three PRs are in flight at once and the interaction is easy to miss.

Verified on Windows 11 / Python 3.11.9 against `main` @ 48b7ee7.

PR #8451, opened 2026-08-09
Use the documented default batch_size in the batched IterableDataset exports

LeSingh1:
`IterableDataset.to_dict()`, `to_pandas()` and `to_polars()` all document `batch_size` as:

> batch_size (`int`, *optional*): The size (number of rows) of the batches if `batched` is `True`. Defaults to `datasets.config.DEFAULT_MAX_BATCH_SIZE`.

but the default is never applied. `batch_size` (`None`) is passed straight through to `iter()`, where it means "don't rebatch", so the whole dataset comes back as a single batch:

```python
ds = Dataset.from_dict({"col": range(2500)})

[len(df) for df in ds.to_pandas(batched=True)]
# [1000, 1000, 500]           <- Dataset applies the default

[len(df) for df in ds.to_iterable_dataset().to_pandas(batched=True)]
# [2500]                      <- IterableDataset returns everything at once
```

Same for `to_dict(batched=True)` and `to_polars(batched=True)`.

This defeats the purpose of `batched=True` on the streaming type specifically: the reason to ask for batches is to avoid holding the dataset in memory, and here `to_pandas(batched=True)` builds one DataFrame over the entire stream. Passing `batch_size` explicitly works fine — only the documented default is missing.

## Fix

Apply `config.DEFAULT_MAX_BATCH_SIZE` when `batch_size` is not set, using the same idiom `Dataset.to_pandas()` already uses:

```python
batch_size = batch_size if batch_size else config.DEFAULT_MAX_BATCH_SIZE
```

It is placed before the `if batched:` branch; the non-batched branches hardcode their own chunk size and are unaffected.

## Test

`test_iterable_dataset_batched_export_default_batch_size` builds a dataset of `2 * DEFAULT_MAX_BATCH_SIZE + 1` rows and asserts `Dataset` and `IterableDataset` produce the same batch sizes from `to_pandas(batched=True)` and `to_dict(batched=True)`. It fails on `main`, where the iterable version yields one batch.

Note: #8382 fixes a separate bug in the same `batched=True` branches of `to_dict` / `to_polars` (a bare `yield` making them return an empty generator), and #8434 fixes the non-batched branches crashing on an empty dataset. The three are complementary.

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -4500,7 +4500,7 @@ class IterableDataset(DatasetInfoMixin):
         >>> ds.to_dict()
         ```
         """
-        batch_size = batch_size if batch_size else config.DEFAULT_MAX_BATCH_SIZE
+        batch_size = batch_size if batch_size is not None else config.DEFAULT_MAX_BATCH_SIZE
         if batched:
             for table in self.with_format("arrow").iter(batch_size=batch_size):
                 yield Dataset(table, fingerprint="unset").to_dict()
@@ -4552,7 +4552,7 @@ class IterableDataset(DatasetInfoMixin):
                 return cast_table_to_features(table, self.features)
             return table
 
-        batch_size = batch_size if batch_size else config.DEFAULT_MAX_BATCH_SIZE
+        batch_size = batch_size if batch_size is not None else config.DEFAULT_MAX_BATCH_SIZE
         if batched:
             return (
                 Dataset(maybe_cast_to_declared_features(table), info=info, fingerprint="unset").to_pandas()
@@ -4594,7 +4594,7 @@ class IterableDataset(DatasetInfoMixin):
         >>> ds.to_polars()
         ```
         """
-        batch_size = batch_size if batch_size else config.DEFAULT_MAX_BATCH_SIZE
+        batch_size = batch_size if batch_size is not None else config.DEFAULT_MAX_BATCH_SIZE
         if batched:
             for table in self.with_format("arrow").iter(batch_size=batch_size):
                 yield Dataset(table, fingerprint="unset").to_polars(schema_overrides=schema_overrides, rechunk=rechunk)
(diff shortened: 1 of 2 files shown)


shashvat-singham (commented):
Confirmed the bug and the fix on this branch — `to_dict`, `to_pandas` and `to_polars` now batch at `DEFAULT_MAX_BATCH_SIZE` and match `Dataset`:

```
DEFAULT_MAX_BATCH_SIZE = 1000
batched=True, no batch_size : [1000, 1000, 500]
batched=True, batch_size=500: [500, 500, 500, 500, 500]
```

One thing to fix: the guard is truthiness, not a `None` check.

```python
batch_size = batch_size if batch_size else config.DEFAULT_MAX_BATCH_SIZE
```

`0` is falsy, so an explicit `batch_size=0` is silently replaced by `1000`:

```
batched=True, batch_size=0  : [1000, 1000, 500]
```

Asking for a zero-sized batch is a caller bug either way, but silently substituting a default is the least helpful of the available responses — it looks like it worked and quietly ignores what was asked for. `batch_size if batch_size is not None else config.DEFAULT_MAX_BATCH_SIZE` keeps `0` flowing through to whatever the intended error is.

That is not academic here: your own #8446 adds "reject a non-positive `batch_size`" for the `Dataset` export methods. As written, this PR makes the `IterableDataset` versions unable to ever see a `0` to reject, so the two PRs would leave the eager and streaming paths disagreeing again — in the opposite direction from the one this PR is fixing. Worth lining them up.

Same three call sites, same one-word change.

Verified on Windows 11 / Python 3.11.9.

LeSingh1:
You are right, and the conflict with #8446 is the important part — thanks for spotting it. Fixed in e8509c1e1: all three call sites now use `batch_size if batch_size is not None else config.DEFAULT_MAX_BATCH_SIZE`, so a `0` reaches whatever validation exists instead of being quietly swapped for the default.

Verified: `None` still defaults to 1000, and `0` now flows through rather than being swallowed. That leaves the streaming methods able to reject it once #8446 lands, so the two PRs point the same way instead of opposite ones.

PR #8455, opened 2026-08-09
close input stream in _in_memory_arrow_table_from_file to prevent file descriptor leaks

HassanRady:
In `src/datasets/table.py`, `_in_memory_arrow_table_from_file` opens a `pa.input_stream(filename)` without a `with` context manager block. 
When exceptions occur during stream parsing or when exception tracebacks retain frame variables in memory (e.g., in logging handlers or data loader error queues), local `in_memory_stream` variables are held alive by stack frame references. This prevents CPython reference counting from closing the underlying OS file descriptor, resulting in open file descriptor leaks (`OSError: [Errno 24] Too many open files`) during multi-shard dataset processing.
This PR wraps `pa.input_stream` in a `with` context manager and executes `opened_stream.read_all()` inside the block to guarantee immediate file descriptor closure upon reading.

### DIFF
--- a/src/datasets/table.py
+++ b/src/datasets/table.py
@@ -31,9 +31,9 @@ def inject_arrow_table_documentation(arrow_table_method):
 
 
 def _in_memory_arrow_table_from_file(filename: str) -> pa.Table:
-    in_memory_stream = pa.input_stream(filename)
-    opened_stream = pa.ipc.open_stream(in_memory_stream)
-    pa_table = opened_stream.read_all()
+    with pa.input_stream(filename) as in_memory_stream:
+        opened_stream = pa.ipc.open_stream(in_memory_stream)
+        pa_table = opened_stream.read_all()
     return pa_table
 
 


shashvat-singham (commented):
I went looking for a case where this actually changes behaviour, because on CPython refcounting closes `in_memory_stream` as soon as the function returns or raises, which makes the change look cosmetic. It is not — there is a real leak, it just needs the traceback to survive.

**Happy path and plain exception: no difference.** Both `main` and this branch release the handle immediately, on success and when `open_stream` raises on a corrupt file.

**Retained traceback: `main` leaks, this branch does not.** An exception object keeps its traceback, which keeps the frame, which keeps `in_memory_stream` alive — so any caller that collects errors rather than letting them propagate holds the file open:

```python
saved = []
try:
    _in_memory_arrow_table_from_file(corrupt_path)
except Exception as e:
    saved.append(e)          # traceback -> frame -> in_memory_stream still referenced

os.remove(corrupt_path)
```

```
main     : handle released while traceback retained: NO (leaked)
pr-8455  : handle released while traceback retained: yes
```

On Windows that surfaces as `PermissionError` on the unlink, which is how I detected it; on Linux it is a silent fd leak instead, and it accumulates across shards. So this is worth having, and that scenario is probably worth a line in the PR description since it is the case that justifies the change.

I also confirmed the table stays fully valid after the stream closes — `read_all()` materialises into memory buffers, so there is no zero-copy reference back into the file. Checked `validate(full=True)`, `to_pylist()`, and reading values after `os.remove` of the source file: all fine on both branches. That was my main worry with wrapping this in a `with`, and it is unfounded.

One thing left on the table: `opened_stream` is still not closed. `pa.ipc.open_stream` returns a `RecordBatchStreamReader`, which is also a context manager, so

```python
with pa.input_stream(filename) as in_memory_stream, pa.ipc.open_stream(in_memory_stream) as opened_stream:
    pa_table = opened_stream.read_all()
```

would close both deterministically. Minor, since closing the underlying stream is what releases the OS handle, but it is the same argument applied one level up.

Tested on Windows 11 / Python 3.11.9 / pyarrow 25.0.1, `main` vs `pr-8455`.

PR #8456, opened 2026-08-09
Fix Dataset.map(batched=True, drop_last_batch=True) returning unchanged input when dataset is smaller than batch_size

### LINKED ISSUE #8386: map(batched=True, drop_last_batch=True) returns the input unchanged when the dataset is smaller than batch_size
### Describe the bug

When `drop_last_batch=True` and the dataset is smaller than `batch_size`, `Dataset.map(batched=True, ...)` returns the **input unchanged** instead of an empty dataset. The mapped function is never called, but the original rows are passed through with the original (unbatched) schema.

The iterable path and `Dataset.iter` both correctly yield nothing, so the map-style and streaming paths disagree.

```python
from datasets import Dataset

out = ds.map(lambda b: b, batched=True, batch_size=5, drop_last_batch=True)
print(len(out), out.features)
# 3 {'a': Value('int64')}          <- expected 0 rows

print(len(ds.batch(5, drop_last_batch=True)))
# 3                                <- expected 0 rows

print(sum(1 for _ in ds.to_iterable_dataset().batch(5, drop_last_batch=True)))
# 0                                <- correct

print(sum(1 for _ in ds.iter(5, drop_last_batch=True)))
# 0                                <- correct
```

Note the progress bar reports `Map: 0 examples` while the result still has 3 rows, which confirms the function was never invoked yet the data was passed through.

`Dataset.batch()` is implemented on top of `map(batched=True)`, so it inherits the same behavior — and its result is doubly wrong, since the rows come back *unbatched* (`Value('int64')` rather than `List(Value('int64'))`).

### Expected behavior

`drop_last_batch=True` means an incomplete final batch is dropped. When the whole dataset is smaller than one batch, every batch is incomplete, so the result should be empty — matching `IterableDataset.batch` and `Dataset.iter`.

### Root cause

In `_map_single`, `update_data` stays `None` because the function is never invoked for an incomplete batch. The writer is therefore never initialized and the code falls through to `yield rank, True, shard`, returning the original shard.

### Open question for maintainers

The right output *schema* is genuinely ambiguous when the function never runs, which is why I'm filing this rather than sending a patch directly. The options I see:

1. Return an empty dataset carrying the **input** features (simple, but wrong for `batch()`, where the output schema should be the batched `List(...)` form).
2. Initialize the writer eagerly from the function's declared/inferred output features.
3. Raise a clear error when `drop_last_batch=True` would drop everything.

Happy to open a PR once there's a preference — I have a fix for (1) working locally, and (2) is doable but a larger change.

### Environment info

- Python 3.12.13
- pyarrow 25.0.0

mayuriphad:
Fixes #8386

## Bug

When `drop_last_batch=True` and the dataset has fewer rows than `batch_size`, `Dataset.map(batched=True, ...)` was returning the **input dataset unchanged** instead of an empty dataset. Every batch is incomplete in this situation, so the mapped function is never invoked and `update_data` stays falsy; `_map_single` then fell through to `yield rank, True, shard`, silently passing the original, unprocessed rows straight through.

This disagreed with `IterableDataset.batch()` and `Dataset.iter()`, which both correctly yield zero rows when every batch is incomplete, and it also affected `Dataset.batch()` since it is implemented on top of `map(batched=True)`.

```python
from datasets import Dataset

out = ds.map(lambda b: b, batched=True, batch_size=5, drop_last_batch=True)
print(len(out))  # 3 before the fix, expected 0

print(len(ds.batch(5, drop_last_batch=True)))  # 3 before the fix, expected 0
```

## Fix

In `Dataset._map_single`, when `batched=True`, `drop_last_batch=True`, and the shard has fewer rows than `batch_size` (so the function was never called and `update_data` is still falsy), return an empty dataset (`shard.select([])`) carrying the input schema instead of the untouched `shard`. This matches option (1) discussed in the issue, and aligns the eager `map`/`batch` behavior with the streaming `IterableDataset.batch`/`Dataset.iter` behavior.

## Test

Added `test_map_batched_drop_last_batch_smaller_than_batch_size` in `tests/test_arrow_dataset.py`, which fails on `main` (`AssertionError: 30 != 0`) and passes with this fix.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -4056,6 +4056,11 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                 yield rank, True, Dataset.from_file(cache_file_name, info=info, split=shard.split)
             else:
                 yield rank, True, Dataset.from_buffer(buf_writer.getvalue(), info=info, split=shard.split)
+        elif batched and drop_last_batch and len(shard) > 0 and len(shard) < batch_size:
+            # Every batch was incomplete and therefore dropped, so the function was never called.
+            # Return an empty dataset (matching `IterableDataset.batch`/`Dataset.iter`) instead of
+            # silently passing through the original, unprocessed shard.
+            yield rank, True, shard.select([])
         else:
             yield rank, True, shard
 
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -4060,7 +4060,18 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
             # Every batch was incomplete and therefore dropped, so the function was never called.
             # Return an empty dataset (matching `IterableDataset.batch`/`Dataset.iter`) instead of
             # silently passing through the original, unprocessed shard.
-            yield rank, True, shard.select([])
+            if features is not None:
+                # An explicit `features` was requested for the output: honor it instead of falling
+                # back to the (pre-map) input schema, since the caller already knows the intended
+                # output schema and it can't be inferred from a function that was never called.
+                empty_shard = Dataset.from_dict({col: [] for col in features}, features=features, split=shard.split)
+            else:
+                empty_shard = shard.select([])
+                if remove_columns:
+                    empty_shard = empty_shard.remove_columns(
+                        [col for col in remove_columns if col in empty_shard.column_names]
+                    )
+            yield rank, True, empty_shard
         else:
             yield rank, True, shard
 
--- a/tests/test_arrow_dataset.py
+++ b/tests/test_arrow_dataset.py
@@ -1319,6 +1319,19 @@ class BaseDatasetTest(TestCase):
                     )
                     assert_arrow_metadata_are_synced_with_dataset_features(dset_test_with_indices)
 
+    def test_map_batched_drop_last_batch_smaller_than_batch_size(self, in_memory):
+        # Regression test: when `drop_last_batch=True` and the dataset has fewer rows than
+        # `batch_size`, every batch is incomplete, so the function is never called and the
+        # result should be an empty dataset (matching `IterableDataset.batch`/`Dataset.iter`),
+        # not the original, unprocessed dataset passed through unchanged.
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            with self._create_dummy_dataset(in_memory, tmp_dir) as dset:
+                self.assertEqual(len(dset), 30)
+                with dset.map(
+                    lambda batch: batch, batched=True, batch_size=1000, drop_last_batch=True
+                ) as dset_test_batched:
+                    self.assertEqual(len(dset_test_batched), 0)
+
     def test_map_batched(self, in_memory):
         def map_batched(example):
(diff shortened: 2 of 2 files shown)


shashvat-singham (commented):
I checked this out and ran it against the cases from #8386. The row count is fixed, but `shard.select([])` keeps the **input** schema, and that is option 1 from the issue — the one the reporter flagged as "wrong for `batch()`". Three things fall out of it, all measured on this branch (`71ddff2`).

**1. `Dataset.batch()` and `IterableDataset.batch()` still disagree, just on schema now instead of length.**

```python
small = Dataset.from_dict({"a": [1, 2, 3]})

small.batch(5, drop_last_batch=True).features
# {'a': Value('int64')}          <- this branch

small.to_iterable_dataset().batch(5, drop_last_batch=True).features
# {'a': List(Value('int64'))}    <- the path #8386 calls correct
```

A non-empty `Dataset.batch(2, drop_last_batch=True)` gives `{'a': List(Value('int64'))}`, so on this branch the same call returns a *different schema depending on how many rows went in*. That is the divergence the issue is about, moved from `len()` to `.features`.

**2. An explicit `features=` argument is ignored.**

```python
ds.map(
    lambda b: {"out": ["x"] * len(b["a"])},
    batched=True, batch_size=5, drop_last_batch=True,
    features=Features({"out": Value("string")}),
    remove_columns=["a"],
)
# got:      {'a': Value('int64')}
# expected: {'out': Value('string')}
```

The user declared the output schema explicitly and it is discarded. This one seems clearly wrong regardless of how the `batch()` question is resolved — the declared features are available without inference.

**3. `remove_columns` is ignored.**

```python
ds.map(lambda b: {"b": b["a"]}, batched=True, batch_size=5, drop_last_batch=True, remove_columns=["a"])
# got:      {'a': Value('int64')}
# expected: {'b': Value('int64')}
```

The dropped column comes back in the empty result.

---

Case 2 looks cheap to handle: when `features` is not None the writer can be initialised from it, which is close to option 2 in the issue for the case where nothing has to be inferred. Cases 1 and 3 need the output schema when the function never ran, which is the genuinely ambiguous part the reporter raised and did not want to decide alone — probably worth a maintainer steer before more code goes in.

One thing I checked and it is **fine**, noting it so nobody else has to: `batch_size=None` does not trip the new `len(shard) < batch_size` comparison. `None` means one single batch, that batch is complete, so the function runs, `update_data` is set and the `elif` is never evaluated — `ds.map(..., batch_size=None, drop_last_batch=True)` still returns all 3 rows.

Verified on Windows 11 / Python 3.11.9 / pyarrow 25.0.1. Not a maintainer, so take the schema call as input rather than a blocker — the length fix itself matches what #8386 asked for.

mayuriphad:
Thanks for the thorough check. Addressed the two clear-cut ones in 4c8ec30:

- When `features=` is explicitly passed, the empty result is now built from that `features` directly instead of falling back to the input schema.
- `remove_columns` is now applied to the empty result, so a removed column no longer reappears.

Left issue 1 (schema divergence between `Dataset.batch()` and `IterableDataset.batch()`, `List` vs scalar `Value`) alone for now — as you noted, that's a real design question about what the "right" inferred schema is when the function never ran, and deserves a maintainer call rather than a guess baked into this fix.

PR #8457, opened 2026-08-09
fix(types): expose PyTorch dataset compatibility

### LINKED ISSUE #7500: Make `with_format` correctly indicate that a `Dataset` is compatible with PyTorch's `Dataset` class
### Feature request

Currently `datasets` does not correctly indicate to the Python type-checker (e.g. `pyright` / `Pylance`) that the output of `with_format` is compatible with PyTorch's `Dataloader` since it does not indicate that the HuggingFace `Dataset` is compatible with the PyTorch `Dataset` class. It would be great if we could get the typing to work nicely.

### Motivation

To avoid casting types in our Python code.

### Your contribution

I would be happy to contribute a PR if this is something that may be accepted and could work with the current approach.
This doesn't have to be for just PyTorch, but I imagine that the same thing would be useful for `tensorflow` and such, but we only have a need for PyTorch at this stage.

uuzzrm:
## What does this PR do?

Make `Dataset` and `IterableDataset` visible to static type checkers as compatible with PyTorch's dataset classes, so Pyright accepts both map-style and iterable datasets as `torch.utils.data.DataLoader` inputs, including `dataset.with_format(torch)`. This addresses the typing gap described in #7500.

The runtime behavior remains unchanged:

- `torch` is only imported under `TYPE_CHECKING` for the new bases; `import datasets` never imports torch.
- `Dataset` is not changed into a runtime subclass of `torch.utils.data.Dataset`.
- The existing runtime parent-class behavior for `IterableDataset` (via `_maybe_add_torch_iterable_dataset_parent_class`) is preserved.

## Runtime behavior / MRO notes

The torch base classes are typing-only. The base used in the class statement is a stand-in that, at runtime, resolves to *no base at all* through PEP 560 `__mro_entries__`. As a result:

- `Dataset.__mro__` and `Dataset.__bases__` are identical to `main`: no private shim class (and no torch class) ever appears in the runtime MRO, including for users who never touch torch.
- `IterableDataset.__bases__` stays exactly `(DatasetInfoMixin,)` until `_maybe_add_torch_iterable_dataset_parent_class` appends the real `torch.utils.data.IterableDataset` when torch is in use — the same bases `main` sees, so the `__bases__` mutation is unaffected and the shim never coexists with the real torch class.

## Testing

- `ruff check src/datasets/arrow_dataset.py src/datasets/iterable_dataset.py tests/test_arrow_dataset.py`
- `ruff format --check src/datasets/arrow_dataset.py src/datasets/iterable_dataset.py tests/test_arrow_dataset.py`
- `git diff --check`
- Focused import regression: `import datasets` does not import torch (test added in this PR)
- New regression test asserting no torch-named base appears in the runtime `__bases__`
- `tests/test_arrow_dataset.py -k "pickle or format"`: 36 passed, 17 skipped (matches the review's verification)
- Pyright 1.1.411 with a minimal PyTorch type stub covering `Dataset.__getitem__`, `IterableDataset`, and `DataLoader`: 0 errors

Fixes #7500

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -158,9 +158,16 @@ if TYPE_CHECKING:
     import polars as pl
     import pyspark
     import sqlalchemy
+    from torch.utils.data import Dataset as _TorchDatasetBase
 
     from .dataset_dict import DatasetDict
     from .iterable_dataset import IterableDataset
+else:
+
+    class _TorchDatasetBase:
+        def __class_getitem__(cls, item):
+            return cls
+
 
 logger = logging.get_logger(__name__)
 
@@ -715,7 +722,7 @@ class Column(Sequence_):
             return value == list(self)
 
 
-class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
+class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin, _TorchDatasetBase[Any]):
     """A Dataset backed by an Arrow table."""
 
     def __init__(
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -165,8 +165,20 @@ if TYPE_CHECKING:
 else:
 
     class _TorchDatasetBase:
+        """Runtime stand-in for :class:`torch.utils.data.Dataset`.
+
+        The class statement below uses ``_TorchDatasetBase[Any]`` as a base so type
+        checkers (which see the real torch class imported under ``TYPE_CHECKING``)
+        accept ``Dataset`` as a ``torch.utils.data.DataLoader`` input. At runtime the
+        base is dropped through PEP 560 ``__mro_entries__``, so ``Dataset.__mro__``
+        stays identical to main and no fake torch class leaks into introspection.
+        """
+
         def __class_getitem__(cls, item):
-            return cls
+            return cls()
+
+        def __mro_entries__(self, bases):
+            return ()
 
 
 logger = logging.get_logger(__name__)
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -93,8 +93,15 @@ if TYPE_CHECKING:
     import polars as pl
     import sqlalchemy
     import torch
+    from torch.utils.data import IterableDataset as _TorchIterableDatasetBase
 
     from .builder import Key as BuilderKey
+else:
+
+    class _TorchIterableDatasetBase:
+        def __class_getitem__(cls, item):
+            return cls
+
 
 logger = get_logger(__name__)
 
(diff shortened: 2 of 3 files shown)


shashvat-singham (commented):
The risk I went looking for is the interaction with `_maybe_add_torch_iterable_dataset_parent_class`, which mutates `__bases__` at runtime:

```python
if torch.utils.data.IterableDataset not in cls.__bases__:
    cls.__bases__ += (torch.utils.data.IterableDataset,)
```

`__bases__` assignment is one of the more fragile things in CPython, and this PR gives that same class an extra base. **It still works**, so this is a "checked, it's fine" rather than a problem — but it is worth someone confirming with torch actually installed, because I could not.

I do not have torch here, so I stood in a `Generic` subclass shaped like `torch.utils.data.IterableDataset` and replayed the mutation against both class definitions:

```
main (DatasetInfoMixin only)       OK   bases=['DatasetInfoMixin', 'FakeTorchIterableDataset']
PR (+ _TorchIterableDatasetBase)   OK   bases=['DatasetInfoMixin', '_TorchIterableDatasetBase', 'FakeTorchIterableDataset']
PR map-style Dataset               OK   bases=[..., '_TorchDatasetBase', 'FakeTorchIterableDataset']
```

`issubclass` holds afterwards in all three, so the MRO does not conflict. That is a proxy, not the real thing — treat it as "no obvious layout conflict" rather than proof.

What I could verify directly, on this branch:

- `import datasets` does not import torch — your new `test_import_datasets_does_not_import_torch` passes, and I confirmed it independently.
- `Dataset.__mro__` becomes `['Dataset', 'DatasetInfoMixin', 'IndexableMixin', 'TensorflowDatasetMixin', '_TorchDatasetBase', 'object']`. Pickle round-trip, `copy.deepcopy`, `map`, indexing and `len` all behave.
- `tests/test_arrow_dataset.py -k "pickle or format"`: 36 passed, 17 skipped.

Two things worth a line in the description rather than changes:

**The shim stays in the MRO forever.** With torch installed, `IterableDataset` ends up with *both* `_TorchIterableDatasetBase` and the real `torch.utils.data.IterableDataset` in `__bases__`, since `TYPE_CHECKING` is false at runtime and the `else` branch always defines the shim. Harmless, but anyone reading `__mro__` while debugging a DataLoader sharding problem will find a class that looks like it should be torch's and is not.

**`Dataset` gains a base it never had.** The map-style `Dataset` is now a runtime subclass of a private empty class. That is the stated intent (no runtime torch subclassing), so no objection — just noting that `Dataset.__mro__` is public-ish surface and this changes it for everyone, including people who never touch torch.

Tested on Windows 11 / Python 3.11.9, no torch installed.

uuzzrm:
Thanks for the thorough review — the replay of the `__bases__` mutation was super helpful, and the MRO observation was spot on. I went a step further than just documenting it: the runtime shims now drop out of the class statement entirely via PEP 560 `__mro_entries__`, so `Dataset.__mro__` / `Dataset.__bases__` and `IterableDataset.__bases__` are byte-for-byte identical to `main` at runtime — no private shim class, and the shim can never coexist with the real torch base when the mutation runs with torch installed. I added a regression test for this and updated the PR description with a runtime-behavior section; `tests/test_arrow_dataset.py -k "pickle or format"` still comes out 36 passed / 17 skipped. Note the CI runs are still showing `ACTION_REQUIRED` awaiting approval since this is a fork PR, so they haven't actually executed yet.

PR #8461, opened 2026-08-11
Add test for tokenizer stride with Dataset.map (issue #4769)

### LINKED ISSUE #4769: Fail to process SQuADv1.1 datasets with max_seq_length=128, doc_stride=96.
## Describe the bug
datasets fail to process SQuADv1.1 with max_seq_length=128, doc_stride=96 when calling datasets["train"].train_dataset.map().

## Steps to reproduce the bug

I used huggingface[ TF2 question-answering examples](https://github.com/huggingface/transformers/tree/main/examples/tensorflow/question-answering). And my scripts are as follows:

```
python run_qa.py \
  --model_name_or_path $BERT_DIR \
  --dataset_name $SQUAD_DIR \
  --do_train \
  --do_eval \
  --per_device_train_batch_size 12 \
  --learning_rate 3e-5 \
  --num_train_epochs 2 \
  --max_seq_length 128 \
  --doc_stride 96 \
  --output_dir $OUTPUT \
  --save_steps 10000 \
  --overwrite_cache \
  --overwrite_output_dir \

```

## Expected results
Normally process SQuADv1.1 datasets with max_seq_length=128, doc_stride=96.

## Actual results
```
INFO:__main__:Padding all batches to max length because argument was set or we're on TPU.
WARNING:datasets.fingerprint:Parameter 'function'=<function main.<locals>.prepare_train_features at 0x7f15bc2d07a0> of the transform datasets.arrow_dataset.Dataset._map_single couldn't be hashed properly, a random hash was used instead. Make sure your transforms and parameters are serializable with pickle or dill for the dataset fingerprinting and caching to work. If you reuse this transform, the caching mechanism will consider it to be different from the previous calls and recompute everything. This warning is only showed once. Subsequent hashing failures won't be showed.
  0%|                                                                                                                                                               | 0/88 [00:00<?, ?ba/s]thread '<unnamed>' panicked at 'assertion failed: stride < max_len', /__w/tokenizers/tokenizers/tokenizers/src/tokenizer/encoding.rs:311:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
  0%|                                                                                                                                                               | 0/88 [00:00<?, ?ba/s]
  File "run_qa.py", line 743, in <module>
    main()
  File "run_qa.py", line 485, in main
    load_from_cache_file=not data_args.overwrite_cache,
  File "/anaconda3/envs/py37/lib/python3.7/site-packages/datasets/arrow_dataset.py", line 2394, in map
    desc=desc,
  File "/anaconda3/envs/py37/lib/python3.7/site-packages/datasets/arrow_dataset.py", line 551, in wrapper
  File "/anaconda3/envs/py37/lib/python3.7/site-packages/datasets/arrow_dataset.py", line 518, in wrapper
  File "/anaconda3/envs/py37/lib/python3.7/site-packages/datasets/fingerprint.py", line 458, in wrapper
  File "anaconda3/envs/py37/lib/python3.7/site-packages/datasets/arrow_dataset.py", line 2768, in _map_single
    offset=offset,
  File "anaconda3/envs/py37/lib/python3.7/site-packages/datasets/arrow_dataset.py", line 2644, in apply_function_on_filtered_inputs
    processed_inputs = function(*fn_args, *additional_args, **fn_kwargs)
  File "anaconda3/envs/py37/lib/python3.7/site-packages/datasets/arrow_dataset.py", line 2336, in decorated
    result = f(decorated_item, *args, **kwargs)
  File "run_qa.py", line 410, in prepare_train_features
    padding=padding,
  File "anaconda3/envs/py37/lib/python3.7/site-packages/transformers/tokenization_utils_base.py", line 2512, in __call__
    **kwargs,
  File "anaconda3/envs/py37/lib/python3.7/site-packages/transformers/tokenization_utils_base.py", line 2703, in batch_encode_plus
    **kwargs,
  File "anaconda3/envs/py37/lib/python3.7/site-packages/transformers/tokenization_utils_fast.py", line 429, in _batch_encode_plus
    is_pretokenized=is_split_into_words,
pyo3_runtime.PanicException: assertion failed: stride < max_len
  File "./data/SQuADv1.1/evaluate-v1.1.py", line 92, in <module>
    with open(args.prediction_file) as prediction_file:
FileNotFoundError: [Errno 2] No such file or directory: './output/bert_base_squadv1.1_tf2/eval_predictions.json'

```

## Environment info

- `datasets` version: 2.3.2
- Platform: Ubuntu, pytorch=1.11.0, tensorflow-gpu=2.9.1
- Python version: 2.7
- PyArrow version: 8.0.0

imjimit07:
## Problem
When processing SQuAD v1.1 datasets with `max_seq_length=128` and `doc_stride=96` using transformers' QA pipeline, users encountered an assertion failure:
```
thread '<unnamed>' panicked at 'assertion failed: stride < max_len'
```

This prevented legitimate use of the stride parameter in question-answering preprocessing pipelines.

## Solution
Added a comprehensive test case `test_map_with_tokenizer_and_stride` that:
- Reproduces the scenario where tokenizers create overlapping chunks with stride
- Validates that `Dataset.map()` correctly handles tokenization with `max_length=128` and `stride=96`
- Confirms that multiple overlapping sequences are properly created and returned

## Test Coverage
The new test:
1. Creates a dataset with text that requires multiple chunks
2. Uses transformers' `AutoTokenizer` with `return_overflowing_tokens=True` and `stride=96`
3. Applies tokenization via `Dataset.map()` with batched processing
4. Verifies the result contains the expected columns and sequences

This ensures users can safely use `Dataset.map()` with tokenizers and stride parameters in their QA pipelines without assertion errors.

Closes #4769


shashvat-singham (commented):
I do not think this test can fail, with or without the bug from #4769, so it would not catch a regression.

The panic is `assertion failed: stride < max_len`, and `max_len` there is not `max_length` — it is the space left for the **context** after the question and the special tokens are subtracted. That is why #4769 hits it with `max_seq_length=128, doc_stride=96`: it is a question-answering pipeline, so each example is a *pair*, and a question long enough to eat more than 32 tokens drops the context budget below the stride.

This test tokenizes a single sequence:

```python
result = tokenizer(
    examples["text"],
    max_length=128,
    truncation=True,
    return_overflowing_tokens=True,
    stride=96,
)
```

No pair, so nothing competes for the budget: `max_len` is `128 - 2` for `[CLS]`/`[SEP]`, and `96 < 126` holds comfortably. The assertion is never reached. The test passes on a tree with the bug present and on one without it.

To actually reproduce #4769 the call needs to be the two-argument form with a question long enough to squeeze the context, roughly:

```python
tokenizer(
    examples["question"],
    examples["context"],
    max_length=128,
    truncation="only_second",
    return_overflowing_tokens=True,
    stride=96,
)
```

with a question of ~35+ tokens. Then `max_len = 128 - len(question) - 3` falls under `96` and the assertion fires.

Two smaller things, if the test stays:

- It asserts `len(result) > 0` and that `input_ids` exists. Both hold for any successful tokenization, so the assertions do not pin down the stride behaviour the test is named for. Asserting the number of overflow chunks for a known input would.
- It downloads `bert-base-uncased` at test time. `@require_transformers` covers the import but not the network; the repo has `@require_transformers` + hub fixtures elsewhere for this, worth matching.

Worth raising the bigger question too: `assertion failed: stride < max_len` is raised inside `tokenizers`, from a function `datasets` only calls through the user's own `map` callable. `datasets` never sees `stride`. If the conclusion is that the fix belongs in `transformers`/`tokenizers`, then a test here would be pinning behaviour this repo does not own — #4769 may simply be filed on the wrong repo, which would also explain why it is still open.

Not a maintainer — flagging it because a regression test that cannot go red is worse than no test, since it reads as coverage.

imjimit07:
Thanks for the review, addressed your feedback. I added a focused integration test (tests/test_stride_qa.py) that uses the two-argument tokenizer form (question + context) and a ~35-token question to exercise the original failure mode. The test now conditionally xfails when the tokenizers Rust backend panics or raises the truncation error (documents upstream behavior) so CI is not blocked. Pushed the changes to the fork branch imjimit07-fix-squad-seq-length-stride. If you prefer the test to be strict when tokenizers is fixed, I can add a version-gated assertion or open a follow-up issue in tokenizers/transformers to track the upstream fix.

PR #8477, opened 2026-08-13
Reject non-positive batch sizes in batch()

### LINKED ISSUE #8472: batch(0) silently returns one batch with the whole dataset instead of raising
### Describe the bug

`Dataset.batch()` and `IterableDataset.batch()` accept `batch_size=0` and negative sizes, and quietly return **one batch holding the whole dataset**. Asking for a zero-sized batch produces the largest possible batch, which is as close to the opposite of the request as the API can get.

### Steps or code to reproduce the bug

```python
from datasets import Dataset

ds = Dataset.from_dict({"a": list(range(5))})
it = ds.to_iterable_dataset()

[len(b["a"]) for b in it.batch(0)]    # [5]
[len(b["a"]) for b in it.batch(-1)]   # [5]
[len(b["a"]) for b in it.batch(2)]    # [2, 2, 1]   <- sane case

len(ds.batch(0))    # 1
len(ds.batch(-1))   # 1
len(ds.batch(2))    # 3
```

Both the eager and the streaming path behave the same way, so this is consistent — just consistently wrong.

### Expected behavior

`batch_size` should be required to be a positive integer, raising something like

```
ValueError: batch_size must be a positive integer, but got 0.
```

Silently substituting "everything in one batch" is the worst of the options: it looks like it worked, and a `batch_size` computed at runtime that comes out as `0` (an empty shard, an integer division, a config default that did not get filled in) turns into a single unbounded batch. On a large dataset that is a memory blow-up rather than an error.

### Note on overlap with open PRs

I checked the open PRs before filing, and this specific path is not covered by either of the two nearby ones:

- #8446 adds `_check_batch_size` but wires it only into the **export** methods (`to_csv`, `to_json`, ...), not `batch()`.
- #8443 validates `by_column` and the "missing `batch_size` or `by_column`" message in `Dataset.batch`, but does not check the value of `batch_size`.

So the helper that `batch()` wants already exists in #8446:

```python
def _check_batch_size(batch_size: Optional[int]):
    if batch_size is not None and batch_size <= 0:
        raise ValueError(f"batch_size must be a positive integer, but got {batch_size}.")
```

and the fix is largely calling it from `Dataset.batch` and `IterableDataset.batch` too. That does mean this is best done after #8446 lands, to avoid two copies of the same helper.

### Related, lower confidence

Same family, listed separately because the right answer is less obvious and I am not proposing a change: `IterableDataset.take(-1)` yields 0 rows and `skip(-1)` yields all rows, i.e. negatives are silently clamped rather than rejected. That is defensible as "clamp like a slice", unlike `batch(0)`, where there is no reading under which one giant batch is the answer.

### Environment info

- `datasets` version: 5.0.2.dev0 (`main` @ 48b7ee7)
- Python version: 3.11.9
- Platform: Windows 11
- PyArrow version: 25.0.1

Anakintano:
## Summary

- reject non-positive `batch_size` values in `Dataset.batch()` and `IterableDataset.batch()`
- raise before mapping or consuming an iterable source
- add focused regression tests for `0` and `-1`

Fixes #8472

## Related work

This includes the small `_check_batch_size` helper also proposed by @LeSingh1 in #8446, then applies it to the two `batch()` APIs that #8446 does not cover. If #8446 lands first, I can rebase this PR and drop the overlapping helper.

## Validation

- original #8472 reproducer now raises the expected `ValueError` for `0` and `-1`
- focused regression tests: 4 passed
- affected batching tests: 20 passed, 2 skipped
- Ruff lint and format checks passed

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -633,6 +633,12 @@ def _check_column_names(column_names: list[str]):
         raise ValueError(f"The table can't have duplicated columns but columns {duplicated_columns} are duplicated.")
 
 
+def _check_batch_size(batch_size: Optional[int]):
+    """Check the `batch_size` of methods that process a dataset by batches."""
+    if batch_size is not None and batch_size <= 0:
+        raise ValueError(f"batch_size must be a positive integer, but got {batch_size}.")
+
+
 def _check_valid_indices_value(index, size):
     if (index < 0 and index + size < 0) or (index >= size):
         raise IndexError(f"Index {index} out of range for dataset of size {size}.")
@@ -4112,6 +4118,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         """
         if batch_size is None and by_column is None:
             raise ValueError("IterableDataset.batch() misses `batch_size` or `by_column` argument.")
+        _check_batch_size(batch_size)
         if by_column is not None:
             if num_proc:
                 raise NotImplementedError("Multiprocessed batching by column is not implemented yet")
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -35,7 +35,7 @@ from huggingface_hub.utils import RepositoryNotFoundError
 from packaging import version
 
 from . import __version__, config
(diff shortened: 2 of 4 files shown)


Anakintano:
@lhoestq, when you have a chance, could you take a look and authorize the CI run? This is a focused fix for #8472, with the overlap with #8446 called out in the description. Thanks!

PR #8490, opened 2026-08-16
Fix archive path check allowing sibling-directory escapes

icn5381:
## Bug

The `safemembers` path check that guards archive extraction (the CVE-2007-4559 mitigation, duplicated across the tar/zip/rar/7z extractors) uses a plain `startswith(base)`. That also matches **sibling directories whose name starts with the output directory's name**:

```py
target = resolved(os.path.join(base, path))
return not target.startswith(base)     # "/cache/extracted_evil/x".startswith("/cache/extracted") → True → allowed
```

Extracting into `.../extracted/<hash>` therefore lets a member named `../<hash>_evil/x` write outside the output directory, into a sibling the archive itself names. On a shared machine this is a cache-poisoning vector: one dataset's archive can overwrite files inside *another* dataset's extraction directory. Reproduced on current `main` with a three-line tar before the fix:

```
TarExtractor.extract(evil.tar, tmp/"extracted")
(tmp/"extracted_evil"/"escaped.txt").exists()  → True   # file written outside the output dir
```

Paths escaping to unrelated directories (`../..`) stay blocked; only the shared-prefix sibling passes the old check.

## Fix

`badpath` now requires the resolved target to be the base itself (a member named `.`) or start with `base + os.sep`, in all four `safemembers` copies. The link checks (`badlink`) share `badpath`, so they get the same tightening.

## Verification

- New regression fixture `tar_file_with_sibling_prefix` added to the parametrized `test_tar_extract_insecure_files`, asserting the member is blocked with the "illegal path" error, alongside the existing dot-dot and symlink cases.
- Manual probe before/after the fix: escape succeeded on `main`, is blocked after, and legitimate nested members (`sub/good.txt`) still extract.
- `tests/test_extract.py`: 20 passed.

### DIFF
--- a/src/datasets/utils/extract.py
+++ b/src/datasets/utils/extract.py
@@ -105,7 +105,11 @@ class TarExtractor(BaseExtractor):
 
         def badpath(path: str, base: str) -> bool:
             # joinpath will ignore base if path is absolute
-            return not resolved(os.path.join(base, path)).startswith(base)
+            # Also, a plain startswith(base) would still match sibling directories
+            # whose name starts with the base name (e.g. "../base_evil/x" when
+            # extracting into ".../base"), so require the base plus a separator
+            target = resolved(os.path.join(base, path))
+            return target != base and not target.startswith(base + os.sep)
 
         def badlink(info: tarfile.TarInfo, base: str) -> bool:
             # Links are interpreted relative to the directory containing the link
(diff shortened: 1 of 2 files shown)


ebarkhordar (commented):
Confirmed at 17b0f16 in a clean container: the escape reproduces on the merge base (836b82e0) and is blocked at head, both through `TarExtractor.extract` directly and through the real `ExtractManager` flow. `tests/test_extract.py` is 14 passed / 6 skipped at head, and the new case fails on the merge base, so it is a genuine regression test.

One correction for the description. Under `ExtractManager` the output directory is `hash_url_to_filename()` output, always 64 hex characters, so no other dataset's extraction directory can have one as a strict prefix. The sibling escape cannot reach another dataset. What it does reach is the live lock file from `extract.py:456`:

```py
f.add(payload, arcname="../" + os.path.basename(output_path) + ".lock")
```

Before, that lands on the `FileLock` path held open around the extraction (lock body becomes `CLOBBERED-BY-ARCHIVE`); after, it is blocked. Narrower than "overwrite files inside another dataset's extraction directory", still worth closing.

lhoestq (approved):
lgtm !

PR #8493, opened 2026-08-17
Don't mask the original error when finalizing the writer fails in Dataset.map

### LINKED ISSUE #8491: `Dataset.map(num_proc=N)` worker crashes with `ValueError: I/O operation on closed file` when `finalize()` is interrupted
### Describe the bug

We hit this while running Megatron-LM training through ms-swift, which uses `datasets` under the hood for preprocessing. Our preprocessing step calls `Dataset.map()` with a fairly aggressive `num_proc=512` on a ~25k-example JSONL file:

```bash
swift --dataset_num_proc 512 --dataset large.jsonl --max_length 131072 ...
```

Near the end of the map (around 94%), one of the worker processes died with a confusing two-stage traceback, and since this runs inside distributed training, the whole 128-GPU job went down with it.

#### Error logs

The worker first hit an `InterruptedError` while finalizing its output shard:

```
Map (num_proc=512):  94%|█████████▍| 24277/25742 [01:00<00:03, 398.48 examples/s]
[rank48]: Traceback (most recent call last):
[rank48]:   File ".../datasets/arrow_dataset.py", line 4022, in _map_single
[rank48]:     writer.finalize()  # close_stream=bool(buf_writer is None))
[rank48]:   File ".../datasets/arrow_writer.py", line 778, in finalize
[rank48]:     self.stream.close()
[rank48]:   File ".../fsspec/implementations/local.py", line 450, in close
[rank48]:     return self.f.close()
[rank48]: InterruptedError: [Errno 4] Interrupted system call
```

That part is at least understandable — with 512 processes being spawned/reaped there's a lot of signal noise, and apparently `close()` can legitimately get interrupted.

The real problem is what happened next. The `except` block in `_map_single()` tried to call `finalize()` again, and that second call blew up with a completely different error:

```
[rank48]: During handling of the above exception, another exception occurred:
[rank48]:
[rank48]: Traceback (most recent call last):
[rank48]:   File ".../multiprocess/pool.py", line 125, in worker
[rank48]:     result = (True, func(*args, **kwds))
[rank48]:   File ".../datasets/utils/py_utils.py", line 585, in _write_generator_to_queue
[rank48]:     for i, result in enumerate(func(**kwargs)):
[rank48]:   File ".../datasets/arrow_dataset.py", line 4027, in _map_single
[rank48]:     writer.finalize()
[rank48]:   File ".../datasets/arrow_writer.py", line 773, in finalize
[rank48]:     self._build_writer(self.schema)
[rank48]:   File ".../datasets/arrow_writer.py", line 601, in _build_writer
[rank48]:     self.pa_writer = pa.RecordBatchStreamWriter(self.stream, self._schema)
[rank48]:   File "pyarrow/ipc.py", line 90, in __init__
[rank48]:     self._open(sink, schema, options=options)
[rank48]: ValueError: I/O operation on closed file
```

So the exception that actually kills the worker (`ValueError: I/O operation on closed file`) is not the original failure — it's produced by the error-handling path itself, which masks the real problem and makes this very confusing to debug.

The same script runs fine with smaller `num_proc` values (we've used 32 and 64 without issues), so this seems related to the high process count.

```python

ds = load_dataset("json", data_files="large.jsonl", split="train")
ds = ds.map(lambda x: x, num_proc=512)
```

It's timing/signal dependent so it doesn't fire 100% of the time, but with a high enough `num_proc` and enough examples it reproduces pretty reliably for us — we've hit it multiple times on ~25k examples with `num_proc=512`.

### Expected behavior

- A transient `InterruptedError` while closing the output stream shouldn't bring down the worker.
- The cleanup path in `_map_single()` shouldn't raise a *new* exception that hides the original one.

### Environment info

- `datasets`: 5.0.0 (I also checked the 5.0.1 source and the relevant code looks unchanged)
- Python: 3.12.11
- OS: Linux (training container)
- PyTorch: 2.8.0
- `multiprocess`: bundled with `datasets`

jayzuccarelli:
## Summary

Fixes #8491.

When `writer.finalize()` raises in `_map_single()`, the `except` block calls `finalize()` again on a writer whose stream is already closed. `finalize()` then hits `self.pa_writer is None and self.schema` and calls `_build_writer(self.schema)`, which raises `ValueError: I/O operation on closed file` from pyarrow. That replaces the exception we were handling, so the traceback the user sees points at the cleanup path instead of the real failure (in #8491 an `InterruptedError` from `stream.close()` with `num_proc=512`). It also skips the `tmp_file.close()` / `os.remove()` below it, leaving the partial shard behind.

Wrap the cleanup `finalize()` so the original error reaches the `raise` and the temp file still gets removed.

Second commit (after @HowardZorn's review): `ArrowWriter.finalize` now ignores `InterruptedError` from `stream.close()`. By then `pa_writer.close()` has flushed everything and, on Linux, the fd is already closed even when `close(2)` returns EINTR, so retrying would be wrong (fd reuse, see `man 2 close`; PEP 475 leaves `close()` out for the same reason). We log and continue, so the worker survives. The cleanup path in `_map_single` logs a `warning` (not `debug`) with the finalize error so it isn't silently lost.

## How tested

New tests in `tests/test_arrow_dataset.py`:

`test_map_ignores_interrupted_close`: patch the stream's `close` to close for real then raise `InterruptedError(EINTR)`, assert `map()` succeeds and the output reads back. Fails on `main` with the `InterruptedError`.

`test_map_keeps_the_error_raised_when_finalizing`: patch `ArrowWriter.finalize` so the first call runs for real and then raises `InterruptedError` (the state in the report: `pa_writer` set to None, stream closed), and assert `map()` propagates that `InterruptedError`. On `main` it fails with `ValueError: I/O operation on closed file` raised from `_build_writer`, same as the report.

```
$ python -m pytest tests/test_arrow_dataset.py -q -n 8
1 failed, 363 passed, 61 skipped, 17 warnings
```

(The one failure is `test_dataset_to_iterable_dataset`, which errors with "PyTorch needs to be installed to be able to return PyTorch tensors" because torch is not installed in my environment. Unrelated to this change.)

`make quality` is clean (`ruff check` passed, `ruff format --check` reports 233 files already formatted).

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -4027,8 +4027,8 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                     if writer is not None:
                         try:
                             writer.finalize()
-                        except Exception:  # the writer may be broken already, don't hide the error we are handling
-                            logger.debug("Failed to finalize the writer.")
+                        except Exception as e:  # the writer may be broken already, don't hide the original error
+                            logger.warning(f"Failed to finalize the writer while handling an error: {e!r}")
                     if tmp_file is not None:
                         tmp_file.close()
                         if os.path.exists(tmp_file.name):
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -4025,7 +4025,10 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                 yield rank, False, num_examples_progress_update
                 if update_data:
                     if writer is not None:
-                        writer.finalize()
+                        try:
+                            writer.finalize()
+                        except Exception:  # the writer may be broken already, don't hide the error we are handling
+                            logger.debug("Failed to finalize the writer.")
                     if tmp_file is not None:
                         tmp_file.close()
                         if os.path.exists(tmp_file.name):
--- a/src/datasets/arrow_writer.py
+++ b/src/datasets/arrow_writer.py
@@ -788,7 +788,13 @@ class ArrowWriter:
             self.pa_writer.close()
             self.pa_writer = None
             if close_stream:
-                self.stream.close()
+                try:
+                    self.stream.close()
+                except InterruptedError:
+                    # close(2) was interrupted by a signal: the data is already flushed and, on Linux, the file
+                    # descriptor is already closed. Retrying close() would be wrong since the descriptor may have
+                    # been reused (see `man 2 close`), so just carry on.
+                    logger.debug("Closing the stream was interrupted by a signal, ignoring.")
         else:
             if close_stream:
                 self.stream.close()
--- a/tests/test_arrow_dataset.py
+++ b/tests/test_arrow_dataset.py
@@ -4830,6 +4830,25 @@ def test_map_keeps_the_error_raised_when_finalizing():
             dset.map(lambda x: x)
 
 
+def test_map_ignores_interrupted_close(tmp_path):
+    dset = Dataset.from_dict({"x": range(10)})
+    finalize = ArrowWriter.finalize
+
+    def finalize_with_interrupted_close(self, *args, **kwargs):
+        stream_close = self.stream.close
+
+        def close():
+            stream_close()
+            raise InterruptedError(errno.EINTR, "Interrupted system call")
+
+        with patch.object(self.stream, "close", close):
+            return finalize(self, *args, **kwargs)
+
+    with patch.object(ArrowWriter, "finalize", finalize_with_interrupted_close):
+        out = dset.map(lambda x: {"y": x["x"] + 1}, cache_file_name=str(tmp_path / "cache.arrow"))
+    assert out["y"] == list(range(1, 11))
+
+
 def test_filter_async():
     dset = Dataset.from_dict({"x": range(100)})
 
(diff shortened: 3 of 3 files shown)


HowardZorn:
This could not fix https://github.com/huggingface/datasets/issues/8491. The `InterruptedError` raised from `file.close()` does not actually need to be processed on the Linux platform as I mentioned in the issue.

src/datasets/arrow_dataset.py
HowardZorn, line 1: The code currently suppresses all errors during writer.finalize(). Since `logger.debug` is usually disabled in production, any failure during cleanup will be silently lost.

jayzuccarelli:
Yep, you're right: this only fixed the masking, the worker still died on the `InterruptedError`.

Pushed ea95838:
- `ArrowWriter.finalize` now ignores `InterruptedError` from `stream.close()`. At that point `pa_writer.close()` already flushed everything and on Linux the fd is already closed, so retrying would be wrong (fd reuse, per `man 2 close`); we just log and continue.
- The cleanup path in `_map_single` now logs a `warning` with the finalize error instead of `debug`, so it isn't silently lost.
- Test: `test_map_ignores_interrupted_close` (close raises EINTR after actually closing, `map` succeeds and the shard reads back).

src/datasets/arrow_dataset.py
jayzuccarelli: Fair, bumped to `logger.warning` with the exception in ea95838.

PR #8496, opened 2026-08-18
Preserve storage options across URL hops

tandede:
## Summary

- return the accumulated storage options for every hop in chained fsspec URLs
- add a regression test covering protocol-specific options on both `zip` and `https` hops

## Problem

`_prepare_path_and_storage_options` builds `prepared_storage_options` while iterating over a multi-hop path, but returned the loop-local `storage_options` value instead. As a result, options for every hop except the last one were silently discarded. This can drop archive settings, credentials, or other protocol-specific configuration from paths such as `zip://data.jsonl::https://example.org/archive.zip`.

## Validation

- `make quality`
- `pytest tests/test_file_utils.py` (138 passed, 3 skipped)
- focused multi-hop and single-hop preparation tests (9 passed)

I also ran the broader test suite: 3,172 tests passed and 18 skipped before it was stopped after 15 minutes; the 97 failures were confined to optional local integrations such as TorchCodec/FFmpeg media decoding, Spark/Java, Faiss, and distributed multiprocessing, unrelated to this change.

### DIFF
--- a/src/datasets/utils/file_utils.py
+++ b/src/datasets/utils/file_utils.py
@@ -886,7 +886,7 @@ def _prepare_path_and_storage_options(
         hop, storage_options = _prepare_single_hop_path_and_storage_options(hop, download_config=download_config)
         prepared_urlpath.append(hop)
         prepared_storage_options.update(storage_options)
-    return "::".join(prepared_urlpath), storage_options
+    return "::".join(prepared_urlpath), prepared_storage_options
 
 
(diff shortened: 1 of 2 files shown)


tandede:
Hi maintainers, I’m following up on this PR in case it was missed. It preserves storage options across multi-hop URL resolution and is ready for review. Please let me know if any changes or additional tests would be useful. Thank you!

lhoestq (approved):
good catch !

PR #8504, opened 2026-08-23
Preserve special NumPy scalar types in NumPy formatting

### LINKED ISSUE #8500: with_format("numpy") returns bool/datetime values as 0-d unhashable arrays and casts duration columns to int64
### Describe the bug

With numpy formatting, row access returns `np.int64`/`np.float64`/`np.str_` scalars for most dtypes, but **bool and timestamp/date values come back as 0-dimensional `ndarray`s** (`array(True)`), which are unhashable and fail `isinstance(x, np.bool_)` checks. Separately, **batch/column access silently casts `duration` columns from `timedelta64[us]` to plain `int64`** (row access keeps `np.timedelta64`), so the same column reports different types depending on how it's read.

Root cause: `NumpyFormatter._tensorize` (`src/datasets/formatting/np_formatter.py`) early-returns only `np.number` scalars — `np.bool_`/`np.datetime64` aren't `np.number`, so they fall through to `np.asarray()` (and `_recursive_tensorize` re-wraps them via `__array__` first). The int64 default-dtype branch uses `np.issubdtype(dtype, np.integer)`, which is True for `timedelta64` in numpy's type hierarchy, so duration arrays get the int64 default.

```python
from datasets import Dataset
import datetime as dt
ds = Dataset.from_dict({'b': [True], 't': [dt.datetime(2024,1,1)],
                        'td': [dt.timedelta(seconds=5)]}).with_format('numpy')
row = ds[0]
# actual: b -> array(True) (ndarray!), t -> array('2024-01-01...') (ndarray!)
#         td -> np.timedelta64(5000000,'us') (scalar, inconsistent with batch below)
hash(row['b'])   # TypeError: unhashable type: 'numpy.ndarray'
print(ds[:]['td'].dtype)  # int64  <- duration unit silently stripped
```

### Expected behavior

`np.True_`, `np.datetime64(...)` scalars for row access (consistent with int/float/str), and `ds[:]['td'].dtype == timedelta64[us]` (consistent between row and batch access).

I have a fix ready (7-line change + regression tests) and will open a PR.

### Environment info

datasets main (5.0.2.dev0), pyarrow 25.0.1, numpy 2.4.6, Python 3.11.14, macOS

medisean:
Fixes #8500.\n\nTreat NumPy scalar types uniformly in NumpyFormatter so boolean, datetime, and duration values remain scalar values during row formatting. Also avoid applying the integer default dtype to timedelta arrays, preserving duration units for batch and column formatting.\n\nAdded regression coverage for boolean, timestamp, and duration formatting.\n\nValidation: git diff --check and python3 -m py_compile passed. The targeted pytest could not be collected locally because pyarrow and the editable datasets package are not installed.

### DIFF
--- a/src/datasets/formatting/np_formatter.py
+++ b/src/datasets/formatting/np_formatter.py
@@ -48,12 +48,16 @@ class NumpyFormatter(TensorFormatter[Mapping, np.ndarray, Mapping]):
             return value
         elif isinstance(value, (np.character, np.ndarray)) and np.issubdtype(value.dtype, np.character):
             return value
-        elif isinstance(value, np.number):
+        elif isinstance(value, np.generic):
             return value
 
         default_dtype = {}
 
-        if isinstance(value, np.ndarray) and np.issubdtype(value.dtype, np.integer):
+        if (
+            isinstance(value, np.ndarray)
+            and np.issubdtype(value.dtype, np.integer)
+            and not np.issubdtype(value.dtype, np.timedelta64)
+        ):
             default_dtype = {"dtype": np.int64}
         elif isinstance(value, np.ndarray) and np.issubdtype(value.dtype, np.floating):
             default_dtype = {"dtype": np.float32}
@@ -86,7 +90,7 @@ class NumpyFormatter(TensorFormatter[Mapping, np.ndarray, Mapping]):
 
             if isinstance(data_struct, torch.Tensor):
                 return self._tensorize(data_struct.detach().cpu().numpy()[()])
-        if hasattr(data_struct, "__array__") and not isinstance(data_struct, (np.ndarray, np.character, np.number)):
+        if hasattr(data_struct, "__array__") and not isinstance(data_struct, (np.ndarray, np.generic)):
             data_struct = data_struct.__array__()
         # support for nested types like struct of list of struct
         if isinstance(data_struct, np.ndarray):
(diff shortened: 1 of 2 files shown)


ebarkhordar (commented):
The PR body says pytest could not be collected locally, so I ran it against `86105122` in a clean `python:3.11-slim` container (numpy 2.4.6, pyarrow 25.0.1). Your new test passes, and so do the other 9 numpy formatter tests.

Copying your test file onto the merge base `836b82e0` to see what it pins, three of the four assertions fail there and one does not:

```
                                base 836b82e0        head 86105122
row['boolean']                  FAIL  -> ndarray      PASS -> bool
row['timestamp']                FAIL  -> ndarray      PASS -> datetime64
row['duration']                 PASS  -> timedelta64  PASS -> timedelta64
batch['duration'].dtype         FAIL  -> int64        PASS -> timedelta64[us]
```

`row['duration']` already held before your change, because `np.timedelta64.__mro__` is `(timedelta64, signedinteger, integer, number, generic)` and the old `isinstance(value, np.number)` returned it early. So the `np.generic` widening is load-bearing for `np.bool_` and `np.datetime64` only.

That same inheritance is why your array-branch guard is doing real work rather than being redundant: `np.issubdtype(np.dtype('m8[us]'), np.integer)` is `True`, which is what coerced durations to int64.

Separately, #8503 looks like the same fix from you without the `_recursive_tensorize` line, so one of the two can probably be closed.

Issue #8499, opened 2026-08-19
TsFile format filter uses an inconsistent URL and is not available initially


ColinLeeo:
## Summary

There appear to be two related issues with the `TsFile` format filter on the Hugging Face Datasets page:

1. The URL parameter for `TsFile` is inconsistent with other formats.
2. `TsFile` is not initially shown in the format filter and only appears after visiting a TsFile-specific URL.

## 1. Inconsistent URL parameter

Other formats use URLs such as:

https://huggingface.co/datasets?format=format%3Ajson&sort=trending

However, the current URL for `TsFile` is:

https://huggingface.co/datasets?format=tsfile&sort=trending

The corresponding URL using the same convention as other formats does not work:

https://huggingface.co/datasets?format=format%3Atsfile&sort=trending

Should `TsFile` use the same convention as the other formats, such as `format=format:tsfile`? If TsFile intentionally requires a different parameter format, could you please clarify why or consider making the URL handling consistent?

## 2. TsFile is not initially shown in the format filter

When opening the Datasets page directly:

https://huggingface.co/datasets

`TsFile` does not appear in the `Format` filter list. :codex-annotation{index="1"}

However, after visiting the following URL:

https://huggingface.co/datasets?format=tsfile&sort=trending

`TsFile` appears in the `Format` filter list. :codex-annotation{index="2"}

In other words, users must first visit a TsFile-specific query URL before `TsFile` becomes available in the format filter. This may indicate that the format is only loaded or registered after the first query.

## Steps to reproduce

1. Open https://huggingface.co/datasets.
2. View the `Format` filter list and confirm that `TsFile` is not listed.
3. Open https://huggingface.co/datasets?format=tsfile&sort=trending.
4. View the `Format` filter list again.
5. `TsFile` is now listed.

## Expected behavior

- `TsFile` should be shown in the `Format` filter list immediately when the Datasets page is opened.
- Users should not need to visit a TsFile-specific URL first.
- The URL generated for `TsFile` should follow the same convention as the URLs for other formats.
- Filtering by `TsFile` should work consistently through both the UI and the URL.

Could you please investigate why `TsFile` is not initially available in the format filter, and whether any additional configuration is required?

AxelNoun:
Thanks for the very precise report â€” the reproduction steps made this easy to investigate.
I looked into both points and I think the two symptoms share a single cause, which is
neither of the two mechanisms one would naively assume (lazy registration / URL parsing).

[...]

Issue #8256, opened 2026-06-10
Dataset Viewer fails on TSFile datasets


gengziyand:
## Description

The Dataset Viewer fails when trying to display a dataset stored in TSFile format.

The error shown by the viewer is:

```python
ModuleNotFoundError: No module named 'tsfile'

### Steps to reproduce the bug

# Dataset Viewer fails to load TSFile dataset due to missing `tsfile` dependency

## Dataset

Dataset URL:

https://huggingface.co/datasets/smilegeng/pusht

## Description

The dataset is uploaded successfully, but the Dataset Viewer cannot display it.

The viewer reports:

```text
SplitsNotFoundError: The split names could not be parsed from the dataset config.
```

After inspecting the traceback, the actual root cause appears to be a missing dependency:

```text
ModuleNotFoundError: No module named 'tsfile'
```

The error originates from the TSFile dataset loader:

```text
File "/usr/local/lib/python3.12/site-packages/datasets/packaged_modules/tsfile/tsfile.py", line 304, in _scan_metadata
    from tsfile.constants import TIME_COLUMN, ColumnCategory

ModuleNotFoundError: No module named 'tsfile'
```

This exception is then wrapped into:

```text
datasets.inspect.SplitsNotFoundError:
The split names could not be parsed from the dataset config.
```

## Expected Behavior

The Dataset Viewer should either:

1. Support TSFile datasets by installing the required `tsfile` dependency, or
2. Return a clearer error message indicating that the dependency is unavailable.

## Full Traceback

```text
Exception: SplitsNotFoundError
Message: The split names could not be parsed from the dataset config.

Traceback:
...
ModuleNotFoundError: No module named 'tsfile'
...
datasets.inspect.SplitsNotFoundError:
The split names could not be parsed from the dataset config.
```

### Expected behavior

The Dataset Viewer should support TSFile datasets.

### Environment info

D:\huggingface_to_tsfile>datasets-cli env

Copy-and-paste the text below in your GitHub issue.

- `datasets` version: 5.0.0
- Platform: Windows-11-10.0.26200-SP0
- Python version: 3.14.0
- `huggingface_hub` version: 1.7.2
- PyArrow version: 23.0.1
- Pandas version: 3.0.1
- `fsspec` version: 2026.2.0

D:\huggingface_to_tsfile>

lhoestq:
Thanks for reporting ! I'm working on it :)

The viewer doesn't use `datasets` V5 yet, I have to do some minor adjustments first

[...]

PR #8160, opened 2026-04-29
feat: add TsFile (Apache IoTDB) packaged builder with per-device wide format

JackieTien97:
Closes #7922

## Summary

Add a packaged builder for [TsFile](https://tsfile.apache.org/) — the columnar time-series file format used as the native storage layer of [Apache IoTDB](https://iotdb.apache.org/). This enables `load_dataset("tsfile", data_files="...")` with automatic `.tsfile` extension detection.

### Data model

Unlike tabular formats, TsFile is **time-series-aware**: each file contains tables with TAG columns (device identifiers), FIELD columns (measurements), and a TIME column. The builder outputs one dataset row per **device**, where:

- TAG columns are scalar `string` values identifying the device.
- The `time` column and every FIELD column are Arrow `list<...>` columns holding the device's full sorted time series.

```text
<tag_1>:    string
<tag_2>:    string
time:       list<timestamp[unit, tz]>
<field_1>:  list<original_type>
<field_2>:  list<original_type>
```

### Key features

- **Per-device reading** — data is fetched via `TsFileReader.query_table` with a push-down `tag_filter`; peak memory is bounded by a single device's payload, not the entire split.
- **Time-range pushdown** — `start_time` / `end_time` are pushed down to TsFile's internal time index. Accepts `int` epochs, `datetime`, `date`, ISO-8601 strings, and `pyarrow.TimestampScalar`.
- **Schema evolution** — when different files expose different FIELD columns, the loader takes the union and fills missing values with nulls. Numeric types are promoted following IoTDB's widening rules (`INT32 → INT64 → DOUBLE`, `INT32 → FLOAT → DOUBLE`).
- **Case-insensitive table names** — table-name lookups use a canonical lowercase form so auto-detected and user-supplied names always match.
- **`on_bad_files`** — `"error"` (default) / `"warn"` / `"skip"` to control handling of unreadable inputs.
- **Configurable batching** — `input_batch_size` (rows per Arrow batch from the reader) and `output_batch_size` (devices per emitted record batch) for memory control.
- **Timestamp unit & timezone** — `timestamp_unit` (`"s"` / `"ms"` / `"us"` / `"ns"`, default `"ms"`) and `timestamp_tz` (optional timezone string).

### Config knobs (`TsFileConfig`)

| Parameter | Default | Description |
|---|---|---|
| `table_name` | auto-detect | Table to read (case-insensitive) |
| `columns` | all | Subset of FIELD columns to keep; TAGs are always returned |
[...]

PR #8514, opened 2026-08-24
fix(tsfile): guard optional Apache TsFile SDK import

AxelNoun:
## Summary
- Guard every `tsfile` SDK import behind `_require_tsfile()` so missing extras raise `ImportError: ... pip install 'datasets[tsfile]'` instead of a bare `ModuleNotFoundError`.
- Add a `tsfile` extra (`tsfile>=2.3.0`), install it in the `tests` extra on Python 3.14 (where it does not conflict with `pyarrow>=21`), and install `datasets[tsfile]` in the py3.14 CI job.
- Document the extra in `loading.mdx`, `tsfile_load.mdx`, `about_dataset_load.mdx`, and the README.
- Add `tests/packaged_modules/test_tsfile_optional.py` for the missing-SDK path. Existing round-trip tests remain behind `pytest.importorskip(tsfile)`.

This is the `datasets`-side slice of #8499 (see also #8256, #8160, and the overlapping #8307). It does **not** register a Hub `format:tsfile` facet: that still needs the SDK on every dataset-viewer worker **and** Hub taxonomy registration.

## Test plan
- [x] `pytest tests/packaged_modules/test_tsfile_optional.py` (no `tsfile` SDK required)
- [ ] CI py3.14 job installs `datasets[tsfile]` and runs `tests/packaged_modules/test_tsfile.py`
- [ ] `pip install 'datasets[tsfile]'` then `load_dataset(tsfile, data_files=...)`
- [ ] Without the extra, `load_dataset(smilegeng/pusht)` raises the actionable `ImportError`

Refs #8499, #8160, #8256, #8307

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,8 +121,8 @@ jobs:
         run: pip install --upgrade uv
       - name: Install dependencies
         run: uv pip install --system "datasets[tests] @ ."
-      - name: Install tsfile (py3.14 only)
-        run: uv pip install --system "tsfile>=2.3.0"
+      - name: Install tsfile extra (py3.14 only)
+        run: uv pip install --system "datasets[tsfile] @ ."
       - name: Print dependencies
         run: uv pip list
       - name: Test with pytest
--- a/README.md
+++ b/README.md
@@ -84,6 +84,9 @@ pip install datasets[vision]
 # For PDFs/NIfTI (pdfplumber, nibabel)
 pip install datasets[pdfs,nibabel]
 
+# For Apache TsFile (time-series)
+pip install datasets[tsfile]
+
 # For PyTorch/TensorFlow/JAX integration
 pip install datasets[torch,tensorflow,jax]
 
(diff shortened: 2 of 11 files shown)


AxelNoun:
Note for the record, since this branch is closed but still linked from #8499.

Two things in it did not survive further checking:

1. The commit message on d11450c says the unguarded SDK import is "also why no
`format:tsfile` facet is derived on the Hub". That is wrong. `tsfile` is not a registered
value in the Hub's format tag taxonomy at all, which accounts for the empty Format bucket
on its own, independently of dataset-viewer job status. Shipping the SDK to the workers
would not make the facet appear.

2. The `tsfile` extra as written here only resolves on Python 3.14 (the SDK requires
`pyarrow<20` on 3.10-3.13, `datasets` requires `pyarrow>=21`). Because that is an
environment marker, `pip install 'datasets[tsfile]'` returns success while installing
nothing on 3.10-3.13, so an error message pointing at that command sends the user in a
loop. Any guard needs to be version-aware.

Full analysis and measurements: https://github.com/huggingface/datasets/issues/8499#issuecomment-5420616135

PR #8515, opened 2026-08-25
Fix offline cache lookup with custom data files

### LINKED ISSUE #7551: Issue with offline mode and partial dataset cached
### Describe the bug

Hi, 

a issue related to #4760 here when loading a single file from a dataset, unable to access it in offline mode afterwards

```python
import os
# os.environ["HF_HUB_OFFLINE"] = "1"
os.environ["HF_TOKEN"] = "xxxxxxxxxxxxxx"

import datasets

dataset_name = "uonlp/CulturaX"
data_files = "fr/fr_part_00038.parquet"

ds = datasets.load_dataset(dataset_name, split='train', data_files=data_files)
print(f"Dataset loaded   : {ds}")
```
Once the file has been cached, I rerun whit the HF_HUB_OFFLINE activated an get this error : 
```
ValueError: Couldn't find cache for uonlp/CulturaX for config 'default-1e725f978350254e'
Available configs in the cache: ['default-2935e8cdcc21c613']
```

### Expected behavior

Should be able to access the previously cached files

### Environment info

- `datasets` version: 3.2.0
- Platform: Linux-5.4.0-215-generic-x86_64-with-glibc2.31
- Python version: 3.12.0
- `huggingface_hub` version: 0.27.0
- PyArrow version: 19.0.0
- Pandas version: 2.2.2
- `fsspec` version: 2024.3.1

mdeff:
## Summary

Fixes #7551.

Datasets loaded from the Hub with a `data_files` override could not be reloaded in offline mode.

During online loading, relative data file patterns are resolved to revision-pinned Hub URLs before computing the cache configuration ID. During offline lookup, only the original unresolved patterns are available, producing a different ID.

This change falls back to the sole cached custom configuration matching the requested base configuration name. If multiple candidates exist, it preserves the existing error rather than choosing ambiguously.

## Tests

- `pytest tests/packaged_modules/test_cache.py -q`
- `ruff check src/datasets/packaged_modules/cache/cache.py tests/packaged_modules/test_cache.py`
- Verified offline loading against an existing partial `allenai/c4` cache without modifying or aliasing its cache directory

### DIFF
--- a/src/datasets/packaged_modules/cache/cache.py
+++ b/src/datasets/packaged_modules/cache/cache.py
@@ -22,6 +22,15 @@ def _get_modification_time(cached_directory_path):
     return (Path(cached_directory_path)).stat().st_mtime
 
 
+def _get_config_name(cached_directory_path: str) -> Optional[str]:
+    try:
+        return json.loads(Path(cached_directory_path, "dataset_info.json").read_text(encoding="utf-8"))[
+            "config_name"
+        ]
+    except (FileNotFoundError, KeyError, json.JSONDecodeError):
+        return None
+
+
 def _find_hash_in_cache(
     dataset_name: str,
     config_name: Optional[str],
@@ -53,6 +62,23 @@ def _find_hash_in_cache(
             == Path(cached_directory_path).parts[-3]  # no extra params => config_id == config_name
         )
     ]
+    if not cached_directory_paths and "data_files" in config_kwargs:
+        # Hub data files are resolved to revision-pinned URLs before their config ID is computed.
+        # In offline mode only the unresolved user patterns are available, so the exact config ID
+        # cannot be reconstructed. Falling back is safe if only one cached custom configuration
+        # corresponds to the requested base config name.
+        requested_config_name = config_name or "default"
+        matching_directory_paths = [
+            cached_directory_path
+            for cached_directory_path in glob.glob(
+                os.path.join(cached_datasets_directory_path_root, "*", "*", "*")
+            )
+            if os.path.isdir(cached_directory_path)
+            and _get_config_name(cached_directory_path) == requested_config_name
+        ]
+        matching_config_ids = {Path(path).parts[-3] for path in matching_directory_paths}
+        if len(matching_config_ids) == 1:
+            cached_directory_paths = matching_directory_paths
     if not cached_directory_paths:
         cached_directory_paths = [
             cached_directory_path
--- a/tests/packaged_modules/test_cache.py
+++ b/tests/packaged_modules/test_cache.py
@@ -1,3 +1,4 @@
+import json
 from pathlib import Path
 
 import pytest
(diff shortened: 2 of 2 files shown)


src/datasets/packaged_modules/cache/cache.py
```diff
@@ -53,6 +62,23 @@ def _find_hash_in_cache(
...
+        # cannot be reconstructed. Falling back is safe if only one cached custom configuration
+        # corresponds to the requested base config name.
+        requested_config_name = config_name or "default"
+        matching_directory_paths = [
+            cached_directory_path
+            for cached_directory_path in glob.glob(
+                os.path.join(cached_datasets_directory_path_root, "*", "*", "*")
+            )
+            if os.path.isdir(cached_directory_path)
+            and _get_config_name(cached_directory_path) == requested_config_name
```
ebarkhordar, line 77: I ran this against a79f45e0 in a clean container (python:3.11-slim, datasets 4.4.1 for deps, source on PYTHONPATH).

The comment above describes the fallback as picking the sole cached custom configuration, but this filter matches on `config_name` alone, and a plain cached config satisfies it too: a directory left by `load_dataset("namespace/dataset", "en")` has config_id `en` and `dataset_info.json["config_name"]` of `en`. Two consequences, both from that.

If only the plain cache is present, a call that explicitly passes `data_files` now resolves to it, where main raised. That cache was built from the dataset's default files, so the wrong data loads with only the usual "Found the latest cached dataset configuration" warning:

```
A. only the plain cache exists (config_id == config_name)
   base 836b82e0: ValueError (Couldn't find cache for namespace/dataset for config 'en-5be9c333aa4ee8e3')
   head a79f45e0: RESOLVED -> en

C. plain and custom both cached
   head a79f45e0: ValueError (Couldn't find cache ...)
```

In C the plain directory is also counted in `matching_config_ids`, so the set holds two entries and the fallback never fires. That is the case where someone loaded the dataset normally before overriding `data_files`, which I would expect to be the common way to arrive here.

Excluding the base config from the candidates covers both:

```python
and Path(cached_directory_path).parts[-3] != requested_config_name
and _get_config_name(cached_directory_path) == requested_config_name
```

With that line, A raises again, C resolves to the custom config, and your two new tests still pass.

I did not test against a real Hub cache. My directories were seeded by hand in the layout your tests use.

caiotheodoro: Excluding the base config from the candidates does not close this — two non-plain custom configs that share `config_name` still collide silently when only one happens to be cached. Cache `en-hashA` from an earlier `data_files=pattern-A` run, then offline-request `en` with an unrelated `data_files=pattern-B` that was never cached: `matching_config_ids` has exactly one entry (`en-hashA`), so it resolves without ever checking B against it. Verified against a79f45e0: the mismatched cache loads with no error, only the usual "Found the latest cached dataset configuration" warning. Matching needs to key on something derived from the requested `data_files` value, not `config_name` alone.

ebarkhordar: You are right, and my line only removes the plain directory. Reading the filter again, `matching_config_ids` still holds a single entry when one custom `en-<hash>` is cached and the request carries `data_files` that were never cached, so it resolves against a cache built from different patterns with only the usual warning. I did not run that case. Agreed the match has to key on something derived from the requested `data_files`, not on `config_name` alone.

PR #8518, opened 2026-08-26
Validate features when a cast only changes schema metadata

### LINKED ISSUE #8494: cast_column to ClassLabel silently accepts out-of-range label indices
### Describe the bug

`Dataset.cast_column` / `Dataset.cast` to a `ClassLabel` column silently keeps label indices that are out of range for the target number of classes. The corruption only shows up later — e.g. on the first `int2str` call — or silently trains on wrong labels:

```python
from datasets import Dataset, ClassLabel

d = Dataset.from_dict({"label": [5, 1]})
casted = d.cast_column("label", ClassLabel(names=["neg", "pos", "oth"]))  # no error, num_classes=3
print(casted["label"])                    # [5, 1]
casted.features["label"].int2str(5)       # ValueError: Invalid integer class label 5
```

The write path does validate the same data — `Dataset.from_dict(..., features=Features({"label": ClassLabel(names=[...])}))` raises `ValueError: Class label 5 greater than configured num_classes 3` — so the cast path bypasses a check the writer performs.

Cause: `table_cast` in `src/datasets/table.py` only runs the feature-level casts when the arrow schema changes, and `pa.Schema.__eq__` ignores metadata. Casting int64 storage to `ClassLabel` leaves the arrow schema unchanged and only alters the metadata, so it takes the `replace_schema_metadata` branch and `ClassLabel.cast_storage`, which does the range check, never runs.

### Expected behavior

`cast_column`/`cast` should raise the same `ValueError` as the write path when a label index is `>= num_classes`.

Kayvan-Zahiri:
Closes #8494.

`pa.Schema.__eq__` ignores metadata, so casting int64 storage to a `ClassLabel`
leaves the arrow schema equal and `table_cast` took the
`replace_schema_metadata` branch. That skips `cast_array_to_feature`, and with
it the range check in `ClassLabel.cast_storage`, so an out-of-range label index
was kept and only surfaced later on the first `int2str`, or not at all.

The clearest argument for treating this as a bug is that the write path already
rejects the same data, so the two disagreed on identical input:

```python
Dataset.from_dict({"label": [5, 1]}, features=Features({"label": ClassLabel(names=[...])}))
# ValueError: Class label 5 greater than configured num_classes 3

Dataset.from_dict({"label": [5, 1]}).cast_column("label", ClassLabel(names=[...]))
# accepted, values stay [5, 1]
```

After the change the cast raises that same `ValueError`, and in-range casts are
untouched.

`table_cast` now runs the feature cast whenever the schema *or* its metadata
differs. `cast_table_to_schema` builds the result with the target schema, so it
already does what the metadata-only branch did.

Worth flagging for review: this means a metadata-only cast now walks the columns
through `cast_array_to_feature` instead of swapping the metadata. That is the
point, since the validation lives there, but it is the one behaviour change a
reviewer should weigh. `tests/test_table.py` and `tests/test_arrow_dataset.py`
are unchanged at 640 passed, 1 pre-existing failure
(`test_dataset_to_iterable_dataset`, which fails the same way on a clean
checkout here). `ruff check` and `ruff format` are clean.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -2429,7 +2429,7 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
         dataset = self.with_format("arrow")
         # capture the PyArrow version here to make the lambda serializable on Windows
         dataset = dataset.map(
-            partial(table_cast, schema=schema),
+            partial(table_cast, schema=schema, validate_features=True),
             batched=True,
             batch_size=batch_size,
             keep_in_memory=keep_in_memory,
--- a/src/datasets/table.py
+++ b/src/datasets/table.py
@@ -2360,7 +2360,7 @@ def embed_table_storage(table: pa.Table, token_per_repo_id=None, local_files: bo
     return pa.Table.from_arrays(arrays, schema=features.arrow_schema)
 
 
-def table_cast(table: pa.Table, schema: pa.Schema):
+def table_cast(table: pa.Table, schema: pa.Schema, validate_features: bool = False):
     """Improved version of `pa.Table.cast`.
 
     It supports casting to feature types stored in the schema metadata.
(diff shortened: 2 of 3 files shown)


ebarkhordar (commented):
You asked for a reviewer to weigh the metadata-only branch, so I measured it against `1426f0a2` in a clean `python:3.11-slim` container (pyarrow 25.0.1).

That branch is not a rare case. A plain parquet load with `features=` takes it on every `table_cast` call:

```
rows loaded: 200000
   20  METADATA ONLY -> was replace_schema_metadata, now full cast
```

The table read from storage carries no `huggingface` metadata while the target schema does, and the arrow types already match, so the metadata comparison is what fires. Every packaged module's `_cast_table` sits on this path, plus `arrow_writer._write_table`.

Cost of one such call, 20 int64 columns by 50k rows:

```
pre-PR  replace_schema_metadata :    0.006 ms/call
PR head cast_table_to_schema    :    2.263 ms/call
```

`cast_table_to_schema` also calls `Features.from_arrow_schema(schema)`, so the metadata JSON is parsed once per batch too.

None of this makes the fix wrong, and the gap in #8494 is real. It does mean the cost lands on every loader rather than on `cast_column`, where the reported bug is. Scoping it to the user-facing cast path would close #8494 without paying this on load.

I did not run the full test suite.

Kayvan-Zahiri:
You're right, and the measurement settles it. Fixed in 532b0ff.

I profiled where the 2.263 ms actually goes, on 20 int64 columns by 50k rows:

```
replace_schema_metadata         0.002 ms
Features.from_arrow_schema(tgt) 0.294 ms
cast_table_to_schema (full)     0.373 ms
```

So the cast itself is ~0.08 ms and the metadata JSON round trip is the rest. That also kills the obvious middle ground: I tried comparing decoded features and only casting when they differ, and it came out at 0.412 ms, worse than just casting, because it parses the metadata twice.

So there is no cheap way to tell the two cases apart from inside `table_cast`, and scoping it the way you suggested is the right call. `table_cast` now takes `validate_features`, defaulting to False, and only `Dataset.cast` passes True. Loaders are back on `replace_schema_metadata`: the metadata-only path measures 0.003 ms/call again, against 0.389 on the previous head.

#8494 stays fixed, since `cast_column` routes through `Dataset.cast`. The regression test still fails when I revert the call site.

tests/test_table.py 278 passed, tests/features + test_arrow_writer.py 451 passed, cast tests 5 passed.

ebarkhordar:
That settles it, and thanks for measuring the middle ground rather than assuming it. `validate_features` defaulting to False with only `Dataset.cast` passing True keeps #8494 fixed without putting the metadata round trip back on the loader path, which was the only thing I was worried about.

PR #8519, opened 2026-08-26
Return numpy scalars from numpy formatting and keep duration dtypes

### LINKED ISSUE #8500: with_format("numpy") returns bool/datetime values as 0-d unhashable arrays and casts duration columns to int64

Kayvan-Zahiri:
Closes #8500.

Two problems in `NumpyFormatter`, both falling out of numpy's type hierarchy.

**Scalars.** `np.bool_` and `np.datetime64` are not `np.number`, so `_tensorize`
did not early-return them and `_recursive_tensorize` ran them through
`__array__` first. Row access handed back 0-d `ndarray`s, which are unhashable
and fail `isinstance(x, np.bool_)`. Both checks now use `np.generic`, which
covers every numpy scalar; `np.character` and `np.number` are already subclasses
of it, so the two existing branches collapse into it cleanly.

**Durations.** `np.issubdtype(np.dtype("timedelta64[us]"), np.integer)` is
`True`, so the int64 default dtype applied to duration columns. That is why the
same column reported `timedelta64` by row and `int64` by batch. The default now
skips timedelta64.

Before and after, on `{"b": True, "t": datetime, "d": timedelta, "i": 3}`:

```
row    b: ndarray(0-d, unhashable) -> np.bool_
       t: ndarray(0-d, unhashable) -> np.datetime64
batch  d: int64                    -> timedelta64[us]
```

`int64` and `float32` defaults for ordinary integer and float columns are
unchanged, and so are strings.

Two tests added, both failing without the change. `tests/test_formatting.py`,
`tests/test_arrow_dataset.py` and `tests/test_table.py` stay at 664 passed with
the same single pre-existing failure (`test_dataset_to_iterable_dataset`, which
fails identically on a clean checkout here). `ruff check` and `ruff format`
clean.

Separate from #8518, which touches `table_cast`.

### DIFF
--- a/src/datasets/formatting/np_formatter.py
+++ b/src/datasets/formatting/np_formatter.py
@@ -48,12 +48,20 @@ class NumpyFormatter(TensorFormatter[Mapping, np.ndarray, Mapping]):
             return value
         elif isinstance(value, (np.character, np.ndarray)) and np.issubdtype(value.dtype, np.character):
             return value
-        elif isinstance(value, np.number):
+        elif isinstance(value, np.generic):
+            # np.bool_ and np.datetime64 are not np.number, so they used to fall
+            # through to np.asarray and come back as unhashable 0-d arrays.
             return value
 
         default_dtype = {}
 
-        if isinstance(value, np.ndarray) and np.issubdtype(value.dtype, np.integer):
+        if (
+            isinstance(value, np.ndarray)
+            and np.issubdtype(value.dtype, np.integer)
+            and not np.issubdtype(value.dtype, np.timedelta64)
+        ):
+            # np.issubdtype counts timedelta64 as an integer subtype, which
+            # would drop the unit and hand back a plain int64 column.
             default_dtype = {"dtype": np.int64}
         elif isinstance(value, np.ndarray) and np.issubdtype(value.dtype, np.floating):
             default_dtype = {"dtype": np.float32}
@@ -86,7 +94,7 @@ class NumpyFormatter(TensorFormatter[Mapping, np.ndarray, Mapping]):
 
             if isinstance(data_struct, torch.Tensor):
                 return self._tensorize(data_struct.detach().cpu().numpy()[()])
-        if hasattr(data_struct, "__array__") and not isinstance(data_struct, (np.ndarray, np.character, np.number)):
+        if hasattr(data_struct, "__array__") and not isinstance(data_struct, (np.ndarray, np.generic)):
             data_struct = data_struct.__array__()
         # support for nested types like struct of list of struct
         if isinstance(data_struct, np.ndarray):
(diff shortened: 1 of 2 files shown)


ebarkhordar (commented):
#8500 has three open PRs against `src/datasets/formatting/np_formatter.py`, so I fetched each head's copy of that file from this repo and diffed them.

```
#8503 5a97bc12  sha256 7210141c7c3c9f13
#8504 86105122  sha256 0cbf9e401c97e48c
#8519 1335ea15  sha256 99d8e045fc467e27

diff 8504 8519 -> 4 added lines, all comments (the np.generic note
                  and the timedelta64 note). No executable difference.
diff 8503 8504 -> 1 line. 8503 leaves _recursive_tensorize checking
                  (np.ndarray, np.character, np.number).
```

So the code change here is the same one #8504 makes, and #8503 is that change without the `_recursive_tensorize` line.

Where this PR does differ is the tests. It also asserts `format_column`, checks each scalar is not an `ndarray`, calls `hash()` on it, and covers a plain int column. #8504's single test asserts the three isinstance checks and the batch duration dtype only.

I have no view on which should land. Flagging it so the same diff does not get reviewed three times.

PR #8520, opened 2026-08-26
Take the image dtype from Arrow storage in Image.cast_storage

### LINKED ISSUE #8511: `Image.cast_storage` warns about downcasting int64 to uint8 for storage that is already uint8
### Describe the bug

Casting a nested-list column whose Arrow value type is already `uint8` to `Image()` emits a spurious warning:

```
UserWarning: Downcasting array dtype int64 to uint8 to be compatible with 'Pillow'
```

No downcast is actually needed. The Arrow storage type is `list<item: list<item: list<item: uint8>>>`, so the dtype information is present and correct; the `int64` is introduced by `cast_storage` itself.

`Image.cast_storage`, list branch (`src/datasets/features/image.py#L264-L268`):

```python
elif pa.types.is_list(storage.type):
    bytes_array = pa.array(
        [encode_np_array(np.array(arr))["bytes"] if arr is not None else None for arr in storage.to_pylist()],
        type=pa.binary(),
    )
```

`storage.to_pylist()` materialises the values as plain Python `int`s, discarding the Arrow value type, and `np.array(arr)` on nested lists of Python `int`s then defaults to `int64`. `encode_np_array` (`src/datasets/features/image.py#L389-L397`) compares that against `|u1` and warns:

```python
# Multi-channel array case (only np.dtype("|u1") is allowed)
if array.shape[2:]:
    ...
    dest_dtype = np.dtype("|u1")
    if dtype != dest_dtype:
        warnings.warn(f"Downcasting array dtype {dtype} to {dest_dtype} to be compatible with 'Pillow'")
```

So the values round-trip `uint8` → Python `int` → `int64` → `uint8`, and the user is warned about a downcast their data never needed. The encoded image is correct; the problem is the misleading message.

```python
import warnings

import numpy as np
from datasets import Dataset
from datasets.features import Image

ds = Dataset.from_list(
    [
        {
            "prompt": [{"role": "user", "content": "x"}],
            "image": np.random.uniform(low=0.0, high=255.0, size=(64, 64, 3)).astype(np.uint8),
        }
        for _ in range(4)
    ]
)
print("inferred features:", ds.features)
print("arrow schema:", ds.data.schema)
with warnings.catch_warnings(record=True) as w:
    warnings.simplefilter("always")
    ds2 = ds.cast_column("image", Image())
    _ = ds2[0]
    print("WARNINGS:", [(x.category.__name__, str(x.message)) for x in w])
```

Output:

```
inferred features: {'prompt': List({'role': Value('string'), 'content': Value('string')}), 'image': List(List(List(Value('uint8'))))}
arrow schema: prompt: list<item: struct<role: string, content: string>>
  child 0, item: struct<role: string, content: string>
      child 0, role: string
      child 1, content: string
image: list<item: list<item: list<item: uint8>>>
  child 0, item: list<item: list<item: uint8>>
      child 0, item: list<item: uint8>
          child 0, item: uint8
-- schema metadata --
huggingface: '{"info": {"features": {"prompt": {"feature": {"role": {"dty' + 241
WARNINGS: [('UserWarning', "Downcasting array dtype int64 to uint8 to be compatible with 'Pillow'"), ('UserWarning', "Downcasting array dtype int64 to uint8 to be compatible with 'Pillow'"), ('UserWarning', "Downcasting array dtype int64 to uint8 to be compatible with 'Pillow'"), ('UserWarning', "Downcasting array dtype int64 to uint8 to be compatible with 'Pillow'")]
```

Note the contradiction between the schema (`uint8`) and the warning (`int64`). The `prompt` column is irrelevant to the issue; it is only there because this is the shape the reproduction was found in.

Both `5.0.1` and current `main` are affected — `cast_storage` is identical in the two.

### Expected behavior

No warning: the Arrow storage type already is a valid image array dtype, so nothing is being lost.

`cast_storage` knows the value type from `storage.type`, so it could build the array with it rather than letting `numpy` guess from Python `int`s — unwrapping the nested `list<...>` types down to the innermost value type and passing it as `dtype` to `np.array`. Genuine downcasts (e.g. list storage whose value type really is `int64`) would keep warning as they do today.

I'm happy to send a PR if this sounds right to you.

### Environment info

- `datasets` 5.0.2.dev0 (editable install of `main` @ 836b82e0544cabf6474b25ade131b4d21e570373); also reproduced with `datasets` 5.0.1 in CI
- Python 3.12.9, Linux
- `numpy` 2.4.4, `pyarrow` 24.0.0, `Pillow` 12.1.1
- CI occurrence with `datasets==5.0.1`, `numpy==2.5.2`, `pyarrow==25.0.1`, `pillow==12.3.0`, Python 3.12: https://github.com/huggingface/trl/actions/runs/32695065358/job/97335526967 (tracked in https://github.com/huggingface/trl/issues/6888)

Kayvan-Zahiri:
Closes #8511.

`storage.to_pylist()` materializes the values as plain Python `int`s, which drops the Arrow value type. `np.array(arr)` on nested lists of Python ints then defaults to `int64`, and `encode_np_array` warns about a downcast to `uint8` that the storage never needed:

```python
arr = np.random.randint(0, 256, (2, 4, 4, 3), dtype=np.uint8)
storage = pa.array(arr.tolist(), type=pa.list_(pa.list_(pa.list_(pa.uint8()))))
Image().cast_storage(storage)
# UserWarning: Downcasting array dtype int64 to uint8 to be compatible with 'Pillow'  (x2)
```

Reading the dtype off the Arrow value type instead keeps `uint8` storage quiet. Storage that genuinely needs a downcast still warns, and still names its own dtype:

| storage | before | after |
|---|---|---|
| `uint8` | warns | silent |
| `int64` | warns | warns |
| `int32` | warns | warns |

The descent loop handles `large_list` and `fixed_size_list` too, and the dtype is only taken for integer and floating value types so anything else falls back to the previous inference.

Encoded bytes are unchanged: decoded pixels are identical to the source array for both the `uint8` and `int64` paths.

Test is parametrized over the three cases above; the `uint8` one fails on `main` and passes here.

### DIFF
--- a/src/datasets/features/image.py
+++ b/src/datasets/features/image.py
@@ -262,8 +262,26 @@ class Image:
                 path_array = pa.array([None] * len(storage), type=pa.string())
             storage = pa.StructArray.from_arrays([bytes_array, path_array], ["bytes", "path"], mask=storage.is_null())
         elif pa.types.is_list(storage.type):
+            # to_pylist() returns plain Python ints, so np.array would infer int64
+            # and encode_np_array would warn about a downcast the storage does not
+            # need. Take the dtype from the Arrow value type instead.
+            value_type = storage.type
+            while (
+                pa.types.is_list(value_type)
+                or pa.types.is_large_list(value_type)
+                or pa.types.is_fixed_size_list(value_type)
+            ):
+                value_type = value_type.value_type
+            dtype = (
+                value_type.to_pandas_dtype()
+                if pa.types.is_integer(value_type) or pa.types.is_floating(value_type)
+                else None
+            )
             bytes_array = pa.array(
-                [encode_np_array(np.array(arr))["bytes"] if arr is not None else None for arr in storage.to_pylist()],
+                [
+                    encode_np_array(np.array(arr, dtype=dtype))["bytes"] if arr is not None else None
+                    for arr in storage.to_pylist()
+                ],
                 type=pa.binary(),
             )
             path_array = pa.array([None] * len(storage), type=pa.string())
--- a/src/datasets/features/image.py
+++ b/src/datasets/features/image.py
@@ -277,6 +277,11 @@ class Image:
                 if pa.types.is_integer(value_type) or pa.types.is_floating(value_type)
                 else None
             )
+            # encode_np_array only narrows within a kind, and int8 and float16 have
+            # no valid target to narrow to, so keep a dtype it can already encode
+            # and otherwise leave the inference alone.
+            if dtype is not None and np.dtype(dtype) not in _VALID_IMAGE_ARRAY_DTPYES:
+                dtype = None
             bytes_array = pa.array(
                 [
                     encode_np_array(np.array(arr, dtype=dtype))["bytes"] if arr is not None else None
--- a/tests/features/test_image.py
+++ b/tests/features/test_image.py
@@ -712,3 +712,22 @@ def test_encode_np_array(array, dtype_cast, expected_image_format):
     decoded_image = Image().decode_example(encoded_image)
     assert decoded_image.format == expected_image_format
     np.testing.assert_array_equal(np.array(decoded_image), array)
+
+
+@require_pil
+@pytest.mark.parametrize(
+    "value_type, warns",
+    [(pa.uint8(), False), (pa.int64(), True), (pa.int32(), True)],
+)
+def test_image_cast_storage_list_keeps_arrow_dtype(value_type, warns):
+    array = np.random.default_rng(0).integers(0, 256, (2, 4, 4, 3), dtype=np.uint8)
+    storage = pa.array(array.tolist(), type=pa.list_(pa.list_(pa.list_(value_type))))
+    if warns:
+        with pytest.warns(UserWarning, match="Downcasting array dtype"):
+            casted = Image().cast_storage(storage)
+    else:
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
+            casted = Image().cast_storage(storage)
+    decoded = np.stack([np.array(Image().decode_example(row)) for row in casted.to_pylist()])
+    np.testing.assert_array_equal(decoded, array)
--- a/tests/features/test_image.py
+++ b/tests/features/test_image.py
@@ -731,3 +731,17 @@ def test_image_cast_storage_list_keeps_arrow_dtype(value_type, warns):
             casted = Image().cast_storage(storage)
     decoded = np.stack([np.array(Image().decode_example(row)) for row in casted.to_pylist()])
     np.testing.assert_array_equal(decoded, array)
+
+
+@require_pil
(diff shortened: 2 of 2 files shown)


src/datasets/features/image.py
```diff
@@ -262,8 +262,26 @@ def cast_storage(self, storage: Union[pa.StringArray, pa.StructArray, pa.ListArr
...
+            # and encode_np_array would warn about a downcast the storage does not
+            # need. Take the dtype from the Arrow value type instead.
+            value_type = storage.type
+            while (
+                pa.types.is_list(value_type)
+                or pa.types.is_large_list(value_type)
+                or pa.types.is_fixed_size_list(value_type)
+            ):
+                value_type = value_type.value_type
+            dtype = (
```
ebarkhordar, line 275: I ran this against 21717cd7 in a clean `python:3.11-slim` container. Taking the dtype from Arrow also narrows it, and `encode_np_array` cannot always land a narrow one: its downcast loop halves the itemsize within the same kind, and `|i1` is deliberately excluded from `_VALID_IMAGE_ARRAY_DTPYES`. int8 goes from 1 to 0 with no candidate left, and float16 halves to `<f1`, which is not a numpy dtype at all. Both are single-channel 2D storage, and both worked on main because `to_pylist()` widened them to int64 or float64 first.

```python
ds = Dataset.from_dict({"col": [[[1, 2], [3, 4]]]},
                       features=Features({"col": Sequence(Sequence(Value("int8")))}))
ds.cast_column("col", Image())
# main cd6a3406: ok, mode I
# 21717cd7:      TypeError: Cannot downcast dtype int8 to a valid image dtype
# same with float16: TypeError: data type '<f1' not understood
```

Guarding with `if dtype is not None and np.dtype(dtype) not in _VALID_IMAGE_ARRAY_DTPYES: dtype = None` restores both and still leaves your three parametrized cases passing. I did not check the multi-channel path for these two types, which takes the other branch and looks unaffected.

Kayvan-Zahiri: You're right, and thanks for the container repro. Fixed in cecc936.

Added the guard you suggested:

```python
if dtype is not None and np.dtype(dtype) not in _VALID_IMAGE_ARRAY_DTPYES:
    dtype = None
```

int8 and float16 are back to main's behavior (mode `I` and `F`), uint8 still encodes as `L` with no downcast warning, and there are two new parametrized cases covering both.

On the multi-channel path you didn't check: it's identical before and after. int8 gives RGB either way, and float16 raises the same TypeError either way, so that one is pre-existing. Its message is misleading though, it says "Only None is supported" because `dest_dtype` is still `None` when the multi-channel branch raises.

ebarkhordar: Thanks, cecc936 is the guard I meant, and the two parametrized cases pin both dtypes. Agreed that the multi-channel TypeError is pre-existing and separate from this PR, including the message naming None when dest_dtype has not been set yet.

PR #8524, opened 2026-08-28
Return None for a null row of a dynamic-shape ArrayXD column

Kayvan-Zahiri:
`ArrayExtensionArray.to_pylist` builds the python values from `to_numpy`, where a null row is a bare `np.nan`. That is fine for the numpy format but not for the python one, and the guard added in #8363 only covered a fixed first dimension.

So for a column like `Array2D(dtype="int32", shape=(None, 2))` with a `None` row:

```python
ds[1]        # {'foo': nan}   instead of {'foo': None}
ds.to_dict() # AttributeError: 'float' object has no attribute 'tolist'
```

The same crash hits `to_list`, `ds[:]`, `Dataset.iter`, `map` and `filter`. The identical column declared with a fixed shape returns `None` correctly. Dynamic first dimensions are a documented feature and `test_array_xd_with_none` already covers them.

The null guard is really about nulls, not about the shape, so it now runs before the numpy conversion and covers both.

`PandasArrayExtensionArray.isna` met the same scalar nan from the other side: `pd.isna` returns a plain bool for it and bool has no `.any()`, which broke `Dataset.to_csv` and `DataFrame.isna()`.

Two regression tests added next to the existing ones. The numpy, pandas and arrow formats are unchanged, verified across all twelve ArrayXD value dtypes.

### DIFF
--- a/src/datasets/features/features.py
+++ b/src/datasets/features/features.py
@@ -948,9 +948,14 @@ class PandasArrayExtensionArray(PandasExtensionArray):
         return self._data.nbytes
 
     def isna(self) -> np.ndarray:
-        # A null row of a dynamic-shape ArrayXD column is a scalar np.nan rather than an
-        # array, and the plain bool `pd.isna` returns for it has no `.any()`.
-        return np.array([np.any(pd.isna(arr)) for arr in self._data])
+        if self._data.dtype == object:
+            # Dynamic shape: a missing row is a scalar sentinel, present rows are
+            # arrays. isna is a per-value mask, so a NaN inside a present row must
+            # not mark the row missing.
+            return np.array([not isinstance(arr, np.ndarray) and bool(pd.isna(arr)) for arr in self._data])
+        # Fixed shape: every row is an array and the null row is materialised as
+        # NaNs, so the validity bitmap is already gone by this point.
+        return np.array([pd.isna(arr).any() for arr in self._data])
 
     def __setitem__(self, key: Union[int, slice, np.ndarray], value: Any) -> None:
         raise NotImplementedError()
--- a/src/datasets/features/features.py
+++ b/src/datasets/features/features.py
@@ -838,17 +838,16 @@ class ArrayExtensionArray(pa.ExtensionArray):
         return numpy_arr
 
     def to_pylist(self, maps_as_pydicts: Optional[Literal["lossy", "strict"]] = None):
+        if self.storage.null_count:
+            # to_numpy writes np.nan in the null rows and the python path cannot undo that: a
+            # fixed shape has its whole column cast to float64, losing integers above 2**53, and
+            # a dynamic first dimension leaves a bare float where a list belongs. The nested-list
+            # storage already carries the exact values and None for the null rows, so use it.
+            return self.storage.to_pylist()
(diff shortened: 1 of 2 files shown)


Jokasa7 (commented):
I found one correctness issue in the pandas-formatting part of this fix. The Python-format change behaves correctly for the tested null-row cases, but the scalar-safe `isna` reduction still conflates an inner NaN with top-level `ArrayXD` nullness, so the newly working `to_csv` path can silently blank a present array value. Details inline.

src/datasets/features/features.py
```diff
@@ -949,7 +948,9 @@ def nbytes(self) -> int:
         return self._data.nbytes
 
     def isna(self) -> np.ndarray:
-        return np.array([pd.isna(arr).any() for arr in self._data])
+        # A null row of a dynamic-shape ArrayXD column is a scalar np.nan rather than an
+        # array, and the plain bool `pd.isna` returns for it has no `.any()`.
+        return np.array([np.any(pd.isna(arr)) for arr in self._data])
```
Jokasa7: **[P2] Preserve present ArrayXD rows that contain NaN**

For a dynamic/ragged float column, `_data` is an object array where a top-level null is scalar `np.nan` but each present row is an ndarray. Reducing `pd.isna(arr)` with `np.any` therefore marks a present row missing as soon as one element is NaN. With `Array2D((None, 2), "float32")` and `[[[1.0, np.nan]], None, [[5.0, 6.0]]]`, Arrow's top-level null mask is `[False, True, False]`, but this line returns `[True, True, False]`; on this head, `to_csv()` writes both of the first two cells as empty strings. Since `ExtensionArray.isna()` is a per-value mask, could the object/ragged path identify only the scalar missing sentinel (or preserve Arrow's validity bitmap) and add an inner-NaN regression test?

Kayvan-Zahiri: Good catch, that is silent data loss and you are right that `isna` is a per-value mask. Fixed in 9746085.

Your repro, before and after, against Arrow's own validity:

```
                arrow      isna (before)   isna (after)
dynamic   [F, T, F]        [T, T, F]       [F, T, F]
csv row 0 was ""                           now [[ 1. nan]]
```

The dynamic path is object dtype, where a missing row is a scalar sentinel and present rows are arrays, so it now tests for the sentinel rather than reducing over the row:

```python
if self._data.dtype == object:
    return np.array([not isinstance(arr, np.ndarray) and bool(pd.isna(arr)) for arr in self._data])
return np.array([pd.isna(arr).any() for arr in self._data])
```

I left the fixed-shape branch alone deliberately. There `_data` is a float ndarray and the null row is already materialised as NaNs, so nothing at that point can tell a null row from a row that happens to be all NaN. Preserving the validity bitmap would fix it properly, but that is a wider change than this PR and I would rather not fold it in. Say the word if you want it as a follow-up.

Added `test_array_xd_dynamic_shape_isna_keeps_rows_with_inner_nan`, which pins isna against `column[i].is_valid` and asserts the present NaN row survives `to_csv`. It fails on the previous head.

tests/features + tests/test_table.py: 646 passed, 106 skipped. ruff clean.

PR #8535, opened 2026-08-30
Validate delimiter and repeated column_names in ConllConfig

LeSingh1:
### The bug

Two `ConllConfig` arguments reach low-level code before anything validates them, so the resulting error never mentions the argument that is wrong. Both arrive wrapped in `DatasetGenerationError: An error occurred while generating the dataset`:

```python
load_dataset("conll", data_files=path, column_names=["a", "a"])
# caused by KeyError: 'Field "a" exists 2 times in schema'      <- from pa.Table.from_arrays

load_dataset("conll", data_files=path, delimiter="")
# caused by ValueError: empty separator                          <- from str.split("")
```

Neither says `column_names` or `delimiter`.

### The fix

Reject both in `__post_init__`:

```
ValueError: `column_names` has repeated entries: ['a']
ValueError: `delimiter` must not be empty; pass None to split on any whitespace
```

The delimiter message points at `None`, which is the documented way to split on any whitespace, since an empty string is the natural but wrong way to ask for that.

### What I deliberately left alone

Empty `column_names` also produces a wrapped error, but `_generate_tables` already raises a clear "must be a non-empty list" for it, and `test_generate_tables_empty_column_names_raises` asserts it comes from there. I had initially validated it in `__post_init__` too, which moved the raise to construction and broke that test — so I dropped it rather than change the existing expectation. The case is already handled well; only the two above were unguarded.

### Verification

`tests/packaged_modules/test_conll.py` is 17 passed. The three new tests fail on `main` and pass with the change. Loading still works unchanged for the default column names, custom column names, and an explicit `delimiter=" "`.

### DIFF
--- a/src/datasets/packaged_modules/conll/conll.py
+++ b/src/datasets/packaged_modules/conll/conll.py
@@ -65,6 +65,18 @@ class ConllConfig(datasets.BuilderConfig):
     skip_docstart: bool = True
     comment_prefix: Optional[str] = None
 
+    def __post_init__(self):
+        super().__post_init__()
+        # Checked here rather than during generation so the error names the argument
+        # instead of arriving wrapped in a DatasetGenerationError.
+        duplicate_columns = sorted({c for c in self.column_names if self.column_names.count(c) > 1})
+        if duplicate_columns:
+            # Otherwise pa.Table.from_arrays fails with KeyError: Field "x" exists 2 times.
+            raise ValueError(f"`column_names` has repeated entries: {duplicate_columns}")
+        if self.delimiter == "":
+            # str.split("") raises "empty separator", which does not mention delimiter.
+            raise ValueError("`delimiter` must not be empty; pass None to split on any whitespace")
+
 
 class Conll(datasets.ArrowBasedBuilder):
     BUILDER_CONFIG_CLASS = ConllConfig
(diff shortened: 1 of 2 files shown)


src/datasets/packaged_modules/conll/conll.py
```diff
@@ -65,6 +65,18 @@ class ConllConfig(datasets.BuilderConfig):
...
+        super().__post_init__()
+        # Checked here rather than during generation so the error names the argument
+        # instead of arriving wrapped in a DatasetGenerationError.
+        duplicate_columns = sorted({c for c in self.column_names if self.column_names.count(c) > 1})
+        if duplicate_columns:
+            # Otherwise pa.Table.from_arrays fails with KeyError: Field "x" exists 2 times.
+            raise ValueError(f"`column_names` has repeated entries: {duplicate_columns}")
+        if self.delimiter == "":
+            # str.split("") raises "empty separator", which does not mention delimiter.
+            raise ValueError("`delimiter` must not be empty; pass None to split on any whitespace")
```
ebarkhordar, line 78: The same argument applies to the check just below: `column_names=[]` still constructs fine and only fails later at `conll.py:128`, wrapped in the `DatasetGenerationError` this PR exists to avoid.

```
>>> ConllConfig(name="n", column_names=["tokens", "tokens"])
ValueError: `column_names` has repeated entries: ['tokens']
>>> ConllConfig(name="n", column_names=[])
constructed OK, column_names=[]
```

That is your branch at 0a3826d, `pip install -e .` in a clean container.

Moving the non-empty check up beside the other two would cover all three cases, and `_generate_tables` could then drop lines 127 and 128.

PR #8538, opened 2026-08-30
Fix batched JSON record export

Marchematics:
Fix Dataset.to_json(lines=False) when records are written in multiple batches.


Marchematics:
Closing in favor of existing fixes.

Issue #8505, opened 2026-08-23
Request to correct language name for ISO code 'blk' from 'Pa'o Karen' to 'Pa'O'


KhunPhanDuae:
### Describe the bug

Dear Hugging Face Team,

I am writing to request a correction regarding the language display name associated with the ISO 639-3 code `blk` on the platform.

Currently, it is displayed as **"Pa'O Karen"**. However, as a native speaker and local member of the Pa'O community, this terminology is incorrect and does not accurately reflect our self-identification and cultural identity. 

The correct and respectful self-designated name for our language and people is simply **"Pa'O"** (or Pa'O language). The term "Karen" appended to it is an external administrative classification that local native speakers do not accept or use.

Could you please update or provide an option to reflect **"Pa'O"** instead of "Pa'O Karen" for the `blk` language code across the platform? 

Thank you very much for your understanding, respect for indigenous identity, and assistance.

Best regards,
KhunPhanDuae

### Steps to reproduce the bug

<img width="1268" height="640" alt="Image" src="https://github.com/user-attachments/assets/40d8412d-f3ca-4a14-a6e2-475a73a780ee" />

### Expected behavior

Update the display name of the language code blk from "Pa'o Karen" to "Pa'O" (ပအိုဝ်ႏ), as the local native community identifies as Pa'O, not Pa'o Karen.

### Environment info

Not applicable (This is a metadata/terminology correction request, not a software bug).

PR #8540, opened 2026-08-31
Fix language name for ISO code 'blk' (#8505)

raviknce0509:
The `blk` entry listed "Pa'o Karen; Pa'O". Per the request from a native speaker in #8505, "Karen" is an external administrative classification rather than a self-designation, so the entry now uses the self-identified name "Pa'O".

### DIFF
--- a/src/datasets/utils/resources/languages.json
+++ b/src/datasets/utils/resources/languages.json
@@ -802,7 +802,7 @@
     "blh": "Kuwaa",
     "bli": "Bolia",
     "blj": "Bolongan",
-    "blk": "Pa'o Karen; Pa'O",
+    "blk": "Pa'O",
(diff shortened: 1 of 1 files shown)


raviknce0509 (commented):
Fixes #8505

## What
Change the language name for ISO 639-3 code `blk` from "Pa'o Karen; Pa'O" to "Pa'O".

## Why
Per #8505, filed by a native Pa'O speaker: "Karen" is an external administrative
classification, not how the community self-identifies. The entry now carries the
self-designated name.

## How tested
- Data-only change to `src/datasets/utils/resources/languages.json`; no code path altered.
- JSON validated (`json.load`, 8024 entries parse).
- `make quality` passing locally.

PR #8542, opened 2026-09-01
feat(features): add BioSequence and BioStructure decodable features

behroozazarkhalili:
Adds two decodable feature types that store a biological file as bytes and decode it to the biopython object the format describes.

| Type | Module | `decode_example` returns | Serves |
|---|---|---|---|
| `BioSequence` | `features/bio_sequence.py` | `Bio.SeqRecord.SeqRecord` | FASTA #7923, FASTQ #7924, GenBank #7951 |
| `BioStructure` | `features/bio_structure.py` | `Bio.PDB.Structure.Structure` | mmCIF #7925, PDB #7926 |

Both follow the contract `Audio`, `Image`, `Pdf` and `Nifti` already use: storage is `struct<bytes, path>`, `decode=False` returns that struct unchanged, and decoding raises `ImportError` when biopython is absent.

### Why two types

One general type would return a different class depending on which file it was handed, which pushes an isinstance check onto every caller. Five would be one class per file format, but FASTA, FASTQ and GenBank all decode to `SeqRecord` and differ only in the parser, so the format is a field rather than a type. The split follows the boundary biopython itself draws between `Bio.SeqIO` and `Bio.PDB`, which is also where the optional imports divide.

Naming agreed with @lhoestq in https://github.com/huggingface/datasets/pull/7923#issuecomment-5451746802.

### biopython stays optional

Declared as the `bio` extra and probed through `config.BIOPYTHON_AVAILABLE`, which checks the import name `Bio` rather than the distribution name `biopython`. Users who never touch biological data pay nothing: the availability flag is a `find_spec` call, and the lazy isinstance dispatch in `_cast_to_python_objects` is additionally guarded on `"Bio" in sys.modules`, so having the package installed never forces the import.

### Registration

Registered everywhere the other decodable features are, so a dataset carrying one survives a round trip through `dataset_info.json`: the `FeatureType` union, the `_FEATURE_TYPES` registry, `features/__init__.py`, the object dispatch that encodes a live `SeqRecord` or `Structure` passed straight to `Dataset.from_dict`, and the docs under `main_classes.mdx`.

### Verification

- 22 tests in `tests/features/test_bio.py`, covering storage type, encoding from path, bytes and dict, the `decode=False` path, `flatten`, `cast_storage`, and the `Features.to_dict` round trip for both types without biopython present, plus decoding to `SeqRecord` and `Structure` when it is.
- Confirmed the tests fail without the implementation (`ImportError: cannot import name 'BioSequence'`) before making them pass.
- `tests/features/test_bio.py` + `tests/features/test_features.py`: 222 passed, 3 skipped, no regressions.
- `ruff check` and `ruff format` clean.

This is the foundation for the five bio-format loaders; wiring each loader to these types follows in their own PRs.

### DIFF
--- a/docs/source/package_reference/main_classes.mdx
+++ b/docs/source/package_reference/main_classes.mdx
@@ -293,6 +293,14 @@ Dictionary with split names as keys ('train', 'test' for example), and `Iterable
 
 [[autodoc]] datasets.Nifti
 
+### BioSequence
+
+[[autodoc]] datasets.BioSequence
+
+### BioStructure
+
+[[autodoc]] datasets.BioStructure
+
 ## Filesystems
 
 [[autodoc]] datasets.filesystems.is_remote_filesystem
--- a/setup.py
+++ b/setup.py
@@ -219,6 +219,8 @@ DOCS_REQUIRE = [
 
 PDFS_REQUIRE = ["pdfplumber>=0.11.4"]
 
+BIO_REQUIRE = ["biopython>=1.80"]
+
 NIBABEL_REQUIRE = ["nibabel>=5.3.2", "ipyniivue==2.4.2"]
 
 ICEBERG_REQUIRE = ["pyiceberg>=0.7.0"]
@@ -241,6 +243,7 @@ EXTRAS_REQUIRE = {
     "benchmarks": BENCHMARKS_REQUIRE,
     "docs": DOCS_REQUIRE,
     "pdfs": PDFS_REQUIRE,
+    "bio": BIO_REQUIRE,
     "nibabel": NIBABEL_REQUIRE,
     "iceberg": ICEBERG_REQUIRE,
 }
--- a/src/datasets/config.py
+++ b/src/datasets/config.py
@@ -139,6 +139,8 @@ IS_MP3_SUPPORTED = True
 TORCHCODEC_AVAILABLE = importlib.util.find_spec("torchcodec") is not None
 TORCHVISION_AVAILABLE = importlib.util.find_spec("torchvision") is not None
 PDFPLUMBER_AVAILABLE = importlib.util.find_spec("pdfplumber") is not None
+# biopython is imported as ``Bio`` (the PyPI package name is ``biopython``).
+BIOPYTHON_AVAILABLE = importlib.util.find_spec("Bio") is not None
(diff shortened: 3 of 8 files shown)


behroozazarkhalili:
Follow-up after a second review pass. Head is ea2682125.

- `encode_bio_structure` wrote mmCIF for every format string other than exactly `"pdb"`, while `decode_example` accepted only `"pdb"` and `"mmcif"`, so `BioStructure(format="PDB")` stored bytes that could never be read back. Encode and decode now share one `(parser, writer)` table and the format is validated in `__post_init__`.
- `string_to_dict` returns `None` for a non-Hub URL, so the `except ValueError` in both `decode_example` methods never ran and a plain https or s3 path raised `TypeError`. Token resolution is one helper shared by decode and embed, with the `None` check the embed path already had.
- The shared embed helper masked on `bytes_array.is_null()`, nulling a path-only row when embedding was disabled; it masks on the input row now, as `Image` does.

Tests: 30 in `tests/features/test_bio.py`, ruff clean. `pdf.py:158` upstream has the same dead `except ValueError`; happy to open a separate issue for it.

behroozazarkhalili:
CI note: the full upstream workflow (ubuntu and windows, py3.10 to 3.14, minimum and latest deps) passed on this exact head (794afb4cf) on my fork: https://github.com/behroozazarkhalili/datasets/actions/runs/33911204131. Approving the workflow run here should be a formality.

lhoestq (approved):
lgtm !

lhoestq:
Merging this one ! feel free to adapt the five loaders https://github.com/huggingface/datasets/pull/7951, https://github.com/huggingface/datasets/pull/7923, https://github.com/huggingface/datasets/pull/7924, https://github.com/huggingface/datasets/pull/7925 and https://github.com/huggingface/datasets/pull/7926

behroozazarkhalili:
Thanks for the review! The only red check here is `test_delete_from_hub`, a hub-ci read timeout; the full workflow passed on this same commit (794afb4cf) on my fork: https://github.com/behroozazarkhalili/datasets/actions/runs/33911204131. Once this lands I'll rebase the five loader PRs onto it and switch them to `BioSequence` / `BioStructure`.

apcamargo:
A quick question: I understand that `biopython` is optional for installing `datasets`, but is it also optional if I want to parse a FASTA file?

PR #8557, opened 2026-09-02
Support JPEG XL images via pillow-jxl-plugin

### LINKED ISSUE #8537: JPEG XL support
It seems that the dataset don't support JXL natively in the same ways as e.g. PNG and JPEG are supported. A JPEG XL is supported by an international ISO standard, gives great compression ratios, finally supports lossless for JPEG and the browser support is now starting to get wide spread (safari has it for some time -> https://caniuse.com/jpegxl, firefox and chrome have it in test mode and publicly stated recently an intention to ship https://groups.google.com/a/mozilla.org/g/dev-platform/c/3YMV4MS34KA?pli=1 and https://groups.google.com/a/chromium.org/g/blink-dev/c/-gDojQbDPRI)

I could find the old PR #5289 which might be the base. But I don't know whether that's sufficient or some more code changes are required to get e.g. the dataset viewer going.

jayzuccarelli:
Fixes #8537.

## What

- `.jxl` added to the `imagefolder` and `webdataset` image extensions.
- `Image.decode_example` now imports the third-party Pillow plugins that are installed (`pillow_jxl` for now) before calling `PIL.Image.open`, via a small `_import_pil_plugins()` helper and a `config.PILLOW_JXL_AVAILABLE` flag, same pattern as the other optional deps.

## Why the second part

Answering the open question in #8537: it is neither the plugin nor a limitation of Pillow's plugin system, the plugin just never gets imported. Pillow only registers its built-in formats on its own; `pillow_jxl` registers `JXL` (with its file signature, so extensionless files are fine) when the module is imported, and nothing in the `load_dataset` path did that. Checked with a fresh interpreter and a `.jxl` file copied to an extensionless path like the Hub cache blobs:

```
no-import:            UnidentifiedImageError
with-import:          JXL (8, 8)
with-import + .jxl:   JXL
```

(`pillow-jxl-plugin` 1.3.8, Pillow 12.3.0)

So with only the extension change, `load_dataset` on a JXL imagefolder still fails at decode time unless the user remembers to `import pillow_jxl` in their own script. With the import, `pip install pillow-jxl-plugin` is all that is needed. If you'd rather keep this PR to the extension change and document the import instead, I'll drop the helper.

## Tests

- `test_image_decode_example_imports_pil_plugins`: stubs `pillow_jxl` and checks `decode_example` imports it.
- `test_image_decode_example_jxl`: real round trip with `pillow-jxl-plugin` (skipped when it is not installed), on both a `.jxl` path and an extensionless copy.

```
$ pytest tests/features/test_image.py tests/packaged_modules/test_imagefolder.py -q
83 passed, 1 skipped
```
(81 passed, 1 skipped on main before the change.) `ruff check` and `ruff format --check` clean.

### DIFF
--- a/src/datasets/config.py
+++ b/src/datasets/config.py
@@ -134,6 +134,7 @@ SQLALCHEMY_AVAILABLE = importlib.util.find_spec("sqlalchemy") is not None
 
 # Optional tools for file parsing and feature decoding
 PIL_AVAILABLE = importlib.util.find_spec("PIL") is not None
+PILLOW_JXL_AVAILABLE = importlib.util.find_spec("pillow_jxl") is not None
 IS_OPUS_SUPPORTED = True
 IS_MP3_SUPPORTED = True
 TORCHCODEC_AVAILABLE = importlib.util.find_spec("torchcodec") is not None
--- a/src/datasets/features/image.py
+++ b/src/datasets/features/image.py
@@ -43,6 +43,17 @@ _VALID_IMAGE_ARRAY_DTPYES = [
 ]
 
 
+def _import_pil_plugins() -> None:
+    """Import the third-party Pillow plugins that are installed.
+
+    Pillow only registers its built-in formats on its own. Plugins register their format (and its file signature,
+    so that extensionless files are recognized too) when they are imported, so import them here to make
+    `PIL.Image.open` accept these formats.
+    """
+    if config.PILLOW_JXL_AVAILABLE:
+        import pillow_jxl  # noqa: F401  # JPEG XL
+
+
 @dataclass
 class Image:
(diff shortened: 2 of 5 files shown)


jayzuccarelli:
@lhoestq ready for a look when you have a moment. Small change: `.jxl` in the image extension lists plus importing installed Pillow plugins before `Image.open`, which is what #8537 actually needed. CI is green.

jayzuccarelli:
@lhoestq when you have a moment: this one (JPEG XL support) and #8493 (don't mask the original error when finalizing the writer fails) are both small. On #8493 the open question is whether you want the fsync-before-close I suggested in #8491. Happy to rebase either.

PR #8558, opened 2026-09-02
Close the Arrow writer when finalize() fails

### LINKED ISSUE #6917: WinError 32 The process cannot access the file during load_dataset
### Describe the bug

When I try to load the opus_book from hugging face (following the [guide on the website](https://huggingface.co/docs/transformers/main/en/tasks/translation))

```python
from datasets import load_dataset, Dataset

dataset = load_dataset("Helsinki-NLP/opus_books", "en-fr", features=["id", "translation"])
```

I get an error:
`PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:/Users/Me/.cache/huggingface/datasets/Helsinki-NLP___parquet/ca-de-a39f1ef185b9b73b/0.0.0/2a3b91fbd88a2c90d1dbbb32b460cf621d31bd5b05b934492fdef7d8d6f236ec.incomplete\\parquet-train-00000-00000-of-NNNNN.arrow'
`

<details><summary>Full stacktrace</summary>
<p>

```python

File c:\Users\Me\.conda\envs\ia\lib\site-packages\datasets\builder.py:1858, in ArrowBasedBuilder._prepare_split_single(self, gen_kwargs, fpath, file_format, max_shard_size, job_id)
   [1857](file:///C:/Users/Me/.conda/envs/ia/lib/site-packages/datasets/builder.py:1857) _time = time.time()
-> [1858](file:///C:/Users/Me/.conda/envs/ia/lib/site-packages/datasets/builder.py:1858) for _, table in generator:
   [1859](file:///C:/Users/Me/.conda/envs/ia/lib/site-packages/datasets/builder.py:1859)     if max_shard_size is not None and writer._num_bytes > max_shard_size:

File c:\Users\Me\.conda\envs\ia\lib\site-packages\datasets\packaged_modules\parquet\parquet.py:59, in Parquet._generate_tables(self, files)
     [58](file:///C:/Users/Me/.conda/envs/ia/lib/site-packages/datasets/packaged_modules/parquet/parquet.py:58) def _generate_tables(self, files):
---> [59](file:///C:/Users/Me/.conda/envs/ia/lib/site-packages/datasets/packaged_modules/parquet/parquet.py:59)     schema = self.config.features.arrow_schema if self.config.features is not None else None
     [60](file:///C:/Users/Me/.conda/envs/ia/lib/site-packages/datasets/packaged_modules/parquet/parquet.py:60)     if self.config.features is not None and self.config.columns is not None:

AttributeError: 'list' object has no attribute 'arrow_schema'

During handling of the above exception, another exception occurred:

File c:\Users\Me\.conda\envs\ia\lib\site-packages\datasets\builder.py:1882, in ArrowBasedBuilder._prepare_split_single(self, gen_kwargs, fpath, file_format, max_shard_size, job_id)
   [1881](file:///C:/Users/Me/.conda/envs/ia/lib/site-packages/datasets/builder.py:1881) num_shards = shard_id + 1
-> [1882](file:///C:/Users/Me/.conda/envs/ia/lib/site-packages/datasets/builder.py:1882) num_examples, num_bytes = writer.finalize()
   [1883](file:///C:/Users/Me/.conda/envs/ia/lib/site-packages/datasets/builder.py:1883) writer.close()

File c:\Users\Me\.conda\envs\ia\lib\site-packages\datasets\arrow_writer.py:584, in ArrowWriter.finalize(self, close_stream)
    [583](file:///C:/Users/Me/.conda/envs/ia/lib/site-packages/datasets/arrow_writer.py:583) # If schema is known, infer features even if no examples were written
--> [584](file:///C:/Users/Me/.conda/envs/ia/lib/site-packages/datasets/arrow_writer.py:584) if self.pa_writer is None and self.schema:
...
--> [627](file:///C:/Users/Me/.conda/envs/ia/lib/shutil.py:627)         os.unlink(fullname)
    [628](file:///C:/Users/Me/.conda/envs/ia/lib/shutil.py:628)     except OSError:
    [629](file:///C:/Users/Me/.conda/envs/ia/lib/shutil.py:629)         onerror(os.unlink, fullname, sys.exc_info())

PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:/Users/Me/.cache/huggingface/datasets/Helsinki-NLP___parquet/ca-de-a39f1ef185b9b73b/0.0.0/2a3b91fbd88a2c90d1dbbb32b460cf621d31bd5b05b934492fdef7d8d6f236ec.incomplete\\parquet-train-00000-00000-of-NNNNN.arrow'
``` 

</p>
</details> 

Steps to reproduce:

Just execute these lines

```python
from datasets import load_dataset, Dataset

dataset = load_dataset("Helsinki-NLP/opus_books", "en-fr", features=["id", "translation"])
```

### Expected behavior

I expect the dataset to be loaded without any errors.

### Environment info

| Package| Version|
|--------|--------|
| transformers| 4.37.2|
| python| 3.9.19|
| pytorch| 2.3.0|
| datasets|2.12.0 | 
| arrow  |  1.2.3| 
                 
I am using Conda on Windows 11.

antnewman:
Fixes #6917.

## Root cause

`_prepare_split_single` finalises the writer inside a `finally` block, in both builders (`builder.py:1558-1563` for `GeneratorBasedBuilder`, `builder.py:1864-1869` for `ArrowBasedBuilder`):

```python
finally:
    yield job_id, False, num_examples_progress_update
    num_shards = shard_id + 1
    num_original_shards = original_shard_id + 1
    writer.close()
```

`ArrowWriter.finalize()` can raise before it reaches its own `stream.close()`: `write_rows_on_file()` (`arrow_writer.py:781`), `write_examples_on_file()` (:783), `_build_writer()` (:786) and `pa_writer.close()` (:788) are all upstream of it. When one of those raises, `writer.close()` on the next line never runs and the Arrow file stays open. (The `SchemaInferenceError` branch at :794-795 is fine, it closes the stream first.)

`incomplete_dir` then cleans up (`builder.py:874-876`):

```python
finally:
    if os.path.exists(tmp_dir):
        shutil.rmtree(tmp_dir)
```

On Linux, unlinking a file that is still open succeeds and nobody notices. On Windows it raises `PermissionError`, **from inside a `finally`**, which replaces the exception that was propagating. That is the two-stage traceback in the issue: a real error, discarded, and a file-lock error in its place.

## Reproduction

Windows 10, Python 3.13.6, `main` at 5.0.2.dev0. Plain data, no mocking:

```python
import tempfile
import pyarrow as pa
from datasets.builder import ArrowBasedBuilder
from datasets.features import Features, Value
from datasets.info import DatasetInfo
from datasets.splits import Split, SplitGenerator

class DummyBuilder(ArrowBasedBuilder):
    def _info(self):
        return DatasetInfo(features=Features({"text": Value("string")}))

        return [SplitGenerator(name=Split.TRAIN)]

    def _generate_tables(self):
        # a single-row table is buffered by write_row() and only cast when finalize() flushes it
        yield 0, pa.table({"text": ["foo"], "extra": ["bar"]})

with tempfile.TemporaryDirectory() as tmp:
    DummyBuilder(cache_dir=tmp).download_and_prepare()
```

Before: `PermissionError: [WinError 32] ... 0.0.0.incomplete\...-00000-of-NNNNN.arrow`.
After: `DatasetGenerationError` wrapping `CastError: Couldn't cast ... because column names don't match`, which is the error the user needed to see.

## The change

Run `writer.close()` in a `finally` at both sites, and do not let a failure to close hide the original error:

```python
try:
finally:
    try:
        writer.close()
    except Exception as e:  # don't let a failure to close hide the original error
        logger.warning(f"Failed to close the writer while handling an error: {e!r}")
```

The inner guard matters. `ArrowWriter.close()` wraps `pa_writer.close()` in `try/except Exception: pass` (`arrow_writer.py:565-568`) but calls `self.stream.close()` unguarded (:569-570), and on a remote fsspec filesystem that close commits the upload and can fail. Without the guard, an `ENOSPC` or a failed commit would demote the user's real error to `__cause__.__context__`, which is the same masking this PR is fixing. This mirrors the shape of #8493 for `Dataset.map`.

The success path is unchanged: `finalize()` sets `pa_writer = None` and closes the stream, so the subsequent `close()` short-circuits on both guards and is a no-op. `ParquetWriter` overrides neither `finalize` nor `close`, so it inherits the same behaviour.

## Scope, and what this does not fix

- **Shard rotation is deliberately untouched, though not for the reason this note first gave.** It originally called wrapping those sites "strictly worse", on the grounds that the outer `finally` then re-finalises a spent writer and a `ValueError` replaces the real `OSError`. @ebarkhordar showed that comparison was against the wrong baseline: the masking is already there without the wrap, so wrapping is neutral rather than a regression. I withdraw "strictly worse". The masking itself is real and pre-existing — `finalize()` clears `pa_writer` at `arrow_writer.py:811` before closing the stream at `:813`, so a second `finalize()` rebuilds against a closed stream and raises `ValueError: I/O operation on closed file` (not "write to closed file", as this note also had it). Confirmed on Windows as well as Linux. Separate defect, not addressed here; see the discussion below.
- **`num_proc > 1` was never affected.** `Pool.__exit__` terminates workers, releasing their handles, before the parent's `rmtree`. I measured both paths; only the single-process path ever showed the bug.
- **Abandoning the generator still leaks.** `_prepare_split_single` is a generator whose outer `finally` begins with `yield`, so `GeneratorExit` (a `KeyboardInterrupt` landing in the consumer loop) skips the block entirely and neither `finalize()` nor `close()` runs. That is a pre-existing generator-lifecycle issue, unchanged by this PR, and fixing it means restructuring writer ownership.
- Also noticed and left alone: the same `finalize(); close()` shape in `packaged_modules/spark/spark.py:239` and `arrow_dataset.py:1955`, and a double-`finalize()` on one writer at `arrow_dataset.py:4014` / `:4019`.

## Tests

`test_builder_download_and_prepare_when_finalize_fails` is parametrised over both builders and fails on `main` on every platform, for different reasons: on Windows with `PermissionError [WinError 32]`, and on Linux on the leaked-handle assertion, since there the original error propagates correctly and only the file handle is left open. So it needs no platform guard. I verified the Linux behaviour by making `rmtree` tolerant to mimic POSIX unlink semantics.

## CI

On this machine, 15 tests in `tests/test_arrow_dataset.py` (`from_parquet`, `to_iterable_dataset`) fail identically with and without this change; I checked them against unmodified `main`. `tests/test_builder.py` and `tests/test_arrow_writer.py` are green (142 passed, 3 skipped).


ebarkhordar (commented):
I ran both arms in a clean container (`python:3.12-slim`, `pip install -e .`) at `4e79da3805cca32c2656138359e2d3920326c066` and at the merge base `bff13713a1e63a57a4793273df7665ce5f17a413`.

The scope note says the rotation sites are fine because "the outer `finally` finalises and closes the same object and the handle is released anyway". That holds for the handle. The masking you measured when you tried wrapping them is already there without the wrap.

`builder.py:1535` calls `finalize()`, which sets `pa_writer = None` at `arrow_writer.py:789` and only then calls `self.stream.close()` at :791. If that close raises, `writer` is not rebound, so the outer `finally` re-finalises the same object, `pa_writer` is now `None`, and `finalize()` goes on into `_build_writer(self.schema)` against a stream that is gone.

I injected the failure at the first shard's `stream`: `close()` marks the stream closed and then raises `OSError(28)`, which is the order fsspec's `AbstractBufferedFile.close` uses, since `self.closed = True` sits in its `finally`. Six 4KB rows, `_writer_batch_size = 1`, `max_shard_size = 1024`, so rotation fires on a real generator.

```
[0] DatasetGenerationError: An error occurred while generating the dataset
[1] ValueError: I/O operation on closed file
[2] OSError: [Errno 28] No space left on device
```

Identical on both arms, so this is pre-existing and your diff neither causes it nor makes it worse. `raise DatasetGenerationError(...) from e` picks [1], so the disk-full error the user needs is the one they do not see. Same demotion this PR removes at the other site, from the other direction.

That puts the "strictly worse" comparison against a baseline that already had it, which may change the answer: is the real problem at those two sites the missing `close()`, or the outer `finally` re-finalising a writer that is already spent?

I did not test on Windows and I did not test against a real remote filesystem.

antnewman:
Thank you — that is a careful piece of work, and you are right.

I have rebased onto main (the branch was 22 commits behind, so my earlier line numbers pointed at other real lines) and corrected the scope note. Line numbers below are current main.

## What I am withdrawing

The note made two claims and you refuted the second. **"Strictly worse" was wrong**, and it was wrong for the reason you give: I compared a wrapped rotation site against a baseline I assumed was clean, when the `ValueError` was already there unwrapped. Wrapping is neutral, not a regression. I have struck that from the description.

The handle half you granted, and it does hold — but it never licensed the conclusion I drew from it.

## Confirming it, including the platform you flagged

You said you had not tested Windows. I have, and I get your trace exactly — Windows 10, Python 3.13.6, pyarrow 25.0.1, six 4KB rows, `_writer_batch_size = 1`, `max_shard_size = 1024`, failing the first shard's `stream.close()`:

```
[0] DatasetGenerationError: An error occurred while generating the dataset
[1] ValueError: I/O operation on closed file
[2] OSError: [Errno 28] No space left on device
```

Isolated from the builder, the mechanism is just the two calls:

```python
writer = ArrowWriter(stream=FailingCloseStream(), schema=schema)
writer.write({"x": 1})
try:
    writer.finalize()          # OSError: [Errno 28]
except OSError:
    writer.finalize()          # ValueError: I/O operation on closed file
```

Worth adding *why* your `[2]` survives at all: that second `finalize()` runs inside the outer `finally` while the `OSError` is still propagating, so Python sets `__context__` implicitly. Called at the top level instead, the same `ValueError` arrives with `__context__` of `None`. So the disk-full error is preserved by accident of position rather than by design — which is thinner than it looks if anyone ever reorders that block.

## Your question: missing `close()`, or the re-finalise?

The re-finalise, and underneath it one flag carrying two meanings.

`pa_writer is None` means "not built yet" at every read that decides whether to build — `arrow_writer.py:740`, `:773`, `:789`, and the one that bites, `:807`. But `finalize()` sets it to `None` at `:811` to mean "finished", and does that *before* `self.stream.close()` at `:813`. A close failure therefore leaves the object looking, by that flag, like one that never had a writer, and the next `finalize()` takes `:807` into `_build_writer(self.schema)` against a stream that is gone.

That makes the missing `close()` at the rotation sites the smaller half: adding one would not stop the second `finalize()` rebuilding, so the masking would survive it. I checked, and in this particular failure mode the added `close()` is inert anyway, because the stream is already closed by the time it runs.

Two shapes would fix it properly:

- **Narrow, builder-local:** stop the outer `finally` re-finalising a writer that was already finalised at the rotation site. Confined to `_prepare_split_single`.
- **General:** make `finalize()` record that it finished, so a second call returns the recorded `(num_examples, num_bytes)` rather than rebuilding. Not a bare no-op — `builder.py` unpacks that return, so an early `return None` swaps your `ValueError` for a `TypeError` and masks the `OSError` just the same. The flag would also have to be set only on the success branch, or an empty schema-less writer would return `(0, 0)` on the second call instead of re-raising `SchemaInferenceError`.

I prefer the general one: it is about six lines in one method, and it fixes the other double-`finalize()` caller too — `arrow_dataset.py:4014` and `:4019` finalise the same writer in the `try` and in the `except (Exception, KeyboardInterrupt)`.

## What I would like to do

Keep this PR to the two sites in `_prepare_split_single` and send the `finalize()` change separately, with a test built from your scenario — with your name on it, if you are happy for me to use it.

@lhoestq that is a preference, not a position: if you would rather have one change than two, say so and I will fold it in here instead.

ebarkhordar:
Yes, please use it, and thanks for running the Windows arm.

One thing on the general fix: "set only on the success branch" has two readings and only one of them fixes this trace. I ran both at `1fed7987` in a clean container (`python:3.12-slim`, `pip install -e .`, pyarrow 25.0.1) with your isolated two-call shape, second `finalize()` in a `finally`:

```
base, and flag set just before the return at the end of finalize():
  [0] ValueError: I/O operation on closed file
  [1] OSError: [Errno 28] No space left on device

flag set right after self.pa_writer = None (:811), above self.stream.close() (:813):
  [0] OSError: [Errno 28] No space left on device
```

Setting it at the end changes nothing, because the run that needs the flag never gets there. `stream.close()` raises at :813 and the assignment sits below it. It has to go inside the `if self.pa_writer is not None:` branch, above the close.

That placement gives you the schema-less constraint for free, since `SchemaInferenceError` comes from the `else` branch and never sets the flag. I checked an empty writer both ways and the first and second calls both re-raise it.

Your other-caller list matches what I get from parsing the package rather than grepping it. `_map_single` at `arrow_dataset.py:4014` and `:4019` is the only other pair on one writer. `spark.py:239` and `:257` look like a third but finalise two different writers either side of a shard rotation.

Not checked: Windows, and I did not run the wider suite against a patched `finalize()`.

antnewman:
You are right on the placement. I had written "set only on the success branch" loosely, and it does have two readings, only one of which survives a close that raises, because the `return` sits below `self.stream.close()`.

Confirmed on Windows, both readings:

| flag recorded | outcome |
|---|---|
| before the `return` at the end | `[0] ValueError: I/O operation on closed file`, `[1] OSError: [Errno 28]` |
| inside the success branch, above the close | `[0] OSError: [Errno 28]`, nothing masked |

And you are right that the correct placement gives the schema-less constraint for free rather than needing its own guard, since `SchemaInferenceError` comes from the `else` branch. I checked an empty writer both ways rather than reasoning about it, and it raises on both calls.

Thanks also for ruling `spark.py` out explicitly. I had listed `:239` in the description as the same `finalize(); close()` shape, but you are right that `:239` and `:257` are two different writers either side of a rotation, so `_map_single` is the only other pair on one object.

Opened as #8642, with your reproduction as the test and credited to you.

One thing changed on the way there, from a review pass over my own branch. The early return as I first wrote it would silently drop rows staged after a `finalize()`: `write; finalize(); write; finalize()` returned a count that omitted the second row, where `main` at least raised. Silently losing a row is worse than a confusing error, so the early return is now guarded on nothing being staged, and writing to a finished writer keeps failing as it does today. There is a test for it, though it passes on `main` too, so it is an invariance guard rather than a regression test.

I have also added `ArrowBasedBuilder._prepare_split_single` to the description over there. It has the same rotation shape as the generator one, and I had only named the latter.

Separate PR rather than folded in here, because this one fixes the call sites and that one fixes the writer. @lhoestq if you would rather have a single change, say so and I will fold it back and close the other.

PR #8562, opened 2026-09-04
Keep the declared dtype for an empty column under numpy format

### LINKED ISSUE #8469: Empty column loses its dtype under numpy format (always float32)
### Describe the bug

Under `numpy` format, a column that holds zero rows comes back as `float32` no matter what dtype it was declared with. The same column keeps its dtype as soon as it holds a single row, and the `pandas` and `arrow` formatters keep it in both cases.

This is easy to hit in normal code: any `filter` that matches nothing, or `select([])`.

### Steps or code to reproduce the bug

```python
import numpy as np

feats = Features({"a": Value("int64")})
ds = Dataset.from_dict({"a": [1, 2, 3]}, features=feats).with_format("numpy")

matched   = ds.filter(lambda x: x["a"] > 1)[:]["a"]
unmatched = ds.filter(lambda x: x["a"] > 99)[:]["a"]

print(matched.dtype)    # int64
print(unmatched.dtype)  # float32
```

Across dtypes:

| declared | non-empty | empty |
|---|---|---|
| `int64` | `int64` | **`float32`** |
| `int32` | `int64` | **`float32`** |
| `float64` | `float32` | `float32` |
| `bool` | `bool` | **`float32`** |
| `string` | `<U1` | **`float32`** |

Only the numpy formatter is affected:

```
   pandas : int64
   arrow  : int64
   numpy  : float32
```

### Why it matters

The dtype silently changes based on how many rows survived a filter, so a numeric pipeline that concatenates per-batch results gets its integers promoted the moment one batch is empty:

```python
np.concatenate([unmatched, matched]).dtype   # float64, from int64 data
```

A `string` column coming back as `float32` is the same bug wearing a more obvious hat.

### Root cause

`NumpyArrowExtractor._arrow_array_to_numpy` goes through a Python list:

https://github.com/huggingface/datasets/blob/main/src/datasets/formatting/formatting.py#L188

```python
array: list = pa_array.to_numpy(zero_copy_only=zero_copy_only).tolist()
```

`.tolist()` drops the arrow type. For a non-empty list numpy re-infers it correctly; for `[]` it falls back to `float64`. `NumpyFormatter._tensorize` then applies its float default on top:

```python
elif isinstance(value, np.ndarray) and np.issubdtype(value.dtype, np.floating):
    default_dtype = {"dtype": np.float32}
```

which turns the accidental `float64` into `float32`. Confirmed by tracing an empty `int64` column:

```
arrow schema      : a: int64
arrow col to_numpy: int64      <- still correct here
extract_column    : float64    <- lost by .tolist()
formatted         : float32
```

To be clear, `int32 -> int64` and `float64 -> float32` on the non-empty rows are the formatter's documented `default_dtype` behaviour and not part of this report.

### Expected behavior

An empty column should keep the dtype the same column has when non-empty, as it already does under `pandas` and `arrow`.

PR: #8470.

### Environment info

- `datasets` version: 5.0.2.dev0 (`main` @ 48b7ee7)
- Python version: 3.11.9
- Platform: Windows 11
- PyArrow version: 25.0.1
- Pandas version: 3.0.5
- NumPy version: 2.4.6

akashjainn:
Fixes #8469.

Under numpy format a column holding zero rows came back as float32 no matter what dtype it was declared with, while the same column kept its dtype as soon as it held one row. pandas and arrow were unaffected.

```python
feats = Features({"a": Value("int64")})
ds = Dataset.from_dict({"a": [1, 2, 3]}, features=feats).with_format("numpy")

ds.filter(lambda x: x["a"] > 1)[:]["a"].dtype    # int64
ds.filter(lambda x: x["a"] > 99)[:]["a"].dtype   # float32
```

I reproduced the whole table from the issue - int64, int32, bool and string all came back float32 when empty. float64 only looked right because it was already floating.

Two steps produce it. _arrow_array_to_numpy builds a python list and calls np.asarray on it, and a zero-length list carries no type, so you get float64 regardless of the column. NumpyFormatter._tensorize then sees a floating dtype and applies its np.float32 default.

The fix takes the dtype from the arrow type when the column is empty, which is effectively what the pandas and arrow extractors already do. Types with no single numpy equivalent - nested and extension types - return None from the helper and fall through to the existing path rather than getting a guessed dtype, so ArrayXD columns are untouched.

What this fixes in practice, since the dtype changed with how many rows survived a filter:

```python
np.concatenate([unmatched, matched]).dtype   # was float64 from int64 data, now int64
```

One behaviour change worth calling out: an empty string column now comes back as object rather than float32. That's what arrow and pandas give and it's well defined for a zero-length array, whereas the non-empty <U1 isn't - there's no natural width for an empty one. Happy to do something different there if you'd rather.

Tests: three new ones in tests/test_formatting.py covering the dtype round-trip across int64/int32/float64/bool, the string case specifically, and the concatenate promotion. All five assertions fail on main and pass with the change.

tests/test_formatting.py: 27 passed, 30 skipped. ruff check and ruff format clean.

One pre-existing failure I ran into while checking wider: tests/test_arrow_dataset.py::test_dataset_to_iterable_dataset fails identically on unmodified main in my environment, and a lot of tests/features/ errors out on missing optional video and audio dependencies.

### DIFF
--- a/src/datasets/formatting/formatting.py
+++ b/src/datasets/formatting/formatting.py
@@ -151,19 +151,31 @@ class PythonArrowExtractor(BaseArrowExtractor[dict, list, dict]):
 
 
 def _numpy_dtype_from_arrow_type(pa_type: pa.DataType) -> Optional[np.dtype]:
-    """The numpy dtype an arrow type maps to, or None if it has no direct equivalent.
+    """The numpy dtype an arrow type maps to, or None if arrow declines to convert it.
 
     Used only for empty columns. A zero-length Python list carries no type, so
     ``np.asarray([])`` yields float64 regardless of what the column was declared
     as, and the numpy formatter's float default then narrows that to float32.
-    Taking the dtype from the arrow type keeps an empty column matching a
-    non-empty one, which is what the pandas and arrow extractors already do.
+    Taking the dtype from a zero-length arrow array of the column's own type
+    keeps an empty column matching a non-empty one, which is what the pandas and
+    arrow extractors already do.
+
+    Arrow has to be the one asked, rather than ``pa_type.to_pandas_dtype()``:
+    pandas has no day-resolution datetime64, so it reports date32 as
+    ``datetime64[ms]`` while every other path here renders it as
+    ``datetime64[D]``.
+
+    Nested types are not excluded from this path: arrow converts list, map and
+    struct columns to object, so they land here rather than falling through. It
+    does not close the gap for them - an empty ``list<int64>`` column is object
+    where a filled one is int64, because the filled side takes its dtype from the
+    stacking path above rather than from here.
     """
     try:
-        return np.dtype(pa_type.to_pandas_dtype())
+        return pa.array([], type=pa_type).to_numpy(zero_copy_only=False).dtype
     except (NotImplementedError, TypeError, ValueError):
-        # Nested and extension types have no single numpy dtype; leave those to
-        # the existing path rather than guessing at one.
+        # ArrowNotImplementedError subclasses NotImplementedError and ArrowInvalid
+        # subclasses ValueError, so arrow's own errors land here.
         return None
 
 
--- a/src/datasets/formatting/formatting.py
+++ b/src/datasets/formatting/formatting.py
@@ -150,6 +150,23 @@ class PythonArrowExtractor(BaseArrowExtractor[dict, list, dict]):
         return pa_table.to_pydict()
 
 
+def _numpy_dtype_from_arrow_type(pa_type: pa.DataType) -> Optional[np.dtype]:
+    """The numpy dtype an arrow type maps to, or None if it has no direct equivalent.
+
+    Used only for empty columns. A zero-length Python list carries no type, so
+    ``np.asarray([])`` yields float64 regardless of what the column was declared
+    as, and the numpy formatter's float default then narrows that to float32.
+    Taking the dtype from the arrow type keeps an empty column matching a
+    non-empty one, which is what the pandas and arrow extractors already do.
+    """
+    try:
+        return np.dtype(pa_type.to_pandas_dtype())
+    except (NotImplementedError, TypeError, ValueError):
+        # Nested and extension types have no single numpy dtype; leave those to
+        # the existing path rather than guessing at one.
+        return None
+
+
 class NumpyArrowExtractor(BaseArrowExtractor[dict, np.ndarray, dict]):
     def __init__(self, **np_array_kwargs):
         self.np_array_kwargs = np_array_kwargs
@@ -212,6 +229,10 @@ class NumpyArrowExtractor(BaseArrowExtractor[dict, np.ndarray, dict]):
                             return np.asarray(array, dtype=object)
                         return np.array(array, copy=False, dtype=object)
                     break
+        if len(array) == 0:
+            dtype = _numpy_dtype_from_arrow_type(pa_array.type)
+            if dtype is not None:
+                return np.empty(0, dtype=dtype)
         if np.lib.NumpyVersion(np.__version__) >= "2.0.0b1":
             return np.asarray(array)
         else:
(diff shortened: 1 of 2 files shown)


src/datasets/formatting/formatting.py
```diff
@@ -150,6 +150,23 @@ def extract_batch(self, pa_table: pa.Table) -> dict:
...
+    """The numpy dtype an arrow type maps to, or None if it has no direct equivalent.
+
+    Used only for empty columns. A zero-length Python list carries no type, so
+    ``np.asarray([])`` yields float64 regardless of what the column was declared
+    as, and the numpy formatter's float default then narrows that to float32.
+    Taking the dtype from the arrow type keeps an empty column matching a
+    non-empty one, which is what the pandas and arrow extractors already do.
+    """
+    try:
+        return np.dtype(pa_type.to_pandas_dtype())
```
ebarkhordar: I ran this against `b76c8b14` in a clean `python:3.12-slim` container, pyarrow 25.0.1 and numpy 2.5.2. The dtype this picks is exact for the fixed-width numeric and boolean types your tests cover. `date32` is the one where it disagrees with the array's own dtype:

```
type      empty.dtype      nonempty.dtype
date32    datetime64[ms]   datetime64[D]
```

`pa.date32().to_pandas_dtype()` is `datetime64[ms]`, but the non-empty path reaches numpy through `chunk.to_numpy()`, which gives `datetime64[D]`. So an empty and a filled `date32` column still disagree, which is the thing the docstring says this keeps.

Taking the dtype from a zero-length array of the same type agrees by construction:

```python
return pa.array([], type=pa_type).to_numpy(zero_copy_only=False).dtype
```

The `except` clause needs no change, since `pa.lib.ArrowNotImplementedError` subclasses `NotImplementedError`.

Patched in locally, `date32` empty and filled are both `datetime64[D]`, and two columns that currently fall through to the float default pick up a real dtype: an empty struct column becomes object, matching its filled form, and an empty `Array2D(shape=(2,2), dtype="float32")` becomes float32 instead of float64. Your three new tests and the rest of `tests/test_formatting.py` pass with it, 27 passed and 30 skipped.

I only exercised the numpy extractor, not the pandas or arrow ones.

src/datasets/formatting/formatting.py
```diff
@@ -150,6 +150,23 @@ def extract_batch(self, pa_table: pa.Table) -> dict:
...
+    except (NotImplementedError, TypeError, ValueError):
+        # Nested and extension types have no single numpy dtype; leave those to
```
ebarkhordar: Separate from the dtype point above: this comment does not describe what pyarrow does, so the fall-through is narrower than it looks.

`pa.list_(pa.int64()).to_pandas_dtype()` returns `object` rather than raising, and so does `pa.map_(pa.string(), pa.int64())`. Only `pa.struct(...)` raises. Measured on pyarrow 25.0.1 with pandas 3.0.5.

So list and map columns do take this branch, and an empty `list<int64>` column comes back `object` where a filled one comes back `int64`. That gap is not closed by changing how the dtype is looked up, since the filled side gets its dtype from the stacking path rather than from here. Worth saying in the comment either way, so the next reader does not assume nested types are excluded.

src/datasets/formatting/formatting.py
akashjainn: Thank you for flagging that! Fixed here: 500db47d26a322a591deaf35b2e65ad3ace8883b. On your caveat about only exercising the numpy extractor: the helper has exactly one call site, inside `NumpyArrowExtractor._arrow_array_to_numpy`, so `PandasArrowExtractor` and `ArrowExtractor` never reach it.

src/datasets/formatting/formatting.py
akashjainn: Updated docstrings to clearly state nested types are not excluded here: 500db47d26a322a591deaf35b2e65ad3ace8883b. Thanks!

src/datasets/formatting/formatting.py
ebarkhordar: Read 500db47d. The single call site does close my caveat, thanks for checking that rather than taking my word for it. date32 and time32 are the two I would have picked to pin.

PR #8563, opened 2026-09-04
Raise clear errors for missing folder metadata files

### LINKED ISSUE #7369: Importing dataset gives bad error message when filename's in metadata.csv are not found in the directory
### Describe the bug

While importing an audiofolder dataset, 

Assume an audiofolder with audiofiles, filename1.mp3, filename2.mp3 etc and a file metadata.csv which contains the columns file_name and sentence. The file_names are formatted like filename1.mp3, filename2.mp3 etc. 

Load the audio
```
load_dataset("audiofolder", data_dir='/path/to/audiofolder')
```
When the file_names in the csv are not in sync with the filenames in the audiofolder, then we get an Error message:
```
File /opt/conda/lib/python3.12/site-packages/datasets/arrow_reader.py:251, in BaseReader.read(self, name, instructions, split_infos, in_memory)
    249 if not files:
    250     msg = f'Instruction "{instructions}" corresponds to no data!'
--> 251     raise ValueError(msg)
    252 return self.read_files(files=files, original_instructions=instructions, in_memory=in_memory)

ValueError: Instruction "train" corresponds to no data!
```
load_dataset has a default value for the argument split = 'train'. 

### Expected behavior

It would be better to get an error report something like:
```
The metadata.csv file has different filenames than the files in the datadirectory. 
```
It would have saved me 4 hours of debugging. 

### Environment info

- `datasets` version: 3.2.0
- Platform: Linux-5.14.0-427.40.1.el9_4.x86_64-x86_64-with-glibc2.39
- Python version: 3.12.8
- `huggingface_hub` version: 0.27.0
- PyArrow version: 18.1.0
- Pandas version: 2.2.3
- `fsspec` version: 2024.9.0

tadad:
## Summary

- validate local files referenced by folder metadata during example generation
- report both the missing file and the metadata file that references it
- preserve lazy access for remote files

## Testing

- `pytest -q tests/packaged_modules/test_folder_based_builder.py tests/packaged_modules/test_imagefolder.py tests/packaged_modules/test_audiofolder.py`
- `ruff check src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py tests/packaged_modules/test_folder_based_builder.py`
- `ruff format --check src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py tests/packaged_modules/test_folder_based_builder.py`

Fixes #7369

### DIFF
--- a/src/datasets/builder.py
+++ b/src/datasets/builder.py
@@ -1568,7 +1568,10 @@ class GeneratorBasedBuilder(DatasetBuilder):
             # Ignore the writer's error for no examples written to the file if this error was caused by the error in _generate_examples before the first example was yielded
             if isinstance(e, SchemaInferenceError) and e.__context__ is not None:
                 e = e.__context__
-            raise DatasetGenerationError("An error occurred while generating the dataset") from e
+            message = "An error occurred while generating the dataset"
+            if isinstance(e, FileNotFoundError):
+                message += f": {e}"
+            raise DatasetGenerationError(message) from e
 
         yield (
             job_id,
--- a/src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
+++ b/src/datasets/packaged_modules/folder_based_builder/folder_based_builder.py
@@ -15,7 +15,7 @@ import datasets
 from datasets import config
 from datasets.builder import Key
 from datasets.features.features import FeatureType, _visit, _visit_with_path, _VisitPath, require_storage_cast
-from datasets.utils.file_utils import readline
+from datasets.utils.file_utils import is_local_path, readline
 
 
 logger = datasets.utils.logging.get_logger(__name__)
@@ -401,6 +401,7 @@ class FolderBasedBuilder(datasets.GeneratorBasedBuilder):
                         )
                     elif len(feature_path) == 0:
                         if item is not None:
+                            original_item = item
                             # Guard against path traversal (CWE-22): a crafted `file_name` such as
                             # "../../etc/passwd" or an absolute path must not be able to escape the
                             # metadata file's directory and read arbitrary files on the host.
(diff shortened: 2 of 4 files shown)


ebarkhordar (commented):
I ran this end to end against `bb664b20` in a clean `python:3.12-slim` container, on an imagefolder whose `metadata.csv` names a file that is not on disk. The check fires, but `load_dataset` wraps it, so the last line the user sees is still the generic one:

```
TOP:   DatasetGenerationError: An error occurred while generating the dataset
CAUSE: FileNotFoundError: File 'missing.png' referenced in metadata file '/tmp/rep/af/metadata.csv' does not exist
```

`builder.py:1571` does `raise DatasetGenerationError(...) from e`, so the message you wrote is the chained cause. It is in the traceback, but the final line is the part people copy into an issue, and #7369 is a report about exactly that line.

The test added here calls `_generate_examples` directly, so it passes without going through that path. A second test at the `load_dataset` level asserting `excinfo.value.__cause__` would pin the behaviour the issue asks for.

One thing I could not reproduce, so the issue link may be worth a second look: on the merge base `bff13713` the same directory loaded with no error at all, one row, `path='/tmp/rep/af/missing.png'`. I never saw the `Instruction "train" corresponds to no data!` that #7369 reports, so that symptom may come from a different layout than a plain missing reference.

Test suites are unaffected: `tests/packaged_modules/test_folder_based_builder.py test_imagefolder.py test_audiofolder.py` give 119 passed at base and 120 at head, same 9 audiofolder failures on both, which are a torchcodec library load in my container and nothing to do with this change.

tadad:
Thanks—addressed both main points in d46f5a899: `load_dataset()` now surfaces the `FileNotFoundError` detail, and the regression test now exercises `load_dataset()` and asserts the cause chain.

ebarkhordar:
Read the new commit, and that covers both points: the top-level message now carries the filename, and the test goes through `load_dataset` and asserts the cause chain. I have not re-run it against my container repro.

PR #8564, opened 2026-09-04
Print the repr these feature examples actually produce

vineethsaivs:
### What breaks

Four feature classes show a repr in their `Examples` block that they cannot print.

```
>>> ds.features["image"]
Image(decode=True, id=None)
```

`Image`, `Pdf`, `Nifti` and `Video` all declare `id: Optional[str] = field(default=None, repr=False)`, so `id=None` never appears. On top of that `Image` has a `mode` field the example omits, and `Video` has five more fields that do print.

What they actually print:

| documented | actual |
|---|---|
| `Image(decode=True, id=None)` | `Image(mode=None, decode=True)` |
| `Pdf(decode=True, id=None)` | `Pdf(decode=True)` |
| `Nifti(decode=True, id=None)` | `Nifti(decode=True)` |
| `Video(decode=True, id=None)` | `Video(decode=True, stream_index=None, dimension_order='NCHW', num_ffmpeg_threads=1, device='cpu', seek_mode='exact')` |

Two more in the same blocks:

- `video.py` ends with `...[0]["video]` (unterminated string), so that line does not parse.
- `image.py`'s last block prints a dict directly after `ds = ds.cast_column(...)`, which is an assignment and outputs nothing. `pdf.py` and `nifti.py` have the `>>> ds[0]["image"]` line it is missing.

### What changed

Docs only, no code. Each shown repr replaced with what the class prints, the string literal closed, and the missing line added.

### Verification

By execution, not by eye: the four classes were built from this branch's source and their `repr()` compared against the docstring text, and every `>>>` line in the four blocks was `compile()`d.

**4 wrong reprs and 1 line that does not compile on main, 0 and 0 after.**

`ruff format --check` and `ruff check` clean on all four files.

### DIFF
--- a/src/datasets/features/features.py
+++ b/src/datasets/features/features.py
@@ -2325,8 +2325,8 @@ class Features(dict):
 
             >>> from datasets import Features, List, Value
             >>> # let's say we have two features with a different order of nested fields (for a and b for example)
-            >>> f1 = Features({"root": {"a": Value("string"), "b": Value("string")}})
-            >>> f2 = Features({"root": {"b": Value("string"), "a": Value("string")}})
+            >>> f1 = Features({"root": List({"a": Value("string"), "b": Value("string")})})
+            >>> f2 = Features({"root": List({"b": Value("string"), "a": Value("string")})})
             >>> assert f1.type != f2.type
             >>> # re-ordering keeps the base structure (here List is defined at the root level), but makes the fields order match
             >>> f1.reorder_fields_as(f2)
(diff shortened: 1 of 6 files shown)


vineethsaivs:
Pushed 101bb1a: the same drift in two files this PR had not reached, plus one example that lost its point.

`iterable_dataset.py` documents `Audio(sampling_rate=8000, mono=True, decode=True, id=None)`. `Audio` has no `mono` field at all now, so that line cannot even be constructed, and `id` carries `repr=False`. Also `Image(mode=None, decode=False, id=None)`.

`features.py` `Features.flatten` prints `List(Value(int32), id=None)` twice, and `List` keeps `id` out of its repr too.

`Features.reorder_fields_as` is a different miss. Its example builds `f1` and `f2` from plain dicts, yet its own comment says "here List is defined at the root level", it imports `List` without using it, and the documented output is `{'root': List({...})}`. The `List` wrapper was dropped from the two constructor lines and the output was never updated. Putting it back makes the example agree with its comment, its import and its output. `f1.reorder_fields_as(f2).type == f2.type` still holds, so the closing assert is unchanged.

Every line here was run against main, not reasoned about.

PR #8565, opened 2026-09-04
Add a strategy argument to split_dataset_by_node for iterable datasets

### LINKED ISSUE #8253: Feature request: Add a streaming_shard operator for early sample-level sharding when file-level sharding is insufficient

behroozazarkhalili:
Closes #8253 by adding the explicit splitting-strategy argument that @lhoestq asked for in that thread.

```python
split_dataset_by_node(ds, rank, world_size, strategy="auto" | "shards" | "examples")
```

- `"auto"`: unchanged. Shards are assigned when `num_shards % world_size == 0`, otherwise each node keeps one example in `world_size`.
- `"shards"`: always assigns shards. Raises `ValueError` naming `num_shards` and `world_size` when they do not divide, at split time and again when the iterable is prepared, because `shuffle()` or `shard()` can change the shard count afterwards.
- `"examples"`: always keeps one example in `world_size`, for the case in the issue where shard-level assignment is possible but the shards are imbalanced.

The value is stored on `DistributedConfig`, so every path that copies the config (`shuffle`, `shard`, `map`, `filter`, `take`, `skip`, `interleave_datasets`) keeps it. A nested `split_dataset_by_node` inherits the strategy; passing the same value again is accepted, a different explicit value raises. Map-style datasets accept only `"auto"`. `num_shards` reports the full shard count under `"examples"`.

Tests in `tests/test_distributed.py` cover both new strategies, the union over ranks, the non-divisible error, the nested cases, the post-shuffle check, invalid values, and the map-style rejection. Docs updated in `use_with_pytorch.mdx`.

Three behaviors were checked and found identical on `main` for the existing example-level path, so they are unchanged here and left for separate issues if wanted: `concatenate_datasets` drops the distributed config; `take`/`skip` after an example-level split apply before the per-node step; DataLoader worker sharding under an example-level split does not partition the node's examples evenly.

### DIFF
--- a/docs/source/use_with_pytorch.mdx
+++ b/docs/source/use_with_pytorch.mdx
@@ -250,9 +250,9 @@ Each node is assigned a chunk of data, e.g. rank 0 is given the first chunk of t
 
 For iterable datasets:
 
-If the dataset has a number of shards that is a factor of `world_size` (i.e. if `dataset.num_shards % world_size == 0`),
-then the shards are evenly assigned across the nodes, which is the most optimized.
-Otherwise, each node keeps 1 example out of `world_size`, skipping the other examples.
+The splitting `strategy` can be `"auto"`, `"shards"`, or `"examples"`. The default `"auto"` assigns shards when
+the number of shards is divisible by `world_size` and otherwise assigns every `world_size`-th example. `"shards"`
+always assigns shards and requires divisibility, while `"examples"` always assigns examples.
 
 This can also be combined with a `torch.utils.data.DataLoader` if you want each node to use multiple workers to load the data.
 
--- a/docs/source/use_with_pytorch.mdx
+++ b/docs/source/use_with_pytorch.mdx
@@ -250,9 +250,11 @@ Each node is assigned a chunk of data, e.g. rank 0 is given the first chunk of t
 
 For iterable datasets:
 
-The splitting `strategy` can be `"auto"`, `"shards"`, or `"examples"`. The default `"auto"` assigns shards when
-the number of shards is divisible by `world_size` and otherwise assigns every `world_size`-th example. `"shards"`
-always assigns shards and requires divisibility, while `"examples"` always assigns examples.
+The splitting `strategy` can be `"auto"`, `"shards"`, or `"examples"`:
+
+* `"shards"`: shards are evenly assigned across the nodes
+* `"examples"`: each node keeps 1 example out of `world_size`, skipping the other examples
+* `"auto"` (default): uses `"shards"` if the dataset has a number of shards that is a factor of `world_size` (i.e. if `dataset.num_shards % world_size == 0`), which is the most optimized. Otherwise uses `"examples"`.
 
 This can also be combined with a `torch.utils.data.DataLoader` if you want each node to use multiple workers to load the data.
 
--- a/src/datasets/distributed.py
+++ b/src/datasets/distributed.py
@@ -1,4 +1,4 @@
-from typing import TypeVar
+from typing import Literal, TypeVar
 
 from .arrow_dataset import Dataset, _split_by_node_map_style_dataset
 from .iterable_dataset import IterableDataset, _split_by_node_iterable_dataset
@@ -7,7 +7,9 @@ from .iterable_dataset import IterableDataset, _split_by_node_iterable_dataset
 DatasetType = TypeVar("DatasetType", Dataset, IterableDataset)
 
 
-def split_dataset_by_node(dataset: DatasetType, rank: int, world_size: int) -> DatasetType:
+def split_dataset_by_node(
+    dataset: DatasetType, rank: int, world_size: int, strategy: Literal["auto", "shards", "examples"] = "auto"
+) -> DatasetType:
     """
     Split a dataset for the node at rank `rank` in a pool of nodes of size `world_size`.
 
@@ -18,9 +20,9 @@ def split_dataset_by_node(dataset: DatasetType, rank: int, world_size: int) -> D
 
     For iterable datasets:
 
-    If the dataset has a number of shards that is a factor of `world_size` (i.e. if `dataset.num_shards % world_size == 0`),
-    then the shards are evenly assigned across the nodes, which is the most optimized.
-    Otherwise, each node keeps 1 example out of `world_size`, skipping the other examples.
+    The splitting `strategy` can be `"auto"`, `"shards"`, or `"examples"`. The default `"auto"` assigns shards
+    when the number of shards is divisible by `world_size` and otherwise assigns every `world_size`-th example.
+    `"shards"` always assigns shards and requires divisibility, while `"examples"` always assigns examples.
 
     > [!WARNING]
     > If you shuffle your iterable dataset in a distributed setup, make sure to set a fixed `seed` in [`IterableDataset.shuffle`]
@@ -33,11 +35,18 @@ def split_dataset_by_node(dataset: DatasetType, rank: int, world_size: int) -> D
             Rank of the current node.
         world_size (`int`):
             Total number of nodes.
+        strategy (`str`, defaults to `"auto"`):
+            How to split an iterable dataset. Must be one of `"auto"`, `"shards"`, or `"examples"`.
+            Map-style datasets only support `"auto"`.
 
     Returns:
         [`Dataset`] or [`IterableDataset`]: The dataset to be used on the node at rank `rank`.
     """
+    if strategy not in {"auto", "shards", "examples"}:
+        raise ValueError(f"Invalid strategy {strategy!r}. Expected one of 'auto', 'shards', or 'examples'.")
     if isinstance(dataset, Dataset):
+        if strategy != "auto":
+            raise ValueError(f"Map-style datasets only support strategy='auto', but got strategy={strategy!r}.")
         return _split_by_node_map_style_dataset(dataset, rank=rank, world_size=world_size)
     else:
-        return _split_by_node_iterable_dataset(dataset, rank=rank, world_size=world_size)
+        return _split_by_node_iterable_dataset(dataset, rank=rank, world_size=world_size, strategy=strategy)
--- a/src/datasets/distributed.py
+++ b/src/datasets/distributed.py
@@ -20,9 +20,12 @@ def split_dataset_by_node(
 
     For iterable datasets:
 
-    The splitting `strategy` can be `"auto"`, `"shards"`, or `"examples"`. The default `"auto"` assigns shards
-    when the number of shards is divisible by `world_size` and otherwise assigns every `world_size`-th example.
-    `"shards"` always assigns shards and requires divisibility, while `"examples"` always assigns examples.
+    The splitting `strategy` can be `"auto"`, `"shards"`, or `"examples"`:
+
+    * `"shards"`: shards are evenly assigned across the nodes
+    * `"examples"`: each node keeps 1 example out of `world_size`, skipping the other examples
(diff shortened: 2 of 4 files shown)


behroozazarkhalili:
CI note: the full upstream workflow (ubuntu and windows, py3.10 to 3.14, minimum and latest deps) passed on this exact head (bd41e7d02) on my fork: https://github.com/behroozazarkhalili/datasets/actions/runs/33910020873. Approving the workflow run here should be a formality.

src/datasets/distributed.py
```diff
@@ -18,9 +20,9 @@ def split_dataset_by_node(dataset: DatasetType, rank: int, world_size: int) -> D
 
     For iterable datasets:
 
-    If the dataset has a number of shards that is a factor of `world_size` (i.e. if `dataset.num_shards % world_size == 0`),
-    then the shards are evenly assigned across the nodes, which is the most optimized.
-    Otherwise, each node keeps 1 example out of `world_size`, skipping the other examples.
+    The splitting `strategy` can be `"auto"`, `"shards"`, or `"examples"`. The default `"auto"` assigns shards
+    when the number of shards is divisible by `world_size` and otherwise assigns every `world_size`-th example.
+    `"shards"` always assigns shards and requires divisibility, while `"examples"` always assigns examples.
```
lhoestq: maybe clearer this way ?

```suggestion
    The splitting `strategy` can be `"auto"`, `"shards"`, or `"examples"`:
    
    * `"shards"`: shards are evenly assigned across the nodes
    * `"examples"`: each node keeps 1 example out of `world_size`, skipping the other examples
    * `"auto"` (default): uses `"shards"` if the dataset has a number of shards that is a factor of `world_size`
    (i.e. if `dataset.num_shards % world_size == 0`), which is the most optimized. Otherwise uses `"examples"`.
```

src/datasets/iterable_dataset.py
```diff
@@ -5428,14 +5429,28 @@ def _split_by_node_iterable_dataset(dataset: IterableDataset, rank: int, world_s
...
+        strategy = dataset._distributed.strategy
         rank = world_size * dataset._distributed.rank + rank
         world_size = world_size * dataset._distributed.world_size
-    distributed = DistributedConfig(rank=rank, world_size=world_size)
+    if strategy == "shards" and dataset._ex_iterable.num_shards % world_size != 0:
+        raise ValueError(
+            f"Cannot split an iterable dataset with num_shards={dataset._ex_iterable.num_shards} "
+            f"across world_size={world_size} nodes "
+            "with strategy='shards' because the number of shards must be divisible by world_size."
+        )
```
lhoestq: I think we can allow it, sometimes it's fine if shards are not assigned perfectly evenly

docs/source/use_with_pytorch.mdx
```diff
@@ -250,9 +250,9 @@ Each node is assigned a chunk of data, e.g. rank 0 is given the first chunk of t
 
 For iterable datasets:
 
-If the dataset has a number of shards that is a factor of `world_size` (i.e. if `dataset.num_shards % world_size == 0`),
-then the shards are evenly assigned across the nodes, which is the most optimized.
-Otherwise, each node keeps 1 example out of `world_size`, skipping the other examples.
+The splitting `strategy` can be `"auto"`, `"shards"`, or `"examples"`. The default `"auto"` assigns shards when
+the number of shards is divisible by `world_size` and otherwise assigns every `world_size`-th example. `"shards"`
+always assigns shards and requires divisibility, while `"examples"` always assigns examples.
```
lhoestq:
```suggestion
The splitting `strategy` can be `"auto"`, `"shards"`, or `"examples"`:
    
* `"shards"`: shards are evenly assigned across the nodes
* `"examples"`: each node keeps 1 example out of `world_size`, skipping the other examples
* `"auto"` (default): uses `"shards"` if the dataset has a number of shards that is a factor of `world_size` (i.e. if `dataset.num_shards % world_size == 0`), which is the most optimized. Otherwise uses `"examples"`.
```

lhoestq (commented):
Great ! I just have some minor comments:

src/datasets/distributed.py
behroozazarkhalili: Applied your wording in 4922727a6 (here, in `_split_by_node_iterable_dataset`, and in the PyTorch guide).

docs/source/use_with_pytorch.mdx
behroozazarkhalili: Applied in 4922727a6.

src/datasets/iterable_dataset.py
behroozazarkhalili: Done in 4922727a6: `strategy="shards"` now accepts a shard count that is not a factor of `world_size`. Shards are assigned round-robin (`shard_data_sources(contiguous=False)`), so some nodes get one more shard than others, and `IterableDataset.num_shards` reports the per-node count in that case (test: 5 shards over 3 nodes gives `[2, 2, 1]`).

I kept one check, `num_shards < world_size`, both at split time and at iteration time. Without it a node with zero shards crashes inside `CyclingMultiSourcesExamplesIterable.__iter__` (`ThreadPoolExecutor(max_workers=0)`) rather than failing with a clear message; this is reachable in practice because `shuffle()` interleaves the shards into a single source, so a 6-shard dataset split 3 ways with `strategy="shards"` and then shuffled ends up with `num_shards=1`. If you would rather have zero-shard nodes yield nothing instead of raising, say so and I will change it.

lhoestq (approved):
lgtm !

PR #8566, opened 2026-09-04
Fix Pdf.decode_example raising TypeError on remote non-Hub paths

behroozazarkhalili:
`Pdf.decode_example` on a remote path outside the Hub raised before any download:

```
>>> Pdf().decode_example({"path": "https://example.com/a.pdf", "bytes": None})
TypeError: 'NoneType' object is not subscriptable   # src/datasets/features/pdf.py:158
```

`string_to_dict` returns `None` when the URL does not match the Hub pattern; it does not raise, so the surrounding `except ValueError` never ran. The lookup now checks for `None`, matching `Pdf.embed_storage` and the Image, Audio, Video and Nifti decoders, which already do this. Verified by execution that https, s3 and `hf://…@revision` paths all decode after the change.

Regression test added in `tests/features/test_pdf.py` (monkeypatched `xopen`, real test PDF). The one failing test in that file on `main`, `test_dataset_with_pdf_feature`, is unrelated (`Column` vs `list` assertion) and unchanged here.

Out of scope, shared by every decoder: a revisionless `hf://datasets/{repo}/{path}` URL does not match `HUB_DATASETS_HFFS_URL` (which requires `@{revision}`), so an explicit per-repo token is not applied for that form.

### DIFF
--- a/src/datasets/features/pdf.py
+++ b/src/datasets/features/pdf.py
@@ -154,11 +154,10 @@ class Pdf:
                         if source_url.startswith(config.HF_ENDPOINT)
                         else config.HUB_DATASETS_HFFS_URL
                     )
-                    try:
-                        repo_id = string_to_dict(source_url, pattern)["repo_id"]
-                        token = token_per_repo_id.get(repo_id)
-                    except ValueError:
-                        token = None
+                    source_url_fields = string_to_dict(source_url, pattern)
+                    token = (
+                        token_per_repo_id.get(source_url_fields["repo_id"]) if source_url_fields is not None else None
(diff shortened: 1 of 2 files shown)


behroozazarkhalili:
CI note: the full upstream workflow (ubuntu and windows, py3.10 to 3.14, minimum and latest deps) passed on this exact head (2f7364e8b) on my fork: https://github.com/behroozazarkhalili/datasets/actions/runs/33910022270. Approving the workflow run here should be a formality.

lhoestq (approved):
lgtm :)

PR #8569, opened 2026-09-06
Shard the sources when skipping a distributed IterableDataset

### LINKED ISSUE #8568: skip() after split_dataset_by_node gives every rank the whole dataset
### Describe the bug

Calling `.skip()` on an iterable dataset that already went through `split_dataset_by_node` cancels the node split. Each rank then iterates every shard, so a distributed run sees each example `world_size` times.

`SkipExamplesIterable.shard_data_sources` returns `self` when `split_when_sharding` is False, so the wrapped iterable never gets sharded. `TakeExamplesIterable` handles the same branch by sharding the sources and leaving `n` alone, which is why `.take()` behaves and `.skip()` does not.

```python
from datasets import IterableDataset

def gen(shards):
    for s in shards:
        for i in range(4):
            yield {"i": s * 4 + i}

ds = IterableDataset.from_generator(gen, gen_kwargs={"shards": list(range(4))})
W = 2
print("no skip:", {r: [x["i"] for x in split_dataset_by_node(ds, r, W)] for r in range(W)})
print("skip(1):", {r: [x["i"] for x in split_dataset_by_node(ds, r, W).skip(1)] for r in range(W)})
print("take(3):", {r: [x["i"] for x in split_dataset_by_node(ds, r, W).take(3)] for r in range(W)})
```

On main at `19f69de53015525126e6d3f96859acb90498e364`:

```
no skip: {0: [0, 1, 2, 3, 8, 9, 10, 11], 1: [4, 5, 6, 7, 12, 13, 14, 15]}
skip(1): {0: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15], 1: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]}
take(3): {0: [0, 1, 2], 1: [4, 5, 6]}
```

Both ranks yield the same 15 examples out of a 16 example dataset, so 30 examples come out of an epoch instead of 16.

### Expected behavior

The two ranks stay disjoint, as they do without the skip and as they do with `take`. Each rank should skip `n` from its own shards.

### Environment info

- `datasets` 5.0.2.dev0, main at `19f69de53015525126e6d3f96859acb90498e364`
- Python 3.11, Linux
- no torch needed, `IterableDataset.from_generator` is enough

I have a fix and will open a PR right after this.

ebarkhordar:
`SkipExamplesIterable.shard_data_sources` returned `self` when `split_when_sharding` is False, so a `.skip()` applied after `split_dataset_by_node` threw the sharding away and every rank iterated the whole dataset. `TakeExamplesIterable` already handles that branch by sharding the sources and leaving `n` alone, so this just mirrors it.

The test meant to cover this, `test_iterable_dataset_skip_or_take_after_split_by_node`, has its conditional expression parenthesised so that the skip half asserts a non-empty list rather than comparing it, which is why it stayed green. Fixed the parentheses too, and the three skip cases fail on main once they actually compare.

`tests/test_iterable_dataset.py` and `tests/test_distributed.py` are otherwise unchanged in outcome for me. I could not run the interleave sharding cases locally, they need torch and my env does not have it, so CI should tell.

Fixes #8568

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -2121,7 +2121,12 @@ class SkipExamplesIterable(_BaseExamplesIterable):
                 split_when_sharding=self.split_when_sharding,
             )
         else:
-            return self
+            return SkipExamplesIterable(
+                self.ex_iterable.shard_data_sources(num_shards, index, contiguous=contiguous),
+                n=self.n,
+                block_sources_order_when_shuffling=self.block_sources_order_when_shuffling,
+                split_when_sharding=self.split_when_sharding,
+            )
 
     def reshard_data_sources(self) -> "SkipExamplesIterable":
         return SkipExamplesIterable(
--- a/tests/test_iterable_dataset.py
+++ b/tests/test_iterable_dataset.py
@@ -2417,11 +2417,10 @@ def test_iterable_dataset_skip_or_take_after_split_by_node(method, after_split_b
     if after_split_by_node:
         distributed_dataset = split_dataset_by_node(distributed_dataset, rank=rank, world_size=world_size)
         distributed_dataset = distributed_dataset.skip(count) if method == "skip" else distributed_dataset.take(count)
-        assert (
-            list(true_distributed_dataset)[count:]
-            if method == "skip"
-            else list(true_distributed_dataset)[:count] == list(distributed_dataset)
+        expected = (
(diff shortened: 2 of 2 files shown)


lhoestq (approved):
good point, thanks for the fix !

PR #8577, opened 2026-09-07
Fix infinite repetition of empty iterable datasets

Jokasa7:
Infinite repetition currently restarts an empty input indefinitely. For example, a streaming dataset filtered down to zero rows hangs in `dataset.repeat(None).take(1)` instead of finishing. The same problem affects an empty distributed shard and an Arrow input containing only zero-row tables.

Stop infinite repetition when a fresh pass over the input produces no examples. Allow the first pass to finish and reset before applying the check, because a restored checkpoint can be positioned just after the last example of a non-empty repetition. Keep the existing state dictionary schema and finite-repeat behavior unchanged, and count Arrow rows rather than yielded tables as progress.

Added regression coverage for empty example/Arrow inputs, zero-row tables, checkpoint restoration at and between repetition boundaries, dynamic inputs with an empty first pass, and filtered empty/non-empty ranks.

Validation:

- Six bounded empty-input regressions fail against the original runtime and pass with the fix. An independent subprocess also reproduces the public-API hang on the baseline.
- `python -m pytest -q tests/test_iterable_dataset.py -m "not integration"`: **484 passed, 32 skipped, 2 deselected** on Python 3.12 / Windows with CPU PyTorch. The two deselected tests require network integration; optional-dependency/platform skips remain.
- Both commands from `make quality` pass: `ruff check tests src benchmarks utils setup.py` and `ruff format --check tests src benchmarks utils setup.py` (236 files).
- `git diff --check` passes.

This change was implemented, tested, and independently reviewed with Codex assistance. Infinite repetition is not a polling API: a dynamic input that remains empty for a complete restarted pass now terminates.

### DIFF
--- a/src/datasets/iterable_dataset.py
+++ b/src/datasets/iterable_dataset.py
@@ -2203,10 +2203,19 @@ class RepeatExamplesIterable(_BaseExamplesIterable):
 
     def __iter__(self):
         repeat_index = self._state_dict["repeat_index"] if self._state_dict else 0
+        first_iteration = True
         while True:
             if self.num_times is not None and repeat_index >= max(self.num_times, 0):
                 break
-            yield from self.ex_iterable
+            has_examples = False
+            for key_example in self.ex_iterable:
+                has_examples = True
+                yield key_example
+            # The first iteration may resume at the end of a non-empty input.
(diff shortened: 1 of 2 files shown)


lhoestq (approved):
great !

PR #8583, opened 2026-09-08
Keep every value of nested columns in Dataset.to_csv()

### LINKED ISSUE #6778: Dataset.to_csv() missing commas in columns with lists
### Describe the bug

The `to_csv()` method does not output commas in lists.  So when the Dataset is loaded back in the data structure of the column with a list is not correct.

Here's an example:

Obviously, it's not as trivial as inserting commas in the list, since its a comma-separated file.  But hopefully there's a way to export the list in a way that it'll be imported by `load_dataset()` correctly.

Here's some code to reproduce the bug:

```python
from datasets import Dataset

ds = Dataset.from_dict(
    {
        "pokemon": ["bulbasaur", "squirtle"],
        "type": ["grass", "water"]
    }
)

def ascii_to_hex(text):
    return [ord(c) for c in text]

ds = ds.map(lambda x: {"int": ascii_to_hex(x['pokemon'])})

ds.to_csv('../output/temp.csv')
```

temp.csv then contains:
```

### Expected behavior

ACTUAL OUTPUT:
```
pokemon,type,int
bulbasaur,grass,[ 98 117 108  98  97 115  97 117 114]
squirtle,water,[115 113 117 105 114 116 108 101]
```

EXPECTED OUTPUT:
```
pokemon,type,int
bulbasaur,grass,[98, 117, 108,  98,  97, 115,  97, 117, 114]
squirtle,water,[115, 113, 117, 105, 114, 116, 108, 101]
```

or probably something more like this since it's a CSV file:
```
pokemon,type,int
bulbasaur,grass,"[98, 117, 108,  98,  97, 115,  97, 117, 114]"
squirtle,water,"[115, 113, 117, 105, 114, 116, 108, 101]"
```

### Environment info

### Package Version
Name: datasets
Version: 2.16.1

### Python
version: 3.10.12


### OS Info
PRETTY_NAME="Ubuntu 22.04.4 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.4 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
...
UBUNTU_CODENAME=jammy

ebarkhordar:
`to_csv()` hands the Arrow batch to pandas, and pandas renders a list cell with numpy's repr. So it comes out space separated, and past numpy's summarization threshold of 1000 the middle is elided: a 1500-element list is written as `[   0    1    2 ... 1497 1498 1499]` and 1494 values are gone. Struct columns and ArrayXD columns lose values the same way. `to_json()` on the same column round-trips exactly.

This replaces the numpy arrays with lists before pandas sees the frame, so `str()` prints all of them. Cells come out as `[0, 1, 2]` instead of `[0 1 2]`, which is what Dref360's `apply(list)` workaround in the issue produces, done in the writer instead.

It is a change to to_csv output for those columns, and Dref360 flagged that as wanting a greenlight from HF first. If you'd rather have `json.dumps` there, or not have it at all, say so and I'll change it.

Left `to_sql` alone: it doesn't lose anything, sqlite stored my 1500-element column as the array's raw buffer.

Tests cover a list of 3 and of 1500, a struct holding a list, a list of lists, and an Array2D. All five fail on main. `tests/io/` and the csv tests in `tests/test_arrow_dataset.py` still pass, ruff is clean.

Fixes #6778

### DIFF
--- a/src/datasets/io/csv.py
+++ b/src/datasets/io/csv.py
@@ -1,8 +1,10 @@
 import multiprocessing
 import os
-from typing import BinaryIO, Optional, Union
+from typing import Any, BinaryIO, Optional, Union
 
 import fsspec
+import numpy as np
+import pyarrow as pa
 
 from .. import Dataset, Features, NamedSplit, config
 from ..formatting import query_table
@@ -10,6 +12,18 @@ from ..packaged_modules.csv.csv import Csv
 from ..utils import tqdm as hf_tqdm
 from ..utils.typing import NestedDataStructureLike, PathLike
 from .abc import AbstractDatasetReader
+
+
+def _arrays_to_lists(obj: Any) -> Any:
+    """Replace the numpy arrays inside `obj` by lists, recursively."""
+    if isinstance(obj, np.ndarray):
+        return _arrays_to_lists(obj.tolist())
+    elif isinstance(obj, list):
+        return [_arrays_to_lists(value) for value in obj]
+    elif isinstance(obj, dict):
+        return {key: _arrays_to_lists(value) for key, value in obj.items()}
+    else:
+        return obj
 
 
 class CsvDatasetReader(AbstractDatasetReader):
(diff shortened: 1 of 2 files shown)


ebarkhordar:
I think `[0, 1, 2]` is the right output here, since it is what Dref360's `apply(list)` workaround in #6778 already produces, but `json.dumps` is defensible too and either way it changes `to_csv` output for list, struct and ArrayXD columns. Which do you want? Nothing has run upstream on this one, both fork workflows are still sitting at `action_required`.

PR #8584, opened 2026-09-08
Fix the xml loader, which cannot load any file

LeSingh1:
### The bug

`load_dataset("xml", ...)` fails for **every** input:

```python
open("a.xml", "w").write("<root><a>1</a></root>")
load_dataset("xml", data_files="a.xml", split="train")
```

```
DatasetGenerationError: An error occurred while generating the dataset
  caused by
FileNotFoundError: [Errno 2] No such file or directory:
  '<datasets.utils.file_utils.FilesIterable object at 0x10e41f260>'
```

The path in that message is a `repr`, which is the tell: a `FilesIterable` is being passed to `open()`.

### Cause

`_split_generators` wraps each file:

```python
files = [dl_manager.iter_files(file) for file in files]
```

so `gen_kwargs["files"]` is a list of `FilesIterable`, not paths. `_generate_tables` used to flatten it, but #7943 ("Add `_generate_shards`") removed that:

```diff
-        for file_idx, file in enumerate(itertools.chain.from_iterable(files)):
+        for file_idx, file in enumerate(files):
```

The `_split_generators` side was left wrapping, so the loop now opens the wrappers. `_generate_shards` has the same problem — it yields `FilesIterable` objects where the builder expects paths, so shard identifiers are wrong too.

### The fix

Every other module that calls `iter_files` — `text`, `csv`, `json`, `conll` — takes `base_files` and `files_iterables` and iterates the latter, yielding `base_files` for shards. The modules that keep the single-argument `files` form — `arrow`, `hdf5`, `pandas`, `cache`, `tsfile` — don't call `iter_files` at all.

`xml` was the only module with one foot in each convention. Moving it to the `base_files`/`files_iterables` form fixes both methods and matches its peers. Restoring the single `chain.from_iterable` call would fix loading but leave `_generate_shards` yielding wrappers.

### Verification

There was no `tests/packaged_modules/test_xml.py`, which is how this stayed broken. Added one covering a single file, multiple files, and `encoding`.

Against `main`: **3 failed, 4 passed** (the 4 are config-validation tests that never reach the loader). With the change: **7 passed**.

### DIFF
--- a/src/datasets/packaged_modules/xml/xml.py
+++ b/src/datasets/packaged_modules/xml/xml.py
@@ -40,8 +40,13 @@ class Xml(datasets.ArrowBasedBuilder):
         for split_name, files in data_files.items():
             if isinstance(files, str):
                 files = [files]
-            files = [dl_manager.iter_files(file) for file in files]
-            splits.append(datasets.SplitGenerator(name=split_name, gen_kwargs={"files": files}))
+            files_iterables = [dl_manager.iter_files(file) for file in files]
+            splits.append(
+                datasets.SplitGenerator(
+                    name=split_name,
+                    gen_kwargs={"base_files": files, "files_iterables": files_iterables},
+                )
+            )
         return splits
 
     def _cast_table(self, pa_table: pa.Table) -> pa.Table:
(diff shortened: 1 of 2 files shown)


src/datasets/packaged_modules/xml/xml.py
```diff
@@ -57,14 +62,15 @@ def _cast_table(self, pa_table: pa.Table) -> pa.Table:
...
-                xml = f.read()
-                pa_table = pa.Table.from_arrays([pa.array([xml])], names=pa_table_names)
-                yield (file_idx, 0), self._cast_table(pa_table)
+        for shard_idx, files_iterable in enumerate(files_iterables):
+            for file_idx, file in enumerate(files_iterable):
+                # open in text mode, by default translates universal newlines ("\n", "\r\n" and "\r") into "\n"
+                with open(file, encoding=self.config.encoding, errors=self.config.encoding_errors) as f:
+                    xml = f.read()
+                    pa_table = pa.Table.from_arrays([pa.array([xml])], names=pa_table_names)
+                    yield (shard_idx, file_idx), self._cast_table(pa_table)
```
ebarkhordar, line 76: I ran this against `17d07e0` in a clean container (`python:3.12-slim`, pyarrow + pandas + runtime deps). The loader fix is right and `test_xml.py` passes 7/7 for me.

One thing the move to the peer convention leaves behind: `_generate_tables` now computes `shard_idx`, but yields a bare tuple, and the builder only reads the shard id off a `Key`.

```python
# builder.py:1828
if isinstance(key, Key):  # old custom builders may not use Key
    original_shard_id = key.original_shard_id
```

So `original_shard_id` stays 0, `_generate_shards` reports three shards and the writer records one. Three single-row files, `config.SAVE_ORIGINAL_SHARD_LENGTHS = True` (it gates the record at `builder.py:1793`), otherwise identical:

```
xml (this PR)   rows=3  _generate_shards->3  original_shard_lengths=None
text (peer)     rows=3  _generate_shards->3  original_shard_lengths=[1, 1, 1]
```

That flag is `False` in `config.py:173`, so nobody hits this today. Latent, not a regression, and I would not hold the PR for it.

At `xml.py:76`, with the import beside the others:

```python
from datasets.builder import Key
yield Key(shard_idx, file_idx), self._cast_table(pa_table)
```

The probe then prints `[1, 1, 1]` and test_xml.py still passes 7/7. Keeping `file_idx` rather than a literal `0` is worth it: `iter_files` expands a directory into several files under one shard, and `file_idx` keeps those keys distinct.

I did not check streaming or `num_proc > 1`.

PR #8585, opened 2026-09-08
Reject a non-callable generator in GeneratorConfig

LeSingh1:
### The bug

`GeneratorConfig` checks that `generator` is not `None`, but not that it is callable. A wrong type is only discovered during generation:

```python
Dataset.from_generator([{"a": 1}])
```

```
DatasetGenerationError: An error occurred while generating the dataset
  caused by
TypeError: 'list' object is not callable
```

Passing the data itself instead of a function that yields it is an easy mistake — `from_generator` reads like it takes a generator, and a list or a generator *object* both look plausible. Neither message names `generator`, so there is nothing pointing at the argument.

### The fix

Check it next to the existing `None` check, which is already in `__post_init__` for exactly this reason:

```
TypeError: generator must be callable, but got list
```

### Verification

The new test covers a string, a list, an int and a dict. All four fail on `main` and pass with the change:

```
without the fix:  FFFF
with the fix:     ....
```

The existing `from_generator` tests are unaffected — 18 passed. The diff to `tests/test_arrow_dataset.py` is purely additive; no existing test or parametrization is touched.

### DIFF
--- a/src/datasets/packaged_modules/generator/generator.py
+++ b/src/datasets/packaged_modules/generator/generator.py
@@ -17,6 +17,11 @@ class GeneratorConfig(datasets.BuilderConfig):
         super().__post_init__()
         if self.generator is None:
             raise ValueError("generator must be specified")
+        if not callable(self.generator):
+            # Otherwise it is only called during generation, and the failure arrives as a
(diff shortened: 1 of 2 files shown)


lhoestq (approved):
thx for the improvement :)

PR #8589, opened 2026-09-08
Advance the user-supplied generator in Dataset.shuffle even on cache hits

### LINKED ISSUE #7127: Caching shuffles by np.random.Generator results in unintiutive behavior
### Describe the bug

Create a dataset. Save it to disk. Load from disk. Shuffle, usning a `np.random.Generator`. Iterate. Shuffle again. Iterate. The iterates are different since the supplied np.random.Generator has progressed between the shuffles.

Load dataset from disk again. Shuffle and Iterate. See same result as before. Shuffle and iterate, and this time it does not have the same shuffling as ion previous run.

The motivation is I have a deep learning loop with 

```
for epoch in range(10):
    for batch in dataset.shuffle(generator=generator).iter(batch_size=32):
        .... # do stuff
```
where I want a new shuffling at every epoch. Instead I get the same shuffling.

Run the code below two times. 

```python
import datasets
import numpy as np

generator = np.random.default_rng(0)
ds = datasets.Dataset.from_dict(mapping={"X":range(1000)})
ds.save_to_disk("tmp")
print("First loop: ", end="")
for _ in range(10):
    print(next(ds.shuffle(generator=generator).iter(batch_size=1))['X'], end=", ")
print("")

print("Second loop: ", end="")
ds = datasets.Dataset.load_from_disk("tmp")
for _ in range(10):
    print(next(ds.shuffle(generator=generator).iter(batch_size=1))['X'], end=", ")
print("")
```

The output is:

```
$ python main.py 
Saving the dataset (1/1 shards): 100%|███████████████████████████████████████████████████████████████████████| 1000/1000 [00:00<00:00, 495019.95 examples/s]
First loop: 459, 739, 72, 943, 241, 181, 845, 830, 896, 334, 
Second loop: 741, 847, 944, 795, 483, 842, 717, 865, 231, 840,
$ python main.py 
Saving the dataset (1/1 shards): 100%|████████████████████████████████████████████████████████████████████████| 1000/1000 [00:00<00:00, 22243.40 examples/s]
First loop: 459, 739, 72, 943, 241, 181, 845, 830, 896, 334, 
Second loop: 741, 741, 741, 741, 741, 741, 741, 741, 741, 741, 
```

The second loop, on the second run, only spits out "741, 741, 741...." which is *not* the desired output

### Expected behavior

I want the dataset to shuffle at every epoch since I provide it with a generator for shuffling. 

### Environment info

Datasets version 2.21.0
Ubuntu linux.

BenYang12:
Fixes #7127.

When a `np.random.Generator` is passed to `Dataset.shuffle`, the indices cache lookup returns before `generator.permutation(...)` is called, so the caller's generator never advances on a cache hit. Because the fingerprint is derived from the generator's state, the next call recomputes the same fingerprint, hits the same cache file and returns the same order — so a training loop like `for epoch in ...: ds.shuffle(generator=g)` yields the identical permutation every epoch on the second and later runs, while the first (cold-cache) run shuffles correctly.

This draws the permutation before the cache lookup when the generator comes from the caller, so cold-cache and warm-cache runs produce identical sequences of shuffles and identical generator progression.

 When `shuffle` builds the generator itself, the draw stays lazy, so seed-based shuffling keeps its cache short-circuit and does no extra work. Caching is otherwise unchanged and the public generator API is preserved.

I added `test_shuffle_generator_advances_on_cache_hit`, which compares three successive shuffles cold vs. warm and asserts both the row orders and the final bit-generator state match. It fails on `main` (on-disk parametrization) and passes with this change.

There is a tradeoff:

 on a cache hit with a caller-supplied generator we now compute and discard one permutation — measured at 6 ms for 1M rows and 110 ms for 10M rows, small next to reading the indices file the hit is saving, and it does not affect the `seed=` path at all. I preferred this over fast-forwarding the generator by a computed number of steps, which would depend on undocumented NumPy internals.

Note: `Dataset.train_test_split` has the same pattern and is not addressed here — happy to follow up if you'd like it fixed in the same PR.

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -4939,6 +4939,10 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                 seed = seed[pos] if pos < 624 else seed[0]
                 _ = np.random.random()  # do 1 step of rng
             generator = np.random.default_rng(seed)
+            permutation = None
+        else:
+            # draw before the cache lookup so the caller's generator advances on cache hits too
+            permutation = generator.permutation(len(self))
 
         # Check if we've already cached this computation (indexed by a hash)
         if self.cache_files:
@@ -4951,7 +4955,8 @@ class Dataset(DatasetInfoMixin, IndexableMixin, TensorflowDatasetMixin):
                     fingerprint=new_fingerprint, indices_cache_file_name=indices_cache_file_name
                 )
 
-        permutation = generator.permutation(len(self))
+        if permutation is None:
+            permutation = generator.permutation(len(self))
 
         return self.select(
             indices=permutation,
--- a/tests/test_arrow_dataset.py
+++ b/tests/test_arrow_dataset.py
@@ -2395,6 +2395,22 @@ class BaseDatasetTest(TestCase):
                         self.assertNotEqual(d3["filename"], d2["filename"])
                         self.assertNotEqual(d3._fingerprint, d2._fingerprint)
 
+    def test_shuffle_generator_advances_on_cache_hit(self, in_memory):
+        def successive_shuffles(dset, generator):
+            orders = []
+            for _ in range(3):
+                with dset.shuffle(generator=generator) as dset_shuffled:
+                    orders.append(list(dset_shuffled["filename"]))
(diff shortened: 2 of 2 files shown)


lhoestq (approved):
lgtm !

Issue #2831, opened 2021-08-24
ArrowInvalid when mapping dataset with missing values


uniquefine:
## Describe the bug
I encountered an `ArrowInvalid` when mapping dataset with missing values. 
Here are the files for a minimal example. The exception is only thrown when the first line in the csv has a missing value (if you move the last line to the top it isn't thrown).
[data_small.csv](https://github.com/huggingface/datasets/files/7037838/data_small.csv)
[data.csv](https://github.com/huggingface/datasets/files/7037842/data.csv)

## Steps to reproduce the bug
```python
from datasets import load_dataset

datasets = load_dataset("csv", data_files=['data_small.csv'])

datasets = datasets.map(lambda e: {'labels': e['match']},
                        remove_columns=['id'])
```

## Expected results
No error

## Actual results
```
File "pyarrow/error.pxi", line 84, in pyarrow.lib.check_status
pyarrow.lib.ArrowInvalid: Invalid null value
```

## Environment info
2021-08-24 10:46:45.452199: W tensorflow/stream_executor/platform/default/dso_loader.cc:60] Could not load dynamic library 'libcudart.so.11.0'; dlerror: libcudart.so.11.0: cannot open shared object file: No such file or directory
2021-08-24 10:46:45.452225: I tensorflow/stream_executor/cuda/cudart_stub.cc:29] Ignore above cudart dlerror if you do not have a GPU set up on your machine.
WARNING:tensorflow:From /home/sebastian/.local/lib/python3.8/site-packages/datasets/commands/env.py:36: is_gpu_available (from tensorflow.python.framework.test_util) is deprecated and will be removed in a future version.
Instructions for updating:
Use `tf.config.list_physical_devices('GPU')` instead.
2021-08-24 10:46:46.523091: I tensorflow/core/platform/cpu_feature_guard.cc:142] This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN) to use the following CPU instructions in performance-critical operations:  AVX2 FMA
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
2021-08-24 10:46:46.528272: I tensorflow/compiler/jit/xla_gpu_device.cc:99] Not creating XLA devices, tf_xla_enable_xla_devices not set
[...]

PR #8591, opened 2026-09-09
Better error when a column written as null gets values in a later batch

ebarkhordar:
On #2831 the map still fails at HEAD, as `TypeError: Couldn't cast array of type double to null` now rather than `ArrowInvalid`. Your 2021 comment on that issue has the mechanism and both workarounds; the error itself says none of it.

So this does not change what the writer accepts. Before the cast, `ArrowWriter` now checks whether the schema has the column as null while the batch has values that are not None, and raises:

```
Couldn't write column 'labels': its type in this file is null, because the first 1000 values
written for it were all None. Set the type of 'labels' with features=, or make sure the first
batch written contains a value that is not None.
```

Both remedies are the ones from your comment and I ran them against the csv in the issue.

I did not try to promote the column instead. The stream schema is fixed once the first batch is on disk, so widening it means rewriting what is already written, and #4485 reads like a deliberate call that this cast should refuse rather than drop values.

The check sits in `_write_batch` and in `_write_table`, since both reach the cast and a batched map over arrow tables only goes through the second one.

Two tests, one per path, both fail on main. A third pins that a column which stays all None is still written as null.

Refs #2831

### DIFF
--- a/src/datasets/arrow_writer.py
+++ b/src/datasets/arrow_writer.py
@@ -712,6 +712,23 @@ class ArrowWriter:
         self.write_examples_on_file()  # in case there are buffered examples to write first
         self._write_batch(batch_examples, writer_batch_size=writer_batch_size, try_original_type=try_original_type)
 
+    def _check_null_column(self, col: str, values):
+        """Raise an informative error if a column written as null gets values that are not None.
+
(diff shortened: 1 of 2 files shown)


lhoestq (approved):
lgtm :)

PR #8592, opened 2026-09-09
fix(typing): Use Sequence instead of List for path_or_paths type annotations

CJstate:
This fixes mypy errors when passing list of paths to `Dataset.from_parquet()`, `from_csv()`, `from_json()`, `from_text()`, and `IterableDataset` equivalents.

**Problem**: `List` is invariant while `Sequence` is covariant, so passing a `List[str]` to a function expecting `Union[PathLike, List[PathLike]]` fails type checking.

**Solution**: Change `Union[PathLike, list[PathLike]]` to `Union[PathLike, Sequence[PathLike]]` in type annotations.

**Changes**:
- `src/datasets/arrow_dataset.py`: Updated imports and 5 method signatures (+ `sort()` method)
- `src/datasets/iterable_dataset.py`: Updated imports and 4 method signatures

Closes #5354

### DIFF
--- a/src/datasets/arrow_dataset.py
+++ b/src/datasets/arrow_dataset.py
@@ -35,8 +35,7 @@ import time
 import warnings
 import weakref
 from collections import Counter, defaultdict
-from collections.abc import Iterable, Iterator, Mapping
-from collections.abc import Sequence as Sequence_
+from collections.abc import Iterable, Iterator, Mapping, Sequence
 from copy import deepcopy
 from functools import partial, wraps
 from math import ceil, floor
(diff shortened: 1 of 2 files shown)


CJstate:
Closing as a duplicate. `path_or_paths: Union[PathLike, list[PathLike]]` → `Sequence` is already proposed by at least #8198, #8361, #8067 and #8282 (plus #8549, #8329, #8321 in the same area) — this was the eighth PR on the same two-line type annotation. Dumping another one on the maintainers' queue makes the queue worse, not the typing better.

PR #8594, opened 2026-09-09
Forward max_buffer_input_shards from IterableDatasetDict.shuffle

vineethsaivs:
## What breaks

`load_dataset(..., streaming=True)` without a split returns an `IterableDatasetDict`, and shuffling it cannot reach the shard-interleaving control:

```python
ds = load_dataset("HuggingFaceFW/fineweb", streaming=True)
ds = ds.shuffle(seed=42, buffer_size=10_000, max_buffer_input_shards=4)
# TypeError: IterableDatasetDict.shuffle() got an unexpected keyword argument 'max_buffer_input_shards'
```

#8194 added `max_buffer_input_shards` to `IterableDataset.shuffle` and updated `docs/source/stream.mdx`, but not `dataset_dict.py`, so the per-split method has the argument and the dict wrapper silently does not. Every other `IterableDatasetDict.shuffle` argument is forwarded.

## Second, smaller thing in the same argument

Its docstring entry in `iterable_dataset.py` reads

```
max_buffer_input_shards (`int`, defaults to `101000`):
```

The signature says `max_buffer_input_shards: int = 10`, and `stream.mdx` says "default is 10". `101000` is the `10` and the neighbouring `buffer_size` default of `1000` run together.

## The change

Accept the argument on `IterableDatasetDict.shuffle` with the same default, document it with the same wording, and pass it to each split. Fix the stated default in the `IterableDataset.shuffle` docstring.

## Test

`test_iterable_dataset_dict_shuffle_forwards_max_buffer_input_shards` in `tests/test_iterable_dataset.py`, next to the existing `test_iterable_dataset_shuffle_buffer_uses_multiple_input_shards` and using its assertion: with `max_buffer_input_shards=1` the first ten examples must come from at most two shards, which only holds if the argument reaches the split.

```
tests/test_iterable_dataset.py -k input_shards
  before: 1 failed, 1 passed
          TypeError: IterableDatasetDict.shuffle() got an unexpected keyword argument 'max_buffer_input_shards'
  after:  2 passed

tests/test_iterable_dataset.py tests/test_dataset_dict.py
  559 passed, 9 skipped
```

`ruff check` and `ruff format --check` are clean on the three files.

### DIFF
--- a/src/datasets/dataset_dict.py
+++ b/src/datasets/dataset_dict.py
@@ -2088,6 +2088,7 @@ class IterableDatasetDict(dict[Union[str, NamedSplit], IterableDataset]):
         seed=None,
         generator: Optional[np.random.Generator] = None,
         buffer_size: int = 1000,
+        max_buffer_input_shards: int = 10,
     ) -> "IterableDatasetDict":
         """
         Randomly shuffles the elements of this dataset.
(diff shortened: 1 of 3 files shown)


lhoestq (approved):
good catch !

PR #8598, opened 2026-09-10
[httpx migration] Import httpx from huggingface_hub, require huggingface_hub>=1.31.0

Wauplin:
Direct httpx imports will break when huggingface_hub migrates to httpx2. This updates them to use the compatibility re-export introduced in https://github.com/huggingface/huggingface_hub/pull/4803, following the migration plan in https://github.com/huggingface/huggingface_hub/issues/4802. Goal is to make datasets compatible with both huggingface_hub 1.x and 2.x when the switch to httpx2 happens.

Following https://github.com/huggingface/datasets/pull/8598#issuecomment-5621739103, this also bumps the minimum `huggingface_hub` version and removes the code that was only there to keep working with older releases.

### Minimum version

Bumped to `1.31.0` (in `setup.py` and in the "minimum versions" CI job) rather than `1.30.0`: `huggingface_hub.utils.httpx` landed in 1.31.0, not in 1.30.0 as the migration plan anticipated.

`requests` and `httpx` are also gone from the dependencies. `requests` was only needed for huggingface_hub 0.x, and pinning `httpx<1.0.0` here would defeat the purpose of the re-export — datasets should not constrain the httpx major version, huggingface_hub owns that now.

### Legacy code removed

- the `config.HF_HUB_VERSION >= 1.6.0` gates in `arrow_dataset.py`, `dataset_dict.py`, `iterable_dataset.py`, `load.py`, `tests/fixtures/hub.py` and `tests/test_iterable_dataset.py`, with the `BucketNotFoundError = None` / `HfFileSystemResolvedRepositoryPath = HfFileSystemResolvedPath` fallbacks and the matching `raise ImportError("... requires huggingface_hub>=1.6.0")` guards
- `except EntryNotFoundError` around `dirfs.glob(...)`, marked `# needed for huggingface_hub<=1.7.1`
- the `requests` exceptions caught next to their `httpx` counterparts in `load.py` and in `CONNECTION_ERRORS_TO_RETRY`
- `proxies=download_config.proxies` passed to `hf_hub_download`, which huggingface_hub has silently ignored since 1.0. `DownloadConfig.proxies` is kept for backward compatibility, with a docstring pointing at `HTTP_PROXY` / `set_client_factory`
- in the test helpers: `IS_HF_HUB_1_x` and its two branches in `offline()`, `require_buckets_support_in_huggingface_hub` (buckets are always available now), the `requests` half of `xfail_if_500_502_http_error`, and the `huggingface_hub.file_download.HUGGINGFACE_CO_URL_TEMPLATE` monkeypatch fallback

Once huggingface_hub 2.0 is out, the last step of the plan is to widen the pin to `<3.0`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,7 @@ jobs:
         run: uv pip install --system --upgrade pyarrow huggingface-hub "dill<0.3.9"
       - name: Install dependencies (minimum versions)
         if: ${{ matrix.deps_versions != 'deps-latest' }}
-        run: uv pip install --system pyarrow==24.0.0 huggingface-hub==0.25.0 transformers dill==0.3.1.1
+        run: uv pip install --system pyarrow==24.0.0 huggingface-hub==1.31.0 transformers dill==0.3.1.1
       - name: Print dependencies
         run: uv pip list
       - name: Test with pytest
--- a/setup.py
+++ b/setup.py
@@ -116,9 +116,6 @@ REQUIRED_PKGS = [
     "dill>=0.3.0,<0.4.2",  # tmp pin until dill has official support for determinism see https://github.com/uqfoundation/dill/issues/19
     # For performance gains with apache arrow
     "pandas",
-    # for downloading datasets over HTTPS
-    "requests>=2.32.2",
-    "httpx<1.0.0",
     # progress bars in downloads and data operations
     "tqdm>=4.66.3",
     # for fast hashing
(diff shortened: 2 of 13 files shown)


Wauplin:
Failing CI seems unrelated 

```
FAILED tests/test_hub.py::test_delete_from_hub - huggingface_hub.errors.HfHubHTTPError: Server error '502 Bad Gateway' for url 'https://hub-ci.huggingface.co/api/datasets/__DUMMY_DATASETS_USER__/test-dataset-14a7d7-17890532107233/commit/main'
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502
==== 1 failed, 3159 passed, 180 skipped, 232 warnings in 322.78s (0:05:22) =====
```

lhoestq:
cool ! I guess we need to bump the minimal version from 0.25.0 to 1.30.0 (both in setup.py and in the "minimum version" CI)

Wauplin:
@lhoestq good idea. PR now removes all dead code that was needed for legacy versions of `huggingface_hub` (especially to gracefully raise on versions not supporting buckets, and on versions running on `requests`)


Failing CI all come from HTTP 500 from staging (tests are flaky on hfh as well nowadays :grimacing:)

lhoestq (approved):
awesome !

PR #8613, opened 2026-09-12
Resolve a legacy dataset name to its canonical repo id

### LINKED ISSUE #8612: load_dataset on a dataset name with no namespace raises HfUriError
`load_dataset` on a dataset name with no namespace fails at HEAD and on 5.0.1:

```python
>>> load_dataset("imdb", split="train[:5]")
...
  File "src/datasets/load.py", line 609, in get_module
    standalone_yaml_path = cached_path(
  File "src/datasets/utils/file_utils.py", line 180, in cached_path
    ).resolve_path(url_or_filename)
huggingface_hub.errors.HfUriError: Invalid HF URI 'hf://datasets/imdb@e6281661ce1c48d982bc483cf8a173c1bbeb5d31/.huggingface.yaml'. Repository id must be 'namespace/name', got 'imdb'.
```

Same for `squad`, `glue`, `rotten_tomatoes`, `ag_news`, `cnn_dailymail`, `mnist` and `wikitext`.

The Hub still redirects these names, and the earlier part of the load path uses them happily: `api.hf_hub_download(repo_id="imdb", ...)` returns the README and a real commit sha. `HubDatasetModuleFactory.get_module` then builds an `hf://` URI out of the same raw string, and `parse_hf_uri` requires `namespace/name`.

Environment: datasets 5.0.2.dev0 at d336dcb8 and released 5.0.1, huggingface_hub 1.31.0, python 3.11.

ebarkhordar:
`load_dataset("imdb")` dies with an `HfUriError` about a `.huggingface.yaml` nobody asked for, and so do `squad`, `glue`, `rotten_tomatoes`, `ag_news`, `cnn_dailymail`, `mnist` and `wikitext`.

The Hub still redirects these names, so the first half of the load works: `hf_hub_download(repo_id="imdb", ...)` returns the card and a real sha. Then `get_module` puts that raw string into an `hf://` URI, and `parse_hf_uri` wants `namespace/name`.

So resolve it once at the top of `get_module` and use it for the two `hf_dataset_url` calls, `base_path`, the viewer lookup and `builder_kwargs`. `base_path` and `repo_id` have to move together: `DatasetBuilder.__init__` checks every data file sits under `hf://datasets/{repo_id}@`, so canonicalising one and not the other trips that check.

Only a name without a `/` pays for the extra `dataset_info` call, so namespaced datasets make the same number of Hub requests as before. I counted them on both sides to be sure.

One thing you may not expect: the viewer lookup 404s with "The dataset has been renamed" for these names, so `get_exported_dataset_infos` was raising and getting swallowed for every legacy name. It returns the export now. I checked that one by hand rather than in the test, since `use_exported_dataset_infos` is off by default in the factory, and for imdb the card carries the same features either way, so you would not have noticed from the outside.

Test is an integration one like the other `HubDatasetModuleFactory` tests, and it fails on main with the error above. I also checked the eight names load end to end, and that `"imdb"` and `"stanfordnlp/imdb"` land in the same cache directory.

Offline is still canonical-only: `HF_HUB_OFFLINE=1` plus `load_dataset("ag_news")` raises `ConnectionError` before any of this runs, same as it does today, since the cached-module lookup globs the cache by the name it was given. Left alone rather than guessed at.

Happy to turn this into a clear error naming the canonical id instead, if you'd rather these names stop resolving.

Fixes #8612

### DIFF
--- a/src/datasets/load.py
+++ b/src/datasets/load.py
@@ -583,9 +583,11 @@ class HubDatasetModuleFactory(_DatasetModuleFactory):
             library_version=__version__,
             user_agent=get_datasets_user_agent(self.download_config.user_agent),
         )
+        # a legacy name like "imdb" is still served by the Hub but is not a valid repo id in an hf:// URI
+        repo_id = self.name if "/" in self.name else api.dataset_info(self.name, revision=self.commit_hash).id
         try:
             dataset_readme_path = api.hf_hub_download(
-                repo_id=self.name,
+                repo_id=repo_id,
                 filename=config.REPOCARD_FILENAME,
                 repo_type="dataset",
                 revision=self.commit_hash,
@@ -598,7 +600,7 @@ class HubDatasetModuleFactory(_DatasetModuleFactory):
             download_config.download_desc = "Downloading standalone yaml"
         try:
             standalone_yaml_path = cached_path(
-                hf_dataset_url(self.name, config.REPOYAML_FILENAME, revision=self.commit_hash),
+                hf_dataset_url(repo_id, config.REPOYAML_FILENAME, revision=self.commit_hash),
                 download_config=download_config,
             )
             with open(standalone_yaml_path, encoding="utf-8") as f:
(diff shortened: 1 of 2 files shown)


ebarkhordar:
The one thing I can't decide for you: should `load_dataset("imdb")` keep resolving, or is a clear error naming `stanfordnlp/imdb` the behaviour you want now? I built the resolving one, but the error is a much smaller change if these names are meant to be dead.

PR #8614, opened 2026-09-12
Fingerprint functions from the user's own modules by value

### LINKED ISSUE #3297: .map() cache is wrongfully reused - only happens when the mapping function is imported
## Describe the bug
When `.map` is used with a mapping function that is imported, the cache is reused even if the mapping function has been modified.
The reason for this is that `dill` that is used for creating the fingerprint [pickles imported functions by reference](https://stackoverflow.com/a/67851411).

I guess it is not a widespread case, but it can still lead to unwanted results unnoticeably. 

## Steps to reproduce the bug
Create files `a.py` and `b.py`:
```python
# a.py

def main():
    squad = load_dataset("squad")
    squad.map(mapping_func, batched=True)

def mapping_func(examples):
    ID_LENGTH = 4
    examples["id"] = [id_[:ID_LENGTH] for id_ in examples["id"]]
    return examples

if __name__ == "__main__":
    main()
```
```python
# b.py
from a import mapping_func

def main():
    squad = load_dataset("squad")
    squad.map(mapping_func,batched=True)

if __name__ == "__main__":
    main()
```
Run `python b.py` twice: In the first run you will see tqdm bars showing that the data is processed, and in the second run you will see "Loading cached processed dataset at...".
Now change ID_LENGTH to another number in order to change the mapping function, and run `python b.py` again. You'll see that `.map` loads from the cache the result of the previous mapping function.

## Expected results
Run `python a.py` twice: In the first run you will see tqdm bars showing that the data is processed, and in the second run you will see "Loading cached processed dataset at...".
Now change `ID_LENGTH` to another number in order to change the mapping function, and run `python a.py` again. You'll see that the dataset is being processed and that there's no reuse of the previous mapping function result.

## Workaround
Put the mapping function inside a dummy class as a static method:
```python
# a.py
class MappingFuncClass:
    @staticmethod
    def mapping_func(examples):
        ID_LENGTH = 4
        examples["id"] = [id_[:ID_LENGTH] for id_ in examples["id"]]
        return examples
```
```python
# b.py
from a import MappingFuncClass

def main():
    squad = load_dataset("squad")
    squad.map(MappingFuncClass.mapping_func,batched=True)

if __name__ == "__main__":
    main()
```

## Environment info

- `datasets` version: 1.15.1
- Platform: Linux-4.4.0-19041-Microsoft-x86_64-with-glibc2.17
- Python version: 3.8.10
- PyArrow version: 4.0.1

ebarkhordar:
`Hasher.hash` builds the `map()` fingerprint from `_dill.dumps`, and dill dumps a function it can import as a module-qualified name. For a mapping function imported from the user's own module the dump is 35 bytes of module and qualname with no code object in it, so editing the body leaves the fingerprint unchanged and `map()` reloads the previous run's data, with no warning. The same function defined in `__main__` is already dumped by value, body included.

`Pickler.save` already intercepts `FunctionType` to unwrap `torch.compile` wrappers. A function whose defining module lives outside the standard library and site-packages is now dumped by value there. That is the split you suggested on the issue in 2021: a module that only changes when its package is reinstalled stays by reference, source the user edits in place does not.

One consequence worth your call: fingerprints change for any `map()` that takes a function imported from a local module, so those caches are recomputed once after upgrading.

Verified:
- The new test in `tests/test_fingerprint.py` fails without the change. Reverting only the one-line call and keeping the new helpers still fails it, so it is pinned to the fix rather than to the rest of the diff.
- End to end on an on-disk dataset, editing the imported function returns the stale rows on main and the correct rows here.
- `tests/test_fingerprint.py` 26 passed, `tests/test_py_utils.py` 72 passed, `tests/test_arrow_dataset.py -k map` 50 passed, and both `check_code_quality` steps, on Python 3.10.

I have not measured the hashing cost for a function with very large globals. An editable install counts as local source, so its functions are dumped by value too.

Fixes #3297

### DIFF
--- a/src/datasets/utils/_dill.py
+++ b/src/datasets/utils/_dill.py
@@ -15,6 +15,8 @@
 
 import os
 import sys
+import sysconfig
+from functools import lru_cache
 from io import BytesIO
 from types import CodeType, FunctionType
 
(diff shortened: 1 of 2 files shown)


ebarkhordar:
The part that needs your call is the one-time cache recompute: after this, any `map()` taking a function from a local module refingerprints once on upgrade. If that is too broad, narrowing it to editable installs only is a couple of lines.

PR #8644, opened 2026-09-18
fix(py_utils): save/restore NPU RNG state in temp_seed

li-lizhe:
## Description

When `set_pytorch=True` is passed to `temp_seed`, the context manager saves and restores the PyTorch CPU and CUDA RNG states. On non-CUDA accelerators (e.g. Ascend NPU), the device RNG state is silently left unseeded during the context, breaking reproducibility across seeded dataset transforms.

This PR adds an NPU branch alongside the existing CUDA branch: when `torch.npu` is available, the NPU RNG state is saved, seeded, and restored just like CUDA.

## Changes

`src/datasets/utils/py_utils.py` — 9 insertions, 1 deletion

- Initialize `torch_cuda_states = None` before the CUDA availability check (makes the restore-side guard more robust)
- Add `torch_npu_states` save/seed before the context and restore after, following the same pattern as CUDA

## Verification

- Tested: Python syntax passes (`py_compile`), existing behavior unchanged (CUDA path is identical, NPU path only activates when `torch.npu.is_available()`)
- Not tested: actual NPU runtime — requires an Ascend NPU environment

Fixes the reproducibility gap on Ascend NPU, part of broader non-CUDA accelerator support in `datasets`.


ebarkhordar (commented):
I went looking for the call path this fixes, at 7fba5ab7.

`temp_seed` has no caller in `src/datasets`. Parsing all 256 `.py` files in the repo for a call to it (ast, so a renamed or attribute-style call cannot hide) returns 8 call sites, every one in tests, and only two of them pass `set_pytorch=True`:

```
tests/test_py_utils.py:149      kwargs=['set_tensorflow']
tests/test_py_utils.py:151      kwargs=['set_tensorflow']
tests/test_py_utils.py:168      kwargs=['set_pytorch']
tests/test_py_utils.py:170      kwargs=['set_pytorch']
tests/test_py_utils.py:181      kwargs=[]
tests/test_py_utils.py:183      kwargs=[]
tests/test_arrow_dataset.py:2391 kwargs=[]
tests/test_arrow_dataset.py:2392 kwargs=[]
```

History says the same on a clone with full history. Two commits ever carried the string `set_pytorch=True`: #466, which added the parameter together with that test, and #3875, which moved the test file during the v2.0 namespace cleanup. The last library caller of `temp_seed` at all was `Metric.compute`, and it called `temp_seed(self.seed)` with no accelerator flags; it went away with #6983.

So the new branch is reachable from a direct call to the utility, not from a dataset transform. Which caller did you have in mind? That is the part I could not answer from the diff, and it is what decides whether a branch CI cannot exercise is worth carrying.

I did not find anything wrong with the branch itself. I have no Ascend device, so I did not run the NPU path.

li-lizhe:
Thanks for the careful trace @ebarkhordar. To answer your question directly: my intended caller is the public `temp_seed` API used by downstream applications to make a seeded dataset-processing step reproducible. You are correct that `temp_seed(..., set_pytorch=True)` has no internal caller in `src/datasets` today — the last internal caller (`Metric.compute`) went away with #6983, and it never passed accelerator flags anyway.

The branch is not reachable from a stock dataset transform; it is reachable from a direct call to the utility. The value of the change is closing a documented reproducibility gap for those direct callers on non-CUDA accelerators (the CPU/CUDA RNG state is already saved/restored; without this branch the NPU device RNG state is silently left unseeded). It is a small, self-contained addition that only activates when `torch.npu` is importable, so it adds no cost on CUDA/CPU. I understand you have no Ascend device to run the NPU path — same here, and I flagged that honestly in the PR body.

If the maintainers prefer not to carry a branch CI cannot exercise, I am happy to close this PR — just let me know. Otherwise, please advise if any changes are wanted.

ebarkhordar:
Thanks, that answers it. Direct callers of the public `temp_seed` are a reachability story the diff on its own did not show, and having it stated in the thread is what I was after. Whether a branch CI cannot exercise is worth carrying is a maintainer call rather than mine, and I have no Ascend device to add coverage with.

PR #8646, opened 2026-09-20
Fix map(drop_last_batch=True) returning the input unchanged when all batches are dropped

### LINKED ISSUE #8386: map(batched=True, drop_last_batch=True) returns the input unchanged when the dataset is smaller than batch_size

SIDDARTHAREDDY8:
Fixes #8386.

When `drop_last_batch=True` and a shard is smaller than `batch_size`, `_map_single` iterates zero batches, so the map function never runs and `update_data` stays `None`. The final branch then yielded the original shard back unchanged (3 rows instead of 0), disagreeing with `IterableDataset.batch` and `Dataset.iter`, which correctly yield nothing. `Dataset.batch()` (built on `map`) inherited the bug and additionally returned the unbatched schema.

The fix adds an `elif update_data is None and len(shard) > 0` branch that yields an empty `Dataset` instead of the input shard. Since the function never ran, the output schema is unknown, so the empty dataset falls back to the input features, honoring explicit `features=` and `remove_columns` (this covers `Dataset.batch()`, which passes `List(...)` features explicitly).

Test notes:
- Reproduced the bug on current main (`3e2c1a6c3`): `ds.map(lambda b: b, batched=True, batch_size=5, drop_last_batch=True)` on a 3-row dataset returned 3 rows; after the fix it returns 0 rows with the expected features (also verified `features=`, `remove_columns=`, `Dataset.batch()`, and `num_proc=2`).
- Added regression test `test_map_drop_last_batch_all_batches_dropped` in `tests/test_arrow_dataset.py` (passes).
- Ran existing tests: `pytest tests/test_arrow_dataset.py -k "map and not num_proc and not remote"` — 48 passed, 10 skipped.


src/datasets/arrow_dataset.py
```diff
@@ -3869,6 +3869,25 @@ async def async_apply_function(pa_inputs, indices, offset=0):
             processed_inputs = await function(*fn_args, *additional_args, **fn_kwargs)
             return prepare_outputs(pa_inputs, inputs, processed_inputs)
 
+        def _empty_mapped_shard() -> "Dataset":
+            # Build the empty dataset returned when the map function never ran
+            # because all the batches were dropped (`drop_last_batch=True`).
+            # The function output schema is unknown, so fall back to the input
+            # features, honoring explicit `features` and `remove_columns`.
+            result_features = features.copy() if features is not None else shard.features.copy()
```
ebarkhordar: I ran this against b3f38996 in a clean `python:3.11-slim` container.

`remove_columns` names input columns dropped before the function runs; an explicit `features` describes the output schema. This line applies the first to the second, so a column named in both goes missing from the empty result. `init_buffer_and_writer` does not do that: when `features` is given it reaches the writer verbatim under `update_features=False`, so the non-empty path keeps the column.

Same call twice, differing only in whether the last batch is dropped:

```python
fn = lambda b: {"text": [str(n) for n in b["n"]]}
feats = Features({"n": Value("int64"), "text": Value("string")})
kw = dict(batched=True, drop_last_batch=True, features=feats, remove_columns=["text"])
Dataset.from_dict({"text": list("abcd"), "n": [1,2,3,4]}).map(fn, batch_size=2, **kw).features
# {'n': Value('int64'), 'text': Value('string')}
Dataset.from_dict({"text": list("abc"), "n": [1,2,3]}).map(fn, batch_size=10, **kw).features
# {'n': Value('int64')}
```

Guarding the strip so it only runs in the `features is None` fallback makes the two match, and leaves your new test passing:

```python
if features is not None:
    result_features = features.copy()
else:
    result_features = shard.features.copy()
    if remove_columns:
        for column in remove_columns:
            if column in result_features:
                del result_features[column]
```

Your test covers `features` and `remove_columns` separately, so the combination slips past it.

SIDDARTHAREDDY8:
Done. The empty-shard path now applies `remove_columns` only in the `features is None` fallback: when an explicit `features` is given, it describes the output schema and is used verbatim, matching the non-empty path (`init_buffer_and_writer`). Also extended the regression test with the combined `features` + `remove_columns` case from your repro (a column named in both now stays in the empty result, same as the non-empty path).

tests/test_arrow_dataset.py
```diff
@@ -4942,6 +4942,49 @@ def test_dataset_batch():
...
+    assert len(mapped) == 0
+    assert list(mapped.features) == ["c"]
+
+    mapped = ds.map(lambda batch: batch, batched=True, batch_size=5, drop_last_batch=True, remove_columns=["b"])
+    assert len(mapped) == 0
+    assert list(mapped.features) == ["a"]
+
+    # Combined `features` and `remove_columns`: an explicit `features` describes
+    # the output schema, so `remove_columns` must not strip columns from it
+    mapped = ds.map(
```
ebarkhordar: I ran the new test against `0b7062bd` in a clean `python:3.11-slim` container. The code change is right: with an explicit `features`, the empty and non-empty paths now return the same schema. The test you added to pin it does not reach that code.

`ds` is `{"a", "b"}`, so `remove_columns=["text"]` names a column the input does not have, and `map` rejects it at `arrow_dataset.py:3392` before any batching happens.

```
$ python -m pytest tests/test_arrow_dataset.py::test_map_drop_last_batch_all_batches_dropped -q
E   ValueError: Column to remove ['text'] not in the dataset. Current columns in the dataset: ['a', 'b']
src/datasets/arrow_dataset.py:3392: ValueError
1 failed in 1.88s
```

For `features` and `remove_columns` to disagree, the receiver has to actually hold the column being removed:

```python
ds_with_text = Dataset.from_dict({"a": [1, 2, 3], "text": ["p", "q", "r"]})
mapped = ds_with_text.map(
    lambda batch: {"text": [str(n) for n in batch["a"]]},
    batched=True, batch_size=5, drop_last_batch=True,
    features=Features({"a": Value("int64"), "text": Value("string")}),
    remove_columns=["text"],
)
```

With only the receiver swapped and your two assertions unchanged, the test passes at this head, and it still fails on the previous one at the `list(mapped.features) == ["a", "text"]` line, so it keeps pinning the fix.

SIDDARTHAREDDY8:
Fixed: the combined `features` + `remove_columns` case now runs on a dataset that actually holds the removed column (`{"a": [1,2,3], "text": ["p","q","r"]}`), so the test reaches the fixed empty-shard path instead of failing early at the `remove_columns` validation. The two assertions are unchanged. Thanks for catching the gap.

PR #8647, opened 2026-09-20
Fix IterableDataset.filter() with no function

YaoxinHuang:
`list(dataset.filter())` raises `TypeError` because the iterator tries to call `None`. The documented default is to keep all rows.

Return a new dataset wrapper when no function is supplied. Returning a new object, rather than `self`, keeps changes to its epoch from affecting the input dataset.

Tests cover the default argument, explicit `None`, batching, input columns and indices, four output formats, empty inputs, and epoch independence.

## Tests

Tested locally on Windows with Python 3.12.14, against main at `3e2c1a6c33b883fd9710f55f7b6fa4ab64d0ee07`:

- New regression cases before the implementation change: **18 failed, 2 passed**.
- The same regression cases after the fix: **20 passed**.
- `python -m pytest tests/test_iterable_dataset.py -q -rs --tb=short -k "not interleave_dataset_with_sharding" --basetemp=<short-temporary-path>`: **505 passed, 46 skipped, 4 deselected**.
- `python -m ruff check src/datasets/iterable_dataset.py tests/test_iterable_dataset.py`: passed.
- `python -m ruff format --check src/datasets/iterable_dataset.py tests/test_iterable_dataset.py`: passed.

The four deselected cases directly import PyTorch, which is not installed in this CPU-only test environment. Existing skips cover optional Torch/TensorFlow/torchdata dependencies and Windows/NumPy limitations. A short pytest temporary directory was used to avoid Windows path-length errors. The full repository suite and GPU tests were not run.


tests/test_iterable_dataset.py
```diff
@@ -2345,6 +2345,37 @@ def test_iterable_dataset_filter(dataset: IterableDataset) -> None:
     assert next(iter(filtered_dataset)) == {"id": 1}
 
 
+@pytest.mark.parametrize("format_type", [None, "numpy", "arrow", "pandas"])
```
ebarkhordar, line 2348: I ran this against `c09152ed` in a clean container (python:3.12-slim, `pip install -e .`, plus polars).

The early return is right, but it changes what a table format does here, and the green `arrow`/`pandas` legs in this matrix should not be read as `filter` working under those formats. At origin/main, `with_format("arrow").filter()` already raised, just with a different error, and `filter` with a function still raises at this head:

```
TABLE arrow    filter(fn)=ValueError: The Arrow-formatted FilteredExamplesIterable has underlying
                          iterable that is a FormattedExamplesIterable instead of a
                          RebatchedArrowExamplesIterable.   | map(fn)=OK(3) | filter()=OK(3)
      numpy    filter(fn)=OK(3)                             | map(fn)=OK(3) | filter()=OK(3)
TABLE pandas   filter(fn)=ValueError: (same)                | map(fn)=OK(3) | filter()=OK(3)
TABLE polars   filter(fn)=ValueError: (same)                | map(fn)=OK(3) | filter()=OK(3)
```

Fixture was `Dataset.from_dict({"id": [0, 1, 2]}).to_iterable_dataset().with_format(fmt)`. The three marked formats are the `TableFormatter` entries in `_FORMAT_TYPES`, read off the registry rather than guessed.

So after this PR the no-function call is the only `filter` that works under a table format, and this test covers exactly that case.

None of it is caused by your change. `_map` has the ordering for it at `iterable_dataset.py:3702-3715`: `FormattedExamplesIterable` first, then `RebatchedArrowExamplesIterable(force_convert_to_arrow=True)` outside it. `filter` at `:3825-3838` nests the two the other way round, so the `formatting.is_table` check at `:1389` can never pass. #7207 added that branch to `map` and not to `filter` in one commit, and its own title names filter.

Worth a line in the test saying the function path is untested for table formats, so a later reader does not take these legs for coverage it does not have. I have not tried the fix itself.

PR #8649, opened 2026-09-20
Apply the bit-to-byte conversion to binary units too

Rodrigo-Palma:
`convert_file_size_to_int` already distinguishes bits from bytes: a unit ending in a lowercase `b` is read as bits and divided by 8. That check is only applied to the decimal branches, though, so the binary ones return a bit size as if it were bytes, eight times larger than asked for:

```python
>>> from datasets.utils.py_utils import convert_file_size_to_int
>>> convert_file_size_to_int("1MB"), convert_file_size_to_int("1Mb")
(1000000, 125000)       # decimal: bits handled
>>> convert_file_size_to_int("1MiB"), convert_file_size_to_int("1Mib")
(1048576, 1048576)      # binary: bits ignored, "1Mib" should be 131072
```

This applies the same test to the `KiB`/`MiB`/`GiB`/`TiB`/`PiB` branches.

Byte sizes are untouched, which covers everything the library itself passes: `MAX_SHARD_SIZE` (`"500MB"`) and `MAX_ROW_GROUP_SIZE` (`"100MB"`) resolve to the same numbers as before. The values that change are only those a user writes with a lowercase `b` on a binary unit, for example `max_shard_size="500Mib"`.

The function had no tests, so this adds a parametrized one covering both unit families in both spellings, plus the `int` passthrough. `tests/test_py_utils.py` is green (85 passed, 2 skipped), and `ruff check` and `ruff format --check` are clean.

Note that the binary branches here go further than the `transformers` copy of this helper, which only has `GIB`/`MIB`/`KIB`: `PiB` and `TiB` were added on this side, so the two have already diverged and this does not put them further apart in behaviour for anything `transformers` accepts.


Rodrigo-Palma:
The runs here are still waiting for maintainer approval, so I ran this repository's CI on my fork at the same commit as this PR (`50a1c3b`):

https://github.com/Rodrigo-Palma/datasets/actions/runs/35598935223

`check_code_quality` passed and every **unit** job passed: ubuntu and windows, `deps-latest` and `deps-minimum`, plus `test_py314` and `test_py314_future` on both platforms. `Build PR Documentation` also passed (https://github.com/Rodrigo-Palma/datasets/actions/runs/35598936211).

To be upfront about what is red there: the four **integration** jobs failed, but not for a reason related to this change. They talk to `hub-ci.huggingface.co` as `__DUMMY_DATASETS_USER__` and get 404s and 500s, because a fork does not have the CI hub credentials. `ci.yml` already marks those jobs `continue-on-error: ${{ matrix.test == 'integration' }}`, which is why the run as a whole is green.

Locally, `pytest tests/test_py_utils.py` is 78 passed / 2 skipped, and `ruff check` and `ruff format --check` are clean on both files.

PR #8658, opened 2026-09-21
Fix push_to_hub for IterableDataset with unknown features

davanstrien:
An IterableDataset can have unknown features (features=None), e.g. when built with IterableDataset.from_generator() without `features=`, or when streaming CSV files (or JSON files that require a full read) without declared features.

IterableDataset._push_parquet_shards_to_hub() then crashed at the decodable-columns check with:

    AttributeError: 'NoneType' object has no attribute 'items'

since self._info.features is None and embed_external_files defaults to True. This affected the single-dataset push (repo and bucket paths) as well as IterableDatasetDict.push_to_hub().

Resolve the features from the first examples at the top of _push_parquet_shards_to_hub(), the choke point shared by all push paths, as already done in concatenate_datasets() and interleave_datasets(). The rest of the push path already handles freshly resolved features.

Feel free to close and treat as an issue if you want to do it differently!


lhoestq (approved):
wow thx for fixing this

Issue #3419, opened 2021-12-11
`.to_json` is extremely slow after `.select`


eladsegal:
## Describe the bug
Saving a dataset to JSON with `to_json` is extremely slow after using `.select` on the original dataset.

## Steps to reproduce the bug
```python
from datasets import load_dataset

original = load_dataset("squad", split="train")
original.to_json("from_original.json")  # Takes 0 seconds

selected_subset1 = original.select([i for i in range(len(original))])
selected_subset1.to_json("from_select1.json")  # Takes 212 seconds

selected_subset2 = original.select([i for i in range(int(len(original) / 2))])
selected_subset2.to_json("from_select2.json")  # Takes 90 seconds
```

## Environment info

- `datasets` version: master (https://github.com/huggingface/datasets/commit/6090f3cfb5c819f441dd4a4bb635e037c875b044)
- Platform: Linux-4.4.0-19041-Microsoft-x86_64-with-glibc2.27
- Python version: 3.9.7
- PyArrow version: 6.0.0

lhoestq:
Hi ! It's slower indeed because a datasets on which `select`/`shard`/`train_test_split`/`shuffle` has been called has to do additional steps to retrieve the data of the dataset table in the right order.

Indeed, if you call `dataset.select([0, 5, 10])`, the underlying table of the dataset is not altered to keep the examples at index 0, 5, and 10. Instead, an indices mapping is added on top of the table, that says that the first example is at index 0, the second at index 5 and the last one at index 10.

Therefore accessing the examples of the dataset is slower because of the additional step that uses the indices mapping.

The step that takes the most time is to query the dataset table from a list of indices here:

https://github.com/huggingface/datasets/blob/047dc756ed20fbf06e6bcaf910464aba0e20610a/src/datasets/formatting/formatting.py#L61-L63

In your case it can be made significantly faster by checking if the indices are contiguous. If they're contiguous, we could pass a python `slice` or `range` instead of a list of integers to `_query_table`. This way `_query_table` will do only one lookup to get the queried batch instead of `batch_size` lookups.

Given that calling `select` with contiguous indices is a common use case I'm in favor of implementing such an optimization :)
Let me know what you think

eladsegal:
Hi, thanks for the response!
I still don't understand why it is so much slower than iterating and saving:
```python
from datasets import load_dataset

original = load_dataset("squad", split="train")
original.to_json("from_original.json")  # Takes 0 seconds

[...]

PR #8662, opened 2026-09-22
Speed up JSON export of indexed datasets

aprylewu:
Fixes the remaining indexed-export slowdown described in #3419.

Non-contiguous selections and shuffles still leave an indices mapping. Querying a JSON export batch then produces one Arrow chunk per selected row, making conversion to pandas expensive. Combine the queried batch before conversion, while keeping the existing path for unindexed data and falling back if Arrow cannot combine a nested array. The source dataset is not flattened or copied as a whole.

The original contiguous-selection example is already optimized on current main; this change targets selections that actually retain indices.

Validation:
- `PYTHONPATH=src pytest tests/io/test_json.py -q`: 63 passed. Added checks cover duplicate/reordered rows, chunked input, nested and Json features, nullable integers above 2**53, JSON orientations, multiprocessing, and the Arrow fallback.
- Ruff checks/format and `git diff --check` pass.
- `PYTHONPATH=src python benchmarks/benchmark_json.py`: 50,000 local synthetic rows, five repetitions, no downloads. On macOS arm64 / Python 3.13 / PyArrow 25, indexed cases improve from 0.38–0.40 s to 0.23–0.24 s (1.62–1.72x). Unindexed cases remain about 0.14 s. Output SHA-256 matches before and after in all eight cases. The benchmark script is included for reproduction; these measurements describe this workload, not a universal speedup.

Developed and reviewed with OpenAI Codex assistance.


src/datasets/io/json.py
```diff
@@ -131,6 +132,14 @@ def _batch_json(self, args):
             key=slice(offset, offset + self.batch_size),
             indices=self.dataset._indices,
         )
+        if self.dataset._indices is not None:
+            # Gathering creates one-row chunks, which are expensive to convert to pandas.
+            # Combine only the selected batch, not the potentially much larger source table.
+            try:
+                batch = batch.combine_chunks()
+            except (pa.ArrowInvalid, pa.ArrowNotImplementedError):
```
ebarkhordar: I ran this against `c11cefc5` in a clean `python:3.11-slim` container on pyarrow 25.0.1.

The fallback is the one part I could not confirm. `test_dataset_to_json_when_chunks_cannot_be_combined` is parametrized over `pa.ArrowInvalid` and `pa.ArrowNotImplementedError`, the two classes this `except` names, and it reaches them by replacing `query_table` with a stub whose `combine_chunks` raises. That shows the handler catches what it says it catches. It cannot show the list is the right one.

I tried to build the case the comment describes and could not. Five chunked shapes, each a two-chunk table, all combined without raising:

```
pyarrow 25.0.1

positive control, same harness:
  concat_arrays_mismatched_types  RAISED ArrowInvalid
  cast_string_to_int              RAISED ArrowInvalid

pa.table({"c": chunked}).combine_chunks():
  dictionary_differing_dicts      OK (no raise)
  large_list                      OK (no raise)
  map                             OK (no raise)
  fixed_size_list                 OK (no raise)
  struct_nested_list              OK (no raise)
```

Which shape did you see this fail on, and what did it raise? `ArrowCapacityError` and `ArrowMemoryError` are not subclasses of either caught class (`ArrowCapacityError.__mro__` is `ArrowCapacityError, ArrowException, Exception`), so if the real failure is one of those the fallback does not engage, and `main` has no `combine_chunks` call at all today.

Five shapes on one pyarrow version is not an exhaustive search, so this is a question rather than a claim that the catch is wrong.

aprylewu: Good catch — the original comment was too vague, and Capacity/Memory were indeed outside the except list.

I dug into this on pyarrow 25.0.1. `query_table` returns a `pa.Table`, so this is `Table.combine_chunks()`.

The shape that actually raises `ArrowInvalid` is nested string/binary data whose total payload across chunks exceeds the 32-bit offset limit (~2 GiB), e.g. `struct<string>` or `list<string>`:

```text
ArrowInvalid: offset overflow while concatenating arrays
```

Flat `string`/`binary` columns don't hit this on `Table.combine_chunks()` — Arrow keeps multiple chunks instead. Nested ones still go through concatenation that overflows.

I couldn't reproduce `ArrowNotImplementedError` on 25.0.1 (dict/map/large_list/union/REE/Array2DExtension all combined fine). That one was real historically for extension concat on pyarrow <12 (datasets Array2D); we still keep it since it's harmless. Capacity/Memory are not subclasses of Invalid/NotImplemented, as you said.

Pushed a follow-up that:
- catches `ArrowCapacityError` / `ArrowMemoryError` as well
- rewrites the comment to match the above
- keeps the stub test for all four exceptions
- adds an opt-in real overflow test (`RUN_LARGE_ARROW_TESTS=1`) that builds 24×100MiB `struct<string>` chunks and checks both the raise and that `to_json` still succeeds via the fallback

ebarkhordar: Thanks, that is the shape I could not build. I ran the new opt-in test at `986d6487` in a clean `python:3.11-slim` container on pyarrow 25.0.1, and it fails before it reaches the fallback:

```
$ RUN_LARGE_ARROW_TESTS=1 python -m pytest tests/io/test_json.py -k offset_overflow -q
tests/io/test_json.py:396: in test_dataset_to_json_nested_string_offset_overflow
    dataset = Dataset(table).select(list(reversed(range(24))))
src/datasets/arrow_dataset.py:763: in __init__
    self._fingerprint = generate_fingerprint(self)
src/datasets/fingerprint.py:242: in generate_fingerprint
    hasher.update(state[key])
src/datasets/utils/_dill.py:163: in _save_arrowTable
    args = (obj.schema, [column.combine_chunks() for column in obj.columns])
E   pyarrow.lib.ArrowInvalid: offset overflow while concatenating arrays
1 failed
```

`Dataset(table)` with no fingerprint hashes the table, and `_save_arrowTable` combines every column so the hash does not depend on chunking (#8327). That call is unguarded, so it hits the same overflow at construction time.

Giving the constructor a fingerprint skips the hash, and the rest of the test then does what you describe. Same container, same head, test body otherwise unchanged:

```
Dataset(table, fingerprint="probe")
COMBINE: ArrowInvalid: offset overflow while concatenating arrays
written 2516582952  stat 2516582952  gt 2**31: True
rows 24 == len(ds): True
```

So the premise reproduces and the fallback works on it. The widened except list holds up as well: `ArrowCapacityError` is not a subclass of `ArrowInvalid`, and `ArrowMemoryError` subclasses `MemoryError`, so neither of the two you added is covered by the other pair.

I have not looked at whether `_save_arrowTable` should tolerate the overflow itself.

aprylewu: Yeah, Dataset(table) was hashing through combine_chunks before we even got to the export path. Fixed the opt-in test to pass a fingerprint so it hits the fallback. Leaving `_save_arrowTable` alone for now.

aprylewu:
If that looks good on your side, mind merging when you get a chance? 🙂

src/datasets/io/json.py
ebarkhordar: Confirmed at `8a3403cb`. Clean `python:3.11-slim` container, pyarrow 25.0.1:

```
$ RUN_LARGE_ARROW_TESTS=1 python -m pytest tests/io/test_json.py -k offset_overflow -q
1 passed, 65 deselected in 103.52s

# same command at 986d6487, the parent commit
1 failed, 65 deselected in 1.92s
E   pyarrow.lib.ArrowInvalid: offset overflow while concatenating arrays
    src/datasets/utils/_dill.py:163: in <listcomp>
```

So the fingerprint argument is what carries it, and the export path is unchanged between the two.

I do not have commit rights here, so a maintainer will need to merge. One thing worth flagging for whoever picks it up: the two workflow runs on this head are sitting at `action_required` and have not started, so there is no CI result on this PR yet.

